PR #23, opened 2014-12-23
added policy statement for S3ReadWriteResource

ank:
The S3ReadWriteResource only allowed operations on the bucket and not on the objects in the bucket. Added a second policy statement to allow put/get/delete on objects in the bucket.


ank:
Thanks for clarifying. You're right. The single statement with the /* in the parameter will give read/write access to the bucket contents. I did need a second statement to allow ListBucket on the bucket so that `aws s3 ls <bucket>` works but I imagine that's an edge case. Closing this as it's unnecessary.

Issue #35, opened 2015-06-30
SGE 'pending jobs' count - ignore held jobs


dougalb:
Hi,

Firstly, thanks Dougal for your efforts with cfn-cluster, it's great. 
I think I've spotted a small issue with the pending jobs count for the sge scheduler. Currently this is calculated as follows:

pending=$(qstat -g d -s p -u '*' | tail -n+3 | awk '{total = total+ $8}END{print total}')

The problem is, this doesn't take into account whether jobs are held pending the completion of other job dependencies (their state is 'hqw' rather than 'qw'). Obviously held jobs don't want to be included in the pending count for scaling purposes since they won't run even if there are slots available. 

The following seems to be a quick fix (*caveat* I know as much about awk as a fish knows about cycling). 

pending=$(qstat -g d -s p -u '*' | tail -n+3 | awk '$5 == "qw" {total = total+ $8} END {print total}')

Hope that's helpful.

Cheers,

Mike

kdaily:
This also would be impacted by 'Eqw' status (those that have an error).

dougalb:
This was merged into develop with #54 and will be rolled into next AMI.

PR #54, opened 2015-07-21
Only count pending jobs with status 'qw'

kdaily:
Fixes Issue #35


dougalb:
Thanks for the fix. Merging into develop and will plan on spinning an AMI shortly.

Issue #67, opened 2015-10-29
Race condition when nodes terminate (SGE)


kgutwin:
There is a race condition in the nodewatcher-sqswatcher system for scaling down. Normal termination flow is:

1. Nodewatcher determines node is eligible for termination
1. Nodewatcher sends termination request to Auto Scaling
1. Auto Scaling terminates the node and sends a notification to SNS
1. SNS puts a message on the SQS queue
1. Sqswatcher on the master receives the message
1. Sqswatcher removes the node from SGE queues and execute host list

The problem happens when a job is submitted and scheduled onto the node any time after step 1 and before step 6. Grid Engine will fail to remove the node from the queue if jobs are scheduled onto that node, even if the node is no longer responding to requests (and the jobs are still in 'transferring' state). If the node fails to remove, Grid Engine will continue to try to send jobs to it.

My suggestion to help resolve this is to, immediately after step 1 above, the node disables its own queue with the command `qmod -d all.q@<node>`. This shortens the race condition window by preventing jobs from being scheduled during the remaining steps. It does not eliminate the race condition entirely.

dougalb:
Thanks for logging the issue and providing some code to fix. I will review and work this into the next release.

digout:
I fear the same race condition might occur on openlava as well.
In case of openlava after step 1 the node should be closed (badmin hclose <node>) so no jobs will be started on the node.

dougalb:
All schedulers have had the same condition addressed. In the next release,
it will be deployed.

szs8:
When is the next release due?

dougalb:
It has been included in cfncluster-node==0.0.7 which is now part of the upcoming cfncluster=1.0 release, which is in the develop branch now and actively being tested. If you wish to test, I can share the link to the beta python sdist.

dougalb:
Released in d14a9fcaec241fd3d87a0ca00e34c22eea0f4187

PR #68, opened 2015-10-30
Avoid race conditions when nodes terminate

kgutwin:
Addresses #67. I attempted to put in reasonable lock commands for all schedulers, but I only have significant experience with SGE, so the other schedulers may need to be tweaked. Also, I have yet to rigorously test this code, so please test and confirm before merging. Thanks!


dougalb:
This has been pulled into the current develop branch.

Issue #57, opened 2015-07-23
master_subnet_id shows up twice in the docs


cancan101:
* https://github.com/awslabs/cfncluster/blob/da3aa2e2e3840f8bd69971168ebd17a10db15f66/docs/source/configuration.rst#master_subnet_id
* https://github.com/awslabs/cfncluster/blob/da3aa2e2e3840f8bd69971168ebd17a10db15f66/docs/source/configuration.rst#master_subnet_id-1

dougalb:
A pull request with a fix is typically the quickest and community approach for fixing doc errors.

PR #70, opened 2015-11-06
Fix compute_subnet_id reference in docs, see #57

kgutwin:
This was bugging me too, when I was looking to run master and compute in different subnets.


dougalb:
Done. Thanks.

PR #71, opened 2015-11-09
Add compute nodes as SGE submit hosts

kgutwin:
We have at least one use case where SGE jobs make `qsub` calls to submit additional jobs to the cluster. This PR adds the "submit host" capability to cfncluster compute nodes.


dougalb:
This has been pulled into the latest develop branch.

PR #72, opened 2015-11-13
Improve logging of nodewatcher and sqswatcher

kgutwin:
When debugging problems with either nodewatcher or sqswatcher, the previous logging system (print statements mostly) left some things to be desired. This PR switches logging to use the common Python logging framework. Log level can be adjusted by an optional `loglevel=` setting in `nodewatcher.cfg` and `sqswatcher.cfg`.

### DIFF
--- a/node/nodewatcher/nodewatcher.py
+++ b/node/nodewatcher/nodewatcher.py
@@ -25,10 +25,6 @@ import sys
 import tempfile
 import logging
 
-logging.basicConfig(
-    level=logging.INFO,
-    format='%(asctime)s %(levelname)s [%(module)s:%(funcName)s] %(message)s'
-)
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


node/nodewatcher/nodewatcher.py
```diff
@@ -23,12 +23,22 @@
 import time
 import sys
 import tempfile
+import logging
+
+logging.basicConfig(
```
cancan101: This should probably not be at module level.

node/sqswatcher/sqswatcher.py
```diff
@@ -27,23 +28,38 @@
 from boto.dynamodb2.fields import HashKey
 from boto.dynamodb2.table import Table
 
+logging.basicConfig(
```
cancan101: ditto

PR #75, opened 2015-11-30
Slurm support

mike-jt79:
Initial support for Slurm scheduler. Adds slurm scripts and templates to bootstrap and the required plugins for sqswatcher and nodewatcher.  Slurm support for dynamic nodes isn't great and so we use a dummy node in the slurm.conf to allow submission of jobs larger than the current cluster size. Also the process of adding nodes relies on the syntax of the slurm.conf - not a great solution and I will think about making it more elegant. 

__Note__
When adding updated boostrap and cfncluster-node packages to the AMI you also need to update the slurm binaries (/opt/slurm). The version in the current AMI is 14.11.2 and there seems to be an issue with the "squeue" command causing a segmentation fault, updating to 14.11.9 resolves this. The cluster will come up with the older binaries but we use the squeue command as part of nodewatcher, this will fail.


dougalb:
Hi Mike, thank you very much for your commit. We recently made the switch to Chef for the installation and bootstrapping of CfnCluster. I will review your PR and see what we can pull in.

mike-jt79:
Hi Dougal, ok no problem - happy to wait for the first release of the chef version and then I can look at integrating Slurm with that.

dougalb:
I have pulled the changes and will also add some of the code to the https://github.com/awslabs/cfncluster-cookbook project.

PR #83, opened 2016-02-19
fixed bug in cloudformation template preventing extra_json from being provided to chef cookbooks

clausman:
command was always blowing away /etc/chef/dna.json with /tmp/dna.jsaon, causing /tmp/extra.json to be ignored.

### DIFF
--- a/cloudformation/cfncluster.cfn.json
+++ b/cloudformation/cfncluster.cfn.json
@@ -1952,7 +1952,7 @@
                 "command" : "touch /etc/chef/ohai/hints/ec2.json"
               },
               "jq" : {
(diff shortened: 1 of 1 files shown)


dougalb:
The behaviour you describe is expected. The `/tmp/dna.json` is built from the CloudFormation parameters and resources and `/tmp/extra.json` is filled from the `extra_json` parameter in the CfnCluster config file http://cfncluster.readthedocs.org/en/latest/configuration.html#extra-json. 

Can you provide more details on what is failing specifically?

dougalb:
Ah yes, I see it now. Is it possible you can make the PR against the develop branch?

PR #125, opened 2016-08-02
Add missing parameters to cfnconfig

kgutwin:
The CustomChefCookbook and CustomChefRunList parameters are present in the stock cfncluster.cfn.json, but are not pulled in from the cfncluster config file. This PR adds the missing arguments.


dougalb:
These are already in the develop branch and will be part of the next release.

https://github.com/awslabs/cfncluster/blob/develop/cli/cfncluster/cfnconfig.py#L213

kgutwin:
I suppose I should have checked there before I submitted the PR :) thanks!

PR #175, opened 2017-03-17
I3 and R4 support

johnbot1:
Added i3 and R4 support.


johnbot1:
just realized that instances with NVMe have difference mount points for the ephemeral so need to update that logic. /dev/nvme0n1

bwbarrett:
We added R4 and I3 with the other new instance types, so this PR is now redundant.  Sorry about ignoring the PR; we're going through some internal org changes and missed that this PR already existed.

PR #187, opened 2017-06-01
Small fixes to improve HTTP config sanity check error messages.

dougalb:
Updates the error message when config_sanity.py is checking URL based resources i.e. the source of the 403 message.

**Before:**
$ cfncluster create foo
Starting: foo
403

**After:**
$ cfncluster create foo
Starting: foo
Config sanity error: https://s3.amazonaws.com/dougalb/cfncluster/cfncluster.cfn.yaml2 403 Forbidden


cgibioinformatics:
hi I'm also getting this message on my pre-install script that's on S3. What was the resolution? thanks

PR #191, opened 2017-06-21
Version update, contrib file, logging, config permissions

cgorski:


cgorski:
I think that we should leave the auth info as is for now, as I don't see any obvious risks. However, I believe that we should encourage, add (if not already available) and make obvious the suggestion that users calling the CLI from AWS should be using role-based auth on an role-assigned EC2 instance.

PR #212, opened 2017-08-08
Add a new CLI option to stop only the ComputeFleet.

rajachan:
This commit adds a new option, --compute-only/-c, to stop a cluster while
leaving the MasterServer intact. This would come in handy when a user wants to
use the MasterServer for compilation, development, or when debugging core files
and such.

```
usage: cfncluster stop [-h] [--compute-only] cluster_name

positional arguments:
  cluster_name        stop a cfncluster with the provided name.

optional arguments:
  -h, --help          show this help message and exit
  --compute-only, -c  Stop the ComputeFleet, but leave the MasterServer
                      running for debugging/development.
```

### DIFF
--- a/cli/cfncluster/cfncluster.py
+++ b/cli/cfncluster/cfncluster.py
@@ -214,6 +214,10 @@ def stop(args):
     asg = get_asg(stack_name=stack_name, config=config)
     set_asg_limits(asg=asg, min=0, max=0, desired=0)
 
+    # Skip the Master shutdown if requested by the user
+    if args.compute_only:
+        return
+
     # Stop master Server
     master_server_id = get_master_server_id(stack_name, config)
     ec2conn = boto.ec2.connect_to_region(config.region,aws_access_key_id=config.aws_access_key_id,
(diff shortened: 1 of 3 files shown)


cli/cfncluster/cfncluster.py
```diff
@@ -215,24 +215,25 @@ def stop(args):
...
-    master_server_id = get_master_server_id(stack_name, config)
-    ec2conn = boto.ec2.connect_to_region(config.region,aws_access_key_id=config.aws_access_key_id,
-                                                    aws_secret_access_key=config.aws_secret_access_key)
-    try:
-        response = ec2conn.stop_instances(master_server_id)
-    except boto.exception.BotoServerError as e:
-        if e.message.endswith("does not exist"):
-            print(e.message)
-            sys.stdout.flush()
+    if not args.compute_only:
```
mohanasudhan: Suggestion:

if args.compute_only:
     return

 will improve the readability

rajachan: Fair enough, I will update the PR.

rajachan: PR updated with this change now.

PR #224, opened 2017-09-05
Fix: G2.8xLarge does not support EBS-optimized launch

mohanasudhan:
cfncluster create fails if the instance is g2.8xlarge because
of EBS-optimized launch is not supported in this type


mohanasudhan:
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html#ebs-optimization-support

PR #240, opened 2017-10-18
Cfncluster to support new instance type by default

mohanasudhan:
Cloudformation template till now needs an update to support
new instance type. Instance type need to be updated in
list of allowed instance types and on a map which contains
capability of instance (EbsOpt & HVM). Following change
will allow all instance type by default (no restriction on
instance type) and revised capability logic as all instance
type are 64HVM (we also don't use this parameter anymore)
and include logic on Non-EbsOpt instance alone (as
future instances will fall under EbsOpt).

Tested cfn template with cc & c4 type both looks fine. 

Notes: In list of allowed values I see i2.large and t2.nano missing comparing map list. Let me know if we want to block those instances from being created. Apart from that changes looks fine in my end.

### DIFF
--- a/cloudformation/cfncluster.cfn.json
+++ b/cloudformation/cfncluster.cfn.json
@@ -709,147 +709,13 @@
       "Description" : "MasterServer EC2 instance type",
       "Type" : "String",
       "Default" : "t2.micro",
-      "ConstraintDescription" : "Must be a valid EC2 instance type, with support for HVM.",
-      "AllowedValues" : [
-        "cc2.8xlarge",
-        "c3.8xlarge",
-        "c3.4xlarge",
-        "c3.2xlarge",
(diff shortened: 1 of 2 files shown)


mohanasudhan:
side note: Fn:Or takes only 2-10 conditions, hence need to split it with batch of 10

bwbarrett:
This works, but any reason that we couldn't define a map of instances that don't support EBSOptimized and then use Fn::FindInMap?  Seems like it would be a bit easier to read.

bwbarrett:
Even better if we did that would be to have two maps, one that was instance families that don't support EBS optimized (t2, for example), and one that's instance types that don't (c3.8xlarge, for example), so that we don't have to push an update next time a new t2 size is released (should that happen).

Issue #176, opened 2017-03-22
Multiple documentation issues & missing IAM permissions


chrisdag:
Creating an issue here but I will also try to improve my git-fu so that I can submit a proper pull request.

The following documentation issues were encountered when I created a fresh IAM user, configured IAM policies for the user and Ec2-IntanceRole and then tried to make an HPC stack using a standalone config file that contained no default entries.

This is what I ran into:

1.  "configuration" page aka configuration.rst has a syntax error in the Tag section. The example provided showing how to add tags has broken syntax. Instead of `tags = {"key": "value", "key2", "value2"}` the proper format is `tags = {"key" : "value", "key2" : "value2"}`

2. "configuration" page has outdated values for the size of volumes used by the master and compute nodes. The doc page says "10" is the default but if you try to launch a cluster with these values you get an error back saying the minimum value for both master and compute root disk is now 15

3. The suggested IAM policy for the CfnCluster user is wildly incorrect and will fail instantly out of the box. There are missing IAM permissions required to launch the stack and MANY missing permissions that enable a clean rollback and delete. This is what I had to add to my user IAM policy in order to get past the create and delete/rollback phases:

`
Add To EC2 Describe:
"ec2:DescribeAvailabilityZones"

Add To EC2Modify:
 “ec2:DeleteNetworkInterface”
 “ec2:DeleteVolume”
 "ec2:TerminateInstances"

Add To SNS Describe:
  “sns:GetTopicAttributes"

Add To SNS Modify:
 “sns:DeleteTopic”

Add To DynamoDBModify:
  “dynamodb:DeleteTable"

Add To SQS Modify:
 “sqs:DeleteQueue”

`

chrisdag:
Update:

Found a new missing IAM permission

You can't "cfncluster delete" a cluster with a live computeFleet unless you also have:

   autoscaling:DeleteLaunchConfiguration

... permission

chrisdag:
More missing IAM permissions required for cfncluster IAM user to delete a cluster:

autoscaling:DescribeScalingActivities
autoscaling:DeleteAutoScalingGroup

mohanasudhan:
Issue 1 & 2 are already fixed. https://github.com/awslabs/cfncluster/pull/242 for iam policy update.

Issue #144, opened 2016-10-17
Accessing the cloud formation templates


kegunder:
I'm having issues accessing the latest cloudformation templates (i.e. templateUrl).  I have only days worth of s3 experience, so I'm having a hard time finding the source of the issue.

When I use boto3, I'm able to download and review the template for cfnCluster v1.2.1:

`s3.meta.client.download_file('cfncluster-us-west-2', 'templates/cfncluster-1.2.1.cfn.json' , 'cfncluster2.json')`

However, finding a 1.3.1 (or 1.3.0) version isn't possible do to:
ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden

Please explain if this is intended behavior (or point to relevant section in the docs.)  I'm not seeing, or not understanding the details that I see here: http://cfncluster.readthedocs.io/en/latest/configuration.html

Thanks!

rvinaybharadwaj:
Similarly confused regarding the 1.3.2 template within the east-1 region. The default template selection appears to fail: 

`TemplateURL must reference a valid S3 object to which you have access.`

Specifying the [template url ](https://s3.amazonaws.com/cfncluster-us-east-1/templates/cfncluster-1.3.2.cfn.json)directly also fails:

`Config sanity error: https://s3.amazonaws.com/cfncluster-us-east-1/templates/cfncluster-1.3.2.cfn.json 403 Forbidden`

bwbarrett:
@rvinaybharadwaj, i'm afraid it's a bit hard to determine what's going on without more information (like, at least a snippet of your config file).  Were you trying to use the CfnCluster default template in the second test, or a personal version?  I ask because the CfnCluster version is at https://s3.amazonaws.com/cfncluster-resources-us-east-1/templates/cfncluster-1.3.2.cfn.json, not https://s3.amazonaws.com/cfncluster-us-east-1/templates/cfncluster-1.3.2.cfn.json.

rvinaybharadwaj:
We have tried multiple solutions, including specifying the template URL ourselves. For sanity purposes we just generated a configuration with cfncluster config

[aws]
aws_region_name = us-east-1
aws_access_key_id = xxx
aws_secret_access_key = xxxx

[cluster xxxx]
vpc_settings = xxxx
key_name = xxxxx

[vpc xxxx]
master_subnet_id = xxx
vpc_id = xxx

[global]
update_check = true
sanity_check = true
cluster_template = xxxx

And we still receive the same error

[...]

PR #242, opened 2017-10-20
Fix: Missing iam permission in policy doc

mohanasudhan:
Updating iam policy doc file as per issue #176 

Signed-off-by: Mohan Gandhi mohgan@amazon.com

### DIFF
--- a/docs/source/iam.rst
+++ b/docs/source/iam.rst
@@ -165,8 +165,12 @@ CfnClusterUserPolicy
                   "ec2:ModifyVolumeAttribute",
                   "ec2:ModifyNetworkInterfaceAttribute",
                   "ec2:DeleteNetworkInterface",
-	          "ec2:DeleteVolume",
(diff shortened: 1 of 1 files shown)


mohanasudhan:
As part of issue #144 I tested latest template (was missing couple of delete policy which I updated now). I can create and delete cfncluster using latest policy. In here I create a custom iam policy (policy statement as in PR) and have attached it to my cfn-user (user doesn't have any other permission). Obviously using cfn-user credential in cfncluster config. 

Going forward I plan to use only this cfn-user for my future testing, so in case if we see any issue we can fix faster.

cfncluster --config ~/.cfncluster/config-test-policy --nowait create tester-s3 --norollback
Starting: tester-s3
Status: CREATE_IN_PROGRESS

cfncluster --config ~/.cfncluster/config-test-policy delete tester-s3 
Deleting: tester-s3
Status: ComputeFleet - DELETE_IN_PROGRESS

PR #265, opened 2017-12-01
docs: Removing openlava from valid scheduler options

mohanasudhan:


rajachan:
We should take it out of the CloudFormation allowed values too then.

mohanasudhan:
Oops !!! Ya will remove from their too.

PR #274, opened 2017-12-06
post-release: Update cfncluster and cookbook version + changelogs

mohanasudhan:
As we released cfncluster-1.4.1, updating to next version
as post-release process. Missed updating changelogs,
hence adding it as well
 

cloudformation/cfncluster.cfn.yaml
```diff
@@ -980,8 +980,8 @@ Mappings:
       RootDevice: /dev/sda1
   CfnClusterVersions:
     default:
-      cfncluster: cfncluster-1.4.1
-      cookbook: cfncluster-cookbook-1.4.0
+      cfncluster: cfncluster-1.4.2
+      cookbook: cfncluster-cookbook-1.4.1
```
bwbarrett: why are we bumping this before we release cookbook 1.4.1?  That seems wrong...

PR #283, opened 2018-01-04
test: Add test to launch and test clusters

bwbarrett:
Add a pre-release verification test that launches the combination
of region, distro, and scheduler of tests.  Each test will
run a "parallel" job (it's sleep, but on two nodes) and cause a
scale up event, verifying that scale-up works.

We limit the number of tests per region to limit how hard we
push on various AWS limits (like instance count, EIPs,
DynamoDB table creation rate, etc.).  This does mean the test
takes many hours to run (likely 5 hours in the average case).

Signed-off-by: Brian Barrett <bbarrett@amazon.com>

### DIFF
--- a/tests/cluster-check.sh
+++ b/tests/cluster-check.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# Copyright 2018      Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy
+# of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+#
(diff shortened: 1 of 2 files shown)


bwbarrett:
This still needs some work (copyrights, comments, etc.) and needs a better way of handling the vpc-id and subnet-id map, but this is about what I'm thinking...

tests/cluster-check.sh
```diff
@@ -0,0 +1,118 @@
...
+EOF
+
+    chmod +x job1.sh job2.sh
+    rm -f job1.done job2.done
+
+    qsub -l nodes=2:ppn=2 ./job1.sh
+    qsub -l nodes=2:ppn=2 ./job2.sh
+
+    done=0
+    while test $done = 0 ; do
```
mohanasudhan: While loop in each schedulers can moved out of if/else block. can avoid duplicate code here.

tests/cluster-check.sh
```diff
@@ -0,0 +1,118 @@
...
+	fi
+    done
+
+elif test "$scheduler" = "sge" ; then
+    ppn=` qhost | grep 'lx-' | head -n 1 | sed -n -e 's/[^[:space:]]*[[:space:]]\+[^[:space:]]*[[:space:]]\+\([0-9]\+\).*/\1/p'`
+    count=$((ppn * 2))
+
+    cat > job1.sh <<EOF
+#!/bin/bash
+#$ -pe mpi $count
```
mohanasudhan: Not clear what we doing here.

PR #285, opened 2018-01-05
cli: Fix configure failure on first use

bwbarrett:
Commit 54926c4 exposed a bug in "cfncluster configure" which
caused configure to fail if the base directory for the config
file didn't exit (in the default case, ~/.cfncluster/).  This
wasn't a common problem prior to 54926c4, because logging was
saved to ~/.cfncluster/, and it *did* create the directory
properly.  So create the directory for the config file basename
and avoid the issue, regardless of logging behavior.

Signed-off-by: Brian Barrett <bbarrett@amazon.com>

### DIFF
--- a/cli/cfncluster/easyconfig.py
+++ b/cli/cfncluster/easyconfig.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 from __future__ import absolute_import
-# Copyright 2013-2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2013-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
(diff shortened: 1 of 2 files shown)


cli/cfncluster/easyconfig.py
```diff
@@ -167,6 +167,10 @@ def configure(args):
             if value is not None and key is not '__name__':
                 config.set(section['__name__'], key, value)
 
+    # ensure that the directory for the config file exists (because
+    # ~/.cfncluster is likely not to exist on first usage)
+    os.makedirs(os.path.dirname(config_file))
```
rajachan: I think os.mkdirs() raises an error when the directory exists already. We might get into that case if a user runs configure after initial bootstrap on the same node.

bwbarrett: Doh!  Good catch.

PR #305, opened 2018-01-29
Fix documentation to reflect correct behavior of publish_pending

rajachan:
The cron job publishes the count of nodes required and not the count of
pending jobs.

Signed-off-by: Raghu Raja <craghun@amazon.com>

### DIFF
--- a/docs/source/processes.rst
+++ b/docs/source/processes.rst
@@ -16,7 +16,10 @@ A cluster's lifecycle begins after it is created by a user.  Typically, this is 
 
 publish_pending_jobs
 --------------------
-Once a cluster is running, a cronjob owned by the root user will monitor the configured scheduler (SGE, Torque, Openlava, etc) and publish the number of pending jobs to CloudWatch.  This is the metric utilized by :ref:`Auto Scaling <auto_scaling>` to add more nodes to the cluster.
+Once a cluster is running, a cronjob owned by the root user will monitor the
+configured scheduler (SGE, Torque, Openlava, etc) and publish to CloudWatch the
+number of nodes required to schedule pending jobs. This is the metric utilized
+by :ref:`Auto Scaling <auto_scaling>` to add more nodes to the cluster.
 
 .. image:: images/publish_pending_jobs.png
     :align: center


rajachan:
I really should also s/publish_pending_jobs/publish_pending/ to reflect what is actually in the code, but the repository does not have the source for the image. Will leave the text be until the image gets fixed too.

dougalb:
It has always published the number of slots that are pending not nodes. Slots are a general concept and aligns with vCPUs.

rajachan:
Not what I am seeing, at least [with SLURM](https://github.com/awslabs/cfncluster-cookbook/blob/develop/templates/default/publish_pending.slurm.erb#L25).

From the man page:

> %D Number of nodes allocated to the job or the minimum number of nodes required by a pending job. The actual number of nodes allocated to a pending job may exceed this number if the job specified a node range count (e.g. minimum and maximum node counts) or the job specifies a processor count instead of a node count and the cluster contains nodes with varying processor counts. As a job is completing this number will reflect the current number of nodes allocated. (Valid for jobs only)

dougalb:
SLURM is the odd one out and honestly doesn’t follow convention compared to
SGE.

rajachan:
Turns out, SLURM's cron job doesn't even publish the number of nodes, but just the number of pending jobs.  Will fix that first before updating the doc to clarify this text. Let's leave this PR open for tracking.

bwbarrett:
@dougalb, ok, slots matches what the autoscaling logic does a bit better (scaling up 20 nodes for an alarm of 200), but does it make sense?  With the drastic slot count differences between a t2.micro and a c5.18xlarge, i'm not sure that assuming instances have about 10 slots makes sense.  It might make more sense to have the master node compute the number of instances needed and publish a node count instead.  It would require the compute nodes to tell the master node about slot configuration, which wouldn't be ideal, but would be better than where we are today.  Thoughts?

dougalb:
@bwbarrett All the pieces are there and just needed to be wired up :-)

When an instance boots, the Chef runs _prep_env.rb which sets the number of vCPUs on the instance as a config parameter - https://github.com/awslabs/cfncluster-cookbook/blob/e3f93c1396dbb223dd507b6388b0df89c12cb13b/recipes/_prep_env.rb#L17-L23. This is then written out as a config file - https://github.com/awslabs/cfncluster-cookbook/blob/develop/templates/default/cfnconfig.erb#L9. This config file is read by the compute_ready script which is run when an instance is ready to join the cluster - https://github.com/awslabs/cfncluster-cookbook/blob/develop/files/default/compute_ready#L8. That results in a message in the queue that contains the number of slots available on that instance, which is this then picked up by sqswatcher - https://github.com/awslabs/cfncluster-node/blob/develop/sqswatcher/sqswatcher.py#L147. It uses this to add the correct "slot" count for an instance to the scheduler. If you also persist this to DDB - https://github.com/awslabs/cfncluster-node/blob/develop/sqswatcher/sqswatcher.py#L149-L152, you could then query it when running publish_pending and you would know how many slots/vCPUs the instances are adding when then join. I think that should cover it.

bwbarrett:
I wonder if we shouldn't just air our dirty laundry and explain that SGE/Torque work one way and SLRUM works another.  We can say where we want to go as well, but I think it will be confusing if we don't explain how it works today.

dougalb:
@bwbarrett Agreed. It should be noted that SLURM was contributed to CfnCluster and not something originally developed as part of the first pass. SLURM is quite a different in how it works and that has resulted in a lot of challenges. Not sure it will ever work as well as SGE.

PR #306, opened 2018-01-30
Clarify ec2_iam_role parameter's input type.

rajachan:
Explicitly state that ec2_iam_role takes the name of a role (to imply not an arn).

Signed-off-by: Raghu Raja <craghun@amazon.com>


bwbarrett:
you can't pass an ARN, right?  I'd add an explicit clarification about what we mean by "name".

rajachan:
Yeah,I should have been more explicit. How's about the modified text?

PR #349, opened 2018-04-12
Support for resizing the shared drive snapshot

brianjbeach:
Updating CloudFormation templates to support resizing the shared drive when created from a snapshot. Previously the volume_size parameter was ignored when creating from a snapshot. With this change, the user can specify volume size that is larger than the original snapshot. If the user specifies a volume size that is smaller than the snapshot, the snapshot size is used instead. It does not cause an error. NOTE: There is a side effect to take into account when reviewing this PR.  If the user has created a snapshot that is smaller than 20GB and does not specify a VolumeSize, the default value of 20GB will be used.

Why is this important? I am working a WRF cluster that is run for different size simulations. We are running pIOPS and storage is a significant cost driver. During a normal day, we run many small (4 node) clusters every few hours. These simulations are low resolution and only need a small volume. Occasionally (e.g. when a storm is coming) we run a much larger simulation and need lots more storage. Managing multiple snapshots for various simulation sizes is a hassle and this would make life better.

### DIFF
--- a/cloudformation/cfncluster.cfn.json
+++ b/cloudformation/cfncluster.cfn.json
@@ -769,7 +769,7 @@
       "Default" : "NONE"
     },
     "VolumeSize" : {
-      "Description" : "Size of EBS volume in GB, if creating a new one",
+      "Description" : "Size of EBS volume in GB",
       "Type" : "Number",
       "Default" : "20"
     },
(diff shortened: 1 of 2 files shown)


bwbarrett:
This patch is a silent change in user behavior for many of the use cases, including the case called out in the commit log.  If we take this patch as is, the next release should have a major version number bump.  Given the likely ability to avoid those behavior changes (say, but using the snapshot size if an explicit size isn't given), it seems like doing more work on the patch is the right path forward.

Regardless, the docs need to be updated to reflect the final change in behavior.

PR #350, opened 2018-04-13
CfnCluster Login

sean-smith:
## Overview
Add cli command `cfncluster ssh mycluster` , this is defined in the `aliases` portion of the config file:

```
[aliases]
ssh = ssh {CFN_USER}@{MASTER_IP} {ARGS}
```

Before execution, the username and ip address are substituted in as well as any provided arguments.

For example:

```
cfncluster ssh mycluster -i ~/.ssh/id_rsa -v
```

Executes as (with username assumed to be `ec2-user` and ip address as `1.1.1.1`):

```
ssh ec2-user@1.1.1.1 -i ~/.ssh/id_rsa -v
```

### DIFF
--- a/cli/cfncluster/cfncluster.py
+++ b/cli/cfncluster/cfncluster.py
@@ -10,12 +10,12 @@ from __future__ import absolute_import
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
-
 from builtins import str
 import sys
 import time
 import logging
 import boto3
+import os
 from botocore.exceptions import ClientError
 
 from . import cfnconfig
@@ -277,6 +277,31 @@ def poll_master_server_state(stack_name, config):
 
     return state
 
+def get_master_server_ip(stack_name, config):
+    ec2 = boto3.client('ec2', region_name=config.region,
+                       aws_access_key_id=config.aws_access_key_id,
+                       aws_secret_access_key=config.aws_secret_access_key)
+
+    master_id = get_master_server_id(stack_name, config)
+
+    try:
+        instance = ec2.describe_instances(InstanceIds=[master_id]) \
+            .get('Reservations')[0] \
+            .get('Instances')[0]
+        ip_address = instance.get('PublicIpAddress')
+        state = instance.get('State').get('Name')
+        if state != 'running' or ip_address is None:
+            logger.info("MasterServer: %s\nCannot get ip address." % state.upper)
+            sys.exit(1)
+        return ip_address
+    except ClientError as e:
+        logger.critical(e.response.get('Error').get('Message'))
+        sys.stdout.flush()
+        sys.exit(1)
+    except KeyboardInterrupt:
+        logger.info('\nExiting...')
+        sys.exit(0)
+
 def get_ec2_instances(stack, config):
     cfn = boto3.client('cloudformation', region_name=config.region,
                        aws_access_key_id=config.aws_access_key_id,
@@ -345,6 +370,56 @@ def instances(args):
 
     for instance in instances:
         print('%s         %s' % (instance[0],instance[1]))
+
+def get_head_user(parameters, template):
+    mappings = template.get("TemplateBody") \
+            .get("Mappings") \
+            .get("OSFeatures")
+    base_os =[i.get('ParameterValue') for i in parameters if i.get('ParameterKey') == "BaseOS"][0]
+    return mappings.get(base_os).get("User")
+
+def command(args, extra_args):
+    stack = ('cfncluster-' + args.cluster_name)
+    config = cfnconfig.CfnClusterConfig(args)
+    if args.command in config.aliases:
+        config_command = config.aliases[args.command]
+    else:
+        config_command = "ssh {CFN_USER}@{MASTER_IP} {ARGS}"
+
+    cfn = boto3.client('cloudformation', region_name=config.region,
+                       aws_access_key_id=config.aws_access_key_id,
+                       aws_secret_access_key=config.aws_secret_access_key)
+    try:
+        status = cfn.describe_stacks(StackName=stack).get("Stacks")[0].get('StackStatus')
+        invalid_status = ['DELETE_COMPLETE', 'DELETE_IN_PROGRESS']
+        if status in invalid_status:
+            logger.info("Stack status: %s. Cannot SSH while in %s" % (status, ' or '.join(invalid_status)))
+            sys.exit(1)
+        ip = get_master_server_ip(stack, config)
+        stack_result = cfn.describe_stacks(StackName=stack).get('Stacks')[0]
+        template = cfn.get_template(StackName=stack)
+        username = get_head_user(stack_result.get('Parameters'), template)
+
+        try:
+            from shlex import quote as cmd_quote
+        except ImportError:
+            from pipes import quote as cmd_quote
+
+        # build command
+        cmd = config_command.format(CFN_USER=username, MASTER_IP=ip, ARGS=' '.join(cmd_quote(str(e)) for e in extra_args))
+
+        # run command
+        if not args.dryrun:
+            os.system(cmd)
+        else:
+            logger.info(cmd)
+    except ClientError as e:
+            logger.critical(e.response.get('Error').get('Message'))
+            sys.stdout.flush()
+            sys.exit(1)
+    except KeyboardInterrupt:
+        logger.info('\nExiting...')
+        sys.exit(0)
 
 def status(args):
     stack = ('cfncluster-' + args.cluster_name)
@@ -439,4 +514,3 @@ def delete(args):
     except KeyboardInterrupt:
         logger.info('\nExiting...')
         sys.exit(0)
-
--- a/cli/cfncluster/cfnconfig.py
+++ b/cli/cfncluster/cfnconfig.py
@@ -338,6 +338,13 @@ class CfnClusterConfig(object):
         except AttributeError:
             pass
 
+        # handle aliases
+        self.aliases = {}
+        self.__alias_section = 'aliases'
+        if __config.has_section(self.__alias_section):
+            for alias in __config.options(self.__alias_section):
+                self.aliases[alias] = __config.get(self.__alias_section, alias)
+
         # Handle extra parameters supplied on command-line
         try:
             if self.args.extra_parameters is not None:
--- a/cli/cfncluster/cli.py
+++ b/cli/cfncluster/cli.py
@@ -21,12 +21,14 @@ import errno
 from . import cfncluster
 from . import easyconfig
 
-
 def create(args):
     cfncluster.create(args)
 
 def configure(args):
     easyconfig.configure(args)
+
+def command(args, extra_args):
+    cfncluster.command(args, extra_args)
 
 def status(args):
     cfncluster.status(args)
@@ -146,7 +148,8 @@ def main():
     addarg_region(pstart)
     pstart.set_defaults(func=start)
 
-    pstop = subparsers.add_parser('stop', help='stop the compute-fleet, but leave the MasterServer running for debugging/development')
+    pstop = subparsers.add_parser('stop', help='stop the compute-fleet, but leave the MasterServer running for '
+                                               'debugging/development')
     pstop.add_argument("cluster_name", type=str, default=None,
                         help='stops the compute-fleet of the provided cluster name.')
     addarg_config(pstop)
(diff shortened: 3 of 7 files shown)


cli/cfncluster/cfncluster.py
```diff
@@ -403,6 +413,54 @@ def instances(args):
...
+    return username_map[base_os]
+
+def login(args):
+    stack = ('cfncluster-' + args.cluster_name)
+    config = cfnconfig.CfnClusterConfig(args)
+    cfnconn = boto.cloudformation.connect_to_region(config.region,aws_access_key_id=config.aws_access_key_id,
+                                                    aws_secret_access_key=config.aws_secret_access_key)
+    try:
+        status = cfnconn.describe_stacks(stack)[0].stack_status
+        if not ((status == 'CREATE_COMPLETE') or (status == 'UPDATE_COMPLETE')):
```
rajachan: How about using DescribeInstanceStatus or some other mechanism to just wait for the master node to come up before allowing logins instead of waiting for the entire cluster to be ready? It will be convenient to login to the master server and prepare my batch scripts/compile my code while my large compute ASG is spinning up.

cli/cfncluster/cli.py
```diff
@@ -21,13 +21,25 @@
 from . import cfncluster
 from . import easyconfig
 
+#
+# configuration
+#
+username_map = { 'alinux' : 'ec2-user',
```
rajachan: You don't need this map here, you aren't using it anyway.

docs/source/commands.rst
```diff
@@ -119,6 +119,25 @@ optional arguments:
...
+Attempts to ssh into head node. Stack needs to be in CREATE_COMPLETE or UPDATE_COMPLETE.
+Head node needs to be running. You can specify an ssh `key_path` yo use in your config or as a cli parameter.
+If you specify it as a cli parameter, this will override the value in the config.
+
+positional arguments:
+  cluster_name  ssh into head node
+
+optional arguments:
+  -h, --help        show this help message and exit
+  -s, --key-path    path to ssh private key
```
rajachan: The shorthand variant of the argument should start with the same character as that of the full argument. Either go for `-k`, or follow the same convention as the ssh command and take the identity file with `-i`.  I'd clarify the help text too, and explicitly say key file instead of just path.  Path would typically refer to the path that contains the file, so `~/.ssh/` instead of ~/.ssh/my_priv_key_file`.

cli/cfncluster/cli.py
sean-smith: Removed. Actually I can get rid of the entire dict by using `get_template()` and looking at the mappings section: 

```
"Mappings" : {
    "AWSRegionOS2AMI" : {
     ...
    },
    "OSFeatures" : {
      "centos6" : {
        "User" : "centos",
        "RootDevice" : "/dev/sda1"
      },
      "centos7" : {
        "User" : "centos",
        "RootDevice" : "/dev/sda1"
      },
      "alinux" : {
        "User" : "ec2-user",
        "RootDevice" : "/dev/xvda"
      },
      "ubuntu1404" : {
        "User" : "ubuntu",
        "RootDevice" : "/dev/sda1"
      },
      "ubuntu1604" : {
        "User" : "ubuntu",
        "RootDevice" : "/dev/sda1"
      }
    },
```

cli/cfncluster/cfncluster.py
```diff
@@ -24,6 +24,15 @@
 
 from . import cfnconfig
 
+#
+# configuration
+#
+username_map = { 'alinux' : 'ec2-user',
```
mohanasudhan: This information is also available in cfn template. Can we have util function to figure mapping from template ?

cli/cfncluster/cfncluster.py
mohanasudhan: Just a thought,  using DescribeInstanceStatus is good idea, just want to make sure associate-address happened w.r.t eip

cli/cfncluster/cfncluster.py
```diff
@@ -403,6 +413,54 @@ def instances(args):
...
+        state = poll_master_server_state(stack, config, args.nowait)
+        if state != 'running':
+            logger.info("MasterServer: %s. Needs to be running to login." % (state.upper))
+            sys.exit(1)
+        key_path = ""
+        if config.key_path is not None:
+            key_path = "-i %s" % config.key_path
+        if args.key_path is not None:
+            key_path = "-i %s" % args.key_path
+        stack_result = cfnconn.describe_stacks(stack)[0]
```
mohanasudhan: In case key_path is empty then throw error message

cli/cfncluster/cfnconfig.py
```diff
@@ -145,6 +145,19 @@ def __init__(self, args):
             sys.exit(1)
         self.parameters.append(('KeyName', self.key_name))
 
+        # Get the EC2 key path
+        try:
+            self.key_path = __config.get(self.__cluster_section, 'key_path')
```
mohanasudhan: so we are making key_path mandatory field in config file. so when customer upgrade, they need to update there config file ? I don't recommend making it mandatory in config file. I have like 20 config file, as a customer i need to go and update in all those config (not a good customer experience even if i don't want to use login option)

cli/cfncluster/cfncluster.py
sean-smith: The idea is to make it optional for customers who have their keys in the default `~/.ssh` folder

cli/cfncluster/cfncluster.py
sean-smith: Yes. See above comment

docs/source/commands.rst
sean-smith: Implementing `ssh` flags is a slippery slope. Ideally there's a way for the customer to pass arbitrary flags and commands that'll just get run.

cli/cfncluster/cfnconfig.py
sean-smith: It's optional in the config file

docs/source/commands.rst
rajachan: I'm not saying we should implement the ssh flags, that is definitely a slippery slope. My comment was merely about the naming of this login command's argument to take a key.

cli/cfncluster/cfncluster.py
sean-smith: I did not address this in the latest pr as I want to get the interface correct first.

docs/source/configuration.rst
```diff
@@ -79,6 +79,12 @@ Name of an existing EC2 KeyPair to enable SSH access to the instances. ::
 
     key_name = mykey
 
+identity_file
```
bwbarrett: I thought we didn't need this now?

docs/source/commands.rst
```diff
@@ -119,6 +119,40 @@ optional arguments:
...
+positional arguments:
+  cluster_name  name of the cluster to set variables for.
+
+optional arguments:
+  -h, --help    show this help message and exit
+  --dryrun, -d  print command and exit.
+
+::
+
+    $cfncluster ssh mycluster --args="-i ~/.ssh/id_rsa -v"
```
bwbarrett: do we have to use --args?  It would be more awesome if `cfncluster ssh mycluster -i ~/.ssh/id_rsa` just worked..

cli/cfncluster/config_sanity.py
```diff
@@ -20,6 +20,7 @@
 from urllib.parse import urlparse
 import boto.exception
 import sys
+import os
```
bwbarrett: still needed?

cli/cfncluster/cli.py
```diff
@@ -156,6 +159,29 @@ def main():
...
 
+    # Add ssh and scp
+    pssh = subparsers.add_parser('ssh', description='run ssh command with username and ip address pre-filled. '
+                                                    'This command may be customized in the aliases section of the config file.'
+                                                    'You may use: {CFN_USER} {MASTER_IP} variables in the command.')
+    pssh.add_argument("cluster_name", type=str, default=None,
+                        help='name of the cluster to set variables for.')
+    pssh.add_argument("--dryrun", "-d", action='store_true', dest="dryrun", default=False,
+                         help='print command and exit.')
+    pssh.add_argument("--args", type=str, dest="args", default="",
```
bwbarrett: `*nargs` instead?, or can you not use that on a subparser?

cli/cfncluster/cfncluster.py
```diff
@@ -251,7 +252,7 @@ def get_master_server_id(stack_name, config):
 
     return [r.physical_resource_id for r in resources if r.logical_resource_id == 'MasterServer'][0]
 
-def poll_master_server_state(stack_name, config):
+def poll_master_server_state(stack_name, config, nowait):
```
bwbarrett: i don't love what you've done to this function.  I think you need to refactor.  You're depending on nowait meaning "don't print", which is weird.  I'm all for reusing code, but maybe splitting it up a bit to handle both use cases would be better.

docs/source/configuration.rst
sean-smith: removed.

docs/source/commands.rst
sean-smith: I can do that. There's a function `parse_known_args()` that allows you to pass arbitrary args through, I can just append them to the command.

cli/cfncluster/config_sanity.py
sean-smith: removed.

cli/cfncluster/cli.py
sean-smith: I'm now passing through all args using `parse_known_args()` so it doesn't throw an exception for unknown arguments.

cli/cfncluster/cfncluster.py
```diff
@@ -403,6 +432,62 @@ def instances(args):
...
+        if status in invalid_status:
+            logger.info("Stack status: %s. Needs to be in %s" % (status, ' or '.join(invalid_status)))
+            sys.exit(1)
+        ip = get_master_server_ip(stack, config)
+        stack_result = cfnconn.describe_stacks(stack)[0]
+        template = cfnconn.get_template(stack)
+        username = get_head_user(stack_result.parameters, template)
+
+        # build command
+        cmd = config_command.format(CFN_USER=username, MASTER_IP=ip, ARGS=' '.join(str(e) for e in extra_args))
```
mauri-melato: Is this going to work if `extra_args` contains spaces, e.g. `-i "/path with spaces/to/pemfile"` ? 
I may be wrong but I'd expect the join would correctly manage the path above as a single param but final result for `ARGS` and so the command would be wrong.
It would be great to have a unit test for it.

cli/cfncluster/cfncluster.py
```diff
@@ -287,6 +292,30 @@ def poll_master_server_state(stack_name, config):
...
     return state
 
+def get_master_server_ip(stack_name, config):
+    ec2conn = boto.ec2.connect_to_region(config.region,aws_access_key_id=config.aws_access_key_id,
+                                                    aws_secret_access_key=config.aws_secret_access_key)
+
+    master_id = get_master_server_id(stack_name, config)
+
+    try:
+        instance = ec2conn.get_only_instances(instance_ids=[master_id])[0]
```
mauri-melato: This direct access to the first element of the list is an error condition that's worth to be checked, to avoid this kind of error to pop-up in front of the user. In this case it would mean we couldn't find the Master instance.

docs/source/commands.rst
```diff
@@ -123,6 +123,39 @@ optional arguments:
...
 
+ssh
+====
+
+Runs ssh to the master node, with username and ip filled in based on the provided cluster.
+
+For example:
+    cfncluster ssh mycluster -i ~/.ssh/id_rsa
+
+Results in an scp command with username and ip address pre-filled.
```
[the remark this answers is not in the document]
sean-smith: fixed.

cli/cfncluster/cli.py
```diff
@@ -154,6 +157,24 @@ def main():
     pversion = subparsers.add_parser('version', help='display version of cfncluster')
     pversion.set_defaults(func=version)
 
-    args = parser.parse_args()
+    # Add ssh and scp
+    pssh = subparsers.add_parser('ssh', description='run ssh command with username and ip address pre-filled. '
+                                                    'This command may be customized in the aliases section of the config file.'
+                                                    'You may use: {CFN_USER} {MASTER_IP} {ARGS} variables in the command.'
```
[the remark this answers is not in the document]
sean-smith: changed.

cli/cfncluster/cfncluster.py
```diff
@@ -346,6 +371,51 @@ def instances(args):
...
+        if status in invalid_status:
+            logger.info("Stack status: %s. Cannot SSH while in %s" % (status, ' or '.join(invalid_status)))
+            sys.exit(1)
+        ip = get_master_server_ip(stack, config)
+        stack_result = cfn.describe_stacks(StackName=stack).get('Stacks')[0]
+        template = cfn.get_template(StackName=stack)
+        username = get_head_user(stack_result.get('Parameters'), template)
+
+        # build command
+        cmd = config_command.format(CFN_USER=username, MASTER_IP=ip, ARGS=' '.join(str(e) for e in extra_args))
```
mauri-melato: I think `extra_args` with spaces, e.g. `-i "/path with spaces/to/pemfile"` is not going to work.
I think the python3 function `shlex.quote()` on the `str(e)` does the job.

sean-smith: fixed. `shlex.quote()` was introduced in 3.3 so I've added code to import either `pipes.quote` or `shlex.quote`

PR #354, opened 2018-04-27
Strip leading and trailing spaces from user's inputs

enricou:
*Issue:* If the user's input for the `cfncluster configure` command (e.g. VPC ID) contains trailing or leading spaces the next query for subnets fails with the message: _"No subnets found in region ..."_

*Description of changes:* The user input is sanitized removing both trailing and leading spaces and this makes the queries work in any case.


sean-smith:
I tested this with both python `2.6` and `3.6`. LGTM

PR #357, opened 2018-05-01
Change scale up threshold to 10

sean-smith:
## Background
The value was originally set to `200` because it referred to vcpus.
We've changed slurm and torque to look at number of nodes instead of vcpu's so
this value needs to change as well.

* Slurm https://github.com/awslabs/cfncluster-cookbook/pull/99
* Torque https://github.com/awslabs/cfncluster-cookbook/pull/102

### DIFF
--- a/cloudformation/cfncluster.cfn.json
+++ b/cloudformation/cfncluster.cfn.json
@@ -941,7 +941,7 @@
     "ScalingThreshold2" : {
       "Description" : "Threshold for triggering CloudWatch ScaleUp2 action",
       "Type" : "String",
-      "Default" : "200"
+      "Default" : "10"
     },
     "ScalingCooldown" : {
       "Description" : "Period in seconds to wait before allowing further scaling actions",


cloudformation/cfncluster.cfn.json
```diff
@@ -941,7 +941,7 @@
     "ScalingThreshold2" : {
       "Description" : "Threshold for triggering CloudWatch ScaleUp2 action",
       "Type" : "String",
-      "Default" : "200"
+      "Default" : "10"
```
mohanasudhan: Please modify cloudformation/cfncluster.cfn.yaml file when modifying json file. we haven't automated this process. we were thinking of having only one format in github and the release process should generate another format while pushing templates to s3 bucket.

lukeseawalker:
Can you please reflect this change in all the other references, for example:
- config template: 
`#scaling_threshold2 = 200`
- docs:
-- configuration.rst
 ```
Defaults to 200 for default template. ::

    scaling_threshold2 = 200
```
-- auto_scaling.png

PR #380, opened 2018-05-22
Add EC2 Describe Volumes Permission

sean-smith:
To fix a bug, we added logic to poll on the status of the EBS volume.
This requires "ec2:DescribeVolumes" permissions in the IAM role.
Since some customers provide their own role, I've added a warning in
the documentation about including the new permission.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,6 +28,10 @@ Please open a GitHub issue for any feedback or issues:
 https://github.com/awslabs/cfncluster.  There is also an active AWS
 HPC forum which may be helpful:https://forums.aws.amazon.com/forum.jspa?forumID=192.
 
+CfnCluster 1.5 IAM Change
+=========================
+Between CfnCluster 1.4.2 and 1.5.0 we made a change to the CfnClusterInstancePolicy that adds “ec2:DescribeVolumes” permissions. If you’re using a custom policy (e.g. you specify "ec2_iam_role" in your config) be sure it includes this new permission. See https://cfncluster.readthedocs.io/en/latest/iam.html
+
 CfnCluster 1.2 and Earlier
 ==========================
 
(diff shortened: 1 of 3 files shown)


docs/source/iam.rst
```diff
@@ -3,6 +3,10 @@
 IAM in CfnCluster
 ========================
 
+.. warning::
```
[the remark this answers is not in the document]
sean-smith: I'll move the warning. See https://cfncluster.readthedocs.io/en/latest/ami_customization.html for how the warning looks.

README.rst
```diff
@@ -28,6 +28,10 @@ Please open a GitHub issue for any feedback or issues:
 https://github.com/awslabs/cfncluster.  There is also an active AWS
 HPC forum which may be helpful:https://forums.aws.amazon.com/forum.jspa?forumID=192.
 
+CfnCluster 1.5 IAM Change
+=========================
+Between CfnCluster 1.4.2 and 1.5.0 we made a change to the CfnClusterInstancePolicy that adds “ec2:DescribeVolumes” permissions. If you’re using a custom policy (e.g. you specify "ec2_iam_role" in your config) you’ll need to make this change in your policy as well.
```
mauri-melato: I'd slightly reword it  to say that is not necessarily needed to change the policy if it already includes the required permissions, something like    "If you’re using a custom policy (...) be sure it already includes the new required permission or you’ll need to make this change in your policy as well"

docs/source/iam.rst
```diff
@@ -3,6 +3,10 @@
...
+    Between CfnCluster 1.4.2 and 1.5.0 we added a change to the `CfnClusterInstancePolicy` that adds "ec2:DescribeVolumes" permissions. If you're using a custom policy you'll need to make this change in your policy as well.
```
mauri-melato: same as comment above

README.rst
sean-smith: done

docs/source/iam.rst
sean-smith: ack

PR #383, opened 2018-05-23
Remove Cloudformation YAML template

enricou:
Signed-off-by: Enrico Usai <usai@amazon.com>


bwbarrett:
The YAML version was added because customers were asking for a yaml version to edit directly, rather than using the json vesion.  I can see both sides to this argument, but we should treat this as a major change and add a readme item and all that.

PR #391, opened 2018-05-31
Fix and improvements for the cfncluster-release-check script

enricou:
+ Fix indentation and trailing spaces 

+ Add python2 requirement in the introduction notes

+ Add `StrictHostKeyChecking=0` to avoid user confirmation during ssh and scp connections

+  Fix return value check for `cfncluster-release-check` script:
    +  `run_test()` doesn't return a value, so `retval` is None.
    +  _None_ means "no value" so the test `retval == 0` can give erratic behaviour.
    + We are going to ignore the return value of `run_test()`, because there is no one
        and set retval to 0 after the run if no Exception are raised.

+ Add input parameter for Key Pair name

### DIFF
--- a/tests/cfncluster-release-check.py
+++ b/tests/cfncluster-release-check.py
@@ -19,10 +19,12 @@
 # against limits in each region, the number of simultaneously built
 # clusters in each region is a configuration parameter.
 #
-# NOTE: To simplify this script, at least one subnet in every region
-# to be tested must have a resource tag named "CfnClusterTestSubnet"
-# (value does not matter).  That subnet will be used as the launch
-# target for the cluster.
+# NOTE:
+# - This script requires python2
+# - To simplify this script, at least one subnet in every region
+#   to be tested must have a resource tag named "CfnClusterTestSubnet"
+#   (value does not matter). That subnet will be used as the launch
+#   target for the cluster.
 
 import os
 import sys
--- a/tests/cfncluster-release-check.py
+++ b/tests/cfncluster-release-check.py
@@ -57,7 +57,7 @@ success = 0
 #
 # run a single test, possibly in parallel
 #
-def run_test(region, distro, scheduler):
+def run_test(region, distro, scheduler, key_name):
     testname = '%s-%s-%s' % (region, distro, scheduler)
     test_filename = "config-%s.cfg" % (testname)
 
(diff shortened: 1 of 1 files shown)


tests/cfncluster-release-check.py
```diff
@@ -19,10 +19,13 @@
 # against limits in each region, the number of simultaneously built
 # clusters in each region is a configuration parameter.
 #
-# NOTE: To simplify this script, at least one subnet in every region
-# to be tested must have a resource tag named "CfnClusterTestSubnet"
-# (value does not matter).  That subnet will be used as the launch
-# target for the cluster.
+# NOTE:
+# - This script requires python2
```
lukeseawalker: so it doesn't work with 3? What's the effort to let the script works also with 3?

enricou:
I tested it from a private EC2 instance, in a _python2_ virtual environment with _boto3_ installed.

I added the _CfnClusterTestSubnet_ tag for a Subnet and created a _cfncluster-test_ Key Pair, as described in the initial comment of the script itself.

Then I executed the script with: 
`cfncluster-release-check.py --regions eu-west-1 --distros alinux --schedulers sge`

tests/cfncluster-release-check.py
enricou: The scope of this PR should be just fix the script it to make it work.
We can upgrade the code to work with python3 by changing the _Queue_ and the _iteritems()_ stuff, but I think we can make it in a different patch.

tests/cfncluster-release-check.py
```diff
@@ -66,7 +69,7 @@ def run_test(region, distro, scheduler):
     file.write("aws_region_name = %s\n" % region)
     file.write("[cluster default]\n")
     file.write("vpc_settings = public\n")
-    file.write("key_name = default\n")
+    file.write("key_name = cfncluster-test\n")
```
bwbarrett: Doh; this was me being lazy.  Let's make this a command line argument instead of just changing the name.

tests/cfncluster-release-check.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #
-# Copyright 2018      Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
bwbarrett: Don't remove the 5 spaces for next year.  It just means more changes next year.

enricou: ok, I'm going to update the patch

enricou:
@bwbarrett Requested changes complete.

bwbarrett:
@enricou Thanks for all the changes!  Looks good to me.

PR #420, opened 2018-06-19
Add instance type parameters to cfncluster-release-check

enricou:
Now it's possible to pass a list of instance types to test.
They will be used for both master and compute nodes.
* `--instance-type` (default: `c4.xlarge`)

I also changed the stdout/stderr/config word from prefix to suffix to
have the files relative to the same test close to each other.


mauri-melato:
We're going to change the patch to have one single instance type parameter that supports a comma separated list of values. This will be used in the cycle as a further iteration 'dimension'

PR #427, opened 2018-06-21
CfnCluster Release Check script improvements

mauri-melato:
Tuning a more aggressive scale-up 
Fix Torque node allocation + job sleep tuning
Added ssh options to better manage ssh connection

Torque: to fully allocate a job on one node we need to get the number of processor from the node list.

Scale-up: we need the system to launch the 2nd instance asap since the whole loop takes minutes.

Concerning job sleep, the sum of the two jbs sleep time should 
the timeout, so that if they would run on the same instance the whole
test would fail. At the same time we cannot simply split the full
timeout by 2 because the intance takes from 3 to 6 minutes to come up.
So we need to absorb this time during the computation of the first job.

Signed-off-by: Maurizio Melato <mmelato@amazon.com>

### DIFF
--- a/tests/cfncluster-release-check.py
+++ b/tests/cfncluster-release-check.py
@@ -58,7 +58,7 @@ success = 0
 # run a single test, possibly in parallel
 #
 def run_test(region, distro, scheduler, instance_type, key_name, key_path):
-    testname = '%s-%s-%s-%s' % (region, distro, scheduler, instance_type.replace('.', '-'))
+    testname = '%s-%s-%s-%s' % (region, distro, scheduler, instance_type.replace('.', ''))
     test_filename = "%s-config.cfg" % testname
 
     sys.stdout.write("--> %s: Starting\n" % (testname))
@@ -82,7 +82,10 @@ def run_test(region, distro, scheduler, instance_type, key_name, key_path):
     file.write("[global]\n")
     file.write("cluster_template = default\n")
     file.write("[scaling custom]\n")
-    file.write("scaling_adjustment = 2\n")
+    file.write("scaling_adjustment = 1\n")
+    file.write("scaling_period = 30\n")
+    file.write("scaling_evaluation_periods = 1\n")
+    file.write("scaling_cooldown = 300\n")
     file.close()
 
     stdout_f = open('%s-stdout.txt' % testname, 'w')
@@ -112,6 +115,7 @@ def run_test(region, distro, scheduler, instance_type, key_name, key_path):
             exception_raised = True
             raise Exception('Master IP not found')
         stdout_f.write("--> %s master ip: %s" % (testname, master_ip))
+        print("--> %s master ip: %s" % (testname, master_ip))
 
         # run test on the cluster...
         # ssh_params = ['-n'] # ssh only, not for scp
--- a/tests/cfncluster-release-check.py
+++ b/tests/cfncluster-release-check.py
@@ -90,6 +90,7 @@ def run_test(region, distro, scheduler, instance_type, key_name, key_path):
 
     master_ip = ''
     username = username_map[distro]
+    exception_raised = False;
 
     try:
         # build the cluster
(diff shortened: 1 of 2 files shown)


tests/cluster-check.sh
```diff
@@ -45,42 +49,42 @@ set -e
...
-    sbatch -N 2 ./job1.sh
-    sbatch -N 2 ./job2.sh
+    sbatch -N 1 ./job1.sh
+    sbatch -N 1 ./job2.sh
 
 elif test "$scheduler" = "sge" ; then
     # get the slots per node count of the first real node (one with a
     # architecture type of lx-?), so that we can reserve an enitre
     # node's worth of slots at a time.
     ppn=` qhost | grep 'lx-' | head -n 1 | sed -n -e 's/[^[:space:]]*[[:space:]]\+[^[:space:]]*[[:space:]]\+\([0-9]\+\).*/\1/p'`
```
enricou: If there are not yet compute nodes, the `ppn` is an empty value, so `count=$((ppn))` results `0` and the job will fail with: `Unable to run job: "job" rejected: pe range must be greater than zero`.

We could check the `count` value and fail with a more clear message or perform a loop waiting for the compute nodes.

tests/cluster-check.sh
```diff
@@ -91,22 +95,26 @@ EOF
     qsub ./job2.sh
 
 elif test "$scheduler" = "torque" ; then
+    _chost=$(pbsnodes -l free | head -n 1 | cut -d ' ' -f1)
```
enricou: If there aren't yet free compute nodes the `_chost` will be an empty value because `pbsnodes -l free` doesn't return any host.
Then in the next command the `_ppn` will be the value of the `np` of the master host or a multiple row value if there are other (not free) hosts. 
E.g.:
```
$ _ppn=$(pbsnodes | tr -d '\t ' | sed -n '/np=/{s/^np=\([0-9]\+\)/\1/;p;}')
1000
4
4
```
Then the following submission will fail:
```
$ qsub -l nodes=1:ppn=${ppn} ./job2.sh
qsub: script file '4' cannot be loaded - No such file or directory
```

We could check the `ppn` value and fail with a more clear message or perform a loop waiting for the free compute nodes.

tests/cfncluster-release-check.py
```diff
@@ -107,29 +111,45 @@ def run_test(region, distro, scheduler, instance_type, key_name, key_path):
...
+        ssh_params += ['-o', 'ServerAliveCountMax=5']
+        ssh_params += ['-o', 'ServerAliveInterval=30']
         if key_path:
             ssh_params.extend(['-i', key_path])
 
         subprocess.check_call(['scp'] + ssh_params + ['cluster-check.sh', '%s@%s:.' % (username, master_ip)],
                               stdout=stdout_f, stderr=stderr_f)
         subprocess.check_call(
-            ['ssh'] + ssh_params + ['%s@%s' % (username, master_ip), '/bin/bash --login cluster-check.sh %s' % scheduler],
+            ['ssh'] + ['-n'] + ssh_params + ['%s@%s' % (username, master_ip), '/bin/bash --login cluster-check.sh %s' % scheduler],
```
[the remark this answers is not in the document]
mauri-melato: done

tests/cfncluster-release-check.py
```diff
@@ -107,29 +111,45 @@ def run_test(region, distro, scheduler, instance_type, key_name, key_path):
...
         if master_ip == '':
-            print('!! %s: Master IP not found; aborting !!' % (testname))
+            stderr_f.write('!! %s: Master IP not found; aborting !!' % (testname))
+            exception_raised = True
             raise Exception('Master IP not found')
+        stdout_f.write("--> %s master ip: %s" % (testname, master_ip))
         print("--> %s master ip: %s" % (testname, master_ip))
 
         # run test on the cluster...
+        # ssh_params = ['-n'] # ssh only, not for scp
```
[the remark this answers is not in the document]
mauri-melato: removed the commented code

tests/cluster-check.sh
mauri-melato: Improved the way we retrieve the slots per node for the rare cases (seen in our tests) where the first compute node was not there when trying to submit the jobs.

tests/cluster-check.sh
mauri-melato: Improved the way we retrieve the slots per node for the rare cases (seen in our tests) where the first compute node was not there when trying to submit the jobs.

PR #439, opened 2018-07-03
Add unit tests for cfncluster

fnubalaj:
Add unit tests for cfncluster.

To run the unit tests, run cli/tests/cfncluster-unittest.py


cli/tests/cfncluster-unittest.py
```diff
@@ -0,0 +1,294 @@
...
+config_file = 'cli/tests/config'
+with open('cloudformation/cfncluster.cfn.json') as f:
+    json_data = json.dumps(json.load(f))
+
+
+def config_logger_test():
+    logging.basicConfig(stream=test_log_stream, level=logging.INFO)
+
+
+def setup_configurations():
```
sean-smith: Is this method just to get a link to the cfncluster cloudformation template? If so you can just grab `template_url` from `config` object after you call: [#L26](https://github.com/awslabs/cfncluster/blob/develop/cli/cfncluster/cfncluster.py#L26)

cli/tests/cfncluster-unittest.py
```diff
@@ -0,0 +1,294 @@
...
+class CFN_cluster_test(unittest.TestCase):
+    def setUp(self):
+        config_logger_test()
+
+    def test_cfn_cluster_version(self):
+        args = BaseArgs()
+        cfncluster.version(args)
+        log = test_log_stream.getvalue()
+        version = re.match(r"^INFO:\w+\.\w+:(\d+\.\d+\.\d+)$", log).group(1)
+        self.assertEqual(version, '1.5.1')
```
sean-smith: Pull the current version from `setup.py` so we don't need to change multiple places.

.travis.yml
```diff
@@ -19,4 +19,6 @@ install:
```
sean-smith: Can you add python 3.7

sean-smith:
Also please squash all the commits into one. This should do the trick:

```git rebase -i upstream/develop```

File: cli/tests/cfncluster-unittest.py (hunk 2 above)
Comment (fnubalaj) [reply cli/tests/cfncluster-unittest.py]: I have decided to take it from the json file instead of the python file.

File: .travis.yml (hunk 3 above)
Comment (fnubalaj) [reply .travis.yml]: It is not yet supported

File: cli/tests/cfncluster-unittest.py [hunk 4]
```diff
@@ -0,0 +1,295 @@
...
+    @mock_ec2
+    @mock_cloudformation
+    @mock_s3
+    def test_cfn_cluster_list_nonempty(self):
+        template_url = setup_configurations()
+        args = CreateClusterArgs(template_url, False)
+        cfncluster.create(args)
+        cfncluster.list(args)
+        log = test_log_stream.getvalue()
+        self.assertTrue(args.cluster_name in log)
```
sean-smith: Can you also test that this is the only thing returned? If we mess up the parsing of cloudformation for example, it could return `cfncluster-cluster_name` instead of just `cluster_name` which will go ahead and throw a "stack not found" exception when the customer goes to list.

cli/tests/cfncluster-unittest.py
```diff
@@ -0,0 +1,295 @@
...
+        self.assertEqual(sys_ex.exception.code, 1)
+        log = test_log_stream.getvalue()
+        error_prefix = "CRITICAL:"
+        self.assertTrue(error_prefix in log)
+
+    @mock_ec2
+    @mock_cloudformation
+    @mock_s3
+    @mock_autoscaling
+    def test_cfn_cluster_delete(self):
```
sean-smith: can you also add a delete unit test for when the cluster doesn't exist?

cli/tests/cfncluster-unittest.py
```diff
@@ -0,0 +1,295 @@
...
+    @mock_s3
+    @mock_autoscaling
+    def test_cfn_cluster_start(self):
+        template_url = setup_configurations()
+        args = CreateClusterArgs(template_url, True)
+        cfncluster.create(args)
+        cfncluster.start(args)
+        log = test_log_stream.getvalue()
+        error_prefix = "CRITICAL:"
+        success_message = "Starting compute fleet"
```
sean-smith: Also check the ASG is set to the correct values (per config)

cli/tests/cfncluster-unittest.py
```diff
@@ -0,0 +1,295 @@
...
+        template_url = setup_configurations()
+        args = CreateClusterArgs(template_url, True)
+        cfncluster.create(args)
+        cfncluster.start(args)
+        cfncluster.stop(args)
+        log = test_log_stream.getvalue()
+        error_prefix = "CRITICAL:"
+        success_message = "Stopping compute fleet"
+        self.assertFalse(error_prefix in log)
+        self.assertTrue(success_message in log)
```
sean-smith: Same here.

cli/tests/config
```diff
@@ -0,0 +1,16 @@
+[aws]
+aws_region_name = us-east-1
+
+[cluster default]
+vpc_settings = public
+key_name = test
+
+[vpc public]
+master_subnet_id = subnet-2ce814d3
```
sean-smith: Do these actually get used?

.travis.yml
```diff
@@ -10,6 +10,10 @@ python:
 
 install: 
   - cd cli
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.3 ]]; then
+    pip install -U setuptools;
```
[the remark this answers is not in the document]
fnubalaj: These are specific to issues in 3.3

cli/tests/config
fnubalaj: Yes, they are written and used. This is only used for the sections other than [vpc public]

cli/tests/cfncluster-unittest.py
fnubalaj: That would require to mock out describe_stack_resources, since we would have to get_asg-instances. It was decided that should be a done a part of larger refactoring.

.travis.yml
```diff
@@ -19,4 +23,8 @@ install:
 
 sudo: false
 
-script: sh tests/test.sh
+script:
+  - if [[ $TRAVIS_PYTHON_VERSION != 2.6 ]];
```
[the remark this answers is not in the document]
fnubalaj: Yes! We are not testing for Python 2.6 because there are issues with moto on Python 2.6

cli/tests/cfncluster-unittest.py
fnubalaj: Added

cli/tests/cfncluster-unittest.py
fnubalaj: That would require to mock out describe_stack_resources (moto does not do that), since we would have to get_asg-instances. It was decided that should be a done a part of larger refactoring

PR #490, opened 2018-08-21
Increase scaledown idle time and the timeout on scale down check

fnubalaj:
A buffer of more than a minute is needed for the instance to be shutdown and detached from queue


mauri-melato:
Why it was needed to increase the `scaledown_idletime`? 
Did the instance get terminated before joining the cluster / receiving the submitted job?

PR #510, opened 2018-08-28
EC2 Launch Template Support

sean-smith:
Switch from using Launch Configurations in the ASG to a Launch Template.

Due to changes in how the stack paramters are encoded, I changed SSH to use stack outputs. This makes it significantly faster:

### Pros:
Much faster
```bash
$ time cfncluster ssh sge -d
ssh ec2-user@18.213.184.153
cfncluster ssh launch-template -d  0.42s user 0.12s system 44% cpu 1.219 total
```
Versus:
```bash
$ time cfncluster ssh sge -d
ssh ec2-user@18.209.144.162
cfncluster ssh sge -d  0.51s user 0.12s system 17% cpu 3.541 total
```

### Cons:
Can't ssh while stack is creating or updating


cli/cfncluster/cfncluster.py
```diff
@@ -384,6 +384,7 @@ def instances(args):
         print('%s         %s' % (instance[0],instance[1]))
 
 def get_head_user(parameters, template):
+    print(template)
```
sean-smith: Will remove 😌

enrico-usai:
It could be useful to add in the commit message why the Launch template is better than Launch configuration (e.g. Launch templates can be updated and versioned, etc...).

sean-smith:
Ok, I'll change the commit message.

Yep, it's tested. One weird side effect is the `cfncluster ssh` is breaking since it's encoding the parameters portion of the cloudformation template separately.

bwbarrett:
Why only support Launch Templates for the compute instances?

Also, it is unclear to me how the user interface would change (or is this step one, where we use launch templates internally, but don't expose to users)?

docs/source/conf.py
```diff
@@ -52,7 +52,11 @@
 # The short X.Y version.
 version = '1.6'
 # The full version, including alpha/beta/rc tags.
+<<<<<<< HEAD
```
fnubalaj: This needs to be fixed?

docs/source/configuration.rst
```diff
@@ -522,7 +522,65 @@ Settings which define how the compute nodes scale. ::
 
 
     [scaling custom]
-    scaledown_idletime = 10
+    scaling_period = 60
+    scaling_cooldown = 300
+
+scaling_threshold
```
fnubalaj: I think you need to rebase here, no?

PR #518, opened 2018-08-30
Remove lines containing removed cloudwatch metric parameters

fnubalaj:
Remove the lines containing removed cloudwatch metric parameters which were previously used for scaling up.
While parameters were removed by a previous commit, few lines were missed which is being addressed by this commit


enrico-usai:
Could you also fix the configuration sample file?
https://github.com/awslabs/cfncluster/blob/develop/cli/cfncluster/examples/config

PR #522, opened 2018-08-30
Merge scale into develop

fnubalaj:
This will be a merge commit that merges all features from scaling branch into develop branch.


enrico-usai:
Did you already check if the "autoscaling" docs part is aligned to the commit?
https://github.com/awslabs/cfncluster/blob/develop/docs/source/autoscaling.rst

sean-smith:
@enrico-usai it's not. we can add it as a seperate PR in develop before releasing. This PR has no "new" node, just merging from scale into develop

PR #528, opened 2018-08-30
Cfncluster 1.6.0rc1

fnubalaj:


CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 CHANGELOG
 =========
 
+1.6.0
+=====
+* Faster scaling up and scaling down
```
enrico-usai: I think we should better highlight scaling changes since this is the reason of the minor version increasing. 
E.g. :
- Refactor scaling down to take advantages of per-seconds billing and to lock hosts before termination.
- Refactor scaling up to take account of the real number of pending/requested jobs/slots and instance slots.

enrico-usai: I'm not sure about that, but we could also think to add the list of closed issues, e.g. https://github.com/awslabs/cfncluster/issues/339, https://github.com/awslabs/cfncluster/issues/334, etc

fnubalaj: Since there is no precedent of adding issues to Changelog, I will refrain from that, but I have added the changes suggested. Thanks for the feedback.

PR #568, opened 2018-10-03
Improve integration tests for the scaling process

demartinofra:
Add to the integration tests defined in cfncluster-release-check
a more granular test to verify that scale-up and scale-down features
work correctly. The integration tests are now able to verify that
the cluster is correctly scaled-up and that all created nodes are
available to the scheduler. It is also possible to customize the
scaling range of the cluster and make sure that, when some of the
compute nodes are configured to be retained in absence of jobs, this
condition is verified.
The change also refactors the integration tests structure by moving all
tests defined in cluster-check bash script into the
cfncluster-release-check python script. This was done in order to
increase the cohesion and future extensibility of integration tests.
Also scaledown_idletime has been increased to 4 minutes in integration tests
in order to minimize the chance that compute nodes are released even 
before the cluster completes its creation.

### DIFF
--- a/tests/parallelcluster-release-check.py
+++ b/tests/parallelcluster-release-check.py
@@ -38,9 +38,11 @@ import sys
 import threading
 import time
 from builtins import exit
+from collections import namedtuple
 
 import boto3
 import process_helper as prochelp
+from hamcrest import *
 
 
 class ReleaseCheckException(Exception):
@@ -50,6 +52,23 @@ class ReleaseCheckException(Exception):
 #
 # configuration
 #
+ClusterConfig = namedtuple(
+    "ClusterConfig",
+    [
+        "config_file",
+        "stack_name",
+        "region",
+        "distro",
+        "instance_type",
+        "scheduler",
+        "username",
+        "key_path",
+        "key_name",
+        "master_node",
+        "scaledown_idletime",
+    ],
+)
+
 username_map = {
     "alinux": "ec2-user",
     "centos6": "centos",
@@ -57,6 +76,25 @@ username_map = {
     "ubuntu1404": "ubuntu",
     "ubuntu1604": "ubuntu",
 }
+
+# commands used to retrieve the number of compute nodes in each scheduler
+get_compute_nodes_command_map = {
+    "sge": '/bin/bash --login -c "qhost | grep -o ip- | wc -l"',
+    "slurm": '/bin/bash --login -c "sinfo --Node --noheader | grep compute | wc -l"',
+    "torque": '/bin/bash --login -c "echo $(( $(/opt/torque/bin/pbsnodes -l all | wc -l) - 1))"',
+}
+
+# default ssh options
+ssh_config_options = [
+    "-o {0}".format(option)
+    for option in [
+        "StrictHostKeyChecking=no",
+        "BatchMode=yes",
+        "ConnectTimeout=60",
+        "ServerAliveCountMax=5",
+        "ServerAliveInterval=30",
+    ]
+]
 
 #
 # global variables (sigh)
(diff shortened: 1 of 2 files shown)


tests/cfncluster-release-check.py
```diff
@@ -83,116 +114,330 @@ def _double_writeln(fileo, message):
...
+        expected_compute_nodes: pair containing the expected compute nodes (min_compute_nodes, max_compute_nodes)
+    """
+    assert_that(asg_capacity_time_series, is_not(empty()), 'asg_capacity_time_series cannot be empty')
+    assert_that(compute_nodes_time_series, is_not(empty()), 'compute_nodes_time_series cannot be empty')
+
+    expected_asg_capacity_min, expected_asg_capacity_max = expected_asg_capacity
+    expected_compute_nodes_min, expected_compute_nodes_max = expected_compute_nodes
+    actual_asg_capacity_min, actual_asg_capacity_max = (min(asg_capacity_time_series), max(asg_capacity_time_series))
+    actual_compute_nodes_min, actual_compute_nodes_max = (min(compute_nodes_time_series), max(compute_nodes_time_series))
+    assert_that(actual_asg_capacity_min, is_(equal_to(expected_asg_capacity_min)),
```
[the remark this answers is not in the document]
demartinofra: Nice catch! I changed it so that we look for the min only among the values that are following the max. do you think this can work?

tests/cfncluster-release-check.py
```diff
@@ -83,116 +114,330 @@ def _double_writeln(fileo, message):
...
-
-        prochelp.exec_command(['ssh', '-n'] + ssh_params + ['%s@%s' % (username, master_ip), '/bin/bash --login cluster-check.sh scaledown_check %s' % scheduler],
-                              stdout=out_f, stderr=sub.STDOUT, universal_newlines=True)
+        _double_writeln(out_f, "--> %s: Cluster created successfully" % testname)
+
+        cluster_config = cluster_config._replace(master_node=_get_master_ip(cluster_config_file=test_filename,
+                                                                            cluster_name=testname,
+                                                                            log_file=out_f))
+
+        jobs_start_time = int(time.time())
```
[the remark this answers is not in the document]
demartinofra: Good point! I moved the logging of the start time into the cluster-check script. Please take a look and let me know if this looks good to you.

tests/parallelcluster-release-check.py
```diff
@@ -90,138 +128,396 @@ def _double_writeln(fileo, message):
...
+        number_of_nodes: number of available compute nodes.
+    """
+    output = _exec_ssh_command(
+        command=get_compute_nodes_command_map[cluster_config.scheduler],
+        username=cluster_config.username,
+        host=cluster_config.master_node,
+        key_path=cluster_config.key_path,
+    )
+    # get last line of the output containing the number of compute nodes
+    return int(output.split()[-1])
```
enrico-usai: We could detect `_exec_ssh_command` failures or catch the `TypeError` in the `int()` conversion.

tests/parallelcluster-release-check.py
```diff
@@ -90,138 +128,396 @@ def _double_writeln(fileo, message):
...
+    """
+    Retrieves the desired capacity of the autoscaling group for a specific cluster.
+    Args:
+        cluster_config: named tuple of type ClusterConfig containing the configuration of the cluster.
+
+    Returns:
+        asg_capacity: the desired capacity of the autoscaling group.
+    """
+    asg_conn = boto3.client("autoscaling", region_name=cluster_config.region)
+    r = asg_conn.describe_tags(Filters=[{"Name": "value", "Values": [cluster_config.stack_name]}])
```
enrico-usai: Minor: I know that the `r` variable is only used here, but in general I'd rather have a more descriptive name.

tests/parallelcluster-release-check.py
demartinofra: We agreed that the global try-catch in the run-test function is enough to record such errors.

PR #581, opened 2018-10-11
Multiple EBS volumes

rexcsn:
- Modified cfnconfig.py to take in the new format of the config file, with multiple ebs sections
- Modified config_sanity.py to add validation for VolumeType and EBSEncryption, as those can no longer be checked in the cloudformation template
- Modified cfncluster.cfn.json to handle multiple EBS volumes; added 1 additional parameter: NumberOfEBSVol, added new resource EBSCfnStack, took out all EBS related conditions, and changed all EBS related parameters to comma separated lists
- Added ebs_vol_substack.cfn.json, the new template for EBS substack, which supports up to 5 EBS volume
- Added generate-ebs-substack.py, which can be used to generate template for EBS substack via troposphere

### DIFF
--- a/cli/cfncluster/cfnconfig.py
+++ b/cli/cfncluster/cfnconfig.py
@@ -20,11 +20,12 @@ import os
 import sys
 import inspect
 import pkg_resources
-import json
 import urllib.request, urllib.error, urllib.parse
 from . import config_sanity
 import boto3
+import json
 from botocore.exceptions import ClientError
+
 
 def getStackTemplate(region, aws_access_key_id, aws_secret_access_key, stack):
     cfn = boto3.client('cloudformation', region_name=region,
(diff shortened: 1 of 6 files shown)


cli/cfncluster/cfnconfig.py
```diff
@@ -268,47 +271,122 @@ def __init__(self, args):
...
                     except configparser.NoOptionError:
                         pass
+                    except configparser.NoSectionError:
+                        print("ERROR: [%s] section defined in [ebs_settings] does not exist" %section)
+                        sys.exit(1)
+        except AttributeError:
+            pass
+
+        try:
+            if len(__temp_dir_list) == len(self.__ebs_section):
```
fnubalaj: What if both the lists are empty?

After talking to Rex, there is no case where __ebs_section can exist as an empty list.

util/generate-ebs-substack.py
```diff
@@ -0,0 +1,100 @@
...
+                                             Type="String",
+                                             Description="Availability Zone the cluster will launch into. THIS IS REQUIRED"))
+VolumeSize = t.add_parameter(Parameter("VolumeSize",
+                                       Type="CommaDelimitedList",
+                                       Description = "Size of EBS volume in GB, if creating a new one",
+                                       Default = "20, 20, 20"))
+VolumeType = t.add_parameter(Parameter("VolumeType",
+                                       Type="CommaDelimitedList",
+                                       Description="Type of volume to create either new or from snapshot",
+                                       Default = "gp2, gp2, gp2"))
```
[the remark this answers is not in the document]
fnubalaj: After looking at the template, these values don't serve any purpose, delete them

cli/cfncluster/config_sanity.py
```diff
@@ -167,3 +167,15 @@ def check_resource(region, aws_access_key_id, aws_secret_access_key, resource_ty
                 sys.exit(1)
             print('Config sanity error: %s' % e.response.get('Error').get('Message'))
             sys.exit(1)
+    # EC2 EBS Volume type
+    elif resource_type == 'EC2VolumeType':
+        __allowedVolumeTypes = ["standard", "gp2", "io1", "st1", "sc1"]
+        if resource_value not in __allowedVolumeTypes:
+            print('Config sanity error: incorrect volume type %s.' % resource_value)
```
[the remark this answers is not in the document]
sean-smith: actually I think this is exactly the right thing to include in config_sanity. Let's say a new volume_type comes out and the customer wants to use it, all they have to do is set `config_sanity = false`

cloudformation/cfncluster.cfn.json
```diff
@@ -734,21 +735,14 @@
     },
     "VolumeSize": {
       "Description": "Size of EBS volume in GB, if creating a new one",
-      "Type": "Number",
-      "Default": "20"
+      "Type": "String",
+      "Default": "20, 20, 20, 20, 20"
     },
     "VolumeType": {
       "Description": "Type of volume to create either new or from snapshot",
```
sean-smith: This should be "Comma delimited list of ..."

cloudformation/cfncluster.cfn.json
```diff
@@ -2827,15 +2748,15 @@
           }
         ],
         "LaunchTemplate": {
-            "LaunchTemplateId" : {
```
sean-smith: why did the whitespace get changed?

cloudformation/cfncluster.cfn.json
```diff
@@ -2943,9 +2864,9 @@
             "AssociatePublicIpAddress" : {
               "Fn::If": [
                 "ComputePublicIps",
-                true,
-                false
-              ],
+                "true",
```
sean-smith: I think this was included from a previous commit. Looks like launch template commit.

cloudformation/cfncluster.cfn.json
```diff
@@ -3887,6 +3736,48 @@
...
+          "EBSSnapshotId": {
+            "Ref" : "EBSSnapshotId"
+          },
+          "NumberOfEBSVol": {
+            "Ref" : "NumberOfEBSVol"
+          }
+        },
+
+        "TemplateURL":
+          "https://s3.us-east-2.amazonaws.com/shuningc-cfncluster/extra_template/ebs_vol_substack.cfn.json"
```
sean-smith: This needs to be dynamically generated and point to non-test bucket :)

util/generate-ebs-substack.py
```diff
@@ -0,0 +1,100 @@
...
+                                        Type="CommaDelimitedList",
+                                        Description= "Existing EBS volume Id",
+                                        Default= "NONE, NONE, NONE"))
+EBSSnapshotId = t.add_parameter(Parameter("EBSSnapshotId",
+                                          Type="CommaDelimitedList",
+                                          Description="Id of EBS snapshot if using snapshot as source for volume",
+                                          Default= "NONE, NONE, NONE"))
+EBSVolumeNum = t.add_parameter(Parameter("NumberOfEBSVol",
+                                         Type="Number",
+                                         Description= "Number of EBS Volumes the user requested, up to 10",
```
sean-smith: Shouldn't this be 5? Also pull this value directly from `numberOfVol `.

util/generate-ebs-substack.py
```diff
@@ -0,0 +1,100 @@
...
+                                         Default="1"))
+
+UseVol = [None]*numberOfVol
+UseExistingEBSVolume = [None]*numberOfVol
+v = [None]*numberOfVol
+
+for i in range(numberOfVol):
+    if i == 0:
+        CreateVol = t.add_condition("Vol" + str(i + 1) + "_CreateEBSVolume",
```
sean-smith: use python string interpolation. It gets rid of need for `str()`. See https://www.programiz.com/python-programming/string-interpolation#%-formatting

cloudformation/cfncluster.cfn.json
```diff
@@ -3887,6 +3736,48 @@
           "id": "ee5f3006-419e-4786-b527-e9503a662e5e"
         }
       }
+    },
+        "EBSCfnStack": {
```
fnubalaj: Fix indentation

cloudformation/cfncluster.cfn.json
fnubalaj: Change this, upload to cfncluster s3 bucket

cli/cfncluster/config_sanity.py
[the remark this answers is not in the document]
fnubalaj: But, the errors are no longer caught while creating the stack.

cloudformation/cfncluster.cfn.json
fnubalaj: Since, we are fixing the maximum number of volumes to 5, the template does not need to be dynamically generated. However, I agree that it should point to a bucket which is not dependent on a developer's account :)

cloudformation/cfncluster.cfn.json
rexcsn: Apparently the Cloud Formation Template does not allow passing comma separated list parameters into a substack. When I changed the type from string to comma separated list the template produced an error.

cli/cfncluster/cfnconfig.py
```diff
@@ -24,8 +24,10 @@
 import urllib.request, urllib.error, urllib.parse
 from . import config_sanity
 import boto3
+import json
```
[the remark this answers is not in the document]
rexcsn: I accidentally imported json twice, that statement is removed

cli/cfncluster/config_sanity.py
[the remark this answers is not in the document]
rexcsn: So should I check it in config_sanity or cfnconfig.py?

cloudformation/cfncluster.cfn.json
rexcsn: The launch template commit fixed a similar issue under "EbsOptimized", but I think this is a separate issue

cli/cfncluster/cfnconfig.py
```diff
@@ -268,47 +270,133 @@ def __init__(self, args):
             pass
 
         # Determine if EBS settings are defined and set section
+
+        # #self.parameters.append(("EBSSubstackURL",
```
demartinofra: I think these two commented lines can be removed

cli/cfncluster/cfnconfig.py
```diff
@@ -268,47 +270,133 @@ def __init__(self, args):
...
+        #                         "https://s3.us-east-2.amazonaws.com/shuningc-cfncluster/extra_template/tropo_vol_substack.cfn.json"))
+
+        # CHANGED: adapt to list of ebs_settings
```
demartinofra: This __init__ function is growing more and more and becoming very hard to read and impossible to test. It would be nice to move at the least the new code portions into separate functions which might be easily unit tested.
As a very general guideline: almost every time a comment is necessary to clarify the intent of the code maybe a function with a self-explanatory name can be created

cli/cfncluster/cfnconfig.py
```diff
@@ -268,47 +270,133 @@ def __init__(self, args):
...
+                                  volume_iops=('VolumeIOPS',None), encrypted=('EBSEncryption', None),
+                                  ebs_volume_id=('EBSVolumeId','EC2Volume'))
+        # EBS options processing
+        try:
+            if self.__ebs_section:
+                for key in self.__ebs_options:
+                    __temp_parameter_list = []
+                    for section in self.__ebs_section:
+                        try:
+                            __temp__ = __config.get(section, key)
```
demartinofra: why the double trailing underscore? could we have a more meaningful name rather then temp only, especially cause this var is used in several places?

cloudformation/cfncluster.cfn.json
```diff
@@ -759,10 +753,9 @@
       "Type": "AWS::EC2::AvailabilityZone::Name"
     },
     "EBSSnapshotId": {
-      "Description": "Id of EBS snapshot if using snapshot as source for volume",
+      "Description": "Comma delimited list of Id of EBS snapshot if using snapshot as source for volume",
       "Type": "String",
-      "Default": "NONE",
-      "AllowedPattern": "(NONE|^snap-[0-9a-z]{8}$|^snap-[0-9a-z]{17}$)"
```
mauri-melato: We can probably also keep this validation by chaning the regex as  `^(NONE|snap-[0-9a-z]{8}|snap-[0-9a-z]{17})(,(NONE|snap-[0-9a-z]{8}|snap-[0-9a-z]{17}))*$`

cloudformation/cfncluster.cfn.json
```diff
@@ -1008,15 +997,19 @@
       "AllowedPattern": "(NONE|^sg-[0-9a-z]{8}$|^sg-[0-9a-z]{17}$)"
     },
     "EBSVolumeId": {
-      "Description": "Existing EBS volume Id",
+      "Description": "Comma delimited list of existing EBS volume Id",
       "Type": "String",
-      "Default": "NONE",
-      "AllowedPattern": "(NONE|^vol-[0-9a-z]{8}$|^vol-[0-9a-z]{17}$)"
```
mauri-melato: Similar comment to EBSSnapshotId

cloudformation/cfncluster.cfn.json
```diff
@@ -3887,6 +3736,48 @@
...
+          "EBSSnapshotId": {
+            "Ref" : "EBSSnapshotId"
+          },
+          "NumberOfEBSVol": {
+            "Ref" : "NumberOfEBSVol"
+          }
+        },
+
+        "TemplateURL":
+        "https://s3.us-east-2.amazonaws.com/shuningc-cfncluster/extra_template/ebs_vol_substack.cfn.json"
```
mauri-melato: I'd have expected a path composition, is the substack going to follow the location of the main CFN template...?

cli/cfncluster/cfnconfig.py
```diff
@@ -268,47 +270,133 @@ def __init__(self, args):
...
+
+        # #self.parameters.append(("EBSSubstackURL",
+        #                         "https://s3.us-east-2.amazonaws.com/shuningc-cfncluster/extra_template/tropo_vol_substack.cfn.json"))
+
+        # CHANGED: adapt to list of ebs_settings
         try:
+            #see if there is a "ebs_settings" section, if there is no such section at all, give configparser.NoOptionError and pass
             self.__ebs_settings = __config.get(self.__cluster_section, 'ebs_settings')
+
+            #there is a ebs_setting section with some setting self.__ebs_settings, check if there is some self.__ebs_settings
```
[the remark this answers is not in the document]
rexcsn: These comments were made when I just started so that I could understand the program, it should not be a part of the modification. I will take these comments out

Issue #600, opened 2018-10-24
instance role is used instead of instance profile


UnquietCode:
When using a custom role (`ec2_iam_role`) the role is passed directly as the instance profile, causing an error. Instead, it should create a new instance profile for the role and use that. This error occurs three times in the CloudFormation template JSON.

UnquietCode:
related pull request: https://github.com/awslabs/cfncluster/pull/602

PR #602, opened 2018-10-24
create instance profile when using a provided role

UnquietCode:
*Issue #, if available:*
addresses issue #600

*Description of changes:*
Creates a new instance profile container when a custom IAM role is provided.

### DIFF
--- a/cloudformation/cfncluster.cfn.json
+++ b/cloudformation/cfncluster.cfn.json
@@ -1976,6 +1976,18 @@
         }
       }
     },
+    "NewInstanceProfile": {
+      "Type": "AWS::IAM::InstanceProfile",
+      "Properties": {
+        "Path": "/",
+        "Roles": [
+          {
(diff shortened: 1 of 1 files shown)


cloudformation/cfncluster.cfn.json
```diff
@@ -1976,6 +1976,18 @@
         }
       }
     },
+    "NewInstanceProfile": {
```
sean-smith: Instead of creating a new instance profile why don't you use the existing `RootInstanceProfile` and use an if statement to reference the role. https://github.com/awslabs/cfncluster/blob/develop/cloudformation/cfncluster.cfn.json#L1853

UnquietCode: Sure, that would also work.

sean-smith:
Tested with a custom `ec2_iam_role`

PR #610, opened 2018-10-30
1.6.1 develop to master merge

fnubalaj:


sean-smith:
You're specifically referencing this commit: https://github.com/awslabs/cfncluster/commit/486a913d29181fe6bbebe76dddee528a7992d77a

You can use a custom `ec2_iam_role` and give it broader `DescribeStacks` permissions. Unfortunately we want to scope the iam permissions down to just cfncluster stacks, so for your use case you'd need a custom role.
See https://cfncluster.readthedocs.io/en/latest/configuration.html#ec2-iam-role
And https://cfncluster.readthedocs.io/en/latest/iam.html

chambm:
I need a custom IAM role for having a custom stack name? If that's the case, then somewhere there needs to be a check that if there is no custom IAM role, the stack name must match `cfncluster-*`. So I would have found out a lot sooner (and cheaper) that I needed to rename the stack. :)

sean-smith:
I guess the fundamental issue here is we name all stacks `cfncluster-` in the cli, and we assume that stacks created by cfncluster follow that naming criteria.

We are not taking into account stacks created outside the cli.

PR #640, opened 2018-11-13
filter out Batch outputs from stack output

enrico-usai:
Now the output is:
```
$ pcluster create batch-test--cluster-template batch
Beginning cluster creation for cluster: batch-test
Creating stack named: parallelcluster-batch-test
Status: parallelcluster-batch-test- CREATE_COMPLETE
MasterPublicIP: 63.32.235.86
ClusterUser: ec2-user
MasterPrivateIP: 172.31.20.167
```

### DIFF
--- a/cli/pcluster/pcluster.py
+++ b/cli/pcluster/pcluster.py
@@ -143,13 +143,7 @@ def create(args):
                                      event.get('ResourceStatusReason')))
             logger.info('')
             result_stack = cfn.describe_stacks(StackName=stack_name).get("Stacks")[0]
-            outputs = result_stack.get('Outputs', [])
-            parameters = result_stack.get('Parameters')
-            ganglia_enabled = is_ganglia_enabled(parameters)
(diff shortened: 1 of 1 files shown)


cli/pcluster/pcluster.py
```diff
@@ -168,6 +162,20 @@ def create(args):
...
         sys.exit(1)
 
+
+def _print_stack_outputs(stack):
+    outputs = stack.get('Outputs', [])
+    parameters = stack.get('Parameters')
+    ganglia_enabled = is_ganglia_enabled(parameters)
+    for output in outputs:
+        output_key = output.get('OutputKey')
+        if not ganglia_enabled and output_key.startswith('Ganglia'):
```
demartinofra: Instead of growing the number of ifs I would have a static list with the Outputs we want to whitelist and print it if the key belongs to this list. (for ganglia we can add the key to the list after we check if ganglia is enabled in this function

PR #647, opened 2018-11-14
Restructure docs directory

demartinofra:
This was done to have a consistent behavior when building the docs
with make and with readthedocs. readthedocs looks for config.py file and
sets the working directory where the file is located. On the contrary
make uses as working directory the location of the Makefile.


docs/conf.py
```diff
@@ -66,7 +66,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = ['staging']
```
enrico-usai: should we add the makefiles to the excluded patterns?

demartinofra: we only need to exclude rst files we don't want to include in the output and directories containing rts files to exclude or directories we don't want autodiscovery to search. Makefile is not included in the output.

PR #651, opened 2018-11-14
Patch for stringify errors

rexcsn:
Error description: str() on strings will sometimes cause a KeyError problem

- Modified cfnconfig.py to remove all occurances of stringify and use python string interpolation instead, remove str package
- Modified pcluster.py to catch KeyError separately so we can better identify the problem in the future

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -13,7 +13,6 @@ from __future__ import absolute_import
 
 from future import standard_library
 standard_library.install_aliases()
-from builtins import str
 from builtins import object
 import configparser
 import os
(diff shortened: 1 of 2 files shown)


cli/pcluster/cfnconfig.py
```diff
@@ -108,12 +107,12 @@ def __load_ebs_options(self, __config):
...
                 sys.exit(1)
             # Modify list
-            self.__ebs_section = str(self.__ebs_settings).split(',')
+            self.__ebs_section = self.__ebs_settings.split(',')
             if (len(self.__ebs_section) > self.__MAX_EBS_VOLUMES):
                 print("ERROR: number of EBS volumes requested is greater than the MAX.\n"
                       "Max number of EBS volumes supported is currently %s" % self.__MAX_EBS_VOLUMES)
                 sys.exit(1)
-            self.parameters["NumberOfEBSVol"] = str(len(self.__ebs_section))
+            self.parameters["NumberOfEBSVol"] = '%s' % len(self.__ebs_section)
```
fnubalaj: The config.get should already return a string.

PR #655, opened 2018-11-15
Fix docs words spelling

demartinofra:
tox -e docs-spellcheck


docs/spelling_wordlist.txt
```diff
@@ -0,0 +1,51 @@
+AWS
```
lukeseawalker: could we add AMI too?

docs/spelling_wordlist.txt
```diff
@@ -0,0 +1,51 @@
...
+online
+parallelcluster
+pem
+pre
+qhost
+qstat
+qsub
+sge
+signalled
+slurm
```
mauri-melato: why also lowercase? and not SGE, Torque?

docs/spelling_wordlist.txt
demartinofra: I think I can drop AWS since capitalized words are skipped.

docs/spelling_wordlist.txt
demartinofra: they refer to config values that are lowercase but I'll move them to the spelling section for the specific file (like the other config values) rather than having them in the global dict.

PR #701, opened 2018-11-19
Changes to support RAID

rexcsn:
If merged, the following changes will take place:
- Changed cfnconfig.py to parse all RAID related variables
- Changed config_sanity.py to check RAID related constraints
- Changed aws-parallelcluster.cfn.json to add new parameter RAIDoptions, with allowed pattern to validate input, and new substack RAIDSubstack, output RAID shared directory via dna.json to master and compute nodes, output RAID volume ids to master. Added new parameter RAIDSharedDir to Batch substack resource
- Changed batch-substack.cfn.json to set RAIDSharedDir as an environment variable so the RAID directory can be mounted
- Added util/generate-raid-substack.py, which generates raid substack template with troposphere
- Added raid-substack.cfn.json, generated with generate-raid-substack.py, supports creating up to 5 volumes
- Changed entrypoint.sh to include RAID directory mounting call
- Added integration test raid-test.py and test files config_raid and raid-check.sh, which tests edge cases of creating with the minimum number of 2 volumes and max number of 5 volumes, as well as the default case of not using RAID at all.
- Batch compatibility is manually tested by submitting a job under RAID shared directory

### DIFF
--- a/cli/pcluster/resources/batch/docker/scripts/entrypoint.sh
+++ b/cli/pcluster/resources/batch/docker/scripts/entrypoint.sh
@@ -39,6 +39,11 @@ if [[ "${PCLUSTER_EFS_FS_ID}" != "NONE" ]] && [[ ! -z "${PCLUSTER_AWS_REGION}" ]
   /parallelcluster/bin/mount_efs.sh "${PCLUSTER_EFS_FS_ID}" "${PCLUSTER_AWS_REGION}" "${PCLUSTER_EFS_SHARED_DIR}"
 fi
 
+# mount RAID via nfs
+if [[ ${PCLUSTER_RAID_SHARED_DIR} != "NONE" ]]; then
+  /parallelcluster/bin/mount_nfs.sh "${_master_ip}" "${PCLUSTER_RAID_SHARED_DIR}"
+fi
+
 # create hostfile if mnp job
 if [ -n "${AWS_BATCH_JOB_NUM_NODES}" ]; then
   /parallelcluster/bin/generate_hostfile.sh "${first_ebs_shared_dir}" "${HOME}"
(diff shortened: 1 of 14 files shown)


util/generate-raid-substack.py
```diff
@@ -0,0 +1,92 @@
...
+        if i == 0:
+            outputs[i] = If(use_vol[i], vol_to_return[i], "NONE")
+        else:
+            outputs[i] = If(use_vol[i], Join(",", vol_to_return[: (i + 1)]), outputs[i - 1])
+
+    t.add_output(
+        Output("Volumeids", Description="Volume IDs of the resulted RAID EBS volumes", Value=outputs[number_of_vol - 1])
+    )
+
+    json_file_path = "/Users/shuningc/Desktop/pcluster-github/cfncluster/cloudformation/raid-substack.cfn.json"
```
fnubalaj: This should be an input. Don't hardcode it.

cli/pcluster/resources/batch/docker/scripts/entrypoint.sh
```diff
@@ -18,6 +18,11 @@ if [[ ! -z "${EFS_FS_ID}" ]] && [[ ! -z "${AWS_REGION}" ]] && [[ "${EFS_SHARED_D
   /parallelcluster/bin/mount_efs.sh "${EFS_FS_ID}" "${AWS_REGION}" "${EFS_SHARED_DIR}"
 fi
 
+# mount RAID via nfs
+if [ ! -z ${RAID_SHARED_DIR} ]; then
```
fnubalaj: If there is always going to be a RAID_SHARED_DIR? If the user does not request RAID, will there be RAID_SHARED_DIR?

PR #708, opened 2018-11-22
Add possibility to export environment variables in the job environment

enrico-usai:
Usage:

```
  $ echo "env | sort" | awsbsub -jn test-env -e CUSTOM1,CUSTOM2
  $ echo "env | sort" | awsbsub -jn test-all-env -e all
  $ echo "env | sort" | awsbsub -jn test-blacklisted-env -e all -eb CUSTOM1,CUSTOM2
```

I also changed the s3 folder for the uploaded script from: `s3_bucket/batch/` to `s3_bucket/batch/jobid/`
This because now we have two files, one for the exported environment and the user script.

### DIFF
--- a/cli/awsbatch/awsbsub.py
+++ b/cli/awsbatch/awsbsub.py
@@ -15,9 +15,11 @@ from __future__ import print_function
 
 import datetime
 import os
+import pipes
 import re
 import shutil
 import sys
+import tempfile
 import time
 
 import argparse
@@ -45,53 +47,59 @@ def _get_parser():
     parser.add_argument(
         "-jn",
         "--job-name",
-        help="The name of the job. The first character must be alphanumeric, "
-        "and up to 128 letters (uppercase and lowercase), "
-        "numbers, hyphens, and underscores are allowed",
+        help="The name of the job. The first character must be alphanumeric, and up to 128 letters "
+        "(uppercase and lowercase), numbers, hyphens, and underscores are allowed",
     )
     parser.add_argument("-c", "--cluster", help="Cluster to use")
     parser.add_argument(
         "-cf",
         "--command-file",
-        help="Identifies that the command is a file to be transferred " "to the compute instances",
+        help="Identifies that the command is a file to be transferred to the compute instances",
         action="store_true",
     )
     parser.add_argument(
         "-p",
         "--vcpus",
-        help="The number of vCPUs to reserve for the container. "
-        "When used in conjunction with --nodes it identifies the number "
-        "of vCPUs per node. Default is 1",
+        help="The number of vCPUs to reserve for the container. When used in conjunction with --nodes it identifies "
+        "the number of vCPUs per node. Default is 1",
         type=int,
         default=1,
     )
     parser.add_argument(
         "-m",
         "--memory",
-        help="The hard limit (in MiB) of memory to present to the job. "
-        "If your job attempts to exceed the memory specified here, "
-        "the job is killed. Default is 128",
+        help="The hard limit (in MiB) of memory to present to the job. If your job attempts to exceed the memory "
+        "specified here, the job is killed. Default is 128",
         type=int,
         default=128,
     )
     parser.add_argument(
+        "-e",
+        "--env",
+        help="Comma separated list of environment variable names to export to the Job environment. "
+        "Use 'all' to export all the environment variables, except the ones listed to the --env-blacklist parameter "
+        "and variables starting with PCLUSTER_* and AWS_BATCH_* prefix.",
+    )
+    parser.add_argument(
+        "-eb",
+        "--env-blacklist",
+        help="Comma separated list of environment variable names to NOT export to the Job environment. "
+        "Default: HOME, PWD, USER, PATH, LD_LIBRARY_PATH, TERM, TERMCAP.",
+    )
+    parser.add_argument(
         "-r",
         "--retry-attempts",
-        help="The number of times to move a job to the RUNNABLE status. "
-        "You may specify between 1 and 10 attempts. "
-        "If the value of attempts is greater than one, "
-        "the job is retried if it fails "
-        "until it has moved to RUNNABLE that many times. "
-        "Default value is 1",
+        help="The number of times to move a job to the RUNNABLE status. You may specify between 1 and 10 attempts. "
+        "If the value of attempts is greater than one, the job is retried if it fails until it has moved to RUNNABLE "
+        "that many times. Default value is 1",
         type=int,
         default=1,
     )
     parser.add_argument(
         "-t",
         "--timeout",
-        help="The time duration in seconds (measured from the job attempt's "
-        "startedAt timestamp) after which AWS Batch terminates your jobs "
-        "if they have not finished. It must be at least 60 seconds",
+        help="The time duration in seconds (measured from the job attempt's startedAt timestamp) after which AWS "
+        "Batch terminates your jobs if they have not finished. It must be at least 60 seconds",
         type=int,
     )
     # MNP parameter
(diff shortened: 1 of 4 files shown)


cli/awsbatch/awsbsub.py
```diff
@@ -59,63 +58,63 @@ def _get_parser():
...
-        help="The hard limit (in MiB) of memory to present to the job. "
-        "If your job attempts to exceed the memory specified here, "
-        "the job is killed. Default is 128",
+        help="The hard limit (in MiB) of memory to present to the job. If your job attempts to exceed the memory "
+        "specified here, the job is killed. Default is 128",
         type=int,
         default=128,
     )
+    parser.add_argument(
+        "-e",
```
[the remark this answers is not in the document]
enrico-usai: My idea is that if you pass: `-e A,B,*` you are enforcing the cli to export A and B also if they are escaped.

So with `-e A,B,* -eb A,B` the A and B variables will be exported.

cli/awsbatch/awsbsub.py
```diff
@@ -348,6 +358,42 @@ def run(
...
+        Append environment variables to the environment dictionary by excluding blacklisted variables.
+
+        The variable will be added in the form {"name": "varname", "value": "varvalue"}
+
+        :param environment: dictionary to update
+        :param env_vars: list of variables to get from the environment and add to the dictionary
+        or use * to get all the environment variables except the blacklisted ones
+        :param blacklisted_vars: list of variable names to exclude when * is passed as env_vars parameter
+        """
+        environment_blacklist = ["HOME", "PWD", "USER", "PATH", "LD_LIBRARY_PATH", "TERM", "TERMCAP"]
```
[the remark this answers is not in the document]
enrico-usai: It is a starting point, please let me know the list of variables to blacklist, thank you.

cli/awsbatch/awsbsub.py
```diff
@@ -348,6 +358,42 @@ def run(
...
+        for var in env_vars.split(","):
+            var_name = var.strip()
+
+            if var_name == "*":
+                # export all the env variables except the blacklisted ones
+                for env_var in os.environ:
+                    if env_var not in blacklisted_vars:
+                        var_value = os.environ[env_var]
+                        environment.append({"name": env_var, "value": var_value})
+                        self.log.info("Exporting environment variable: (%s) value: (%s)." % (env_var, var_value))
```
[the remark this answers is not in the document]
enrico-usai: It is not so readable because `environment` is a json and there are a lot of escaping

cli/awsbatch/awsbsub.py
```diff
@@ -348,6 +358,42 @@ def run(
...
+        :param blacklisted_vars: list of variable names to exclude when * is passed as env_vars parameter
+        """
+        environment_blacklist = ["HOME", "PWD", "USER", "PATH", "LD_LIBRARY_PATH", "TERM", "TERMCAP"]
+
+        blacklisted_vars = (
+            [var.strip() for var in env_blacklist_vars.split(",")] if env_blacklist_vars else environment_blacklist
+        )
+
+        self.log.info("Environment blacklist is (%s)", blacklisted_vars)
+        for var in env_vars.split(","):
```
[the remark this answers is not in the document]
enrico-usai: The alternative could be `for var in [v.strip() for v in env_vars.split(",")]` but it is less efficient.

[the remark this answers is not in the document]
enrico-usai: The alternative could be `for var in [v.strip() for v in env_vars.split(",")]` but it is less efficient.

cli/awsbatch/awsbsub.py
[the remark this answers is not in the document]
enrico-usai: I changed the patch, now it's only possible to pass `-e all` or a list of env variables.

cli/awsbatch/awsbsub.py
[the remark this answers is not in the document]
enrico-usai: We choose to filter out PCLUSTER_* variables. This prefix must be used for reserved variables.

cli/awsbatch/awsbsub.py
```diff
@@ -179,67 +189,50 @@ def _upload_to_s3(boto3_factory, s3_bucket, file_path, key_name, timeout):
...
-    if args.command_file or not sys.stdin.isatty():
+    if args.command_file or not sys.stdin.isatty() or args.env:
         # define job script name
-        job_script = "job-{0}-{1}.sh".format(job_name, int(time.time() * 1000))
+        job_id = "job-{0}-{1}".format(job_name, int(time.time() * 1000))
+        job_folder = _get_job_folder(job_id)
+        job_script = job_id + ".sh"
         log.info("Using command-file option or stdin. Job script name: %s" % job_script)
 
+        env_file = job_id + ".env.sh"
```
demartinofra: What about leaving env_file to None when not used rather than repeating the check on args.env in the _compose_bash_command function?

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+    Create file from STDIN and upload to S3.
+
+    :param boto3_factory: initialized Boto3ClientFactory object
+    :param s3_bucket: S3 bucket
+    :param job_folder: S3 bucket subfolder
+    :param job_script: job script name
+    :param timeout: S3 expiration time in seconds
+    """
+    try:
+        # copy stdin to temporary file
```
demartinofra: what about replacing this comment with a function?

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+        with os.fdopen(sys.stdin.fileno(), "rb") as src:
+            with open(job_script, "wb") as dst:
+                shutil.copyfileobj(src, dst)
+
+        _upload_to_s3(boto3_factory, s3_bucket, job_script, job_folder + job_script, timeout)
```
demartinofra: I'm totally fine with leaving it as it is but this function could become part of an S3Uploader class that holds the factory and the bucket as part of its attributes so that these don't need to be passed around at every invocation.

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+    except (OSError, Exception) as e:
```
demartinofra: OSError should be already included in the broader Exception

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+    :param boto3_factory: initialized Boto3ClientFactory object
+    :param s3_bucket: S3 bucket
+    :param job_folder: S3 bucket subfolder
+    :param job_script: job script name
+    :param timeout: S3 expiration time in seconds
+    """
+    try:
+        # copy stdin to temporary file
+        with os.fdopen(sys.stdin.fileno(), "rb") as src:
+            with open(job_script, "wb") as dst:
```
demartinofra: we could use a temporary file and leverage the ContextManager for clean-up: `with tempfile.TemporaryFile() as tmp:`

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+    :param s3_bucket: S3 bucket
+    :param job_folder: S3 bucket subfolder
+    :param env_file: environment file name
+    :param timeout: S3 expiration time in seconds
+    :param log: log
+    """
+    key_value_list = _get_env_key_value_list(env, log, env_blacklist)
+    try:
+        # copy env to temporary file
+        with open(env_file, "wb") as dst:
```
demartinofra: Same here for the temporary file

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+            dst.write("\n".join(key_value_list) + "\n")
+
+        _upload_to_s3(boto3_factory, s3_bucket, env_file, job_folder + env_file, timeout)
+    except (OSError, Exception) as e:
```
demartinofra: Same here for Exception

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+    :param s3_bucket: S3 bucket
+    :param region: AWS region
+    :param job_folder: S3 bucket subfolder
+    :param job_script: job script file
+    :param env_file: environment file
+    :return: composed bash command
+    """
+    command_args = shell_join(args.arguments)
+    # download awscli, if required.
+    bash_command = (
```
demartinofra: I was wondering if we should pipe the different commands with && rather than ; so that we break if a command fails

cli/awsbatch/awsbsub.py
```diff
@@ -249,6 +242,153 @@ def _upload_and_get_command(boto3_factory, args, job_name, region, s3_bucket, lo
...
+
+    Skip internal variables and functions.
+
+    :param key_value_list: list to update
+    :param var_name: var name
+    :param log: log file
+    """
+    # exclude reserved variables and functions
+    if (
+        not var_name.upper().startswith("PCLUSTER_")
```
demartinofra: Small comment on why we are excluding each of these?

var_name.upper could ve called just once before if statement

else: log.warn variable discarded/excluded?

cli/awsbatch/awsbsub.py
enrico-usai: I totally agree

PR #724, opened 2018-11-30
awsbstat: optimize calls to describe-jobs in detailed mode

demartinofra:
The patch contains 3 optimisations:

1. Reduce the amount of calls to describe-jobs when awsbstat is used in detailed mode by calling __add_jobs function only after all jobs are retrieved.
In order to keep jobs sorted by status in the output table, output is now sorted right before being displayed rather than relying on the order given by the api calls performed.

2. Avoid a duplicate call to describe-jobs when awsbstat is called with a list of job ids.

3. Replace list_jobs calls with a single call to describe_jobs when expanding arrays.

### DIFF
--- a/cli/awsbatch/awsbstat.py
+++ b/cli/awsbatch/awsbstat.py
@@ -23,6 +23,8 @@ import argparse
 from awsbatch.common import AWSBatchCliConfig, Boto3ClientFactory, Output, config_logger
 from awsbatch.utils import convert_to_date, fail, get_job_definition_name_by_arn, is_job_array, shell_join
 
+AWS_BATCH_JOB_STATUS = ["SUBMITTED", "PENDING", "RUNNABLE", "STARTING", "RUNNING", "SUCCEEDED", "FAILED"]
+
 
 def _get_parser():
     """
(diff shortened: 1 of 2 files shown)


cli/awsbatch/awsbstat.py
```diff
@@ -172,7 +174,10 @@ def run(self, job_status, expand_arrays, job_queue=None, job_ids=None, show_deta
         if details_required:
             self.output.show()
         else:
-            self.output.show_table(["jobId", "jobName", "status", "startedAt", "stoppedAt", "exitCode"])
+            table_keys = ["jobId", "jobName", "status", "startedAt", "stoppedAt", "exitCode"]
+            self.output.show_table(
+                keys=table_keys, sort_keys_function=lambda x: AWS_BATCH_JOB_STATUS.index(x[table_keys.index("status")])
```
enrico-usai: Could you add a comment for this lambda function?

cli/awsbatch/awsbstat.py
```diff
@@ -172,60 +174,66 @@ def run(self, job_status, expand_arrays, job_queue=None, job_ids=None, show_deta
...
+            if expanded_job_array_ids:
+                jobs = self.__chunked_describe_jobs(expanded_job_array_ids)
+                # forcing details to be False since already retrieved.
+                self.__add_jobs(jobs)
         except Exception as e:
-            fail("Error listing job array children for job (%s). Failed with exception: %s" % (job_array_id, e))
+            fail("Error listing job array children. Failed with exception: %s" % e)
 
-    def __add_jobs(self, jobs, details):  # noqa: C901 FIXME
+    def __chunked_describe_jobs(self, job_ids):
```
enrico-usai: Minor: We should add a comment to this method too.

We could explain that describe_jobs accepts a limit of 100 jobs.

demartinofra: Done!

PR #735, opened 2018-12-04
Update issue templates

mauri-melato:


.github/ISSUE_TEMPLATE/issue-report.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Issue report
```
demartinofra: What's the difference between the bug report and the issue report? Shall we just keep the more generic issue report template?

mauri-melato: Yes I meant to only keep the issue term.

PR #754, opened 2018-12-11
Get MASTER_IP internally and avoid to use the variable set by awsbsub

enrico-usai:
- In future we could remove the MASTER_IP setting from the awsbsub. We are keeping it to avoid to broke the behaviour with existing clusters.

- Use PCLUSTER_ prefix for internal variables


cloudformation/batch-substack.cfn.json
```diff
@@ -145,6 +145,25 @@
...
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Action": [
+                    "cloudformation:DescribeStacks"
+                  ],
+                  "Effect": "Allow",
+                  "Resource": [
+                    {
+                      "Fn::Sub": "arn:${AWS::Partition}:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/${ClusterName}*"
```
mauri-melato: why the final '*' ?

enrico-usai: Without the `*` I have this error:
```
An error occurred (AccessDenied) when calling the DescribeStacks operation: 
User: arn:aws:sts::<accountid>:assumed-role/parallelcluster-<name>-AWSBatchStack-CCW1ZPE-JobRole-XK7BOLJ50HZR/2ea928bf-31ab-4672-aa9b-9af415480909 
is not authorized to perform: cloudformation:DescribeStacks on resource: 
arn:aws:cloudformation:eu-west-1:499649255560:stack/parallelcluster-<name>/*
```

I changed it to: `...${ClusterName}/*`

PR #758, opened 2018-12-11
Patch for upload template logic

rexcsn:
- Modified upload-cfn-template.py so that the program checks for override flag before dryrun flag, this way we can check if the template already exist even in dryrun mode


util/upload-cfn-templates.py
```diff
@@ -38,6 +34,10 @@ def upload_to_s3(args, region):
                     continue
```
fnubalaj: Don't continue here. If you continue, what happens when 'dryrun' and 'override' both are enabled.

In a case where both dry run and override are enabled, there should be a clear indication of which buckets needoverriding

util/upload-cfn-templates.py
```diff
@@ -38,6 +34,10 @@ def upload_to_s3(args, region):
...
                 except ClientError:
                     pass
+            if args.dryrun:
+                key = key_path + "%s-%s.cfn.json" % (t, args.version)
+                print("Skipping upload %s to s3://%s/%s" % (template_name, bucket, key))
```
fnubalaj: Instead of saying skipping upload, say "If dryrun is disabled, %s will be uploaded to s3://%s/%s"

util/upload-cfn-templates.py
```diff
@@ -26,45 +27,48 @@ def upload_to_s3(args, region):
...
-                    res = b.put_object(Body=data, ACL="public-read", Key=key)
-                    print(res)
-                else:
-                    print("Couldn't upload %s to bucket s3://%s/%s" % (template_name, bucket, key))
-                    if e.response["Error"]["Code"] == "NoSuchBucket":
-                        print("Bucket is not present.")
+        for bucket in buckets:
+            if args.dryrun:
+                key = key_path + "%s-%s.cfn.json" % (t, args.version)
+                print("Skipping upload %s to s3://%s/%s" % (template_name, bucket, key))
```
fnubalaj: Change message

PR #760, opened 2018-12-12
awsbstat: show parent when passing ids

demartinofra:
* `awsbstat —expand-children` - returns all jobs but only children in case of arrays and mnps
* `awsbstat <job-array-id> (o <job-mnp-id>)` - returns the table with parents and relative children
* `awsbstat -d <job-array-id> (o <job-mnp-id>)` - same as above but with a detailed view
* In case of multiple job-ids we keep the same conventions listed above

Ordering:
* In case we specify job-ids the order of the output follows the list of ids provided as input.
* In case we don't specify job-ids the order is by (status, startedAt, jobId)
* Order applies both to tabular and detailed view


cli/awsbatch/awsbstat.py
```diff
@@ -299,22 +300,59 @@ def run(self, job_status, expand_children, job_queue=None, job_ids=None, show_de
...
+
+        :param ordered_keys: list containing the sorted keys.
+        :return: a function to be used as key argument of the sorted function.
+        """
+
+        def _sort_by_key(item):
+            job_id = item.id
+            try:
+                # in case the parent id was provided as input, sort children based on parent id position
+                job_position = ordered_keys.index(job_id[: AWSBstatCommand.__JOB_ID_LENGTH])
```
enrico-usai: We could split by space instead of using a fixed length

demartinofra: Replaced with: `parent_id = re.findall(r"[\w-]+", job_id)[0]`

PR #763, opened 2018-12-13
Patch for simplifying efs-check

rexcsn:
- Modified efs-check.sh to simplify logic, since only have 1 efs path, remove logic originally used for multiple path

### DIFF
--- a/tests/efs-check.sh
+++ b/tests/efs-check.sh
@@ -2,50 +2,45 @@
 
 set -e
 
-paths=${1}
+path=${1}
 fs_id=${2}
-paths=$(echo ${paths} | tr "," " ")
 status=0
 
-pathCount=$(cat /etc/fstab | grep -E "$(echo ${paths} | tr " " "|")" | wc -l | xargs)
-fsGrep=$(cat /etc/fstab | grep -E "fs-" | wc -l | xargs)
-truePathCount=$(echo ${paths} | tr " " "\n"| wc -l |xargs)
+path_exist=$(cat /etc/fstab | grep "$(echo ${path})";echo $?)
+fs_grep=$(cat /etc/fstab | grep "fs-";echo $?)
 
-if [ ${pathCount} == ${truePathCount} ] && [ ${fsGrep} == ${truePathCount} ]; then
-    echo "found the file system"
+if [[ ${path_exist} != 1 ]] && [[ ${fs_grep} != 1 ]]; then
+  echo "found the file system"
 else
-    echo "error: file system not found"
-    exit 1
+  echo "error: file system not found"
+  exit 1
 fi
 
 if [[ ${fs_id} ]]; then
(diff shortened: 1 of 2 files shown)


tests/efs-check.sh
```diff
@@ -2,50 +2,45 @@
...
+path=${1}
 fs_id=${2}
-paths=$(echo ${paths} | tr "," " ")
 status=0
 
-pathCount=$(cat /etc/fstab | grep -E "$(echo ${paths} | tr " " "|")" | wc -l | xargs)
-fsGrep=$(cat /etc/fstab | grep -E "fs-" | wc -l | xargs)
-truePathCount=$(echo ${paths} | tr " " "\n"| wc -l |xargs)
+path_exist=$(cat /etc/fstab | grep "$(echo ${path})";echo $?)
+fs_grep=$(cat /etc/fstab | grep "fs-";echo $?)
```
fnubalaj: Why are you relying on $? instead of if -z $fs_grep?

tests/efs-check.sh
```diff
@@ -2,50 +2,45 @@
...
+compute_check=$(ls ${path} | grep "compute_file";echo $?)
 
-    if [[ ${computeCheck} ]]; then
-        echo "${path} is correctly mounted on compute"
-        rm ${path}/compute_file
-    else
-        echo "error: ${path} is not found on compute"
-        status=1
-    fi
+if [[ ${compute_check} != 1 ]]; then
```
fnubalaj: Is the exit code guaranteed to 0 or 1?

tests/efs-test.py
```diff
@@ -132,10 +132,10 @@ def prepare_subnets(vpcrelated, region):
     )
     subnets["a2"] = response_a2["Subnet"]["SubnetId"]
 
-    response_b = ec2.create_subnet(
-        AvailabilityZone=region + "b", CidrBlock="177.31.16.0/20", VpcId=vpcrelated["vpc_id"]
+    response_c = ec2.create_subnet(
+        AvailabilityZone=region + "c", CidrBlock="177.31.16.0/20", VpcId=vpcrelated["vpc_id"]
     )
-    subnets["b"] = response_b["Subnet"]["SubnetId"]
+    subnets["c"] = response_c["Subnet"]["SubnetId"]
```
fnubalaj: Will this work in all regions?

tests/efs-check.sh
rexcsn: I had to do echo $? because if grep does not find a match it will not return an empty string but rather it will exit 1 and the shell script will stop right on the failure, so I had to do that in order to keep the script running and print error messages.

tests/efs-check.sh
rexcsn: grep returns 1 if it cannot find the message and 0 if it can, but I check for != 1 because if grep finds a match the variable will be "matchstring"0, if it fails to find a match it would just be 1

PR #764, opened 2018-12-13
Tag Instances

sean-smith:
### Overview

Adds two new tags to ec2 instances launched with ParallelCluster:

```
aws-parallelcluster-attributes = base_os, scheduler, version 
aws-parallelcluster-filesystem = efs=true/false, multiebs=[1-5], raid=true/false
```

### Testing

1. Created an awsbatch cluster, verified that both master and compute instances are tagged
2. Created a cluster with traditional scheduler, verified both master and compute are properly tagged


sean-smith:
Corresponding cookbook changes in: https://github.com/aws/aws-parallelcluster-cookbook/pull/234

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1883,6 +1883,47 @@
...
+          {
+            "Key": "aws-parallelcluster-filesystem",
+            "Value": {
+              "Fn::Sub": [
+                "efs=${efs}, multiebs=${NumberOfEBSVol}, raid=${raid}",
+                {
+                  "efs": {
+                    "Fn::If": [
+                      "CreateEFSSubstack",
+                      "true",
```
demartinofra: What if we use 0 and 1 instead of true and false? It might be easier to query, aggregate and plot the values, isn't it?

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -2465,6 +2506,49 @@
             "Key": "Name",
```
demartinofra: For consistency shall we add the Application tag also to the compute nodes?

cloudformation/batch-substack.cfn.json
```diff
@@ -270,6 +278,14 @@
                 "Ref": "AWS::NoValue"
               }
             ]
+          },
+          "Tags": {
```
demartinofra: Shall we also add the Application (parallelcluster-$ClusterName) and the Name (something like AWS Batch Compute) tags?

cloudformation/aws-parallelcluster.cfn.json
sean-smith: changed

cloudformation/aws-parallelcluster.cfn.json
sean-smith: Added 👍

PR #786, opened 2018-12-19
Add partition parameter and remove Paris (eu-west-3) from unsupported

lukeseawalker:

### DIFF
--- a/util/batch-instance-whitelist.py
+++ b/util/batch-instance-whitelist.py
@@ -26,12 +26,24 @@ import sys
 import boto3
 from botocore.exceptions import ClientError
 
-# commercial regions unsupported by aws batch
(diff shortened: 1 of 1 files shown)


util/batch-instance-whitelist.py
```diff
@@ -113,8 +116,18 @@ def main(args):
     parser.add_argument("--dryrun", type=str, help="Doesn't push anything to S3, just outputs", required=True)
     args = parser.parse_args()
 
+    if args.partition == "commercial":
```
demartinofra: I would move this selection of the region to the get_all_aws_regions function

lukeseawalker: done

PR #790, opened 2018-12-19
Refactor logic

lukeseawalker:
The previous logic was exiting the script at the first occurrence of an
object already present into a bucket, without considering the dryrun flag

The new logic put the object to S3 only when at least one of the two
case are verified:
- the object already exist but the override is true and dryrun false
- the object do not exist and dryrun is false (independently from the
value of the override flag)

### DIFF
--- a/util/upload-cfn-templates.py
+++ b/util/upload-cfn-templates.py
@@ -22,6 +22,33 @@ def get_all_aws_regions(partition):
     return set(sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions")))
 
 
+def put_object_to_s3(s3_client, bucket, key, region, data, template_name):
+    try:
+        object = s3_client.Object(bucket, key)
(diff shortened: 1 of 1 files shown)


util/upload-cfn-templates.py
```diff
@@ -36,49 +63,23 @@ def upload_to_s3(args, region):
...
-                        print(res)
-                    else:
-                        print("Couldn't upload %s to bucket s3://%s/%s" % (template_name, bucket, key))
-                        if e.response["Error"]["Code"] == "NoSuchBucket":
-                            print("Bucket is not present.")
-                        else:
-                            raise e
-                    pass
+                print(
+                    "Not uploading %s to bucket %s, override is %s, dryrun is %s"
```
mauri-melato: I'd also add here, "object exists true|false", to have everything in one line here

PR #841, opened 2019-01-31
Run as ec2-user

sean-smith:
* On the host system, user `ec2-user` has UID 500. By creating `ec2-user`
with UID 500 in the container, we can access files shared via nfs, owned
by `ec2-user`. This is specific to amazonlinux machines.
* The `mount_nfs.sh` script needs to have root permissions to execute `rpcbind`
and the `mount` commands. This means that we need to either run the script as
root or switch to root before running that script. I opted for the latter option.
* The `/home` directory when mounted overwrites `/home/ec2-user` which is the
traditional home directrory for user `ec2-user`. Hence I changed the home directory
to `/opt/ec2-user`.


cli/pcluster/resources/batch/docker/alinux/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM amazonlinux:latest
 
-ENV USER root
+ENV USER ec2-user
+ENV HOME /opt/ec2-user
```
enrico-usai: I'd add a comment in the code about the home overwrite, like you did in the PR.

cli/pcluster/resources/batch/docker/alinux/Dockerfile
```diff
@@ -35,9 +37,11 @@ RUN sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /etc/ssh/ss
     && echo "export VISIBLE=now" >> /etc/profile
 
 # create ssh keys
```
enrico-usai: I would also add here a comment about the reason we are creating the ec2-user.

cli/pcluster/resources/batch/docker/alinux/Dockerfile
```diff
@@ -35,9 +37,11 @@ RUN sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /etc/ssh/ss
...
-ENV SSHDIR /root/.ssh
+ENV SSHDIR ${HOME}/.ssh
 RUN echo "${USER} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers \
```
demartinofra: do we want to add ec2-user to the list of sudoers?

sean-smith:
@demartinofra 

> Hostfile is still owned by root. Can this be a problem for mpi jobs?

I've added a `chown -R hostfile`

> If I try to submit an mpi job (and I think this applies to every job that has a command file) I get the following error: `fatal error: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied`

This was because `su - ${USER} -c "$@"` wipes out the environment variables, by switching to `su -c "$@*" ${USER}` it preserves environment variables. Downside is I have to manually `cd ${HOME}`

> Is passwordless ssh to the other containers in an MNP job working for ec2-user? I'm afraid this also can be a problem for MPI jobs.

Fixed by adding the to the `/etc/ssh/sshd_config` file the lines:
```
AuthorizedKeysFile    ${HOME}/.ssh/authorized_keys
HostKey ${HOME}/.ssh/ssh_host_rsa_key
```

> Did you verify if the mounted EBS, RAID and EFS are still working for ec2-user since they get mounted as root? I expect them to work since you said you verified the shared NFS volume.

See testing above ^

> ENV PATH "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/usr/lib64/openmpi/bin/" is now set for root and not ec2-user

Preserved now, see above.

cli/pcluster/resources/batch/docker/scripts/entrypoint.sh
```diff
@@ -50,5 +50,5 @@ if [ -n "${AWS_BATCH_JOB_NUM_NODES}" ]; then
 fi
 
 # run the user's script
-echo "Starting the job..."
-exec "$@"
+echo "Starting the job as ${USER}..."
+su - ${USER} -c "$@"
```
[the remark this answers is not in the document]
demartinofra: There might be still an issue with this cause if I submit: `awsbsub -n 2 -cf test.sh` I get the following error:

```
2019-02-08T11:31:43+00:00: Starting the job as ec2-user...
2019-02-08T11:31:43+00:00: mkdir: missing operand
2019-02-08T11:31:43+00:00: Try 'mkdir --help' for more information.
```

Also could you verify that the environment variables passed with the -e awsbsub option are visible to the ec2-user, although I'm pretty confident they are given how we source them.

cli/pcluster/resources/batch/docker/scripts/entrypoint.sh
```diff
@@ -50,5 +50,11 @@ if [ -n "${AWS_BATCH_JOB_NUM_NODES}" ]; then
 fi
 
 # run the user's script
-echo "Starting the job..."
-exec "$@"
+echo "Starting the job as ${USER}..."
+cd ${HOME}
+cat << EOF | su ${USER} -c "/bin/sh"
```
demartinofra: I believe this was some debugging code and needs to be removed right?

PR #849, opened 2019-02-08
Implementation of a framework to write and orchestrate integration tests

demartinofra:
See documentation: https://github.com/aws/aws-parallelcluster/blob/b8ff268c108a45bc09eb52fc3508a0cb6d5ecc8a/tests/integration-tests/README.md

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -1,16 +1,77 @@
 import logging
 
+import boto3
+import configparser
 from retrying import retry
 
 from utils import random_alphanumeric, retry_if_subprocess_error, run_command
 
 
+class Cluster:
+    """Contain all static and dynamic data related to a cluster instance."""
+
+    def __init__(self, name, config_file, ssh_key):
+        self.name = name
+        self.config_file = config_file
+        self.ssh_key = ssh_key
+        self.config = configparser.ConfigParser()
+        self.config.read(config_file)
+        self.__cfn_outputs = None
+
+    @property
+    def cfn_name(self):
+        """Return the name of the CloudFormation stack associated to the cluster."""
+        return "parallelcluster-" + self.name
+
+    @property
+    def region(self):
+        """Return the aws region the cluster is created in."""
+        return self.config.get("aws", "aws_region_name", fallback="us-east-1")
+
+    @property
+    def master_ip(self):
+        """Return the public ip of the cluster master node."""
+        return self.cfn_outputs["MasterPublicIP"]
+
+    @property
+    def os(self):
+        """Return the os used for the cluster."""
+        cluster_template = self.config.get("global", "cluster_template", fallback="default")
+        return self.config.get("cluster {0}".format(cluster_template), "base_os", fallback="alinux")
+
+    @property
+    def cfn_outputs(self):
+        """
+        Return the CloudFormation stack outputs for the cluster.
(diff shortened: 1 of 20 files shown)


tests/integration-tests/clusters_factory.py
```diff
@@ -0,0 +1,136 @@
...
+        cluster_template = self.config.get("global", "cluster_template", fallback="default")
+        return self.config.get("cluster {0}".format(cluster_template), "base_os", fallback="alinux")
+
+    @property
+    def cfn_outputs(self):
+        """
+        Return the CloudFormation stack outputs for the cluster.
+        Outputs are retrieved only once and then cached.
+        """
+        if self.__cfn_outputs:
```
enrico-usai: This can be simplified as:

```python
if not self.__cfn_outputs:
    self.__cfn_outputs = self.__retrieve_cfn_outputs()
return self.__cfn_outputs
```

tests/integration-tests/clusters_factory.py
```diff
@@ -0,0 +1,136 @@
...
+            name = "integ-tests-" + random_alphanumeric()
+
+        if name in self.__created_clusters:
+            raise ValueError("Cluster {0} already exists".format(name))
+
+        # create the cluster
+        logging.info("Creating cluster {0} with config {1}".format(name, config))
+        self.__created_clusters[name] = Cluster(name, config, self.__ssh_key)
+        result = run_command(["pcluster", "create", "--config", config, name])
+        if "CREATE_COMPLETE" not in result.stdout:
```
enrico-usai: You could define an enum for the cluster status.

tests/integration-tests/conftest.py
```diff
@@ -0,0 +1,195 @@
...
+    DIMENSIONS_MARKER_ARGS,
+    add_default_markers,
+    check_marker_dimensions,
+    check_marker_list,
+    check_marker_skip_dimensions,
+    check_marker_skip_list,
+)
+from jinja2 import Environment, FileSystemLoader
+
+
```
enrico-usai: I'd add a comment on top to understand the purpose of this important file.

tests/integration-tests/test_runner.py
```diff
@@ -0,0 +1,279 @@
...
+
+LOGS_DIR = "{0}.logs".format(START_TIME)
+OUT_DIR = "{0}.out".format(START_TIME)
+
+TEST_DEFAULTS = {
+    "parallelism": None,
+    "retry_on_failures": False,
+    "features": "",  # empty string means all
+    "regions": [
+        "us-east-1",
```
enrico-usai: What about gov-cloud and other partitions?

demartinofra: regions will be selected explicitly when running in the other partitions with the --regions option. It doesn't make much sense to have regions belonging to different partitions in the default since in the end the account used for testing can't have cross partition access.

tests/integration-tests/conftest.py
demartinofra: done!

tests/integration-tests/clusters_factory.py
demartinofra: done!

tests/integration-tests/clusters_factory.py
demartinofra: good idea. I'll apply this in a later review

PR #850, opened 2019-02-08
SSH Improvements

sean-smith:
### Description
* SSH using private IP when public ip isn't accessible
* This pulls in logic from https://github.com/aws/aws-parallelcluster/pull/350
to ssh into the MasterServer when the stack is creating, updating,
or stack has rolled back. This is particulary useful to diagnose cookbook errors.

### Testing
1. SSH into a cluster with `use_public_ips = false`  while MasterServer is in `CREATE_IN_PROGRESS`
2. SSH into a cluster with `use_public_ips = false`  while MasterServer is in `CREATE_COMPLETE`
3. SSH into a cluster while stack is in `DELETE_IN_PROGRESS`

### DIFF
--- a/cli/pcluster/pcluster.py
+++ b/cli/pcluster/pcluster.py
@@ -577,7 +577,59 @@ def instances(args):
         LOGGER.info("Run 'awsbhosts --cluster %s' to list the compute instances", args.cluster_name)
 
 
-def command(args, extra_args):
+def _get_master_server_ip(stack_name, config):
+    """
+    Get the IP Address of the MasterServer.
+
+    :param stack_name: The name of the cloudformation stack
+    :param config: Config object
+    :return private/public ip address
+    """
+    ec2 = boto3.client(
+        "ec2",
+        region_name=config.region,
+        aws_access_key_id=config.aws_access_key_id,
+        aws_secret_access_key=config.aws_secret_access_key,
+    )
+
+    master_id = get_master_server_id(stack_name, config)
+    if master_id is []:
+        LOGGER.info("MasterServer not running. Can't SSH")
+        sys.exit(1)
+    instance = ec2.describe_instances(InstanceIds=[master_id]).get("Reservations")[0].get("Instances")[0]
+    ip_address = instance.get("PublicIpAddress")
+    if ip_address is None:
+        ip_address = instance.get("PrivateIpAddress")
+    state = instance.get("State").get("Name")
+    if state != "running" or ip_address is None:
+        LOGGER.info("MasterServer: %s\nCannot get ip address.", state.upper())
+        sys.exit(1)
+    return ip_address
+
+
+def _get_output_value(outputs, key_name):
+    """
+    Get output value from Cloudformation Stack Output.
+
+    :param outputs: Cloudformation Stack Outputs
+    :param key_name: Output Key
+    :return: OutputValue if that output exists, otherwise None
+    """
+    return next((o.get("OutputValue") for o in outputs if o.get("OutputKey") == key_name), None)
+
+
+def _get_param_value(params, key_name):
+    """
+    Get parameter value from Cloudformation Stack Parameters.
+
+    :param outputs: Cloudformation Stack Parameters
+    :param key_name: Parameter Key
+    :return: ParameterValue if that parameter exists, otherwise None
+    """
+    return next((i.get("ParameterValue") for i in params if i.get("ParameterKey") == key_name), None)
+
+
+def command(args, extra_args):  # noqa: C901 FIXME!!!
     stack = "parallelcluster-" + args.cluster_name
     config = cfnconfig.ParallelClusterConfig(args)
     if args.command in config.aliases:
@@ -595,12 +647,34 @@ def command(args, extra_args):
         stack_result = cfn.describe_stacks(StackName=stack).get("Stacks")[0]
         status = stack_result.get("StackStatus")
         valid_status = ["CREATE_COMPLETE", "UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE"]
-        if status not in valid_status:
-            LOGGER.info("Stack status: %s. Stack needs to be in %s", status, " or ".join(valid_status))
+        invalid_status = ["DELETE_COMPLETE", "DELETE_IN_PROGRESS"]
+
+        if status in invalid_status:
+            LOGGER.info("Stack status: %s. Cannot SSH while in %s", status, " or ".join(invalid_status))
             sys.exit(1)
-        outputs = stack_result.get("Outputs")
-        username = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "ClusterUser"][0]
-        ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPublicIP"][0]
+        elif status in valid_status:
+            outputs = stack_result.get("Outputs")
+            username = _get_output_value(outputs, "ClusterUser")
+            ip = (
+                _get_output_value(outputs, "MasterPublicIP")
+                if _get_output_value(outputs, "MasterPublicIP")
+                else _get_output_value(outputs, "MasterPrivateIP")
+            )
+
+            if not username:
+                LOGGER.info("Failed to get cluster %s username.", args.cluster_name)
+                sys.exit(1)
+
+            if not ip:
+                LOGGER.info("Failed to get cluster %s ip.", args.cluster_name)
+                sys.exit(1)
+        else:
+            # Stack is in CREATING, CREATED_FAILED, or ROLLBACK_COMPLETE but MasterServer is running
+            ip = _get_master_server_ip(stack, config)
+            template = cfn.get_template(StackName=stack)
+            mappings = template.get("TemplateBody").get("Mappings").get("OSFeatures")
+            base_os = _get_param_value(stack_result.get("Parameters"), "BaseOS")
+            username = mappings.get(base_os).get("User")
 
         try:
             from shlex import quote as cmd_quote


cli/pcluster/pcluster.py
```diff
@@ -576,6 +576,37 @@ def instances(args):
...
+        return ip_address
+    except ClientError as e:
+        LOGGER.critical(e.response.get('Error').get('Message'))
+        sys.stdout.flush()
+        sys.exit(1)
+    except KeyboardInterrupt:
+        LOGGER.info('\nExiting...')
+        sys.exit(0)
+
+def get_head_user(parameters, template):
```
enrico-usai: Could you add the docstring?

cli/pcluster/pcluster.py
```diff
@@ -576,6 +576,37 @@ def instances(args):
...
+        instance = ec2.describe_instances(InstanceIds=[master_id]) \
+            .get('Reservations')[0] \
+            .get('Instances')[0]
+        ip_address = instance.get('PublicIpAddress')
+        state = instance.get('State').get('Name')
+        if state != 'running' or ip_address is None:
+            LOGGER.info("MasterServer: %s\nCannot get ip address." % state.upper)
+            sys.exit(1)
+        return ip_address
+    except ClientError as e:
```
enrico-usai: We could avoid to catch and log for the `ClientError` and the `KeyboardInterrupt` exceptions since the `get_master_server_ip` is called inside the try/catch block of the `command` function that already catches these exceptions.

cli/pcluster/pcluster.py
```diff
@@ -595,12 +626,25 @@ def command(args, extra_args):
...
+        elif status in valid_status:
+            outputs = stack_result.get("Outputs")
+            username = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "ClusterUser"][0]
+
+            try:
+                ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPublicIP"][0]
+            except IndexError:
+                # If they're not using public ip's use the private ip instead
+                ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPrivateIP"][0]
+        else:
```
enrico-usai: Could you add a comment for the `else` block?

cli/pcluster/pcluster.py
```diff
@@ -576,6 +576,37 @@ def instances(args):
...
+    except ClientError as e:
+        LOGGER.critical(e.response.get('Error').get('Message'))
+        sys.stdout.flush()
+        sys.exit(1)
+    except KeyboardInterrupt:
+        LOGGER.info('\nExiting...')
+        sys.exit(0)
+
+def get_head_user(parameters, template):
+    mappings = template.get("TemplateBody") \
```
enrico-usai: Minor: this fits in one single line.

cli/pcluster/pcluster.py
sean-smith: I removed the `get_head_user` function completely

cli/pcluster/pcluster.py
sean-smith: fixed 👍🏻

cli/pcluster/pcluster.py
```diff
@@ -595,12 +626,25 @@ def command(args, extra_args):
...
             sys.exit(1)
-        outputs = stack_result.get("Outputs")
-        username = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "ClusterUser"][0]
-        ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPublicIP"][0]
+        elif status in valid_status:
+            outputs = stack_result.get("Outputs")
+            username = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "ClusterUser"][0]
+
+            try:
+                ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPublicIP"][0]
```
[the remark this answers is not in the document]
sean-smith: done

cli/pcluster/pcluster.py
sean-smith: sure

cli/pcluster/pcluster.py
```diff
@@ -577,7 +577,36 @@ def instances(args):
...
+        aws_access_key_id=config.aws_access_key_id,
+        aws_secret_access_key=config.aws_secret_access_key,
+    )
+
+    master_id = get_master_server_id(stack_name, config)
+    if master_id is []:
+        LOGGER.info("MasterServer not running. Can't SSH")
+        sys.exit(1)
+    instance = ec2.describe_instances(InstanceIds=[master_id]).get("Reservations")[0].get("Instances")[0]
+    ip_address = instance.get("PublicIpAddress")
```
enrico-usai: Should we add the fallback logic to the PrivateIpAddress here too?

cli/pcluster/pcluster.py
```diff
@@ -577,7 +577,36 @@ def instances(args):
         LOGGER.info("Run 'awsbhosts --cluster %s' to list the compute instances", args.cluster_name)
 
 
-def command(args, extra_args):
+def get_master_server_ip(stack_name, config):
```
enrico-usai: I'd add the underscore prefix since it is an internal function and the docstrings too.

cli/pcluster/pcluster.py
sean-smith: added

cli/pcluster/pcluster.py
sean-smith: 👍🏻 done

cli/pcluster/pcluster.py
```diff
@@ -577,7 +577,45 @@ def instances(args):
...
+    if master_id is []:
+        LOGGER.info("MasterServer not running. Can't SSH")
+        sys.exit(1)
+    instance = ec2.describe_instances(InstanceIds=[master_id]).get("Reservations")[0].get("Instances")[0]
+    ip_address = instance.get("PublicIpAddress")
+    if ip_address is None:
+        ip_address = instance.get("PrivateIpAddress")
+    state = instance.get("State").get("Name")
+    if state != "running" or ip_address is None:
+        LOGGER.info("MasterServer: %s\nCannot get ip address." % state.upper)
```
demartinofra: upper is missing the `()`

Also you can use lazy string formatting in LOGGER by simply adding the parameters to replace as arguments:
`LOGGER.info("MasterServer: %s\nCannot get ip address.", state.upper)`

cli/pcluster/pcluster.py
```diff
@@ -595,12 +633,28 @@ def command(args, extra_args):
         stack_result = cfn.describe_stacks(StackName=stack).get("Stacks")[0]
         status = stack_result.get("StackStatus")
         valid_status = ["CREATE_COMPLETE", "UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE"]
-        if status not in valid_status:
-            LOGGER.info("Stack status: %s. Stack needs to be in %s", status, " or ".join(valid_status))
+        invalid_status = ["DELETE_COMPLETE", "DELETE_IN_PROGRESS"]
+
+        if status in invalid_status:
+            LOGGER.info("Stack status: %s. Cannot SSH while in %s" % (status, " or ".join(invalid_status)))
```
demartinofra: Same here for lazy logging

cli/pcluster/pcluster.py
```diff
@@ -577,7 +577,45 @@ def instances(args):
...
+        ip_address = instance.get("PrivateIpAddress")
+    state = instance.get("State").get("Name")
+    if state != "running" or ip_address is None:
+        LOGGER.info("MasterServer: %s\nCannot get ip address." % state.upper)
+        sys.exit(1)
+    return ip_address
+
+
+def _get_output_value(outputs, key_name):
+    return [o.get("OutputValue") for o in outputs if o.get("OutputKey") == key_name][0]
```
demartinofra: This throws IndexError when the key is not found. I would either go with a more meaningful exception or simply handle the case and return None

A potential solution is: `next((o.get("OutputValue") for o in outputs if o.get("OutputKey") == key_name), None)` which also stops the loop when it finds the first element rather than iterating over the full list.

Same applies to `_get_param_value`

cli/pcluster/pcluster.py
```diff
@@ -595,12 +633,28 @@ def command(args, extra_args):
...
-        ip = [o.get("OutputValue") for o in outputs if o.get("OutputKey") == "MasterPublicIP"][0]
+        elif status in valid_status:
+            outputs = stack_result.get("Outputs")
+            username = _get_output_value(outputs, "ClusterUser")
+
+            try:
+                ip = _get_output_value(outputs, "MasterPublicIP")
+            except IndexError:
+                # If they're not using public ip's use the private ip instead
+                ip = _get_output_value(outputs, "MasterPrivateIP")
```
demartinofra: This can also throw.

cli/pcluster/pcluster.py
```diff
@@ -595,12 +633,28 @@ def command(args, extra_args):
...
+        else:
+            # Stack is in CREATING, CREATED_FAILED, or ROLLBACK_COMPLETE but MasterServer is running
+            ip = _get_master_server_ip(stack, config)
+            stack_result = cfn.describe_stacks(StackName=stack).get("Stacks")[0]
```
demartinofra: Can we reuse the stack_result we already retrieved above?

cli/pcluster/pcluster.py
sean-smith: fixed and changed 👍🏻

cli/pcluster/pcluster.py
sean-smith: fixed

cli/pcluster/pcluster.py
sean-smith: yes, changed

cli/pcluster/pcluster.py
sean-smith: Thanks for the suggestion 👌🏻

cli/pcluster/pcluster.py
sean-smith: Changed

Issue #862, opened 2019-02-17
Tags defined in the cluster configuration file cannot applied to EBS volumes


RodneyMarable:
**Environment:**
 - AWS ParallelCluster / CfnCluster version 2.1.1
 - OS: ALL
 - Scheduler: ALL
 - Master instance type: ALL
 - Compute instance type: ALL

**Bug description and how to reproduce:**
I would like to tag EBS volumes using data supplied from the "tags" parameter that is defined in the pcluster config file.  This is important for showback/chargeback purposes, as large provisioned EBS RAID volumes could represent a significant portion of a cluster's operational costs.

Since there does not appear to be a mechanism that supports tagging EBS resources, this should be possible through the postinstall script by parsing the instanceId from EC2 metadata, listing all EBS volumes associated with that instanceId, and tagging the EBS volumes with the tags defined in the pcluster-config file. 

However, the default IAM role defined in the CloudFormation template does not permit EC2CreateTags, EC2DescribeTags, or EC2DeleteTags, and since I do not want to use a custom CloudFormation template or enable a custom_iam_role, this approach won't work.

For a short term solution, I am going to submit a pull request that will add these API calls to the default Cloudformation stack and provide documentation that describes how to perform arbitrary EBS volume tagging through the postinstall script.

**Additional context:**
A better long term fix would be to have Chef perform this tagging as part of "aws-parallelcluster-cookbook/blob/develop/recipes/base_config.rb" or in a separate recipe (tag_ebs_volumes.rb, or something to that effect).

I will open another feature request for this:

> Tags that are predefined in the ParallelCluster configuration file should be applied to all EBS volumes associated with the cluster stack.

enrico-usai:
Closing in favour of: https://github.com/aws/aws-parallelcluster/issues/865

PR #864, opened 2019-02-17
EBS volume tagging in postinstall

RodneyMarable:
*Issue #, if available:* #862 

*Description of changes:* 
- Modified the default IAM role to permit EC2CreateTags, EC2DescribeTags, and EC2DeleteTags so the user doesn't have to use custom_iam_role or a custom CloudFormation template to apply tags to EBS resources.  

- Updated the documentation to include an example of applying EC2 tags to EBS volumes associated with a pcluster stack as part of the postinstall process.

- Applied some other formatting and phrasing changes to cli/pcluster/cli.py and cloudformation/README.rst for improved readability.

*Notes:*
A better long term fix would be to have Chef perform this tagging as part of "aws-parallelcluster-cookbook/blob/develop/recipes/base_config.rb" or in a separate recipe (tag_ebs_volumes.rb, or something along those lines).  It would also be great if tags defined in the pcluster config file could be applied to all EBS volumes associated with a pcluster stack.


sean-smith:
Closing this because we can add tags to EBS volumes without using a post-install script.

In the `AWS::EC2::Volume` resource, there's a [Tags](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-ebs-volume.html#cfn-ec2-ebs-volume-tags) attribute. If we pass the tags from the cli or config (see [tags](https://aws-parallelcluster.readthedocs.io/en/latest/configuration.html#tags)), we can add them in the cloudformation template.

extremoburo:
Hi Sean ,  can you explain it better how to do this from the config file ?

sean-smith:
@extremoburo You can add tags to your config (or via the cli). For example:

```
[cluster mycluster]
...
tags = {"key" : "value", "key2" : "value2"}
```

With the caveat that for ebs-backed instances this does not tag the master root volume, only ebs or raid volumes created by ParallelCluster.

See https://aws-parallelcluster.readthedocs.io/en/latest/configuration.html#tags

extremoburo:
ok thanks , do you think it will be possible to also tag existing volumes attached to the master in future ?

sean-smith:
Volumes attached with `ebs_snapshot_id = snap-xxxxx`?

For now no. Your best bet is to tag them with a post-install script. This PR has the code to do so.

PR #869, opened 2019-02-19
Integ tests: custom templates and packages support

demartinofra:
Add support for custom templates and packages when running the tests: https://github.com/demartinofra/aws-parallelcluster/blob/adab75703c1a92eeb5643c6055b86f31ef266e79/tests/integration-tests/README.md#custom-templates-and-packages

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -13,11 +13,13 @@
 # This file has a special meaning for pytest. See https://docs.pytest.org/en/2.7.3/plugins.html for
 # additional details.
 
+import json
 import logging
 import os
 import re
 from shutil import copyfile
 
+import configparser
 import pytest
 
 from cfn_stacks_factory import CfnStack, CfnStacksFactory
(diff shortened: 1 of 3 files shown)


tests/integration-tests/conftest.py
```diff
@@ -202,22 +206,49 @@ def pcluster_config_reader(test_datadir, vpc_stacks, region, request):
...
     return _config_renderer
 
 
+def _add_custom_packages_configs(cluster_config, request):
+    config = configparser.ConfigParser()
+    config.read(cluster_config)
+    cluster_template = "cluster {0}".format(config.get("global", "cluster_template", fallback="default"))
+
+    for custom_option in ["template_url", "custom_awsbatch_template_url", "custom_chef_cookbook"]:
+        if request.config.getoption(custom_option) and custom_option not in config[cluster_template]:
```
enrico-usai: It seems the value configured in the file takes precedence over the option passed by command line. 
Is that what you wanted? If yes could you add a comment in the readme?

tests/integration-tests/conftest.py
```diff
@@ -42,12 +44,14 @@ def pytest_addoption(parser):
...
     parser.addoption("--schedulers", help="schedulers under test", default=["slurm"], nargs="+")
     parser.addoption("--tests-log-file", help="file used to write test logs", default="pytest.log")
-    parser.addoption("--custom-node-url", help="url to a custom node package")
-    parser.addoption("--custom-cookbook-url", help="url to a custom cookbook package")
-    parser.addoption("--custom-template-url", help="url to a custom cfn template")
     parser.addoption("--output-dir", help="output dir for tests artifacts")
     parser.addoption("--key-name", help="key to use for EC2 instances", type=str, required=True)
     parser.addoption("--key-path", help="key path to use for SSH connections", type=str)
+    parser.addoption("--custom-chef-cookbook", help="url to a custom cookbook package")
+    parser.addoption("--custom-awsbatch-template-url", help="url to a custom awsbatch template")
```
enrico-usai: I'd remove the url suffix

tests/integration-tests/conftest.py
```diff
@@ -42,12 +44,14 @@ def pytest_addoption(parser):
...
+    parser.addoption("--template-url", help="url to a custom cfn template")
```
enrico-usai: I'd remove the `url `suffix and add the `custom` prefix to be aligned with other parameters

tests/integration-tests/conftest.py
demartinofra: Yes that is the intended behaviour. I left a comment where I define the pcluster_config_reader: The current renderer injects options for custom templates and packages in case these are passed to the cli and not present already in the cluster config.

Will clarify in the README as well

tests/integration-tests/conftest.py
demartinofra: The idea was to align the pytest options with the parameters we use in the pcluster config file while having the test_runner options defined with more aligned and explicit names.

PR #870, opened 2019-02-19
Integ tests: various fixes and improvements

demartinofra:
* sort requirements.txt file
* use stack_id when checking stack deletion: when using name and the cluster transitions to `DELETE_COMPLETED` then the describe_stacks raises an exception. This doesn't happen if we use stack id.
* unmark required pytest options since pytest-xdist does not support it. Behaviour is preserved by test_runner 
* fix cluster creation error condition: `CREATE_COMPLETE` string was appearing next to other resources created in the stack so was not a very meaningful check.


tests/integration-tests/cfn_stacks_factory.py
```diff
@@ -60,8 +61,9 @@ def create_stack(self, stack):
         self.__created_stacks[id] = stack
         try:
             cfn_client = boto3.client("cloudformation", region_name=region)
-            cfn_client.create_stack(StackName=name, TemplateBody=stack.template)
-            final_status = self.__wait_for_stack_creation(name, cfn_client)
+            result = cfn_client.create_stack(StackName=name, TemplateBody=stack.template)
+            stack.cfn_stack_id = result["StackId"]
```
enrico-usai: Memo: we could differentiate the Stack and the StackDefinition

PR #858, opened 2019-02-15
FSx Support

sean-smith:
### Fsx Integration

* Incorporates Rex's changes and gets rid of the custom resource implementation
* Supported options are:

```
[fsx fs]
shared_dir = /fsx
# must be in the same vpc as cluster and have post 988 open in it's SG
# fsx_fs_id = fs-073c3803dca3e28a6
storage_capacity = 3600
fsx_kms_key_id = 9e8a129c-0e85-459-865b-3a5be974a22b
imported_file_chunk_size = 1024
export_path = s3://bucket/folder
import_path =  s3://bucket
weekly_maintenance_start_time = 1:00:00
```

### Testing
* Created a `centos7` cluster with `fsx_fs_id` specified
* Created a `centos7` cluster with `storage_capacity`, `imported_file_chunk_size`, `import_path` and `weekly_maintenance_start_time` specified.

### Todo
Verify that `base_os` is `centos6/7` or `ubuntu1604`

### Resources
1. [CreateFileSystem Docs](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystem.html)
2. [Lustre Configuration Docs](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemLustreConfiguration.html)

Signed-off-by: Sean Smith <seaam@amazon.com>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #872, opened 2019-02-20
Check AWS Account limits before cluster creation

enrico-usai:
At cluster creation/update/configuration time the AWS account capacity
will be verified by trying to instantiate a number of instances
equal to the configured `max_queue_size`.

I added the `ssm:GetParametersByPath` policy to the documentation, since
it is required to get the latest amazon linux ami id to use for the test.

In the _awsbatch_ case we are calculating the number of instances for the
test according to the instance_type, the number of vcpus for the instance_type
and the selected `max_vcpus` parameter.
The test will be skipped if the compute instance type is set to "_optimal_".

### Tests

I tested both python 3.6 and 2.7, awsbatch (with optimal and standard instance_type) and the other schedulers.

**Examples:**
```
$ pcluster create test-limit --cluster-template sge-centos6
Beginning cluster creation for cluster: test-limit
ERROR: The configured max size parameter 1000 exceeds the AWS Account limit in the eu-west-1 region.
Your quota allows for 27 more running instance(s). You requested at least 1000

$ pcluster create test-limit --cluster-template batch-mnp
Beginning cluster creation for cluster: test-limit
ERROR: The configured max size parameter 1000 exceeds the number of free private IP addresses available in the Compute Subnet.
```

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -32,38 +32,6 @@ from botocore.exceptions import ClientError
 from pcluster.config_sanity import ResourceValidator
 
 
-def get_stack_template(region, aws_access_key_id, aws_secret_access_key, cluster_name):
-    """
-    Get stack template corresponding to the given cluster name.
-
-    :param region: AWS Region
-    :param aws_access_key_id: AWS access key
-    :param aws_secret_access_key: AWS secret access key
-    :param cluster_name: The cluster name to search for
-    :return: the corresponding stack template
-    """
-    cfn = boto3.client(
-        "cloudformation",
-        region_name=region,
-        aws_access_key_id=aws_access_key_id,
-        aws_secret_access_key=aws_secret_access_key,
-    )
-    stack_name = "parallelcluster-" + cluster_name
-
-    try:
-        stack = cfn.describe_stacks(StackName=stack_name).get("Stacks")[0]
-    except ClientError as e:
-        print(e.response.get("Error").get("Message"))
-        sys.stdout.flush()
-        sys.exit(1)
-
-    cli_template = [p.get("ParameterValue") for p in stack.get("Parameters") if p.get("ParameterKey") == "CLITemplate"][
-        0
-    ]
-
-    return cli_template
-
-
 class ParallelClusterConfig(object):
     """Manage ParallelCluster Config."""
 
@@ -201,6 +169,32 @@ class ParallelClusterConfig(object):
         self.aws_access_key_id = self._get_config_value("aws", "aws_access_key_id")
         self.aws_secret_access_key = self._get_config_value("aws", "aws_secret_access_key")
 
+    def __get_stack_name(self):
+        """Return stack name"""
+        return "parallelcluster-" + self.args.cluster_name
+
+    def __get_stack_template(self):
+        """Get stack template."""
+        cfn = boto3.client(
+            "cloudformation",
+            region_name=self.region,
+            aws_access_key_id=self.aws_access_key_id,
+            aws_secret_access_key=self.aws_secret_access_key,
+        )
+
+        try:
+            stack = cfn.describe_stacks(StackName=self.__get_stack_name()).get("Stacks")[0]
+        except ClientError as e:
+            print(e.response.get("Error").get("Message"))
+            sys.stdout.flush()
+            sys.exit(1)
+
+        cli_template = [
+            p.get("ParameterValue") for p in stack.get("Parameters") if p.get("ParameterKey") == "CLITemplate"
+        ][0]
+
+        return cli_template
+
     def __get_cluster_template(self):
         """
         Determine which cluster template will be used and return it.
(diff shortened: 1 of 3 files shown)


cli/pcluster/cfnconfig.py
```diff
@@ -347,19 +336,89 @@ def __init_vpc_parameters(self):
...
+                    % e.response.get("Error").get("Message")
+                )
+
+    def __get_latest_alinux_ami_id(self):
+        try:
+            # get latest alinux ami id to use as test image
+            ssm = boto3.client("ssm", region_name=self.region)
+            test_ami_id = (
+                ssm.get_parameters_by_path(Path="/aws/service/ami-amazon-linux-latest")
+                    .get("Parameters")[0]
```
sean-smith: Failing Travis, move left one tab.

cli/pcluster/cfnconfig.py
```diff
@@ -347,19 +336,89 @@ def __init_vpc_parameters(self):
...
+                        max_size, self.region
+                    )
+                )
+            elif code == "InsufficientInstanceCapacity":
+                self.__fail(
+                    "The configured max size parameter {0} exceeds the On-Demand capacity on AWS.".format(max_size)
+                )
+            elif code == "InsufficientFreeAddressesInSubnet":
+                self.__fail(
+                    "The configured max size parameter {0} exceeds the number of free private IP addresses "
```
sean-smith: Is it possible to get the # of free addresses?

cli/pcluster/cfnconfig.py
```diff
@@ -347,19 +336,89 @@ def __init_vpc_parameters(self):
...
-                    "ERROR: VPC section [%s] used in [%s] section is not defined"
-                    % (vpc_section, self.__cluster_section)
+                self.__fail(
+                    "VPC section [%s] used in [%s] section is not defined" % (vpc_section, self.__cluster_section)
                 )
-                sys.exit(1)
+
+    def __check_account_capacity(self):
+        """Try to launch the requested number of instances to verify Account limits."""
+        test_ami_id = self.__get_latest_alinux_ami_id()
```
sean-smith: Why do you need this ami?

cli/pcluster/cfnconfig.py
```diff
@@ -135,6 +108,12 @@ def __init__(self, args):
         except AttributeError:
             pass
 
+    @staticmethod
+    def __fail(message):
+        """Print an error message and exit."""
+        print("ERROR: {0}".format(message))
```
sean-smith: Use `LOGGER.critical()` as it logs to both stdout and the cli log file `~/.parallelcluster/pcluster-cli.log`

sean-smith: Also I've defined a `__fail()` function in https://github.com/aws/aws-parallelcluster/pull/858

cli/pcluster/cfnconfig.py
```diff
@@ -151,19 +130,13 @@ def __init_config(self):
...
-                        os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))),
-                        os.path.sep,
-                        os.path.sep,
-                    )
+                msg = "Default config {0} not found.".format(config_file)
+                msg += "\nYou can copy a template from here: {0}{1}examples{1}config".format(
+                    os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), os.path.sep
                 )
-                sys.exit(1)
+                self.__fail(msg)
```
sean-smith: Can you do this as one call?

```
self.__fail(
     "Default config {0} not found.".format(config_file) + \
     "\nYou can copy a template from here: {0}{1}examples{1}config".format(
      os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), os.path.sep
)
```

cli/pcluster/cfnconfig.py
enrico-usai: Fixed, thank you.

cli/pcluster/cfnconfig.py
enrico-usai: Unfortunately no, it is a ClientError exception with just a generic message like:
```
There are not enough free addresses in subnet 'subnet-xxx' to satisfy the requested number of instances.
```
but thanks to your suggestion I improved the _InstanceLimitExceeded_ and the _InsufficientInstanceCapacity_ cases by adding the Exception message:

```
$ pcluster create test-limit3 --cluster-template sge-centos6
Beginning cluster creation for cluster: test-limit3
ERROR: The configured max size parameter 1000 exceeds the AWS Account limit in the eu-west-1 region.
Your quota allows for 27 more running instance(s). You requested at least 1000
```

cli/pcluster/cfnconfig.py
enrico-usai: Because the `run_instances` command requires an AMI id, so it is a way to found an ami id for the test.

cli/pcluster/cfnconfig.py
enrico-usai: I know you already defined the `__fail()` function. Let's rebase it when required.

cli/pcluster/cfnconfig.py
enrico-usai: Done.

cli/pcluster/utils.py
```diff
@@ -111,3 +112,42 @@ def upload_resources_artifacts(bucket_name, root, aws_client_config):
...
+    :raises ClientError if unable to download the pricing file.
+    """
+    s3 = boto3.resource("s3", region_name=region)
+    bucket_name = "%s-aws-parallelcluster" % region
+    file_name = "instances/instances.json"
+
+    try:
+        file_contents = s3.Object(bucket_name, file_name).get()["Body"].read().decode("utf-8")
+        instances = json.loads(file_contents)
+    except ClientError:
```
demartinofra: not needed

enrico-usai: fixed, thank you

cli/pcluster/cfnconfig.py
```diff
@@ -151,19 +130,14 @@ def __init_config(self):
...
         if not os.path.isfile(config_file):
             if default_config:
-                print("Default config %s not found" % config_file)
-                print(
-                    "You can copy a template from here: %s%sexamples%sconfig"
-                    % (
-                        os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))),
-                        os.path.sep,
+                self.__fail(
+                    "Default config {0} not found.\nYou can copy a template from here: {1}{2}examples{2}config".format(
```
demartinofra: is there a missing param in the format call?

enrico-usai: good catch, I pasted the `config_file `parameter in the wrong place, fixed.

PR #904, opened 2019-03-01
Validate FSx Params

sean-smith:
I've added two validations:
1. `import_path` needs to be specified when using `imported_file_chunk_size` or `export_path`
2. Only Supported OS is `centos7`

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -580,10 +580,15 @@ class ParallelClusterConfig(object):
         if self.__config.has_option(self.__cluster_section, option):
             self.__fail("option %s cannot be used with awsbatch" % option)
 
-    def __validate_awsbatch_os(self, baseos):
-        supported_batch_oses = ["alinux"]
-        if baseos not in supported_batch_oses:
-            self.__fail("awsbatch scheduler supports following OSes: %s" % supported_batch_oses)
+    def __get_os(self):
+        base_os = "alinux"
+        if self.__config.has_option(self.__cluster_section, "base_os"):
+            base_os = self.__config.get(self.__cluster_section, "base_os")
+        return base_os
+
+    def __validate_os(self, service, baseos, supported_oses):
+        if baseos not in supported_oses:
+            self.__fail("%s supports following OSes: %s" % (service, supported_oses))
 
     def __init_batch_parameters(self):  # noqa: C901 FIXME!!!
         """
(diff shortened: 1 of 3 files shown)


cli/pcluster/cfnconfig.py
```diff
@@ -801,6 +800,10 @@ def __init_fsx_parameters(self):
         if not fsx_section:
             return
 
+        # Check that the base_os is supported
+        if self.__config.has_option(self.__cluster_section, "base_os"):
```
demartinofra: we should consider also the case where `base_os` is not set but uses default `alinux` value

cli/pcluster/cfnconfig.py
```diff
@@ -824,6 +827,11 @@ def __init_fsx_parameters(self):
                 # Separate sanity_check for fs_id, need to pass in fs_id and subnet_id
                 if self.__sanity_check and fsx_options.get(key)[1] == "fsx_fs_id":
                     self.__validate_resource("fsx_fs_id", (value, self.__master_subnet))
+                elif self.__sanity_check and fsx_options.get(key)[1] in [
```
demartinofra: we could check `if self.__sanity_check` just once at the beginning

cli/pcluster/cfnconfig.py
```diff
@@ -824,6 +827,11 @@ def __init_fsx_parameters(self):
...
+                    "FSx_imported_file_chunk_size",
+                    "FSx_export_path",
+                ]:
+                    self.__validate_resource(fsx_options.get(key)[1], (value, fsx_options.get("import_path")[1]))
```
demartinofra: isn't `fsx_options.get("import_path")[1]` always None which means validation at line 207 always fails?

cli/pcluster/config_sanity.py
```diff
@@ -201,10 +201,14 @@ def __validate_fsx_parameters(self, resource_type, resource_value):
                 self.__fail(
                     resource_type, "Capacity for FSx lustre filesystem, minimum of 3,600 GB, increments of 3,600 GB"
                 )
+        # Check to see if import_path is specified with imported_file_chunk_size and export_path
+        elif resource_type in ["FSx_imported_file_chunk_size", "FSx_export_path"]:
+            if resource_value[1] is None:
+                self.__fail(resource_type, "import_path must be specified.")
         # FSX file chunk size check
-        elif resource_type == "FSx_imported_file_chunk_size":
+        if resource_type == "FSx_imported_file_chunk_size":
```
demartinofra: This could be moved inside the elif above

cli/pcluster/cfnconfig.py
sean-smith: whoops. ok updated now

cli/pcluster/cfnconfig.py
```diff
@@ -819,10 +825,16 @@ def __init_fsx_parameters(self):
             if not value:
                 temp_fsx_options.append("NONE")
             else:
+                if not self.__sanity_check:
+                    pass
```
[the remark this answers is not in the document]
sean-smith: changed

PR #907, opened 2019-03-04
Implement integration tests for EBS

demartinofra:
Following tests are implemented:
* single EBS volume
* multiple EBS volumes

Additional changes:
* reorganize common scheduler commands that are shared across several test cases in dedicated classes
* replace paramiko with fabric (a higher level library built on top of paramiko). This simplifies the logic in the remote_command_executor
* wait for 60 seconds after cluster creation since under certain circumstances the cluster seems not to be ready right after stack completion. Waiting for additional 60 seconds as temporary workaround.

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -10,6 +10,7 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
+import time
 
 import configparser
 from retrying import retry
@@ -85,6 +86,11 @@ class ClustersFactory:
             logging.error(error)
             raise Exception(error)
         logging.info("Cluster {0} created successfully".format(name))
+
+        # FIXME: temporary workaround since in certain circumstances the cluster isn't ready for
+        # job submission right after creation. We need to investigate this further.
+        logging.info("Sleeping for 60 seconds in case cluster is not ready yet")
+        time.sleep(60)
 
     @retry(stop_max_attempt_number=10, wait_fixed=5000, retry_on_exception=retry_if_subprocess_error)
     def destroy_cluster(self, name):
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -1,0 +1,147 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from abc import ABCMeta, abstractmethod
+
+from retrying import retry
+
+from assertpy import assert_that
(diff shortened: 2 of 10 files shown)


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, remote_command_executor):
+        super().__init__(remote_command_executor)
+
+    @retry(
+        retry_on_result=lambda result: "FAILED" not in result and any(status != "SUCCEEDED" for status in result),
+        wait_fixed=seconds(7),
+        stop_max_delay=minutes(15),
+    )
+    def wait_job_completed(self, job_id):  # noqa: D102
+        result = self._remote_command_executor.run_remote_command("awsbstat -d {0}".format(job_id))
```
sean-smith: Do we need to run this remotely? Why not run locally and pass in the cluster name

tests/integration-tests/tests/test_ebs.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pytest
+
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from tests.common.schedulers_common import get_scheduler_commands
+
+
+@pytest.mark.regions(["us-east-1", "eu-west-1", "cn-north-1", "us-gov-west-1"])
+@pytest.mark.instances(["c5.xlarge"])
```
sean-smith: I'd also add `c4.xlarge` so we have a non-nvme based instance. This has caused problems with mounting in the past, so I think we should check both.

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -0,0 +1,147 @@
...
+
+class AWSBatchCommands(SchedulerCommands):
+    """Implement commands for awsbatch scheduler."""
+
+    def __init__(self, remote_command_executor):
+        super().__init__(remote_command_executor)
+
+    @retry(
+        retry_on_result=lambda result: "FAILED" not in result and any(status != "SUCCEEDED" for status in result),
+        wait_fixed=seconds(7),
```
sean-smith: Why waiting 7 seconds?

tests/integration-tests/tests/common/schedulers_common.py
demartinofra: I decided to run these commands remotely for the following reasons:
* verify that the pcluster batch cli is correctly installed, configured, and upgraded on the master node
* verify that the default IAM role assumed by the master role has the required permissions to execute all batch commands.

tests/integration-tests/tests/test_ebs.py
demartinofra: good point!

tests/integration-tests/tests/common/schedulers_common.py
demartinofra: This means it checks for job completion at regular intervals of 7 seconds.

Why 7 seconds? After a series of runs it seemed to be a good compromise which allows us to proceed with the test soon enough after job completion and at the same time it doesn't overload the master node with remote commands submission

tests/integration-tests/clusters_factory.py
```diff
@@ -86,6 +87,11 @@ def create_cluster(self, cluster):
             raise Exception(error)
         logging.info("Cluster {0} created successfully".format(name))
 
+        # FIXME: temporary workaround since in certain circumstances the cluster isn't ready for
+        # job submission right after creation. We need to investigate this further.
+        logging.info("Sleeping for 60 seconds in case cluster is not ready yet".format(name))
```
enrico-usai: I think you are missing the `{0}`, or you should remove the `format`

demartinofra: Removed the format. Thanks!

demartinofra:
@sean-smith I'm going to merge this one. Let me know if you have any additional comment and I'll address it in a separate PR.

PR #916, opened 2019-03-07
Migrate parallelcluster release-check to the new testing framework

demartinofra:
Old version of the release-check: https://github.com/aws/aws-parallelcluster/blob/develop/tests/parallelcluster-release-check.py

Diff between Old and New version: https://gist.github.com/demartinofra/b28d4a34452fb0d859529e712a74fe42

cluster-check.sh was only moved.

From 872 lines to 180 lines!

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -16,6 +16,7 @@
 import json
 import logging
 import os
+import random
 import re
 from shutil import copyfile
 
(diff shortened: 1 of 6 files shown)


sean-smith:
Amazing 👌🏻

tests/integration-tests/conftest.py
```diff
@@ -271,29 +272,41 @@ def cfn_stacks_factory():
     factory.delete_all_stacks()
 
 
+_AVAILABILITY_ZONE_OVERRIDES = {
+    # c5.xlarge is not supported in us-east-1e
+    # FSx Lustre file system creation is currently not supported for us-east-1e
+    "us-east-1": ["us-east-1a", "us-east-1b", "us-east-1c", "us-east-1d", "us-east-1f"],
```
sean-smith: We could add a link to this on our documentation, as a see what az's fsx is in.

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -109,6 +114,9 @@ def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
         assert_that(status).is_length(1 + children_number)
         assert_that(status).contains_only("SUCCEEDED")
 
+    def compute_nodes_count(self):  # noqa: D102
```
sean-smith: Do you have to implement this because it's in the interface?

tests/integration-tests/tests/test_scaling/test_multiple_jobs_submission/cluster-check.sh
```diff
@@ -0,0 +1,265 @@
+#!/bin/bash
```
sean-smith: Is this the same the file before, just moved?

tests/integration-tests/tests/common/schedulers_common.py
demartinofra: Yes and at the moment no test is using it.

tests/integration-tests/tests/test_scaling/test_multiple_jobs_submission/cluster-check.sh
demartinofra: Exactly the same. Did't do a git move since I'm still preserving the old release-check

tests/integration-tests/conftest.py
```diff
@@ -271,29 +272,41 @@ def cfn_stacks_factory():
...
-        has_nat_gateway=False,
-        default_gateway=Gateways.NAT_GATEWAY,
-    )
-    vpc_config = VPCConfig(subnets=[public_subnet, private_subnet])
-    template = VPCTemplateBuilder(vpc_config).build()
-
     regions = request.config.getoption("regions")
     vpc_stacks = {}
     for region in regions:
+        public_subnet = SubnetConfig(
```
enrico-usai: I know the reason but I think a comment could be useful to remember why you moved the creation of the vpc into the loop.

tests/integration-tests/conftest.py
demartinofra: according to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html: 

> An Availability Zone is represented by a region code followed by a letter identifier; for example, us-east-1a. To ensure that resources are distributed across the Availability Zones for a region, we independently map Availability Zones to names for each AWS account. For example, the Availability Zone us-east-1a for your AWS account might not be the same location as us-east-1a for another AWS account.

which means probably also having it hardcoded like this is not the best option. Maybe an account based config file might be a better way. Shall we leave it like this for now?

tests/integration-tests/conftest.py
```diff
@@ -271,29 +272,41 @@ def cfn_stacks_factory():
...
+            name="PublicSubnet",
+            cidr="10.0.0.0/24",
+            map_public_ip_on_launch=True,
+            has_nat_gateway=True,
+            default_gateway=Gateways.INTERNET_GATEWAY,
+            availability_zone=random.choice(_AVAILABILITY_ZONE_OVERRIDES.get(region, [None])),
```
enrico-usai: Why `random` ?

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -141,7 +149,70 @@ def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
         status = self.get_job_exit_status(job_id)
         assert_that(status).is_equal_to("0")
 
+    def compute_nodes_count(self):  # noqa: D102
+        result = self._remote_command_executor.run_remote_command("qhost | grep -o ip- | wc -l")
+        return int(result.stdout.split()[-1])
```
enrico-usai: Why do you need to split the stdout?

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -141,7 +149,70 @@ def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
...
+
+    def submit_command(self, command):  # noqa: D102
+        raise NotImplementedError
+
+    def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
+        raise NotImplementedError
+
+    def compute_nodes_count(self):  # noqa: D102
+        result = self._remote_command_executor.run_remote_command("sinfo --Node --noheader | grep compute | wc -l")
+        return int(result.stdout.split()[-1])
```
enrico-usai: Same question

tests/integration-tests/conftest.py
demartinofra: Cause I want to randomly choose one of the available AZs so that we don't test in a single one over time.

tests/integration-tests/tests/common/schedulers_common.py
demartinofra: To extract the last line and trim whitespaces:
```
>>> a = "a\n b"
>>> a.split()[-1]
'b'
```

tests/integration-tests/conftest.py
demartinofra: added a FIXME comment

tests/integration-tests/conftest.py
demartinofra: done

tests/integration-tests/tests/common/schedulers_common.py
demartinofra: added comment

tests/integration-tests/tests/common/schedulers_common.py
enrico-usai: It is a very important reason.

PR #918, opened 2019-03-08
IAM User Policy Improvements

sean-smith:
* No trailing commas allowed in JSON 😢
* Fixes an issue:
```
  - AWS::IAM::Role RootRole API: iam:CreateRole User: arn:aws:iam::[Account Id]:user/pcluster is not authorized to perform: iam:CreateRole on resource: arn:aws:iam::[Account Id]:role/parallelcluster-test3-RootRole-TIPZ93ZZSH59
```

### DIFF
--- a/docs/iam.rst
+++ b/docs/iam.rst
@@ -207,6 +207,17 @@ ParallelClusterUserPolicy
 
 In case you are using SGE, Slurm or Torque as a scheduler:
 
+Note, if you use a custom role, ``ec2_iam_role = role_name``, you'll need to change the IAM resource to include
+the name of that role.
+
+From: ::
+
+  "Resource": "arn:aws:iam::<AWS ACCOUNT ID>:role/parallelcluster-*"
+
+To: ::
+
+  "Resource": "arn:aws:iam::<AWS ACCOUNT ID>:role/<role_name>"
+
 ::
 
   {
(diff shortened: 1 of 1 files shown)


docs/iam.rst
```diff
@@ -390,7 +390,7 @@ In case you are using SGE, Slurm or Torque as a scheduler:
                   "iam:SimulatePrincipalPolicy"
               ],
               "Effect": "Allow",
-              "Resource": "arn:aws:iam::<AWS ACCOUNT ID>:role/<PARALLELCLUSTER EC2 ROLE NAME>"
+              "Resource": "arn:aws:iam::<AWS ACCOUNT ID>:role/parallelcluster-*"
```
demartinofra: what if the EC2 role name is created by the customer and doesn't start with parallelcluster-?

sean-smith: This is only the customer's role if they're using `ec2_iam_role`. Perhaps a note is best. The current wording `<PARALLELCLUSTER EC2 ROLE NAME>` leads one to believe it's the user role (aka the role this policy is attached too) when in fact it's the root role created by the cloudformation template.

sean-smith: We also use `parallelcluster-*` in the batch version of this policy. See [#L561](https://github.com/aws/aws-parallelcluster/blame/develop/docs/iam.rst#L561)

sean-smith: I've updated this with a comment

demartinofra: Yeah i was referring exactly to the case when the user is using a custom ec2 role. I think the comment is good enough

PR #921, opened 2019-03-12
Test Python 3.7

sean-smith:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: required
 dist: xenial
 
 python:
-  - "2.6"
   - "2.7"
   - "3.4"
   - "3.5"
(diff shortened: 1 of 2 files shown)


sean-smith:
Python 3.7 has bugs with travis, fix is here: https://github.com/travis-ci/travis-ci/issues/9069

Closing as it's either 3.7 or 2.6 and we have to support 2.6 for Centos 6.

.travis.yml
```diff
@@ -1,12 +1,13 @@
 language: python
-sudo: false
+sudo: required
```
demartinofra: I remember sudo is required to use xenial dist. Will this be always like this? can we add a comment or a link to an issue we can track to know when we can remove sudo?

sean-smith: `sudo: required` forces it to use a non-containerized infrastructure. Travis is deprecating their containerized infrastructure in favor of VM based.

See https://stackoverflow.com/questions/26299552/travis-sudo-is-disabled

PR #929, opened 2019-03-14
batch: validate max vcpus size against vpcus of the compute instance type

enrico-usai:
`max_vcpus` must be greater than or equal to the number of vcpus available for the selected `compute_instance_type`.

I'm skipping the test for _optimal_, instance families or instances list.

Without this patch if you have a configuration like the following, the batch jobs remain _RUNNABLE_ forever:
```
[cluster batch]
scheduler = awsbatch
compute_instance_type = c4.8xlarge
min_vcpus = 0
desired_vcpus = 18
max_vcpus = 18
...
```

From now, the creation will be stopped with the message:
```
$ pcluster create batch-test
Beginning cluster creation for cluster: batch-test
Config sanity error on resource AWSBatch_Parameters: Max vcpus must be greater than or equal to 36, that is the number of vcpus available for the c4.8xlarge you selected as compute instance type
```

I also improved the code to avoid a failure if the instance type cannot be found in the pricing file or in the batch instances file or if the files are not available.

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -30,7 +30,7 @@ import pkg_resources
 from botocore.exceptions import ClientError
 
 from pcluster.config_sanity import ResourceValidator
-from pcluster.utils import get_vcpus_from_pricing_file
+from pcluster.utils import get_instance_vcpus
 
 
 class ParallelClusterConfig(object):
@@ -412,7 +412,7 @@ class ParallelClusterConfig(object):
         try:
             max_size = int(self.parameters.get("MaxSize"))
             if self.parameters.get("Scheduler") == "awsbatch":
-                vcpus = get_vcpus_from_pricing_file(self.region, instance_type)
+                vcpus = get_instance_vcpus(self.region, instance_type)
                 max_size = -(-max_size // vcpus)
         except ValueError:
             self.__fail("Unable to convert max size parameter to an integer")
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -112,6 +112,11 @@ class ParallelClusterConfig(object):
             pass
 
     @staticmethod
+    def __warn(message):
+        """Print a warning message."""
+        print("WARNING: {0}".format(message))
+
+    @staticmethod
     def __fail(message):
         """Print an error message and exit."""
         print("ERROR: {0}".format(message))
@@ -358,6 +363,10 @@ class ParallelClusterConfig(object):
 
         instance_type = self.parameters.get("ComputeInstanceType", "t2.micro")
         max_size = self.__get_max_number_of_instances(instance_type)
+        if max_size < 0:
+            self.__warn("Unable to check AWS account capacity. Skipping limits validation")
+            return
+
         try:
             # Check for insufficient Account capacity
             ec2 = boto3.client("ec2", region_name=self.region)
--- a/cli/pcluster/config_sanity.py
+++ b/cli/pcluster/config_sanity.py
@@ -539,15 +539,6 @@ class ResourceValidator(object):
                 if spot_price > 100 or spot_price < 0:
                     self.__fail(resource_type, "Spot bid percentage needs to be between 0 and 100")
 
-            # Check sanity on desired, min and max vcpus
-            if (
-                "DesiredSize" not in resource_value
-                or "MinSize" not in resource_value
-                or "MaxSize" not in resource_value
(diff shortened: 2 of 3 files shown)


cli/pcluster/config_sanity.py
```diff
@@ -532,16 +534,59 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
                     try:
                         file_contents = s3.Object(bucket_name, file_name).get()["Body"].read().decode("utf-8")
                         supported_instances = json.loads(file_contents)
-                        for instance in resource_value["ComputeInstanceType"].split(","):
+
+                        if "," not in compute_instance_type and "." in compute_instance_type:
+                            # if the type is not a list, and contains dot (nor optimal, nor a family)
+                            # validate instance type against max_vcpus limit
+                            vcpus = get_vcpus_from_pricing_file(self.region, compute_instance_type)
+                            if max_size < vcpus:
```
demartinofra: Shall we check if vcpus is -1 and print invalid instance type in that case? what if it's a new insstance type? we periodically update our list, right?

cli/pcluster/config_sanity.py
```diff
@@ -532,16 +534,59 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+                    self.__fail(resource_type, "Spot bid percentage needs to be between 0 and 100")
+
+            # Check sanity on desired, min and max vcpus
+            if (
+                "DesiredSize" not in resource_value
+                or "MinSize" not in resource_value
+                or "DesiredSize" not in resource_value
+            ):
+                # this should never occur
+                self.__fail(resource_type, "Both min, desired and max vcpus parameters must be set")
```
demartinofra: Is this true? can't I specify only some of them?

Also you are checking twice for DesiredSize

sean-smith: Looking at `__init_batch_parameters()`, there's no way these values would not be set. There is a chance they're set to a string type, but this doesn't validate that.

cli/pcluster/config_sanity.py
```diff
@@ -532,16 +534,59 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
             # Check compute instance types
             if "ComputeInstanceType" in resource_value:
+                compute_instance_type = resource_value["ComputeInstanceType"]
                 try:
                     s3 = boto3.resource("s3", region_name=self.region)
                     bucket_name = "%s-aws-parallelcluster" % self.region
                     file_name = "instances/batch_instances.json"
                     try:
                         file_contents = s3.Object(bucket_name, file_name).get()["Body"].read().decode("utf-8")
                         supported_instances = json.loads(file_contents)
```
sean-smith: We should add logic to print a warning and not fail if it's unable to fetch the `instances.json` file. This file is created based on the return error message from batch and could change outside of our control. So adding logic to not fail if it's null or if it's not json would be a boon here.

cli/pcluster/config_sanity.py
enrico-usai: fixed

cli/pcluster/config_sanity.py
enrico-usai: done

cli/pcluster/cfnconfig.py
```diff
@@ -111,6 +111,11 @@ def __init__(self, args):
         except AttributeError:
             pass
 
+    @staticmethod
+    def __warn(message):
```
demartinofra: it would be nice to use the logging module to log to stdout but we can leave it as is for now

cli/pcluster/config_sanity.py
```diff
@@ -532,47 +533,84 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+                if spot_price > 100 or spot_price < 0:
+                    self.__fail(resource_type, "Spot bid percentage needs to be between 0 and 100")
+
+            # Check sanity on desired, min and max vcpus
+            if (
+                "DesiredSize" not in resource_value
+                or "MinSize" not in resource_value
+                or "MaxSize" not in resource_value
+            ):
+                # this should never occur
```
demartinofra: If this never occurs I would just remove this if

enrico-usai: Done

cli/pcluster/cfnconfig.py
enrico-usai: Yeah we can leave as is for now

PR #930, opened 2019-03-15
Fix to use empty AMI (not bootstrapped) at runtime

lukeseawalker:
The ParallelCluster cookbook needs to be vendored in two cases:
- using a empty AMI (not bootstrapped)
- using a custom cookbook at cluster creation time

In the other case of an already bootstrapped AMI, there is not even the
need to download the cookbook

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -2188,6 +2188,16 @@
                 "\n",
                 "  exit 1\n",
                 "}\n",
+                "function vendor_cookbook\n",
+                "{\n",
+                "  mkdir /tmp/cookbooks\n",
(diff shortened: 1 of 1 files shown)


sean-smith:
For future reference, this resolves the following issue:
```
================================================================================
Error Resolving Cookbooks for Run List:
================================================================================

Missing Cookbooks:
------------------
No such cookbook: aws-parallelcluster

Expanded Run List:
------------------
* aws-parallelcluster::_prep_env

System Info:
------------
chef_version=14.2.0
platform=amazon
platform_version=2018.03
ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
program_name=/usr/bin/chef-client
executable=/opt/chef/bin/chef-client
```

PR #942, opened 2019-03-19
Disable FSx Updates

sean-smith:
* Currently an update to an FSx filesystem requires replacement
  See [AWS::FSx:FileSystem LustreConfiguration](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-fsx-filesystem-lustreconfiguration.html)
* To avoid customers accidently deleting their filesystems we're
disabling updates when `sanity_check = true`
* To enable updates, the customer can set `sanity_check = false`

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -188,8 +188,8 @@ class ParallelClusterConfig(object):
         """Return stack name."""
         return "parallelcluster-" + self.args.cluster_name
 
-    def __get_stack_template(self):
-        """Get stack template."""
+    def __get_stack_parameter(self, parameter):
+        """Get a parameter from stack."""
         cfn = boto3.client(
             "cloudformation",
             region_name=self.region,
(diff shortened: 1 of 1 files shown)


cli/pcluster/cfnconfig.py
```diff
@@ -804,6 +804,26 @@ def __init_raid_parameters(self):  # noqa: C901 FIXME!!!
...
+    def __check_fsx_updates(self, fsx_section, fsx_options):
+        """
+        Check FSx Parameters for updates.
+
+        :param fsx_section: name of FSx Section
+        :param fsx_options: dictionary of options
+        :return: True if updates were made else False
+        """
+        parameters = self.__get_stack_parameter("FSXOptions").split(",")
+        for i, key in enumerate(fsx_options):
```
demartinofra: This might be simplified to the following (need to check if it really works :))
```
new_values = [self.__get_option_in_section(fsx_section, option) or "NONE" for option in fsx_options]
old_values = self.__get_stack_parameter("FSXOptions").split(",")
            
if new_values != old_values:
    self.__fail()
```

sean-smith: Thanks! Confirmed it works

PR #945, opened 2019-03-20
 Amazon Linux with Fsx Document Custom AMI

sean-smith:
When using a `custom_ami`, with `alinux`, the kernel needs to be >= `4.14.104-78.84.amzn1.x86_64` for the FSx mount to succeed.
 

docs/configuration.rst
```diff
@@ -991,6 +991,10 @@ Configuration for an attached FSx Lustre file system. See `FSx CreateFileSystem
 
 FSx Lustre is supported when ``base_os = centos7 | alinux``.
 
+When using an Amazon Linux ``custom_ami``, the kernel must be >= ``4.14.104-78.84.amzn1.x86_64``.
```
enrico-usai: I'd write: _When using an Amazon Linux  please read the requirements here..._ avoiding to write  the kernel version in our doc but demanding this info to the Fsx doc

sean-smith: Actually I think there's a lot of value in leaving the kernel version here. In the future when the new kernel comes out, users can just compare their kernel version with the one in our docs. Making it an explicit requirement in our docs, ensures people read it and don't just glance over that.

PR #947, opened 2019-03-22
Implement Slurm integration tests

demartinofra:
- Test installed Slurm version
- Test dummy nodes are updated when cluster max size changes
- Test cluster does not scale when there are pending job dependencies
- Test cluster does not scale when job requirements exceed cluster capacity
- Test dummy nodes are dinamically updated when real nodes join cluster


tests/integration-tests/utils.py
```diff
@@ -66,6 +66,22 @@ def retrieve_cfn_outputs(stack_name, region):
...
+    logging.debug("Retrieving stack resources for stack {}".format(stack_name))
+    try:
+        cfn = boto3.client("cloudformation", region_name=region)
+        stack_resources = cfn.list_stack_resources(StackName=stack_name)["StackResourceSummaries"]
+        resources = {}
+        for resource in stack_resources:
+            resources[resource.get("LogicalResourceId")] = resource.get("PhysicalResourceId")
+        return resources
+    except Exception as e:
+        logging.warning("Failed retrieving stack resources for stack {} with exception: {}".format(stack_name, e))
```
enrico-usai: Just a note: Python 2.6 and Python 3.0 require that you number the placeholders:
* _Changed in version 2.7: The positional argument specifiers can be omitted for str.format() and unicode.format(), so '{} {}' is equivalent to '{0} {1}', u'{} {}' is equivalent to u'{0} {1}'._
* _Changed in version 3.1: The positional argument specifiers can be omitted for str.format(), so '{} {}'.format(a, b) is equivalent to '{0} {1}'.format(a, b)._

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -0,0 +1,153 @@
...
+    _assert_dummy_nodes(remote_command_executor, curent_max_size)
+
+    # Change ASG value and check dummy-nodes settings
+    new_max_size = 1
+    asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name, MaxSize=new_max_size)
+    time.sleep(40)
+    _assert_dummy_nodes(remote_command_executor, new_max_size)
+
+    # Restore initial cluster size
+    asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name, MaxSize=curent_max_size)
```
enrico-usai: typo: current

tests/integration-tests/utils.py
demartinofra: Thanks 👍 In any case the integration tests code is not compatible with older python versions for several other features.

tests/integration-tests/tests/schedulers/test_slurm.py
demartinofra: fixed

PR #992, opened 2019-04-08
Integ Tests: Add missing region fixture to update test

enrico-usai:
It was failing with:
`botocore.exceptions.NoRegionError: You must specify a region.`


tests/integration-tests/tests/update/test_update.py
```diff
@@ -26,7 +26,7 @@
 @pytest.mark.regions(["us-west-1"])
 @pytest.mark.schedulers(["slurm"])
 @pytest.mark.oss(["alinux"])
-@pytest.mark.usefixtures("os", "scheduler")
+@pytest.mark.usefixtures("region", "os", "scheduler")
```
demartinofra: are you sure this was an issue, cause you are already specifying this fixture as a function argument.

enrico-usai: no, in fact I tried to solve the issue by adding this fixture but the issue is still there

PR #1000, opened 2019-04-12
 Integ test: fix and improve update_test

enrico-usai:
+ fix update_test by adding missing region to the boto3 call
+ add instance fixture to be able to skip the test if the instance is not available (e.g. specific partition)
+ use eu-west-1 region to avoid to overload the us-west-1
+ restore use of node_id to have the test file name in the config
  + node.nodeid is: `tests_outputs/.../test_slurm.py::test_slurm[c5.xlarge-us-west-1-alinux-slurm].config`
  + instead node.name is: `tests_outputs/.../test_slurm[c5.xlarge-us-west-1-alinux-slurm].config`
  + We changed it to node.name in a previous commit because the "::" characters was a problem for gitbash (windows): (OSError, Invalid argument) 
  + Now we are restoring the node_id by replacing the bad characters because the node_id contains both the test file name (test_slurm.py) and the test name (test_slurm), so the configuration file name is unique.
+ move update method from ClustersFactory to Cluster object.
  + The idea is to give the factory the power to create and destroy the clusters, but let the Cluster the power to manage itself.


tests/integration-tests/clusters_factory.py
```diff
@@ -30,11 +30,35 @@ def __init__(self, name, config_file, ssh_key):
         self.__cfn_outputs = None
         self.__cfn_resources = None
 
-    def update(self):
-        """Rewrite the configuration file starting from the self.config object."""
+    def update(self, reset_desired=None, extra_params=None):
```
demartinofra: Shall we set reset_desired=False?

tests/integration-tests/clusters_factory.py
```diff
@@ -30,11 +30,35 @@ def __init__(self, name, config_file, ssh_key):
...
         with open(self.config_file, "w") as configfile:
             self.config.write(configfile)
 
+        # update the cluster
+        logging.info("Updating cluster {0} with config {1}".format(self.name, self.config_file))
+        command = ["pcluster", "update", "--config", self.config_file]
+        if reset_desired:
+            command.append("--reset-desired")
+        if extra_params:
+            command.extend(["--extra-parameters", extra_params])
```
demartinofra: if extra_params is a list this might not work as expected

tests/integration-tests/tests/update/test_update.py
```diff
@@ -23,49 +23,48 @@
 PclusterConfig = namedtuple("PclusterConfig", ["max_queue_size", "compute_instance"])
```
demartinofra: minor: PClusterConfig?

tests/integration-tests/clusters_factory.py
```diff
@@ -30,11 +30,35 @@ def __init__(self, name, config_file, ssh_key):
...
 
-    def update(self):
-        """Rewrite the configuration file starting from the self.config object."""
+    def update(self, reset_desired=None, extra_params=None):
+        """
+        Update a cluster with an already updated config.
+        :param reset_desired: reset the current ASG desired capacity to initial config values
+        :param extra_params: extra parameters to pass to stack update
+        """
+        # Rewrite the configuration file starting from the already updated self.config object
```
demartinofra: since the self.config is updated in a separate method shall we rewrite the file where the config is updated? For instance in _update_cluster_property?

tests/integration-tests/conftest.py
```diff
@@ -176,11 +176,11 @@ def _cluster_factory(cluster_config):
 def _write_cluster_config_to_outdir(request, cluster_config):
     out_dir = request.config.getoption("output_dir")
     os.makedirs(
-        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(request.node.name)),
+        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(request.node.nodeid)),
         exist_ok=True,
     )
     cluster_config_dst = "{out_dir}/clusters_configs/{test_name}.config".format(
-        out_dir=out_dir, test_name=request.node.name
+        out_dir=out_dir, test_name=request.node.nodeid.replace("::", "_")
```
demartinofra: Shall we use a different separator since `_` is already in the test file name and in the test function name? Also `-` might work for me

tests/integration-tests/tests/update/test_update.py
```diff
@@ -23,49 +23,48 @@
...
 
 
-@pytest.mark.regions(["us-west-1"])
+@pytest.mark.regions(["eu-west-1"])
```
demartinofra: No need to change but just to mention it, an alternative to the 4 annotations could have been:
`@pytest.mark.dimensions("eu-west-1", "c5.xlarge", "alinux", "slurm")`

tests/integration-tests/tests/update/test_update.py
```diff
@@ -91,11 +90,11 @@ def _test_update_compute_instance_type(cluster, new_compute_instance):
...
     )
-    _test_compute_instance_type(cluster.cfn_name, new_compute_instance)
+    _test_compute_instance_type(region, cluster.cfn_name, new_compute_instance)
 
 
-def _test_compute_instance_type(stack_name, compute_instance_type):
-    ec2_client = boto3.resource("ec2")
+def _test_compute_instance_type(region, stack_name, compute_instance_type):
+    ec2_client = boto3.resource("ec2", region_name=region)
     instance_ids = []
```
demartinofra: is this used anywhere?

tests/integration-tests/clusters_factory.py
enrico-usai: Fixed, thanks

tests/integration-tests/clusters_factory.py
enrico-usai: extra_params is a json, so a string

tests/integration-tests/tests/update/test_update.py
enrico-usai: fixed

tests/integration-tests/clusters_factory.py
enrico-usai: Moved in the `_update_cluster` to avoid to repeat the write multiple times in the `_update_cluster_property`

tests/integration-tests/conftest.py
enrico-usai: fixed

tests/integration-tests/tests/update/test_update.py
enrico-usai: changed, thanks

tests/integration-tests/tests/update/test_update.py
enrico-usai: removed

PR #1009, opened 2019-04-16
Change sanity_check to not enforce 0.0.0.0/0 CIDR

sean-smith:
Previously, sanity_check forced users to open up their security group to `0.0.0.0/0`, in order
to use FSx or EFS. This isn't necessary to use the service and many users wanted more granularity
in their security_group permissions.

This patch checks the port only, for FSx that's `988` and `2049` for EFS.

[1] https://docs.aws.amazon.com/efs/latest/ug/security-considerations.html#network-access
[2] https://docs.aws.amazon.com/fsx/latest/LustreGuide/limit-access-security-groups.html


cli/pcluster/config_sanity.py
```diff
@@ -49,26 +49,23 @@ def __get_partition(self):
...
-            ip = ip_rule.get("CidrIp")
-            # An existing rule is valid for EFS if, it allows all traffic(0.0.0.0/0)
-            # from all ports or the given port, and does not have a security group restriction
-            if (not port or port == port_to_check) and ip == "0.0.0.0/0" and not group:
-                is_valid = True
-                break
-
-        return is_valid
+        from_port = rule.get("FromPort")
+        to_port = rule.get("FromPort")
```
enrico-usai: you are assigning`get("FromPort")` for both "from_port" and "to_port" variables.

cli/pcluster/config_sanity.py
```diff
@@ -49,26 +49,23 @@ def __get_partition(self):
...
+        ip_protocol = rule.get("IpProtocol")
+
+        # Check to see if the FS accepts connects on port `port_to_check`
+        # checking the sg is open to the CIDR range 0.0.0.0/0 was too stringent
+        # customers need to be able to lock down their sg to their subnet CIDR
+        if (ip_protocol == '-1') or (from_port and from_port <= port_to_check <= to_port):
```
enrico-usai: From the documentation it seems the `-1` value could be also assigned to the `ToPort` and `FromPort` attributes, so the `port_to_check` could not be in the range.

cli/pcluster/config_sanity.py
sean-smith: fixed

cli/pcluster/config_sanity.py
sean-smith: this should be covered by the `(ip_protocol == "-1") ` part of the `or` statement

enrico-usai: From the documentation it seems that is possible to use -1 in the FromPort/ToPort attribute also when IpProtocol is icmp (or [1](http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) as a number):

> FromPort/ToPort: Start/End of port range for the TCP and UDP protocols, or an ICMP type number. If you specify icmp for the IpProtocol property, you can specify -1 as a wildcard (i.e., any ICMP type number).

Furthermore, we can return _True_ when the ip_protocol is -1 or other than icmp (1), tcp (6), udp (17) (we have to verify both string and numbers):

> IpProtocol: The IP protocol name (tcp, udp, icmp) or number (see Protocol Numbers). To specify icmpv6, use a set of IP permissions.
[VPC only] Use -1 to specify all protocols. If you specify -1 or a protocol other than tcp, udp, or icmp, traffic on all ports is allowed, regardless of any ports you specify.
Alternatively, use a set of IP permissions to specify multiple rules and a description for the rule.

Instead we have to return False when the ip_protocol is a value other than tcp, 6 or -1.

sean-smith: Ok, I've updated the check to take into account the `FromPort/ToPort` being set too `-1`, and checking for `IpProtocol == [-1|tcp]`.
### Todo
Add validation for if `IpProtocol` is set to a number.

PR #1019, opened 2019-04-26
Version ParallelCluster Stacks

sean-smith:
* Adds a version tag to newly created stacks
* Disable using a template **other** than the one used to create the
stack in an update
* List the version and status during `pcluster list`
* Color the output based on stack status

image

### DIFF
--- a/cli/pcluster/cli.py
+++ b/cli/pcluster/cli.py
@@ -22,6 +22,8 @@ from botocore.exceptions import NoCredentialsError
 
 from pcluster import easyconfig, pcluster
 
+LOGGER = logging.getLogger("pcluster.pcluster")
+
 
 def create(args):
     pcluster.create(args)
@@ -56,7 +58,8 @@ def update(args):
 
 
 def version(args):
-    pcluster.version(args)
+    version = pcluster.version()
+    LOGGER.info(version)
 
 
 def start(args):
@@ -162,7 +165,7 @@ Examples::
     # update command subparser
     pupdate = subparsers.add_parser(
         "update",
-        help="Updates a running cluster using the values in the config " "file or in a TEMPLATE_URL provided.",
+        help="Updates a running cluster using the values in the config file.",
         epilog="When the command is called and it begins polling for the status of that call, "
         'it is safe to "Ctrl-C" out. You can always return to that status by '
         'calling "pcluster status mycluster".',
@@ -178,7 +181,6 @@ Examples::
         default=False,
         help="Disable CloudFormation stack rollback on error.",
     )
-    pupdate.add_argument("-u", "--template-url", help="Specifies the URL for a custom CloudFormation template.")
     pupdate.add_argument("-t", "--cluster-template", help="Indicates which cluster template to use.")
     pupdate.add_argument("-p", "--extra-parameters", help="Adds extra parameters to the stack update.")
     pupdate.add_argument(
@@ -245,6 +247,7 @@ Examples::
         help="Displays a list of stacks associated with AWS ParallelCluster.",
         epilog="This command lists the names of any CloudFormation stacks named parallelcluster-*",
     )
+    plist.add_argument("--color", action="store_true", default=False, help="Display the cluster status in color.")
     _addarg_config(plist)
     _addarg_region(plist)
     plist.set_defaults(func=list_stacks)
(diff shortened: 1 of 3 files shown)


cli/pcluster/pcluster.py
```diff
@@ -370,6 +372,31 @@ def stop(args):
...
+    Get the version of the stack if tagged.
+
+    :param stack: stack object
+    :return: version or empty string
+    """
+    tags = filter(lambda x: x.get("Key") == "Version", stack.get("Tags"))
+    return tags[0].get("Value") if len(tags) > 0 else ""
+
+
+def colorize(status):
```
demartinofra: This is super cool :) I'm just wondering if this specific syntax is incompatible with any shell/os, if so we might want to provide a flag to enable or disable this feature

cli/pcluster/pcluster.py
```diff
@@ -65,7 +66,7 @@ def create_bucket_with_batch_resources(stack_name, aws_client_config, resources_
 
 def version(args):
```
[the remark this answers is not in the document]
sean-smith: We have to pass it to the `version()` function in `cli.py` because argparser needs it to accept args. From there it's being passed through just to maintain consistency. 

I can remove `args` from `version()` in `pcluster.py`, then remove it from the function call [#L59](https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/cli.py#L59)

cli/pcluster/pcluster.py
sean-smith: It should work across mac/linux and it looks like windows supports it as of windows 10.

https://en.wikipedia.org/wiki/ANSI_escape_code#Platform_support

cli/pcluster/pcluster.py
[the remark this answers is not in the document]
sean-smith: done

cli/pcluster/pcluster.py
```diff
@@ -370,6 +372,31 @@ def stop(args):
...
+def colorize(status):
+    """
+    Color the output, COMPLETE = green, FAILED = red, IN_PROGRESS = yellow.
+
+    :param status: stack status
+    :return: colorized status string
+    """
+    end = "0m"
+    colors = {"COMPLETE": "0;32m", "FAILED": "0;31m", "IN_PROGRESS": "10;33m"}
+    for color in colors:
```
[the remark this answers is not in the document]
sean-smith: done

cli/pcluster/pcluster.py
sean-smith: @enrico-usai Can you test it on windows?

Totally agree on ability to turn it off. Maybe optional cli arg (`--no-color`?)

cli/pcluster/cli.py
```diff
@@ -56,7 +58,8 @@ def update(args):
 
 
 def version(args):
```
enrico-usai: You can remove the `args` parameter from here too

cli/pcluster/pcluster.py
```diff
@@ -370,6 +372,31 @@ def stop(args):
...
 
+def get_version(stack):
+    """
+    Get the version of the stack if tagged.
+
+    :param stack: stack object
+    :return: version or empty string
+    """
+    tags = filter(lambda x: x.get("Key") == "Version", stack.get("Tags"))
+    return tags[0].get("Value") if len(tags) > 0 else ""
```
enrico-usai: `Unexpected error of type TypeError: object of type 'filter' has no len()`
we have to use `len(list(tags))`

cli/pcluster/cli.py
sean-smith: It's required by the way we call `argparse`, each function is called via `args.func(args)`. Removing it results in:

```bash
Unexpected error of type TypeError: version() takes no arguments (1 given)
```

I want to keep the invocation of all cli functions the same (I know command is different but that's for a good reason) hence leave the extra `args` parameter.

cli/pcluster/pcluster.py
sean-smith: fixed

cli/pcluster/pcluster.py
sean-smith: Ok, added `--color` flag

PR #1037, opened 2019-05-06
EFA

sean-smith:
### Background

This add the `enable_efa = compute` flag and validates the parameters to make sure they're sane.

### Validates

```ini
[cluster efa]
base_os = [alinux | centos7]
enable_efa = compute
placement_group != NONE
compute_instance_type = [c5n.18xlarge | p3dn.24xlarge]

[vpc public]
vpc_security_group_id = sg-00b54f8b1f050a52e # if set, checks ingress and egress allow all traffic to itself [1]
```

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -104,6 +104,9 @@ class ParallelClusterConfig(object):
         # Initialize aliases public attributes
         self.__init_aliases()
 
+        # efa checks
+        self.__init_efa_parameters()
+
         # Handle extra parameters supplied on command-line
         try:
             if self.args.extra_parameters is not None:
@@ -518,7 +521,6 @@ class ParallelClusterConfig(object):
             custom_chef_runlist=("CustomChefRunList", None),
             additional_cfn_template=("AdditionalCfnTemplate", None),
             custom_awsbatch_template_url=("CustomAWSBatchTemplateURL", None),
-            enable_efa=("EFA", None),
         )
         for key in cluster_options:
             try:
@@ -530,6 +532,19 @@ class ParallelClusterConfig(object):
                 self.parameters[cluster_options.get(key)[0]] = __temp__
             except configparser.NoOptionError:
                 pass
+
+    def __init_efa_parameters(self):
+        try:
+            __temp__ = self.__config.get(self.__cluster_section, "enable_efa")
+            if __temp__ != "compute":
+                self.__fail("valid values for enable_efa = compute")
+
+            self.__validate_os("EFA", self.__get_os(), ["alinux", "centos7"])
+            self.__validate_scheduler("EFA", self.__get_scheduler(), ["sge", "slurm", "torque"])
+            self.__validate_resource("EFA", self.parameters)
+            self.parameters["EFA"] = __temp__
+        except configparser.NoOptionError:
+            pass
 
     def __init_extra_json_parameter(self):
         """Check for extra_json = { "cluster" : ... } configuration parameters and map to "cfncluster"."""
@@ -597,11 +612,21 @@ class ParallelClusterConfig(object):
         if self.__config.has_option(self.__cluster_section, option):
             self.__fail("option %s cannot be used with awsbatch" % option)
 
+    def __get_scheduler(self):
+        scheduler = "sge"
+        if self.__config.has_option(self.__cluster_section, "scheduler"):
+            scheduler = self.__config.get(self.__cluster_section, "scheduler")
+        return scheduler
+
     def __get_os(self):
         base_os = "alinux"
         if self.__config.has_option(self.__cluster_section, "base_os"):
             base_os = self.__config.get(self.__cluster_section, "base_os")
         return base_os
+
+    def __validate_scheduler(self, service, scheduler, supported_schedulers):
+        if scheduler not in supported_schedulers:
+            self.__fail("%s supports following Schedulers: %s" % (service, supported_schedulers))
 
     def __validate_os(self, service, baseos, supported_oses):
         if baseos not in supported_oses:
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -518,6 +518,7 @@ class ParallelClusterConfig(object):
             custom_chef_runlist=("CustomChefRunList", None),
             additional_cfn_template=("AdditionalCfnTemplate", None),
             custom_awsbatch_template_url=("CustomAWSBatchTemplateURL", None),
+            enable_efa=("EFA", None),
         )
         for key in cluster_options:
             try:
--- a/cli/pcluster/config_sanity.py
+++ b/cli/pcluster/config_sanity.py
@@ -23,7 +23,7 @@ from urllib.parse import urlparse
 import boto3
 from botocore.exceptions import ClientError
 
-from pcluster.utils import get_instance_vcpus, get_supported_batch_instances
+from pcluster.utils import get_instance_vcpus, get_supported_features
 
 
 class ResourceValidator(object):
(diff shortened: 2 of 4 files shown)


sean-smith:
Note, the tox error is:
```
E3002 Invalid Property Resources/ComputeServerLaunchTemplate/Properties/LaunchTemplateData/NetworkInterfaces/0/InterfaceType
aws-parallelcluster.cfn.json:3430:15
```
Which is just saying it doesn't know about the `InterfaceType = EFA` flag yet.

cli/pcluster/config_sanity.py
```diff
@@ -519,6 +519,61 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
         # FSX FS Id check
         elif resource_type in ["fsx_fs_id", "FSx_storage_capacity", "FSx_imported_file_chunk_size", "FSx_export_path"]:
             self.__validate_fsx_parameters(resource_type, resource_value)
+        elif resource_type == "EFA":
+            valid_instances = get_supported_efa_instances(self.region)
+            if "ComputeInstanceType" in resource_value:
+                compute_instance_type = resource_value["ComputeInstanceType"]
```
enrico-usai: we could simplify with something like: 
```python
if resource_value.get("ComputeInstanceType") not in valid_instances
```

cli/pcluster/config_sanity.py
```diff
@@ -519,6 +519,61 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+                if compute_instance_type not in valid_instances:
+                    self.__fail(resource_type, "Compute Instance needs to be one of %s" % valid_instances)
+            else:
+                self.__fail(resource_type, "Compute Instance needs to be one of %s" % valid_instances)
+            if "PlacementGroup" in resource_value:
```
enrico-usai: we could simplify with something like:
```python
if resource_value.get("PlacementGroup", "NONE") == "NONE"
```

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1195,6 +1200,14 @@
           ]
         }
       ]
+    },
+    "EnableEFA": {
+      "Fn::Equals": [
+        {
+          "Ref": "EFA"
+        },
+        "compute"
```
demartinofra: shall we do a check for not NONE as done for the others so that we don't need to change this if we support additional options?

cli/pcluster/config_sanity.py
```diff
@@ -530,6 +530,56 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
         # FSX FS Id check
         elif resource_type in ["fsx_fs_id", "FSx_storage_capacity", "FSx_imported_file_chunk_size", "FSx_export_path"]:
             self.__validate_fsx_parameters(resource_type, resource_value)
+        elif resource_type == "EFA":
+            valid_instances = get_supported_efa_instances(self.region)
+            if "ComputeInstanceType" in resource_value:
```
demartinofra: can you simplify the if else branch with a default in the get?
```
if resource_value.get("ComputeInstanceType", None) not in valid_instances:
```

cli/pcluster/config_sanity.py
```diff
@@ -530,6 +530,56 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
         # FSX FS Id check
         elif resource_type in ["fsx_fs_id", "FSx_storage_capacity", "FSx_imported_file_chunk_size", "FSx_export_path"]:
             self.__validate_fsx_parameters(resource_type, resource_value)
+        elif resource_type == "EFA":
+            valid_instances = get_supported_efa_instances(self.region)
```
demartinofra: shall we move this to a __validate_efa_parameters function? Ideally we should also start writing unit tests for the new code we add to pcluster cli.

cli/pcluster/config_sanity.py
```diff
@@ -530,6 +530,56 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+                    in_rules = sg.get("IpPermissions")
+                    out_rules = sg.get("IpPermissionsEgress")
+
+                    allowed_in = False
+                    allowed_out = False
+                    for rule in in_rules:
+                        if (
+                            rule.get("IpProtocol") == "-1"
+                            and len(rule.get("UserIdGroupPairs")) > 0
+                            and rule.get("UserIdGroupPairs")[0].get("GroupId") == sg_id
```
demartinofra: Can this have more than one GroupPair?

cli/pcluster/config_sanity.py
```diff
@@ -530,6 +530,56 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+                            break
+                    if not (allowed_in and allowed_out):
+                        self.__fail(
+                            resource_type,
+                            "VPC Security Group %s must allow all traffic in and out from itself. "
+                            "See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security"
+                            % sg_id,
+                        )
+                except ClientError as e:
+                    self.__fail(resource_type, e.response.get("Error").get("Message"))
```
demartinofra: Since we don't have any stacktrace for this failures let's maybe leave an additional message such as failed while validating EFA resources with exception ....

cli/pcluster/utils.py
```diff
@@ -131,6 +131,23 @@ def _get_json_from_s3(region, file_name):
...
+    """
+    Get a json object containing the instances supported by efa.
+
+    :param region: AWS Region
+    :param instance_type: the instance type to search for.
+    :return: json object containing the instances supported by efa
+    or an empty object if unable to parse/get the instance list file
+    """
+    try:
+        instances = _get_json_from_s3(region, "instances/efa_instances.json")
```
demartinofra: would it make sense to add efa details to the current instances.json file or have a more general purpose json file we can use to dynamically enable support for different features so that we don't need to pull a single file per feature?

cli/pcluster/config_sanity.py
sean-smith: In the `fail()` function we print out the `resource_type`. For example, a misconfigured security group results in:

```bash
$ pcluster create efa
Beginning cluster creation for cluster: efa
Config sanity error on resource EFA: VPC Security Group sg-00e715cfdebcb97bb must allow all traffic in and out from itself. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security
```

cloudformation/aws-parallelcluster.cfn.json
sean-smith: done

cli/pcluster/utils.py
sean-smith: yes, let's talk about the best way to do this

cli/pcluster/config_sanity.py
sean-smith: Here's what the `rule` object looks like:

```
{u'PrefixListIds': [], u'FromPort': 0, u'IpRanges': [], u'ToPort': 65535, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [{u'UserId': '822857487308', u'GroupId': 'sg-00e715cfdebcb97bb'}], u'Ipv6Ranges': []}
{u'PrefixListIds': [], u'FromPort': 22, u'IpRanges': [{u'CidrIp': '0.0.0.0/0'}], u'ToPort': 22, u'IpProtocol': 'tcp', u'UserIdGroupPairs': [], u'Ipv6Ranges': []}
```

I've done some testing to see if I can extend `UserIdGroupPairs` to make it longer than 1. I can't so I've added a comment explaining this behavior.

cli/pcluster/utils.py
sean-smith: I've revised the creation of this file in https://github.com/aws/aws-parallelcluster/pull/1067

demartinofra: Will you change this in order to call get_supported_instances? Also shall we fetch also schedulers and oses?

sean-smith: Yep 👍

Changed to `get_supported_instances()` everywhere.

cli/pcluster/cfnconfig.py
```diff
@@ -529,6 +532,18 @@ def __init_cluster_parameters(self):
             except configparser.NoOptionError:
                 pass
 
+    def __init_efa_parameters(self):
+        try:
+            __temp__ = self.__config.get(self.__cluster_section, "enable_efa")
+            if __temp__ != "compute":
+                self.__fail("valid values for enable_efa = compute")
+
+            self.__validate_os("EFA", self.__get_os(), ["alinux", "centos7"])
```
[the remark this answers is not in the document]
sean-smith: yep, added

sean-smith:
@demartinofra Ok, I’ve updated the patch to incorporate your feedback:
1. s3 path changed to `features/feature_whitelist.json`
2. validate `base_os` and `scheduler` based on features set in `feature_whitelist.json`

cli/pcluster/cfnconfig.py
```diff
@@ -530,6 +533,23 @@ def __init_cluster_parameters(self):
...
 
+    def __init_efa_parameters(self):
+        try:
+            __temp__ = self.__config.get(self.__cluster_section, "enable_efa")
+            if __temp__ != "compute":
+                self.__fail("valid values for enable_efa = compute")
+
+            supported_features = get_supported_features(self.region, "efa")
+            valid_oses = supported_features.get("baseos")
+            valid_schedulers = supported_features.get("baseos")
```
demartinofra: Forgot to replace “baseos” with schedulers

cli/pcluster/utils.py
```diff
@@ -131,21 +131,35 @@ def _get_json_from_s3(region, file_name):
...
+    :return: json object containing all the attributes supported by feature
     or an empty object if unable to parse/get the instance list file
     """
     try:
-        instances = _get_json_from_s3(region, "instances/batch_instances.json")
+        features = _get_json_from_s3(region, "features/feature_whitelist.json")
+        supported_features = features.get("Features").get(feature)
     except (ValueError, ClientError):
-        instances = ""
+        supported_features = {}
```
demartinofra: I would actually fail with unable to retrieve... rather than returning empty object otherwise it seems the validation failed for some other unmet requirement.
Also shall we catch the broader Exception so that we also consider KeyError?

cli/pcluster/cfnconfig.py
sean-smith: done, also I've removed the fetching the json from this portion to just do hardcoded validation (for now) as discussed.

cli/pcluster/utils.py
sean-smith: Ok, I've changed the exception logic to give customers a workaround should something fail. This is coupled with the above change to have all calls to `get_supported_features()` in config_sanity.py so it can be disabled if anything breaks.

PR #1041, opened 2019-05-08
AWS Batch GPU Support

sean-smith:
### AWS Batch GPU Jobs 
To submit a GPU job, just do:

```
$ awsbsub --gpus 1 nvidia-smi
```
For this to work, there's three requirements:

1. **AMI** - AWS Batch uses the ecs GPU-optimized ami for any P Family instance. See [1]
2. **Container** - nvidia-docker or deep learning container needs to be used here. See [2] and [3]. Container will be a separate PR.
3. **Instance** - Needs to be one of `p2`, `p3` families

[1] https://docs.aws.amazon.com/batch/latest/userguide/batch-gpu-ami.html
[2] https://docs.aws.amazon.com/dlami/latest/devguide/deep-learning-containers-images.html
[3] https://github.com/NVIDIA/nvidia-docker

### DIFF
--- a/cli/awsbatch/awsbsub.py
+++ b/cli/awsbatch/awsbsub.py
@@ -127,6 +127,12 @@ def _get_parser():
         help="The number of nodes to reserve for the job. It enables Multi-Node Parallel submission",
         type=int,
     )
+    parser.add_argument(
+        "-g",
+        "--gpus",
+        help="The number of gpus to attach to the container. GPU jobs will only run p2 or p3 instance families.",
+        type=int,
+    )
     # array parameters
     parser.add_argument(
         "-a",
@@ -438,9 +444,11 @@ class AWSBsubCommand(object):
         job_definition,
         job_name,
         job_queue,
+        compute_environment,
         command,
         nodes=None,
         vcpus=None,
+        gpus=None,
         memory=None,
         array_size=None,
         retry_attempts=1,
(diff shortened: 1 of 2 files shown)


cli/awsbatch/awsbsub.py
```diff
@@ -127,6 +127,12 @@ def _get_parser():
         help="The number of nodes to reserve for the job. It enables Multi-Node Parallel submission",
         type=int,
     )
+    parser.add_argument(
+        "-g",
+        "--gpus",
+        help="The number of gpus to attach to the container. GPU jobs will only run on P family instances.",
```
enrico-usai: I'd write: _"only run on p2 and p3 instance families"_ according to the [documentation](https://docs.aws.amazon.com/en_us/batch/latest/userguide/batch-gpu-ami.html). What do you think?

cli/awsbatch/awsbsub.py
```diff
@@ -508,6 +519,27 @@ def run(  # noqa: C901 FIXME
...
+        :return: None for jobs
+        """
+        ce = self.batch_client.describe_compute_environments(computeEnvironments=[self.config.compute_environment]).get(
+            "computeEnvironments"
+        )[0]
+
+        instances = ce.get("computeResources").get("instanceTypes")
+
+        for instance in instances:
+            if str(instance).startswith("p"):
```
enrico-usai: Does it works with p3dn.24xlarge instances?

cli/awsbatch/awsbsub.py
```diff
@@ -465,6 +473,9 @@ def run(  # noqa: C901 FIXME
                 container_overrides.update(vcpus=vcpus)
             if memory:
                 container_overrides.update(memory=memory)
+            if gpus:
+                self.__validate_instance_type()
```
demartinofra: Can't we just let the boto3 batch client perform the validations and fail on the API call instead of having them explicitly performed in our code? Otherwise we cannot guarantee to be aligned with batch requirements and will be hard to update our validations whenever batch does that

cli/awsbatch/awsbsub.py
demartinofra: Or just mention the fact that these jobs run on a subset of instances and link to the batch docs

PR #1049, opened 2019-05-14
cli: show info message for new package version only with create command

enrico-usai:
I'm changing warning to info and removing the message from all the commands but the create one.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ x.x.x
 **CHANGES**
 
 * Increase default EBS volume size from 17GB to 20GB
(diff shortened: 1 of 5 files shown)


sean-smith:
Why create only? Also the create will go ahead without the new version correct?

enrico-usai:
> Why create only? Also the create will go ahead without the new version correct?

The idea is to show the message that there is a new available version only when the user is going to create a new cluster. In this way he could think to update the product and create a new cluster with the new cli.

With this patch we also have a more consinstent output of the *list*, *status*, *instances*, etc commands, without warning messages in the middle.

PR #1054, opened 2019-05-16
Integ tests: add compute_root_volume_size test

enrico-usai:
Added a new test to verify the `compute_root_volume_size` is updated in the new compute nodes and it is not updated in old compute nodes.

Other improvements: 
+ added functionality to retrieve compute nodes list
+ added functionality to submit a job in a specific node
+ create a function to add new compute nodes
+ verify updated params on both OLD and NEW compute nodes
+ improve compute_instance_type test to check a specific node


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -195,14 +201,18 @@ def assert_job_submitted(self, sbatch_output):  # noqa: D102
...
 
-    def submit_script(self, script, nodes=1):  # noqa: D102
+    def submit_script(self, script, nodes=1, host=None):  # noqa: D102
         script_name = os.path.basename(script)
-        return self._remote_command_executor.run_remote_command(
-            "sbatch -N {0} {1}".format(nodes, script_name), additional_files=[script]
-        )
+        submission_command = "sbatch -N {0} {1}".format(nodes, script_name)
+        if host:
+            submission_command += " --nodelist={0}".format(host)
```
demartinofra: does this work if we append the --nodelist after the scriptname?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -110,16 +162,43 @@ def _test_update_compute_instance_type(region, cluster, new_compute_instance):
...
+        assert_that(instance_types).does_not_contain(compute_instance_type)
+    else:
+        assert_that(instance_types).contains(compute_instance_type)
+
+
+def _test_compute_root_volume_size(
+        command_executor, slurm_commands, test_datadir, compute_root_volume_size, host, must_fail=False
+):
+    # submit a job to retrieve compute root volume size and save in a file
+    result = slurm_commands.submit_script(str(test_datadir / "slurm_get_root_volume_size.sh"), host=host)
```
demartinofra: Did you forget to add the script to the PR?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -110,16 +162,43 @@ def _test_update_compute_instance_type(region, cluster, new_compute_instance):
         max_monitoring_time=minutes(estimated_scaleup_time),
         number_of_nodes=number_of_nodes,
     )
-    _test_compute_instance_type(region, cluster.cfn_name, new_compute_instance)
+
+    return [node for node in slurm_commands.get_compute_nodes() if node not in initial_compute_nodes]
 
 
-def _test_compute_instance_type(region, stack_name, compute_instance_type):
+def _test_compute_instance_type(region, stack_name, compute_instance_type, host, must_fail=False):
```
demartinofra: must_fail -> expect_to_fail?
same for the other function

tests/integration-tests/tests/update/test_update.py
```diff
@@ -21,63 +22,106 @@
...
+    # verify updated parameters
     _test_max_queue(region, cluster.cfn_name, updated_config.max_queue_size)
-    _test_update_compute_instance_type(region, cluster, updated_config.compute_instance)
     _test_s3_read_resource(region, cluster, updated_config.s3_read_resource)
     _test_s3_read_write_resource(region, cluster, updated_config.s3_read_write_resource)
 
+    # verify updated params in OLD compute nodes
+    compute_nodes = slurm_commands.get_compute_nodes()
+    _test_compute_instance_type(
+        region, cluster.cfn_name, updated_config.compute_instance_type, compute_nodes[0], must_fail=True
```
demartinofra: rather than having the must_fail logic you could assert that the instance type and the size still match the previous values.

tests/integration-tests/tests/common/schedulers_common.py
enrico-usai: it works because the integration test works, but I changed it to be aligned with the documentation: https://slurm.schedmd.com/sbatch.html

tests/integration-tests/tests/update/test_update.py
enrico-usai: Good point

tests/integration-tests/tests/update/test_update.py
enrico-usai: Removed since now I'm matching the previous value.

tests/integration-tests/tests/update/test_update.py
enrico-usai: Added

PR #1056, opened 2019-05-16
doc: update network configuration for compute_subnet_cidr case

matteofiordarancio:
I verified with a test that for compute_subnet_cidr no Nat is needed

### DIFF
--- a/docs/networking.rst
+++ b/docs/networking.rst
@@ -47,6 +47,7 @@ The configuration to create a new private subnet for compute instances requires 
   vpc_id = vpc-xxxxxx
   master_subnet_id = subnet-<public>
   compute_subnet_cidr = 10.0.1.0/24
+  use_public_ips = true
 
 The configuration to use an existing private network requires the following settings:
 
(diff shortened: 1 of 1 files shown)


docs/networking.rst
```diff
@@ -57,7 +57,7 @@ The configuration to use an existing private network requires the following sett
   master_subnet_id = subnet-<public>
   compute_subnet_id = subnet-<private>
 
-Both these configuration require to have a `NAT Gateway
+Note that this second configuration require to have a `NAT Gateway
```
enrico-usai: Could you please verify if it works by using `use_public_ips = false`?

matteofiordarancio: it does not work.

enrico-usai: require --> requires

enrico-usai: Thank you.
We could use the `..note::` syntax to show the message as an annotation:
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

Then: require --> requires

matteofiordarancio: Done

docs/networking.rst
```diff
@@ -57,9 +58,10 @@ The configuration to use an existing private network requires the following sett
   master_subnet_id = subnet-<public>
   compute_subnet_id = subnet-<private>
 
-Both these configuration require to have a `NAT Gateway
-<https://docs.aws.amazon.com/vpc/latest/userguide/vpc-nat-gateway.html>`_
-or an internal PROXY to enable web access for compute instances.
+.. note::
+    This second configuration require to have a `NAT Gateway
```
enrico-usai: "This second configuration require to have a" --> "This second configuration requires a"

matteofiordarancio: Done

PR #1057, opened 2019-05-17
update sanity check to solve cidr_value and public_ips conflicts

matteofiordarancio:
I verified with a test that for compute_subnet_cidr public_ips has to be
set to true, so I added the check into the code.

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -354,6 +354,11 @@ class ParallelClusterConfig(object):
                     "VPC section [%s] used in [%s] section is not defined" % (vpc_section, self.__cluster_section)
                 )
 
+        # Check that cidr and public ips are not both set
+        cidr_value = self.__config.get(vpc_section, "compute_subnet_cidr", fallback=None)
+        public_ips = self.__config.getboolean(vpc_section, "use_public_ips", fallback=True)
+        ResourceValidator.validate_vpc_coherence(cidr_value, public_ips)
+
     def __check_account_capacity(self):
         """Try to launch the requested number of instances to verify Account limits."""
         if self.parameters.get("Scheduler") == "awsbatch" or self.parameters.get("ClusterType", "ondemand") == "spot":
--- a/cli/pcluster/config_sanity.py
+++ b/cli/pcluster/config_sanity.py
@@ -47,6 +47,17 @@ class ResourceValidator(object):
         return "aws"
 
     @staticmethod
+    def validate_vpc_coherence(cidr_value, public_ip):
+        """
+        Check that cidr_value and public_ip parameters are not conflicting.
+
+        :param cidr_value: the value of compute_subnet_cidr set by the user (default should be None)
+        :param public_ip: the value of use_public_ips set by the user (default should be True)
+        """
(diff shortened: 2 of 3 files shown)


cli/pcluster/config_sanity.py
```diff
@@ -343,6 +343,11 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
             ):
                 self.__fail(resource_type, "DNS Hostnames not enabled in %s" % resource_value)
         # VPC Subnet Id
+        elif resource_type == "VPC_COHERENCE":
```
enrico-usai: I think you are missing the cfnconfig.py file in this PR

matteofiordarancio: done

cli/pcluster/config_sanity.py
```diff
@@ -343,6 +343,11 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
...
+            cidr_value = resource_value[0]
+            public_ip = resource_value[1]
+            if cidr_value and public_ip is False:
+                self.__fail(resource_type, "compute_subnet_cidr needs use_public_ips to be true")
```
demartinofra: why is this required? is it always true? I'm wondering if it's a bug or a necessary requirement

cli/pcluster/cfnconfig.py
```diff
@@ -354,6 +354,11 @@ def __init_vpc_parameters(self):
                     "VPC section [%s] used in [%s] section is not defined" % (vpc_section, self.__cluster_section)
                 )
 
+        # Check that cidr and public ips are not both set
+        cidr_value = self.__config.get(vpc_section, "compute_subnet_cidr", fallback=None)
+        public_ips = self.__config.getboolean(vpc_section,"use_public_ips", fallback=True)
+        self.__validate_resource("VPC_COHERENCE", (cidr_value, public_ips))
```
demartinofra: if possible let's avoid adding more and more code to this __validate_resource function :)

matteofiordarancio: how would you do that if not adding code there?

demartinofra: You can extract the code under the `elif resource_type == "VPC_COHERENCE":` branch into a separate function (something like validate_vpc_coherence) and call that in place of the huge validate function.

matteofiordarancio: Done, check if it is what you meant

cli/pcluster/cfnconfig.py
```diff
@@ -354,6 +354,11 @@ def __init_vpc_parameters(self):
                     "VPC section [%s] used in [%s] section is not defined" % (vpc_section, self.__cluster_section)
                 )
 
+        # Check that cidr and public ips are not both set
+        cidr_value = self.__config.get(vpc_section, "compute_subnet_cidr", fallback=None)
+        public_ips = self.__config.getboolean(vpc_section, "use_public_ips", fallback=True)
+        ResourceValidator.validate_vpc_coherence(cidr_value, public_ips)
```
demartinofra: we forgot an if self.__sanity_check:

Issue #202, opened 2017-07-07
post_install timeout


shenkers:
I have a very long running post_install script, and it seems that cfncluster kills the cluster before the script completes. When I run the long running commands as a separate nohup'd process the cluster boots up successfully. Is there a configuration to allow for long-running startup scripts?

Thanks,
Sol

bwbarrett:
The timeout is hard coded (in two places, one for the master server and one for the compute servers) in the CloudFormation template.  You will need to use a custom CloudFormation template to increase the timeout.  The default template is [here](https://github.com/awslabs/cfncluster/blob/develop/cloudformation/cfncluster.cfn.json), and then you'll need to use the `template_url` option, documented [here](http://cfncluster.readthedocs.io/en/latest/configuration.html#template-url).

If you have any other questions, reopen this issue or open a new issue.

Good luck,

Brian

shenkers:
Thanks Brian, that clarifies it! Just to put more of the details in this thread, the section of the default template that needs to be edited will look something like this:

```
"CreationPolicy" : {
        "ResourceSignal" : {
          "Timeout" : "PT30M",
          "Count" : {
            "Ref" : "ComputeWaitConditionCount"
          }
        }
      }
```

Where the "PT30M" above means it will wait 30M before timing-out

http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-general.html#command-options-general-autoscalingupdatepolicyrollingupdate``

Issue #147, opened 2016-11-03
CUDA not installed at startup on Ubuntu 16.04


skoocda:
Created a cluster with (abridged) config:

>[cluster default]
>master_instance_type = m4.xlarge
>compute_instance_type = g2.2xlarge
>base_os = ubuntu1604
>post_install = s3://<PATH_TO_FILE>.sh

NVIDIA driver installed correctly, but CUDA toolkit did not. 

No entry exists in `/usr/lib`, `/usr/bin`, `/usr/local` or any other logical location

Nor was nvcc shown as installed, as listed by `nvcc -V`

However, `sudo apt install nvcc` worked correctly when run manually in the compute instance.

From issue #140 and #123 I surmise that this is not intended behaviour, as CUDA should be installed upon creation.

skoocda:
Is there an option to specify a variable for CustomChefRunList or can I potentially add this to extraJson:
`
“nvidia” : {
        “enabled” : ‘yes’
    }
`

skoocda:
Seems to be caused by the c++ compiler version, which is incompatible with CUDA 7.5

mohanasudhan:
We don't plan to add native cuda support in cfncluster.

mohanasudhan:
My bad !!! I will take a deep look at this issue tomorrow.

mohanasudhan:
Apologies for delay in looking at this issue (was working on cfncluster new release). As part of cfncluster 1.4.0 we upgrade to cuda9 and nvidia driver. Can you confirm you still see this issue? Note : we are not installing nvcc as part of cookbook

jprobichaud:
I'm using cfncluster 1.4.1 and I've put the following snippet in my cluster definition:

```
base_os = ubuntu1604
extra_json : { "nvidia" : { "enabled" : "yes" } }
compute_instance_type = p3.8xlarge
master_instance_type = c5.4xlarge
```
On the compute nodes, /opt has no traces of CUDA of any sort.  What am I doing wrong?

demartinofra:
As part of ParallelCluster 2.5.0 we updated Nvidia drivers and CUDA libraries and introduced tests that verify their setup. I'm going to resolve this one.

Issue #105, opened 2016-05-26
Document `tenancy` option


guicho271828:
Amazon clusters generally have varying range of performance instability, due to hyperthreading cores, effect of hypervisors, and the other users running a program on the same machine. While this is ok with batch jobs i.e. the jobs that solely aim to get the jobs done, this is problematic for running massive computational experiments [1]. Unless using dedicated instances, time measurements are unreliable and the reproducibility is not guaranteed.

currently `cluster_type` supports `ondemand` and `spot` only.

[1] Kotthoff, Lars. "Reliability of computational experiments on virtualised hardware." Journal of Experimental & Theoretical Artificial Intelligence 26.1 (2014): 33-49.

dougalb:
Hi, I think you will find the following AWS re:Invent session very helpful in understanding performance on EC2 - https://www.youtube.com/watch?v=SZAvtbrIBAk. 

To address your specific question, the `cluster_type` parameter refers to the purchasing model. Customer who have a need for dedicated tenancy, would use the `tenancy` parameter. More details of Dedicated Tenancy can be found here - http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/dedicated-instance.html. Please note, EC2 performance is not related to tenancy.

guicho271828:
is `tenancy` parameter some sort of undocumented option? I couldnt find it here.
http://cfncluster.readthedocs.io/en/latest/configuration.html#cluster-type

dougalb:
Oops, looks like it's missing in the docs. It is a separate parameter in the `cluster` section. It supports the values __default__ and __dedicated__. We'll get the docs updated.

Again please note, tenancy is not related performance or resource allocations of an instance.

guicho271828:
thanks for the quick response. ok, it was just not in the doc.

Performance is surely affected when there are other users on the same physical machine. For example, if one user is running a very memory-intensive application with a frequent read/write to the main memory, it results in a memory bus contention which affects the performance [2]. So, as other users read/write more memory, my program gets slower.

Again, note that the problem is not the absolute performance i.e. that my program gets slower, but that the performance is unstable/unpredictable. It is perfectly OK if the instance is consistently slow. However, I cannot predict if it gets slower or not, since I cannot control the behavior of the other users on the same phisical machine.

[...]

PR #1058, opened 2019-05-17
Remove undocumented CFN input parameters

enrico-usai:
**Tenancy** was not exposed to the users in the configuration file nor documented:

- https://github.com/aws/aws-parallelcluster/issues/105 Document tenancy option
> you shouldn't use dedicated instances for performance reasons.

___

**CustomChefRunList**  was not exposed to the users in the configuration file nor documented:

- https://github.com/aws/aws-parallelcluster/pull/125 Add missing parameters to cfnconfig
- https://github.com/aws/aws-parallelcluster/issues/147 CUDA not installed at startup on Ubuntu 16.04

___

**ComputeWaitConditionCount** was always set to DesiredSize at creation time and not set in the update. It was used only for [CreationPolicy](https://docs.aws.amazon.com/en_us/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html)

- https://github.com/aws/aws-parallelcluster/issues/100 CfnCluster generated CloudFormation template
> These two values must be the same during a StackCreate call and afterwards may be different depending on the UpdateStack call.
- https://github.com/aws/aws-parallelcluster/issues/202 post_install timeout


sean-smith:
Any reason this hasn’t been merged in?

PR #1074, opened 2019-05-27
Extend pcluster configure to ask for more config parameters

matteofiordarancio:
This pull request is a re-factorization of pcluster configure. 

1.  Support for scheduler, os, master instance type and compute instance type has been added
2. General re-factorization of the code
3. Fixed a bug for which the configuration was updated even if there was an error in it

### DIFF
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -29,10 +29,12 @@ from botocore.exceptions import BotoCoreError, ClientError
 
 from . import cfnconfig
 from pcluster.utils import get_supported_schedulers
+from pcluster.utils import get_supported_os
 
 logger = logging.getLogger("pcluster.pcluster")
 unsupported_regions = ["ap-northeast-3"]
 default_region = "us-east-1"
+
 
 def handle_client_exception(func):
     @functools.wraps(func)
@@ -185,6 +187,18 @@ def configure(args):  # noqa: C901 FIXME!!!
         options=get_supported_schedulers(),
         check_validity=True,
     )
+    is_batch = _is_aws_batch(scheduler)
+
+    if is_batch:
+        operating_system = "alinux"
+    else:
+        operating_system = prompt(
+            "Operating System",
+            config.get("cluster " + cluster_template, "base_os") if config.has_option("cluster " + cluster_template,
+                                                                                      "base_os") else "alinux",
+            options=get_supported_os(),
+            check_validity=True,
+        )
 
     vpcname = prompt(
         "VPC Name",
@@ -290,3 +304,14 @@ def _is_config_valid(args, config):
         if is_file_ok:
             return True
         return False
+
+
+def _is_aws_batch(scheduler):
+    """
+    Return true if the scheduler is awsbatch
+    :param scheduler: the scheduler to check
+    :return: true if the scheduler is awsbatch
+    """
+    if scheduler == "awsbatch":
+        return True
+    return False
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -84,20 +84,26 @@ def get_regions():
 
 
 @handle_client_exception
-def ec2_conn(aws_region_name):
+def ec2_get_region(aws_region_name):
     if aws_region_name:
         region = aws_region_name
     elif os.environ.get("AWS_DEFAULT_REGION"):
         region = os.environ.get("AWS_DEFAULT_REGION")
     else:
         region = "us-east-1"
+    return region
 
+
+@handle_client_exception
+def ec2_conn(aws_region_name):
+    region = ec2_get_region(aws_region_name)
     ec2 = boto3.client("ec2", region_name=region)
     return ec2
 
 
 @handle_client_exception
 def list_keys(aws_region_name):
+    aws_region_name = ec2_get_region(aws_region_name)  # we get the default if not present
     conn = ec2_conn(aws_region_name)
     keypairs = conn.describe_key_pairs()
     keynames = []
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -24,6 +24,7 @@ from builtins import input
 
 import boto3
 import configparser
+import tempfile
 from botocore.exceptions import BotoCoreError, ClientError
 
 from . import cfnconfig
@@ -220,7 +221,7 @@ def configure(args):  # noqa: C901 FIXME!!!
     s_aws = {"__name__": "aws", "aws_region_name": aws_region_name}
     s_aliases = {"__name__": "aliases", "ssh": "ssh {CFN_USER}@{MASTER_IP} {ARGS}"}
     s_cluster = {"__name__": "cluster " + cluster_template, "key_name": key_name, "vpc_settings": vpcname,
-                 "scheduler": scheduler}
+                 "scheduler": scheduler, "base_os": operating_system}
     s_vpc = {"__name__": "vpc " + vpcname, "vpc_id": vpc_id, "master_subnet_id": master_subnet_id}
 
     sections = [s_aws, s_cluster, s_vpc, s_global, s_aliases]
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
         if e.errno != errno.EEXIST:
             raise  # can safely ignore EEXISTS for this purpose...
 
+    if not _is_config_valid(args, config):
+        sys.exit(1)
+
+    # If we are here, than the file it's correct and we can override it.
     # Write configuration to disk
     open(config_file, "a").close()
     os.chmod(config_file, stat.S_IRUSR | stat.S_IWUSR)
     with open(config_file, "w") as cf:
         config.write(cf)
 
+
+def _is_config_valid(args, config):
+    """
+    Validate the configuration of the pcluster configure.
+    :param args: the arguments passed with the command line
+    :param config: the configParser
+    :return True if the configuration is valid, false otherwise
+    """
+    # We create a temp_file to validate before overriding the original config
+    path = os.path.join(tempfile.gettempdir(), "temp_config")
+    temp_file = path
+    temp_args = args
+    temp_args.config_file = path
+    open(temp_file, "a").close()
+    os.chmod(temp_file, stat.S_IRUSR | stat.S_IWUSR)
+    with open(temp_file, "w") as cf:
+        config.write(cf)
     # Verify the configuration
-    cfnconfig.ParallelClusterConfig(args)
+    is_file_ok = True
+    try:
+        cfnconfig.ParallelClusterConfig(temp_args)
+    except SystemExit as e:
+        is_file_ok = False
+    finally:
+        os.remove(path)
+        if is_file_ok:
+            return True
+        return False
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -15,26 +15,38 @@ from future import standard_library  # isort:skip
 standard_library.install_aliases()
 # fmt: on
 
+import copy
 import errno
 import functools
 import logging
 import os
 import stat
 import sys
+import tempfile
 from builtins import input
 
 import boto3
 import configparser
-import tempfile
 from botocore.exceptions import BotoCoreError, ClientError
 
+from pcluster.utils import get_supported_os, get_supported_schedulers
+
 from . import cfnconfig
-from pcluster.utils import get_supported_schedulers
-from pcluster.utils import get_supported_os
 
 logger = logging.getLogger("pcluster.pcluster")
 unsupported_regions = ["ap-northeast-3"]
-default_region = "us-east-1"
+DEFAULT_VALUES = {
+    "cluster_template": "default",
+    "aws_region_name": "us-east-1",
+    "scheduler": "sge",
+    "os": "alinux",
+    "max_queue_size": "10",
+    "master_instance_type": "t2.micro",
+    "compute_instance_type": "t2.micro",
+    "vpc_name": "public",
+    "initial_size": "1",
+}
+FORCED_BATCH_VALUES = {"os": "alinux", "compute_instance_type": "optimal"}
 
 
 def handle_client_exception(func):
@@ -51,18 +63,18 @@ def handle_client_exception(func):
     return wrapper
 
 
-def prompt(prompt, default_value=None, hidden=False, options=None, check_validity=False):
+def prompt(message, default_value=None, hidden=False, options=None, check_validity=False):
     if hidden and default_value is not None:
-        user_prompt = prompt + " [*******" + default_value[-4:] + "]: "
+        user_prompt = message + " [*******" + default_value[-4:] + "]: "
     else:
-        user_prompt = prompt + " ["
+        user_prompt = message + " ["
         if default_value is not None:
             user_prompt = user_prompt + default_value + "]: "
         else:
             user_prompt = user_prompt + "]: "
 
     if isinstance(options, (list, tuple)):
-        print("Acceptable Values for %s: " % prompt)
+        print("Acceptable Values for %s: " % message)
         for o in options:
             print("    %s" % o)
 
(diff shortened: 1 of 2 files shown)


cli/pcluster/easyconfig.py
```diff
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    :param args: the arguments passed with the command line
+    :param config: the configParser
+    :return True if the configuration is valid, false otherwise
+    """
+    # We create a temp_file to validate before overriding the original config
+    path = os.path.join(tempfile.gettempdir(), "temp_config")
+    temp_file = path
+    temp_args = args
+    temp_args.config_file = path
+    open(temp_file, "a").close()
```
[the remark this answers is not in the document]
enrico-usai: We can avoid the append since the file doesn't exist and just open the file for write in the following lines.

cli/pcluster/easyconfig.py
```diff
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    os.chmod(temp_file, stat.S_IRUSR | stat.S_IWUSR)
+    with open(temp_file, "w") as cf:
+        config.write(cf)
     # Verify the configuration
-    cfnconfig.ParallelClusterConfig(args)
+    is_file_ok = True
```
enrico-usai: I'd rename to `is_valid`

cli/pcluster/easyconfig.py
```diff
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    try:
+        cfnconfig.ParallelClusterConfig(temp_args)
+    except SystemExit as e:
```
enrico-usai: we can avoid the `as e` since we are not using it.

cli/pcluster/easyconfig.py
```diff
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+        is_file_ok = False
+    finally:
+        os.remove(path)
+        if is_file_ok:
```
enrico-usai: you can write: `return is_valid` without the if block

cli/pcluster/easyconfig.py
```diff
@@ -84,20 +84,26 @@ def get_regions():
 
 
 @handle_client_exception
-def ec2_conn(aws_region_name):
+def ec2_get_region(aws_region_name):
```
enrico-usai: we could rename as `_evaluate_aws_region` or `_get_aws_region`

cli/pcluster/easyconfig.py
```diff
@@ -84,20 +84,26 @@ def get_regions():
...
+@handle_client_exception
+def ec2_conn(aws_region_name):
+    region = ec2_get_region(aws_region_name)
     ec2 = boto3.client("ec2", region_name=region)
     return ec2
 
 
 @handle_client_exception
 def list_keys(aws_region_name):
+    aws_region_name = ec2_get_region(aws_region_name)  # we get the default if not present
```
enrico-usai: we can avoid the comment since it is described in the called method

cli/pcluster/easyconfig.py
```diff
@@ -32,7 +32,7 @@
 
 logger = logging.getLogger("pcluster.pcluster")
 unsupported_regions = ["ap-northeast-3"]
-
+default_region = "us-east-1"
```
enrico-usai: global variables should be uppercase

cli/pcluster/utils.py
```diff
@@ -166,6 +166,14 @@ def get_instance_vcpus(region, instance_type):
     return vcpus
 
 
+def get_supported_os():
+    """
+    Return a tuple of the os supported by parallelcluster.
+    :return: a tuple of strings of the supported os
+    """
+    return "alinux", "centos6", "centos7", "ubuntu1404", "ubuntu1604"
```
enrico-usai: we can fixup this commit with the one on which you are using this function.
NOTE instead of a function we could have a global variable and use it directly.

cli/pcluster/easyconfig.py
```diff
@@ -29,6 +29,7 @@
 
 from . import cfnconfig
 from pcluster.utils import get_supported_schedulers
+from pcluster.utils import get_supported_os
```
enrico-usai: you can write: `from pcluster.utils import get_supported_os, get_supported_schedulers`

cli/pcluster/easyconfig.py
```diff
@@ -290,3 +303,14 @@ def _is_config_valid(args, config):
         if is_file_ok:
             return True
         return False
+
+
+def _is_aws_batch(scheduler):
```
enrico-usai: the name should be `_is_awsbatch`, in any case I think we can avoid this function since it is used in a single place.

cli/pcluster/easyconfig.py
```diff
@@ -290,3 +303,14 @@ def _is_config_valid(args, config):
...
+    """
+    Return true if the scheduler is awsbatch
+    :param scheduler: the scheduler to check
+    :return: true if the scheduler is awsbatch
+    """
+    if scheduler == "awsbatch":
+        return True
```
enrico-usai: you can write:
`return scheduler == "awsbatch"`
so I think we can avoid this simple function that just perform an equality test

cli/pcluster/easyconfig.py
```diff
@@ -160,73 +160,66 @@ def configure(args):  # noqa: C901 FIXME!!!
         config_file = os.path.expanduser(os.path.join("~", ".parallelcluster", "config"))
 
     config = configparser.ConfigParser()
-
     # Check if configuration file exists
     if os.path.isfile(config_file):
         config.read(config_file)
 
+    s_cluster = {}
```
enrico-usai: I'd use a more explicative prefix, in place of "s_"

cli/pcluster/easyconfig.py
```diff
@@ -305,6 +296,73 @@ def _is_config_valid(args, config):
         return False
 
 
+def get_parameter(config, section, parameter_name, default_value):
+    """
+    Prompt the user to ask question without validation
```
enrico-usai: The comment is wrong

cli/pcluster/easyconfig.py
```diff
@@ -305,6 +296,73 @@ def _is_config_valid(args, config):
         return False
 
 
+def get_parameter(config, section, parameter_name, default_value):
```
enrico-usai: what about `get_config_param`?

cli/pcluster/utils.py
enrico-usai: Instead we can have a `get_supported_os(scheduler)` function filtering the oses according to the selected scheduler

cli/pcluster/easyconfig.py
```diff
@@ -305,6 +296,73 @@ def _is_config_valid(args, config):
...
+        "Max Queue Size",
+        get_parameter(config, "cluster " + cluster_template, "max_queue_size", "10")
+    )
+
+    scheduler_dict["max_queue_size"] = max_queue_size
+    scheduler_dict["initial_queue_size"] = "1"
+    return scheduler_dict
+
+
+def aws_batch_handler(config, cluster_template):
```
enrico-usai: the name of the scheduler in our config is "awsbatch" so the function should be "awsbatch_*"

enrico-usai: We are modifying the `config` object so I'd rename the function as: `_init_awsbatch_params` or see the comment below for `_init_scaling_params(..., scheduler)`

cli/pcluster/easyconfig.py
```diff
@@ -305,6 +296,73 @@ def _is_config_valid(args, config):
...
+    """
+    Return a dictionary containing the values asked to the user for aws_batch
+    :param config the configuration parser
+    :param cluster_template the name of the cluster
+    :return: a dictionary with the updated values
+    """
+    batch_dict = {"base_os": "alinux", "desired_vcpus": "1"}
+
+    # We first remove unnecessary parameters from the past configurations
+    non_batch_parameters = "max_queue_size", "initial_queue_size", "maintain_initial_size"
```
enrico-usai: we can define two global variables:
```
SCALING_PARAMS = "max_queue_size", "initial_queue_size", "maintain_initial_size"
AWSBATCH_SCALING_PARAMS =  "max_vcpus", "desired_vcpus", "min_vcpus"
```

cli/pcluster/easyconfig.py
```diff
@@ -305,6 +296,73 @@ def _is_config_valid(args, config):
...
+    :param config the configuration parser
+    :param section the name of the section
+    :param parameter_name: the name of the parameter
+    :param default_value: the default string to ask the user
+    :return:
+    """
+    return config.get(section, parameter_name) if config.has_option(section, parameter_name) else default_value
+
+
+def general_scheduler_handler(config, cluster_template):
```
enrico-usai: I'd create a function `_init_scaling_params(..., scheduler)` that remove or add the correct parameters according to the scheduler

enrico-usai:
I'd change the title of the PR to be more explicative

cli/pcluster/easyconfig.py
```diff
@@ -244,11 +245,41 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    """
+    Validate the configuration of the pcluster configure.
+    :param args: the arguments passed with the command line
+    :param config: the configParser
+    :return True if the configuration is valid, false otherwise
+    """
+    # We create a temp_file to validate before overriding the original config
+    path = os.path.join(tempfile.gettempdir(), "temp_config")
+    temp_file = path
+    temp_args = args
```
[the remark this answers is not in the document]
matteofiordarancio: Actually there is a bug with that code because of the lack of defensive copy, which means that if the user has selected a path that would be overwritten

cli/pcluster/easyconfig.py
```diff
@@ -151,51 +163,57 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+
+    keys = list_keys(aws_region_name)
     # Query EC2 for available keys as options
     key_name = prompt(
         "Key Name",
-        config.get("cluster " + cluster_template, "key_name")
-        if config.has_option("cluster " + cluster_template, "key_name")
-        else None,
-        options=list_keys(aws_region_name),
+        get_config_parameter(config, "cluster " + cluster_template, "key_name", keys[0]),
```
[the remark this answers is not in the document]
matteofiordarancio: If no keys are present an exception will be thrown before the code reach this line of code

cli/pcluster/easyconfig.py
```diff
@@ -233,11 +250,101 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    :param config the configuration parser
+    :param section the name of the section
+    :param parameter_name: the name of the parameter
+    :param default_value: the default string to propose the user
+    :return:
+    """
+    return config.get(section, parameter_name) if config.has_option(section, parameter_name) else default_value
+
+
+def scheduler_handler(config, cluster_label, scheduler):
```
demartinofra: Dropping this comment here although it is more generic: I would decouple as much as possible the logic that prompts questions to the customers from the logic that maps the read values to the parallelcluster representation.

Also try to understand how easy it is, after the refactoring of the config parser and validator, to reuse the config definitions and default values from that logic. We would like to avoid as much as possible to have to modify multiple parts of the code when we change default values or allowed values for a given field.

cli/pcluster/easyconfig.py
```diff
@@ -82,21 +88,26 @@ def get_regions():
...
+def ec2_conn(aws_region_name):
+    region = _evaluate_aws_region(aws_region_name)
     ec2 = boto3.client("ec2", region_name=region)
     return ec2
 
 
 @handle_client_exception
 def list_keys(aws_region_name):
-    conn = ec2_conn(aws_region_name)
+    conn = ec2_conn(_evaluate_aws_region(aws_region_name))
```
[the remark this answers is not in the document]
matteofiordarancio: I removed it from list key so that also list_subnet and list_vpc will still be checked. We don't know if in the future we may change the order of this function and in this way we prevent a future bug.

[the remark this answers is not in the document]
demartinofra: 👍 if you think about the function responsibilities it seemed more appropriate for the ec2_conn func to receive an already resolved region value, but I'm fine with that

cli/pcluster/utils.py
```diff
@@ -164,3 +163,22 @@ def get_instance_vcpus(region, instance_type):
...
+
+
+def get_supported_os(scheduler):
+    """
+    Return a tuple of the os supported by parallelcluster for the specific scheduler.
+
+    :param scheduler: the scheduler for which we want to know the supported os
+    :return: a tuple of strings of the supported os
+    """
+    return "alinux" if scheduler == "awsbatch" else "alinux", "centos6", "centos7", "ubuntu1404", "ubuntu1604"
```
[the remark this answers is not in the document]
matteofiordarancio: I checked and there isn't another definition (unless I am wrong and don't know that)

I decided to use a tuple because it's immutable and the function calling supported_os and schedulers should not change them. 

Dunno about good python practice though, that's a java style way.

cli/pcluster/easyconfig.py
```diff
@@ -11,25 +11,42 @@
...
+DEFAULT_VALUES = {
+    "cluster_template": "default",
+    "aws_region_name": "us-east-1",
+    "scheduler": "sge",
+    "os": "alinux",
+    "max_queue_size": "10",
+    "master_instance_type": "t2.micro",
+    "compute_instance_type": "t2.micro",
+    "vpc_name": "public",
+    "initial_cpus": "1",
```
enrico-usai: I'd call `initial_size` because it could be  `initial_queue_size ` or  `desired_vcpus `

cli/pcluster/easyconfig.py
```diff
@@ -233,11 +315,75 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    """
+    scheduler_info = {}
+    if scheduler == "awsbatch":
+        scheduler_info["parameters_to_remove"] = (
+            "max_queue_size",
+            "initial_queue_size",
+            "maintain_initial_size",
+            "compute_instance_type",
+        )
+        scheduler_info["max_size_param"] = "max_vcpus"
```
enrico-usai: I'd call "max_cluster_size" or "max_size"

cli/pcluster/easyconfig.py
```diff
@@ -233,11 +315,75 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+        scheduler_info["initial_cpus"] = "desired_vcpus"
+    else:
+        scheduler_info["parameters_to_remove"] = "max_vcpus", "desired_vcpus", "min_vcpus", "compute_instance_type"
+        scheduler_info["max_size_param"] = "max_queue_size"
+        scheduler_info["initial_cpus"] = "initial_queue_size"
```
enrico-usai: I'd call "initial_cluster_size" or "desired_size" because it can be initial_queue_size or max_vcpus

cli/pcluster/easyconfig.py
```diff
@@ -233,11 +315,75 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+        scheduler_info["parameters_to_remove"] = (
+            "max_queue_size",
+            "initial_queue_size",
+            "maintain_initial_size",
+            "compute_instance_type",
+        )
+        scheduler_info["max_size"] = "max_vcpus"
+        scheduler_info["initial_size"] = "desired_vcpus"
+    else:
+        scheduler_info["parameters_to_remove"] = "max_vcpus", "desired_vcpus", "min_vcpus", "compute_instance_type"
```
enrico-usai: We shouldn't remove "compute_instance_type" because we are already replacing it

cli/pcluster/easyconfig.py
```diff
@@ -233,11 +315,75 @@ def configure(args):  # noqa: C901 FIXME!!!
...
+    :param scheduler the target scheduler
+    :return: a dictionary with containing the information
+    """
+    scheduler_info = {}
+    if scheduler == "awsbatch":
+        scheduler_info["parameters_to_remove"] = (
+            "max_queue_size",
+            "initial_queue_size",
+            "maintain_initial_size",
+            "compute_instance_type",
```
enrico-usai: We shouldn't remove "compute_instance_type" because we are already replacing it

cli/pcluster/easyconfig.py
```diff
@@ -143,77 +164,138 @@ def list_subnets(aws_region_name, vpc_id):
...
+        scheduler_info["max_size"]: max_queue_size,
+        scheduler_info["initial_size"]: DEFAULT_VALUES["initial_size"],
+    }
+    aliases_parameters = {"__name__": "aliases", "ssh": "ssh {CFN_USER}@{MASTER_IP} {ARGS}"}
+    vpc_parameters = {"__name__": vpc_label, "vpc_id": vpc_id, "master_subnet_id": master_subnet_id}
+    sections = [aws_parameters, cluster_parameters, vpc_parameters, global_parameters, aliases_parameters]
 
+    # We first remove unnecessary parameters from the past configurations
+    for par in scheduler_info["parameters_to_remove"]:
+        config.remove_option(cluster_label, par)
```
enrico-usai: we are removing compute_instance_type too.

PR #1081, opened 2019-05-30
Add enable_efa to dna.json

sean-smith:

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -2404,6 +2404,9 @@
                     "stack_name": {
                       "Ref": "AWS::StackName"
                     },
+                    "enable_efa": {
+                      "Ref": "EFA"
+                    },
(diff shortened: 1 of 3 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -3242,6 +3242,9 @@
                     "stack_name": {
                       "Ref": "AWS::StackName"
                     },
+                    "enable_efa": {
```
demartinofra: enable_efa is not set in the /tmp/dna.json for the master server which might make this conditional check in the cookbook to fail: `if node['cfncluster']['enable_efa'] == 'compute' && node['cfncluster']['cfn_node_type'] == 'ComputeFleet'`.

We need to either check for ComputeFleet before enable_efa or set the value for the master server too. I would probably go for the second option and keep the dna.json aligned for master and compute nodes.

sean-smith: Added

PR #1091, opened 2019-06-04
Extended pcluster configure to speed up the configuration

matteofiordarancio:
Extended the pcluster configure command to also show vpc and subnets name. Added also the option to select an option with numbers on the left.

### DIFF
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -29,7 +29,7 @@ import boto3
 import configparser
 from botocore.exceptions import BotoCoreError, ClientError
 
-from pcluster.utils import get_supported_os, get_supported_schedulers
+from pcluster.utils import NumberedList, get_supported_os, get_supported_schedulers
 
 from . import cfnconfig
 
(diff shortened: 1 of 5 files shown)


cli/pcluster/numbers_to_list.py
```diff
@@ -0,0 +1,47 @@
...
+                for j in range(1, len(items[i - 1])):
+                    self.param_to_items[items[i - 1][j]] = items[i - 1][0]
+                self.param_to_items[i] = items[i - 1][0]
+            else:
+                self.param_to_items[i] = items[i - 1]
+            self.numbered_items.append((i, items[i - 1]))
+
+    def get(self):
+        """
+        Return an ordered dict of the items and the extra-identifier with the numbers added.
```
enrico-usai: update the doc

cli/pcluster/numbers_to_list.py
```diff
@@ -0,0 +1,36 @@
...
+        self.numbered_items = collections.OrderedDict()
+        for i in range(1, len(items) + 1):
+            self.numbered_items[i] = items[i - 1]
+
+    def get(self):
+        """
+        Return an ordered dict of the items with the numbers added.
+        :return: the ordered dict
+        """
+        return copy.copy(self.numbered_items)
```
lukeseawalker: self.numbered_items.copy()

demartinofra: Also what is the point of this copy? If you want to have an immutable list be aware of the difference between shallow and deep copy: https://docs.python.org/3.7/library/copy.html

cli/pcluster/numbers_to_list.py
```diff
@@ -0,0 +1,57 @@
...
+            else:
+                self.param_to_items[number] = element
+            self.numbered_items.append((number, element))
+
+    def get(self):
+        """
+        Return an ordered list of the items and the extra-identifiers with the numbers added.
+
+        So for example if items was [["id1","synonym1"], ["id2","synonym2"]], the result list will return
+        ["1",[id1","synonym1"], ["2",[id2","synonym2"]]
```
lukeseawalker: the result is a list of tuple, not a list of list like in the example

cli/pcluster/numbers_to_list.py
```diff
@@ -0,0 +1,57 @@
+import copy
+
+
+class NumberedList:
+    """
+    This class adds number to a list of string and associate each one with one element of the list.
```
lukeseawalker: rephrase as we discussed

cli/pcluster/numbers_to_list.py
```diff
@@ -0,0 +1,57 @@
...
+        Return an ordered list of the items and the extra-identifiers with the numbers added.
+
+        So for example if items was [["id1","synonym1"], ["id2","synonym2"]], the result list will return
+        ["1",[id1","synonym1"], ["2",[id2","synonym2"]]
+
+        :return: the ordered dict
+        """
+        return copy.copy(self.numbered_items)
+
+    def get_item_by(self, key):
```
lukeseawalker: I don't like much that if you provide directly string that you are searching for, it will return a ValueError.

cli/pcluster/easyconfig.py
```diff
@@ -63,32 +64,69 @@ def wrapper(*args, **kwargs):
...
-            print("    %s" % o)
+        allowed_values = set([item[0] for item in options])
+        numbered_items = NumberedList(options)
+        for key, value in numbered_items.get().items():
+            print("{0}. {1}".format(key, " | ".join(value)))
+
+        result = get_correct_answer(allowed_values, default_value, message, numbered_items, user_prompt)
+    else:
+        user_input = input(user_prompt).strip()
+        result = user_input if user_input != "" else default_value
```
enrico-usai: you could write:
` result = user_input if user_input else default_value`
or
` result = user_input or default_value`

cli/pcluster/easyconfig.py
```diff
@@ -63,32 +64,69 @@ def wrapper(*args, **kwargs):
...
     else:
-        if check_validity and options is not None and var not in options:
-            print("ERROR: The value (%s) is not valid " % var)
-            print("Please select one of the Acceptable Values listed above.")
-            sys.exit(1)
-        else:
-            return var
+        # We give the user the possibility to try again if wrong
+        while True:
+            result = numbered_items.get_item_by_index(int(user_input))[0] if user_input.isdigit() else user_input
```
enrico-usai: we could define a `get_item_id_by_index`

cli/pcluster/easyconfig.py
```diff
@@ -115,52 +153,47 @@ def ec2_conn(aws_region_name):
...
 
+def extract_tag_from_resource(resource, tag_name):
+    tags = resource.get("Tags")
+    if tags:
+        for item in tags:
+            if item["Key"] == tag_name:
+                return item["Value"]
+
+
+def list_resource(resource, aws_region_name, resource_name, resource_id_name):
```
enrico-usai: It would be useful to have a comment for this function

cli/pcluster/easyconfig.py
```diff
@@ -252,23 +282,23 @@ def configure(args):  # noqa: C901 FIXME!!!
     keys = list_keys(aws_region_name)
     key_name = prompt(
         "Key Name",
-        get_config_parameter(config, section=cluster_label, parameter_name="key_name", default_value=keys[0]),
+        get_config_parameter(config, section=cluster_label, parameter_name="key_name", default_value=keys[0][0]),
```
enrico-usai: keys[0][0] is not clear, I think we can define a var for this.
e.g.
```py
default_key = list_keys(aws_region_name)[0]
default_key_id = default_key[0]
```

cli/pcluster/easyconfig.py
```diff
@@ -252,23 +279,23 @@ def configure(args):  # noqa: C901 FIXME!!!
...
-        check_validity=True,
     )
 
+    vpcs = list_vpcs(aws_region_name)
     vpc_id = prompt(
         "VPC ID",
-        get_config_parameter(config, section=vpc_label, parameter_name="vpc_id", default_value=None),
-        options=list_vpcs(aws_region_name),
-        check_validity=True,
+        get_config_parameter(config, section=vpc_label, parameter_name="vpc_id", default_value=None) or vpcs[0][0],
```
enrico-usai: Same as before --> vpcs[0][0] --> default_vpc_id

Then, could you pass default_vpcs_id as default_value to the get_config_parameter?

cli/pcluster/easyconfig.py
```diff
@@ -252,23 +279,23 @@ def configure(args):  # noqa: C901 FIXME!!!
...
     )
 
+    subnets = list_subnets(aws_region_name, vpc_id)
     master_subnet_id = prompt(
         "Master Subnet ID",
-        get_config_parameter(config, section=vpc_label, parameter_name="master_subnet_id", default_value=None),
-        options=list_subnets(aws_region_name, vpc_id),
-        check_validity=True,
+        get_config_parameter(config, section=vpc_label, parameter_name="master_subnet_id", default_value=None)
+        or subnets[0][0],
```
enrico-usai: Same as before

cli/pcluster/easyconfig.py
matteofiordarancio: I don't know, if you think about it NumbersUpdate knows nothing about the items he elaborates, only that he creates an ordered dict from it. Therefore, it's a job of the caller to handle the division id | name

cli/pcluster/easyconfig.py
matteofiordarancio: Should be a private function

cli/pcluster/easyconfig.py
matteofiordarancio: it's better 

`
    keys = _list_keys(aws_region_name)
    default_key = keys[0]
    default_key_id = default_key[0]`

To avoid calling _list_keys() two times (one here, one in prompt)

matteofiordarancio: Also note that in that example if you change vpc as a default you'll get the one of the previous subnet which is not present in that vpc. Fixed it

PR #1101, opened 2019-06-06
Added notificationARNs parameter to cluster config

MaxiIncluit:
Signed-off-by: Maximiliano Aguirre <maximiliano.aguirre@outlook.es>

Added SNSs parameter into config file for CloudFormation stack creation.


sean-smith:
What's the use case here?

MaxiIncluit:
> What's the use case here?

I would like to automate some event tracking of cloudformation stack, and there is an option of cloudformation to attach one or more SNSs to send events as SNS messages (which can trigger a lambda for example) but this option is currently not accessible from pcluster. This is a cloudformation stack configuration, that is why I included it in the config file.

cli/pcluster/cfnconfig.py
```diff
@@ -578,6 +581,17 @@ def __init_tags(self):
                     self.tags[key] = self.args.tags[key]
         except AttributeError:
             pass
+    
+    def __init_notificationARNs(self):
+        """
+        Get notification ARNs for cluster's event notification service (SNS)
+        """
+        self.notificationARNs = {}
```
jason-mccloskey: A dict is not a good default for notificationARNs as the CloudFormation API expects a list. In the case that notificationARNs is not included in the config, this will cause an error (as it did in my testing). To fix, please change to `self.notificationARNs = []`

sean-smith:
Hi @MaxiIncluit 

Looks good! just two things we need to change before we merge.

1. Please squash all the commits into one commit. Assuming you have `upstream` configured like this:

```
$ git remote -vv
upstream        git@github.com:aws/aws-parallelcluster-cookbook.git (fetch)
upstream        git@github.com:aws/aws-parallelcluster-cookbook.git (push)
```
Then you can rebase & squash:
```bash
git fetch upstream && git rebase upstream/develop
git rebase -i upstream/develop
```
Then change `pick` to `squash` for all but 1 commit, for example:

image

2. Next run `tox -e cfn-autoformat` to make travis happy:

```bash
pip install tox
cd cli && tox -e cfn-autoformat
```

Commit the changes and force push:

```bash
git commit -a --amend
git push [your_fork] [your_branch] --force
```

PR #1104, opened 2019-06-07
Allow re-use of Cluster in Tests

sean-smith:
* `--vpc-cluster` allows re-use of the vpc cluster between tests
* `--cluster` allows re-use of the cluster between tests
* `--no-delete` makes it easy to not delete created cluster, and re-use them

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -56,6 +56,11 @@ def pytest_addoption(parser):
     parser.addoption("--custom-awsbatchcli-package", help="url to a custom awsbatch cli package")
     parser.addoption("--custom-node-package", help="url to a custom node package")
     parser.addoption("--custom-ami", help="custom AMI to use in the tests")
+    parser.addoption("--vpc-stack", help="Name of an existing vpc stack.")
+    parser.addoption("--cluster", help="Use an existing cluster instead of creating one.")
+    parser.addoption(
+        "--no-delete", action="store_true", default=False, help="Don't delete stacks after tests are complete."
+    )
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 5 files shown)


tests/integration-tests/conftest.py
```diff
@@ -56,6 +56,9 @@ def pytest_addoption(parser):
     parser.addoption("--custom-awsbatchcli-package", help="url to a custom awsbatch cli package")
     parser.addoption("--custom-node-package", help="url to a custom node package")
     parser.addoption("--custom-ami", help="custom AMI to use in the tests")
+    parser.addoption("--vpc-cluster", help="Name of an existing vpc cluster.")
+    parser.addoption("--cluster", help="Use an existing cluster instead of creating one.")
+    parser.addoption("--no-delete", help="Don't delete stacks after tests are complete. Also enables --no-rollback.")
```
demartinofra: you are not enabling --no-rollback. We could also consider to enable --no-rollback always cause the stack will be deleted in any case by the tests if --no-delete is not specified

demartinofra: We can have --no-delete defined as a flag with `action="store_true", default=False,`

tests/integration-tests/test_runner.py
```diff
@@ -63,6 +63,9 @@
     "custom_awsbatch_template_url": None,
     "custom_awsbatchcli_url": None,
     "custom_ami": None,
+    "vpc-cluster": None,
```
demartinofra: let's keep the same convention used above with underscores. Same for no-delete

tests/integration-tests/test_runner.py
```diff
@@ -152,6 +155,18 @@ def _init_argparser():
...
+    parser.add_argument(
+        "--vpc-cluster", help="Name of an existing vpc cluster.", default=TEST_DEFAULTS.get("vpc-cluster")
+    )
+    parser.add_argument(
+        "--cluster", help="Use an existing cluster instead of creating one.", default=TEST_DEFAULTS.get("cluster")
+    )
+    parser.add_argument(
+        "--no-delete",
+        action="store_true",
+        help="Don't delete stacks after tests are complete. Also enables --no-rollback",
```
demartinofra: same here for enables no-rollback

tests/integration-tests/test_runner.py
```diff
@@ -223,6 +239,17 @@ def _set_custom_packages_args(args, pytest_args):
...
 
+def _set_custom_stack_args(args, pytest_args):
+    if args.vpc_cluster:
+        pytest_args.extend(["--vpc-cluster", args.vpc_cluster])
+
+    if args.cluster:
+        pytest_args.extend(["--cluster", args.cluster])
+
+    if args.no_delete:
+        pytest_args.extend(["--no-delete", True])
```
demartinofra: this will be `pytest_args.append("--no-delete")` if we go for the flag option

tests/integration-tests/conftest.py
sean-smith: I've gone ahead and made it default... I think this is the best option

sean-smith: changed

tests/integration-tests/test_runner.py
sean-smith: changed

tests/integration-tests/test_runner.py
sean-smith: changed

tests/integration-tests/test_runner.py
sean-smith: changed

PR #1124, opened 2019-06-11
Extended pcluster configure to speed up the configuration

matteofiordarancio:
Extended the pcluster configure command to also show vpc and subnet
name.
Added also the possibility to select an option with numbers on the left.

Continuation of: https://github.com/aws/aws-parallelcluster/pull/1091

Signed-off-by: Matteo Fiordarancio <mfiordar@amazon.com>

Before
image

After
image

### DIFF
--- a/cli/pcluster/easyconfig.py
+++ b/cli/pcluster/easyconfig.py
@@ -63,32 +63,38 @@ def handle_client_exception(func):
     return wrapper
 
 
-def prompt(message, default_value=None, hidden=False, options=None, check_validity=False):
-    if hidden and default_value is not None:
-        user_prompt = message + " [*******" + default_value[-4:] + "]: "
-    else:
-        user_prompt = message + " ["
-        if default_value is not None:
-            user_prompt = user_prompt + default_value + "]: "
+def prompt(message, validator=lambda x: True, input_to_option=lambda x: x, default_value=None, options_to_print=None):
+    """
+    Prompt the user a message with optionally some options.
+
+    :param message: the message to show to the user
+    :param validator: a function that predicates if the input is correct
+    :param input_to_option: a function that given the input transforms it in something else
+    :param default_value: the value to return as the default if the user doesn't insert anything
+    :param options_to_print: the options to print if necessary
+    :return: the value inserted by the user validated
+    """
+    if options_to_print:
+        print("Allowed values for {0}:".format(message))
+        for item in options_to_print:
+            print(item)
+    user_prompt = "{0} [{1}]: ".format(message, default_value or "")
+
+    valid_user_input = False
+    result = default_value
+    # We give the user the possibility to try again if wrong
+    while not valid_user_input:
+        user_input = input(user_prompt).strip()
+        if user_input == "":
+            result = default_value
+            valid_user_input = True
         else:
-            user_prompt = user_prompt + "]: "
-
-    if isinstance(options, (list, tuple)):
-        print("Acceptable Values for %s: " % message)
-        for o in options:
-            print("    %s" % o)
-
-    var = input(user_prompt).strip()
-
-    if var == "":
-        return default_value
-    else:
-        if check_validity and options is not None and var not in options:
-            print("ERROR: The value (%s) is not valid " % var)
-            print("Please select one of the Acceptable Values listed above.")
-            sys.exit(1)
-        else:
-            return var
+            result = input_to_option(user_input)
+            if validator(result):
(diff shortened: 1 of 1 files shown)


matteofiordarancio:
@enrico-usai already made the first review in the #1091, need the second one

cli/pcluster/utils.py
```diff
@@ -199,3 +200,41 @@ def get_supported_schedulers():
...
+        for index, item in enumerate(items, start=1):
+            self.numbered_items[index] = item
+
+    def get(self):
+        """
+        Return an ordered dict of the items with the numbers added.
+
+        :return: the ordered dict
+        """
+        return self.numbered_items.copy()
```
matteofiordarancio: Note that in this case a deepcopy is not needed

cli/pcluster/utils.py
```diff
@@ -199,3 +200,41 @@ def get_supported_schedulers():
     :return: a tuple of strings of the supported scheduler
     """
     return "sge", "torque", "slurm", "awsbatch"
+
+
+class NumberedList:
```
demartinofra: Not clear what's the real value added by this class over a simple list other than accessing values starting from the index 1

cli/pcluster/easyconfig.py
```diff
@@ -63,32 +63,37 @@ def wrapper(*args, **kwargs):
...
+    :param message: the message to show to the user
+    :param validator: a function that predicates if the input is correct
+    :param input_to_option: a function that given the input transforms it in something else
+    :param default_value: the value to return as the default if the user doesn't insert anything
+    :param options_to_print: the options to print if necessary
+    :return: the value inserted by the user validated
+    """
+    if options_to_print:
+        print("Allowed values for {0}:".format(message))
+        [print(item) for item in options_to_print]
```
demartinofra: List comprehensions are used for creating new lists from other iterables. Let's not abuse this construct when not needed

cli/pcluster/easyconfig.py
```diff
@@ -383,7 +373,86 @@ def scheduler_handler(scheduler):
...
+    """
+    Wrap prompt to use it for list.
+
+    :param message: the message to show the user
+    :param options: the list of item to show the user
+    :param default_value: the default value
+    :return: the validate value
+    """
+    if not options:
+        print("ERROR: No options found for {0}".format(message))
```
demartinofra: what if we don't find options because the user is not authorized to list such resource but he is allowed to use it? We should allow the user to input the text for that option.

cli/pcluster/easyconfig.py
```diff
@@ -383,7 +373,86 @@ def scheduler_handler(scheduler):
...
+    def input_to_parameter(to_transform):
+        try:
+            item = numbered_items.get_item_by_index(int(to_transform))
+        except ValueError:
+            item = to_transform
+        return item
+
+    return prompt(
+        message,
+        validator=lambda x: x in options,
```
demartinofra: same here. Users usually have restricted IAM roles that might allow them to use a resource but not list all. I think we should accept textual input or print an option such as `0. enter manually`

cli/pcluster/easyconfig.py
```diff
@@ -63,32 +63,37 @@ def wrapper(*args, **kwargs):
...
+        if user_input == "":
+            result = default_value
+            valid_user_input = True
         else:
-            return var
+            result = input_to_option(user_input)
+            if validator(result):
+                valid_user_input = True
+            else:
+                print("WARNING: {0} is not an acceptable value for {1}".format(user_input, message))
```
demartinofra: ERROR rather than WARNING?

matteofiordarancio:
> looks very good overall. my main concern regards the fact that now we require an enlarged set of IAM permissions to use the configure command. The user should be able to manually input parameters in case the aws call fails or in case the resource he intends to use is not in the list. Maybe we can print a warning in that case.

The set of IAM permission is the same as before. Moreover, If the user has to input parameter all alone, I kinda think it's the same as writing it alone. We can't list vpc or subnet if the we can't list keys, we can't even create a network configuration at that point.

cli/pcluster/utils.py
matteofiordarancio: The real values it that given a list, it adds a number to each element and allows to retrieve each one, knowing the number, in O(1). It also returns the list with the numbers as an ordered dict, making it easier to print for the user. It's not a very complicated class, that's for sure. I think that in this way is more organized

PR #1131, opened 2019-06-11
Ssh from public ips

sean-smith:
SSH without an elastic ip

Enables ssh to the master instance even when `use_public_ips = false`

This commit goes with https://github.com/aws/aws-parallelcluster/pull/1130

### DIFF
--- a/cli/pcluster/pcluster.py
+++ b/cli/pcluster/pcluster.py
@@ -717,11 +717,7 @@ def command(args, extra_args):  # noqa: C901 FIXME!!!
         elif status in valid_status:
             outputs = stack_result.get("Outputs")
             username = _get_output_value(outputs, "ClusterUser")
-            ip = (
-                _get_output_value(outputs, "MasterPublicIP")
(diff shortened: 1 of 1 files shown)


cli/pcluster/pcluster.py
```diff
@@ -691,6 +691,29 @@ def _get_param_value(params, key_name):
...
+    if ip:
+        return ip
+
+    # get master ip even if use_public_ips = false but vpc creates a public ip
+    ip = _get_master_server_ip(stack, config)
+    if ip:
+        return ip
+
+    # fall back to private ip for users that can resolve a private ip locally
+    return _get_output_value(outputs, "MasterPrivateIP")
```
demartinofra: I think `_get_master_server_ip` is already trying to get private IP and exiting if unsuccessful. 

This function could be reduced to:
```
return _get_output_value(outputs, "MasterPublicIP") or _get_master_server_ip(stack, config)
```

(probably we could drop the function completely)

sean-smith: yep, you're right

PR #1140, opened 2019-06-13
Add ec2:DescribeNetworkInterfaceAttribute to User policy

### LINKED ISSUE #1139: Additional permission needed for EFS sanity checks
**Environment:**
 - AWS ParallelCluster / CfnCluster version 2.4.0
 - OS: ubuntu
 - Scheduler: sge
 - Master instance type: r3.2xlarge
 - Compute instance type: r3.2xlarge

**Bug description and how to reproduce:**
Using the ParallelClusterUserPolicy permissions specified in docs, I get an error during sanity checking if an EFS volume is specified. 
`Config sanity error on resource EFSFSId: User is not authorized to perform that action on the specified resource`

In CloudTrail, it looks like this is due to a call to describeMountTargetSecurityGroups. I can recapitulate the error by calling that directly in awscli while using an instance role with ParallelClusterUserPolicy attached:
`aws efs describe-mount-target-security-groups --region us-east-1 --mount-target-id fsmt-abcd1234`
`An error occurred (AccessDeniedException) when calling the DescribeMountTargetSecurityGroups operation: User is not authorized to perform that action on the specified resource`

In the [AWS docs for that endpoint](https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargetSecurityGroups.html), it looks like there is an additional EC2 permission required (ec2:DescribeNetworkInterfaceAttribute) which is not specified in the parallelcluster IAM docs. I'm adding a pull request momentarily to include that permission.

tgjohnst:
Fixes #1139 

Cluster creation will fail during sanity checks (specifically the DescribeMountTargetSecurityGroups call) if an EFS volume is specified in config. This additional permission should allow that endpoint to be called without error. See https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargetSecurityGroups.html and #1139 for more details.

### DIFF
--- a/docs/iam.rst
+++ b/docs/iam.rst
@@ -241,6 +241,7 @@ To: ::
                   "ec2:DescribeAddresses",
                   "ec2:CreateTags",
                   "ec2:DescribeNetworkInterfaces",
+                  "ec2:DescribeNetworkInterfaceAttribute",
                   "ec2:DescribeAvailabilityZones"
               ],
               "Effect": "Allow",


tgjohnst:
This PR is no longer useful now that the docs are moving to a different repo. Fix is addressed here: https://github.com/awsdocs/aws-parallelcluster-user-guide/pull/1

PR #1166, opened 2019-07-01
Add new parameter --credential to integration tests runner

lukeseawalker:
Integration tests can be now launched in new optin region, specifying the --credential parameter for each new region to test.
Credential parameter is a comma separated list in the format region,endpoint,ARN,externalId.

### DIFF
--- a/tests/integration-tests/README.md
+++ b/tests/integration-tests/README.md
@@ -193,6 +193,27 @@ tests_outputs
             └── ...
 ```
 
+## Cross Account Integration Tests
+If you want to distribute integration tests across multiple accounts you can make use of the `--credential` flag. 
+This is useful to overcome restrictions related to account limits and be compliant with a multi-region, multi-account 
+setup.
+
+When the `--credential` flag is specified and STS assume-role call is made in order to fetch temporary credentials to 
+be used to run tests in a specific region. 
+
+The `--credential` flag is in the form `<region>,<endpoint_url>,<ARN>,<external_id>` and needs to be specified for each 
+region you want to use with an STS assumed role (that usually means for every region you want to have in a separate 
+account).
+
+ * `region` is the region you want to test with an assumed STS role (which is in the target account where you want to 
+ launch the integration tests)
+ * `endpoint_url` is the STS endpoint url of the main account to be called in order to assume the delegation role
+ * `ARN` is the ARN of the delegation role in the optin region account to be assumed by the main account
+ * `external_id` is the external ID of the delegation role  
+
+By default, the delegation role lifetime last for one hour. Mind that if you are planning to launch tests that last 
+more than one hour.
+
 ## Write Integration Tests
 
 All integration tests are defined in the `integration-tests/tests` directory.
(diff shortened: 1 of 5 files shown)


tests/integration-tests/utils.py
```diff
@@ -116,3 +119,72 @@ def delete_s3_bucket(bucket_name, region):
...
+            )
+
+            # Set credential for all cli command e.g. pcluster create
+            os.environ["AWS_ACCESS_KEY_ID"] = aws_credentials["AccessKeyId"]
+            os.environ["AWS_SECRET_ACCESS_KEY"] = aws_credentials["SecretAccessKey"]
+            os.environ["AWS_SESSION_TOKEN"] = aws_credentials["SessionToken"]
+
+
+def _retrieve_sts_credential(credentials, region):
+    credential_endpoint = credentials.get(region)[0]
```
demartinofra: minor you could do something like:
```
a, b, c = credentials.get(region)
```

also you could consider passing already the credentials for the specific region to the function

tests/integration-tests/README.md
```diff
@@ -193,6 +193,21 @@ tests_outputs
             └── ...
 ```
 
+## Cross Account Integration Tests
```
demartinofra: I would avoid to mention optin region also since this feature is more generic. we can rephrase with something like:

If you want to distribute integration tests across multiple accounts you can make use of the `—credential` flag. This is useful to overcome restrictions related to account limits and be compliant with a multi-region, multi-account setup.
When the `—credential` flag is specified and STS assume-role call is made in order to fetch temporary credentials to be used to run tests in a specific region. 

The `--credential` flag is in the form `<region>,<endpoint_url>,<ARN>,<external_id>` and needs to be specified for each region you want to use with an STS assumed role (that usually means for every region you want to have in a separate account).

* `region` is the region you want to test with an assumed STS role (which is in the target account where you want to launch the integration tests)
...

tests/integration-tests/conftest.py
```diff
@@ -35,13 +35,26 @@
...
 from vpc_builder import Gateways, SubnetConfig, VPCConfig, VPCTemplateBuilder
 
 
 def pytest_addoption(parser):
     """Register argparse-style options and ini-style config values, called once at the beginning of a test run."""
     parser.addoption("--regions", help="aws region where tests are executed", default=["us-east-1"], nargs="+")
+    parser.addoption(
+        "--credential",
+        help="STS credential endpoint, in the format <region>,<endpoint>,<ARN>,<externalId>. "
+        "Could be specified multiple times.",
```
demartinofra: not really true with nargs+. let's just remove this line

tests/integration-tests/conftest.py
```diff
@@ -323,6 +337,14 @@ def cfn_stacks_factory(request):
 }
 
 
+@pytest.fixture(scope="class", autouse=True)
+def env_setup(region, request):
```
demartinofra: setup_sts_credentials?

PR #1171, opened 2019-07-03
Created Integration Tests for the refactored pcluster configure

matteofiordarancio:
Created Integration Tests for the refactored pcluster configure. 

This pr integrates the changes made in pcluster configure into the integrations tests.

Now tests use network_template_builder instead of vpc_builder.
Created a test for pcluster configure 


tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
demartinofra: Let's move test functions at the top of the file and use the _ prefix for all local functions

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+
+TIMESTAMP = "-{:%Y%m%d%H%M%S}".format(datetime.datetime.utcnow())
+
+
+@pytest.fixture()
+def networking_stack_factory(request):
+    """Define a fixture to manage the creation and destruction of CloudFormation stacks."""
+    factory = CfnStacksFactory()
+    yield factory
+    if not request.config.getoption("no_delete"):
```
demartinofra: For the moment I wouldn't keep these stacks in case no-delete flag is used

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+    internet_gateway_id = get_internet_gateway_id(ec2_client, vpc_id)
+    stack = CfnStack(
+        name="integ-tests-networking-public" + TIMESTAMP,
+        region=region,
+        template=load_public_template(),
+        parameters=get_cfn_parameters(
+            availability_zone, internet_gateway_id=internet_gateway_id, vpc_id=vpc_id, public_cidr=public_subnet_cidr
+        ),
+    )
+    networking_stack_factory.create_stack(stack)
```
demartinofra: Can we abstract some of the common functionalities inside the networking_stack_factory and have something like:

```
def networking_stack_factory():
    factory = CfnStacksFactory()

    def _create_network(region, template_path, parameters):
         stack = CfnStack(...)
         factory.create_stack()
    yield _create_network
    factory.delete_all_stacks()
```

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def test_public_configuration_no_internet_gateway(region, vpc_stacks, networking_stack_factory):
+    ec2_client = ec2_conn(region)
+    vpc_id = vpc_stacks[region].cfn_outputs["VpcId"]
+    public_subnet_cidr = "10.0.3.0/24"
+    availability_zone = _AVAILABILITY_ZONE_OVERRIDES.get(region, "")
+    internet_gateway_id = get_internet_gateway_id(ec2_client, vpc_id)
+    stack = CfnStack(
+        name="integ-tests-networking-public" + TIMESTAMP,
```
demartinofra: I know TIMESTAMP might be better but for the moment let's use the same convention adopted in the rest of the tests `name="integ-tests-networking-" + random_alphanumeric()`

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+    ) as cfn_file:
+        file_content = cfn_file.read()
+    return file_content
+
+
+def test_public_configuration_no_internet_gateway(region, vpc_stacks, networking_stack_factory):
+    ec2_client = ec2_conn(region)
+    vpc_id = vpc_stacks[region].cfn_outputs["VpcId"]
+    public_subnet_cidr = "10.0.3.0/24"
+    availability_zone = _AVAILABILITY_ZONE_OVERRIDES.get(region, "")
```
demartinofra: please rename _AVAILABILITY_ZONE_OVERRIDES to AVAILABILITY_ZONE_OVERRIDES

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+    internet_gateway_id = get_internet_gateway_id(ec2_client, vpc_id)
```
demartinofra: I would add this code to CfnStack and read all the resources from the Stacks Resources whenever possible. This reduces boto3 calls and also requires less IAM policies

```
@property
    def cfn_resources(self):
        """
        Return the CloudFormation stack resources for the cluster.
        Resources are retrieved only once and then cached.
        """
        if not self.__cfn_resources:
            self.__cfn_resources = retrieve_cfn_resources(self.cfn_name, self.region)
        return self.__cfn_resources
```

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+        region, public_subnet_id, expected_autoassign_ip_value=True, expected_availability_zone=avaialibility_zone
+    )
+    check_subnet_property(
+        region, private_subnet_id, expected_autoassign_ip_value=False, expected_availability_zone=avaialibility_zone
+    )
+    check_nat_in_subnet(ec2_client, public_subnet_id)
+    check_nat_in_subnet_route(ec2_client, private_subnet_id)
+
+
+def check_internet_gateway_in_subnet_route(ec2_client, subnet_id, expected_internet_gateway_id=""):
```
demartinofra: minor: you could rename all check* to assert*

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+        {"ParameterKey": "InternetGatewayId", "ParameterValue": internet_gateway_id},
+        {"ParameterKey": "PublicCIDR", "ParameterValue": public_cidr},
+        {"ParameterKey": "VpcId", "ParameterValue": vpc_id},
+    ]
+    if private_cidr:
+        parameters.append({"ParameterKey": "PrivateCIDR", "ParameterValue": private_cidr})
+    return parameters
+
+
+def load_public_template():
```
demartinofra: load_public_template and load_public_private_template can become a single function such as:

```
def read_text(path):
    """Read the content of a file."""
    with path.open() as f:
        return f.read()
```

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+
+def load_public_private_template():
+    with open(
+        os.path.join("..", "..", "cloudformation", "networking_configuration", "public-private.cfn.json")
+    ) as cfn_file:
+        file_content = cfn_file.read()
+    return file_content
+
+
+def test_public_configuration_no_internet_gateway(region, vpc_stacks, networking_stack_factory):
```
demartinofra: minor: I would chose a slightly better name for the test such as:
test_public_network_topology/configuration

without mentioning the no_internet_gateway cause otherwise it seems the network does not have an IGW at all.

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+    public_subnet_id = stack.cfn_outputs["PublicSubnetId"]
+    check_subnet_cidr(ec2_client, public_subnet_id, expected_subnet_cidr=public_subnet_cidr)
+    check_internet_gateway_id(ec2_client, vpc_id, expected_internet_gateway_id=internet_gateway_id)
+    check_internet_gateway_in_subnet_route(ec2_client, public_subnet_id, internet_gateway_id)
+    check_subnet_property(
+        region, public_subnet_id, expected_autoassign_ip_value=False, expected_availability_zone=availability_zone
+    )
+
+
+def test_private_configuration_no_internet_gateway(region, vpc_stacks, networking_stack_factory):
```
demartinofra: same here for the name

tests/integration-tests/tests/networking/test_networking.py
demartinofra: minor you could have a fixture vpc_stack(vpc_stacks, region) that returns only the vpc_stack for the specific region under test

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def check_subnet_cidr(ec2_client, subnet_id, expected_subnet_cidr):
+    """Check that the given subnet has the same subnet cidr."""
+    response = ec2_client.describe_subnets(Filters=[{"Name": "subnet-id", "Values": [subnet_id]}])
+    subnet_cidr = response["Subnets"][0]["CidrBlock"]
+    assert_that(subnet_cidr).is_equal_to(expected_subnet_cidr)
+
+
+def check_internet_gateway_id(ec2_client, vpc_id, expected_internet_gateway_id=""):
```
demartinofra: minor: we could have None as default instead of "". Same everywhere else

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+def check_nat_in_subnet_route(ec2_client, subnet_id):
+    """Check that the route of the given subnet contains a Nat Gateway."""
+    response = ec2_client.describe_route_tables(Filters=[{"Name": "association.subnet-id", "Values": [subnet_id]}])
+    routes = response["RouteTables"][0]["Routes"]
+    assert_that(
+        "NatGatewayId" in next(route for route in routes if route["DestinationCidrBlock"] == "0.0.0.0/0")
+    ).is_true()
+
+
+def ec2_conn(region):
```
demartinofra: probably we can get rid of this since 

ec2_conn(region) is the same as boto3.client("ec2", region_name=region)

and the second one is more explicit

tests/integration-tests/tests/networking/test_networking.py
```diff
@@ -0,0 +1,188 @@
...
+    """Check that there is a nat in the given subnet."""
+    response = ec2_client.describe_nat_gateways(Filters=[{"Name": "subnet-id", "Values": [subnet_id]}])
+    assert_that(len(response["NatGateways"])).is_greater_than(0)
+
+
+def check_nat_in_subnet_route(ec2_client, subnet_id):
+    """Check that the route of the given subnet contains a Nat Gateway."""
+    response = ec2_client.describe_route_tables(Filters=[{"Name": "association.subnet-id", "Values": [subnet_id]}])
+    routes = response["RouteTables"][0]["Routes"]
+    assert_that(
```
demartinofra: minor:
```
assert_that(
        next(route for route in routes if route["DestinationCidrBlock"] == "0.0.0.0/0")
    ).contains("NatGatewayId")
```

tests/integration-tests/cfn_stacks_factory.py
```diff
@@ -20,10 +20,11 @@
 class CfnStack:
     """Identify a CloudFormation stack."""
 
-    def __init__(self, name, region, template):
+    def __init__(self, name, region, template, parameters=None):
         self.name = name
         self.region = region
         self.template = template
+        self.parameters = parameters
```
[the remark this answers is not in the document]
matteofiordarancio: Tested, it works like a flower

PR #1175, opened 2019-07-05
integ tests: increase test subnets IPs

demartinofra:


tests/integration-tests/conftest.py
```diff
@@ -352,15 +352,15 @@ def vpc_stacks(cfn_stacks_factory, request):
         availability_zone = random.choice(_AVAILABILITY_ZONE_OVERRIDES.get(region, [None]))
         public_subnet = SubnetConfig(
             name="PublicSubnet",
-            cidr="10.0.0.0/24",
+            cidr="10.0.0.0/22",  # 1,022 IPs
```
matteofiordarancio: I would change this to 10.0.124.0/22 so that the ip you allocate are nearby from 128.0 to 255.255. 

This will also prevent further bug in nested stack creation in vpc that will probably start from 10.0.0.0

PR #1181, opened 2019-07-08
Use regionalized STS endpoint

lukeseawalker:

### DIFF
--- a/cli/pcluster/config_sanity.py
+++ b/cli/pcluster/config_sanity.py
@@ -41,6 +41,11 @@ class ResourceValidator(object):
         self.aws_access_key_id = aws_access_key_id
         self.aws_secret_access_key = aws_secret_access_key
 
+    def __get_sts_endpoint(self):
+        return "https://sts.{0}.{1}".format(
+            self.region, "amazonaws.com.cn" if self.region.startswith("cn-") else "amazonaws.com"
+        )
+
     def __get_partition(self):
         if self.region.startswith("us-gov"):
             return "aws-us-gov"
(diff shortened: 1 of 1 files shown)


cli/pcluster/config_sanity.py
```diff
@@ -38,9 +38,15 @@ def __init__(self, region, aws_access_key_id, aws_secret_access_key):
...
         """
         self.region = region
+        self.sts_endpoint_url = "https://sts." + region + "." + self.__get_sts_domain()
         self.aws_access_key_id = aws_access_key_id
         self.aws_secret_access_key = aws_secret_access_key
 
+    def __get_sts_domain(self):
+        if self.region.startswith("cn-"):
+            return "amazonaws.com.cn"
+        return "amazonaws.com."
```
demartinofra: extra dot at the end

cli/pcluster/config_sanity.py
```diff
@@ -38,9 +38,15 @@ def __init__(self, region, aws_access_key_id, aws_secret_access_key):
         :param aws_secret_access_key: AWS secret access key
         """
         self.region = region
+        self.sts_endpoint_url = "https://sts." + region + "." + self.__get_sts_domain()
```
enrico-usai: we could define a `__get_sts_endpoint_url()` to create the whole url:
```py
self.sts_endpoint_url = self.__get_sts_endpoint_url()
```

and I'd use `format` rather than a list of append:
```py
sts_endpoint_url = "https://sts.{0}.{1}".format(self.region, "amazonaws.com.cn" if self.region.startswith("cn-") else "amazonaws.com")
```

cli/pcluster/config_sanity.py
```diff
@@ -291,6 +297,7 @@ def validate(self, resource_type, resource_value):  # noqa: C901 FIXME
                     boto3.client(
                         "sts",
                         region_name=self.region,
+                        endpoint_url=self.sts_endpoint_url,
```
enrico-usai: Then, if the sts_endpoint_url is only used here we can avoid to define an object attribute and just call here the `__get_sts_endpoint_url()` function

PR #1190, opened 2019-07-11
Need to add LocationConstraint to create bucket outside of IAD

stephenmsachs:
Below is the error message when this is omitted

stesachs@:~/git/aws-parallelcluster/util$ aws s3api create-bucket --bucket "stesachs-test" --region eu-west-1

An error occurred (IllegalLocationConstraintException) when calling the CreateBucket operation: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.
stesachs@:~/git/aws-parallelcluster/util$ aws s3api create-bucket --bucket "stesachs-test" --region eu-west-1 --create-bucket-configuration LocationConstraint=eu-west-1

*Issue #, if available:*

*Description of changes:*


demartinofra:
I might be wrong but from what I remember LocationContraint cannot be specified with us-east-1 region but only works with the other regions. We might need an if statement to handle this as done here: https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/utils.py#L52

PR #1198, opened 2019-07-15
Make FSx Substack depend on ComputeSecurityGroupIngress to keep FSx f…

jason-mccloskey:
I encountered a race condition when running ParallelCluster 2.4.0. The FSx Substack was created prior to the updating of the ComputeSecurityGroup rules to allow access within the security group. This resulted in the error:
The file system cannot be created because the default security group in the subnet provided or the provided security groups do not permit Lustre LNET network traffic on port 988 (Service: AmazonFSx; Status Code: 400; Error Code: InvalidNetworkSettings; Request ID: 00c909db-c895-4734-a9ad-636c7355a6f1)

We can ensure that this does not happen if the FSXSubstack resource depends on the ComputeSecurityGroupIngress resource. This pull request fixes that issue.

*Issue #, if available:*
N/A

*Description of changes:*
Add DependsOn Condition of ComputeSecurityGroupIngress to FSXSubstack.


sean-smith:
Thanks for the pr @jason-mccloskey !

There's a problem here:

The `ComputeSecurityGroupIngress` has a condition `CreateSecurityGroups`:

```json
"ComputeSecurityGroupIngress": {
      "Type": "AWS::EC2::SecurityGroupIngress",
      "Properties": {
        "IpProtocol": "-1",
        "FromPort": 0,
        "ToPort": 65535,
        "SourceSecurityGroupId": {
          "Ref": "ComputeSecurityGroup"
        },
        "GroupId": {
          "Ref": "ComputeSecurityGroup"
        }
      },
      "Condition": "CreateSecurityGroups"  # <---- This
    },
```

The `CreateSecurityGroups` is dependent on `VPCSecurityGroupId` which is a user provided parameter. 

```json
"CreateSecurityGroups": {
      "Fn::Equals": [
        {
          "Ref": "VPCSecurityGroupId"
        },
        "NONE"
      ]
    },
```

So if we merge this patch, users who create FSx filesystems with the `VPCSecurityGroupId` parameter will be stuck in an infinite loop.

I don't have a good solution for you, but open to ideas.

PR #1207, opened 2019-07-18
Intel MPI Integration Tests

sean-smith:

### DIFF
--- a/tests/integration-tests/tests/efa/test_efa.py
+++ b/tests/integration-tests/tests/efa/test_efa.py
@@ -41,7 +41,8 @@ def test_efa(region, scheduler, instance, os, pcluster_config_reader, clusters_f
 
     _test_efa_installed(scheduler_commands, remote_command_executor)
     _test_mpi(remote_command_executor, slots_per_instance, scheduler, os)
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -77,3 +78,21 @@ def _test_osu_benchmarks_openmpi(remote_command_executor, scheduler_commands, te
     output = remote_command_executor.run_remote_command("cat /shared/osu.out").stdout
     latency = re.search(r"0\s+(\d\d)\.", output).group(1)
     assert_that(int(latency)).is_less_than(20)
+
+
+def _test_osu_benchmarks_intelmpi(remote_command_executor, scheduler_commands, test_datadir, slots_per_instance):
```
demartinofra: since this code is exactly the same as the previous function can we create a third parametrized function (_run_osu_benchmarks) that accepts as parameter the mpi library to use and then _test_osu_benchmarks_openmpi and _test_osu_benchmarks_intelmpi will simply call that function?

sean-smith: Updated!

PR #1248, opened 2019-08-01
Use Placement Group in Dryrun

sean-smith:

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -360,7 +360,7 @@ class ParallelClusterConfig(object):
                     "VPC section [%s] used in [%s] section is not defined" % (vpc_section, self.__cluster_section)
                 )
 
-    def __check_account_capacity(self):
+    def __check_account_capacity(self):  # noqa: C901
         """Try to launch the requested number of instances to verify Account limits."""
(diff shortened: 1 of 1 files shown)


cli/pcluster/cfnconfig.py
```diff
@@ -381,14 +381,27 @@ def __check_account_capacity(self):
...
-                DryRun=True,
-            )
+            if self.parameters.get("PlacementGroup"):
+                ec2.run_instances(
+                    InstanceType=instance_type,
+                    MinCount=max_size,
+                    MaxCount=max_size,
+                    ImageId=test_ami_id,
+                    SubnetId=subnet_id,
+                    Placement={
```
enrico-usai: Is it possible to avoid the if else block? With something like:
```py
Placement={"GroupName": self.parameters.get("PlacementGroup")} if self.parameters.get("PlacementGroup") else {}
```

sean-smith: works brilliantly. Thanks @enrico-usai !

sean-smith:
### This config
```ini
[cluster test]
key_name = amzn
vpc_settings = public
placement_group = test
```
### Results in
```bash
$ pcluster create test -nr
Beginning cluster creation for cluster: test
ERROR: Cluster placement groups are not supported by the 't2.micro' instance type. Specify a supported instance type or change the placement group strategy, and try again.
```

PR #1259, opened 2019-08-07
python script to push cookbook files

ddeidda:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/util/upload-cookbook.py
+++ b/util/upload-cookbook.py
@@ -1,0 +1,257 @@
+#!/usr/bin/python
+#
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not
+# use this file except in compliance with the License. A copy of the License
+# is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+#
+#
+# Upload cookbook
+#
+# usage: ./upload-cookbook.py --regions "<region>[,<region>, ...]" --full_name "<cookbook full name without extension>" \
+# --partition <partition> \
+# [--unsupportedregions "<region>[, <region>, ...]"] [--dryrun] [--override] \
+# [--credential <region>,<endpoint>,<arn>,<role>]*
+
+import re
+from datetime import datetime
+
+import argparse
+import boto3
+from botocore.exceptions import ClientError
+
+_BACKUP_DIR = "backup"
+_COOKBOOKS_DIR = "cookbooks"
+_bck_date = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+_bck_error_array = set()
+_cp_error_array = set()
+_ls_error_array = set()
+_credentials = []
+_main_region = None
+
+
+def _get_all_aws_regions(region):
+    ec2 = boto3.client("ec2", region_name=region)
+    return set(sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions")))
+
+
+def _aws_s3_ls(s3, region, bucket_name, key):
+    out = s3.list_objects_v2(Bucket=bucket_name, Prefix=key)
+    if len(out.get("Contents", [])) > 0:
+        _ls_error_array.add(region)
+
+
+def _aws_s3_bck(s3, args, region, bucket_name, full_name):
+    if args.dryrun:
+        print(
+            "Not backing up {0} to bucket {1} override is {2}, dryrun is {3}".format(
+                full_name, bucket_name, args.override, args.dryrun
+            )
+        )
+    else:
+        try:
+            copy_source = {"Bucket": bucket_name, "Key": _COOKBOOKS_DIR + "/" + full_name}
+            s3.copy(copy_source, bucket_name, _BACKUP_DIR + "/" + full_name + _bck_date)
+        except ClientError as e:
+            print("Couldn't backup {0}".format(full_name))
+            if e.response["Error"]["Code"] == "NoSuchBucket":
+                print("Bucket is not present.")
+            _bck_error_array.add(region)
+
+
+def _aws_s3_cp(s3, args, region, bucket_name, folder, full_name):
+    if args.dryrun:
+        print(
+            "Not uploading {0} to bucket {1}, override is {2}, dryrun is {3}".format(
+                full_name, bucket_name, args.override, args.dryrun
+            )
+        )
+    else:
+        key = folder + "/" + full_name
+        try:
+            s3.upload_file(full_name, bucket_name, key, ExtraArgs={"ACL": "public-read"})
+
+            print("Successfully uploaded {0} to s3://{1}/{2}".format(full_name, bucket_name, key))
+        except ClientError as e:
+            print("Couldn't upload {0} to bucket s3://{1}/{2}".format(full_name, bucket_name, key))
+            _cp_error_array.add(region)
+            if e.response["Error"]["Code"] == "NoSuchBucket":
+                print("Bucket is not present.")
+
+            raise e
+
+
+def _create_s3_client(region):
+    reg_credentials = [c for c in _credentials if c[0] == region]
+
+    if reg_credentials:
+        credential = reg_credentials[0]
+        credential_region = credential[0]
+        credential_endpoint = credential[1]
+        credential_arn = credential[2]
+        credential_external_id = credential[3]
+
+        try:
+            sts = boto3.client("sts", region_name=_main_region, endpoint_url=credential_endpoint)
+
+            assumed_role_object = sts.assume_role(
+                RoleArn=credential_arn,
+                ExternalId=credential_external_id,
+                RoleSessionName=credential_region + "upload_cfn_templates_sts_session",
+            )
+            aws_credentials = assumed_role_object["Credentials"]
+            s3 = boto3.client(
+                "s3",
+                region_name=credential_region,
+                aws_access_key_id=aws_credentials.get("AccessKeyId"),
+                aws_secret_access_key=aws_credentials.get("SecretAccessKey"),
+                aws_session_token=aws_credentials.get("SessionToken"),
+            )
+
+        except ClientError as e:
+            print("Warning: non authorized in region '{0}', skipping".format(credential_region))
+            raise e
+    else:
+        s3 = boto3.client("s3")
+    return s3
+
+
+def _get_bucket_name(args, region):
+    return region + "-aws-parallelcluster" if not args.bucket else args.bucket
+
+
+def _parse_args():
+    global _credentials
+    global _main_region
+    parser = argparse.ArgumentParser(description="Uploads cookbook to S3")
+
+    parser.add_argument(
+        "--regions",
+        type=str,
+        help='Valid Regions, can include "all", or comma separated list of regions',
+        required=True,
+    )
+    parser.add_argument(
+        "--unsupportedregions", type=str, help="Unsupported regions, comma separated", default="", required=False
+    )
+    parser.add_argument(
+        "--override",
+        action="store_true",
+        help="If override is false, the file will not be pushed if it already exists in the bucket",
+        default=False,
+        required=False,
+    )
+    parser.add_argument(
+        "--bucket", type=str, help="Buckets to upload to, defaults to [region]-aws-parallelcluster", required=False
+    )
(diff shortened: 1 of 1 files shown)


util/upload-cookbook.py
```diff
@@ -0,0 +1,237 @@
...
+    if len(out.get("Contents", [])) > 0:
+        _ls_error_array.add(region)
+
+
+def put_object_to_s3(s3_client, bucket, folder, full_name):
+    key = folder + "/" + full_name
+    try:
+        s3_client.upload_file(full_name, bucket, key, ExtraArgs={"ACL": "public-read"})
+
+        print("Successfully uploaded %s to s3://%s/%s" % (full_name, bucket, key))
```
sean-smith: use `.format()` instead

util/upload-cookbook.py
```diff
@@ -0,0 +1,237 @@
...
+                print("Bucket is not present.")
+            _bck_error_array.add(region)
+
+
+def _aws_cp(s3, args, region, bucket_name, folder, full_name):
+    if args.dryrun:
+        print(
+            "Not uploading %s to bucket %s, override is %s, dryrun is %s"
+            % (full_name, bucket_name, args.override, args.dryrun)
+        )
```
sean-smith: you can add a return after this line and then you don't need the `else` indent

util/upload-cookbook.py
```diff
@@ -0,0 +1,237 @@
...
+            print("Couldn't upload %s to bucket s3://%s/%s" % (full_name, bucket_name, key))
+            _cp_error_array.add(region)
+            if e.response["Error"]["Code"] == "NoSuchBucket":
+                print("Bucket is not present.")
+
+            raise e
+
+
+def _create_s3_client(region):
+    reg_credentials = [c for c in credentials if c[0] == region]
```
sean-smith: where does the `credentials` object come from?

util/upload-cookbook.py
```diff
@@ -0,0 +1,237 @@
...
+        except ClientError:
+            print("Warning: non authorized in region '{0}', skipping".format(credential_region))
+            pass
+    else:
+        s3 = boto3.client("s3")
+    return s3
+
+
+def _get_bucket_name(args, region):
+    return region + "-aws-parallelcluster" if not args.bucket else args.bucket
```
sean-smith: use `.format()`

util/upload-cookbook.py
```diff
@@ -0,0 +1,237 @@
...
+            s3.copy(copy_source, bucket_name, _BACKUP_DIR + "/" + full_name + _bck_date)
+        except ClientError as e:
+            print("Couldn't backup %s" % full_name)
+            if e.response["Error"]["Code"] == "NoSuchBucket":
+                print("Bucket is not present.")
+            _bck_error_array.add(region)
+
+
+def _aws_cp(s3, args, region, bucket_name, folder, full_name):
+    if args.dryrun:
```
sean-smith: can you move this check to the main function? It's simpler that way.

ddeidda: My purpose was to make the dryrun option transparent to the main function, so to leave the business logic flow intact while preventing modifications if dryrun is set. Otherwise I'd have to fill the main code of "if" statements... even better could have been to use the "state" pattern, delegating the execution to a DryRunState vs a NormalState but seems a bit of overkill in this case. Or maybe you meant something different?

util/upload-cookbook.py
ddeidda: was created at the bottom of the file, moving it to top

util/upload-cookbook.py
ddeidda: In general I prefer having a single point of return to simplify debugging and code reading but np if we have different conventions

util/upload-cookbook.py
ddeidda: ok

util/upload-cookbook.py
ddeidda: ok

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
+import re
+from datetime import datetime
+
+import argparse
+import boto3
+from botocore.exceptions import ClientError
+
+_BACKUP_DIR = "backup"
```
enrico-usai: we can avoid this global variable since it is used in a single place

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+_COOKBOOKS_DIR = "cookbooks"
+_bck_date = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
```
enrico-usai: same as above: we can avoid this global variable

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+    if args.dryrun:
+        print(
+            "Not uploading {0} to bucket {1}, override is {2}, dryrun is {3}".format(
+                full_name, bucket_name, args.override, args.dryrun
+            )
+        )
+    else:
+        key = folder + "/" + full_name
+        try:
+            s3.upload_file(full_name, bucket_name, key, ExtraArgs={"ACL": "public-read"})
```
enrico-usai: this code seems the same of: put_object_to_s3

Could we put it in common?

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+
+        except ClientError:
+            print("Warning: non authorized in region '{0}', skipping".format(credential_region))
+            pass
+    else:
+        s3 = boto3.client("s3")
+    return s3
+
+
+def _get_bucket_name(args, region):
```
enrico-usai: I'd use bucket=Null as first parameter in place of args that is a more complex structure
```python
def _get_bucket_name(region, bucket=Null):

# then you can call it with:
_get_bucket_name(region, args.bucket)
```

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+    for _region in args.regions:
+        s3 = _create_s3_client(_region)
+        bucket1_name = _get_bucket_name(args, _region)
+
+        print("Listing cookbook for region ({0})".format(_region))
+        _aws_ls(s3, _region, bucket1_name, _COOKBOOKS_DIR + "/" + args.full_name + ".tgz")
+
+    if len(_ls_error_array) > 0 and not args.override:
+        print("We know the cookbook archives are already there, in this round we need to upload the .date files!")
+        print("Failed to push cookbook, already present for region " + " ".join(_ls_error_array))
```
enrico-usai: you can use format

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+import argparse
+import boto3
+from botocore.exceptions import ClientError
+
+_BACKUP_DIR = "backup"
+_COOKBOOKS_DIR = "cookbooks"
+_bck_date = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+_bck_error_array = set()
+_cp_error_array = set()
+_ls_error_array = set()
```
enrico-usai: I'd avoid these global sets and I'd move in the main function, since they are used only there.

The _aws_ls function could return the errors (the same is valid for bck and cp functions):
```python
def _aws_ls(s3, region, bucket_name, key):
    errors = set()
    out = s3.list_objects_v2(Bucket=bucket_name, Prefix=key)
    if len(out.get("Contents", [])) > 0:
        errors.add(region)

   return errors
```

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+    if len(_ls_error_array) > 0 and not args.override:
+        print("We know the cookbook archives are already there, in this round we need to upload the .date files!")
+        print("Failed to push cookbook, already present for region " + " ".join(_ls_error_array))
+        exit(1)
+    elif len(_ls_error_array) > 0 and args.override:
+        print("Some or all of the cookbook archives are already there but OVERRIDE=true")
+
+    for _region in args.regions:
+        s3 = _create_s3_client(_region)
+        bucket1_name = _get_bucket_name(args, _region)
```
enrico-usai: I'd avoid the number in the variable name, we could call it just "bucket_name"

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+        _aws_ls(s3, _region, bucket1_name, _COOKBOOKS_DIR + "/" + args.full_name + ".tgz")
+
+    if len(_ls_error_array) > 0 and not args.override:
+        print("We know the cookbook archives are already there, in this round we need to upload the .date files!")
+        print("Failed to push cookbook, already present for region " + " ".join(_ls_error_array))
+        exit(1)
+    elif len(_ls_error_array) > 0 and args.override:
+        print("Some or all of the cookbook archives are already there but OVERRIDE=true")
+
+    for _region in args.regions:
```
enrico-usai: we generally avoid to use underscore prefix in local variables

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+        _aws_cp(s3, args, _region, bucket1_name, _COOKBOOKS_DIR, args.full_name + ".tgz")
+        _aws_cp(s3, args, _region, bucket1_name, _COOKBOOKS_DIR, args.full_name + ".md5")
+
+        if not args.dryrun:
+            # Stores LastModified info into .tgz.date file and uploads it back to bucket
+            with (open(args.full_name + ".tgz.date", "w+")) as f:
+                response = s3.head_object(Bucket=bucket1_name, Key=_COOKBOOKS_DIR + "/" + args.full_name + ".tgz")
+                f.write(response.get("LastModified").strftime("%Y-%m-%d_%H-%M-%S"))
+
+            _aws_cp(s3, args, _region, bucket1_name, _COOKBOOKS_DIR, args.full_name + ".tgz.date")
```
enrico-usai: I'd rename _aws_cp to copy_to_s3 or something more explicit

util/upload-cookbook.py
```diff
@@ -0,0 +1,241 @@
...
+    if len(_bck_error_array) > 0:
+        print("Failed to backup cookbook for region ({0})".format(" ".join(_bck_error_array)))
+
+    if len(_cp_error_array) > 0:
+        print("Failed to push cookbook for region ({0})".format(" ".join(_cp_error_array)))
+        exit(1)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Uploads cookbook to S3")
```
enrico-usai: I'd move all these parsing in a _get_parser function, see for example:
https://github.com/aws/aws-parallelcluster/blob/v2.4.1/cli/awsbatch/awsbstat.py#L38

enrico-usai: Then I'd move the _get_parser into the main function

util/upload-cookbook.py
ddeidda: Sets are re-used inside loops, then a global error message is generated

util/upload-cookbook.py
ddeidda: In general I follow the rule of using explicit constants to simplify maintenance and for better readability

util/upload-cookbook.py
ddeidda: put_object_to_s3 was a copy/paste from elsewhere, not used here. Removing it

util/upload-cookbook.py
ddeidda: ok

util/upload-cookbook.py
ddeidda: sure, ok

util/upload-cookbook.py
ddeidda: ok

util/upload-cookbook.py
ddeidda: Good idea, renaming also other functions to mimic aws' command s3 syntax, like below:

- _aws_s3_cp
- _aws_s3_bck
- _aws_s3_ls

util/upload-cookbook.py
ddeidda: ok

util/upload-cookbook.py
```diff
@@ -0,0 +1,234 @@
...
+        _credentials = [
+            tuple(credential_tuple.strip().split(","))
+            for credential_tuple in args.credential
+            if credential_tuple.strip()
+        ]
+
+    if args.regions == "all":
+        args.regions = _get_all_aws_regions(_main_region)
+    else:
+        args.regions = re.split("\s+|,+", args.regions)
```
lukeseawalker: args.regions = [x.strip() for x in args.regions.split(",")]

util/upload-cookbook.py
```diff
@@ -0,0 +1,234 @@
...
+
+    args.unsupportedregions = re.split("\s+|,+", args.unsupportedregions)
```
lukeseawalker: same as above

util/upload-cookbook.py
ddeidda: OK

util/upload-cookbook.py
ddeidda: OK

PR #1268, opened 2019-08-13
Check for InvalidAccessKeyId in get_supported_features

sean-smith:
### Before

```bash
$ pcluster create efa-test
Beginning cluster creation for cluster: efa-test
Failed validate efa. This is probably a bug on our end. Please set sanity_check = false and retry
```

### After

```bash
$ pcluster create efa-test
Beginning cluster creation for cluster: efa-test
The AWS Access Key Id you provided does not exist in our records.
```

### DIFF
--- a/cli/pcluster/utils.py
+++ b/cli/pcluster/utils.py
@@ -166,9 +166,15 @@ def get_supported_features(region, feature):
     try:
         features = _get_json_from_s3(region, "features/feature_whitelist.json")
         supported_features = features.get("Features").get(feature)
-    except (ValueError, ClientError, KeyError):
+    except (ValueError, ClientError, KeyError) as e:
+        if type(e) is ClientError:
+            code = e.response.get("Error").get("Code")
+            if code == "InvalidAccessKeyId":
+                print(e.response.get("Error").get("Message"))
+                exit(1)
         print(
-            "Failed validate %s. This is probably a bug on our end. Please set sanity_check = false and retry" % feature
+            "Failed validate {0}. This is probably a bug on our end. Please submit an issue "
+            "https://github.com/aws/aws-parallelcluster/issues/new/choose".format(feature)
         )
         exit(1)
 


cli/pcluster/utils.py
```diff
@@ -166,11 +166,18 @@ def get_supported_features(region, feature):
     try:
         features = _get_json_from_s3(region, "features/feature_whitelist.json")
         supported_features = features.get("Features").get(feature)
-    except (ValueError, ClientError, KeyError):
-        print(
-            "Failed validate %s. This is probably a bug on our end. Please set sanity_check = false and retry" % feature
-        )
-        exit(1)
+    except (ValueError, ClientError, KeyError) as e:
+        code = e.response.get("Error").get("Code")
```
demartinofra: I don't think all the exceptions we are catching here have the Error.Code and Error.Message fields

cli/pcluster/utils.py
```diff
@@ -166,11 +166,18 @@ def get_supported_features(region, feature):
...
+        message = e.response.get("Error").get("Message")
+        if code == "InvalidAccessKeyId":
+            print(message)
+            exit(1)
+        else:
+            print(
+                "Failed validate %s. This is probably a bug on our end. Please set sanity_check = false and retry"
```
demartinofra: Shall we be more specific with something like: Failed to retrieve feature whitelisting? Also sanity_check to false doesn't always fix the problem

cli/pcluster/utils.py
sean-smith: Ok, I've switched this to check only when it's `ClientError`. For example, with invalid credentials this results in:

```bash
$ pcluster create efa-test
Beginning cluster creation for cluster: efa-test
The AWS Access Key Id you provided does not exist in our records.
```
When I change the file to a non-existant file, for example `features/feature_whitelist2.json`
```bash
$ pcluster create efa-test
Beginning cluster creation for cluster: efa-test
Failed validate efa. This is probably a bug on our end. Please submit an issue https://github.com/aws/aws-parallelcluster/issues/new/choose
```

cli/pcluster/utils.py
sean-smith: I've changed this too:

```bash
Failed validate efa. This is probably a bug on our end. Please submit an issue https://github.com/aws/aws-parallelcluster/issues/new/choose
```

And yes, you're right, turning off `sanity_check` only works for awsbatch.

cli/pcluster/utils.py
```diff
@@ -166,9 +166,16 @@ def get_supported_features(region, feature):
...
+    except (ValueError, ClientError, KeyError) as e:
+        if type(e) is ClientError:
+            code = e.response.get("Error").get("Code")
+            message = e.response.get("Error").get("Message")
+            if code == "InvalidAccessKeyId":
+                print(message)
+                exit(1)
         print(
-            "Failed validate %s. This is probably a bug on our end. Please set sanity_check = false and retry" % feature
+            "Failed validate %s. This is probably a bug on our end. Please submit an issue "
```
enrico-usai: You could use "format".

cli/pcluster/utils.py
```diff
@@ -166,9 +166,16 @@ def get_supported_features(region, feature):
     try:
         features = _get_json_from_s3(region, "features/feature_whitelist.json")
         supported_features = features.get("Features").get(feature)
-    except (ValueError, ClientError, KeyError):
+    except (ValueError, ClientError, KeyError) as e:
+        if type(e) is ClientError:
+            code = e.response.get("Error").get("Code")
+            message = e.response.get("Error").get("Message")
```
enrico-usai: you could avoid to define the message variable since we are using it only if `code == InvalidAccessKeyId`

PR #1278, opened 2019-08-23
Only Check for valid instance type if sanity_check = true

sean-smith:


cli/pcluster/cfnconfig.py
```diff
@@ -537,7 +537,8 @@ def __init_efa_parameters(self):
             supported_features = get_supported_features(self.region, "efa")
             valid_instances = supported_features.get("instances")
 
-            self.__validate_instance("EFA", self.parameters.get("ComputeInstanceType"), valid_instances)
+            if self.__sanity_check:
```
enrico-usai: It is useful to have a comment describing why we are skipping just the compute instance validation and not the others

enrico-usai: Thank you!

PR #1290, opened 2019-08-29
Changed amis.txt to yaml format.

joeydumont:
This is a very simple PR that changes the format of the amis.txt file to yaml. The text file was already almost formatted as a yaml file, and it makes parsing that file much easier.

I am working on a script that parses this file to get the latest AMI for a given OS and region, hence the motivation for the change.


sean-smith:
This is a great idea!

We publish this file automagically. Can you edit: https://github.com/aws/aws-parallelcluster/blob/develop/util/generate-ami-list.py#L140

Then going forward every release will produce this yaml file.

joeydumont:
I would like to test my changes to `generate-ami-list.py`, but cannot run the script on my machine. Could you provide an example JSON output file so I can test these functions.

I have noticed that a couple files will need to be updated to reflect the changes. I cannot do it right away as there is no tag associated with this PR. Here is the list of files:
  - `docs/tutorials/02_ami_customization.rst`
  - `util/get-ami-list.py`

I'm not sure which repo is the canonical repo for documentation, but `doc_source/cluster-definition.md` in aws-docs/aws-parallelcluser-user-guide` should probably be changed, as well as the 02_ami_customization file that's there.

joeydumont:
Closing as there doesn't seem to be a lot of interest in this, and I couldn't test the change. Parsing the text file is not that hard either.

Issue #1299, opened 2019-09-04
Why do template_url and additional_cfn_template have to be public objects?


joeydumont:
**Environment:**
 - aws-parallelcluster-2.4.1
 - OS: centos7
 - Scheduler: slurm
 - Master instance type: t2.micro
 - Compute instance type: t2.micro

It is only because pcluster uses `TemplateURL` instead of `TemplateBody` in the `create_stack` call [here](https://github.com/aws/aws-parallelcluster/blob/d2c9eb311a3b9156dedc4ba23d5e96015db8c8c4/cli/pcluster/pcluster.py#L121) and [here](https://github.com/aws/aws-parallelcluster/blob/a0eb85336483565805df73981f1f3d1d9ab61221/cloudformation/aws-parallelcluster.cfn.json#L3636)?

If so, I have a feature request. I would like to be able to specify these two fields as S3 URLs, with the bucket being accessible by the user running the pcluster CLI. From what I've seen, it should be fairly easy for `template_url`, as that stack is created directly with a `create_stack` call.

For `additonal_cfn_template`, I haven't seen any documentation about using a `TemplateBody` inside a `AWS::CloudFormation::Stack` clause, or how to create a nested stack from the CloudFormation CLI. It would be possible to create a completely separate stack, but that doesn't seem ideal to me.

If you have other ideas, I'd be glad to hear them. I'm also open to opening a PR myself for this work, if there is an agreed-upon solution for this.

sean-smith:
Hi @joeydumont 

Yep, you nailed it. You could technically use pre-signed urls that are not public but that's a huge pain. 

`template_url` is a simple feature, so I'd go ahead and create a PR for that. Looking at the docs I don't see a simple way to do `additonal_cfn_template`. Can you try with a s3 uri? I really doubt that'll work but I'm at a loss for a better way to do it.

joeydumont:
I tried with an S3 URI for both `template_url` and `addditional_cfn_template`, like this:
```ini
template_url = s3://<bucket_name>/pcluster_scripts/aws-parallelcluster-2.4.1.cfn.json
additional_cfn_template = s3://<bucket_name>/pcluster_scripts/additional_cfn_template.json
```
and both give the same error:
```bash
S3 error: Domain name specified in <bucket_name> is not a valid S3 domain
```
I will create a PR to load the `template_url` parameter and then use `TemplateBody` in the create_stack call, but, as you said, this cannot work for `additional_cfn_template`.

[...]

PR #1307, opened 2019-09-10
Added a CLI arg for additional_cfn_template.

joeydumont:
Related issue #1299

*Description of changes:*

I simply add a new command-line argumement `-a`, to set the AdditionalCfnTemplate attribute. This would allow the use of a dynamically generated pre-signed URL for both the TemplateURL and AdditionalCfnTemplate attributes without the need to manually edit the config file.

My next step would be to document this in the wiki. Since I can't edit the wiki, I was planning on proceeding like this [gist](https://gist.github.com/larrybotha/10650410) suggests.

I ran `tox -e autoformat` and `tox` on my machine and every test passed (for the python2.7 and python3.7, I don't have other versions of python installed).

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -69,6 +69,9 @@ class ParallelClusterConfig(object):
         # Initialize template url public attribute
         self.__init_template_url()
 
+        # Initialize additional_cfn_template url attribute.
+        self.__init_additional_cfn_template_url()
+
         # Validate VPC configuration settings and initialize corresponding parameters
         self.__init_vpc_parameters()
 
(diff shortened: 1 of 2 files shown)


joeydumont:
I just want to clarify that this is ready for review. If the PR is accepted, then I will document the workflow in the wiki, in case reviews change something.

demartinofra:
Hi Joey,

thanks for your contribution and sorry if this hasn't been reviewed yet. 

We are in the process of releasing a new version of ParallelCluster and also moreover we are re-designing the entire internal logic for pcluster config parsing and validation as part of https://github.com/aws/aws-parallelcluster/pull/1298. We'd like to include your changes once the above mentioned changes are completed if you are ok with that.

joeydumont:
Hi Francesco, 

No worries, I didn't want to press for review, just make clear that it was ready for it.

I can make sure that the changes are properly integrated within the new config parsing implementation once your changes are merged.

joeydumont:
I finally noticed that #1298 was merged. Is there still interest in adding that feature?

demartinofra:
Hi Joey,

sorry for the late reply. We discussed this internally and our conclusion was that for now we would like not to expose a very reduced subset of cluster configuration options through the pcluster CLI. The main motivations were the fact that we would like to keep consistency with our current model, where the cluster configuration is defined in the config file, and we want to avoid adding a feature to the CLI that does not apply to all configuration options.

I know that editing the pcluster configuration file is not the same as adding a flag to the cli, when building some automation around pcluster commands, but, at the same time, programmatically changing the configuration values, by using the native Python configparser module, shouldn't introduce much complexity. But please let me know if I'm missing something here.

In any case we are more than willing to take on a feature enhancement request to add to the pcluster cli a generic mechanism to allow overrides of configuration values.

joeydumont:
Ok, that makes sense if you want to go that route. On my end, I'll continue building automation on top of pcluster.

If you want, I can open an issue with that more general feature request and close this PR.

PR #1246, opened 2019-08-01
Fix broken sanity check for custom EC2 role

demartinofra:
https://github.com/aws/aws-parallelcluster/issues/1241

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1241, opened 2019-07-30
"ValueError: too many values to unpack" during 2.4.1 cluster creation


tgjohnst:
I just updated to 2.4.1 and am now also experiencing the same ValueError during cluster creation as was referenced in #1236

**Environment:**
 - AWS ParallelCluster version 2.4.1
 - OS: ubuntu1604
 - Scheduler: sge
 - Master instance type: r5.2xlarge
 - Compute instance type: r5.2xlarge

**Bug description and how to reproduce:**
Using the same template that worked fine in 2.4.0 (the only change being using 2.4.1 AMI), I now receive an error when running *pcluster create*
```
$ pcluster create rosalind
Beginning cluster creation for cluster: rosalind
Unexpected error of type ValueError: too many values to unpack (expected 2)
```

See cluster config attached:
[pcluster_config_RCE.txt](https://github.com/aws/aws-parallelcluster/files/3448734/pcluster_config_RCE.txt)

Any idea why 2.4.1 is now throwing this error?

demartinofra:
Hi,
thanks for reporting this. Unfortunately I was unable to reproduce the issue. Could you please install this version and rerun the failing command? [aws-parallelcluster-2.4.1.tar.gz](https://github.com/aws/aws-parallelcluster/files/3451736/aws-parallelcluster-2.4.1.tar.gz)

`pip install --upgrade aws-parallelcluster-2.4.1.tar.gz`

I simply removed the exception handling so that we should be able to see the stack trace on unexpected failures.

Thanks,
Francesco

tgjohnst:
Thanks Francesco, here's the stack trace:
```Beginning cluster creation for cluster: rosalind
Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/bin/pcluster", line 10, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/pcluster/cli.py", line 371, in main
    args.func(args)
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/pcluster/cli.py", line 29, in create
    pcluster.create(args)
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/pcluster/pcluster.py", line 77, in create
    config = cfnconfig.ParallelClusterConfig(args)
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/pcluster/cfnconfig.py", line 77, in __init__
    self.__init_cluster_parameters()
[...]

PR #1329, opened 2019-09-25
Fixed parsing issue when using ec2_iam_role option

stemors:
Fixed issue with parsing error when trying to use a custom IAM role with the ec2_iam_role option:

"Unexpected error of type ValueError: too many values to unpack (expected 2)"

The code is looking for two values for this option, but there should only be one (the name of the IAM role). This change has been tested in my environment, and has resolved my issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/pcluster/cfnconfig.py
+++ b/cli/pcluster/cfnconfig.py
@@ -515,7 +515,7 @@ class ParallelClusterConfig(object):
             master_root_volume_size=("MasterRootVolumeSize", None),
             compute_root_volume_size=("ComputeRootVolumeSize", None),
             base_os=("BaseOS", None),
-            ec2_iam_role=("EC2IAMRoleName", "EC2IAMRoleName"),
+            ec2_iam_role=("EC2IAMRoleName", None),
             extra_json=("ExtraJson", None),
             custom_chef_cookbook=("CustomChefCookbook", None),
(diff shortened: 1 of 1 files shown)


enrico-usai:
Hi @stemors , thank you for your efford.
Did you have the issue with the _develop_ branch or with the released version?

I'm asking this because this issue should be already solved with: https://github.com/aws/aws-parallelcluster/pull/1246

See source problem: https://github.com/aws/aws-parallelcluster/issues/1241

stemors:
My issue is with the released version 2.4.1 I looked, but had missed that bug fix you mentioned. That said, unless I'm missing something the code is still wrong, and should really be fixed? The fix in your link relates to the sanity checking.

enrico-usai:
Hi @stemors , 
with your patch you're removing the sanity check validation for the _EC2IAMRoleName_ resource but we actually need to validate it.

The real issue was in the sanity check code.
We fixed it with the # 1246, it will be available in the next release of ParallelCluster.

With that patch it is not more necessary to turn off the sanity check.

Thank you for your contribute.

stemors:
Right, had a look again and I follow. So my change has disabled the sanity checking for a custom IAM role. Whereas the bug was actually in the sanity check of the custom role in config_sanity.py.

My mistake, thanks for the clarification, you can close this.

Steve

PR #1340, opened 2019-10-02
Change Fsx Lustre Storage Defaults

sean-smith:


enrico-usai:
Hi @sean-smith the patch looks good for me. 
Could you describe in the comment why you changed the default? (E.g. fsx changed its defaults)
(I would have added in the commit message too, but it's ok)

PR #1345, opened 2019-10-04
Fixes to logging and pcluster configure validation

demartinofra:
The Python logging module organizes loggers in a hierarchy. We were configuring the logger for "pcluster.pcluster" and then use logging.getLogger(__name__) in some of the modules. This was causing messages to use the default root logger, hence the output was not being displayed. I uniformed this behaviour by configuring the logger on "pcluster" and then using it with logging.getLogger("pcluster." + __name__). As an additional future enhancement it would be nice to configure the logger by using a configuration file so that an expert user can customize its behaviour if necessary.

When validating the cluster configuration as last step of pcluster configure we were not using the region selected by the user, hence the validation was failing if the default aws region set by the user (for example with the AWS_DEFAULT_REGION env var) was different. 

I also added some additional output at the end of pcluster configure workflow to explicitly report the path of the generated config file and to suggest what command to use to create a cluster.

### DIFF
--- a/cli/pcluster/cli.py
+++ b/cli/pcluster/cli.py
@@ -23,7 +23,7 @@ from botocore.exceptions import NoCredentialsError
 from pcluster import pcluster
 from pcluster.configure import easyconfig
 
-LOGGER = logging.getLogger("pcluster.pcluster")
+LOGGER = logging.getLogger("pcluster." + __name__)
 
 
 def create(args):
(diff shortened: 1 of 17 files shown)


cli/pcluster/cli.py
```diff
@@ -23,7 +23,7 @@
 from pcluster import pcluster
 from pcluster.configure import easyconfig
 
-LOGGER = logging.getLogger("pcluster.pcluster")
+LOGGER = logging.getLogger("pcluster." + __name__)
```
sean-smith: Can you use `.format()`

demartinofra:
Discarding..will reopen a new one once it's rebased..

PR #1348, opened 2019-10-08
Add GPU info to instances.json/Validate instances.json

rexcsn:
- Added GPU attributes to instanes.json
- Added logic to validate instance-detail file is a valid JSON file
- Added logic to pull old instances.json and compare new instances.json with previous version to make sure no instance info is loss

### DIFF
--- a/util/upload-instance-slot-map.py
+++ b/util/upload-instance-slot-map.py
@@ -26,26 +26,82 @@ import boto3
 from botocore.exceptions import ClientError
 
 
-def dump_instances(instance_details):
-    with open(instance_details) as data_file:
-        data = json.load(data_file)
+def dump_instances(instance_details, region, aws_credentials=None):
+    """
+    Get instance information from pricing file and dump to instances.json
+    """
+    # Added exception block to check if pricing information is a valid JSON file
+    try:
+        with open(instance_details) as data_file:
+            data = json.load(data_file)
+    except json.JSONDecodeError as e:
+        print("Error when parsing pricing information: {0}".format(e.msg))
+        raise
+
     instances = {}
-    for sku, product in data.get("products").iteritems():
+    for sku, product in data.get("products").items():
         if "Compute Instance" in product.get("productFamily"):
             instance = product.get("attributes")
             instances[instance.get("instanceType")] = {"vcpus": instance.get("vcpu")}
+            # Adding instance's gpu information to instances.json
+            if "gpu" in instance:
+                instances[instance.get("instanceType")]["gpu"] = instance.get("gpu")
+
+    # Check if all info from old instances.json is included in the new instances.json
+    validate_against_old_version(instances, region, aws_credentials)
+
     print(json.dumps(instances))
     json.dump(instances, open("instances.json", "w"))
 
 
+def validate_against_old_version(instances_info, region, aws_credentials=None):
+    """
+    Validate that old instances information are all contained in the new version
+    """
+    old_instances_info = get_old_instances_info(region, aws_credentials)
+    for instance_type in old_instances_info:
+        if instance_type not in instances_info:
+            _fail("new file does not contain {0} instance type!".format(instance_type))
+        for resource in old_instances_info[instance_type]:
+            if resource not in instances_info[instance_type]:
+                _fail("new file does not contain {0} resource for {1} instance type!".format(resource, instance_type))
+
+
 def get_all_aws_regions(region):
+    """Return all AWS regions"""
     ec2 = boto3.client("ec2", region_name=region)
     return sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions"))
 
 
-def push_to_s3(region, aws_credentials=None):
-    bucket_name = region + "-aws-parallelcluster"
+def get_old_instances_info(region, aws_credentials=None):
+    """
+    Retrieve and parse old instances.json from aws-parallelcluster S3 bucket
+    """
+    bucket_name = "{0}-aws-parallelcluster".format(region)
     print(bucket_name)
+    s3 = return_s3_object(region, aws_credentials, bucket_name)
+    old_instances_file_content = s3.Object(bucket_name, "instances/instances.json").get()["Body"].read()
+    try:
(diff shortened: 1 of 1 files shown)


util/upload-instance-slot-map.py
```diff
@@ -26,14 +26,36 @@
...
+            # Adding instance's gpu information to instances.json
+            if "gpu" in instance:
+                instances[instance.get("instanceType")]["gpu"] = instance.get("gpu")
+    # print(json.dumps(instances))
+
+    # Check if all info from old instances.json is included in the new instances.json
+    old_instances_info = get_old_instances_info(region, aws_credentials)
+    for instance_type in old_instances_info:
+        if instance_type not in instances:
+            print("ERROR: new file does not contain %s instance type!" % instance_type)
```
sean-smith: Create `__fail()` function instead of `print`, `exit(1)`

util/upload-instance-slot-map.py
```diff
@@ -43,9 +65,25 @@ def get_all_aws_regions(region):
     return sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions"))
 
 
+def get_old_instances_info(region, aws_credentials=None):
+    bucket_name = region + "-aws-parallelcluster"
```
sean-smith: use `.format()`

util/upload-instance-slot-map.py
```diff
@@ -26,14 +26,36 @@
...
         if "Compute Instance" in product.get("productFamily"):
             instance = product.get("attributes")
             instances[instance.get("instanceType")] = {"vcpus": instance.get("vcpu")}
+            # Adding instance's gpu information to instances.json
+            if "gpu" in instance:
+                instances[instance.get("instanceType")]["gpu"] = instance.get("gpu")
+    # print(json.dumps(instances))
+
+    # Check if all info from old instances.json is included in the new instances.json
+    old_instances_info = get_old_instances_info(region, aws_credentials)
```
sean-smith: Add try/catch blocks

util/upload-instance-slot-map.py
```diff
@@ -43,9 +65,25 @@ def get_all_aws_regions(region):
...
 
+def get_old_instances_info(region, aws_credentials=None):
+    bucket_name = region + "-aws-parallelcluster"
+    print(bucket_name)
+    s3 = return_s3_object(region, aws_credentials, bucket_name)
+    old_instances_file_content = s3.Object(bucket_name, "instances/instances.json").get()["Body"].read()
+    return json.loads(old_instances_file_content)
+
+
 def push_to_s3(region, aws_credentials=None):
```
sean-smith: rename function... doesn't look like it pushes to S3

rexcsn: Does push to s3 with `bucket.upload_file()`

util/upload-instance-slot-map.py
```diff
@@ -26,26 +26,74 @@
...
+                )
+                __fail(msg)
+
+
 def get_all_aws_regions(region):
     ec2 = boto3.client("ec2", region_name=region)
     return sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions"))
 
 
+def get_old_instances_info(region, aws_credentials=None):
```
sean-smith: can you add docstrings to each new function? Tox should complain...

util/upload-instance-slot-map.py
```diff
@@ -26,26 +26,74 @@
...
+
     print(json.dumps(instances))
     json.dump(instances, open("instances.json", "w"))
 
 
+def validate_against_old_version(instances_info, region, aws_credentials=None):
+    old_instances_info = get_old_instances_info(region, aws_credentials)
+    for instance_type in old_instances_info:
+        if instance_type not in instances_info:
+            msg = "ERROR: new file does not contain {0} instance type!".format(instance_type)
```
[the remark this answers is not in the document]
sean-smith: wait it's already there, so this does `ERROR: ERROR: ...`

util/upload-instance-slot-map.py
```diff
@@ -99,6 +152,12 @@ def upload(regions, main_region, credentials):
                     pass
 
 
+def __fail(msg):
```
enrico-usai: minor: 
We are using the convention to have a single underscore as prefix of module's private functions.
We use double underscore for private class methods.

util/upload-instance-slot-map.py
```diff
@@ -99,6 +152,12 @@ def upload(regions, main_region, credentials):
...
 
+def _fail(msg):
+    """Print an error message and exit."""
+    print("ERROR: {0}".format(msg))
+    sys.exit(1)
+
+
 if __name__ == "__main__":
     # parse inputs
     parser = argparse.ArgumentParser(description="Upload instance slot map")
```
demartinofra: we should add a `--dryrun` flag to perform all steps but skipping the upload to s3. Or even better a `--deploy` flag so that by default it is disabled. 

Also it would be nice to always print the old instances.json (that is probably printed already) and the new one so that in case of errors, it's easier to debug.

util/upload-instance-slot-map.py
```diff
@@ -26,26 +26,82 @@
...
+                instances[instance.get("instanceType")]["gpu"] = instance.get("gpu")
+
+    # Check if all info from old instances.json is included in the new instances.json
+    validate_against_old_version(instances, region, aws_credentials)
+
     print(json.dumps(instances))
     json.dump(instances, open("instances.json", "w"))
 
 
+def validate_against_old_version(instances_info, region, aws_credentials=None):
```
demartinofra: what about comparing with something as simple as?
```
dict(new, **old) == new
```

This should also verifies the value linked to every nested key

rexcsn: The comparison you suggested is definitely simpler and works just as well, but I would argue that the double for loop helps with debugging because we would know right away which instance type of resource is missing.

demartinofra: agreed but at the same time we are not checking the values associated to resources keys and this test needs to be updated if we modify the file. 

printing the two files (even better if sorted) should be a good way to debug failures

PR #1350, opened 2019-10-08
Add Options to FSx Lustre tests

sean-smith:

### DIFF
--- a/tests/integration-tests/tests/storage/test_fsx_lustre.py
+++ b/tests/integration-tests/tests/storage/test_fsx_lustre.py
@@ -24,7 +24,7 @@ from tests.common.schedulers_common import SgeCommands
 @pytest.mark.oss(["centos7", "alinux"])
 @pytest.mark.schedulers(["sge"])
 @pytest.mark.usefixtures("os", "instance", "scheduler")
-def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_factory, test_datadir):
+def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_factory, test_datadir, os):
     """
     Test all FSx Lustre related features.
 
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -38,21 +38,26 @@ def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_
...
 
 
-def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir):
+def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os):
     logging.info("Testing fsx lustre is correctly mounted")
     result = remote_command_executor.run_remote_command("df -h -t lustre | tail -n +2 | awk '{print $1, $2, $6}'")
     assert_that(result.stdout).matches(r"[0-9\.]+@tcp:/fsx\s+3\.4T\s+{mount_dir}".format(mount_dir=mount_dir))
 
     result = remote_command_executor.run_remote_command("cat /etc/fstab")
+    if os == "centos7":
```
demartinofra: can we move this mapping into a dict so that it's easy to extend support for other OSs? Also mount_options might not be initialized if we run this on a non supported os. We probably want to handle that case by defaulting to "" and asserting that mount_options is not empty (just so we have a clear error message in case tests are ran against non supported os)

sean-smith: done and done

tilne:
might be bad form to give an approval when the CI check is failing, but seeing as how it's a "line is too long" error i'll do so anyways.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -38,21 +38,27 @@ def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_
...
+        "centos7": "defaults,_netdev,flock,user_xattr,noatime,noauto,x-systemd.automount,"
+        "x-systemd.requires=lnet.service",
+        "alinux": "defaults,_netdev,flock,user_xattr,noatime",
+    }
+
     assert_that(result.stdout).matches(
-        r"fs-[0-9a-z]+\.fsx\.[a-z1-9\-]+\.amazonaws\.com@tcp:/fsx {mount_dir} lustre defaults,_netdev 0 0".format(
-            mount_dir=mount_dir
+        r"fs-[0-9a-z]+\.fsx\.[a-z1-9\-]+\.amazonaws\.com@tcp:/fsx {mount_dir} lustre {mount_options} 0 0".format(
+            mount_dir=mount_dir, mount_options=mount_options[os] if os in mount_options else ""
```
enrico-usai: minor: the alternative was to use `mount_options.get(os, "")`

PR #1354, opened 2019-10-10
Default to optimal compute_instance_type when using batch

tilne:
Update the revamped configuration logic so that it uses a default value of optimal for the ComputeInstanceType/compute_instance_type parameter when the scheduler is awsbatch.

Modify existing tests.


demartinofra:
minor issues detected by flake8 linter. You can run the linters with tox -e code-linters

tilne:
Sorry about that. I'll be sure to run all of the checks locally first on future PRs.

cli/pcluster/config/param_types.py
```diff
@@ -519,6 +519,38 @@ def to_cfn(self):
...
+        Use different default value when the scheduler is awsbatch.
+
+        If the scheduler is awsbatch and the config file does not specify a
+        value for this parameter, use "optimal" instead of the usual default.
+        """
+        section_name = _get_file_section_name(self.section_key, self.section_label)
+        if config_parser.has_option(section_name, self.key):
+            self.value = config_parser.get(section_name, self.key).strip()
+        elif (
+            config_parser.has_option(section_name, "scheduler")
```
sean-smith: Use `get_param_value()` which handles the option not-existing scenario. So this becomes just:

```python
elif self.get_param_value("scheduler") == "awsbatch":
      self.value = "optimal"
```

tilne: The reason I opted for this is because when the `compute_instance_type` parameter is parsed before the `scheduler` parameter, the value returned by `get_param_value` would be the default value. Sean suggested adding a method to the `Param` class for this type of check, given that other default values for parameters within the same section is a potentially common scenario.

tilne:
The failures are intermittent, so it's taking a bit to track them down.

cli/pcluster/config/param_types.py
```diff
@@ -181,7 +181,15 @@ def to_cfn(self):
 
     def get_default_value(self):
         """Get default value from the Param definition if there, None otherwise."""
-        return self.definition.get("default", None)
+        default = self.definition.get("default", None)
+        try:
```
enrico-usai: I really like the [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-eafp) approach, but probably in this case we could use the [LBYL](https://docs.python.org/3/glossary.html?highlight=eafp#term-lbyl) to test if the default is function.
I say this because most of the time we have a value, not a function, so we can avoid to call the `get_section` and speed up the flow.

PR #1367, opened 2019-10-18
Create cloudwatch log group via cfn template.

tilne:
Add cloudwatch log group to the cfn template.

Note that in addition to adding the new resource, I also marked it as a dependency for the MasterServer and ComputeFleet resources. The motivation for this is to ensure that those resources are deleted before the log group is deleted. If the log group is deleted first, what happens is that the EC2 instances that outlive it continue to send log events for a short while. These log events cause a log group to be created with an infinite retention time. In this scenario a user has to delete this log group manually, otherwise the next time a log group of the same name is attempted to be created as part of the creation of a cluster it will fail.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -1126,6 +1126,32 @@
       },
       "Condition": "CreateEFSSubstack"
     },
+    "CloudWatchLogGroup": {
+      "Type": "AWS::Logs::LogGroup",
+      "Properties": {
+        "LogGroupName": {
+          "Fn::Sub": [
+            "/aws/parallelcluster/${cluster_name}",
+            {
+              "cluster_name": {
+                "Fn::Select": [
+                  "1",
+                  {
+                    "Fn::Split": [
+                      "parallelcluster-",
+                      {
+                        "Ref": "AWS::StackName"
+                      }
+                    ]
+                  }
+                ]
+              }
+            }
+          ]
+        },
+        "RetentionInDays": 14
+      }
+    },
     "SQS": {
       "Type": "AWS::SQS::Queue",
       "Properties": {
@@ -1618,6 +1644,9 @@
     },
     "MasterServer": {
       "Type": "AWS::EC2::Instance",
+      "DependsOn": [
+        "CloudWatchLogGroup"
+      ],
       "Properties": {
         "LaunchTemplate": {
           "LaunchTemplateId": {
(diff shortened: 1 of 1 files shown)


tilne:
cfn-lint and cfn-format-check both pass on my laptop. not sure what's going on with that failure but i'll look again tomorrow.

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1126,6 +1126,22 @@
...
     },
+    "CloudWatchLogGroup": {
+      "Type": "AWS::Logs::LogGroup",
+      "Properties": {
+          "LogGroupName": {
+            "Fn::Sub": [
+              "/aws/${cluster_name}",
+              {
+                "cluster_name": {
+                  "Ref": "AWS::StackName"
```
sean-smith: This is `parallelcluster-[cluster_name]` not the cluster name. You can Fn::Split to get just the cluster name. Search the code for examples.

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1126,6 +1126,22 @@
...
+                }
+              }
+            ]
+          },
+          "RetentionInDays": 14
```
sean-smith: Make into a parameter, or use ssm

tilne: Agreed that needs to happen eventually. My intent was to merge this version into the feature branch first, and then do that as part of the PR that modifies the user-facing configuration. Thoughts?

cloudformation/aws-parallelcluster.cfn.json
tilne: That was on purpose. The intent was to have something in the log group's name signifying that it was a cluster resource.

sean-smith: I'd prefer `/aws/parallelcluster/[cluster_name]` since it follows similar services, such as:
* `/aws/batch/job`
* `/aws/lambda/[lambda_name]`
* `/aws/apigateway/[api_name]`
* `/aws/codebuild/[project_name]` 

So it seems like it's the standard format. I know it's much messier to implement in the cfn template.

cloudformation/aws-parallelcluster.cfn.json
sean-smith: sounds good

cloudformation/aws-parallelcluster.cfn.json
tilne: should be good to go now

PR #1372, opened 2019-10-22
Disable Hyperthreading Integration Test Fixes

sean-smith:

### DIFF
--- a/tests/integration-tests/tests/disable_hyperthreading/test_disable_hyperthreading.py
+++ b/tests/integration-tests/tests/disable_hyperthreading/test_disable_hyperthreading.py
@@ -23,6 +23,7 @@ from tests.common.utils import fetch_instance_slots
 # t2's do not support CpuOptions and hence do not support disable_hyperthreading
 @pytest.mark.regions(["us-east-1"])
 @pytest.mark.skip_instances(["t2.micro"])
+@pytest.mark.skip_schedulers(["awsbatch"])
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/disable_hyperthreading/test_disable_hyperthreading.py
```diff
@@ -61,7 +62,9 @@ def _test_disable_hyperthreading(remote_command_executor, scheduler_commands, sl
     assert_that(int(result)).is_equal_to(slots_per_instance // 2)
 
     # check scale up to 2 nodes
-    result = scheduler_commands.submit_command("hostname > /shared/hostname.out", slots=slots_per_instance)
+    nodes = 1 if scheduler == "sge" else 2
```
enrico-usai: Could you add a comment describing why for SGE we have a different behaviour?

demartinofra: You could also simply use slots for all schedulers and set slots to 2*slots_per_instance as done in test_efa.py

sean-smith: This doesn’t work, torque and slurm won’t scale up. Jobwatcher thinks it can’t satisfy the requirements and doesn’t scale. The scheduler would probably schedule the job, however the instance never scales up.

PR #1373, opened 2019-10-23
Translate cluster to cfncluster

sean-smith:
* Internally we use cfncluster, but the new format is cluster

For example:

```ini
extra_json = { "cluster": {"cfn_scheduler_slots": "cores"}}
```
Becomes
```json
{ "cfncluster": {"cfn_scheduler_slots": "cores"}}
```
before it's passed to the cookbook.

Regression introduced with: #1298


cli/tests/pcluster/config/test_param_to_cfn.py
```diff
@@ -30,6 +30,13 @@
         (SCALING, "scaledown_idletime", 10, "10"),
         (SCALING, "scaledown_idletime", 10, "10"),
         (SCALING, "scaledown_idletime", 3, "3"),
+        (
+            CLUSTER,
+            "extra_json",
+            {"cluster": {"cfn_scheduler_slots": "cores"}},
+            '{"cfncluster": {"cfn_scheduler_slots": "cores"}}',
+        ),
+        (CLUSTER, "extra_json", {"cluster": {"enable_ganglia": "true"}}, '{"cfncluster": {"enable_ganglia": "true"}}'),
```
enrico-usai: We could add another test by using "cfncluster" too as starting point.

cli/pcluster/config/param_types.py
```diff
@@ -397,6 +397,9 @@ def from_file(self, config_parser):
         if config_parser.has_option(section_name, self.key):
             config_value = config_parser.get(section_name, self.key)
             self.value = self.get_value_from_string(config_value)
+            if self.key == "extra_json" and "cfncluster" in self.value:
```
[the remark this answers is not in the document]
sean-smith: Ok, I've created another class `ExtraJsonParam` and overriden the methods `to_cfn` and `to_file`

cli/tests/pcluster/config/test_param_to_cfn.py
sean-smith: done

cli/tests/pcluster/config/test_param_from_cfn.py
```diff
@@ -46,6 +46,12 @@
             '{ "cluster" : { "cfn_scheduler_slots" : "cores" } }',
             {"cluster": {"cfn_scheduler_slots": "cores"}},
         ),
+        (
+            CLUSTER,
+            "extra_json",
+            '{ "cfncluster" : { "enable_ganglia" : "true" } }',
+            {"cfncluster": {"enable_ganglia": "true"}},
+        ),
```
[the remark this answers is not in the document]
sean-smith: done

cli/pcluster/config/param_types.py
```diff
@@ -429,6 +429,80 @@ def get_cfn_value(self):
...
+
+        if cfn_converter:
+            if self.key == "extra_json" and "cluster" in self.value:
+                # support parallelcluster syntax by replacing the key
+                self.value = {"cfncluster": self.value.pop("cluster")}
+            cfn_params[cfn_converter] = str(self.get_string_value())
+
+        return cfn_params
+
+    def to_file(self, config_parser):
```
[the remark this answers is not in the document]
sean-smith: updated

cli/pcluster/config/param_types.py
```diff
@@ -434,6 +434,26 @@ def get_cfn_value(self):
...
+        """Convert param to CFN representation, if "cfn_param_mapping" attribute is present in the Param definition."""
+        if "cluster" in self.value:
+            self.value = {"cfncluster": self.value.pop("cluster")}
+        return super(ExtraJsonParam, self).to_cfn()
+
+    def to_file(self, config_parser):
+        """Set parameter in the config_parser in the right section.
+
+        The extra_json configuration parameter can contain both "cfncluster" or "cluster" keys but
+        we are writing "cluster" in the file to suggest the u
```
enrico-usai: You are missing the latest part of the comment: `... suggest the users the recommended syntax.`

cli/pcluster/config/param_types.py
```diff
@@ -434,6 +434,26 @@ def get_cfn_value(self):
 # that require some custom action during CFN/file conversion
 
 
+class ExtraJsonParam(JsonParam):
+    """Class to manage extra_json configuration parameters."""
+
+    def to_cfn(self):
+        """Convert param to CFN representation, if "cfn_param_mapping" attribute is present in the Param definition."""
```
enrico-usai: could you please update the comment to be more specific?
```python
        """
        Convert param to CFN representation.

        The extra_json configuration parameter can contain both "cfncluster" or "cluster" keys but cookbook 
        recipes require "cfncluster" as key.
        """
```

cli/pcluster/config/param_types.py
```diff
@@ -434,6 +434,26 @@ def get_cfn_value(self):
...
+        if "cluster" in self.value:
```
enrico-usai: We need to check the `self.value` because `NoneType` is not iterable (sorry I missed it in my previous comment).
```python
if self.value and "cluster" in self.value:
```

sean-smith: changed

cli/pcluster/config/param_types.py
sean-smith: updated

cli/pcluster/config/param_types.py
sean-smith: updated

cli/pcluster/config/param_types.py
```diff
@@ -434,6 +434,31 @@ def get_cfn_value(self):
...
+    """Class to manage extra_json configuration parameters."""
+
+    def to_cfn(self):
+        """
+        Convert param to CFN representation.
+
+        The extra_json configuration parameter can contain both "cfncluster" or "cluster" keys but cookbook
+        recipes require "cfncluster" as key.
+        """
+        if self.value and "cluster" in self.value:
```
enrico-usai: I looked back to the code and I think the best way to replace the `self.value` during CFN conversion is by overriding the `get_cfn_value` and not the `to_cfn` method.

This means that we can avoid to override the `to_cfn` and simply override the:
```python
    def get_cfn_value(self):
        """
        Convert parameter value into CFN value.

        The extra_json configuration parameter can contain both "cfncluster" or "cluster" keys but cookbook
        recipes require "cfncluster" as key.
       """
       self.value = {"cfncluster": self.value.pop("cluster")}
       return self.get_string_value()
```

then it will be automatically called by the `to_cfn` of the `Param` class (see: https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/config/param_types.py#L185).

I think it is the best way because I don't like that in the `get_cfn_value` we are emitting the parameter not sanitized, and by overriding this method we'll have the right sanitization in both `to_cfn` and `get_cfn_value` outputs.
The unit test below must be updated accordingly to it, see my following comment.

cli/tests/pcluster/config/test_param_to_cfn.py
```diff
@@ -30,6 +30,18 @@
         (SCALING, "scaledown_idletime", 10, "10"),
         (SCALING, "scaledown_idletime", 10, "10"),
         (SCALING, "scaledown_idletime", 3, "3"),
+        (
+            CLUSTER,
+            "extra_json",
+            {"cluster": {"cfn_scheduler_slots": "cores"}},
+            '{"cluster": {"cfn_scheduler_slots": "cores"}}',
```
enrico-usai: here we should have "cfncluster" since it is the result of the `get_cfn_value`. See my previous comment.

PR #1374, opened 2019-10-23
Sanitize test name

enrico-usai:
The `request.node.nodeid` contains the test name, composed by using the
parameters passed to the test. 

Some parameters can contain the "/" character that breaks the tests on Windows.

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -204,12 +204,19 @@ def clusters_factory(request):
 
 def _write_cluster_config_to_outdir(request, cluster_config):
     out_dir = request.config.getoption("output_dir")
+
+    # Sanitize config file name to make it Windows compatible
+    # request.node.nodeid example:
+    # 'dcv/test_dcv.py::test_dcv_configuration[eu-west-1-c5.xlarge-centos7-sge-8443-0.0.0.0/0-/shared]'
+    test_file, test_name = request.node.nodeid.split("::", 1)
+    config_file_name = "{0}-{1}".format(test_file, test_name.replace("/", "_"))
+
     os.makedirs(
-        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(request.node.nodeid)),
+        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(test_file)),
         exist_ok=True,
     )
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest.py
```diff
@@ -204,13 +204,17 @@ def _cluster_factory(cluster_config):
 
 def _write_cluster_config_to_outdir(request, cluster_config):
     out_dir = request.config.getoption("output_dir")
+
+    # sanitize test name
+    test = request.node.nodeid.split("::", 1)
+    test_file = test[0]
+    test_name = test[1].replace("/", "-")
```
demartinofra: shall we do a replace on the os.path.sep? Also I'm wondering if we should simply escape this or use a different separator since `-` is already used by pytest to separate test parameters values

tests/integration-tests/conftest.py
```diff
@@ -204,13 +204,17 @@ def _cluster_factory(cluster_config):
...
+
     os.makedirs(
-        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(request.node.nodeid)),
+        "{out_dir}/clusters_configs/{test_dir}".format(out_dir=out_dir, test_dir=os.path.dirname(test_file)),
         exist_ok=True,
     )
-    cluster_config_dst = "{out_dir}/clusters_configs/{test_name}.config".format(
-        out_dir=out_dir, test_name=request.node.nodeid.replace("::", "-")
-    )
+    cluster_config_dst = "{out_dir}/clusters_configs/{test_name}.config".format(out_dir=out_dir, test_name=test_name)
```
demartinofra: are you sure test_name is enough?

tests/integration-tests/conftest.py
enrico-usai: The "/" is not the os.path.sep in Windows, but I'm facing an error like:

```
FileNotFoundError: [Errno 2] No such file or directory: 'tests_outputs/1571825679.8591163.out/clusters_configs/dcv/test_dcv.py-test_dcv_configuration[eu-west-1-c5.xlarge-centos7-sge-5678-192.168.1.1/32-/myshared].config'
```

enrico-usai: I changed the separator to "_"

tests/integration-tests/conftest.py
enrico-usai: Fixed by concatenating test_file and test_name sanitized.

PR #1381, opened 2019-10-24
Modifications for instances.json

rexcsn:
- Added memory info to instances.json
- Added deploy flag to upload-instance-slot-map.py
- Added more print statements for better debugging

*Issue #, if available:*

*Description of changes:*


util/upload-instance-slot-map.py
```diff
@@ -152,8 +159,10 @@ def upload(regions, main_region, credentials):
                     pass
 
 
-def _fail(msg):
-    """Print an error message and exit."""
+def _fail(msg, old_instances, new_instances):
+    """Print old and new instances.json and error message then exit."""
+    print("Old instances.json:\n{0}".format(old_instances))
```
demartinofra: I'd print these files always so that when we do a dryrun we can compare the two

util/upload-instance-slot-map.py
```diff
@@ -61,10 +62,16 @@ def validate_against_old_version(instances_info, region, aws_credentials=None):
     old_instances_info = get_old_instances_info(region, aws_credentials)
     for instance_type in old_instances_info:
```
demartinofra: is this validation logic also comparing the values of the resources. e.g. by mistake vcpus for an instance type are changed during an update? If not we might consider using the approach suggested in the previous PR

util/upload-instance-slot-map.py
```diff
@@ -43,6 +43,7 @@ def dump_instances(instance_details, region, aws_credentials=None):
         if "Compute Instance" in product.get("productFamily"):
             instance = product.get("attributes")
             instances[instance.get("instanceType")] = {"vcpus": instance.get("vcpu")}
+            instances[instance.get("instanceType")]["memory"] = instance.get("memory")
```
demartinofra: minor: can you report somewhere in this file an extract from the initial doc used to read this data? What is the format for memory? is it in Gb and is it in the format we need in the daemons?

util/upload-instance-slot-map.py
rexcsn: Added value comparison

util/upload-instance-slot-map.py
rexcsn: Added comment

PR #1384, opened 2019-10-25
Drop dependency on requests in AWSBatch CustomResources

demartinofra:
requests library has been removed from the botocore.vendored libs: https://aws.amazon.com/blogs/developer/removing-the-vendored-version-of-requests-from-botocore/

We are using that library in two different parts of our code, both related to the code executed by AWS Lambda resources used by the batch substack:
- crhelper.py: a script to simplify the implementation of CustomResources
- send_build_notification.py: the lambda that sends the notification to CloudFormation when the build of Docker images is completed.

I imported the newer version of the crhelper from the aws-cloudformation repo, by copying the files from https://github.com/aws-cloudformation/custom-resource-helper/tree/master/crhelper
This newer version still presents a dependency on requests library. To drop this I updated one of the imported scripts to use python native libraries to perform the HTTP call. I also submitted a PR to apply this change upstream: https://github.com/aws-cloudformation/custom-resource-helper/pull/26

As last change I updated our send_build_notification.py to drop the dependency on the requests library.


cli/pcluster/resources/batch/custom_resources_code/crhelper/utils.py
```diff
@@ -20,9 +22,14 @@ def _send_response(response_url, response_body, put=requests.put):
     logger.debug("CFN response URL: {}".format(response_url))
     logger.debug(json_response_body)
     headers = {'content-type': '', 'content-length': str(len(json_response_body))}
+    split_url = urlsplit(response_url)
+    host = split_url.netloc
+    url = urlunsplit(("", "", *split_url[2:]))
```
lukeseawalker: please add comment

PR #1386, opened 2019-10-25
Enable cloudwatch logging config

tilne:
* Add CWLog cfn parameter
* Use CWLog cfn parameter to conditionally create the log group with the
correct retention time for log events
* Implement new cw_log section in the config file
* Create new tests for the cw_log_section
* Add tests to the existing cluster section
* Fix tests broken by the addition of CloudWatchAgentServerPolicy to
default EC2IAMPolicies

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -18,15 +18,8 @@ import re
 from configparser import NoSectionError
 
 import yaml
-from pcluster.utils import (
-    error,
-    get_avail_zone,
-    get_cfn_param,
-    get_efs_mount_target_id,
-    get_instance_vcpus,
-    get_partition,
-    warn,
-)
+from pcluster.config.iam_policy_rules import AWSBatchFullAccessInclusionRule, CloudWatchAgentServerPolicyInclusionRule
+from pcluster.utils import error, get_avail_zone, get_cfn_param, get_efs_mount_target_id, get_instance_vcpus, warn
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


cli/tests/pcluster/config/defaults.py
```diff
@@ -123,10 +123,13 @@
     "efs_settings": None,
     "raid_settings": None,
     "fsx_settings": None,
+    "cw_log_settings": None,
 }
 
 DEFAULT_PCLUSTER_DICT = {"cluster": DEFAULT_CLUSTER_DICT}
 
+DEFAULT_CW_LOG_DICT = {"enable": True, "retention_days": 14}
```
enrico-usai: minor: move before PCLUSTER_DICT

cli/tests/pcluster/config/defaults.py
```diff
@@ -142,12 +145,13 @@ class DefaultDict(Enum):
     raid = DEFAULT_RAID_DICT
     fsx = DEFAULT_FSX_DICT
     pcluster = DEFAULT_PCLUSTER_DICT
+    cw_log = DEFAULT_CW_LOG_DICT
```
enrico-usai: minor: move before PCLUSTER_DICT

cli/tests/pcluster/config/test_section_efs.py
```diff
@@ -173,7 +173,11 @@ def test_efs_section_to_cfn(mocker, section_dict, expected_cfn_params):
             utils.merge_dicts(
                 DefaultCfnParams["cluster"].value,
                 DefaultCfnParams["efs"].value,
-                {"MasterSubnetId": "subnet-12345678", "AvailabilityZone": "mocked_avail_zone"},
+                {
+                    "MasterSubnetId": "subnet-12345678",
+                    "AvailabilityZone": "mocked_avail_zone",
+                    "EC2IAMPolicies": "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy",
```
enrico-usai: You could add the `CloudWatchAgentServerPolicy` to the `DefaultCfnParams["cluster"]` since by default it is included in the EC2IAMPolicies

cli/pcluster/config/param_types.py
```diff
@@ -648,25 +648,89 @@ def to_cfn(self):
         return cfn_params
 
 
+class EC2IAMPolicyInclusionRule(object):
```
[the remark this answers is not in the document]
tilne: I think you're right that it's a good idea to move the classes to another module. I moved them to pcluster.config.iam_policy_rules.py. Let me know if you think they should go somewhere else.

The reason I did not make this an abstract base class is because the class is only used via class methods. So even if I were to use that, the "abstract" class method could still be invoked because there is never an attempt to create an instance of the class:
```
import abc

class TestABC(abc.ABC):
    @classmethod
    @abc.abstractmethod
    def someabstractmethod(cls):
        print('This is an abstract method')
```
```
>>> import test_abc
>>> test_abc.TestABC.someabstractmethod()
This is an abstract method
```

In some of the examples I've seen, they prevent this by having the base class's classmethod return an instance of the class:
```
import abc

class TestABC(abc.ABC):
    @classmethod
    @abc.abstractmethod
    def someabstractmethod(cls):
        return cls()
```
```
>>> import test_abc
>>> test_abc.TestABC.someabstractmethod()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tilne/repos/aws-parallelcluster-cookbook/util/test_abc.py", line 7, in someabstractmethod
    return cls()
TypeError: Can't instantiate abstract class TestABC with abstract methods someabstractmethod
```

I can make the change, but the process for ensuring that classes don't inherit the interface without implementing it is still 'manual'.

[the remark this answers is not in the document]
enrico-usai: You right, thank you for the detailed explanation.
I agree we can avoid this "manual" check.

PR #1389, opened 2019-10-28
Define get_stack utility function and pylint fixes

enrico-usai:
## cli.py
C0103: Variable name "ch" doesn't conform to snake_case naming style (invalid-name)
C0103: Variable name "fh" doesn't conform to snake_case naming style (invalid-name)

## utils.py
W0107: Unnecessary pass statement (unnecessary-pass)
C0103: Variable name "s3" doesn't conform to snake_case naming style (invalid-name)
C0123: Using type() instead of isinstance() for a typecheck. (unidiomatic-typecheck)

## param_types.py
R1705: Unnecessary "else" after "return" (no-else-return)
W0107: Unnecessary pass statement (unnecessary-pass)
R0916: Too many boolean expressions in if statement (7/5) (too-many-boolean-expressions)
E1003: Bad first argument 'CommaSeparatedParam' given to super() (bad-super-call)
W0711: Exception to catch is the result of a binary "or" operation (binary-op-exception)
R1719: The if expression can be replaced with 'bool(test)' (simplifiable-if-expression)

## pcluster_config.py
C0103: Variable name "cf" doesn't conform to snake_case naming style (invalid-name)


cli/pcluster/utils.py
```diff
@@ -235,18 +231,34 @@ def get_stack_output_value(stack_outputs, output_key):
...
+
+    :param stack_name: the CFN Stack name
+    :param cfn_client: boto3 cloudformation client
+    :return: the Stack data type
+    """
+    try:
+        if not cfn_client:
+            cfn_client = boto3.client("cloudformation")
+        return cfn_client.describe_stacks(StackName=stack_name).get("Stacks")[0]
+    except ClientError as e:
```
sean-smith: What if the stack doesn't exist? Will this throw a `ClientError` or an `IndexError` which we don't handle.

cli/pcluster/commands.py
```diff
@@ -133,12 +133,12 @@ def create(args):  # noqa: C901 FIXME!!!
...
         if not args.nowait:
-            utils.verify_stack_creation(cfn, stack_name)
+            utils.verify_stack_creation(stack_name, cfn)
             LOGGER.info("")
-            result_stack = cfn.describe_stacks(StackName=stack_name).get("Stacks")[0]
+            result_stack = utils.get_stack(stack_name, cfn)
             _print_stack_outputs(result_stack)
         else:
-            stack_status = cfn.describe_stacks(StackName=stack_name).get("Stacks")[0].get("StackStatus")
+            stack_status = utils.get_stack(stack_name, cfn).get("StackStatus")
```
sean-smith: same for `get_stack` function

cli/pcluster/commands.py
```diff
@@ -133,12 +133,12 @@ def create(args):  # noqa: C901 FIXME!!!
         LOGGER.debug("StackId: %s", stack.get("StackId"))
 
         if not args.nowait:
-            utils.verify_stack_creation(cfn, stack_name)
+            utils.verify_stack_creation(stack_name, cfn)
```
sean-smith: Rename `cfn` to `cfn_client` to stay consistent with `verify_stack_creation(stack_name, cfn_client=None)`

cli/pcluster/utils.py
enrico-usai: If the stack doesn't exist the describe_stacks raises a `ClientError` so we cannot incur in the `IndexError`, in any case I added the handle for the IndexError too.

```
$ pcluster instances test
ERROR: Stack with id parallelcluster-test does not exist
```

cli/pcluster/commands.py
enrico-usai: Done

cli/pcluster/commands.py
enrico-usai: If you mean: rename cfn to cfn_client --> done.

PR #1393, opened 2019-10-29
Add optional filter to retrieve AMI list given a build date

lukeseawalker:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/util/generate-ami-list.py
+++ b/util/generate-ami-list.py
@@ -55,7 +55,9 @@ def get_ami_list_from_file(regions, cfn_template_file):
     return amis_json
 
 
-def get_ami_list_from_ec2(main_region, regions, date, cookbook_git_ref, node_git_ref, version, owner, credentials):
+def get_ami_list_from_ec2(
+    main_region, regions, date, build_date, cookbook_git_ref, node_git_ref, version, owner, credentials
+):
     amis_json = {}
 
     for region_name in regions:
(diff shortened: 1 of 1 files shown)


util/generate-ami-list.py
```diff
@@ -65,7 +65,11 @@ def get_ami_list_from_ec2(main_region, regions, date, cookbook_git_ref, node_git
         elif cookbook_git_ref and node_git_ref:
             filters.append({"Name": "tag:parallelcluster_cookbook_ref", "Values": ["%s" % cookbook_git_ref]})
             filters.append({"Name": "tag:parallelcluster_node_ref", "Values": ["%s" % node_git_ref]})
-            filters.append({"Name": "name", "Values": ["aws-parallelcluster-*"]})
+            if build_date:
```
enrico-usai: minor: you can write:
```python
filters.append({"Name": "name", "Values": ["aws-parallelcluster-*%s" % build_date if build_date else ""]})
```

lukeseawalker: Thanks for the suggestion.
Btw the correct line is
```
filters.append({"Name": "name", "Values": ["aws-parallelcluster-*%s" % (build_date if build_date else "")]})
```
otherwise you end up in having a Values="" if build_date is not defined

util/generate-ami-list.py
```diff
@@ -215,6 +218,9 @@ def update_amis_txt(amis_txt_file, amis):
     group2 = parser.add_argument_group("Retrieve instances from EC2 searching by cookbook and node git reference")
     group2.add_argument("--cookbook-git-ref", type=str, help="cookbook git hash reference", required=False)
     group2.add_argument("--node-git-ref", type=str, help="node git hash reference", required=False)
+    group2.add_argument(
```
enrico-usai: minor: `type=str` is the default so we can remove it from all the arguments.
minor2: Why did you add `(optional)` only for this parameter?

lukeseawalker: because you could not specify it and the latest will be taken

PR #1395, opened 2019-10-29
Exit with error on get_keys failure

ddeidda:
Exit gracefully when no key pair is found in the provided region. Previously an error like "no index found for item..." was written after the information for no key pair found.


enrico-usai:
Could you please add in the PR description what was the error you was facing and the current output?

PR #1400, opened 2019-10-30
Fix for logging configuration

ddeidda:
Logging configuration was addressing the hierarchy "pcluster.cli"
instead of just "pcluster", hence preventing many log lines from
being printed out

### DIFF
--- a/cli/pcluster/cli.py
+++ b/cli/pcluster/cli.py
@@ -16,6 +16,7 @@ import logging
 import os
 import sys
 import textwrap
+from logging.handlers import RotatingFileHandler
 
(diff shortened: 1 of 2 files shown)


cli/pcluster/cli.py
```diff
@@ -92,7 +93,7 @@ def config_logger():
     log_file_handler = logging.FileHandler(logfile)
```
enrico-usai: Could you please add a log file rotation? 
See: https://github.com/aws/aws-parallelcluster/blob/develop/cli/awsbatch/common.py#L351

PR #1402, opened 2019-10-31
Upload Scripts to S3

sean-smith:

### DIFF
--- a/util/s3_factory.py
+++ b/util/s3_factory.py
@@ -1,0 +1,120 @@
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+
+from pcluster.utils import error
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(format="%(asctime)s - %(levelname)s - %(module)s - %(message)s", level=logging.INFO)
+
+
+class S3DocumentManager:
+    """
+    Class to manage S3 Operations.
+    """
+
+    def __init__(self, _region, _credentials=None):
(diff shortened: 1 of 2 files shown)


util/s3_factory.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Upload file to S3, if it exists will only upload if override = true.
+
+        :param region: region to upload too.
+        :return: None
+        """
+
+        key = "%s/%s" % (self.key_path, self.script)
+        bucket = self.get_bucket(self.bucket)
+        data = open(self.script, "rb")
```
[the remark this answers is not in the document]
sean-smith: I call `put_object` in two places, in the rollback method, I need to get the data from the previous version then pass that to `put_object`, hence the need to get it in `upload_to_s3`

util/s3_factory.py
```diff
@@ -0,0 +1,120 @@
...
+        Get the most recent version of the s3 file.
+
+        :param s3_bucket: bucket
+        :param document_s3_path: s3 key path to the object
+        :return: versionId if it's versioned, else None
+        """
+        response = boto3.client("s3", region_name=self._region, **self._credentials).head_object(
+            Bucket=s3_bucket, Key=document_s3_path
+        )
+        if not response.get("VersionId"):
```
tilne: I think the default behavior should  be to raise an exception here.

sean-smith: changed

PR #1404, opened 2019-10-31
Add debug messages for create and ssh commands

enrico-usai:
To debug customer issues it is useful to see stack status and SSH command when using `pcluster ssh` command and CFN parameters when creating the cluster.

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -121,6 +121,7 @@ def create(args):  # noqa: C901 FIXME!!!
             cfn_params.update(dict(args.extra_parameters))
 
         # prepare input parameters for stack creation and create the stack
+        LOGGER.debug(cfn_params)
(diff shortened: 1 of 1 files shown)


cli/pcluster/commands.py
```diff
@@ -642,9 +643,10 @@ def command(args, extra_args):  # noqa: C901 FIXME!!!
 
         # run command
         if not args.dryrun:
+            LOGGER.debug("Stack status: {0}, SSH command: {1}".format(stack_status, cmd))
             os.system(cmd)
         else:
-            LOGGER.info(cmd)
+            LOGGER.info("Stack status: {0}, SSH command: {1}".format(stack_status, cmd))
```
ddeidda: Could this line be left outside the If?

enrico-usai: moved

PR #1406, opened 2019-11-01
Implement CloudWatch logging integration tests.

tilne:
These tests verify:
* whether or not the log files that were expected to exist on the master
and compute instances actually do exist
* whether or not a log group was created depending on whether the
feature was enabled
* whether or not the log streams for non-empty logs on all of the
instances were created as expected
* whether or not these same log streams contain data
* whether or not these same log streams contain the last line read from
the logs that they represent

There are known issues with these tests:
* They haven't yet been tested with awsbatch as the scheduler
* ~~Tests fail on centos7 currently, due to an issue where the fully
qualified domain name is used where the local domain name is expected~~ (this is no longer true)
* The tests do not verify that the log groups are deleted when the
cluster is deleted
* The tests do not verify that the correctness of the timestamp formats
used to set the event times for log events.

Despite the issues, I thought it worthwhile to get feedback on this
initial implementation.

### DIFF
--- a/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
+++ b/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
@@ -1,0 +1,469 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import json
+import logging
+import random
+import re
+import string
+from os import environ
+from pathlib import Path
+
+import boto3
+import pytest
+
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from tests.common.schedulers_common import get_scheduler_commands
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CloudWatchLoggingClusterState:
+    """
+    Encapsulate the state of a running cluster as it pertains to the CloudWatch logging feature.
+
+    The state is stored in the self._cluster_log_state dict. Here is an example of what that
+    structure might look like for a cluster with one compute node, each containing one log:
+    {
+        "MasterServer": {
+            "node_role": "MasterServer",
+            "hostname": "ip-10-0-127-157.us-west-1.compute.internal",
+            "instance_id": "i-01be4c67943df785d",
+            "logs": {
+                "/var/log/cfn-init-cmd.log": {
+                    "file_path": "/var/log/cfn-init-cmd.log",
+                    "log_stream_name": "cfn-init-cmd",
+                    "exists": true,
+                    "is_empty": false,
+                    "tail": "2019-11-01 21:58:07,407 P1974 [INFO] Completed successfully."
+                }
+            },
+            "agent_status": "running"
+        },
+        "ComputeFleet": {
+            "ip-10-0-155-82.us-west-1.compute.internal": {
+                "node_role": "ComputeFleet",
+                "hostname": "ip-10-0-155-82.us-west-1.compute.internal",
+                "instance_id": "i-0d80db39340ce94d8",
+                "logs": {
+                    "/var/spool/torque/client_logs/*": {
+                        "file_path": "/var/spool/torque/client_logs/*",
+                        "log_stream_name": "torque-client",
+                        "exists": true,
+                        "is_empty": false,
+                        "tail": "2019-11-01 22:00:14.7647 TORQUE authd daemon started"
+                    }
+                },
+                "agent_status": "running"
+            }
+        }
+    }
+    """
+
+    def __init__(self, scheduler, os, cluster, data_retriever):
+        """Get the state of the cluster as it pertains to the CloudWatch logging feature."""
+        self.scheduler = scheduler
+        self.platform = self._base_os_to_platform(os)
+        self.cluster = cluster
+        self.remote_command_executor = RemoteCommandExecutor(self.cluster)
+        self.scheduler_commands = get_scheduler_commands(self.scheduler, self.remote_command_executor)
+        self.data_retriever = data_retriever
+        self._relevant_logs = {"MasterServer": [], "ComputeFleet": []}
+        self._cluster_log_state = {"MasterServer": {}, "ComputeFleet": {}}
+        self._set_cluster_log_state()
+
+    def get_logs_state(self):
+        """Get the state of the log files applicable to each of the cluster's EC2 instances."""
+        desired_keys = ["hostname", "instance_id", "node_role", "agent_status", "logs"]
+        states = [{key: self._cluster_log_state.get("MasterServer").get(key) for key in desired_keys}]
+        states.extend(
+            [
+                {key: host_dict[key] for key in desired_keys}
+                for hostname, host_dict in self._cluster_log_state.get("ComputeFleet").items()
+            ]
+        )
+        assert_that(states).is_length(self.scheduler_commands.compute_nodes_count() + 1)  # computes + master
+        return states
+
+    def _dump_cluster_log_state(self):
+        """Dump the JSON-ified string of self._cluster_log_state for debugging purposes."""
+        return json.dumps(self._cluster_log_state, indent=4)
+
+    @staticmethod
+    def _base_os_to_platform(base_os):
+        """Turn the name of a base OS into the platform."""
+        # Special case: alinux is how the config file refers to amazon linux, but in the chef cookbook
+        # (and the cloudwatch log config produced by it) the platform is "amazon".
+        translations = {"alinux": "amazon"}
+        no_digits = base_os.rstrip(string.digits)
+        return translations.get(no_digits, no_digits)
+
+    def _set_master_instance(self, instance):
+        """Set the master instance field in self.cluster_log_state."""
+        self._cluster_log_state.get("MasterServer").update(
+            {
+                "node_role": "MasterServer",
+                "hostname": instance.get("PrivateDnsName"),
+                "instance_id": instance.get("InstanceId"),
+            }
+        )
+
+    def _add_compute_instance(self, instance):
+        """Update the cluster's log state by adding a compute node."""
+        self._cluster_log_state["ComputeFleet"][instance.get("PrivateDnsName")] = {
+            "node_role": "ComputeFleet",
+            "hostname": instance.get("PrivateDnsName"),
+            "instance_id": instance.get("InstanceId"),
+        }
+
+    def _get_initial_cluster_log_state(self):
+        """Get EC2 instances belonging to this cluster. Figure out their roles in the cluster."""
+        for instance in self.data_retriever.get_ec2_instances():
+            tags = {tag.get("Key"): tag.get("Value") for tag in instance.get("Tags", [])}
+            if tags.get("ClusterName", "") != self.cluster.name:
+                continue
+            elif tags.get("Name", "") == "Master":
+                self._set_master_instance(instance)
+            else:
+                self._add_compute_instance(instance)
+        LOGGER.debug("After getting initial cluster state:\n{}".format(self._dump_cluster_log_state()))
+
+    def _read_log_configs_from_master(self):
+        """Read the log configs file at /usr/local/etc/cloudwatch_log_files.json."""
+        read_cmd = "cat /usr/local/etc/cloudwatch_log_files.json"
+        config = json.loads(self._run_command_on_master(read_cmd))
+        return config.get("log_configs")
+
+    @staticmethod
+    def _clean_log_config(log):
+        """Remove unnecessary fields from the given log dict."""
+        desired_keys = ["file_path", "log_stream_name"]
+        return {key: log[key] for key in desired_keys}
+
+    def _get_relevant_logs(self):
+        """Get subset of all log configs that apply to this cluster's scheduler/os combo."""
+        # Figure out which logs are relevant to the master and computes for this cluster
+        logs = self._read_log_configs_from_master()
+        for log in logs:
+            if self.scheduler not in log.get("schedulers") or self.platform not in log.get("platforms"):
+                continue
+            for node_role in log.get("node_roles"):
+                self._relevant_logs[node_role].append(self._clean_log_config(log))
+        # Give each nodes representative dict in self._cluster_log_state a copy
+        self._cluster_log_state["MasterServer"]["logs"] = {
+            log.get("file_path"): log for log in self._relevant_logs.get("MasterServer")
+        }
+        for _hostname, compute_instance_dict in self._cluster_log_state.get("ComputeFleet").items():
+            compute_instance_dict["logs"] = {
+                log.get("file_path"): log.copy() for log in self._relevant_logs.get("ComputeFleet")
+            }
+        LOGGER.debug("After populating relevant logs:\n{}".format(self._dump_cluster_log_state()))
+
+    def _run_command_on_master(self, cmd):
+        """Run cmd on cluster's MasterServer."""
+        return self.remote_command_executor.run_remote_command(cmd).stdout.strip()
+
+    def _run_command_on_computes(self, cmd, assert_success=True):
+        """Run cmd on all computes in the cluster."""
+        # Create directory in /shared to direct outputs to
+        out_dir = Path(self._run_command_on_master("mktemp -d -p /shared"))
+        redirect = " > {out_dir}/$(hostname -f) ".format(out_dir=out_dir)
+        remote_cmd = cmd.format(redirect=redirect)
+
+        # Run the command, wait for it to complete
+        submit_out = self.scheduler_commands.submit_command(
+            remote_cmd, nodes=self.scheduler_commands.compute_nodes_count()
+        )
+        job_id = self.scheduler_commands.assert_job_submitted(submit_out.stdout)
+        self.scheduler_commands.wait_job_completed(job_id)
+        if assert_success:
+            self.scheduler_commands.assert_job_succeeded(job_id)
+
+        # Read the output and map it to the hostname
+        outputs = {}
+        result_files = self._run_command_on_master("ls {}".format(out_dir))
+        for hostname in result_files.split():
+            outputs[hostname] = self._run_command_on_master("sudo cat {}".format(out_dir / hostname))
+        self._run_command_on_master("rm -rf {}".format(out_dir))
+        return outputs
+
+    def _populate_master_log_existence(self):
+        """Figure out which of the relevant logs for the MasterServer don't exist."""
+        for log_path, log_dict in self._cluster_log_state.get("MasterServer").get("logs").items():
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.ini
```diff
@@ -0,0 +1,26 @@
...
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = {{ scheduler }}
+master_instance_type = {{ instance }}
+compute_instance_type = {{ instance }}
+initial_queue_size = {{ queue_size }}
+max_queue_size = {{ queue_size }}
```
sean-smith: Why do you need `max_queue_size `?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+        if self.enabled:  # Log streams are only relevant when the feature is enabled
+            observed_streams = self.data_retriever.get_log_streams(self.log_group_name)
+            expected_stream_index = self._get_expected_log_stream_index(logs_state)
+            self.verify_log_streams_exist(logs_state, expected_stream_index, observed_streams)
+            self.verify_log_streams_data(logs_state, expected_stream_index, observed_streams)
+
+
+@pytest.mark.parametrize("cw_logging_enabled", [True, False])
+@pytest.mark.regions(["us-west-1", "us-west-2", "us-east-1", "us-east-2"])
+@pytest.mark.instances(["t2.micro", "c4.xlarge"])
```
sean-smith: let's just go with `t2.micro` for now

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+@pytest.mark.oss(["alinux", "centos6", "centos7", "ubuntu1604", "ubuntu1804"])
```
sean-smith: remove this - all oss are included by default

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+@pytest.mark.schedulers(["awsbatch", "sge", "slurm", "torque"])
```
sean-smith: same

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+        os,
+        pcluster_config_reader,
+        clusters_factory,
+        cw_logging_enabled):
+    """
+    Test all CloudWatch logging features.
+
+    All tests are grouped in a single function so that the cluster can be reused for all of them.
+    """
+    # Allow certain params to be set via environment variable in case re-testing on an existing cluster
```
sean-smith: Assume there's 1 test per cluster

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def _add_compute_instance(self, instance):
+        """Update the cluster's log state by adding a compute node."""
+        self._cluster_log_state["ComputeFleet"][instance["PrivateDnsName"]] = {
+            "node_role": "ComputeFleet",
+            "hostname": instance["PrivateDnsName"],
+            "instance_id": instance["InstanceId"]
+        }
+
+    def _get_initial_cluster_log_state(self):
```
sean-smith: There's a much easier way to do this - see https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/commands.py#L466

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+    @staticmethod
+    def _base_os_to_platform(base_os):
+        """Turn the name of a base OS into the platform."""
+        # Special case: alinux is how the config file refers to amazon linux, but in the chef cookbook
+        # (and the cloudwatch log config produced by it) the platform is "amazon".
+        translations = {'alinux': 'amazon'}
+        no_digits = base_os.rstrip(string.digits)
+        return translations.get(no_digits, no_digits)
+
+    def _set_master_instance(self, instance):
```
sean-smith: You can just get this from cloudformation

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def get_log_streams(self, log_group_name):
+        """Get list of log streams."""
+        logs_client = boto3.client('logs', region_name=self.region)
+        streams = logs_client.describe_log_streams(logGroupName=log_group_name)['logStreams']
+        LOGGER.debug("Log streams for {group}:\n{streams}".format(
+            group=log_group_name, streams=json.dumps(streams, indent=4)))
+        return streams
+
+    def get_log_events(self, log_group_name, log_stream_name):
```
sean-smith: What if there's 10,000,000 events?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,461 @@
...
+        """Get log events for the given log_stream_name."""
+        logs_client = boto3.client('logs', region_name=self.region)
+        events = logs_client.get_log_events(logGroupName=log_group_name, logStreamName=log_stream_name)
+        LOGGER.debug("Log events for {group}/{stream}:\n{events}".format(
+            group=log_group_name, stream=log_stream_name, events=json.dumps(events, indent=4)))
+        return events
+
+    def get_ec2_instances(self):
```
sean-smith: There's probably an easier way to get the information you need.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: The reason I included c4.xlarge is because in the Jenkins build_and_test pipeline, `c4.xlarge` is the smallest instance included in the the commercial and china instances lists. In looking at the pipeline again, my inclination is to enable it for c5.xlarge as well, given that's the only one in the default govcloud instances list.

I want these tests to be enabled in all three partitions. Do the build_and_test pipelines get launched with non-default instances lists on a regular basis?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.ini
tilne: I don't.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
sean-smith: Not very often.

It's run on every instance type passed in, hence I would limit it to one instance type that runs across partitions. It looks like `c5.xlarge` is the best one to do that with.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.ini
sean-smith: please remove

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Configuring `retention_days` via the environment is useful when manually re-running tests on an already-existing cluster. I changed the comment to clarify the use case.

Configuring `queue_size` via the environment is useful because it enables testing this feature on clusters with a lot of computes by setting the environment variable rather than changing the code. It's not something I anticipate using a lot, but it's nice for one-off testing for larger clusters.

Is there a better way to do this?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Okay. I'll include that one, and I'll keep `t2.micro` for the sake of manual testing without having to whitelist it first.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.ini
tilne: I did. I just haven't pushed it yet. I was going to incorporate the necessary changes for all of the feedback before pushing.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: I didn't use an iterator for pagination because the log streams that I'm currently getting events from do not get big enough to exceed the 1MB limit enforced by the `get_log_events` method.

But now that I'm reconsidering the decision again, it probably makes sense to just do it now, given that it seems very possible that a log that does so might be added in the future.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: I'm gonna resolve this because it's really closely related to your comment on `_get_initial_cluster_log_state`.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: For each instance I need:
* its role in the cluster (master or compute),
* its instance ID
* its hostname

I can get the first two items in that list via the function you pointed me to (using the `LogicalResourceId` and `PhysicalResourceId` keys of the dicts in the returned list). What would be the best way to then map the instance IDs to hostnames? Call `ec2.describe_instances` with an `InstanceIds` arg obtained from the returned value of the first call? That makes this a two-step process instead of one (albeit two more reasonably sized processes).

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Resolving this because I think it's very closely related to the conversation under `_get_initial_cluster_log_state`.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Actually now that I look at it again, the decision to ignore pagination was motivated by the fact that the data is only used to ensure:
* the log stream is not empty
* the last event in the corresponding log file on the instance has been ingested by the log stream

So if the assumption "less than ~8,000 log events will be ingested for any given log in-between the time the 'test tail event' is read and the time the log stream is queried for events" seems safe, I think this can probably remain as is.

Thoughts?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CloudWatchLoggingClusterState:
+    """
+    Encapsulate the state of a running cluster as it pertains to the CloudWatch logging feature.
+
+    The state is stored in the self._cluster_log_state dict. Here is an example of what that
+    structure might look like for a cluster with one comptue node, each containing one log:
```
sean-smith: minor: `comptue` -> `compute`

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
sean-smith: That makes sense. You can either call `ec2.descibe_instances()` or `asg.describe_autoscaling_instances()` which both take the instance Id's as input so it's a moot point.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+        """Set the master instance field in self.cluster_log_state."""
+        self._cluster_log_state["MasterServer"].update(
+            {"node_role": "MasterServer", "hostname": instance["PrivateDnsName"], "instance_id": instance["InstanceId"]}
+        )
+
+    def _add_compute_instance(self, instance):
+        """Update the cluster's log state by adding a compute node."""
+        self._cluster_log_state["ComputeFleet"][instance["PrivateDnsName"]] = {
+            "node_role": "ComputeFleet",
+            "hostname": instance["PrivateDnsName"],
```
sean-smith: Use `.get()` instead of `instance["PrivateDnsName"]`

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+            "instance_id": instance["InstanceId"],
+        }
+
+    def _get_initial_cluster_log_state(self):
+        """Get EC2 instances belonging to this cluster. Figure out their roles in the cluster."""
+        for reservation in self.data_retriever.get_ec2_instances()["Reservations"]:
```
sean-smith: You may need to paginate this call. If the number of instances is high, this could lead to not getting all the instances in the cluster. See https://github.com/aws/aws-parallelcluster/blob/f842e47fa5dca64e98c44e932ba549605f35e345/cli/pcluster/utils.py#L49

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+            for instance in reservation["Instances"]:
```
sean-smith: use `.get()` basically everywhere

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+        job_id = self.scheduler_commands.assert_job_submitted(submit_out.stdout)
+        self.scheduler_commands.wait_job_completed(job_id)
+        if assert_success:
+            self.scheduler_commands.assert_job_succeeded(job_id)
+
+        # Read the output and map it to the hostname
+        outputs = {}
+        result_files = self._run_command_on_master("ls {}".format(out_dir))
+        for hostname in result_files.split():
+            path = "/".join((out_dir, hostname))
```
sean-smith: Use a path object for `out_dir`, then you can just do `out_dir / hostname`. See https://stackoverflow.com/questions/3751900/create-file-path-from-variables#answer-56620931

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+    def _populate_compute_log_emptiness_and_tail(self):
+        """Figure out which of the relevant logs for the ComputeFleet nodes are empty."""
+        for log_dict in self._relevant_logs["ComputeFleet"]:
+            # If this file doesn't exist on any of the computes, don't assert success
+            assert_success = True
+            for _, compute_dict in self._cluster_log_state["ComputeFleet"].items():
+                if not compute_dict["logs"][log_dict["file_path"]]["exists"]:
+                    assert_success = False
+                    break
+            cmd = "sudo tail {{redirect}} -n 1 {path}".format(path=log_dict["file_path"])
```
sean-smith: Where does `{{redirect}}` come from?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
sean-smith: Agreed

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -0,0 +1,460 @@
...
+            expected_stream_index = self._get_expected_log_stream_index(logs_state)
+            self.verify_log_streams_exist(logs_state, expected_stream_index, observed_streams)
+            self.verify_log_streams_data(logs_state, expected_stream_index, observed_streams)
+
+        if self.failures:
+            pytest.fail("Failures: {}".format(", ".join(self.failures)), pytrace=False)
+
+
+@pytest.mark.parametrize("cw_logging_enabled", [True, False])
+@pytest.mark.regions(["us-west-1", "us-west-2", "us-east-1", "us-east-2"])
```
sean-smith: Stick with 1 commercial region, 1 china region and 1 govcloud region.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: It gets `format`'ed in by the `_run_command_on_computes`. I did it this way because I couldn't figure out a better way for commands that required piping (like the one in `_populate_compute_log_existence`).

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done. Resulted in a lot of changes because of how I stored everything in a dict, so I need to test the changes first.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done.

tilne:
@sean-smith I think I've incorporated all of your feedback. This is ready for another review at your convenience.

PR #1420, opened 2019-11-06
Updated README guide with new pcluster flow

ddeidda:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,16 @@ You can build higher level workflows, such as a Genomics portal that automates t
 
 Quick Start
 -----------
-First, install the library:
+**IMPORTANT**: you will need an **Amazon EC2 Key Pair** to be able to complete the following steps.
+Please see the `Official AWS Guide <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html>`_.
+
+First, make sure you have installed the `AWS Command Line Interface <http://>`_:
+
+.. code-block:: sh
+
+    $ pip install awscli
+
+Then you can install AWS ParallelCluster:
 
 .. code-block:: sh
 
@@ -35,34 +44,59 @@ Next, configure your aws credentials and default region:
     Default region name [us-east-1]:
     Default output format [None]:
 
-Then, run pcluster configure:
+Then, run ``pcluster configure``. A list of valid options will be displayed for each
+configuration parameter. Type an option number and press ``Enter`` to select a specific option,
+or just press ``Enter`` to accept the default option.
 
 .. code-block:: ini
 
   $ pcluster configure
-  Cluster Template [default]:
-  Acceptable Values for AWS Region ID:
-      ap-south-1
-      ...
-      us-west-2
-  AWS Region ID [us-east-1]:
-  VPC Name [myvpc]:
-  Acceptable Values for Key Name:
-    keypair1
-    keypair-test
-    production-key
-  Key Name []:
-  Acceptable Values for VPC ID:
-    vpc-1kd24879
-    vpc-blk4982d
-  VPC ID []:
-  Acceptable Values for Master Subnet ID:
-    subnet-9k284a6f
-    subnet-1k01g357
-    subnet-b921nv04
-  Master Subnet ID []:
+  INFO: Configuration file /dir/conf_file will be written.
+  Press CTRL-C to interrupt the procedure.
 
-Now you can create your first cluster;
+
+  Allowed values for AWS Region ID:
+  1. eu-north-1
+  ...
+  15. us-west-1
+  16. us-west-2
+  AWS Region ID [us-east-1]:
+  ...
+
+Be sure to select a region containing the EC2 key pair you wish to use. You can also import a public key using
+`these instructions <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html#how-to-generate-your-own-key-and-import-it-to-aws>`_.
+
+During the process you will be asked to set up your networking environment. The wizard will offer you the choice of
+using an existing VPC or creating a new one on the fly.
+
+.. code-block:: ini
+
+  Automate VPC creation? (y/n) [n]:
+
+Enter '``n``' if you already have a VPC suitable for the cluster. Otherwise you can let ``pcluster configure``
+create a VPC for you. The same choice is given for the subnet: you can select a valid subnet ID for
+both the master and compute nodes, or you can let ``pcluster configure`` set up everything for you.
+The same choice is given for the subnet configuration: you can select a valid subnet ID for both
+the master and compute nodes, or you can let pcluster configure set up everything for you.
+In the latter case, just select the configuration you prefer.
+
+.. code-block:: ini
+
+  Automate Subnet creation? (y/n) [y]: y
+  Allowed values for Network Configuration:
+  1. Master in a public subnet and compute fleet in a private subnet
+  2. Master and compute fleet in the same public subnet
+
+
+At the end of the process a message like this one will be shown:
+
+.. code-block:: ini
+
+  Configuration file written to /dir/conf_file
+  You can edit your configuration file or simply run 'pcluster create -c /dir/conf_file cluster-name' to create your cluster
+
+
+Now you can create your first cluster:
 
 .. code-block:: sh
 


README.rst
```diff
@@ -19,6 +19,10 @@ You can build higher level workflows, such as a Genomics portal that automates t
 
 Quick Start
 -----------
+**IMPORTANT**: you will need an **Amazon EC2 Key Pair** to be able to complete the following steps.
+Please see the `official AWS guide <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html>`_
+if you need to create a Key Pair.
```
enrico-usai: I'd remove "if you need to create a Key Pair."

README.rst
```diff
@@ -19,6 +19,10 @@ You can build higher level workflows, such as a Genomics portal that automates t
...
+
 First, install the library:
 
 .. code-block:: sh
```
enrico-usai: should we add the aws cli as requirement?

README.rst
```diff
@@ -35,32 +39,59 @@ Next, configure your aws credentials and default region:
...
-    vpc-blk4982d
-  VPC ID []:
-  Acceptable Values for Master Subnet ID:
-    subnet-9k284a6f
-    subnet-1k01g357
-    subnet-b921nv04
-  Master Subnet ID []:
+  ...
+
+Make sure to select a region where you own at least a key pair otherwise the process will end with this message.
```
enrico-usai: where -> on which ?

Then I'd remove the error message, because in future we can change the error message and this help will became outdated.

I'd write: "Make sure to select a region on which you own at least an EC2 key pair."

README.rst
```diff
@@ -35,32 +39,59 @@ Next, configure your aws credentials and default region:
...
 
 .. code-block:: ini
 
   $ pcluster configure
-  Cluster Template [default]:
-  Acceptable Values for AWS Region ID:
-      ap-south-1
-      ...
-      us-west-2
+  INFO: Configuration file /tmp/test will be written.
```
enrico-usai: let's use "/dir/conf_file" as in the following examples.

README.rst
ddeidda: Done

README.rst
ddeidda: Done

README.rst
ddeidda: Done

README.rst
ddeidda: Done

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
-    vpc-blk4982d
-  VPC ID []:
-  Acceptable Values for Master Subnet ID:
-    subnet-9k284a6f
-    subnet-1k01g357
-    subnet-b921nv04
-  Master Subnet ID []:
+  ...
+
+Make sure to select a region on which you own at least an EC2 key pair.
```
tilne: Two comments on this line:
* I would instead say: "Be sure to select a region containing the EC2 key pair you wish to use."
* This suggests that a user is limited to regions where they have already uploaded the EC2 key they want to use. I would add a second sentence that says "You can also import a public key using [these instructions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html#how-to-generate-your-own-key-and-import-it-to-aws)."

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
+
+During the process you will be asked to setup your networking environment. The wizard will
```
tilne: "setup" should be "set up"

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
+offer you the possibility of creating a new VPC on the fly.
+
+.. code-block:: ini
+
+  Automate VPC creation? (y/n) [n]:
+
+Enter '``n``' if you already have a correctly set up VPC. Otherwise you can let ``pcluster configure``
```
tilne: "correctly set up" is vague. Does it mean that a customer has already specified in the config file the VPC and subnet ID that they want to use? If so, I'd say something like "...if you have already specified the VPC configuration that you wish to use. Otherwise..."

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
+
+At the end of the process a message like this one will be shown:
+
+.. code-block:: ini
+
+  Configuration file written to /dir/conf_file
+  You can edit your configuration file or simply run 'pcluster create -c /dir/conf_file cluster-name' to create your cluster
+
 
 Now you can create your first cluster;
```
tilne: This wasn't part of the changes introduced by this PR, but "cluster;" should be "cluster:"

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
-    subnet-9k284a6f
-    subnet-1k01g357
-    subnet-b921nv04
-  Master Subnet ID []:
+  ...
+
+Make sure to select a region on which you own at least an EC2 key pair.
+
+During the process you will be asked to setup your networking environment. The wizard will
+offer you the possibility of creating a new VPC on the fly.
```
tilne: I would use "choice" instead of "possibility" here. I would also, describe the other option: "The wizard will offer you the choice of using an existing VPC or creating a new one on the fly."

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
...
+
+.. code-block:: ini
+
+  Automate VPC creation? (y/n) [n]:
+
+Enter '``n``' if you already have a correctly set up VPC. Otherwise you can let ``pcluster configure``
+create a VPC for you. The same possibility is given for the subnet: you can select a valid subnet id for
+master and compute nodes or you can let ``pcluster configure`` set up everything for you. In that case,
+just select the configuration you prefer.
```
tilne: I'd make a these changes to these two sentences:
* instead of "possibility" I'd use "choice"
* "id" should be capitalized: "ID"
* I'd put a "both the" in front of "master and compute nodes"
* I'd put a comma before or: "...nodes, or you can..."
* Instead of "In that case," I'd say "In the latter case," just to be more specific about which case the further configuration would apply to (although that is illustrated afterwards).
So I would have this: "The same choice is given for the subnet configuration: you can select a valid subnet ID for both the master and compute nodes, or you can let `pcluster configure` set up everything for you. In the latter case, just select the configuration you prefer."

README.rst
ddeidda: What I meant here is: "If you already have a VPC on the AWS cloud that you are sure you can install the cluster on". For instance, I have my VPC with two subnets and I usually create all my clusters there. Could you suggest me a better way to express this concept?

README.rst
```diff
@@ -19,7 +19,16 @@ You can build higher level workflows, such as a Genomics portal that automates t
 
 Quick Start
 -----------
-First, install the library:
+**IMPORTANT**: you will need an **Amazon EC2 Key Pair** to be able to complete the following steps.
+Please see the `Official AWS Guide <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html>`_.
+
+First, make sure you installed the `AWS Command Line Interface <http://>`_:
```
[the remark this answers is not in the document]
ddeidda: :+1:

README.rst
```diff
@@ -35,32 +44,54 @@ Next, configure your aws credentials and default region:
     Default region name [us-east-1]:
     Default output format [None]:
 
-Then, run pcluster configure:
+Then, run ``pcluster configure``. A list of valid options will be displayed for each
+configuration parameter: type an option number and press ``Enter`` to confirm your choice, or just press ``Enter``
```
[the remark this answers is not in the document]
ddeidda: :+1:

README.rst
ddeidda: :+1:

README.rst
ddeidda: :+1:

README.rst
ddeidda: :+1:

README.rst
ddeidda: :+1:

README.rst
ddeidda: :+1:

ddeidda:
Thanks for the careful review!

README.rst
tilne: Okay, I think that sounds good. "...if you already have a VPC suitable for the cluster."

tilne:
> Thanks for the careful review!

Sure thing. Hopefully my responses weren't too annoying. They were mostly nitpicks.

README.rst
ddeidda: :+1:

ddeidda:
> Sure thing. Hopefully my responses weren't too annoying. They were mostly nitpicks.

Not at all! Much appreciated! Thanks

PR #1435, opened 2019-11-08
Add script to get archive of cluster's log group

tilne:
This is the initial check-in for a script that enables a user to extract
all of the logs in a given cluster's CloudWatch log group and create a
tarball of the extracted data. This allows for ease in attaching logs to
tickets.

Potentially useful enhancements include:
* Allow for optionally deleting data from the s3 bucket that serves as
the intermediary between CloudWatch and the local filesystem.
* Enable a user to specify that the data only be exported to s3 and not
downloaded to the local filesystem.

### DIFF
--- a/util/logging/retrieve-cluster-logs.py
+++ b/util/logging/retrieve-cluster-logs.py
@@ -1,0 +1,273 @@
+#!/usr/bin/env python
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+#
+# Future enhancements:
+#   * Catch KeyboardInterrupt exceptions while waiting for export tasks to start or finish and clean up accordingly.
+
+import gzip
+import json
+import logging
+import os
+import sys
+import tarfile
+import tempfile
+import time
+from datetime import datetime
+
+import argparse
+import boto3
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__file__)
+
+
+def err_and_exit(message):
+    """Log the given error message and exit nonzero."""
+    LOGGER.error(message)
+    sys.exit(1)
+
+
+def get_default_bucket_prefix(cluster_name):
+    """Get default keypath to store logs under in an s3 bucket for the given cluster."""
+    return "{cluster_name}-logs-{timestamp}".format(cluster_name=cluster_name, timestamp=datetime.now().timestamp())
+
+
+def get_log_group(args):
+    """Get log group for args.cluster."""
+    log_group_name = "/aws/parallelcluster/{}".format(args.cluster)
+    logs = boto3.client("logs", region_name=args.region)
+    paginator = logs.get_paginator("describe_log_groups")
+    for result in paginator.paginate(logGroupNamePrefix=log_group_name):
+        for group in result.get("logGroups"):
+            if group.get("logGroupName") == log_group_name:
+                return group
+    err_and_exit(
+        "Unable to find log group for cluster {cluster}. Expected to find one named {log_group_name}".format(
+            cluster=args.cluster, log_group_name=log_group_name
+        )
+    )
+
+
+def verify_bucket_exists_in_region(args):
+    """Verify that bucket exists."""
+    bucket = boto3.resource("s3", region_name=args.region).Bucket(args.bucket)
+    if bucket.creation_date is None:
+        err_and_exit("s3 bucket {} does not exist.".format(args.bucket))
+    bucket_region = bucket.meta.client.get_bucket_location(Bucket=bucket.name).get("LocationConstraint")
+    expected_region = args.region if args.region != "us-east-1" else None
+    if bucket_region != expected_region:
+        err_and_exit(
+            "CloudWatch requires buckets to use for exporting logs to be in the same region as the log group. "
+            "The given cluster's log group is in {region}, but the given bucket's region is {bucket_region} ".format(
+                region=expected_region, bucket_region=bucket_region
+            )
+        )
+
+
+def verify_times(args):
+    """Verify that the start and end times represent a window containing at least one log event."""
+    if args.from_time >= args.to_time:
+        err_and_exit("Start time must be earlier than end time.")
+    logs_client = boto3.client("logs", region_name=args.region)
+    event_in_window = logs_client.filter_log_events(
+        logGroupName=args.log_group.get("logGroupName"),
+        startTime=int(1000 * args.from_time.timestamp()),
+        endTime=int(1000 * args.to_time.timestamp()),
+        limit=1,
+    ).get("events")
+    if not event_in_window:
+        err_and_exit(
+            "No log events in the log group {log_group} in interval starting at {start} and ending at {end}".format(
+                log_group=args.log_group.get("logGroupName"), start=args.from_time, end=args.to_time
+            )
+        )
+
+
+def parse_args():
+    """Parse command line args."""
+
+    def timestamp_from_arg(arg):
+        """Convert arg into a UNIX timestamp."""
+        return datetime.fromtimestamp(int(arg))
+
+    parser = argparse.ArgumentParser(description="Create an archive for a ParallelCluster's CloudWatch logs.")
+    parser.add_argument("--bucket", required=True, help="s3 bucket to export CloudWatch logs data to.")
+    parser.add_argument("--cluster", required=True, help="Name of cluster whose logs to get.")
+    parser.add_argument(
+        "--bucket-prefix", help="Keypath under which exported CloudWatch logs data will be stored in s3 bucket."
+    )
+    parser.add_argument(
+        "--from-time",
+        type=timestamp_from_arg,
+        help="Start time of interval of interest for log events, as number of seconds since the epoch. Deafults to "
+        "cluster's start time.",
+    )
+    parser.add_argument("--region", required=True, help="Region in which the CloudWatch log group exists.")
+    parser.add_argument(
+        "--to-time",
+        type=timestamp_from_arg,
+        help="End time of interval of interest for log events, as number of seconds since the epoch. Defaults to the "
+        "current time.",
+    )
+    args = parser.parse_args()
+
+    # Set defaults that require other args
+    if not args.bucket_prefix:
+        args.bucket_prefix = get_default_bucket_prefix(args.cluster)
+    args.log_group = get_log_group(args)
+    if args.from_time is None:
+        args.from_time = datetime.fromtimestamp(args.log_group.get("creationTime") / 1000)
+    if args.to_time is None:
+        args.to_time = datetime.now()
+
+    # Verify args
+    verify_bucket_exists_in_region(args)
+    verify_times(args)
+
+    return args
+
+
+def start_export_task(logs_client, args):
+    """Start the task that will export the cluster's logs to an s3 bucket, and return the task ID."""
+    LOGGER.info(
+        "Starting export of logs from log group {log_group} to s3 bucket {bucket}".format(
+            log_group=args.log_group.get("logGroupName"), bucket=args.bucket
+        )
+    )
+    try:
+        response = logs_client.create_export_task(
+            logGroupName=args.log_group.get("logGroupName"),
+            fromTime=int(1000 * args.from_time.timestamp()),
+            to=int(1000 * args.to_time.timestamp()),
+            destination=args.bucket,
+            destinationPrefix=args.bucket_prefix,
+        )
+    except Exception as err:
+        if "Please check if CloudWatch Logs has been granted permission to perform this operation." in str(err):
+            err_and_exit(
+                "CloudWatch Logs needs GetBucketAcl and PutObject permisson for the s3 bucket {bucket}. See {url} for "
+                "more details.".format(
+                    bucket=args.bucket,
+                    url="https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/S3ExportTasks.html#S3Permissions",
+                )
+            )
+        else:
+            err_and_exit("Unexpected error when starting export task: {}".format(err))
+    return response.get("taskId")
+
+
+def get_status_for_export_task(logs_client, task_id):
+    """Get the status for the CloudWatch export task with the given task_id."""
+    tasks = logs_client.describe_export_tasks(taskId=task_id).get("exportTasks")
+    if not tasks:
+        err_and_exit("Unable to get status for CloudWatch logs export task with ID={}".format(task_id))
+    elif len(tasks) > 2:
+        err_and_exit(
+            "More than one CloudWatch logs export task with ID={task_id}:\n{tasks}".format(
+                task_id=task_id, tasks=json.dumps(tasks, indent=2)
+            )
+        )
+    return tasks[0].get("status").get("code")
+
+
+def wait_for_task_completion(logs_client, task_id):
+    """Wait for the CloudWatch logs export task given by task_id to finish."""
+    LOGGER.info("Waiting for export task with task ID {} to finish.".format(task_id))
+    status = "PENDING"
+    still_running_statuses = ("PENDING", "PENDING_CANCEL", "RUNNING")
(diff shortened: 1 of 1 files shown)


util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+def get_default_bucket_prefix(cluster_name):
+    """Get default keypath to store logs under in an s3 bucket for the given cluster."""
+    return "{cluster_name}-logs-{end_time}".format(
+        cluster_name=cluster_name, end_time=datetime.now().strftime("%Y%m%d%H%M%S")
+    )
+
+
+def get_log_group(args):
+    """Get log group for args.cluster."""
+    log_group_name = "/aws/parallelcluster/{}".format(args.cluster)
```
enrico-usai: We should have a global variable defined on top of the module for the group name path or use a global variable from the source code

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+            if group.get("logGroupName") == log_group_name:
+                return group
+    err_and_exit(
+        "Unable to find log group for cluster {cluster}. Expected to find one named {log_group_name}".format(
+            cluster=args.cluster, log_group_name=log_group_name
+        )
+    )
+
+
+def verify_bucket_exists_in_region(args):
```
enrico-usai: minor: as internal convention we are using one underscore prefix for functions that are only used in the module itself.

tilne: Since this is meant to be used as a script, that would mean all of the functions should have that prefix, correct?

util/retrieve-cluster-logs.py
tilne: What's the motivation for making it a global?

Is your concern that this function is too expensive to call every time the log group name is needed? I agree that it is. This function is only called once when parsing the command line. It has to be done then because it requires the cluster's name as input. The returned value is saved in the args object that's passed around the rest of the script, so the function is not called again.

enrico-usai: The "/aws/parallelcluster/" path is something like a configuration, that must correspond to the one used in the cloudwatch resource creation and in the rest of the code, so I think we should have a single place on which it is defined, if in future we want to change it.

If we want to keep this script separate from the rest of the code I'd move the definition of the log group name on top, to be clear when you open the file what are the "configuration" parameters.

What do you think?

util/retrieve-cluster-logs.py
enrico-usai: If the script remains isolated from the rest of the code it is ok for me to keep the code as is.
If we'll move this functionality in our cli I'd add the underscore prefix.

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+    """Get default keypath to store logs under in an s3 bucket for the given cluster."""
+    return "{cluster_name}-logs-{end_time}".format(
+        cluster_name=cluster_name, end_time=datetime.now().strftime("%Y%m%d%H%M%S")
+    )
+
+
+def get_log_group(args):
+    """Get log group for args.cluster."""
+    log_group_name = "/aws/parallelcluster/{}".format(args.cluster)
+    logs = boto3.client("logs", region_name=args.region)
```
demartinofra: if you don't want to pass this region around you could set it as an env variable named AWS_DEFAULT_REGION. boto3 will pick that up automatically.

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+def err_and_exit(message):
+    """Log the given error message and exit nonzero."""
+    LOGGER.error(message)
+    sys.exit(1)
+
+
+def get_default_bucket_prefix(cluster_name):
+    """Get default keypath to store logs under in an s3 bucket for the given cluster."""
+    return "{cluster_name}-logs-{end_time}".format(
+        cluster_name=cluster_name, end_time=datetime.now().strftime("%Y%m%d%H%M%S")
```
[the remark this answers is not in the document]
demartinofra: for the format shall we use a delimiter or use the unix timestamp format directly?

util/retrieve-cluster-logs.py
tilne: Agreed. Since I'm checking this in as a standalone script I'll resolve this, but I'll be sure to adhere to this convention if adding this code to the CLI package.

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+    )
+
+
+def verify_bucket_exists_in_region(args):
+    """Verify that bucket exists."""
+    bucket = boto3.resource("s3", region_name=args.region).Bucket(args.bucket)
+    if bucket.creation_date is None:
+        err_and_exit("s3 bucket {} does not exist.".format(args.bucket))
+    bucket_region = bucket.meta.client.get_bucket_location(Bucket=bucket.name).get("LocationConstraint")
```
demartinofra: just double check that this works in us-east-1 as well

demartinofra: Also a more general comment: we should collect all IAM policies required to run this script and see what happens when the boto3 calls are failing because of permissions. Are we handling this failures?

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+
+def start_export_task(logs_client, args):
+    """Start the task that will export the cluster's logs to an s3 bucket, and return the task ID."""
+    LOGGER.info(
+        "Starting export of logs from log group {log_group} to s3 bucket {bucket}".format(
+            log_group=args.log_group.get("logGroupName"), bucket=args.bucket
+        )
+    )
+    start_time = datetime.fromtimestamp(args.log_group.get("creationTime") / 1000)
+    end_time = datetime.now()
```
demartinofra: same here for utc and everywhere else we are using a timestamp which depends on the timezone

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+    if bucket_region != args.region:
+        err_and_exit(
+            "CloudWatch requires buckets to use for exporting logs to be in the same region as the log group. "
+            "The given cluster's log group is in {region}, but the given bucket's region is {bucket_region} ".format(
+                region=args.region, bucket_region=bucket_region
+            )
+        )
+
+
+def parse_args():
```
demartinofra: future enhancement: --from-time --to-time options

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+    return tasks[0].get("status").get("code")
+
+
+def wait_for_task_completion(logs_client, task_id):
+    """Wait for the CloudWatch logs export task given by task_id to finish."""
+    LOGGER.info("Waiting for export task with task ID {} to finish.".format(task_id))
+    status = "PENDING"
+    still_running_statuses = ("PENDING", "PENDING_CANCEL", "RUNNING")
+    while status in still_running_statuses:
+        status = get_status_for_export_task(logs_client, task_id)
```
demartinofra: we probably want to set a sleep between consecutive calls.

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
...
+    return status
+
+
+def export_logs_to_s3(args):
+    """Export the contents of a cluster's CloudWatch log group to an s3 bucket."""
+    logs_client = boto3.client("logs", region_name=args.region)
+    task_id = start_export_task(logs_client, args)
```
demartinofra: Shall we also check that there isn't a running export before starting a new one?

util/retrieve-cluster-logs.py
tilne: The prefix is used in four places within the code base:
* cookbook
* cfn template
* integration tests
* this script

I don't know of a way to share with the first two items in this list a definition for this prefix. If you know of a way I could do that I'd be open to trying it out.

For sharing the definition between the last two items, my approach would be to create a subpackage `pcluster.cloudwatch_logs` with a `utils` module that contains a function that wraps the prefix. I've made this change and am pushing it now. Let me know what you think.

util/retrieve-cluster-logs.py
tilne: That is a very good idea, depending on how we want to handle that situation. I need to confirm this, but my understanding is that the logs client's `create_export_task` queues the export task if there's already one running in that region.

My initial reaction would be to not explicitly check for an ongoing task, but instead check if the newly created export task is queued based on the export task's status code. Then if it is, log that information and any information made available by the API regarding where in the queue the task is (which is something else I'd need to look into).

The alternative would be to check for ongoing export tasks and then fail out of the script if one exists.

Your thoughts?

util/retrieve-cluster-logs.py
enrico-usai: It is good for me. 👍 
Alas we don't have a way (yet?) to share information between CFN, cookbook and python.

For the mapping file I created a test_source_consistency.py unit test that verifies if the default set in CFN are the same used for the defaults in the mappings.py, but cookbook files are totally separated.

util/retrieve-cluster-logs.py
tilne: Makes sense. I'll set it to sleep for one second in between calls.

util/retrieve-cluster-logs.py
tilne: The security/access issues that I handled are:
* the constraint that the bucket and log group must exist in the same region
* the bucket must allow the CloudWatch logs service to get its ACL and put an object where it needs to based on the key prefix used.

There are quite a few boto3 calls that I do not ensure a user has permission for first:
* In addition to the use of the s3 bucket resource to list and download objects, it also call's the s3 client's `get_bucket_location`
* the logs client's `describe_log_groups`, `create_export_task`, `describe_export_tasks` are all called at least once

As of right now, I'm not sure how I would verify a user has permission to make those calls. But I'll look into that today.

util/retrieve-cluster-logs.py
tilne: Honestly I'd prefer to specify it as a kwarg. Do we usually set the env? If so I'll change it.

util/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,213 @@
+import gzip
```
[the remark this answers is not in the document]
tilne: Yeah that's makes sense. I'll do that.

util/retrieve-cluster-logs.py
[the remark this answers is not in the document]
tilne: I prefer to use a formatted string as opposed to an epoch timestamp, but I'm not sure if that's what you meant. Does the ISO 8601 format work:
```
>>> datetime.now().isoformat()
'2019-11-11T10:24:47.899582'
```

util/retrieve-cluster-logs.py
tilne: Ah, I didn't think about the consistency tests. I'll add one.

tilne: Actually after looking at those tests I don't think it goes well there since the name is not passed as a parameter. I'll omit a test for now.

util/retrieve-cluster-logs.py
tilne: This did not work in us-east-1. There, the `LocationRestraint` key in the JSON returned by `get_bucket_location` was None instead of the region. Do you know why that is? Is that something I need to worry about for any of the other regions?

tilne: Never mind my previous comment. I found the information here: https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

util/retrieve-cluster-logs.py
tilne: I thought the same thing, but it works to pass the start and end times around as timezone unaware `datetime`s and then convert them to a timestamp as they are passed into `create_export_task`.

The `datetime`'s `timestamp` method adjusts its return value based on the instance's timezone information, so passing around the UTC versions of the objects without explicitly attaching that timezone information results in times that are 9 hours later than intended when run from a PST locale.

util/retrieve-cluster-logs.py
tilne: I can just add them real quick.

util/retrieve-cluster-logs.py
tilne: To circle back on my response to the general comment: I looked into what it would take to figure out whether or not a user has the permissions to perform all of the actions listed above.

I think it would be possible to use `iam.simulate_principal_policy` function to do this, but I don't necessarily think the work required to do this well is worth the benfit it gives us over just letting the exceptions bubble up when they occur, given that the error messages tend to be pretty informative for AccessDenied issues.

demartinofra: Yes for us-east-1 things are always handled differently there.

For the policy/permissions I think it's enough for now just to be clear in the error message when the script fails due to missing permissions.

util/retrieve-cluster-logs.py
demartinofra: I'd say if we decide to queue additional `create_export_task`s then let's make sure the script waits for it to be completed doesn't fail because there are multiple tasks in the queue. Alternatively we could just simplify the approach for now and fail fast if a task is already in the queue.

A future enhancement might be to also SIGINT and perform a cleanhup when this occurs. But we can leave this for later.

util/retrieve-cluster-logs.py
demartinofra: In the CLI we are transitioning to this approach since passing region, proxy and credentials args to all functions calling boto3 was probably not the best approach. At the same time you might forget to set this parameters when initializing a boto3 client. 
Of course at the same time you might argue that this hides the fact that this boto3 calls are using specific parameters loaded from the environment.

Feel free to leave it as is here if you find this approach reasonable for this use case.

util/retrieve-cluster-logs.py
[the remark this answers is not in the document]
demartinofra: Yes I just meant to use some delimiters to identify the different sections of the datetime. ISO format works for me as long as `-:.` chars are allowed by S3

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -21,6 +21,7 @@
 import pytest
 
 from assertpy import assert_that
+from pcluster.cloudwatch_logs.utils import get_log_group_name
```
demartinofra: I would allow to create dependencies between integration tests and the code defined in the cli. One of the reasons being we can move these tests in a separate package and the other being that sometimes we want to use the test from the develop branch and run them against an older version of the cli deployed through PyPI which might not contain the required dependencies.

(p.s. same needs to be done for the dcv test in case you want to open a quick patch :D)

util/logging/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,271 @@
...
+import sys
+import tarfile
+import tempfile
+import time
+from datetime import datetime
+
+import argparse
+import boto3
+
+from pcluster.cloudwatch_logs.utils import get_log_group_name
```
demartinofra: Same here. As long as this is a util script let's drop any dependency on the cli code.

util/logging/retrieve-cluster-logs.py
```diff
@@ -0,0 +1,271 @@
...
+
+def err_and_exit(message):
+    """Log the given error message and exit nonzero."""
+    LOGGER.error(message)
+    sys.exit(1)
+
+
+def get_default_bucket_prefix(cluster_name):
+    """Get default keypath to store logs under in an s3 bucket for the given cluster."""
+    return "{cluster_name}-logs-{end_time}".format(cluster_name=cluster_name, end_time=datetime.now().isoformat())
```
demartinofra: what about this datetime? Shall we have it always in utc and not in local time?

util/logging/retrieve-cluster-logs.py
tilne: Done.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: Done. And noted ;).

util/logging/retrieve-cluster-logs.py
tilne: How about I just make it a timestamp as well? The main point of including it there was just to make sure that the output file. `end_time` is a bit misleading in that way.

tilne: That's done.

util/retrieve-cluster-logs.py
tilne: The script currently waits for the task to start and complete, so I'll leave it as is for now unless you feel strongly about changing it. I added a comment at the top of the script with a TODO for catching the SIGINT and cleaning up appropriately.

demartinofra:
Nice job! 👍

PR #1444, opened 2019-11-11
Skip g3 in non-GPU tests

rexcsn:
* Only need to run test_slurm_gpu and DCV tests with g3.8xlarge
* Adding skip g3.8xlarge to all other tests so that they are not run with GPU instance
* Adding markers to DCV tests to run on all schedulers

### DIFF
--- a/tests/integration-tests/tests/dcv/test_dcv.py
+++ b/tests/integration-tests/tests/dcv/test_dcv.py
@@ -26,7 +26,8 @@ DCV_CONNECT_SCRIPT = "/opt/parallelcluster/scripts/pcluster_dcv_connect.sh"
 )
 @pytest.mark.regions(["eu-west-1", "cn-northwest-1"])  # DCV license bucket not present in us-gov
 @pytest.mark.oss(["centos7"])
-@pytest.mark.schedulers(["sge", "awsbatch"])
+@pytest.mark.instances(["c4.xlarge", "g3.8xlarge"])
+@pytest.mark.schedulers(["sge"])
 def test_dcv_configuration(
     dcv_port, access_from, shared_dir, region, instance, os, scheduler, pcluster_config_reader, clusters_factory
 ):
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -25,8 +25,9 @@
     "dcv_port, access_from, shared_dir", [(8443, "0.0.0.0/0", "/shared"), (5678, "192.168.1.1/32", "/myshared")]
 )
 @pytest.mark.regions(["eu-west-1", "cn-northwest-1"])  # DCV license bucket not present in us-gov
-@pytest.mark.oss(["centos7"])
-@pytest.mark.schedulers(["sge", "awsbatch"])
+@pytest.mark.oss(get_supported_dcv_os())
```
enrico-usai: You need to rebase the code, let's use "centos7" and not the "get_supported_dcv_os" function to split code and tests

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -25,8 +25,9 @@
...
+@pytest.mark.instances(["c4.xlarge", "g3.8xlarge"])
```
enrico-usai: Probably we should add the g4 in place of the g3 since it is the one that requires a specific driver to work (the same is valid for slurm) what do you think?

rexcsn: Some of the slurm_gpu tests are specific to `g3.8xlarge`(32 cpus/2 gpus), there is no instance in the g4 family with the same number of cpu/gpu. To change to g4 I will have to rewrite the logic we are testing in slurm_gpu tests. I would suggest to change to g4 later in a separate PR.

tests/integration-tests/tests/dcv/test_dcv.py
rexcsn: Done.

PR #1451, opened 2019-11-12
Improve EC2 instance types validator to get the list from pricing file

enrico-usai:
The boto3 metadata are not updated, it is required to upgrade botocore library to see the newest instances. The pricing file is a list specific for the product.

From now the `compute_instance_type_validator` automatically calls `ec2_instance_type_validator` if the scheduler is not awsbatch.

I'm using two different methods to retrieve compute and master instance types because compute instance type list must contain also families and "optimal" keyword in case of awsbatch scheduler.

+ Remove useless parameter from `cluster_section_to_file` test

#### awsbatch, compute_instance_type
- t2.micro --> compute_instance_type 't2.micro' is not supported by awsbatch in region 'eu-west-1'
- t2 --> compute_instance_type 't2' is not supported by awsbatch in region 'eu-west-1'
- test --> compute_instance_type 'test' is not supported by awsbatch in region 'eu-west-1'
- c4 --> ok
- c4.large --> ok
#### awsbatch, master_instance_type
- c4 --> The instance type 'c4' used for the 'master_instance_type' parameter is not supported by AWS ParallelCluster.
- t2.micro --> ok
- c4.large --> ok
#### sge, *_instance_type
- c4 --> The instance type 'c4' used for the 'master_instance_type' parameter is not supported by AWS ParallelCluster.
- c4 --> The instance type 'c4' used for the 'compute_instance_type' parameter is not supported by AWS ParallelCluster.
- test --> The instance type 'test' used for the 'master_instance_type' parameter is not supported by AWS ParallelCluster.
- test --> The instance type 'test' used for the 'compute_instance_type' parameter is not supported by AWS ParallelCluster.
- t2.micro --> ok
- c4.large --> ok

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -22,7 +22,9 @@ from pcluster.utils import (
     get_instance_vcpus,
     get_partition,
     get_region,
+    get_supported_compute_instance_types,
     get_supported_features,
+    get_supported_instance_types,
 )
 
 
@@ -420,6 +422,19 @@ def _get_pcluster_user_policy(partition, region, account_id):
     ]
 
 
+def ec2_instance_type_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if param_value not in get_supported_instance_types():
+        errors.append(
+            "The instance type '{0}' used for the '{1}' parameter is not supported by AWS ParallelCluster.".format(
+                param_value, param_key
+            )
+        )
+    return errors, warnings
+
+
 def ec2_vpc_id_validator(param_key, param_value, pcluster_config):
     errors = []
     warnings = []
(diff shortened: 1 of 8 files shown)


cli/pcluster/config/validators.py
```diff
@@ -421,7 +421,7 @@ def ec2_instance_type_validator(param_key, param_value, pcluster_config):
     warnings = []
 
     try:
-        if param_value not in list_ec2_instance_types():
+        if param_value not in get_supported_instance_types():
```
demartinofra: get_supported_instance_types is exiting on error. Are we ok with that?

cli/pcluster/config/validators.py
```diff
@@ -421,7 +421,7 @@ def ec2_instance_type_validator(param_key, param_value, pcluster_config):
...
             errors.append(
                 "The instance type '{0}' used for the '{1}' parameter is not supported by EC2.".format(
```
demartinofra: this probably should become: it's not supported by AWS ParallelCluster

enrico-usai: Fixed, thanks

cli/pcluster/config/validators.py
```diff
@@ -420,6 +422,24 @@ def _get_pcluster_user_policy(partition, region, account_id):
...
+
+    try:
+        if param_value not in get_supported_instance_types():
+            errors.append(
+                "The instance type '{0}' used for the '{1}' parameter is not supported by AWS ParallelCluster.".format(
+                    param_value, param_key
+                )
+            )
+
+    except ClientError as e:
```
demartinofra: get_supported_instance_types is exiting with error on ClientError so this code can never be reached.

cli/pcluster/config/validators.py
```diff
@@ -672,14 +692,15 @@ def cluster_validator(section_key, section_label, pcluster_config):
...
     warnings = []
 
     cluster_config = pcluster_config.get_section("cluster")
-    if cluster_config.get_param_value("scheduler") == "awsbatch":
-
+    scheduler = cluster_config.get_param_value("scheduler")
+    if scheduler == "awsbatch":
         try:
-            supported_instances = get_supported_features(pcluster_config.region, "batch").get("instances")
+            supported_instances = get_supported_compute_instance_types(scheduler)
```
demartinofra: same happens here with get_supported_compute_instance_types exiting on ClientError

cli/pcluster/config/validators.py
enrico-usai: Good catch! resolved

PR #1458, opened 2019-11-13
Add possibility to add suffix to integration tests stack names

lukeseawalker:
Default is to use the username of the user launching the tests suite

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -74,6 +74,7 @@ def pytest_addoption(parser):
     )
     parser.addoption("--benchmarks-target-capacity", help="set the target capacity for benchmarks tests", type=int)
     parser.addoption("--benchmarks-max-time", help="set the max waiting time in minutes for benchmarks tests", type=int)
+    parser.addoption("--stackname-suffix", help="set a suffix in the integration tests stack names")
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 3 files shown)


tests/integration-tests/conftest.py
```diff
@@ -189,7 +190,7 @@ def _cluster_factory(cluster_config):
         cluster = Cluster(
             name=request.config.getoption("cluster")
             if request.config.getoption("cluster")
-            else "integ-tests-" + random_alphanumeric(),
+            else "integ-tests-" + random_alphanumeric() + request.config.getoption("stackname_suffix"),
```
demartinofra: minor: I think we should add a `-` here before concatenating the suffix

tests/integration-tests/test_runner.py
```diff
@@ -72,6 +73,7 @@
     "benchmarks": False,
     "benchmarks_target_capacity": 200,
     "benchmarks_max_time": 30,
+    "stackname_suffix": "-" + getpass.getuser(),
```
demartinofra: if it doesn't add much value I'd rather leave an empty suffix as default

tests/integration-tests/test_runner.py
```diff
@@ -244,6 +251,7 @@ def _get_pytest_args(args, regions, log_file, out_dir):
     pytest_args.extend(["--output-dir", "{0}/{1}".format(args.output_dir, out_dir)])
     pytest_args.extend(["--key-name", args.key_name])
     pytest_args.extend(["--key-path", args.key_path])
+    pytest_args.append("--stackname-suffix={0}".format(args.stackname_suffix))
```
demartinofra: minor: just for consistency we can use the extend and an array as in the previous line

tests/integration-tests/conftest.py
lukeseawalker: I think the suffix should contain the - 
Let's discuss it

tests/integration-tests/test_runner.py
lukeseawalker: If you extend the array like in the previous one and the suffix contains a -, it will be seen as a parameter

tests/integration-tests/test_runner.py
lukeseawalker: ok, this will impact the other comments

PR #1459, opened 2019-11-13
Remove ec2 instance type validator

enrico-usai:
We are removing it since it is based on botocore metadata.


sean-smith:
Closing in favor of https://github.com/aws/aws-parallelcluster/pull/1460

PR #1462, opened 2019-11-14
Handle NoSectionError directly from PclusterConfig

ddeidda:

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -10,7 +10,6 @@
 # limitations under the License.
 from future.moves.collections import OrderedDict
 
-import functools
 import json
 import logging
 import re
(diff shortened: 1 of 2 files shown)


cli/pcluster/config/param_types.py
```diff
@@ -34,10 +33,11 @@
 def handle_no_section_error(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
+        param = args[0]  # "self" argument
```
demartinofra: I'm afraid this is not safe cause the first arg is not always self but it depends how you invoke this function.
Why don't we move this functions outside of the PclusterConfig class?

ddeidda: The wrapper has been totally removed now; the exception is handled directly from PClusterConfig

PR #1475, opened 2019-11-18
Perform checks and do m5sum before uploading cookbook file

ddeidda:
This commit applies the following changes:
- Check for archive file existence before going on even in dryrun mode
- Do md5sum directly inside python script
- Receive the more informative "cookbook-archive-file" param instead
of the old "full_name" one

*Issue #, if available:*

*Description of changes:*


ddeidda:
> The push to S3 doesn't work if the archive is not in the same path from where the script is called

Fixed

PR #1488, opened 2019-11-22
Fix CloudWatch awsbatch integration tests

tilne:
The tests did not account for the fact that log data is not collected
from the "computes" when AWS Batch is the scheduler. This PR fixes that.

### DIFF
--- a/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
+++ b/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
@@ -19,6 +19,7 @@ from pathlib import Path
 
 import boto3
 import pytest
+from retrying import retry
 
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -149,23 +158,31 @@ def _clean_log_config(log):
...
-        logs = self._read_log_configs_from_master()
+    def _filter_logs(self, logs):
+        """Populate self._relevant_logs with logs appropriate for the two different node types."""
+        # Batch doesn't really have compute nodes in the traditional sense.
+        desired_node_roles = {"MasterServer"} if self.scheduler == "awsbatch" else {"MasterServer", "ComputeFleet"}
         for log in logs:
             if self.scheduler not in log.get("schedulers") or self.platform not in log.get("platforms"):
                 continue
-            for node_role in log.get("node_roles"):
+            for node_role in set(log.get("node_roles")) & desired_node_roles:
```
rexcsn: Minor: is the `ComputeFleet` node_role in `cloudwatch_log_files.json` used when running a cluster with awsbatch? If not should we remove the role?

tilne: On this line, `log` represents a JSON object with this format:
```
{
      "timestamp_format_key": "default",
      "file_path": "/var/log/cfn-init-cmd.log",
      "log_stream_name": "cfn-init-cmd",
      "schedulers": [
        "awsbatch",
        "sge",
        "slurm",
        "torque"
      ],
      "platforms": [
        "amazon",
        "centos",
        "ubuntu"
      ],
      "node_roles": [
        "ComputeFleet",
        "MasterServer"
      ]
}
```

`desired_node_roles` is a set that only contains `MasterServer` when batch is the scheduler. I think that has the effect that you're getting at with your question.

Removing it from the log entry in the data file itself wouldn't make sense because it is a log that would be present on compute nodes for traditional schedulers.

rexcsn: I would suggest to add a comment like, "ComputeFleet role is defined in log config for awsbatch, currently not used, hence skipping tests for ComputeFleet role when running with awsbatch"

PR #1490, opened 2019-11-22
Fix invalid spot_price default value

ddeidda:
This commit makes default value for SpotPriceParam to be "0" replacing
"0.0" which prevented the no-value condition to work properly and caused
the invalid 0.0 value to be passed to CloudFormation

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -495,14 +495,21 @@ class SpotPriceParam(FloatParam):
 
         return self
 
+    def get_cfn_value(self):
+        """
+        Convert parameter value into CFN value.
+
+        Insignificant trailing zeros removed to correctly match CloudFormation conditions using "0" as test value
+        """
+        return str("{0:g}".format(self.value) if self.value is not None else self.definition.get("default", "NONE"))
+
     def to_cfn(self):
         """Convert parameter to CFN representation."""
         cfn_params = {}
 
         cluster_config = self.pcluster_config.get_section(self.section_key)
         if cluster_config.get_param_value("scheduler") != "awsbatch":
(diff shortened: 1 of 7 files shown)


cli/tests/pcluster/config/test_param_from_cfn.py
```diff
@@ -64,7 +64,7 @@
         (CLUSTER, "spot_price", "10", 10),
         (CLUSTER, "spot_price", "3", 3),
         (CLUSTER, "spot_price", "0.0009", 0.0009),
-        (CLUSTER, "spot_price", "0.0", 0.0),
+        (CLUSTER, "spot_price", "0", 0.0),
```
enrico-usai: I'd add also a test for 0.0 as input:
```python
(CLUSTER, "spot_price", "0.0", 0.0),
```

cli/pcluster/config/param_types.py
```diff
@@ -495,14 +495,21 @@ def from_cfn_params(self, cfn_params):
 
         return self
 
+    def get_cfn_value(self):
+        """
+        Convert parameter value into CFN value.
+
+        Insignificant trailing zeros removed to correctly match UseSpotPrice with default value
```
enrico-usai: I'd change the comment with:
```
Insignificant trailing zeros removed to simplify comparison.
```

cli/tests/pcluster/config/test_param_from_cfn.py
ddeidda: OK :+1:

cli/pcluster/config/param_types.py
ddeidda: Updated with: "Insignificant trailing zeros removed to correctly match CloudFormation conditions using "0" as test value"

PR #1496, opened 2019-11-26
Add config option for passing Spot block duration to the compute autoscaling group

isobit:
Currently there is no way to specify a Spot block duration for the compute autoscaling group to use when launching spot instances. Having the ability to pass this option would improve the experience of using Spot requests with parallelcluster, since Spot instance interruptions can be detrimental when trying to run jobs in a timely manner.


demartinofra:
Thank you for the very valid contribution!

Before going ahead and merging this change we would like to better understand how to support such feature in ParallelCluster. Having a static spot duration defined in the LaunchTemplate can be a starting point but I'd say it's not a perfect fit for the current scaling logic of PC which allows to share spot nodes across several jobs or to reuse spot nodes for subsequent runs. Reusing or sharing a spot node with a given block duration might be somehow misleading and result in unexpected termination of nodes.
An alternative approach might be to reserve spot nodes with a given block duration to individual jobs and prevent these from being re-used. Of course this change requires major changes to the current scaling logic.

demartinofra:
From https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances

Spot Instances with a defined duration (also known as Spot blocks) are no longer available to new customers from July 1, 2021. For customers who have previously used the feature, we will continue to support Spot Instances with a defined duration until December 31, 2022.

PR #1497, opened 2019-11-27
Fix mounting EBS to /NONE

rexcsn:
* Fixed `shared_dir` logic. The complete logic: if `ebs_settings` is not specified or if only using 1 volume and `shared_dir` is not specified under EBS section, use `shared_dir` from cluster section. Otherwise always use `shared_dir` specified under EBS section(s)
* Added validator to check limits on maximum number of EBS volumes, and to check that `shared_dir` is specified under each EBS section when using more than 1 EBS volume
* Added test cases in `test_section_cluster` to test the above mentioned cases


cli/tests/pcluster/config/test_section_cluster/test_cluster_from_file_to_cfn/pcluster.config.ini
```diff
@@ -148,8 +148,29 @@ ebs_settings = ebs1
 # Test: multiple ebs settings
 [cluster ebs-multiple]
 vpc_settings = default
+shared_dir = /work
 ebs_settings = ebs1,ebs2
```
demartinofra: what happens when multiple EBS sections are used but one of them does not declare shared_dir? Can we add a test for this?

enrico-usai:
I'd also add an improvement to the cookbook, to avoid to mount when the shared dir is NONE or /NONE.

cli/tests/pcluster/config/test_section_cluster/test_cluster_from_file_to_cfn/pcluster.config.ini
rexcsn: That should be caught with `ebs_settings_validator`, I will look today to see how we can test this.

cli/pcluster/config/validators.py
```diff
@@ -531,6 +531,37 @@ def ec2_ebs_snapshot_validator(param_key, param_value, pcluster_config):
     return errors, warnings
 
 
+def ebs_settings_validator(param_key, param_value, pcluster_config):
```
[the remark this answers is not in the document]
rexcsn: `test_ebs_settings_validator` added in `test_validators.py`

cli/pcluster/config/validators.py
```diff
@@ -531,6 +531,49 @@ def ec2_ebs_snapshot_validator(param_key, param_value, pcluster_config):
...
+
+    return errors, warnings
+
+
+def shared_dir_validator(param_key, param_value, pcluster_config):
+    """Validate that user is not specifying /NONE or NONE as shared_dir for any filesystem."""
+    errors = []
+    warnings = []
+
+    if param_value == "/NONE" or param_value == "NONE":
```
enrico-usai: you can use a regex and print the param_value instead of a fixed string:
```python
if re.match("/?NONE"):
    errors.append("{0} cannot be used as a shared directory".format(param_value))
```

cli/tests/pcluster/config/test_validators.py
```diff
@@ -400,6 +400,8 @@ def test_ec2_security_group_validator(mocker, boto3_stubber):
             "When specifying 'provisioned_throughput', the 'throughput_mode' must be set to 'provisioned'",
         ),
         ({"throughput_mode": "provisioned", "provisioned_throughput": 1024}, None),
+        ({"shared_dir": "NONE"}, "/NONE cannot be used as a shared directory"),
```
enrico-usai: Let's align to the previous comment

cli/tests/pcluster/config/test_validators.py
```diff
@@ -617,3 +625,57 @@ def test_efa_validator_with_vpc_security_group(
...
+        for vol in ebs_section_dict:
+            config_parser_dict["ebs {0}".format(vol)] = ebs_section_dict.get(vol)
+    utils.assert_param_validator(mocker, config_parser_dict, expected_message)
+
+
+@pytest.mark.parametrize(
+    "section_dict, expected_message",
+    [
+        ({"shared_dir": "NONE"}, "/NONE cannot be used as a shared directory"),
+        ({"shared_dir": "/NONE"}, "/NONE cannot be used as a shared directory"),
```
enrico-usai: Let's also add a test for a valid value.

cli/pcluster/config/validators.py
rexcsn: In this case do we want to use `re.fullmatch("/?NONE", param_value)`? `re.match("/?NONE", param_value)` will also match strings starting with `/?NONE` like `/NONEshared`

enrico-usai: let's use `fullmatch` 👍

rexcsn: `fullmatch` not compatible with Python 2.7, using `re.match("^/?NONE$", param_value)`

PR #1503, opened 2019-11-27
Handle feature-specific logs in CW integ tests

tilne:
[This cookbook PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/442) adds DCV logs to the config file used to generate the
CloudWatch agent configuration. This same config file is used by the
CloudWatch logs integration tests to generate the list of logs which are
expected to (a) exist on the various nodes in a cluster and (b) be
represented in the log streams present in a cluster's log group (if the
log file on the node is not empty).

This PR ensures that feature-specific logs do not result in test
failures if those logs do not exist. The reason for this behavior is to
avoid duplicating functionality that should be present in the
integration tests specific to the same feature as the logs.

As an example: some of the DCV logs for which support was added are not
created until `pcluster dcv connect` is called to start a session. It's
preferable to keep the logic that exercises this in the DCV integration
tests.

The CloudWatch logs integration tests will be modified to allow other
integration tests to verify that the logs they expect to be collected
are, but that can be added later.

Note that feature-specific logs are still validated to an extent. If
they are present and not empty, their log streams are validated per
usual.

### DIFF
--- a/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
+++ b/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
@@ -155,7 +155,7 @@ class CloudWatchLoggingClusterState:
     @staticmethod
     def _clean_log_config(log):
         """Remove unnecessary fields from the given log dict."""
-        desired_keys = ["file_path", "log_stream_name"]
+        desired_keys = ["file_path", "log_stream_name", "feature_conditions"]
         return {key: log[key] for key in desired_keys}
 
     def _filter_logs(self, logs):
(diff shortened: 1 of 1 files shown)


sean-smith:
I think we should have some sort of test for these logs.

When you say:
> The CloudWatch logs integration tests will be modified to allow other
integration tests to verify that the logs they expect to be collected
are, but that can be added later.

How much effort is that to implement?

tilne:
> I think we should have some sort of test for these logs.
> 
> When you say:
> 
> > The CloudWatch logs integration tests will be modified to allow other
> > integration tests to verify that the logs they expect to be collected
> > are, but that can be added later.
> 
> How much effort is that to implement?

A couple of days worth. The reason I didn't want to include it in this one is because I had some other changes already in flight for this same file. I'd like to include them before making the changes I'll need to for other tests to hook into this one.

Issue #166, opened 2017-02-12
Instances command raises exception if cluster rollbacks


snorfalorpagus:
If the instances command is called after the cluster has died an exception is raised, rather than something useful.
```
$ cfncluster status mycluster
Status: ROLLBACK_COMPLETE
2017-02-12 22:13:30.189000 CREATE_FAILED AWS::AutoScaling::AutoScalingGroup ComputeFleet Received 1 FAILURE signal(s) out of 1.  Unable to satisfy 100% MinSuccessfulInstancesPercent requirement
$ cfncluster instances mycluster
Traceback (most recent call last):
  File "/Users/snorf/miniconda3/bin/cfncluster", line 9, in <module>
    load_entry_point('cfncluster', 'console_scripts', 'cfncluster')()
  File "/Users/snorf/Desktop/cfncluster/cli/cfncluster/cli.py", line 160, in main
    args.func(args)
  File "/Users/snorf/Desktop/cfncluster/cli/cfncluster/cli.py", line 38, in instances
    cfncluster.instances(args)
  File "/Users/snorf/Desktop/cfncluster/cli/cfncluster/cfncluster.py", line 406, in instances
    instances.extend(get_asg_instances(stack, config))
  File "/Users/snorf/Desktop/cfncluster/cli/cfncluster/cfncluster.py", line 394, in get_asg_instances
    instances = asgconn.get_all_groups(names=[asg[1]])[0].instances
```

jocherry:
So just to clarify, it would be helpful if the instance call on a cluster with no instances returned a helpful message, as opposed to a stack trace?

snorfalorpagus:
@jocherry Yes. I don't think the user of the CLI should even need to know cfncluster is written in Python. The command should return "no instances", and perhaps "check status?". I say this because, unlike the example above, I called the instances command first. It took me a little while to figure out the cluster hadn't started, rather than it being a bug in cfncluster per-say.

demartinofra:
Fixed in https://github.com/aws/aws-parallelcluster/releases/tag/v2.5.1

PR #1506, opened 2019-11-28
pcluster instances: handle ASG not found error

demartinofra:
Better handle `pcluster instances` errors when ASG is not found

This fixes https://github.com/aws/aws-parallelcluster/issues/166

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -486,7 +486,11 @@ def _set_asg_limits(asg_name, min, max, desired):
 def _get_asg_instances(stack):
     asg = boto3.client("autoscaling")
     asg_name = _get_asg_name(stack)
-    asg = asg.describe_auto_scaling_groups(AutoScalingGroupNames=[asg_name]).get("AutoScalingGroups")[0]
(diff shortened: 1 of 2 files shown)


cli/pcluster/commands.py
```diff
@@ -486,7 +486,11 @@ def _set_asg_limits(asg_name, min, max, desired):
 def _get_asg_instances(stack):
     asg = boto3.client("autoscaling")
     asg_name = _get_asg_name(stack)
-    asg = asg.describe_auto_scaling_groups(AutoScalingGroupNames=[asg_name]).get("AutoScalingGroups")[0]
+    auto_scaling_groups = asg.describe_auto_scaling_groups(AutoScalingGroupNames=[asg_name]).get("AutoScalingGroups")
+    if not auto_scaling_groups:
+        LOGGER.error("Unable to retrieve ASG info. Please check cluster status.")
```
enrico-usai: Should we use LOGGER.critical ?

demartinofra: I see critical to be more appropriate for unexpected exceptions/unhandled exceptions. In this case we know that probably the cluster is being deleted or failed to create and that is why we are recommending to check its status. What do you think?

enrico-usai: ok, I agree.

PR #1516, opened 2019-12-04
Enable validation of dockerfile entry point deps

tilne:
This PR enables more thorough validation of the dependencies (and
functionality) of the docker image generated when a cluster uses
awsbatch as its scheduler. It does this via two changes:
1. The `batch_dockerfile_deps` mark has been added to the awsbatch
scheduler test function, in order to allow for selecting only this test.
2. An EFS filesystem is configured in the awsbatch scheduler test
function's config file, in order to exercies the `mount_efs.sh` entry
point script in the generated docker image.

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_awsbatch.py
+++ b/tests/integration-tests/tests/schedulers/test_awsbatch.py
@@ -18,6 +18,7 @@ from remote_command_executor import RemoteCommandExecutor
 from tests.common.schedulers_common import AWSBatchCommands
 
 
+@pytest.mark.batch_dockerfile_deps
 @pytest.mark.skip_regions(["ap-northeast-3", "us-gov-east-1", "us-gov-west-1"])
 @pytest.mark.instances(["c5.xlarge", "t2.large"])
 @pytest.mark.dimensions("*", "*", "alinux", "awsbatch")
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/schedulers/test_awsbatch.py
```diff
@@ -18,6 +18,7 @@
 from tests.common.schedulers_common import AWSBatchCommands
 
 
+@pytest.mark.batch_dockerfile_deps
```
demartinofra: An awsbatch mark is already added to this function by the testing framework (default markers are based on the filename). I'm ok with keeping this here so that if additional functions are added this is the only one being invoked for this kind of tests

tilne: Doesn't using the awsbatch mark result in running all of the tests that support that scheduler?

demartinofra: nope..a default marker is added based on the name of the file defining the test functions. E.g. test_awsbatch.py -> all functions have a default `@pytest.mark.awsbatch` marker

tilne: Okay. I suppose I'll keep it in case we need to add tests for this in different files in the future.

PR #1522, opened 2019-12-05
Integ tests minor improvements

demartinofra:
- On cluster teardown do not fail on delete if cluster does not exist (e.g. if create already failed before cluster got created)
- Add the name of the test function to the logging messages so that it's easier to retrieve all logs related to a specific failure
- improve assertions and messaging in scaling tests

### DIFF
--- a/tests/integration-tests/tests/common/assertions.py
+++ b/tests/integration-tests/tests/common/assertions.py
@@ -10,7 +10,7 @@
 # limitations under the License.
 import boto3
 
-from assertpy import assert_that
+from assertpy import assert_that, soft_assertions
 from tests.common.scaling_common import get_compute_nodes_allocation
 from time_utils import minutes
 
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/common/assertions.py
```diff
@@ -63,9 +63,19 @@ def assert_scaling_worked(
...
         + minutes(max_scaledown_time),
     )
-    if assert_asg:
-        assert_that(max(asg_capacity_time_series)).is_equal_to(expected_max)
-        assert_that(asg_capacity_time_series[-1]).is_equal_to(expected_final)
-    if assert_scheduler:
-        assert_that(max(compute_nodes_time_series)).is_equal_to(expected_max)
-        assert_that(compute_nodes_time_series[-1]).is_equal_to(expected_final)
+
+    with soft_assertions():
```
enrico-usai: I like soft_assertions 👍

tests/integration-tests/tests/common/assertions.py
```diff
@@ -63,9 +63,19 @@ def assert_scaling_worked(
...
+        if assert_asg:
+            assert_that(max(asg_capacity_time_series)).described_as(
+                f"asg_capacity_time_series={asg_capacity_time_series}"
```
enrico-usai: Is it right you are using the same description for the two expected_max and expected_final comparison?

If yes we could set a variable for the description (`f"asg_capacity_time_series={asg_capacity_time_series}"`) and then use it for the two asserts. The same below for `f"compute_nodes_time_series={compute_nodes_time_series}"`

PR #1523, opened 2019-12-05
Add test for multi CIDR VPC

rexcsn:
* Added test to test that a cluster is created when using VPC with 2 CIDR and compute subnet in the 2nd CIDR block of VPC.
* Testing changes made in https://github.com/aws/aws-parallelcluster-cookbook/pull/449. Cluster creation will fail without this change

### DIFF
--- a/tests/integration-tests/tests/networking/test_multi_cidr.py
+++ b/tests/integration-tests/tests/networking/test_multi_cidr.py
@@ -1,0 +1,15 @@
+import pytest
+
+
+@pytest.mark.regions(["us-east-2"])
+@pytest.mark.instances(["c5.xlarge"])
+@pytest.mark.schedulers(["slurm", "awsbatch"])
+@pytest.mark.usefixtures("os", "instance", "scheduler", "region")
+def test_multi_cidr(pcluster_config_reader, clusters_factory):
+    """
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/networking/test_multi_cidr.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+
+@pytest.mark.regions(["us-east-2"])
+@pytest.mark.instances(["t2.micro", "c5.xlarge"])
+@pytest.mark.schedulers(["slurm", "awsbatch"])
+@pytest.mark.usefixtures("os", "instance", "scheduler")
+def test_multi_cidr(region, pcluster_config_reader, clusters_factory):
```
demartinofra: minor: you can move region to the usefixtures annotation

tests/integration-tests/tests/networking/test_multi_cidr.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+
+@pytest.mark.regions(["us-east-2"])
+@pytest.mark.instances(["t2.micro", "c5.xlarge"])
```
demartinofra: we could leave only c5.xlarge

PR #1524, opened 2019-12-05
Fix easy config not preserving initial_queue_size and max_queue_size

rexcsn:
* Added string casting to get_value_from_string calls to avoid error
* Added error message in get_value_from_string to make user aware that the parameter is not set when the input is not a string
* Changed logic in test_pcluster_configure to test that the expected config is a subset of resulting config after running pcluster configure. Before this change we were testing the other way around and we would never know what the resulting config actually is.
* Added test_no_input_no_automation_no_error_with_config_file and test_with_input_no_automation_no_error_with_config_file to test that resulting config is as expected
* Changed majority of the test config files to adapt to above mentioned changes. Basically commenting out fields with default values, because these field will not be written into the config after we run pcluster configure.

### DIFF
--- a/cli/tests/pcluster/configure/test_pcluster_configure.py
+++ b/cli/tests/pcluster/configure/test_pcluster_configure.py
@@ -254,7 +254,6 @@ def test_no_automation_no_awsbatch_no_errors(mocker, capsys, test_datadir):
 
 
 def _run_input_test_with_config(mocker, config, old_config_file, error, output, capsys, with_input=False):
-    """Helper function to run input tests."""
     if with_input:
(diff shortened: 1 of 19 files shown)


cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -237,6 +253,46 @@ def test_no_automation_no_awsbatch_no_errors(mocker, capsys, test_datadir):
     _verify_test(mocker, capsys, output, error, config, TEMP_PATH_FOR_CONFIG)
 
 
+def test_no_input_no_automation_no_errors_with_config_file(mocker, capsys, test_datadir):
```
enrico-usai: this seems very similar to the second one, I think we can have a single parametrized test

PR #1531, opened 2019-12-09
Added logic to check existing config values are valid for easy config

rexcsn:
* Added checks for None in get_value_from_string
* Added logic to check that values from existing config file is still valid under the new partition/region/vpc setup
* Added unit test to test the case that the user switched partition and region, key_name, vpc_id, master/compute_subnet_id are not found in available list


cli/pcluster/configure/easyconfig.py
```diff
@@ -122,14 +122,20 @@ def configure(args):
     aws_region_name = prompt_iterable(
         "AWS Region ID",
         get_regions(),
-        default_value=pcluster_config.get_section("aws").get_param_value("aws_region_name"),
+        default_value=pcluster_config.get_section("aws").get_param_value("aws_region_name")
+        if pcluster_config.get_section("aws").get_param_value("aws_region_name") in get_regions()
```
enrico-usai: we should avoid to call get_regions two times since this query performs a boto3 call, I'd store default value and regions in variables.

```python
regions = get_regions()
default_region = pcluster_config.get_section("aws").get_param_value("aws_region_name")
aws_region_name = prompt_iterable(
        "AWS Region ID",
        regions,
        default_value=default_region if default_region in regions else None,
)
```

cli/pcluster/configure/easyconfig.py
```diff
@@ -122,14 +122,20 @@ def configure(args):
...
     )
     # Set provided region into os environment for suggestions and validations from here on
     os.environ["AWS_DEFAULT_REGION"] = aws_region_name
 
     # Get the key name from the current region, if any
     key_name = prompt_iterable(
-        "EC2 Key Pair Name", _get_keys(), default_value=cluster_section.get_param_value("key_name")
+        "EC2 Key Pair Name",
+        _get_keys(),
+        default_value=cluster_section.get_param_value("key_name")
```
enrico-usai: same here

cli/pcluster/configure/easyconfig.py
```diff
@@ -119,17 +119,21 @@ def configure(args):
...
-        default_value=pcluster_config.get_section("aws").get_param_value("aws_region_name"),
+        available_regions,
+        default_value=default_region if default_region in available_regions else None,
     )
     # Set provided region into os environment for suggestions and validations from here on
     os.environ["AWS_DEFAULT_REGION"] = aws_region_name
 
     # Get the key name from the current region, if any
+    available_keys = _get_keys()
+    default_keys = cluster_section.get_param_value("key_name")
```
enrico-usai: minor: default_key

cli/pcluster/configure/easyconfig.py
```diff
@@ -219,14 +228,25 @@ def _create_vpc_parameters(vpc_section, scheduler, min_subnet_size, automate_vpc
...
     )
 
+    default_compute_subnet = vpc_section.get_param_value("compute_subnet_id")
     compute_subnet_id = prompt_iterable(
         "Compute Subnet ID",
         subnet_list,
-        default_value=vpc_section.get_param_value("compute_subnet_id") or master_subnet_id,
+        default_value=(
+            default_compute_subnet
+            if default_compute_subnet in [subnet_entry[0] for subnet_entry in subnet_list]
```
enrico-usai: what about defining a `available_subnets= [subnet_entry[0] for subnet_entry in subnet_list]`
given that it is used in two places inside the _ask_for_subnets function?

cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -43,16 +43,46 @@ def _mock_aws_region(mocker):
     mocker.patch(EASYCONFIG + "get_regions", return_value=regions)
 
 
-def _mock_list_keys(mocker):
+def _mock_aws_china_region(mocker):
```
enrico-usai: instead of having two functions to mock the region we could define a dict and return the right list according to the partition value.

```python
def _mock_aws_region(mocker, partition="commercial"):
  regions = {
  "commercial" : [
        "eu-north-1",
        "ap-south-1",
         ...
    ] 
   "china" : [
        "cn-north-1",
        "cn-northwest-1",
         ...
    ] 
  }
  return regions.get(partition)
```

What do you think?

We can do the same for mock_list_keys, etc, instead of having a boolean flag.
It could be useful if in future we want to moke more partitions.

cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -21,60 +21,90 @@ def _mock_input(mocker, input_in_order):
...
-                ("vpc-23456789", "ParallelClusterVPC-20190624105051", "0 subnets inside"),
-                ("vpc-34567891", "default", "3 subnets inside"),
-                ("vpc-45678912", "ParallelClusterVPC-20190626095403", "1 subnets inside"),
-            ],
-            "vpc_subnets": {
-                "vpc-12345678": [
-                    ("subnet-12345678", "ParallelClusterPublicSubnet", "Subnet size: 256"),
-                    ("subnet-23456789", "ParallelClusterPrivateSubnet", "Subnet size: 4096"),
+        response_dict = {
+            "commercial": {
```
enrico-usai: you could add another "fake" or "empty" key to return the empty region response, and avoid to have another flag:

```python
_mock_list_vpcs_and_subnets(mocker, partition="commercial"):
 vpcs = {
    "commercial": {...},
    "china": {...},
    "fake|empty": {"vpc_list": [], "vpc_to_subnets": {}}

}
```

rexcsn: We are passing in partition when we call `MockHandler(mocker, partition="china")` which calls all the`_mock...` functions with `partition`, so we will need a third variable for the special case of empty vpc and subnets anyway.

PR #1538, opened 2019-12-10
Add tests for --show-url

sean-smith:

### DIFF
--- a/tests/integration-tests/tests/dcv/test_dcv.py
+++ b/tests/integration-tests/tests/dcv/test_dcv.py
@@ -16,6 +16,7 @@ import pytest
 
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
+from utils import run_command
 
 SERVER_URL = "https://localhost"
 DCV_CONNECT_SCRIPT = "/opt/parallelcluster/scripts/pcluster_dcv_connect.sh"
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -39,6 +40,13 @@ def test_dcv_configuration(
     # check configuration parameters
     _check_security_group(region, cluster, dcv_port, expected_cidr=access_from)
 
+    # dcv connect show url
+    result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"])
+    assert_that(result.stdout).matches(
+        r"Please use the following one-time URL in your browser within 30 seconds:\n"
+        r"https:\/\/(\b(?:\d{1,3}\.){3}\d{1,3}\b):(\d\d\d\d)\?authToken=(.*)"
```
enrico-usai: To verify the url is the right one we should use the `dcv_port` in the regex string or let's simplify `(\d\d\d\d)` with `\d+`, since the `dcv_port` is an input of the test, so in future we could add a test with a port with more or less digits.

sean-smith: I've changed to use `dcv_port`. Thanks!

Issue #1540, opened 2019-12-10
ERROR: The configuration parameter 'disable_hyperthreading' has an invalid value 'True'


PhilipMorrisJones:
**Environment:**
 - AWS ParallelCluster 2.5.0
 - OS: centos
 - Scheduler: SGE

**Bug description and how to reproduce:**

```
$ pcluster create -c config_250_noHT pmj-tue-test2
Beginning cluster creation for cluster: pmj-tue-test2
ERROR: The configuration parameter 'disable_hyperthreading' has an invalid value 'True'
cfn_scheduler_slots cannot be set in addition to disable_hyperthreading = true
Creating stack named: parallelcluster-pmj-tue-test2
Status: parallelcluster-pmj-tue-test2 - CREATE_COMPLETE                         
MasterPublicIP: 3.213.63.193
ClusterUser: centos
MasterPrivateIP: 10.0.0.114
```

I have a line in my config to turn hyperthreading off and I get an error message. The error says "true" but my config says "true"

It still seems to work, so the default config has c5.xlarge with 4 CPUs and all online

```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    2
Core(s) per socket:    2
Socket(s):             1
NUMA node(s):          1
```

and with the flag, I get just 2 online

```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                2
On-line CPU(s) list:   0,1
Thread(s) per core:    1
Core(s) per socket:    2
Socket(s):             1
NUMA node(s):          1
```

```
[cluster default]
vpc_settings = public
fsx_settings = fsxshared
base_os = centos7
enable_efa = compute
scaling_settings = custom
key_name = pmj_private_key
compute_instance_type = c5n.18xlarge
master_instance_type = c5.xlarge
extra_json = { "cluster" : { "cfn_scheduler_slots" : "cores" } }
s3_read_write_resource = *
cluster_type = ondemand
placement_group = DYNAMIC
placement = compute
max_queue_size = 30
initial_queue_size = 0
maintain_initial_size = true
disable_hyperthreading = true
```

enrico-usai:
Hi @PhilipMorrisJones,
the problem is explained in the second row of the error message:
[...]

PR #1543, opened 2019-12-10
Improve validation messages

enrico-usai:
We were showing the param.value in the error message, but it is different from the configuration value.

E.g. param.value is a Boolean (True|False), but the configuration value is true|false

Previous behaviour:
```
$ pcluster create testcluster -t testcluster
Beginning cluster creation for cluster: testcluster
ERROR: The configuration parameter 'disable_hyperthreading' has an invalid value 'True'
cfn_scheduler_slots cannot be set in addition to disable_hyperthreading = true
```

Current behaviour:
```
$ pcluster create testcluster -t testcluster
Beginning cluster creation for cluster: testcluster
ERROR: The configuration parameter 'disable_hyperthreading' generated the following errors:
cfn_scheduler_slots cannot be set in addition to disable_hyperthreading = true
```

See: https://github.com/aws/aws-parallelcluster/issues/1540

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -128,13 +128,14 @@ class Param(object):
                 errors, warnings = validation_func(self.key, self.value, self.pcluster_config)
                 if errors:
                     self.pcluster_config.error(
-                        "The configuration parameter '{0}' has an invalid value '{1}'\n"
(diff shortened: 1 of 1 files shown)


PhilipMorrisJones:
Thanks, looks much better.

PR #1554, opened 2019-12-12
Fix error message when running createami command without a config file in place

lukeseawalker:
Before the patch
```
pcluster createami -ai foo -os foo
Building AWS ParallelCluster AMI. This could take a while...

No custom AMI created
ERROR: Configuration file {0} not found.
You can copy a template from /Users/carrogu/.virtualenvs/parallelcluster-2.5.1/lib/python3.7/site-packages/pcluster/config/examples/config or execute the 'pcluster configure' command
```

After the patch
```
pcluster createami -ai foo -os foo
Building AWS ParallelCluster AMI. This could take a while...

No custom AMI created
ERROR: Configuration file /Users/carrogu/.parallelcluster/config not found.
You can execute the 'pcluster configure' command or see https://docs.aws.amazon.com/parallelcluster/latest/ug/configuration.html
```


cli/pcluster/config/pcluster_config.py
```diff
@@ -21,6 +21,7 @@
 import configparser
 from botocore.exceptions import ClientError
 
+import pcluster
```
demartinofra: do we need to import pcluster? if so can we import it locally if the error is triggered?

lukeseawalker: I have changed the output string and the import it's not needed anymore

PR #1557, opened 2019-12-12
Improve dcv connect command by retrying in case of errors

enrico-usai:
* Fix error message when unable to get ip address
* Add retry to the `retrieve_dcv_session_url` function. It could happen that the dcv session hasn't started yet so we have to retry.
* Show url message when `webbrowser.open_new` returns `False`

### Previous behaviour:
```
$ pcluster dcv connect testenv --key-path key
ERROR: Something went wrong during DCV connection.
ERROR: Unable to obtain the Request Token from the NICE DCV external authenticator.. Wrong JSON.
```

### Current behaviour
It tries 4 times before failing, then: 
```
$ pcluster dcv connect testenv --key-path key
ERROR: Something went wrong during DCV connection.
ERROR: Unable to obtain the Request Token from the NICE DCV external authenticator.. Wrong JSON.

Please check the logs in the /var/log/parallelcluster/ folder of the master instance and submit an issue https://github.com/aws/aws-parallelcluster/issues.
```

### Previous behaviour:
No answer from the cli:
```
$ pcluster dcv connect testenv --key-path key 
```

### Current behaviour:
```
$ pcluster dcv connect testenv --key-path key 
Unable to open the Web browser.
Please use the following one-time URL in your browser within 30 seconds:
https://url
```

### DIFF
--- a/cli/pcluster/dcv/connect.py
+++ b/cli/pcluster/dcv/connect.py
@@ -22,9 +22,16 @@ from pcluster.utils import (
     get_master_ip_and_username,
     get_stack,
     get_stack_name,
+    retry,
 )
 
 LOGGER = logging.getLogger(__name__)
+
+
+class DCVConnectionError(Exception):
+    """Error raised with DCV connection fails."""
+
+    pass
 
 
 def _check_command_output(cmd):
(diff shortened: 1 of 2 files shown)


cli/pcluster/dcv/connect.py
```diff
@@ -71,31 +100,18 @@ def dcv_connect(args):
...
-                "of the master instance and submit an issue {1}.".format(cmd, PCLUSTER_ISSUES_LINK)
+                "of the master instance and submit an issue {1}".format(ssh_cmd, PCLUSTER_ISSUES_LINK)
             )
 
     except sub.CalledProcessError as e:
         if "{0}: No such file or directory".format(DCV_CONNECT_SCRIPT) in e.output:
             error(
                 "The cluster {0} has been created with an old version of ParallelCluster "
-                "without the DCV support.".format(args.cluster_name)
+                "without the DCV support.".format(cluster_name)
```
sean-smith: "Without DCV support"

cli/pcluster/dcv/connect.py
```diff
@@ -71,31 +100,18 @@ def dcv_connect(args):
...
             )
         else:
-            error("Something went wrong during DCV connection.\n{0}".format(e.output))
+            raise DcvConnectionError(e.output)
```
sean-smith: DCVConnectionError

cli/pcluster/dcv/connect.py
```diff
@@ -52,13 +59,35 @@ def dcv_connect(args):
         DCV_SHARED_DIR=shared_dir,
     )
 
-    # Connect by ssh to the master instance and prepare DCV session
     try:
-        LOGGER.debug("SSH command: {0}".format(cmd))
-        output = _check_command_output(cmd)
+        url = retry(_retrieve_dcv_session_url, func_args=[cmd, args.cluster_name, master_ip], attempts=4)
```
sean-smith: Why `retry()` function and not `@retry` decorator? Also why 4 attempts?

cli/pcluster/dcv/connect.py
```diff
@@ -22,11 +22,18 @@
...
+    retry,
 )
 
 LOGGER = logging.getLogger(__name__)
 
 
+class DcvConnectionError(Exception):
+    """Error raised with DCV connection fails."""
+
+    pass
```
sean-smith: Whitespace

cli/pcluster/dcv/connect.py
enrico-usai: I don't want to include the retrying library in this patch release. We can decide to add it in the next round.

enrico-usai: Each round has an internal retry of 5 seconds (see pcluster_dcv_authenticator.py code) so a retry of 4 means 4*5 = 20 seconds

cli/pcluster/dcv/connect.py
enrico-usai: "without" is a continuation of the previous row ;)

cli/pcluster/dcv/connect.py
enrico-usai: It is required by pylint

cli/pcluster/dcv/connect.py
enrico-usai: Done

PR #1563, opened 2019-12-18
EFS - Add logic to create additional mount target in compute subnet

rexcsn:
* EFS can only be mounted if there is a mount target in the AZ of the instance's subnet.
* Previously we only create mount target in the master's subnet. In some cases the compute subnet could be in a different AZ than the master subnet, which means EFS cannot be mounted to the compute in this case.
* Added logic to create mount target in the AZ of the compute subnet, if needed.

### DIFF
--- a/tests/integration-tests/tests/storage/test_efs.py
+++ b/tests/integration-tests/tests/storage/test_efs.py
@@ -1,0 +1,84 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import logging
+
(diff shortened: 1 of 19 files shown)


tests/integration-tests/tests/storage/test_efs/test_efs_compute_az/pcluster.config.ini
```diff
@@ -0,0 +1,31 @@
...
+{% else %}
+initial_queue_size = 1
+maintain_initial_size = true
+{% endif %}
+efs_settings = efs
+
+[vpc parallelcluster-vpc]
+vpc_id = {{ vpc_id }}
+master_subnet_id = {{ public_subnet_id }}
+# This compute subnet would be in a different AZ than master for regions defined in AVAILABILITY_ZONE_OVERRIDES
```
demartinofra: we should verify this precondition with an assertion in the test

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -0,0 +1,84 @@
...
+
+import pytest
+
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from tests.common.schedulers_common import get_scheduler_commands
+from tests.storage.storage_common import verify_directory_correctly_shared
+
+
+# For EFS tests, only use regions defined in AVAILABILITY_ZONE_OVERRIDES in conftest
```
demartinofra: we could also create an ad hoc compute subnet that is in a different AZ (for this test only) rather than relying on the overrides

PR #1564, opened 2019-12-18
Fix update master instance

rexcsn:
* Moved all resources from MasterLaunchTemplate to MasterServer to avoid "Interface: [eni-*] in use" error.
* Added test to test_update to test that updating specification of master instance will work.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -1714,539 +1714,512 @@
         "CloudWatchLogsSubstack"
       ],
       "Properties": {
-        "LaunchTemplate": {
-          "LaunchTemplateId": {
-            "Ref": "MasterServerLaunchTemplate"
+        "InstanceType": {
+          "Ref": "MasterInstanceType"
+        },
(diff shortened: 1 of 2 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1706,543 +1706,513 @@
     },
     "MasterServer": {
       "Type": "AWS::EC2::Instance",
-      "DependsOn": [
-        "CloudWatchLogsSubstack"
```
demartinofra: isn't this a requirement anymore?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -58,6 +58,7 @@ def test_update(instance, region, pcluster_config_reader, clusters_factory, test
     s3_arn_updated = "arn:aws:s3:::fake_bucket/fake_folder/*"
     updated_config = PClusterConfig(
         max_queue_size=10,
+        master_instance_type="c4.xlarge",
```
demartinofra: don't we need to verify that this is actually changed after the update?

PR #1568, opened 2019-12-19
Allow Arbitrary Args in pcluster dcv connect

sean-smith:
This patch allows arbitrary ssh arguments to get passed into dcv connect.

```bash
pcluster dcv connect mycluster -o StrictHostKeyChecking=no
```

Which becomes:

```bash
ssh centos@54.89.10.53 -o StrictHostKeyChecking=no  "/opt/parallelcluster/scripts/pcluster_dcv_connect.sh /shared"
```

### DIFF
--- a/cli/pcluster/dcv/connect.py
+++ b/cli/pcluster/dcv/connect.py
@@ -38,7 +38,7 @@ def _check_command_output(cmd):
     return sub.check_output(cmd, shell=True, universal_newlines=True, stderr=sub.STDOUT).strip()
 
 
-def dcv_connect(args):
+def dcv_connect(args, extra_args):
     """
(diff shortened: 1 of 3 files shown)


cli/pcluster/dcv/connect.py
```diff
@@ -38,7 +38,7 @@ def _check_command_output(cmd):
     return sub.check_output(cmd, shell=True, universal_newlines=True, stderr=sub.STDOUT).strip()
 
 
-def dcv_connect(args):
+def dcv_connect(args, extra_args):
```
enrico-usai: We have to remember in future to change this, because must be possible to pass the extra_args only to the "dcv connect" subcommand not the others subcommands.

PR #1572, opened 2019-12-27
Enable feature-specific CW Logging Tests

tilne:
This commit implements the FeatureSpecificCloudWatchLoggingTestRunner
class that other integration tests can use to test how CloudWatch (CW)
logging integration is behaving with respect to only a specific feature's logs
(as opposed to validating CW logs integration for all logs that should
be collected for a given cluster).

### DIFF
--- a/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
+++ b/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
@@ -26,6 +26,21 @@ from tests.cloudwatch_logging import cloudwatch_logging_boto3_utils as cw_logs_u
 from tests.common.schedulers_common import get_scheduler_commands
 
 LOGGER = logging.getLogger(__name__)
+DEFAULT_RETENTION_DAYS = 14
+NODE_CONFIG_PATH = "/etc/chef/dna.json"
+MASTER_NODE_ROLE_NAME = "MasterServer"
+COMPUTE_NODE_ROLE_NAME = "ComputeFleet"
+NODE_ROLE_NAMES = {MASTER_NODE_ROLE_NAME, COMPUTE_NODE_ROLE_NAME}
+
+
+def _get_log_group_name_for_cluster(cluster_name):
+    """Return the name of the log group to be created for the given cluster if CloudWatch logging is enabled."""
+    return "/aws/parallelcluster/{0}".format(cluster_name)
+
+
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -151,31 +172,116 @@ def _read_log_configs_from_master(self):
...
+            return self._filter_logs_on_feature_key(logs)
+        else:
+            return [log for log in logs if not log.get("feature_conditions")]
+
+    def _populate_relevant_logs_for_node_roles(self, logs):
+        """Populate self._relevant_logs with the entries of logs."""
         # When the scheduler is AWS Batch, only keep log that whose config's node_role value is MasterServer, since
         # Batch doesn't have compute nodes in the traditional sense.
-        desired_node_roles = {"MasterServer"} if self.scheduler == "awsbatch" else {"MasterServer", "ComputeFleet"}
+        desired_node_roles = {MASTER_NODE_ROLE_NAME} if self.scheduler == "awsbatch" else set(NODE_ROLE_NAMES)
```
rexcsn: Minor: I think this is the only place `NODE_ROLE_NAMES` is used? Maybe declare `NODE_ROLE_NAMES` as a set instead of a list to avoid the set casting here?

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -151,31 +172,116 @@ def _read_log_configs_from_master(self):
...
+        """Return a boolean describing whether log contains a feature_conditions entry relevant to self.feature_key."""
+        is_relevant_for_feature = False
+        for feature_condition in log.get("feature_conditions", []):
+            if all(
+                [
+                    self.feature_key == feature_condition.get("dna_key"),
+                    node_config.get(self.feature_key) in feature_condition.get("satisfying_values"),
+                ]
+            ):
+                is_relevant_for_feature = True
```
rexcsn: Minor: no need to use the variable `is_relevant_for_feature` and `break`. `return True` and a `return False` at the end of the function should function the same.

tilne: Yeah the thought was to have only one exit point for the function. I'll change it, though.

PR #1575, opened 2020-01-03
Add AZ assertion to EFS integration tests

rexcsn:
* Add pre-assertion to check the AZs of master and compute subnets to make sure subnets are set up as expected
* Add back efs settings to batch tests so that batch scripts are tested more often


tests/integration-tests/utils.py
```diff
@@ -244,3 +244,9 @@ def paginate_boto3(method, **kwargs):
     for page in paginator.paginate(**kwargs).result_key_iters():
         for result in page:
             yield result
+
+
+def get_vpc_snakecase_value(region, vpc_stacks, vpc_output_dict):
+    vpc = vpc_stacks[region]
+    for key, value in vpc.cfn_outputs.items():
+        vpc_output_dict[to_snake_case(key)] = value
```
demartinofra: I would rather have this function return a dict and then have the caller merge in its own if necessary with an update

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -26,12 +28,17 @@
...
 @pytest.mark.usefixtures("region", "os", "instance")
-def test_efs_compute_az(scheduler, pcluster_config_reader, clusters_factory):
+def test_efs_compute_az(region, scheduler, pcluster_config_reader, clusters_factory, vpc_stacks):
     """
     Test when compute subnet is in a different AZ from master subnet.
 
     A compute mount target should be created and the efs correctly mounted on compute.
     """
+    vpc = {}
+    get_vpc_snakecase_value(region, vpc_stacks, vpc)
```
demartinofra: you could also move this logic to the assert function

tests/integration-tests/utils.py
demartinofra: sorry I meant without passing the vpc_output_dict as input

PR #1577, opened 2020-01-06
Pass shared dir path to CloudWatch logs test class

tilne:
Enable passing the path to the shared directory configured for a cluster
to the feature-specific CloudWatch logging tests. This is required
because the CloudWatch logging tests get output from commands run on the
compute nodes by creating a temporary directory in the cluster's shared
directory and re-directing all output to files in that directory.

These changes were tested by running the affected integration tests locally.


tilne:
Travis failures are due to the issue fixed by [PR 1578](https://github.com/aws/aws-parallelcluster/pull/1578)

tilne:
> It’s possible to use shared_dir = shared with no leading /. In this case I think the logic here fails, can you handle that case?

should be handled now

PR #1578, opened 2020-01-06
Fix linter errors related to exception catching

tilne:
* Fix linter failures related to style used for catching exceptions
* Also pin to a version of PYYaml that is compatible with python 3.4

Tested by running the linters locally.

### DIFF
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -28,7 +28,7 @@ REQUIRES = [
     "tabulate>=0.8.2,<=0.8.3",
     "ipaddress>=1.0.22",
     "enum34>=1.1.6",
-    "PyYAML>=5.1.2",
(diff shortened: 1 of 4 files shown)


cli/setup.py
```diff
@@ -28,7 +28,7 @@ def readme():
     "tabulate>=0.8.2,<=0.8.3",
     "ipaddress>=1.0.22",
     "enum34>=1.1.6",
-    "PyYAML>=5.1.2",
+    "PyYAML==5.2",
```
demartinofra: let's pin this only in case of Python 3.4 so that we do not enforce an older version when not strictly required

tilne: addressed here: https://github.com/aws/aws-parallelcluster/pull/1586

PR #1586, opened 2020-01-08
Only pin specific PyYAML for python 3.4

tilne:
This commit conditionally specifies the PyYAML version installed as a
ParallelCluster dependency based on the version of python used.

PyYAML 5.3 is not compatible with python 3.4, so for this version of
python, pin to 5.2. Otherwise, use the latest compatible version.

Note that the corresponding `requirements.txt` still pins to 5.2.

### DIFF
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -28,7 +28,7 @@ REQUIRES = [
     "tabulate>=0.8.2,<=0.8.3",
     "ipaddress>=1.0.22",
     "enum34>=1.1.6",
-    "PyYAML==5.2",
+    "PyYAML==5.2" if sys.version_info.major == 3 and sys.version_info.minor <= 4 else "PyYAML>=5.1.2",
 ]
 
 if sys.version_info[0] == 2:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,13 @@
 CHANGELOG
 =========
 
+X.X.X
+====
+
+**CHANGES**
+
+* Use version 5.2 of PyYAML for python 3 versions of 3.4 or earlier.
+
 2.5.1
 =====
 
--- a/cli/requirements34.txt
+++ b/cli/requirements34.txt
@@ -4,4 +4,4 @@ tabulate>=0.8.2
 ipaddress>=1.0.22
 enum34>=1.1.6
 configparser>=3.5.0
-PyYAML>=5.1.2
+PyYAML==5.2


cli/setup.py
```diff
@@ -28,7 +28,7 @@ def readme():
     "tabulate>=0.8.2,<=0.8.3",
     "ipaddress>=1.0.22",
     "enum34>=1.1.6",
-    "PyYAML==5.2",
+    "PyYAML==5.2" if sys.version_info.major == 3 and sys.version_info.minor == 4 else "PyYAML>=5.2",
```
lukeseawalker: Does PyYAML 5.3 not work only with Python 3.4 or even with previous version like 3.3?

tilne: It doesn't work with any of 3.0-3.4. Here's what you get when you try to install 5.3 with python3.4:
```
❯ pip install pyyaml==5.3
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
Collecting pyyaml==5.3
  Using cached https://files.pythonhosted.org/packages/3d/d9/ea9816aea31beeadccd03f1f8b625ecf8f645bd66744484d162d84803ce5/PyYAML-5.3.tar.gz
ERROR: PyYAML requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.10
```

sean-smith: https://github.com/yaml/pyyaml/blob/master/setup.py#L314

tilne:
> Can you please explain what you mean with
> "Note that the corresponding requirements.txt still pins to 5.2."
> I see in the requirements.txt there is a
> `PyYAML>=5.1.2`

The most recent version pins to 5.2. It was merged into develop as part of this PR: https://github.com/aws/aws-parallelcluster/pull/1578.

cli/setup.py
lukeseawalker: Ok can we put `sys.version_info.minor <= 4` then?

tilne: Sure thing.

lukeseawalker:
> > Can you please explain what you mean with
> > "Note that the corresponding requirements.txt still pins to 5.2."
> > I see in the requirements.txt there is a
> > `PyYAML>=5.1.2`
> 
> The most recent version pins to 5.2. It was merged into develop as part of this PR: #1578.

OK, I was looking at the wrong code version =)
In the past we had different requirements file depending on the python version, see https://github.com/aws/aws-parallelcluster/tree/v2.0.0/cli
not sure but maybe an idea could be to have a `requirements34.txt` with "PyYAML>=5.1.2, <=5.2", and a `requirements.txt` with just "PyYAML>=5.1.2"

PR #1590, opened 2020-01-09
Extend DCV support to Ubuntu1804

rexcsn:
* Change dcv validator to allow using DCV with Ubuntu1804
* Update validator test to adapt to changes
* Update DCV integration test to adapt to changes

### DIFF
--- a/cli/tests/pcluster/config/test_validators.py
+++ b/cli/tests/pcluster/config/test_validators.py
@@ -937,3 +937,21 @@ def test_ebs_settings_validator(mocker, cluster_section_dict, ebs_section_dict, 
 def test_shared_dir_validator(mocker, section_dict, expected_message):
     config_parser_dict = {"cluster default": section_dict}
     utils.assert_param_validator(mocker, config_parser_dict, expected_message)
+
+
+@pytest.mark.parametrize(
+    "base_os, expected_message",
+    [
+        ("alinux", "Please double check the 'base_os' configuration parameter"),
+        ("centos6", "Please double check the 'base_os' configuration parameter"),
+        ("ubuntu1604", "Please double check the 'base_os' configuration parameter"),
+        ("centos7", None),
+        ("ubuntu1804", None),
+    ],
+)
+def test_dcv_enabled_validator(mocker, base_os, expected_message):
+    config_parser_dict = {
+        "cluster default": {"base_os": base_os, "dcv_settings": "dcv"},
+        "dcv dcv": {"enable": "master"},
+    }
+    utils.assert_param_validator(mocker, config_parser_dict, expected_message)
--- a/tests/integration-tests/tests/dcv/test_dcv.py
+++ b/tests/integration-tests/tests/dcv/test_dcv.py
@@ -18,7 +18,7 @@ import pytest
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
 from tests.cloudwatch_logging.test_cloudwatch_logging import FeatureSpecificCloudWatchLoggingTestRunner
-from utils import run_command
+from utils import get_username_for_os, run_command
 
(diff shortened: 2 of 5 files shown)


tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -115,13 +115,20 @@ def _check_shared_dir(remote_command_executor, shared_dir):
     ).is_greater_than(0)
 
 
-def _check_auth_ok(remote_command_executor, external_authenticator_port, session_id, session_token):
+def _check_auth_ok(remote_command_executor, external_authenticator_port, session_id, session_token, os):
+    if os == "centos7":
```
sean-smith: Instead of a `if/else` can you just do a `string.format()` and substitute username. This makes it more extensible for the future.

cli/tests/pcluster/config/test_validators.py
```diff
@@ -937,3 +937,30 @@ def test_ebs_settings_validator(mocker, cluster_section_dict, ebs_section_dict,
...
+        (
+            {"base_os": "ubuntu1604", "dcv_settings": "dcv"},
+            {"enable": "master"},
+            "Please double check the 'base_os' configuration parameter",
+        ),
+        ({"base_os": "centos7", "dcv_settings": "dcv"}, {"enable": "master"}, None),
+        ({"base_os": "ubuntu1804", "dcv_settings": "dcv"}, {"enable": "master"}, None),
+    ],
+)
+def test_dcv_enabled_validator(mocker, cluster_section_dict, dcv_section_dict, expected_message):
```
enrico-usai: You can avoid to have `dcv_section_dict` and `cluster_section_dict` as a test parameters, since they are not changing.
The only real parameter for the test is the `base_os`.

tests/integration-tests/tests/dcv/test_dcv.py
enrico-usai: I'd create an utility function to retrieve the username according to the os and use it also for [remote_command_executor.py](https://github.com/aws/aws-parallelcluster/blob/v2.5.1/tests/integration-tests/remote_command_executor.py#L29)

tests/integration-tests/utils.py
```diff
@@ -244,3 +244,14 @@ def paginate_boto3(method, **kwargs):
...
+
+def get_username_for_os(os):
+    usernames = {
+        "alinux": "ec2-user",
+        "centos6": "centos",
+        "centos7": "centos",
+        "ubuntu1604": "ubuntu",
+        "ubuntu1804": "ubuntu",
+    }
+    return usernames[os]
```
enrico-usai: Minor: I'd use `usernames.get(os)`.
With `usernames[os]`, if you pass unsupported OS you'll obtain a `KeyError`, instead with `get()` you'll have `None` as result or it's possible to add a default returned value.

cli/tests/pcluster/config/test_validators.py
```diff
@@ -937,3 +937,21 @@ def test_ebs_settings_validator(mocker, cluster_section_dict, ebs_section_dict,
 def test_shared_dir_validator(mocker, section_dict, expected_message):
     config_parser_dict = {"cluster default": section_dict}
     utils.assert_param_validator(mocker, config_parser_dict, expected_message)
+
+
+@pytest.mark.parametrize(
+    "base_os, expected_message",
+    [
+        ("alinux", "Please double check the 'base_os' configuration parameter"),
```
sean-smith: What is this checking for?

rexcsn: It's checking that when using dcv with OSs that are not supported we are printing out error message. Full error message is `"NICE DCV can be used with one of the following operating systems: ["centos7", "ubuntu1804"]. Please double check the 'base_os' configuration parameter"`

PR #1596, opened 2020-01-14
Retry on throttling from CloudFormation API

demartinofra:
This is likely to happen when several compute nodes are being bootstrapped concurrently

I manually tested cli commands that are not covered by unit tests.


cli/pcluster/utils.py
```diff
@@ -580,3 +582,13 @@ def retry(func, func_args, attempts=1, wait=0):
...
+
+
+def retry_on_boto3_throttling(func, wait=5, *args, **kwargs):
+    while True:
+        try:
+            return func(*args, **kwargs)
+        except ClientError as e:
+            if e.response["Error"]["Code"] != "Throttling":
+                raise
+            time.sleep(wait)
```
sean-smith: Let's make this dynamically extend like an exponential backoff or the like

demartinofra: do you see a real value in having a exponential backoff here? In the end the retry happens every 5 seconds and I wouldn't extend it more than that.
Also this call is not the real responsible of the throttling but rather the call from the compute nodes where I agree an exponential backoff is the more appropriate choice.

cli/pcluster/utils.py
```diff
@@ -580,3 +598,13 @@ def retry(func, func_args, attempts=1, wait=0):
...
```
enrico-usai: I'd add a warning message in the log file (or at least a debug message).
In this way if the CLI hangs we are able to understand the cause is the Throttling by checking the CLI logs.

cli/tests/common.py
```diff
@@ -1,6 +1,9 @@
 from collections import namedtuple
 
-MockedBoto3Request = namedtuple("MockedBoto3Request", ["method", "response", "expected_params"])
+MockedBoto3Request = namedtuple(
+    "MockedBoto3Request", ["method", "response", "expected_params", "generate_error", "error_code"]
+)
+MockedBoto3Request.__new__.__defaults__ = (False, None)
```
enrico-usai: I'd add a comment to describe this row

cli/tests/pcluster/config/test_validators.py
```diff
@@ -524,7 +524,7 @@ def test_raid_validators(mocker, section_dict, expected_message):
 @pytest.mark.parametrize(
     "kms_key_id, expected_message",
     [
-        ("9e8a129be-0e46-459d-865b-3a5bf974a22k", None,),
+        ("9e8a129be-0e46-459d-865b-3a5bf974a22k", None),
```
enrico-usai: Why the linter was not able to detect this error before?
Did we forget to enable something in tox?

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -189,14 +204,142 @@ def test_get_stack_resources(boto3_stubber, resources, error_message):
...
     else:
         with pytest.raises(SystemExit, match=response) as sysexit:
             utils.get_stack_resources(FAKE_STACK_NAME)
         assert_that(sysexit.value.code).is_not_equal_to(0)
+
+
+def test_retry_on_boto3_throttling(boto3_stubber, mocker):
+    sleep_mock = mocker.patch("pcluster.utils.time.sleep")
+    mocked_requests = [
+        MockedBoto3Request(
```
enrico-usai: What about defining a constant for this MockedBoto3Request since it is used in many places?
Like you did for `_generate_stack_event`

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -189,14 +204,142 @@ def test_get_stack_resources(boto3_stubber, resources, error_message):
...
+            generate_error=True,
+            error_code="Throttling",
+        ),
+        MockedBoto3Request(
+            method="describe_stack_resources", response={}, expected_params={"StackName": FAKE_STACK_NAME}
+        ),
+    ]
+    client = boto3_stubber("cloudformation", mocked_requests)
+    utils.retry_on_boto3_throttling(client.describe_stack_resources, StackName=FAKE_STACK_NAME)
+    sleep_mock.assert_called_with(5)
```
enrico-usai: I'd also add a check for the number of call, to be sure it reflects the number of throttling failures.
Something like: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count

cli/tests/pcluster/config/test_validators.py
demartinofra: probably it's just a newer version of black that implemented this

cli/tests/pcluster/utils/test_pcluster_utils.py
demartinofra: tried to add this but other libraries are also calling the time.sleep function. We'll rely on the boto3 stubber to guarantee that the number of calls is correct

cli/tests/pcluster/utils/test_pcluster_utils.py
demartinofra: we decided to leave as is since they are testing different methods with different boto3 calls and having some duplication here helps the readability of tests

PR #1597, opened 2020-01-15
Fix DCV test with known_host entries

rexcsn:
* Add DCV master server to global known hosts file `/etc/ssh/ssh_known_hosts`

### DIFF
--- a/tests/integration-tests/tests/dcv/test_dcv.py
+++ b/tests/integration-tests/tests/dcv/test_dcv.py
@@ -18,7 +18,7 @@ import pytest
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
 from tests.cloudwatch_logging.test_cloudwatch_logging import FeatureSpecificCloudWatchLoggingTestRunner
-from utils import get_username_for_os, run_command
+from utils import add_keys_to_known_hosts, get_username_for_os, remove_keys_from_known_hosts, run_command
 
 SERVER_URL = "https://localhost"
 DCV_CONNECT_SCRIPT = "/opt/parallelcluster/scripts/pcluster_dcv_connect.sh"
@@ -45,15 +45,21 @@ def test_dcv_configuration(
     # dcv connect show url
     env = operating_system.environ.copy()
     env["AWS_DEFAULT_REGION"] = region
-    # Disable StrictHostKeyChecking by adding option to end of file
-    operating_system.system(r"echo '    StrictHostKeyChecking no' >> /etc/ssh/ssh_config")
-    result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
+
+    # add ssh key to jenkins user known hosts file to avoid ssh keychecking prompt
+    host_keys_file = operating_system.path.expanduser("~/.ssh/known_hosts")
+    add_keys_to_known_hosts(cluster.master_ip, host_keys_file)
+
+    try:
+        result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
+    finally:
+        # remove ssh key from jenkins user known hosts file
+        remove_keys_from_known_hosts(cluster.master_ip, host_keys_file, env=env)
+
     assert_that(result.stdout).matches(
         r"Please use the following one-time URL in your browser within 30 seconds:\n"
         r"https:\/\/(\b(?:\d{1,3}\.){3}\d{1,3}\b):" + str(dcv_port) + r"\?authToken=(.*)"
     )
-    # Enable StrictHostKeyChecking again by removing last line
-    operating_system.system(r"sed -i '$ d' /etc/ssh/ssh_config")
 
     # check error cases
     _check_auth_ko(
--- a/tests/integration-tests/utils.py
+++ b/tests/integration-tests/utils.py
@@ -14,6 +14,7 @@ import os
 import random
 import re
 import shlex
+import socket
 import string
 import subprocess
 
@@ -265,3 +266,14 @@ def get_username_for_os(os):
         "ubuntu1804": "ubuntu",
     }
     return usernames.get(os)
+
+
+def add_keys_to_known_hosts(hostname, host_keys_file):
+    """Add ssh key for a host to a known_hosts file."""
+    os.system("ssh-keyscan -t rsa {0} >> {1}".format(hostname, host_keys_file))
+
+
+def remove_keys_from_known_hosts(hostname, host_keys_file, env):
+    """Remove ssh key for a host from a known_hosts file."""
+    for host in hostname, "{0}.".format(hostname), socket.gethostbyname(hostname):
+        run_command("ssh-keygen -R {0} -f {1}".format(host, host_keys_file), env=env)


tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -45,15 +49,19 @@ def test_dcv_configuration(
...
     result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
     assert_that(result.stdout).matches(
         r"Please use the following one-time URL in your browser within 30 seconds:\n"
         r"https:\/\/(\b(?:\d{1,3}\.){3}\d{1,3}\b):" + str(dcv_port) + r"\?authToken=(.*)"
     )
-    # Enable StrictHostKeyChecking again by removing last line
-    operating_system.system(r"sed -i '$ d' /etc/ssh/ssh_config")
+
+    # remove ssh key
+    _remove_keys_from_known_hosts(cluster.master_ip, host_keys_file)
```
enrico-usai: This code is not executed if the previous step raises an error.

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+    # ubuntu1404 ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # ubuntu1604 ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # centos7    ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # centos6    ssh-rsa
+    key_type_list = ["ssh-rsa"]
+
+    for key_type in key_type_list:
+        transport = None
+        try:
+            sock = socket.socket()
```
enrico-usai: could you add a comment to describe "why" we are using socket and transport?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+            sock.settimeout(5)
+            sock.connect((hostname, 22))
+            transport = Transport(sock)
+            transport._preferred_keys = [key_type]
+            transport.start_client()
+            server_keys.append(transport.get_remote_server_key())
+        except Exception:
+            pass
```
enrico-usai: I'd add a warning/debug message to log the exception case.

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
     ips = response["SecurityGroups"][0]["IpPermissions"]
     target = next(filter(lambda x: x.get("FromPort", -1) == port, ips), {})
     assert_that(target["IpRanges"][0]["CidrIp"]).is_equal_to(expected_cidr)
+
+
+def _get_server_keys(hostname):
```
ddeidda: What about moving these functions into a common/utils module, so that we can re-use it more easily if needed?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+
+    return {hostname: [(server_key.get_base64(), server_key.get_name()) for server_key in server_keys]}
+
+
+def _add_keys_to_known_hosts(server_keys, host_keys_file):
+    try:
+        if not operating_system.path.isfile(host_keys_file):
+            host_keys = HostKeys()
+        else:
+            host_keys = HostKeys(filename=host_keys_file)
```
enrico-usai: This code can be rewritten with something like:
```python
filename = host_keys_file if operating_system.path.isfile(host_keys_file) else None
host_keys = HostKeys(filename=filename)
```

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+
+        for hostname, key_list in server_keys.items():
+            try:
+                for key_tuple in key_list:
```
enrico-usai: you could be more explicit with:
```python
for key_base64, key_type in key_list:
    key = RSAKey(data=base64.b64decode(key_base64))
    host_keys.add(hostname=hostname, key=key, keytype=key_type)
    ...
```

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+            except Exception as e:
+                logging.error(
+                    "Failed adding keys to known hosts file for host '{0}', with exception: {1}".format(hostname, e)
+                )
+    except Exception as e:
+        logging.error("Failed adding keys to known hosts file '{0}', with exception: {1}".format(host_keys_file, e))
+
+
+def _remove_keys_from_known_hosts(hostname, host_keys_file):
+    command = "ssh-keygen -R " + hostname + " -f " + host_keys_file
```
enrico-usai: You can avoid code duplication by writing something like:
```python
for host in [hostname, "{0}.".format(hostname), socket.gethostbyname(hostname)]:
    run_command("ssh-keygen -R {0} -f {1}".format(host, host_keys_file))
```

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -130,3 +138,72 @@ def _check_security_group(region, cluster, port, expected_cidr):
...
+        if not operating_system.path.isfile(host_keys_file):
+            host_keys = HostKeys()
+        else:
+            host_keys = HostKeys(filename=host_keys_file)
+
+        for hostname, key_list in server_keys.items():
+            try:
+                for key_tuple in key_list:
+                    key = RSAKey(data=base64.b64decode(key_tuple[0]))
+                    host_keys.add(hostname=hostname, key=key, keytype=key_tuple[1])
```
ddeidda: Maybe code duplication could be avoided also here:
for host in hostname, "{0}.".format(hostname), socket.gethostbyname(hostname):
  host_keys.add(hostname=host, key=key, keytype=key_tuple[1])

tests/integration-tests/tests/dcv/test_dcv.py
rexcsn: This logic was taken from the node package, and we used socket/transport there so we could use threading to retrieve keys from multiple compute nodes. But since we are only dealing with the master here I agree we can simplify the logic a lot and just use `ssh-keyscan`.

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -45,15 +45,21 @@ def test_dcv_configuration(
     # dcv connect show url
     env = operating_system.environ.copy()
     env["AWS_DEFAULT_REGION"] = region
-    # Disable StrictHostKeyChecking by adding option to end of file
-    operating_system.system(r"echo '    StrictHostKeyChecking no' >> /etc/ssh/ssh_config")
-    result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
+
+    # add ssh key to global known hosts file to avoid ssh keychecking prompt
+    host_keys_file = "/etc/ssh/ssh_known_hosts"
+    operating_system.system("ssh-keyscan -t rsa {0} > {1}".format(cluster.master_ip, host_keys_file))
```
ddeidda: Could we extract also this part into utils as add_keys_to_known_hosts(host_name, key_file)?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -27,7 +27,7 @@
 @pytest.mark.parametrize(
     "dcv_port, access_from, shared_dir", [(8443, "0.0.0.0/0", "/shared"), (5678, "192.168.1.1/32", "/myshared")]
 )
-@pytest.mark.regions(["eu-west-1", "cn-northwest-1"])  # DCV license bucket not present in us-gov
+@pytest.mark.regions(["us-east-2", "cn-northwest-1"])  # DCV license bucket not present in us-gov
```
tilne: just curious: why did this need to be changed?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -45,15 +45,21 @@ def test_dcv_configuration(
     # dcv connect show url
     env = operating_system.environ.copy()
     env["AWS_DEFAULT_REGION"] = region
-    # Disable StrictHostKeyChecking by adding option to end of file
-    operating_system.system(r"echo '    StrictHostKeyChecking no' >> /etc/ssh/ssh_config")
-    result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
+
+    # add ssh key to global known hosts file to avoid ssh keychecking prompt
```
tilne: should this comment be changed to reflect the fact that the user known hosts file is now being used?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -45,15 +45,21 @@ def test_dcv_configuration(
...
+    host_keys_file = operating_system.path.expanduser("~/.ssh/known_hosts")
+    add_keys_to_known_hosts(cluster.master_ip, host_keys_file)
+
+    try:
+        result = run_command(["pcluster", "dcv", "connect", cluster.name, "--show-url"], env=env)
+    finally:
+        # remove ssh key from global known hosts file
```
tilne: same thing here: should this comment be changed to reflect the fact that the user known hosts file is now being used?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -45,15 +45,21 @@ def test_dcv_configuration(
...
+        remove_keys_from_known_hosts(cluster.master_ip, host_keys_file, env=env)
```
tilne: is there any risk of multiple processes trying to modify the hosts file at the same time?

tests/integration-tests/tests/dcv/test_dcv.py
rexcsn: Just because my AMI is in `us-east-2`. Generally this could be any commercial region, but I will change in back just for consistency.

tests/integration-tests/tests/dcv/test_dcv.py
rexcsn: This should not be a problem, because to add keys, we are only appending to file and not overwriting. To remove, `ssh-keygen` scans the file and only removes the matching entries. So multiple add/remove at the same time should not interfere with each other.

Passing Jenkins test had 2 tests in parallel and it seems to be fine.

tests/integration-tests/tests/dcv/test_dcv.py
rexcsn: Absolutely

tests/integration-tests/tests/dcv/test_dcv.py
tilne: Okay. I do think this could cause issues if we get unlucky, but I agree that the chances of it happening are very small considering only two of these tests will ever be running at one time.

PR #1600, opened 2020-01-15
Require FSx Storage Capacity

sean-smith:

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -136,6 +136,9 @@ def fsx_validator(section_key, section_label, pcluster_config):
     if fsx_export_path and not fsx_import_path:
         errors.append("When specifying 'export_path', the 'import_path' option must be specified")
 
+    if not fsx_section.get_param_value("storage_capacity") and not fsx_section.get_param_value("fsx_fs_id"):
+        errors.append("When specifying 'fsx' section, the 'storage_capacity' option must be specified")
+
     return errors, warnings
 
 
--- a/cli/tests/pcluster/config/test_validators.py
+++ b/cli/tests/pcluster/config/test_validators.py
@@ -484,16 +484,20 @@ def test_raid_validators(mocker, section_dict, expected_message):
 @pytest.mark.parametrize(
     "section_dict, expected_message",
     [
-        ({"imported_file_chunk_size": 1024, "import_path": "test"}, None),
+        ({"imported_file_chunk_size": 1024, "import_path": "test", "storage_capacity": 1200}, None),
         (
-            {"imported_file_chunk_size": 1024},
+            {"imported_file_chunk_size": 1024, "storage_capacity": 1200},
             "When specifying 'imported_file_chunk_size', the 'import_path' option must be specified",
         ),
-        ({"export_path": "test", "import_path": "test"}, None),
-        ({"export_path": "test"}, "When specifying 'export_path', the 'import_path' option must be specified"),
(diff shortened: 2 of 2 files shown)


cli/pcluster/commands.py
```diff
@@ -75,7 +75,7 @@ def _check_for_updates(pcluster_config):
         utils.check_if_latest_version()
 
 
-def create(args):  # noqa: C901 FIXME!!!
+def create(args):  # noqa: C901 FIXME!!!w
```
rexcsn: typo?

cli/pcluster/config/validators.py
```diff
@@ -136,6 +136,9 @@ def fsx_validator(section_key, section_label, pcluster_config):
     if fsx_export_path and not fsx_import_path:
         errors.append("When specifying 'export_path', the 'import_path' option must be specified")
 
+    if not fsx_section.get_param_value("storage_capacity") and not fsx_section.get_param_value("fsx_fs_id"):
+        errors.append("When specifying 'fsx' section, the 'storage_capacity' option must be specified")
```
rexcsn: Should we also check for the specific storage capacity as outlined by FSx: `For Lustre file systems, valid values are 1,200, 2,400, 3,600, then continuing in increments of 3600 GiB.`

We should at least provide this info in the error message so the user does not have to look it up again

cli/tests/pcluster/config/test_validators.py
```diff
@@ -484,16 +484,20 @@ def test_raid_validators(mocker, section_dict, expected_message):
...
-        ({"shared_dir": "/NONE"}, "/NONE cannot be used as a shared directory"),
-        ({"shared_dir": "/fsx"}, None),
+        ({"export_path": "test", "import_path": "test", "storage_capacity": 1200}, None),
+        (
+            {"export_path": "test", "storage_capacity": 1200},
+            "When specifying 'export_path', the 'import_path' option must be specified",
+        ),
+        ({"shared_dir": "NONE", "storage_capacity": 1200}, "NONE cannot be used as a shared directory"),
+        ({"shared_dir": "/NONE", "storage_capacity": 1200}, "/NONE cannot be used as a shared directory"),
+        ({"shared_dir": "/fsx"}, "the 'storage_capacity' option must be specified"),
```
rexcsn: Probably need to change this or add more tests according to comment above.

cli/pcluster/commands.py
sean-smith: yep

cli/pcluster/config/validators.py
sean-smith: That's done here https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/config/validators.py#L194

cli/tests/pcluster/config/test_validators.py
sean-smith: That's done here https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/config/test_validators.py#L714

enrico-usai:
@sean-smith could you describe in the comment of the PR why this change is necessary? 
Has anything changed in FSx side or were we missing this already required parameter?

In future please add this kind of description in the commit message too.

sean-smith:
@enrico-usai Previously the stack failed CloudFormation create with a very opaque error message "Filesystem failed to create". This was despite our docs saying the parameter was optional. It's only optional if the customer provides a `fsx_fs_id` for an existing filesystem.

This patch fixes that behaviour by validating `storage_capacity` is provided when no `fsx_fs_id` is given.

PR #1603, opened 2020-01-16
Add warning message in cli logs when dcv port is open to 0.0.0.0/0

demartinofra:
Add warning message in cli logs when dcv port is open to 0.0.0.0/0

Codecov: https://codecov.io/gh/aws/aws-parallelcluster

### DIFF
--- a/cli/pcluster/cli.py
+++ b/cli/pcluster/cli.py
@@ -83,6 +83,7 @@ def create_ami(args):
 
 def config_logger():
     logger = logging.getLogger("pcluster")
+    file_only_logger = logging.getLogger("cli_log_file")
     logger.setLevel(logging.DEBUG)
 
     log_stream_handler = logging.StreamHandler(sys.stdout)
@@ -99,8 +100,9 @@ def config_logger():
 
     log_file_handler = RotatingFileHandler(logfile, maxBytes=5 * 1024 * 1024, backupCount=1)
     log_file_handler.setLevel(logging.DEBUG)
-    log_file_handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(name)s - %(message)s"))
+    log_file_handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(module)s - %(message)s"))
     logger.addHandler(log_file_handler)
+    file_only_logger.addHandler(log_file_handler)
 
 
 def _addarg_config(subparser):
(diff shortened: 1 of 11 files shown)


cli/pcluster/config/validators.py
```diff
@@ -28,6 +29,15 @@
     get_supported_os,
 )
 
+LOGFILE_LOGGER = logging.getLogger("cli_log_file")
+
+DCV_MESSAGES = {
+    "warnings": {
+        "access_from_world": "With this configuration you are opening dcv port ({port}) to the word (0.0.0.0/0). "
```
lukeseawalker: word/world

demartinofra: nice catch. thanks!

cli/pcluster/cli.py
```diff
@@ -99,8 +100,9 @@ def config_logger():
 
     log_file_handler = RotatingFileHandler(logfile, maxBytes=5 * 1024 * 1024, backupCount=1)
     log_file_handler.setLevel(logging.DEBUG)
-    log_file_handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(name)s - %(message)s"))
+    log_file_handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(module)s - %(message)s"))
```
ddeidda: Just out of curiosity, what is the goal of this change?

cli/pcluster/config/validators.py
```diff
@@ -237,6 +247,13 @@ def dcv_enabled_validator(param_key, param_value, pcluster_config):
         if get_partition() not in get_supported_dcv_partition():
             errors.append("NICE DCV is not supported in the selected region '{0}'".format(get_region()))
 
+        if pcluster_config.get_section("dcv").get_param_value("access_from") == "0.0.0.0/0":
```
ddeidda: Minor, could we use a meaningful utils constant name like CIDR_ALL_IPS?

cli/tests/requirements.txt
```diff
@@ -5,4 +5,4 @@ pytest-datadir
 pytest-html
 pytest-mock
 assertpy
-discover # Needed for Python 2.6 compatibility
+codecov
```
ddeidda: Should we maybe split this part about codecov on a separate commit?

cli/pcluster/cli.py
demartinofra: It's to always have the name of the module that originated the log record rather than the name of the logger that can be arbitrary

cli/pcluster/config/validators.py
demartinofra: done

cli/pcluster/constants.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ddeidda: Good idea, better than in the utils file.

PR #1610, opened 2020-01-22
Validate S3 URI's

sean-smith:

### DIFF
--- a/cli/pcluster/config/mappings.py
+++ b/cli/pcluster/config/mappings.py
@@ -383,7 +383,9 @@ FSX = {
             ("import_path", {
                 "validators": [s3_bucket_validator],
             }),
-            ("weekly_maintenance_start_time", {}),  # TODO add regex
+            ("weekly_maintenance_start_time", {
+                "allowed_values": r"NONE|^[1-7]:([01]\d|2[0-3]):?([0-5]\d)$",
+            }),
         ]
     )
 }
--- a/cli/pcluster/config/mappings.py
+++ b/cli/pcluster/config/mappings.py
@@ -58,6 +58,7 @@ from pcluster.config.validators import (
     intel_hpc_validator,
     kms_key_validator,
     raid_volume_iops_validator,
+    s3_bucket_validator,
     scheduler_validator,
     shared_dir_validator,
     url_validator,
@@ -376,8 +377,12 @@ FSX = {
                 "type": IntParam,
                 "validators": [fsx_imported_file_chunk_size_validator]
             }),
-            ("export_path", {}),  # TODO add regex
-            ("import_path", {}),  # TODO add regex
+            ("export_path", {
+                "validators": [s3_bucket_validator],
+            }),
+            ("import_path", {
+                "validators": [s3_bucket_validator],
+            }),
             ("weekly_maintenance_start_time", {}),  # TODO add regex
         ]
     )
@@ -538,10 +543,12 @@ CLUSTER = {
                 "validators": [ec2_iam_role_validator],  # TODO add regex
             }),
             ("s3_read_resource", {
-                "cfn_param_mapping": "S3ReadResource",  # TODO add validator
+                "cfn_param_mapping": "S3ReadResource",
+                "validators": [s3_bucket_validator],
             }),
             ("s3_read_write_resource", {
-                "cfn_param_mapping": "S3ReadWriteResource",  # TODO add validator
+                "cfn_param_mapping": "S3ReadWriteResource",
+                "validators": [s3_bucket_validator],
             }),
             (
                 "disable_hyperthreading",
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -581,7 +581,7 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
             bucket = param_value.split("/")[2]
             boto3.client("s3").head_bucket(Bucket=bucket)
         except ClientError:
-            errors.append("The S3 bucket '{0}' does not exist or you do not have access to it.".format(param_value))
+            warnings.append("The S3 bucket '{0}' does not exist or you do not have access to it.".format(param_value))
     else:
         errors.append(
             "The value '{0}' used for the parameter '{1}' is not a valid S3 URI.".format(param_value, param_key)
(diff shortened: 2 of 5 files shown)


cli/tests/pcluster/config/test_validators.py
```diff
@@ -319,6 +319,66 @@ def test_url_validator(mocker, boto3_stubber):
...
+                    "export_path": "s3://test/test1/test2",
+                },
+            },
+            3,
+            {"Bucket": "test"},
+            "The value 'http://test/test.json' used for the parameter 'import_path' is not a valid S3 URI.",
+        ),
+    ],
+)
+def test_s3_validator(mocker, boto3_stubber, config, num_calls, bucket, expected_message):
```
enrico-usai: test_s3_url_validator ?

cli/tests/pcluster/config/test_validators.py
```diff
@@ -516,25 +576,48 @@ def test_kms_key_validator(mocker, boto3_stubber):
...
+            "AcceptRanges": "bytes",
+            "ContentType": "text/html",
+            "LastModified": "Thu, 16 Apr 2015 18:19:14 GMT",
+            "ContentLength": 77,
+            "VersionId": "null",
+            "ETag": '"30a6ec7e1a9ad79c203d05a589c8b400"',
+            "Metadata": {},
+        }
+    }
+    mocked_requests = [MockedBoto3Request(method="head_bucket", response=head_bucket_response, expected_params=bucket)]
```
enrico-usai: We can create an utility method to mock s3_head_bucket call and use it everywhere. The method can have the bucket name as parameter and return the mocked_requests to use in the boto3_stubber

cli/pcluster/config/validators.py
```diff
@@ -572,6 +572,24 @@ def url_validator(param_key, param_value, pcluster_config):
...
+def s3_bucket_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if urlparse(param_value).scheme == "s3":
+        try:
+            bucket = param_value.split("/")[2]
+            boto3.client("s3").head_bucket(Bucket=bucket)
+        except ClientError:
+            errors.append("The S3 bucket '{0}' does not exist or you do not have access to it.".format(param_value))
```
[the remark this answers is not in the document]
sean-smith: good point

cli/pcluster/config/mappings.py
```diff
@@ -58,6 +58,7 @@
     intel_hpc_validator,
     kms_key_validator,
     raid_volume_iops_validator,
+    s3_bucket_validator,
```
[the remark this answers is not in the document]
sean-smith: it actually does a `head_bucket` so doesn't care about the path

cli/tests/pcluster/config/test_validators.py
sean-smith: good idea

cli/pcluster/config/mappings.py
```diff
@@ -376,9 +377,15 @@
...
-            ("import_path", {}),  # TODO add regex
-            ("weekly_maintenance_start_time", {}),  # TODO add regex
+            ("export_path", {
+                "validators": [s3_bucket_validator],
+            }),
+            ("import_path", {
+                "validators": [s3_bucket_validator],
+            }),
+            ("weekly_maintenance_start_time", {
+                "allowed_values": r"^[1-7]:([01]\d|2[0-3]):?([0-5]\d)$",
```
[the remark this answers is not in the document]
sean-smith: done!

cli/tests/pcluster/config/test_validators.py
sean-smith: done

cli/pcluster/config/validators.py
```diff
@@ -572,6 +572,24 @@ def url_validator(param_key, param_value, pcluster_config):
...
+def s3_bucket_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if urlparse(param_value).scheme == "s3":
+        try:
+            bucket = param_value.split("/")[2]
+            boto3.client("s3").head_bucket(Bucket=bucket)
+        except ClientError:
+            warnings.append("The S3 bucket '{0}' does not exist or you do not have access to it.".format(param_value))
```
enrico-usai: What about adding a message like: _"Please be sure the cluster nodes have access to it."_?

sean-smith:
@enrico-usai See https://github.com/aws/aws-parallelcluster/pull/1613

cli/pcluster/config/mappings.py
```diff
@@ -538,10 +545,12 @@
                 "validators": [ec2_iam_role_validator],  # TODO add regex
             }),
             ("s3_read_resource", {
-                "cfn_param_mapping": "S3ReadResource",  # TODO add validator
+                "cfn_param_mapping": "S3ReadResource",
+                "validators": [s3_bucket_validator],
```
demartinofra: s3_read_resource and s3_read_write_resource are arns pointing to a bucket or bucket objects and cannot be validated by s3_bucket_validator

cli/pcluster/config/validators.py
```diff
@@ -572,6 +572,24 @@ def url_validator(param_key, param_value, pcluster_config):
...
 
 
+def s3_bucket_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if urlparse(param_value).scheme == "s3":
+        try:
+            bucket = param_value.split("/")[2]
+            boto3.client("s3").head_bucket(Bucket=bucket)
```
demartinofra: this validation does not guarantee that the nodes of the cluster have access to the bucket (because the user policy is different that the role assumed by cluster nodes) and also enforces the user that creates the cluster to have head_bucket access to the s3 bucket itself.

cli/pcluster/config/validators.py
demartinofra: It might be also useful at this point to add an assertion in the integration tests that verifies that no warnings are returned from the pcluster create command. This could be embedded in the ClustersFactory for now so that it gets applied to all tests, then we can add a flag to disable it in case some of the tests require a warning to be displayed.

enrico-usai: Fixed by https://github.com/aws/aws-parallelcluster/pull/1613

cli/pcluster/config/mappings.py
enrico-usai: Good catch!

PR #1615, opened 2020-01-25
S3 Bucket Validator fixes

sean-smith:
* Removes the bucket validator from `s3_read_resource` and `s3_read_write_resource`
* Adds a check in the integration tests to fail if WARNINGS are created during cluster create

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -162,6 +162,10 @@ class ClustersFactory:
             error = "Cluster creation failed for {0} with output: {1}".format(name, result.stdout)
             logging.error(error)
             raise Exception(error)
+        elif "WARNING" in result.stdout:
+            error = "Cluster creation for {0} generated a warning: {1}".format(name, result.stdout)
+            logging.error(error)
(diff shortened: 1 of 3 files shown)


tests/integration-tests/clusters_factory.py
```diff
@@ -162,6 +162,10 @@ def create_cluster(self, cluster):
             error = "Cluster creation failed for {0} with output: {1}".format(name, result.stdout)
             logging.error(error)
             raise Exception(error)
+        elif "WARNING" in result.stdout:
+            error = "Cluster creation generated a warning: {1}".format(name, result.stdout)
```
enrico-usai: in the format you only have the {1} but you're passing two variables

sean-smith: fixed

enrico-usai: I'd add the name of the cluster like in the error case.
`"Cluster creation for {0} generated a warning: {1}".format(name, result.stdout)`

sean-smith: done

PR #1619, opened 2020-01-27
Add symlink to alinux batch resources for alinux2

tilne:
The plan is to support using batch with alinux2. Use a symlink to the
alinux batch docker resources directory.


tilne:
> LGTM. What about copying the artefacts to alinux2 and symlinking the other way around? Just cause the current Dockerfile is actually using alinux2

Makes sense. I'll do that now.

PR #1630, opened 2020-02-05
Validate Data Repository Export works

sean-smith:
* Verify POSIX permissions of a file on S3

### DIFF
--- a/tests/integration-tests/tests/storage/test_fsx_lustre.py
+++ b/tests/integration-tests/tests/storage/test_fsx_lustre.py
@@ -13,10 +13,13 @@ import logging
 
 import boto3
 import pytest
+from retrying import retry
 
+import utils
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
 from tests.common.schedulers_common import SgeCommands
+from time_utils import minutes, seconds
 
 
 @pytest.mark.regions(["us-east-1"])
@@ -37,11 +40,13 @@ def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_
     cluster_config = pcluster_config_reader(bucket_name=bucket_name, mount_dir=mount_dir)
     cluster = clusters_factory(cluster_config)
     remote_command_executor = RemoteCommandExecutor(cluster)
+    fsx_fs_id = get_fsx_fs_id(cluster, region)
 
     _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os)
     _test_import_path(remote_command_executor, mount_dir)
     _test_fsx_lustre_correctly_shared(remote_command_executor, mount_dir)
     _test_export_path(remote_command_executor, mount_dir, bucket_name)
+    _test_data_repository_task(remote_command_executor, mount_dir, bucket_name, fsx_fs_id, region)
 
 
 def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os):
@@ -61,6 +66,11 @@ def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os):
             mount_dir=mount_dir, mount_options=mount_options.get(os, mount_options["default"])
         )
     )
+
+
+def get_fsx_fs_id(cluster, region):
+    fsx_stack = utils.get_substacks(cluster.cfn_name, region=region, sub_stack_name="FSXSubstack")[0]
+    return utils.retrieve_cfn_outputs(fsx_stack, region).get("FileSystemId")
 
 
 def _test_import_path(remote_command_executor, mount_dir):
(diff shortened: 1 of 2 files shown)


tests/integration-tests/utils.py
```diff
@@ -92,14 +92,24 @@ def retrieve_cfn_resources(stack_name, region):
...
+            r.get("PhysicalResourceId")
+            for r in stack_resources
+            if (r.get("ResourceType") == "AWS::CloudFormation::Stack" and r.get("LogicalResourceId") == sub_stack_name)
+        ]
+    else:
+        return [
+            r.get("PhysicalResourceId")
+            for r in stack_resources
+            if r.get("ResourceType") == "AWS::CloudFormation::Stack"
+        ]
```
tilne: it's a minor point, but i'd prefer a two-phased approach where you first filter on the resource type and then filter for sub stack if needed:
```
substacks = [s for s in stacks if <logic for selecting substacks>]
if sub_stack_name:
    substacks = [s for s in substacks if <condition for identifying desired substack>]
```
that way the logic for selecting the resources that aren't substacks isn't duplicated.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -37,14 +39,16 @@ def test_fsx_lustre(region, pcluster_config_reader, clusters_factory, s3_bucket_
...
-    _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os)
+    _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os, region, fsx_fs_id)
     _test_import_path(remote_command_executor, mount_dir)
     _test_fsx_lustre_correctly_shared(remote_command_executor, mount_dir)
     _test_export_path(remote_command_executor, mount_dir, bucket_name)
+    _test_data_repository_task(remote_command_executor, mount_dir, bucket_name, fsx_fs_id, region)
 
 
-def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os):
+def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os, region, fsx_fs_id):
```
tilne: these are unused.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -63,6 +67,12 @@ def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os):
     )
 
 
+def get_fsx_fs_id(cluster, region):
+    cluster_name = "parallelcluster-{0}".format(cluster.name)
```
tilne: `cluster.cfn_name` would give you the same thing

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -98,3 +108,50 @@ def _test_export_path(remote_command_executor, mount_dir, bucket_name):
...
+
+    fsx = boto3.client("fsx", region_name=region)
+    task = fsx.create_data_repository_task(
+        FileSystemId=fsx_fs_id, Type="EXPORT_TO_REPOSITORY", Paths=["file_to_export"], Report={"Enabled": False}
+    ).get("DataRepositoryTask")
+
+    status = task.get("Lifecycle")
+    task_id = task.get("TaskId")
+
+    while status in ["PENDING", "EXECUTING"]:
```
tilne: would it be good to include a `CANCELLING` here as well? within the context of the test it probably doesn't matter, but I'm wondering what would happen if this test existed while the task was still cancelling and the cluster was deleted.

tilne: also, is there a timeout here? maybe wrap this polling in a function decorated with a retrying?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -98,3 +108,50 @@ def _test_export_path(remote_command_executor, mount_dir, bucket_name):
...
+        status = fsx.describe_data_repository_tasks(TaskIds=[task_id]).get("DataRepositoryTasks")[0].get("Lifecycle")
+        time.sleep(5)
+
+    assert_that(status).is_equal_to("SUCCEEDED")
+
+    remote_command_executor.run_remote_command(
+        "aws s3 cp s3://{bucket_name}/export_dir/file_to_export ./file_to_export".format(bucket_name=bucket_name)
+    )
+    result = remote_command_executor.run_remote_command("cat ./file_to_export")
+    assert_that(result.stdout).is_equal_to("Exported by FSx Lustre")
```
tilne: minor: it looks like you use this string a few places in your test. it might be nice to just define a constant for it.

tests/integration-tests/tests/storage/test_fsx_lustre.py
sean-smith: changed

tests/integration-tests/tests/storage/test_fsx_lustre.py
sean-smith: done

tests/integration-tests/tests/storage/test_fsx_lustre.py
sean-smith: done

sean-smith: > would it be good to include a `CANCELLING` here as well? within the context of the test it probably doesn't matter, but I'm wondering what would happen if this test existed while the task was still cancelling and the cluster was deleted.

great idea

PR #1657, opened 2020-02-20
Fix bucket creation utility function

enrico-usai:
From now the `generate_random_bucket_name` function will raise the `ClientError` exception in each case and it is up to the caller to manage the different cases.

The function is only used to create the internal s3 bucket used for the awsbatch scheduler.

Since the awsbatch bucket is removed a cluster deletion time I removed the condition to skip the `BucketAlreadyOwnedByYou` exception, because we need to be sure that the bucket belongs to a specific cluster and is not a bucket previously created, already owned by the user.

Then I split the `try` blocks for the cluster creation and the resource artifacts upload, because it can happen that the bucket creation fails and the code in the catch block was trying to delete a nonexistent bucket.

Changed the log level to `CRITICAL`.

Refactored batch bucket creation function to be more testable:
* Moved bucket name generation in a separate function: `generate_random_bucket_name`.
* Moved resources folder discovery into the `create_bucket_with_batch_resources` function.

Added unit test for: 
* `generate_random_bucket_name`,
* `create_s3_bucket` 
* `create_bucket_with_batch_resources`

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -20,9 +20,7 @@ import datetime
 import json
 import logging
 import os
-import random
 import shlex
-import string
 import subprocess as sub
 import sys
 import tarfile
@@ -48,20 +46,29 @@ else:
 LOGGER = logging.getLogger(__name__)
 
 
-def _create_bucket_with_batch_resources(stack_name, resources_dir, region):
-    random_string = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(16))
-    # bucket names must be at least 3 and no more than 63 characters long
-    s3_bucket_name = "-".join([stack_name.lower()[: 63 - len(random_string) - 1], random_string])
+def _create_bucket_with_batch_resources(stack_name, region):
+    """
+    Create a bucket associated to the given stack and upload batch resources.
+
+    Returns the bucket name if both creation and upload succeed.
+    """
+    batch_resources = pkg_resources.resource_filename(__name__, "resources/batch")
+    s3_bucket_name = utils.generate_random_bucket_name(stack_name)
+    LOGGER.debug("Creating S3 bucket for AWS Batch resources, named %s", s3_bucket_name)
 
     try:
-        utils.create_s3_bucket(bucket_name=s3_bucket_name, region=region)
-        utils.upload_resources_artifacts(bucket_name=s3_bucket_name, root=resources_dir)
-    except boto3.client("s3").exceptions.BucketAlreadyExists:
-        LOGGER.error("Bucket %s already exists. Please retry cluster creation.", s3_bucket_name)
+        utils.create_s3_bucket(s3_bucket_name, region)
+    except ClientError:
+        LOGGER.error("Unable to create S3 bucket %s.", s3_bucket_name)
         raise
+
+    try:
+        utils.upload_resources_artifacts(s3_bucket_name, root=batch_resources)
     except Exception:
-        utils.delete_s3_bucket(bucket_name=s3_bucket_name)
+        LOGGER.error("Unable to upload AWS Batch resources to the S3 bucket %s.", s3_bucket_name)
+        utils.delete_s3_bucket(s3_bucket_name)
         raise
+
     return s3_bucket_name
 
 
@@ -91,18 +98,15 @@ def create(args):  # noqa: C901 FIXME!!!
 
     _check_for_updates(pcluster_config)
 
-    batch_temporary_bucket = None
+    batch_bucket_name = None
     try:
         cfn_client = boto3.client("cloudformation")
         stack_name = utils.get_stack_name(args.cluster_name)
 
         # If scheduler is awsbatch create bucket with resources
         if cluster_section.get_param_value("scheduler") == "awsbatch":
-            batch_resources = pkg_resources.resource_filename(__name__, "resources/batch")
-            batch_temporary_bucket = _create_bucket_with_batch_resources(
-                stack_name=stack_name, resources_dir=batch_resources, region=pcluster_config.region
-            )
-            cfn_params["ResourcesS3Bucket"] = batch_temporary_bucket
+            batch_bucket_name = _create_bucket_with_batch_resources(stack_name, pcluster_config.region)
+            cfn_params["ResourcesS3Bucket"] = batch_bucket_name
 
         LOGGER.info("Creating stack named: %s", stack_name)
         LOGGER.debug(cfn_params)
@@ -144,22 +148,21 @@ def create(args):  # noqa: C901 FIXME!!!
     except ClientError as e:
         LOGGER.critical(e.response.get("Error").get("Message"))
         sys.stdout.flush()
-        if batch_temporary_bucket:
-            utils.delete_s3_bucket(bucket_name=batch_temporary_bucket)
+        if batch_bucket_name:
+            utils.delete_s3_bucket(batch_bucket_name)
         sys.exit(1)
     except KeyboardInterrupt:
         LOGGER.info("\nExiting...")
         sys.exit(0)
     except KeyError as e:
-        LOGGER.critical("ERROR: KeyError - reason:")
-        LOGGER.critical(e)
-        if batch_temporary_bucket:
-            utils.delete_s3_bucket(bucket_name=batch_temporary_bucket)
+        LOGGER.critical("ERROR: KeyError - reason:\n%s", e)
+        if batch_bucket_name:
+            utils.delete_s3_bucket(batch_bucket_name)
         sys.exit(1)
     except Exception as e:
         LOGGER.critical(e)
-        if batch_temporary_bucket:
-            utils.delete_s3_bucket(bucket_name=batch_temporary_bucket)
+        if batch_bucket_name:
+            utils.delete_s3_bucket(batch_bucket_name)
         sys.exit(1)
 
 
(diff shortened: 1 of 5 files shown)


tilne:
Worth adding unit tests?

cli/pcluster/utils.py
```diff
@@ -105,22 +107,31 @@ def paginate_boto3(method, **kwargs):
             yield result
 
 
+def generate_random_bucket_name(bucket_name_prefix):
+    """
+    Generate a random bucket name, with the given prefix.
+
+    Bucket name must be at least 3 and no more than 63 characters long.
+    """
+    random_string = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(16))
```
ddeidda: Could you please add a small comment / example of generated bucket name? Is this mechanism safe enough to avoid name collisions?

cli/pcluster/commands.py
```diff
@@ -48,21 +46,24 @@
...
-    except boto3.client("s3").exceptions.BucketAlreadyExists:
-        LOGGER.error("Bucket %s already exists. Please retry cluster creation.", s3_bucket_name)
+    except ClientError:
+        LOGGER.critical("Unable to create internal bucket %s. Please retry cluster creation.", s3_bucket_name)
         raise
+
+    try:
+        utils.upload_resources_artifacts(bucket_name=s3_bucket_name, root=resources_dir)
     except Exception:
         utils.delete_s3_bucket(bucket_name=s3_bucket_name)
```
ddeidda: Why is the bucket still deleted here? 
Probably the problem here was to make the function return True to allow bucket deletion outside even if the upload of resources failed. An idea could be to catch the exceptions locally and just return a tuple Boolean, Boolean with bucket creation result and bucket upload result. Or also trap the error from delete_s3_bucket.

cli/pcluster/commands.py
```diff
@@ -144,22 +147,22 @@ def create(args):  # noqa: C901 FIXME!!!
     except ClientError as e:
         LOGGER.critical(e.response.get("Error").get("Message"))
         sys.stdout.flush()
-        if batch_temporary_bucket:
-            utils.delete_s3_bucket(bucket_name=batch_temporary_bucket)
+        if batch_bucket_created:
```
ddeidda: Look like a bit of code duplication. Could this block be be moved into a finally block?

cli/tests/pcluster/test_commands.py
```diff
@@ -0,0 +1,58 @@
...
+        ("eu-west-1", "An error occurred", "Unable to create internal bucket"),
+        # TODO add a working test
+    ],
+)
+def test_create_bucket_with_batch_resources(region, client_error, log_message, boto3_stubber, caplog):
+    """Verify that create_bucket_with_batch_resources behaves as expected."""
+    bucket_name = "test"
+    resources_dir = "."
+    expected_params = {"Bucket": bucket_name}
+    if region != "us-east-1":
```
ddeidda: Could you explain here with a comment the reason of this region check?

cli/pcluster/commands.py
enrico-usai: It's a good idea, unfortunately the `KeyboardInterrupt` has a different behaviour, so we cannot put the code in common.

cli/tests/pcluster/test_commands.py
enrico-usai: done

cli/pcluster/utils.py
enrico-usai: Example added.
This random generation is using the cluster name as prefix and adding a random string of 16 characters.

Clearly a name collision could happen, in that case the entire stack creation will fail logging an error and the exception, before this patch it was failing without a clear message for the user.

demartinofra: it's not perfect but if you only consider the randomly generated substring (last 16 chars) we have the permutation of 36 unique chars (string.ascii_lowercase + string.digits) over 16 positions which gives us 7.9586611099e+24 unique strings and if we assume that the random function has a uniform distribution (which is not completely true but I think we are quite close to that) then probability of a collision is really low.
If we think this is not enough we could handle the error code saying that the bucket name is already in use and generate a different name.

cli/pcluster/commands.py
```diff
@@ -48,20 +46,31 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def _create_bucket_with_batch_resources(stack_name, resources_dir, region):
-    random_string = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(16))
-    # bucket names must be at least 3 and no more than 63 characters long
-    s3_bucket_name = "-".join([stack_name.lower()[: 63 - len(random_string) - 1], random_string])
+def create_bucket_with_batch_resources(stack_name, region):
```
demartinofra: Minor: why removing the leading underscore?

cli/pcluster/commands.py
```diff
@@ -48,20 +46,31 @@
...
+    LOGGER.debug("Creating batch bucket named %s", s3_bucket_name)
 
     try:
-        utils.create_s3_bucket(bucket_name=s3_bucket_name, region=region)
-        utils.upload_resources_artifacts(bucket_name=s3_bucket_name, root=resources_dir)
-    except boto3.client("s3").exceptions.BucketAlreadyExists:
-        LOGGER.error("Bucket %s already exists. Please retry cluster creation.", s3_bucket_name)
+        utils.create_s3_bucket(s3_bucket_name, region)
+    except ClientError:
+        LOGGER.critical("Unable to create internal bucket %s. Please retry cluster creation.", s3_bucket_name)
```
demartinofra: I would keep this as error and print the message that comes with ClientError. A retry is not always the solution, for example missing IAM permissions might trigger the same error. Printing the message associated to ClientError might help debugging the root cause.

cli/pcluster/commands.py
```diff
@@ -48,20 +46,31 @@
...
         raise
+
+    try:
+        utils.upload_resources_artifacts(s3_bucket_name, root=batch_resources)
     except Exception:
-        utils.delete_s3_bucket(bucket_name=s3_bucket_name)
+        LOGGER.critical(
```
demartinofra: Same here I would say

Minor: upload to the

cli/pcluster/utils.py
```diff
@@ -136,7 +148,7 @@ def delete_s3_bucket(bucket_name):
     except boto3.client("s3").exceptions.NoSuchBucket:
         pass
     except ClientError:
-        print("Failed to delete bucket %s. Please delete it manually." % bucket_name)
+        LOGGER.critical("Failed to delete bucket %s. Please delete it manually." % bucket_name)
```
demartinofra: What about setting this as a warning and also printing the reason why the deletion failed?

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -343,3 +344,40 @@ def _generate_stack_event():
...
     }
+
+
+@pytest.mark.parametrize(
+    "bucket_prefix", ["test", "test-", "prefix-63-characters-long--------------------------------to-cut"]
+)
+def test_generate_random_bucket_name(bucket_prefix):
+    bucket_name = utils.generate_random_bucket_name(bucket_prefix)
+    assert_that(bucket_name).starts_with(bucket_prefix[: 63 - len(bucket_prefix) - 16])
+    assert_that(len(bucket_name)).is_between(3, 63)
```
demartinofra: Why not asserting the exact expected length?

cli/tests/pcluster/test_commands.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.parametrize(
+    "region,error,action",
+    [
+        ("eu-west-1", "create_bucket", "BucketAlreadyExists"),  # bucket creation failure
+        ("eu-west-1", "upload_fileobj", "ExpiredToken"),  # resources upload failure
+        ("eu-west-1", "delete", "AccessDenied"),  # bucket deletion failure
+        ("eu-west-1", None, None),
+    ],
+)
+def test_create_bucket_with_batch_resources(region, action, error, mocker, caplog):
```
demartinofra: Optional: in this casa separate test functions rather than a parametrize one might have simplified the test function

cli/pcluster/commands.py
enrico-usai: Because now it's included from unit tests.

cli/pcluster/commands.py
enrico-usai: I changed it to critical because the cluster creation will fail.

In any case this method raises a `ClientError` exception catched in the `create` method and [displayed as critical](https://github.com/aws/aws-parallelcluster/blob/v2.6.0/cli/pcluster/commands.py#L145) so we can avoid to show the reason here, it's managed by the caller.

cli/pcluster/commands.py
enrico-usai: 1. See previous comment for the exception management.
2. in --> to --> DONE

cli/tests/pcluster/utils/test_pcluster_utils.py
enrico-usai: test added

cli/tests/pcluster/test_commands.py
enrico-usai: you right. splitted.

cli/pcluster/commands.py
demartinofra: imo this should not determine the scope of the method. If it's a "private" function in our application we should keep it as such and include it with the trailing underscore in the unit tests

cli/pcluster/utils.py
enrico-usai: done

PR #1665, opened 2020-02-25
Test for Default FSx Configuration

sean-smith:


sean-smith:
When a stack is created with the following config:

```ini
[fsx test5]
shared_dir = /fsx
storage_capacity = 3600
```
The parameter passed to cloudformation is as follows:

image

This parameter means that the [DeploymentType](https://github.com/aws/aws-parallelcluster/blob/develop/cloudformation/fsx-substack.cfn.json#L299) cloudformation parameter is set to `AWS::NoValue` which according to the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-fsx-filesystem-lustreconfiguration.html#cfn-fsx-filesystem-lustreconfiguration-deploymenttype) is a optional parameter.

PR #1673, opened 2020-03-03
Improve log retrieval script

tilne:
Changes made in this commit:
* Specify the region in which the script was looking for a cluster's log
group if it can't be found.
* Enable configuration of output path for resulting archive.
* Minor grammatical changes to a couple log messages.
* Delete objects CloudWatch exports to S3 unless user requests
otherwise.

When this is merged I'll also need to add `s3:DeleteObject` to the list of required permissions in [the corresponding wiki page](https://github.com/aws/aws-parallelcluster/wiki/Creating-an-Archive-of-a-Cluster's-Logs)

### DIFF
--- a/util/logging/retrieve-cluster-logs.py
+++ b/util/logging/retrieve-cluster-logs.py
@@ -30,16 +30,14 @@ import boto3
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger(__file__)
 
+DEFAULT_BUCKET_PREFIX_FORMAT = "{{cluster_name}}-logs-{timestamp}".format(timestamp=datetime.now().timestamp())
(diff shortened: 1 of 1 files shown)


util/logging/retrieve-cluster-logs.py
```diff
@@ -30,18 +30,16 @@
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger(__file__)
 
+DEFAULT_BUCKET_PREFIX_FORMAT = "{{cluster_name}}-logs-{timestamp}".format(timestamp=datetime.now().timestamp())
```
rexcsn: Interesting use of python string formatting 👍

demartinofra: is this compatible with Python 2.7. I know this script is not part of our CLI but it would be nice to guarantee same python compatibility if possible

tilne: Yes it is.

PR #1676, opened 2020-03-03
Reduce number of CloudWatch Logs integration tests

tilne:
Reduce the number of CloudWatch logs integration tests run from 80 to
10:
* Remove dimension for enabling CloudWatch logging. Only test when the case where the feature is enabled.
* Remove dimension for keeping logs after a cluster is deleted. Only test the case where logs are kept.
* Don't test on centos 6, amazon linux 1, or ubuntu1604.
  * Since the logic for selecting logs of interest relies on platform and not a specific OS version, this won't result in any real loss of coverage.

### DIFF
--- a/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
+++ b/tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
@@ -555,16 +555,7 @@ class FeatureSpecificCloudWatchLoggingTestRunner(CloudWatchLoggingTestRunner):
         assert_that(observed_stream_names).contains(*expected_stream_index)
 
     @classmethod
-    def run_tests_for_feature(
-        cls,
-        cluster,
-        scheduler,
-        os,
-        feature_key,
-        region,
-        shared_dir=DEFAULT_SHARED_DIR,
-        retention_days=DEFAULT_RETENTION_DAYS,
-    ):
+    def run_tests_for_feature(cls, cluster, scheduler, os, feature_key, region, shared_dir=DEFAULT_SHARED_DIR):
         """Verify that the logs for the given feature are present on the cluster and are stored in cloudwatch."""
         environ["AWS_DEFAULT_REGION"] = region
         cluster_logs_state = CloudWatchLoggingClusterState(
@@ -574,7 +565,7 @@ class FeatureSpecificCloudWatchLoggingTestRunner(CloudWatchLoggingTestRunner):
         test_runner.run_tests(cluster_logs_state)
 
 
-def get_config_param_vals(cw_logging_enabled):
+def get_config_param_vals():
     """Return a dict used to set values for config file parameters."""
     # Allow certain params to be set via environment variable in case manually re-testing on an existing cluster
     retention_days = int(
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -584,51 +575,41 @@ def get_config_param_vals(cw_logging_enabled):
...
-    return {"enable": str(cw_logging_enabled).lower(), "retention_days": retention_days, "queue_size": queue_size}
+    return {"enable": "true", "retention_days": retention_days, "queue_size": queue_size}
 
 
-@pytest.mark.parametrize("cw_logging_enabled", [True, False])
-@pytest.mark.parametrize("cw_logs_persist_after_delete", [True, False])
-@pytest.mark.regions(["us-east-1", "cn-north-1", "us-gov-west-1"])
+@pytest.mark.regions(["us-east-1"])
+@pytest.mark.oss(["centos7", "ubuntu1804", "alinux2"])
 @pytest.mark.instances(["t2.micro", "c5.xlarge"])
```
sean-smith: Can we just do c5's?

tilne: Done.

tilne:
Making sure tests pass before merging. Thanks for the review Sean.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -584,51 +575,41 @@ def get_config_param_vals(cw_logging_enabled):
...
-    pcluster_config_reader,
-    clusters_factory,
-    cw_logging_enabled,
-    cw_logs_persist_after_delete,
-):
+    return {"enable": "true", "retention_days": retention_days, "queue_size": queue_size}
+
+
+@pytest.mark.regions(["us-east-1"])
+@pytest.mark.oss(["centos7", "ubuntu1804", "alinux2"])
```
demartinofra: I'd probably keep testing on all OSes. At most if schedulers logs are OS agnostic we could test only slurm on all OSes and then sge and torque on a single one

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -584,51 +575,41 @@ def get_config_param_vals(cw_logging_enabled):
...
     cluster = clusters_factory(cluster_config)
     test_runner = CloudWatchLoggingTestRunner(
-        _get_log_group_name_for_cluster(cluster.name),
-        cw_logging_enabled,
-        config_params.get("retention_days"),
-        cw_logs_persist_after_delete,
+        log_group_name=_get_log_group_name_for_cluster(cluster.name),
+        enabled=True,
+        retention_days=config_params.get("retention_days"),
+        logs_persist_after_delete=True,
```
demartinofra: wondering if we should have this set to False so that we test the main use case and the proper logs cleanup. At most we could have a separate test that does not iterate on all dimensions (but eventually includes awsbatch) only to test log retention

tilne: Why not test this as well? I don't think it adds a lot of time given that the delete operation will need to take place anyways.

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
tilne: The reason I did this is because for each scheduler the same log streams should be created for the pairs (centos 6 , centos 7), (ubuntu16, ubuntu18), and (AL1, AL2).

tilne: I made this change. AL2 is tested with all the schedulers and the other OSes are tested with only slurm.

PR #1677, opened 2020-03-03
Add slurm scaling test cases for manual actions

rexcsn:
* Test that node is replaced if node is terminated manually
* Test that node is replaced if node is placed in down state
* Test that node is NOT replaced if node is placed in drain and there is job running on the node
* Add assertion to make sure nodewatcher will replace compute node when slurmd is terminated

### DIFF
--- a/tests/integration-tests/tests/common/compute_logs_common.py
+++ b/tests/integration-tests/tests/common/compute_logs_common.py
@@ -17,11 +17,15 @@ from time_utils import minutes, seconds
 @retry(
     retry_on_exception=lambda exception: isinstance(exception, RemoteCommandExecutionError),
     wait_fixed=seconds(30),
-    stop_max_delay=minutes(10),
+    stop_max_delay=minutes(15),
 )
-def wait_compute_log(remote_command_executor):
+def wait_compute_log(remote_command_executor, expected_num_nodes=1):
+    """Return list of compute node instance_ids in case of failure."""
     remote_command_executor.run_remote_command("test -d /home/logs/compute", log_error=False)
-    # return instance-id
-    return remote_command_executor.run_remote_command(
-        "find /home/logs/compute/ -type f -printf '%f\\n' -quit  | head -1 | cut -d. -f1", log_error=False
-    ).stdout
+    output = remote_command_executor.run_remote_command("ls /home/logs/compute/", log_error=False).stdout
+    # sample output: "i-049ce596aa69ac988.tar.gz  i-064f07c373d926ba4.tar.gz"
+    instance_ids = [instance.replace(".tar.gz", "") for instance in output.split()]
+    # make sure we got all the expected failing compute nodes
+    if len(instance_ids) != expected_num_nodes:
+        raise RemoteCommandExecutionError(result="Not enough nodes in /home/logs/compute/")
+    return instance_ids
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -105,6 +105,16 @@ class SchedulerCommands(metaclass=ABCMeta):
         """Get number of slots per instance."""
         pass
 
+    @abstractmethod
+    def set_nodes_state(self, compute_nodes, state):
+        """Set nodes to down state in scheduler"""
+        pass
+
+    @abstractmethod
+    def get_nodes_status(self):
+        """Retrieve node state/status from scheduler"""
+        pass
+
 
 class AWSBatchCommands(SchedulerCommands):
     """Implement commands for awsbatch scheduler."""
@@ -152,6 +162,14 @@ class AWSBatchCommands(SchedulerCommands):
         raise NotImplementedError
 
     def get_node_cores(self):  # noqa: D102
+        raise NotImplementedError
+
+    def set_nodes_state(self, compute_nodes, state):
+        """Not implemented."""
+        raise NotImplementedError
+
+    def get_nodes_status(self):
+        """Not implemented."""
         raise NotImplementedError
 
 
@@ -236,6 +254,14 @@ class SgeCommands(SchedulerCommands):
         """Return number of slots from the scheduler."""
         result = self._remote_command_executor.run_remote_command("qhost -F | grep hl:m_core")
         return re.search(r"hl:m_core=(\d+).000000", result.stdout).group(1)
+
+    def set_nodes_state(self, compute_nodes, state):
+        """Not implemented."""
+        raise NotImplementedError
+
+    def get_nodes_status(self):
+        """Not implemented."""
+        raise NotImplementedError
 
 
 class SlurmCommands(SchedulerCommands):
@@ -332,6 +358,24 @@ class SlurmCommands(SchedulerCommands):
         """Cancel a job"""
         return self._remote_command_executor.run_remote_command("scancel {}".format(job_id))
 
+    def set_nodes_state(self, compute_nodes, state):
+        """Put nodes into down state."""
+        self._remote_command_executor.run_remote_command(
+            "sudo /opt/slurm/bin/scontrol update NodeName={} state={} reason=testing".format(
+                ",".join(compute_nodes), state
+            )
+        )
+
+    def get_nodes_status(self, nodes):
+        """Retrieve node state/status from scheduler"""
+        result = self._remote_command_executor.run_remote_command(
+            "/opt/slurm/bin/sinfo -O nodehost,statelong | tail -n +2"
+        ).stdout.splitlines()
+        current_node_states = {}
+        for entry in result:
+            current_node_states[entry.split()[0]] = entry.split()[1]
+        return {node: current_node_states.get(node, "Unable to retrieve state") for node in nodes}
+
 
 class TorqueCommands(SchedulerCommands):
     """Implement commands for torque scheduler."""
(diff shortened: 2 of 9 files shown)


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -328,6 +367,30 @@ def get_job_info(self, job_id):
...
         return self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id)).stdout
 
+    def set_nodes_to_down_state(self, nodes):
+        """Put nodes into down state."""
+        for node in nodes:
+            self._remote_command_executor.run_remote_command(
+                "sudo /opt/slurm/bin/scontrol update NodeName={0} state=down reason=testing".format(node)
+            )
+
+    def set_nodes_to_debug_state(self, nodes):
```
rexcsn: suspend state rather than debug

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -84,16 +88,133 @@ def test_nodewatcher_terminates_failing_node(scheduler, region, pcluster_config_
...
+    )
+
+    logging.info("Testing that nodes are NOT terminated when set to debug state and there is job running on the nodes")
+    scheduler_commands.submit_command(command="sleep 100000", nodes=num_compute_nodes)
+    # sleep for 10 seconds to avoid case of node is put into drain before job is assigned to the node
+    time.sleep(10)
+    # Set n-1 nodes to drain
+    _set_nodes_to_debug_state_manually(scheduler_commands, compute_nodes[:-1])
+    # assert all nodes are retained correctly
+    _assert_nodes_not_terminated_by_nodewatcher(scheduler_commands, compute_nodes)
```
rexcsn: Test that nodes are terminated after job finishes

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -84,16 +88,133 @@ def test_nodewatcher_terminates_failing_node(scheduler, region, pcluster_config_
     assert_instance_replaced_or_terminating(instance_id, region)
     # verify that desired capacity is still 1
     assert_that(get_desired_asg_capacity(region, cluster.cfn_name)).is_equal_to(1)
-    _assert_nodes_removed_from_scheduler(scheduler_commands, compute_nodes)
+    _assert_nodes_removed_and_replaced_in_scheduler(
```
rexcsn: make this more complicated, 5 nodes kill 4, so on

tests/integration-tests/tests/common/compute_logs_common.py
```diff
@@ -17,11 +17,15 @@
 @retry(
     retry_on_exception=lambda exception: isinstance(exception, RemoteCommandExecutionError),
     wait_fixed=seconds(30),
-    stop_max_delay=minutes(10),
+    stop_max_delay=minutes(15),
 )
-def wait_compute_log(remote_command_executor):
+def wait_compute_log(remote_command_executor, expected_num_nodes=1):
```
demartinofra: ok keeping this for this PR. But we should remove these logs now that we have CW, hence updating these tests.

tests/integration-tests/tests/common/compute_logs_common.py
```diff
@@ -17,11 +17,15 @@
...
     remote_command_executor.run_remote_command("test -d /home/logs/compute", log_error=False)
-    # return instance-id
-    return remote_command_executor.run_remote_command(
-        "find /home/logs/compute/ -type f -printf '%f\\n' -quit  | head -1 | cut -d. -f1", log_error=False
-    ).stdout
+    output = remote_command_executor.run_remote_command("ls /home/logs/compute/", log_error=False).stdout
+    # sample output: "i-049ce596aa69ac988.tar.gz  i-064f07c373d926ba4.tar.gz"
+    instance_ids = [instance.replace(".tar.gz", "") for instance in output.split()]
+    # make sure we got all the expected failing compute nodes
+    if len(instance_ids) != expected_num_nodes:
```
demartinofra: is relying on the number enough? What if you reuse the same cluster for multiple tests? Would it make sense to pass instance ids to check?

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -332,6 +371,30 @@ def cancel_job(self, job_id):
         """Cancel a job"""
         return self._remote_command_executor.run_remote_command("scancel {}".format(job_id))
 
+    def set_nodes_to_down_state(self, nodes):
```
demartinofra: nit: single function set_nodes_state?

also I think you can pass multiple nodes as NodeName so that you can run the remote command just once.

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -69,31 +72,195 @@ def test_multiple_jobs_submission(scheduler, region, pcluster_config_reader, clu
 @pytest.mark.usefixtures("region", "os", "instance")
 @pytest.mark.nodewatcher
 def test_nodewatcher_terminates_failing_node(scheduler, region, pcluster_config_reader, clusters_factory, test_datadir):
-    cluster_config = pcluster_config_reader()
+    initial_queue_size = 5 if scheduler == "slurm" else 1
```
demartinofra: why do we need this many for slurm? and why a different number for other scheduler? let's at least drop a comment and if possible let's use fewer nodes.

tests/integration-tests/tests/scaling/test_scaling/test_scaling_with_manual_actions/pcluster.config.ini
```diff
@@ -0,0 +1,19 @@
...
+aws_region_name = {{ region }}
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = {{ scheduler }}
+master_instance_type = {{ instance }}
+compute_instance_type = {{ instance }}
+initial_queue_size = 5
```
demartinofra: same here can we use fewer nodes?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -69,31 +72,195 @@ def test_multiple_jobs_submission(scheduler, region, pcluster_config_reader, clu
...
+
+def _test_replace_terminated_nodes(scheduler_commands, num_compute_nodes, instance_ids):
+    """Test that slurm nodes are replaced if instances are terminated manually."""
+    logging.info("Testing that nodes are replaced when terminated manually")
+    compute_nodes = _assert_initial_conditions(scheduler_commands, num_compute_nodes)
+    instance_ids_to_hostname = get_instance_ids_to_compute_hostnames_dict(instance_ids)
+
+    nodes_to_retain = [instance_ids_to_hostname[instance_ids[0]]]
+    compute_nodes.remove(instance_ids_to_hostname[instance_ids[0]])
+    # terminate n-1 nodes manually
```
demartinofra: can we terminate a single node?
same for the other tests

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -69,31 +72,195 @@ def test_multiple_jobs_submission(scheduler, region, pcluster_config_reader, clu
...
+    """Test that slurm-specific scaling logic is resistent to manual actions and failures."""
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    num_compute_nodes = 5
+
+    instance_ids = get_compute_nodes_instance_ids(cluster.cfn_name, region)
+
+    _test_replace_terminated_nodes(scheduler_commands, num_compute_nodes, instance_ids)
```
demartinofra: shall we run a job on the nodes while testing termination and down state?

rexcsn: Yes that's a good idea! Addressed.

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: General comment: agreed to keep number of nodes during internal discussion

tests/integration-tests/tests/scaling/test_scaling/test_scaling_with_manual_actions/pcluster.config.ini
tilne: Even though this test is only being run for slurm, which we've , it might still be a good idea to parametrize this since you're setting a variable for it anyways in the test.

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -69,31 +72,209 @@ def test_multiple_jobs_submission(scheduler, region, pcluster_config_reader, clu
...
+@pytest.mark.os(["ubuntu1804"])
+@pytest.mark.usefixtures("region", "os", "instance")
+@pytest.mark.scaling_with_manual_actions
+def test_scaling_with_manual_actions(scheduler, region, pcluster_config_reader, clusters_factory):
+    """Test that slurm-specific scaling logic is resistent to manual actions and failures."""
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    num_compute_nodes = 5
```
tilne: I dropped a comment in corresponding config file about why I think it'd be a good idea to pass this to pcluster_config_reader.

tests/integration-tests/utils.py
```diff
@@ -125,6 +125,21 @@ def get_compute_nodes_instance_ids(stack_name, region):
         raise
 
 
+def get_instance_ids_to_compute_hostnames_dict(instance_ids):
+    """Return dict of instanceIDs to hostnames."""
+    try:
+        instance_id_to_hostname = {}
+        ec2_client = boto3.client("ec2")
+        response = ec2_client.describe_instances(InstanceIds=instance_ids).get("Reservations")
```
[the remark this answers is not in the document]
tilne: Ah missed the last sentence in MaxResults. My bad.

PR #1696, opened 2020-03-15
Test MPI stray processes for Slurm

rexcsn:
* MPI job is known to leave stray processes after job termination if slurm process tracking is not setup correctly
* Add test cases for this bug


tests/integration-tests/tests/schedulers/test_slurm/test_slurm_mpi/mpi_job.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+#SBATCH --export=ALL
+#SBATCH --nodes=5
+#SBATCH --ntasks-per-node=3
+#SBATCH --job-name=imb5
+#SBATCH --time=1
+#SBATCH --output=runscript.out
+
+module load intelmpi
```
demartinofra: what about using openmpi that is available on all OSes? 
On centos6 it is probably not available as a modulefile if I recall correctly. Take a look at the `_test_mpi` function in the `common/mpi_common.py` module

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -79,6 +79,67 @@ def test_slurm_gpu(region, pcluster_config_reader, clusters_factory):
...
+    assert_no_errors_in_logs(remote_command_executor, ["/var/log/sqswatcher", "/var/log/jobwatcher"])
+
+
+def _test_mpi_job_termination(remote_command_executor, test_datadir):
+    """Test canceling mpirun job will not leave stray processes."""
+    logging.info("Testing no stray process left behind after mpirun job is terminated")
+    slurm_commands = SlurmCommands(remote_command_executor)
+
+    # Submit mpi_job, leaving 1 vcpu on each node idle so that the process check job can run while mpi_job is running
+    # Note: using 5 compute nodes for mpi_job so that the job takes more time to complete than the jobs's time limit
```
demartinofra: can we use fewer compute nodes? (1 would be ideal) and maybe issue an scancel if we can't rely on the timeout?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -79,6 +79,67 @@ def test_slurm_gpu(region, pcluster_config_reader, clusters_factory):
...
+
+    # Check that mpi processes are terminated
+    _check_mpi_process(remote_command_executor, slurm_commands, test_datadir, after_completion=True)
+
+
+def _check_mpi_process(remote_command_executor, slurm_commands, test_datadir, after_completion):
+    """Submit script and check for MPI processes."""
+    # Clean up old datafiles
+    remote_command_executor.run_remote_command("rm -f /shared/check_proc.out")
+    result = slurm_commands.submit_script(str(test_datadir / "check_mpi_process.sh"))
```
demartinofra: you could also inline this script and use submit_command (the advantage is that we have the command in the code without having to open a separate file)

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -28,7 +28,7 @@
 @pytest.mark.instances(["c5.xlarge"])
 @pytest.mark.schedulers(["slurm"])
 @pytest.mark.usefixtures("os", "instance", "scheduler")
-def test_slurm(region, pcluster_config_reader, clusters_factory):
+def test_slurm(region, os, pcluster_config_reader, clusters_factory, test_datadir):
```
demartinofra: minor: you can remove os from usefixtures if you are declaring it here

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -79,6 +84,53 @@ def test_slurm_gpu(region, pcluster_config_reader, clusters_factory):
...
+    slurm_commands.cancel_job(job_id)
+
+    # Make sure mpirun job is cancelled
+    _assert_job_state(remote_command_executor, job_id, job_state="CANCELLED")
+
+    # Check that mpi processes are terminated
+    _check_mpi_process(remote_command_executor, slurm_commands, test_datadir, num_nodes=2, after_completion=True)
+
+    # Wait for cluster to scale compute back down to 0
+    _assert_no_nodes_in_scheduler(slurm_commands)
```
demartinofra: if this is a precondition for the next test I would move it there, so no matter how we reorder them they should always work.

PR #1707, opened 2020-03-21
Fix scaling manual test

rexcsn:
* Fix scaling manual test by adding region to EC2 calls

### DIFF
--- a/tests/integration-tests/tests/scaling/test_scaling.py
+++ b/tests/integration-tests/tests/scaling/test_scaling.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
 import time
+from os import environ
 
 import boto3
 import pytest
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -126,7 +126,7 @@ def test_scaling_with_manual_actions(scheduler, region, pcluster_config_reader,
 
     instance_ids = get_compute_nodes_instance_ids(cluster.cfn_name, region)
 
-    _test_replace_terminated_nodes(scheduler_commands, num_compute_nodes, instance_ids)
+    _test_replace_terminated_nodes(scheduler_commands, region, num_compute_nodes, instance_ids)
```
tilne: the guidance i've gotten in the past is to just set the AWS_DEFAULT_REGION environ so that you don't have to pass the region arg around.

tilne:
Sorry, I should have caught the region issue during my initial review.

tests/integration-tests/utils.py
```diff
@@ -125,11 +125,11 @@ def get_compute_nodes_instance_ids(stack_name, region):
         raise
 
 
-def get_instance_ids_to_compute_hostnames_dict(instance_ids):
+def get_instance_ids_to_compute_hostnames_dict(instance_ids, region):
```
tilne: Since this is in a utils module, what about making this new arg an optional with a default of None?

PR #1710, opened 2020-03-24
Fix CentOS6 slurm test

rexcsn:
* Change initial queue size for CentOS6 slurm test from 2 to 0 so test can run right away without waiting for a scaledown


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -36,7 +36,10 @@ def test_slurm(region, os, pcluster_config_reader, clusters_factory, test_datadi
     """
     scaledown_idletime = 3
     max_queue_size = 5
-    cluster_config = pcluster_config_reader(scaledown_idletime=scaledown_idletime, max_queue_size=max_queue_size)
+    initial_queue_size = 2 if os != "centos6" else 0
```
tilne: nitpick: since you use the centos6 string literal in two places based on the fact that IMPI isn't available on it, it might make sense to make a list like `doesnt_support_impi = ["centos6"]` and maybe add a comment explaining why that matters for the test.

then the context would be easier to understand for the conditionals that check whether the OS is in that list.

PR #1736, opened 2020-04-16
Add skip_install_recipes=no to integration test config

rexcsn:
* Add skip_install_recipes=no to integration test config when using extra json, so that integration test specification like custom node packages can take effect


tests/integration-tests/conftest.py
```diff
@@ -321,6 +321,8 @@ def _add_custom_packages_configs(cluster_config, request):
                 cluster[extra_json_custom_option] = request.config.getoption(extra_json_custom_option)
                 extra_json["cluster"] = cluster
     if extra_json:
+        # Do not skip install recipes so that custom node package can take effect
+        extra_json["cluster"]["skip_install_recipes"] = "no"
```
tilne, line 325: Is the use of custom node packages the only thing in extra_json that is prevented from working due to skipping install recipes? If so, can this be set conditional on that?

PR #1749, opened 2020-04-28
Add support for S3 URI in custom_chef_cookbook.

### LINKED ISSUE #1640: Support S3 URI for custom_chef_cookbook
**Environment:**
 - aws-parallelcluster-2.5.1

I use a `custom_chef_cookbook` to configure my nodes. However, my organization would like to disallow objects with `public-read` ACLs in S3 buckets. Would it be possible to support S3 URIs for the `custom_chef_cookbook` field?

In the past, I've used presigned URLs to fix that issue with the custom template URL and the additional stack, but the Chef cookbook must be accessible at all times, since new compute nodes are constantly being instantiated.

The only downside I can see is that validation will be a more complicated, as `pcluster` will have to make sure that the MasterServer and the ComputeFleet has read access to the S3 object.

joeydumont:
*Issue #, if available:* Fixes #1640

*Description of changes:* 
I added support for using S3 URI in the `custom_chef_cookbook` section. I modified the CFN template to create a presigned URL when the `custom_cookbook` variable starts with `s3://`. This minimizes the changed needed to the UserData fields.

I have also added a custom validator that checks that the S3 bucket in the URI is also in the `s3_read_resource` field. The assumption is that the primary use of S3 URIs for `custom_chef_cookbook` is for private use, so it would make sense to list the bucket there. More advanced validation based on policies could be performed, if deemed necessary.

I ran `tox -e autoformat` and `tox`. All tests were successful.

I also tried to create a cluster with
```ini
[cluster default]
s3_read_resource = {*, arn:aws:s3:::dumontj-cfn-test2, arn:aws:s3:::dumontj-cfn-test23}
s3_read_write_resource = {*}
custom_chef_cookbook = {s3://dumontj-cfn-test2/cookbooks/aws-parallelcluster-cookbook-2.6.1.tgz}
```
where `{}` indicate a selection. The validation phase emitted warnings as expected.

Note that the names of the bucket cannot be quoted, as quotes are parsed literally by configparser.


cli/pcluster/config/validators.py
```diff
@@ -839,3 +839,41 @@ def intel_hpc_validator(param_key, param_value, pcluster_config):
         )
 
     return errors, warnings
+
+
+def custom_chef_cookbook_validator(param_key, param_value, pcluster_config):
```
enrico-usai, line 844: I think we can move this logic in the `url_validator` given that we're using it for pre_install, post_install and other parameters that can be stored in an S3 bucket.

cli/pcluster/config/validators.py
```diff
@@ -839,3 +839,41 @@ def intel_hpc_validator(param_key, param_value, pcluster_config):
...
+    warnings = []
+
+    if urlparse(param_value).scheme in ["https", "http"]:
+        errors, warnings = url_validator(param_key, param_value, pcluster_config)
+
+    elif urlparse(param_value).scheme == "s3":
+
+        # Check that bucket is in s3_read_resource.
+        cluster_section = pcluster_config.get_section("cluster")
+        s3_read_resource = cluster_section.get_param_value("s3_read_resource")
```
enrico-usai, line 861: I'd check also the `s3_read_write_resource`.

cli/pcluster/config/validators.py
```diff
@@ -839,3 +839,41 @@ def intel_hpc_validator(param_key, param_value, pcluster_config):
...
+            match = re.match(r"s3://(.*?)/(.*)", param_value)
+            if not match or len(match.groups()) < 2:
+                errors.append("S3 url is invalid.")
+            bucket = match.group(1)
+
+            # Parse bucket name from S3 URI.
+            # Match anything after arn prefix, until it hits / or *.
+            match = re.search(r"(?<=arn:aws:s3:::)([^*/]*)", s3_read_resource)
+            if bucket != match.group(0):
+                errors.append("Bucket {} is not in your s3_read_resource parameter.".format(bucket))
```
enrico-usai, line 875: Instead of creating an error, I'd use a warning, like we're doing in the url validator because for example the user could upload the cookbook in a bucket and make it public or use a policy to access to it.

cli/pcluster/config/validators.py
```diff
@@ -839,3 +839,41 @@ def intel_hpc_validator(param_key, param_value, pcluster_config):
...
+    else:
+        errors.append("custom_chef_cookbook only supports HTTP/HTTPS and S3 URIs, was given {}".format(param_value))
```
enrico-usai, line 877: In the messages let's use positional arguments, e.g. `{0}` in place of `{}`.
I know we don't support 2.7 but in this way we'll keep the code consistent.

cli/pcluster/config/validators.py
enrico-usai: Then when adding a new validator, let's also add a test in the `tests/pcluster/config/test_validators.py` module.

We already have the `test_url_validator` so if you're going to merge your code on it you can simply extend it to cover your use cases.

joeydumont: I did move the logic in `url_validator`, but I think that `custom_chef_cookbook` is the only field that accepts `s3://` and also depends on the cluster being able to read the object, so I added a conditional check on `custom_chef_cookbook`.

As for the test, I wanted to test a failure scenario where the cookbook is not in a bucket that the cluster can read (`head_object` fails, and the bucket is not in either `s3_read_resource` or `s3_read_write_resource`). Unfortunately, I couldn't figure out how to make `MockedBoto3Request` raise a `ClientError`. I left the test in the commit, but commented out, in case you can help me with this.

cli/pcluster/config/validators.py
joeydumont: Done.

cli/pcluster/config/validators.py
joeydumont: Done.

cli/pcluster/config/validators.py
joeydumont: Done.

cli/pcluster/config/validators.py
```diff
@@ -593,7 +593,36 @@ def url_validator(param_key, param_value, pcluster_config):
             bucket, key = match.group(1), match.group(2)
             boto3.client("s3").head_object(Bucket=bucket, Key=key)
         except ClientError:
-            warnings.append("The S3 object does not exist or you do not have access to it.")
+
+            # custom_chef_cookbook accepts S3 URIs. Check that the referenced bucket
+            # is in s3_read_resource or s3_read_write_resource.
+            if param_key == "custom_chef_cookbook":
```
enrico-usai, line 599: Let's remove this check. Your improvements are valid in general. 

The s3 uri can be used for pre and post install script parameters too.

cli/pcluster/config/validators.py
```diff
@@ -593,7 +593,36 @@ def url_validator(param_key, param_value, pcluster_config):
...
+                    )
+                    s3_write_bucket = (
+                        re.search(match_bucket_from_arn, s3_read_write_resource).group(0)
+                        if s3_read_write_resource
+                        else None
+                    )
+
+                    if bucket not in [s3_read_bucket, s3_write_bucket]:
+                        warnings.append(
+                            "The S3 object does not exist, or you do not have access it. Consider adding the {0} bucket in either s3_read_resource or s3_read_write_resource.".format(
```
enrico-usai, line 619: `"The S3 object does not exist, or you do not have access it. Consider adding the {0} bucket in either s3_read_resource or s3_read_write_resource."` --> `"The S3 object does not exist, or you do not have access it. Please be sure the cluster nodes have access to it."`  

I'd use a more generic message because there are different ways to achieve the same result. E.g. a custom IAM policy for the instances.

cli/pcluster/config/validators.py
enrico-usai: Thanks for addressing the comments, I added some new minor ones.

There are other parameters accepting `s3://` urls (i.e. `pre_install` and `post_install`) so I think your improvement can be generalized.

To raise an error in the `MockedBoto3Request` you can pass `generate_error=True`, see example [here](https://github.com/aws/aws-parallelcluster/blob/v2.7.0/cli/tests/pcluster/utils/test_pcluster_utils.py#L381).

enrico-usai:
Some minor comments:
* Travis is failing with the message: `pcluster/config/validators.py:592:1: C901 'url_validator' is too complex (11)`.
To fix this you could create a new function to be called in the url_validator one in the s3 case (or eventually add `# noqa: C901` at the end of the row with the function definition)
* Then I'd ask you to [fixup](https://git-scm.com/docs/git-rebase) all the commits into a single one.

Thanks!

joeydumont:
Hi Enrico,

Sorry about the double rebase,  I figured there was a simpler way of doing things than my original solution. I seperated S3 URI validation in its own function to fix the C901 linting error, and rebased on the latest develop.

joeydumont:
@enrico-usai, for context, the clusters I spin up are joined to Active Directory. I use additional recipes to setup/manage `/etc/security/access.conf` and sudoers.d configuration files (among others). I also setup/manage custom `profile.d` files. I also modify the default OpenSSH configuration to allow for X forwarding, and for password authentication to be allowed except for the cluster user.

Also, I wanted to setup X11 forwarding with slurm, but slurm printed warnings when `PrologFlags=X11` was set, which broke nodewatcher. So I patched slurm and modified the slurm recipe to apply the patch before compiling and installing.

I say manage above because I set up `cfn-hup` with a custom `update` configSet to be able to manage all of the configuration items named above (and more) with `pcluster update`.

Hope this helps.

PR #1765, opened 2020-05-07
Removes @timeout decorator.

patemotter:
*Description of changes:*
The Wrapt package has requirements that temporarily install versions of python that are not compatible with a strict 3.6 or 3.7 Python install. This issue was noticed when attempting to create an AWS Lambda layer from ParallelCluster's testing framework. Many other timeout mechanisms exist and the existing package is only used in a single case.

The Wrapt requirements can be seen at https://github.com/bitranox/wrapt_timeout_decorator/blob/master/requirements.txt.


tests/integration-tests/tests/scaling/test_mpi.py
```diff
@@ -69,7 +68,6 @@ def test_mpi_ssh(scheduler, os, pcluster_config_reader, clusters_factory, test_d
     _test_mpi_ssh(remote_command_executor, scheduler, os, test_datadir)
 
 
-@timeout(10, use_signals=False, timeout_exception=RemoteCommandExecutionError)
```
demartinofra, line 72: The two calls to `run_remote_script` might get stuck in case of errors due to the nature of the test. I'm ok with removing this annotation but can we add an optional timeout parameter to `run_remote_command` and `run_remote_command` functions and have this timeout propagated to the run() call? So that in this test function only when we invoke the run_remote_script we also pass `timeout=10`

Here is the doc for the run command which seems to support a timeout argument: http://docs.pyinvoke.org/en/latest/api/runners.html#invoke.runners.Runner.run

patemotter: Added in latest revision

PR #1769, opened 2020-05-09
Make scheduler and OS required parameters

tilne:
Also in this PR:
* print a warning that torque and sge will be deprecated starting in version 3.0.0 when either is used as the scheduler
* tests for the validator that ensures scheduler is required and that prints the deprecation warning
* change to the CFN template such that slurm is the default scheduler and alinux2 is the default OS
* changes required for the other tests


cli/pcluster/config/validators.py
```diff
@@ -745,6 +745,14 @@ def scheduler_validator(param_key, param_value, pcluster_config):
     if pcluster_config.get_section("cluster").get_param_value("base_os") not in supported_os:
         errors.append("'{0}' scheduler supports the following Operating Systems: {1}".format(param_value, supported_os))
 
+    deprecated_in_30 = ["sge", "torque"]
+    if param_value in deprecated_in_30:
```
enrico-usai, line 749: An alternative is to use a new attribute to manage the "deprecated" values too. 
e.g.
`"deprecation": {"values": ["sge", "torque"], "version": "3.0.0", "message": "sometext" }`

Then in the validate() function we can have:
```python
from packaging import version
...
def validate(self):

    deprecation = self.definition.get("deprecation", {})
    if deprecation and self.value in deprecation.get("values", []):

        deprecation_version = deprecation.get("version", None)
        if deprecation_version and version.parse(utils.get_installed_version()) < version.parse(deprecation_version):
             message = deprecation.get("message ", None)
             self.pcluster_config.warn(message or "ERROR: Starting in version {0} the value '{1}' for the parameter '{2}' will no longer be supported".format(deprecation_version, self.value, self.key)) 
       else:
           sys.exit("ERROR: Starting from version {0} the value '{1}' is no longer supported for the parameter '{2}".format(deprecation_version, self.value, self.key))
```

The code is a draft and it could be refactored in a specific function `check_deprecation` called by the validate one, but with an approach like this we'll have this information confinated in the `mappings.py` file and we don't need to change the code once we release the 3.0.0.

This kind of code could be helpful if in future we need to deprecate other values.

Library to use: https://packaging.pypa.io/en/latest/version/#packaging.version.parse

cli/tests/conftest.py
```diff
@@ -87,8 +87,8 @@ def boto3_stubber(mocker, boto3_stubber_path):
     # e.g. boto3.client("ec2", region_name=region, ...) --> x = ec2
     mocked_client_factory.client.side_effect = lambda x, **kwargs: mocked_clients[x]
 
-    def _boto3_stubber(service, mocked_requests):
-        client = boto3.client(service)
+    def _boto3_stubber(service, mocked_requests, region="us-east-1"):
```
enrico-usai, line 90: to be reverted?

cli/pcluster/config/mappings.py
```diff
@@ -460,10 +461,10 @@
                 "allowed_values": ["alinux", "alinux2", "ubuntu1604", "ubuntu1804", "centos6", "centos7"],
             }),
             ("scheduler", {
-                "default": "sge",
                 "cfn_param_mapping": "Scheduler",
                 "allowed_values": ["awsbatch", "sge", "slurm", "torque"],
                 "validators": [scheduler_validator],
+                "type": RequiredParam,
```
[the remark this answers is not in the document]
enrico-usai: Note: if following this approach remember to adapt the test_source_consistency.py too.

cli/tests/pcluster/config/test_required_params.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.mark.parametrize(
+    "cluster_label, expected_message",
+    [
+        ("no_scheduler", "Configuration parameter 'scheduler' must have a value"),
+        ("empty_string_scheduler", "The configuration parameter 'scheduler' has an invalid value"),
+    ],
+)
+def test_param_is_required(mocker, pcluster_config_reader, cluster_label, expected_message):
```
enrico-usai, line 24: If using the "required" attribute we can remove this test module and moving the test of required parameters in the test for the sections. E.g. `test_section_cluster.py` --> `test_cluster_param_from_file`

cli/pcluster/config/validators.py
enrico-usai: an alternative is to define a new validator for the deprecation stuff.

cli/pcluster/config/mappings.py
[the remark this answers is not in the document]
tilne: > By subclassing the (string)Param type we're limiting required parameters for the String type.

Good point. I hadn't considered this.

cli/tests/conftest.py
tilne: Yes.

cli/tests/pcluster/config/test_required_params.py
tilne: I moved the test cases to the same file, but I created a new function for them because the check happens in the validate method, which is not called by assert_param_from_file (wrapped by test_cluster_param_from_file).

cli/pcluster/config/validators.py
tilne: Per a discussion between Enrico and I, we are going to table this idea for now.

cli/pcluster/config/validators.py
```diff
@@ -745,6 +745,14 @@ def scheduler_validator(param_key, param_value, pcluster_config):
...
+        warnings.append(
+            "Starting in version 3.0, {0} will no longer be supported schedulers.".format(
```
tilne, line 751: TODO: clear this message with @nstornetta

tilne: Actually, i'm going to update this to `3.0.0`.

tilne:
Addressing linter issue in #1772 since those files were not touched in this PR.

tilne:
Last push was for the changes to two integration tests that weren't explicitly setting a scheduler.

Sorry for all of the noise on this PR, should have left it in draft until I had those in.

tilne:
> but we can push them in a separate patch

@enrico-usai I'd like to handle it in a different patch if that works for you.

tilne:
Merge is pending a change to the deprecation warning message printed when sge or torque is used as the scheduler.

cli/pcluster/utils.py
```diff
@@ -296,7 +296,8 @@ def get_supported_os(scheduler):
     :param scheduler: the scheduler for which we want to know the supported os
     :return: an array of strings of the supported OSes
     """
-    oses = ["alinux", "alinux2"]
+    # NOTE: alinux2 is first because it should be the default in `pcluster configure`
+    oses = ["alinux2", "alinux"]
     if scheduler != "awsbatch":
```
tilne, line 301: There's got to be a better way to handle this. Would it make sense to define an `easyconfig_default` in mappings.py where easyconfig could get the default value from?

tilne:
Given that we're also changing `base_os` to be a required parameter and we were done with this review minus the deprecation warning, i added the changes required for base_os as a series of 4 more commits on top of this.

cli/tests/pcluster/config/utils.py
```diff
@@ -40,28 +40,55 @@ def get_pcluster_config_example():
     return os.path.join(current_dir, "..", "..", "..", "pcluster", "examples", "config")
 
 
-def assert_param_from_file(mocker, section_definition, param_key, param_value, expected_value, expected_message):
+def get_default_values_for_required_clsuter_section_params(cluster_section_dict, only_if_not_present=False):
```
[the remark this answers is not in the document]
enrico-usai: Then instead of "get" I'd rename the method to "set" or "populate" to be clear we're currently setting parameters in the cluster_section_dict

cli/pcluster/utils.py
enrico-usai: Yes I'd define a mapping like this, but in the `easyconfig.py` module.
Let's use a structure similar to the one used in the funciton `get_default_values_for_required_clsuter_section_params` you defined .

But for now we can keep the code as is, it's up to you.

cli/tests/pcluster/config/utils.py
```diff
@@ -40,28 +40,55 @@ def get_pcluster_config_example():
...
+    Provide default values for required parameters for the cluster section.
+
+    This is useful for the cluster section of the config file because, unlike the CFN template,
+    there are no default values defined in mappings.py or in the defaults module used for testing.
+
+    If only_if_not_present is set, then the default values are only added if the
+    cluster_section_dict does not contain keys for the required parameters.
+    """
+    required_cluster_params = [
+        {"config_key": "scheduler", "value": "slurm"},
```
enrico-usai, line 54: config_key --> key

or we can simplify it with: 
```python
required_cluster_params = [{"scheduler": "slurm"}, { "base_os": "alinux2" }]
```

cli/tests/pcluster/config/utils.py
```diff
@@ -40,28 +40,55 @@ def get_pcluster_config_example():
...
+        {"config_key": "base_os", "value": "alinux2"},
+    ]
+    for required_param in required_cluster_params:
+        if (
```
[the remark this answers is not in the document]
enrico-usai: or if using the simplified `required_cluster_params` structure:
```python
for param_key, param_value in required_cluster_params:
    if only_if_not_present:
        cluster_section_dict.setdefault(param_key), param_value)
    else:
        cluster_section_dict[param_key] = param_value  
```

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -555,6 +557,9 @@ def test_cluster_section_from_200_cfn(mocker, cfn_params_dict, expected_section_
     ],
 )
 def test_cluster_section_from_file(mocker, config_parser_dict, expected_dict_params, expected_message):
+    utils.get_default_values_for_required_clsuter_section_params(
+        config_parser_dict.get("cluster default"), only_if_not_present=True
```
enrico-usai, line 561: I think we can get rid of the `only_if_not_present` parameter by always adding a value in our file used for the tests.
It is set to true only here and I think it is more valuable to have the file populated with the required value than adding them here.

In the other cases `test_cluster_section_to_cfn`, `test_cluster_section_from_241_cfn`, `assert_param_validator` etc I agree this function is useful. 👍

tilne: Okay. Makes sense.

tilne: After looking again, it looks like it's also used in `assert_param_validator` so I'm inclined to keep it.

cli/tests/pcluster/config/utils.py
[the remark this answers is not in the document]
tilne: Done.

cli/tests/pcluster/config/utils.py
tilne: Went with the first option.

cli/tests/pcluster/config/test_param_to_file.py
```diff
@@ -23,8 +23,7 @@
         (CLUSTER, "key_name", None, None),
         (CLUSTER, "key_name", "test", "test"),
         (CLUSTER, "key_name", "NONE", "NONE"),
-        (CLUSTER, "base_os", "alinux", None),  # alinux is the default so it will not be added to the file
-        (CLUSTER, "base_os", "alinux2", "alinux2"),
+        (CLUSTER, "base_os", "alinux", "alinux"),
```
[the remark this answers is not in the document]
tilne: It's a redundant test case now. The two test cases were there to ensure that a non-default is written back, while a default is not. There's no more default anymore so they exercise the same path.

cli/tests/pcluster/config/utils.py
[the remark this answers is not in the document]
tilne: Done.

cli/pcluster/utils.py
tilne: Addressed in the last commit I pushed.

PR #1783, opened 2020-05-14
Improve validator for NICE DCV to show a warning when using micro or nano instances

enrico-usai:
The creation of the stack is not interrupted.

Output:
```
$ pcluster create test-dcv
Beginning cluster creation for cluster: private-dcv2
WARNING: The configuration parameter 'enable' generated the following warnings:
The packages required for desktop virtualization in the selected instance type 't2.micro' 
may cause instability of the master instance. 
If you want to use NICE DCV it is recommended to use an instance 
type with at least 1.7 GB of memory.
Creating stack named: parallelcluster-private-dcv2
...
```

Tested with:
* t2.nano --> warning
* t2.medium --> no warning


tilne:
Can you add a test case for the change to the validator?

cli/pcluster/config/validators.py
```diff
@@ -285,6 +285,14 @@ def dcv_enabled_validator(param_key, param_value, pcluster_config):
         if get_partition() not in get_supported_dcv_partition():
             errors.append("NICE DCV is not supported in the selected region '{0}'".format(get_region()))
 
+        master_instance_type = cluster_section.get_param_value("master_instance_type")
+        if re.search(r"(micro)|(nano)", master_instance_type):
```
rexcsn, line 289: Do we want to add tests for this validator, including a couple of simple test cases to test the default and so on?

enrico-usai: Yes I'll do it.

enrico-usai: Done

PR #1796, opened 2020-05-21
Retrieve MasterIP from stack output when stack status is valid

lukeseawalker:
When Cloudformation stack is in a valid status, the master IP must be retrieved from the stack output.
In this way, the selection to whether to take the public IP or the private IP is left on what is reported by the stack output.

This solves an issue with the combination
use_public_ip = false in the cluster configuration and Auto-assign-ipv4 = true in the master subnet configuration:
in this case the Cloudformation stack output is reporting only the private Master IP, but the IP taken and used by the command `pcluster ssh` and `pcluster dcv connect` was the public Master IP.

*Issue #, if available:*

*Description of changes:*


enrico-usai:
What about adding a unit test for the `get_master_ip_and_username` function in the [test_pcluster_utils.py](https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/utils/test_pcluster_utils.py) file?

lukeseawalker:
PR doesn't work when using two subnets with `use_public_ips= false`: Master in public and Compute in private. Stack output will just report private Master IP, but in order to connecto to the Master instance, the public IP must be used

Issue #1799, opened 2020-05-22
Config file with a .ini extension


sean-smith:
As a feature request, I'm suggesting reading the default config file `~/.parallelcluster/config.ini` with a `.ini` extension. 

Newly created config files, with `pcluster configure`, will be created with this suffix and this will support backwards compatibility for the no-extension `~/.parallelcluster/config`. 

### Why is this useful?

This enables syntax highlighting in your favorite editor, which leads to fewer mistakes in editing the config file. For example, a common action such as commenting out a line, can be done in an editor easily with `Command + /` if the editor knows that the file is a `.ini`.

For example, in VIM:

![image](https://user-images.githubusercontent.com/5545980/82693201-ee087780-9c15-11ea-903d-bfb28c72cd03.png)

Versus

![image](https://user-images.githubusercontent.com/5545980/82693620-b3eba580-9c16-11ea-8d42-3f2be8a9bceb.png)

demartinofra:
Currently this can be achieved with `pcluster create --config ~/.parallelcluster/config.ini`

sean-smith:
@demartinofra Yes that's possible and setting the environment variable is possible, however this feature request is requesting the default gets changed.

enrico-usai:
I'm going to close this patch, since in ParallelCluster 3.0 we are changing the format of the configuration file to be a YAML file. Thanks!

PR #1800, opened 2020-05-22
Use config.ini as default

sean-smith:
https://github.com/aws/aws-parallelcluster/issues/1799


rexcsn:
Hi Sean,

Thanks for opening this PR! The patch presents a short way of defaulting to using `config.ini` config files over the tradition `config` file.

I like the idea of defaulting to files with `.ini` extension for a friendlier user experience.
If we were to make this change, we would have to change to using `config.ini` files everywhere, especially for the default config file produced by our `pcluster configure` functionality.

As you are probably aware, the modifications presented in this patch will not change the behavior of `pcluster configure` and write to `~/.parallelcluster/config.ini` by default.
If the user does not have a config file and runs `pcluster configure`, `_init_config_parser` is called and will still fall back to `~/.parallelcluster/config`, so the config file generated by `pcluster configure` will still be written to `~/.parallelcluster/config`.

Unfortunately, with this inconsistency, I am unable to approve your PR at this time. Please let us know if you have idea on how to handle this issue, and feel free to update or submit new PRs.

Thank you!

PR #1808, opened 2020-05-27
Disable validation for ParallelCluster required additional IAM policies

ddeidda:

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -20,6 +20,7 @@ from botocore.exceptions import ClientError
 from pcluster.constants import CIDR_ALL_IPS
 from pcluster.dcv.utils import get_supported_dcv_os, get_supported_dcv_partition
 from pcluster.utils import (
+    get_base_additional_iam_policies,
     get_efs_mount_target_id,
     get_instance_vcpus,
     get_partition,
@@ -445,9 +446,10 @@ def ec2_iam_policies_validator(param_key, param_value, pcluster_config):
     warnings = []
     try:
         if param_value:
-            iam = boto3.client("iam")
             for iam_policy in param_value:
-                iam.get_policy(PolicyArn=iam_policy.strip())
+                if iam_policy not in get_base_additional_iam_policies():
+                    iam = boto3.client("iam")
+                    iam.get_policy(PolicyArn=iam_policy.strip())
     except ClientError as e:
         errors.append(e.response.get("Error").get("Message"))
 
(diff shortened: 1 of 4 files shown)


cli/pcluster/config/validators.py
```diff
@@ -48,6 +48,11 @@
 
 FSX_SUPPORTED_OSES = ["centos7", "ubuntu1604", "ubuntu1804", "alinux", "alinux2"]
 
+BASE_ADDITIONAL_IAM_POLICIES = [
+    "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy",
```
demartinofra, line 52: please check if this arns are compatible with govcloud and cn regions

ddeidda: You made me also notice that the ``get_partition()`` method was not detecting cn-partition. Please check the updated commit.

cli/pcluster/utils.py
```diff
@@ -93,7 +93,9 @@ def get_region():
 
 def get_partition():
     """Get partition for the AWS_DEFAULT_REGION set in the environment."""
-    return "aws-us-gov" if get_region().startswith("us-gov") else "aws"
+    region = get_region()
+    partition = [partition for partition in ["us-gov", "cn"] if region.startswith(partition + "-")]
```
demartinofra, line 97: I think there is still a problem because we need to prepend aws to the partition: `aws-us-gov`, `aws-cn`

quick tip: you could simplify by using next with something like:
```
return next(("aws" + partition for partition in ["us-gov", "cn"] if region.startswith(partition)), "aws")
```

ddeidda: cool one-liner :)
Just very minor fix: 
```
return next(("aws-" + ...
```

cli/tests/pcluster/config/utils.py
```diff
@@ -111,17 +111,6 @@ def mock_pcluster_config(mocker, scheduler=None):
...
-    accepted_policies = [
-        "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy",
-        "arn:aws:iam::aws:policy/AWSBatchFullAccess",
-    ]
-
-    def mock_iam_policies_validate(self):
-        """Mock validation: just check that the policy is among the accepted ones."""
-        assert set(self.value).issubset(accepted_policies)
-
-    mocker.patch("pcluster.config.param_types.AdditionalIamPoliciesParam.validate", new=mock_iam_policies_validate)
```
ddeidda, line 124: Got rid of mocker since we do not need to validate out IAM policies anymore.

PR #1825, opened 2020-06-05
Remove obsolete unit tests on custer section

ddeidda:
Unit tests on cluster section conversion from cfn templates of old ParallelCluster versions
are now obsolete because loading configuration from different versions is prevented.


ddeidda:
> Let's remove the logic in the EBSSettings.from_cfn_params that was only used to recreate the configuration from old cfn configurations.
> 
> There are two FIXME in the [code](https://github.com/aws/aws-parallelcluster/blob/develop/cli/pcluster/config/param_types.py#L1185) that I think can be now fixed.

I think those FIXME are referred to the possibility of removing the NumberOfEBSVol cfn param, which is a different story. Maybe you were thinking about [this line](https://github.com/aws/aws-parallelcluster/blob/7cfc2a02cfd04bfc9a946361b179d1a588e6c76a/cli/pcluster/config/param_types.py#L1024). BTW I wouldn't remove this line either because I think there are many other tests creating sections in that way. I think in the future we should look for al these tests and modify them to make them creating sections in the canonical way, which is through the PclusterConfig.

PR #1743, opened 2020-04-24
Support for arm instances

tilne:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1828, opened 2020-06-09
Rebase on develop branch

tilne:
The purpose of this PR is to rebase on the large set of changes merged in by #1583 

There are only two differences from the most recent commit merged into the feature branch in #1743. These two changes are both in the most recent commit on this branch. Specifically:
* on line 910 of param_types.py, the leading underscore was removed from `_get_file_section_name`
* and `update_policy` attribute was added to the `architecture` parameter, whose configuration starts at line 835 of mappings.py.


tilne:
force updating the base branch since i've rewritten parts of its history

PR #1830, opened 2020-06-09
ARM compute images on batch

tilne:
* Make dependencies installed in Batch Dockerfile architecture agnostic
* Modify image build accordingly when using ARM compute instances

### DIFF
--- a/cli/tox.ini
+++ b/cli/tox.ini
@@ -211,7 +211,10 @@ changedir =
     ../cloudformation
 deps = cfn-lint
 commands =
-    cfn-lint --info *.cfn.json --ignore-templates fsx-substack.cfn.json
+    # E3030 is ignored because version 0.33.0 labels Environment->Type value of
+    # ARM_CONTAINER for AWS::CodeBuild::Project resource as an invalid value.
+    # See https://github.com/aws-cloudformation/cfn-python-lint/issues/1588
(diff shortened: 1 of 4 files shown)


cli/tox.ini
```diff
@@ -211,7 +211,7 @@ changedir =
     ../cloudformation
 deps = cfn-lint
 commands =
-    cfn-lint --info *.cfn.json --ignore-templates fsx-substack.cfn.json
+    cfn-lint --info *.cfn.json --ignore-templates fsx-substack.cfn.json --ignore-checks E3030
```
demartinofra, line 214: if `E3030` cannot be fixed can we add a comment explaining why?

tilne: Done. Linked to [the issue for it](https://github.com/aws-cloudformation/cfn-python-lint/issues/1588).

tilne:
Clicking through to the travis status link shows that all tests have passed. Because of this and the fact that this is being merged into a feature branch and not develop I'm going to go ahead and merge.

cli/tox.ini
PatMyron: Thanks for the report! [Rules can also be ignored at a resource level](https://github.com/aws-cloudformation/cfn-python-lint/#resource-based-metadata) if you just want to ignore it for one resource

tilne: I wasn't aware of that. Thanks for the tip.

PR #1834, opened 2020-06-11
Move all LogGroup definitions to CW substack

demartinofra:
With this change all LogGroup definitions are moved to the `cw-logs-substack`. The main reason of this change is that it allows an easier and more resilient stack update when `--keep-logs` flag is used and log groups need to be retained. If log groups are spread across several stacks we need to update each of them and this increases the probability that one of those substacks are in a failed state.

Some important details to justify some of the choices:
- When defining a lambda function you cannot set the LogGroup to use but AWS Lambda by default uses a log group named `/aws/lambda/${function_name}`. To avoid introducing too many resources dependencies across stacks the name of the Lambda function is set by us in the resource definition and not randomly generated.
- The name of each Lambda function we define in CloudFormation is composed by using a `parallelcluster-` prefix + an arbitrary function name + the final 12 chars of the CloudFormation stack id that are meant to work as a hash that confers uniqueness to the resource name. The alternative could have been to use the stack name but unfortunately Lambda function names can only be 64 characters long.
- RetentionInDays for all log groups can be now configured through the CW config settings.

Also for some reason when we set the function name a new policy is required in the role for the user that creates the cluster: `lambda:GetFunction`. Need to update public docs before releasing this change.

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -533,20 +533,27 @@ def _persist_stack_resources(stack, template, keys):
     """Set the resources in template identified by keys to have a DeletionPolicy of 'Retain'."""
     for key in keys:
         template["Resources"][key]["DeletionPolicy"] = "Retain"
-    utils.update_stack_template(stack.get("StackName"), template, stack.get("Parameters"))
+    try:
+        utils.update_stack_template(stack.get("StackName"), template, stack.get("Parameters"))
+    except ClientError as client_err:
+        utils.error(
+            "Unable to persist logs on cluster deletion, failed with error: {emsg}.\n"
+            "If you want to continue, please retry without the --keep-logs flag.".format(
+                emsg=client_err.response.get("Error").get("Message")
+            )
+        )
 
 
 def _persist_cloudwatch_log_groups(cluster_name):
     """Enable cluster's CloudWatch log groups to persist past cluster deletion."""
     LOGGER.info("Configuring {0}'s CloudWatch log groups to persist past cluster deletion.".format(cluster_name))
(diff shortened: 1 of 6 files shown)


cli/pcluster/commands.py
```diff
@@ -540,13 +540,12 @@ def _persist_cloudwatch_log_groups(cluster_name):
...
-    ]
-    for stack, template, keys in substack_template_keys_triplets:
-        if keys:  # Only persist the CloudWatch group
-            _persist_stack_resources(stack, template, keys)
+    cw_substack = next((stack for stack in substacks if "CloudWatchLogsSubstack" in stack.get("StackName")), None)
+    if cw_substack:
+        cw_substack_template = utils.get_stack_template(cw_substack.get("StackName"))
+        log_group_keys = _get_unretained_cw_log_group_resource_keys(cw_substack_template)
+        if log_group_keys:  # Only persist the CloudWatch group
+            _persist_stack_resources(cw_substack, cw_substack_template, log_group_keys)
```
tilne, line 548: Is there anything that might cause creation of the CloudWatch substack to fail, causing the update to fail for the same reason that's motivating this change?

Looking at the [CloudWatch quotas](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html) this seems unlikely, as the only thing on that page that seems like it would prevent  the creation of log group would be hitting the limit for number of log groups in a region, and 20,000 is a pretty large number for that.

demartinofra: It's very unlikely but it could still happen. I slightly modified the error message in case of failure so that the user knows that he has to retry without keep-logs if he wants to delete anyway

PR #1835, opened 2020-06-15
Use describe_instance_types to get efa enabled instance types

ddeidda:
This commit replaces the usage of the features json file with a call to describe_instanceTypes
to get the list of efa enabled instance types.

### DIFF
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -24,7 +24,7 @@ def readme():
 VERSION = "2.7.0"
 REQUIRES = [
     "setuptools",
-    "boto3>=1.10.15",
+    "boto3>=1.14.3",
     "future>=0.16.0,<=0.18.2",
     "tabulate>=0.8.2,<=0.8.3",
     "ipaddress>=1.0.22",
--- a/cli/tests/pcluster/config/test_section_cluster.py
+++ b/cli/tests/pcluster/config/test_section_cluster.py
@@ -825,10 +825,7 @@ def test_cluster_from_file_to_cfn(mocker, pcluster_config_reader, settings_label
         "pcluster.config.param_types.get_avail_zone",
         side_effect=lambda subnet: "mocked_avail_zone" if subnet == "subnet-12345678" else "some_other_az",
     )
-    mocker.patch(
-        "pcluster.config.validators.get_supported_features",
-        return_value={"instances": ["t2.large"], "baseos": ["ubuntu1804"], "schedulers": ["slurm"]},
-    )
+
     mocker.patch("pcluster.config.param_types.get_instance_vcpus", return_value=2)
     utils.assert_section_params(mocker, pcluster_config_reader, settings_label, expected_cfn_params)
 
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -26,9 +26,9 @@ from pcluster.utils import (
     get_partition,
     get_region,
     get_supported_compute_instance_types,
-    get_supported_features,
     get_supported_instance_types,
     get_supported_os,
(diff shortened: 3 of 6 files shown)


sean-smith:
Nice!

cli/pcluster/utils.py
```diff
@@ -754,3 +754,18 @@ def get_cluster_capacity(stack_name):
...
         else get_asg_settings(stack_name).get("DesiredCapacity")
     )
+
+
+def get_efa_enabled_instance_types():
+    client = boto3.client("ec2")
+    instance_types = []
+    next_token = ""
+
+    while next_token is not None:
```
demartinofra, line 764: Shall we use the paginator: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Paginator.DescribeInstanceTypes

cli/pcluster/utils.py
```diff
@@ -754,3 +754,18 @@ def get_cluster_capacity(stack_name):
...
+        call_args = {"Filters": [{"Name": "network-info.efa-supported", "Values": ["true"]}], "NextToken": next_token}
+
+        response = client.describe_instance_types(**{k: v for k, v in call_args.items() if v})
```
enrico-usai, line 767: What about adding the arguments directly in the `describe_instance_types` call ?
You don't need to check if `next_token` is `None` because you're inside the `while` that is already checking for it.
Am I missing something? If so, could you please add a comment describing what are you doing within this line?

For the pagination (see @demartinofra suggestion) we have an utility function `paginate_boto3` in the `pcluster/utils.py` file.

ddeidda: Thanks, switched to paginate_boto3.

cli/pcluster/utils.py
ddeidda: Thanks, switched to paginate_boto3.

cli/setup.py
```diff
@@ -24,7 +24,7 @@ def readme():
 VERSION = "2.7.0"
 REQUIRES = [
     "setuptools",
-    "boto3>=1.10.15",
+    "boto3>=1.14.2",
```
demartinofra, line 27: need to update requiremets files

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -825,10 +825,7 @@ def test_cluster_from_file_to_cfn(mocker, pcluster_config_reader, settings_label
         "pcluster.config.param_types.get_avail_zone",
         side_effect=lambda subnet: "mocked_avail_zone" if subnet == "subnet-12345678" else "some_other_az",
     )
-    mocker.patch(
-        "pcluster.config.validators.get_supported_features",
-        return_value={"instances": ["t2.large"], "baseos": ["ubuntu1804"], "schedulers": ["slurm"]},
-    )
+    mocker.patch("pcluster.config.validators.get_efa_enabled_instance_types", return_value=["t2.large"])
```
demartinofra, line 828: can we use the boto3_stubber in the tests to mock the call to describe_instance_types so that we perform some extra validations?

cli/pcluster/utils.py
```diff
@@ -754,3 +754,15 @@ def get_cluster_capacity(stack_name):
         if scheduler == "awsbatch"
         else get_asg_settings(stack_name).get("DesiredCapacity")
     )
+
+
+def get_efa_enabled_instance_types():
+    instance_types = []
+
+    for response in paginate_boto3(
```
demartinofra, line 762: shall we add a try/except to print a specific error message + the message in the ClientError rather than propagating the exception?

PR #1857, opened 2020-06-30
Adds validators for newly added FSx Lustre backup parameters

siddharthsalot:

*Issue #, if available:*

*Description of changes:
Following constraints are added:
* automatic_backup_retention_days can’t be 0 if copy_tags_to_backups or daily_automatic_backup_start_time is specified.
* Parameters related to automatic backup are only valid if deployment_type is PERSISTENT_1.
* S3-linked parameters cannot be specified along with backup parameters in Create File System request.

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -163,6 +163,22 @@ def fsx_validator(section_key, section_label, pcluster_config):
         if fsx_section.get_param_value("per_unit_storage_throughput"):
             errors.append("'per_unit_storage_throughput' can only be used when 'deployment_type = PERSISTENT_1'")
 
+    fsx_daily_automatic_backup_start_time = fsx_section.get_param_value("daily_automatic_backup_start_time")
+    fsx_automatic_backup_retention_days = fsx_section.get_param_value("automatic_backup_retention_days")
+    fsx_copy_tags_to_backups = fsx_section.get_param_value("copy_tags_to_backups")
+
+    if not fsx_automatic_backup_retention_days and (fsx_daily_automatic_backup_start_time or fsx_copy_tags_to_backups):
+        errors.append(
+            "'automatic_backup_retention_days' must be greater than 0 if \
+                'daily_automatic_backup_start_time' or 'copy_tags_to_backups' parameters are provided."
+        )
+
+    if fsx_section.get_param_value("deployment_type") != "PERSISTENT_1" and fsx_automatic_backup_retention_days:
+        errors.append("FSx automatic backup features can be used only with 'PERSISTENT_1' file systems")
+
+    if (fsx_imported_file_chunk_size or fsx_import_path or fsx_export_path) and fsx_automatic_backup_retention_days:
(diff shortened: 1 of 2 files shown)


cli/tests/pcluster/config/test_validators.py
```diff
@@ -679,6 +679,76 @@ def _kms_key_stubber(mocker, boto3_stubber, kms_key_id, expected_message, num_ca
...
+                "import_path": "s3://test",
+                "storage_capacity": 1200,
+            },
+            {"Bucket": "test"},
+            "Backups cannot be created on S3-linked file systems",
+            0,
+        ),
     ],
 )
 def test_fsx_validator(mocker, boto3_stubber, section_dict, bucket, expected_error, num_calls):
```
tilne, line 754: Since we're going to be adding even more validation for this section in the next PRs, it's worth considering if there's a cleaner way to separate this validation into different functions (e.g., one validator for the backup parameters, one validator for the restore from backup parameter constraints, one validator for the deployment type constraints, etc.). But maybe since it's all intertwined keeping it all in the same function is the best way to do it.

Either way, just something to think about. Let's keep the current PR as is.

cli/tests/pcluster/config/test_validators.py
```diff
@@ -679,6 +679,76 @@ def _kms_key_stubber(mocker, boto3_stubber, kms_key_id, expected_message, num_ca
             "'per_unit_storage_throughput' can only be used when 'deployment_type = PERSISTENT_1'",
             0,
         ),
+        (
+            {"storage_capacity": 1200, "deployment_type": "PERSISTENT_1", "automatic_backup_retention_days": 2},
+            None,
+            None,
+            1,
```
tilne, line 686: My understanding is that the `num_calls` value should only be nonzero if either the `export_path` or `import_path` parameter is specified, which isn't the case for the majority of these test cases. Can this be 0?

cli/pcluster/config/validators.py
```diff
@@ -163,6 +163,22 @@ def fsx_validator(section_key, section_label, pcluster_config):
...
             errors.append("'per_unit_storage_throughput' can only be used when 'deployment_type = PERSISTENT_1'")
 
+    fsx_daily_automatic_backup_start_time = fsx_section.get_param_value("daily_automatic_backup_start_time")
+    fsx_automatic_backup_retention_days = fsx_section.get_param_value("automatic_backup_retention_days")
+    fsx_copy_tags_to_backups = fsx_section.get_param_value("copy_tags_to_backups")
+
+    if not fsx_automatic_backup_retention_days and (fsx_daily_automatic_backup_start_time or fsx_copy_tags_to_backups):
+        errors.append(
+            "'automatic_backup_retention_days' must be greater than 0 if \
+                'daily_automatic_backup_start_time' or 'copy_tags_to_backups' parameters are provided."
```
tilne, line 173: If a multiline string is needed, do it without the `\`:
```python
"'automatic_backup_retention_days' must be greater than 0 if "
"'daily_automatic_backup_start_time' or 'copy_tags_to_backups' parameters are provided."
```

PR #1871, opened 2020-07-08
Fix issues in script that generates AMI mappings

tilne:
Before this change, the script would avoid updating a mapping if a new AMI was not found when querying EC2. This change ensures that all entries in mappings specific to the regions specified via the command line arg are re-written entirely. If a an AMI cannot be found for the given region-OS-architecture combination, an "UNSUPPORTED" placeholder value is used.

This PR also fixes an issue where AMI mapping data was being errantly written to other mappings in the CFN template.

There was also some minor refactoring.

### DIFF
--- a/util/generate-ami-list.py
+++ b/util/generate-ami-list.py
@@ -270,8 +270,7 @@ def update_cfn_template(cfn_template_file, amis_to_update):
     # Read in existing CFN template
     cfn_data = read_cfn_template(cfn_template_file)
     # update id for new amis without removing regions that are not in the amis_to_update dict
-    current_amis = get_initialized_mappings_dicts()
-    for mapping_name in current_amis:
+    for mapping_name in ARCHITECTURES_TO_MAPPING_NAME.values():
         current_amis_for_mapping = cfn_data.get("Mappings").get(mapping_name, {})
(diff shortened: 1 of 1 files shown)


util/generate-ami-list.py
```diff
@@ -370,22 +353,32 @@ def update_amis_txt(amis_txt_file, amis):
             if credential_tuple.strip()
         ]
 
-    if (args.version and args.date) or (args.cookbook_git_ref and args.node_git_ref):
-        regions = get_all_aws_regions_from_ec2(region)
-        amis_dict = get_ami_list_from_ec2(
+    if args.version and args.date:
```
rexcsn, line 356: nit: I don't think this logic is used anywhere in Jenkins? I don't think we ever call this script with the version. Can we double check why this is needed?

Ideally it might also help to add comment for when each condition is used in Jenkins and what we are trying to achieve, so we don't have to look at Jenkins script.

For example, we would be using `elif args.cookbook_git_ref and args.node_git_ref` logic in build_and_test to enforce that AMIs needs to be created with the same git ref(and the same build_date actually, since build_date is provided in the call). This might not be immediately clear when we look at this script or the Jenkins file alone.

tilne: This is a very good idea - I found myself doing this a lot while working on it. I'll add another commit with this change.

PR #1875, opened 2020-07-09
Add validator for fsx create-file-system-from-backup parameters

siddharthsalot:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -49,6 +49,8 @@ FSX_MESSAGES = {
         "unsupported_architecture": "FSX Lustre can be used only with instance types and AMIs that support these "
         "architectures: {supported_architectures}. Please double check the 'master_instance_type', "
         "'compute_instance_type' and/or 'custom_ami' configuration parameters.",
+        "unsupported_backup_param": "When restoring an FSx Lustre file system from backup, '{name}' "
+        "cannot be specified.",
     }
 }
 
@@ -177,7 +179,8 @@ def fsx_validator(section_key, section_label, pcluster_config):
 
     if not fsx_automatic_backup_retention_days and fsx_daily_automatic_backup_start_time:
         errors.append(
-            "When specifying 'daily_automatic_backup_start_time', the 'automatic_backup_retention_days' option must be specified"
+            "When specifying 'daily_automatic_backup_start_time', "
+            "the 'automatic_backup_retention_days' option must be specified"
         )
 
     if not fsx_automatic_backup_retention_days and fsx_copy_tags_to_backups is not None:
@@ -280,6 +283,9 @@ def fsx_storage_capacity_validator(section_key, section_label, pcluster_config):
 
     if fsx_section.get_param_value("fsx_fs_id"):
         # if fsx_fs_id is provided, don't validate storage_capacity
+        return errors, warnings
+    elif fsx_section.get_param_value("backup_id"):
+        # if backup_id is provided, validation for storage_capacity will be done in fsx_lustre_backup_validator.
         return errors, warnings
     elif not storage_capacity:
         # if fsx_fs_id is not provided, storage_capacity must be provided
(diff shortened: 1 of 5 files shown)


cli/pcluster/config/validators.py
```diff
@@ -1066,3 +1069,52 @@ def _get_efa_enabled_instance_types(errors):
...
+    try:
+        backups = boto3.client("fsx").describe_backups(BackupIds=[param_value]).get("Backups")
+        if backups is None or len(backups) <= 0:
+            errors.append("Backup '{0}' does not exist.".format(param_value))
+            return errors, warnings
+
+        backup = backups[0]
+        fsx_section = pcluster_config.get_section("fsx")
+        backup_deployment_type = backup.get("FileSystem").get("LustreConfiguration").get("DeploymentType")
+        deployment_type = fsx_section.get_param_value("deployment_type")
```
tilne, line 1087: In my opinion instead of requiring customers to copy the correct values for `deployment_type`, `per_unit_storage_throughput`, and `storage_capacity` the parameters into the config file, we should just populate those parameters using values read from the DescribeBackup API when restoring a backup.

siddharthsalot: Sorry my bad. I think FSx already does inherit deployment_type, per_unit_storage_throughput and storage_capacity from the backup. So we need these checks only when the customer explicitly configures these parameters. There should be None value check for each of these parameters. If its not None, only then we will check for equality.

tilne: In my opinion, if a backup to restore from is specified, then specifying any of the parameters associated with that  backup should be a validation error.

cli/pcluster/config/validators.py
```diff
@@ -49,6 +49,8 @@
         "unsupported_architecture": "FSX Lustre can be used only with instance types and AMIs that support these "
         "architectures: {supported_architectures}. Please double check the 'master_instance_type', "
         "'compute_instance_type' and/or 'custom_ami' configuration parameters.",
+        "unsupported_backup_param_update": "Modifying the {name} '{value}' from the original file system's "
```
tilne, line 52: It would be nice to avoid use of the term update here.

cli/pcluster/config/validators.py
```diff
@@ -1066,3 +1072,59 @@ def _get_efa_enabled_instance_types(errors):
...
+            )
+
+        storage_capacity = fsx_section.get_param_value("storage_capacity")
+        backup_storage_capacity = backup.get("FileSystem").get("StorageCapacity")
+        if storage_capacity is not None:
+            errors.append(
+                FSX_MESSAGES["errors"]["unsupported_backup_param"].format(
+                    name="StorageCapacity", value=backup_storage_capacity
+                )
+            )
```
tilne, line 1116: Can you combine the above into a loop?
```python
config_api_param_name_pairs = [
....
]
for (config_param_name, api_param_name) in config_api_param_name_pairs:
  # make sure config_param_name isn't set in config file. Otherwise append error.
```

cli/pcluster/config/validators.py
```diff
@@ -49,6 +49,8 @@
         "unsupported_architecture": "FSX Lustre can be used only with instance types and AMIs that support these "
         "architectures: {supported_architectures}. Please double check the 'master_instance_type', "
         "'compute_instance_type' and/or 'custom_ami' configuration parameters.",
+        "unsupported_backup_param": "When restoring file system from backup, {name} {value} "
```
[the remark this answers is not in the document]
tilne: One more nitpick on this: instead of saying `restoring file system from backup` can you say `restoring an FSx Lustre file system from backup`? The motivation behind this request is to avoid ambiguity between the other file system types we support.

cli/pcluster/config/validators.py
```diff
@@ -1066,3 +1072,59 @@ def _get_efa_enabled_instance_types(errors):
...
+
+        if fsx_section.get_param_value("import_path"):
+            errors.append("When restoring file system from backup, import_path cannot be specified.")
+
+        if fsx_section.get_param_value("export_path"):
+            errors.append("When restoring file system from backup, export_path cannot be specified.")
+
+        if fsx_section.get_param_value("imported_file_chunk_size"):
+            errors.append("When restoring file system from backup, imported_file_chunk_size cannot be specified.")
+
```
tilne, line 1126: Same for these checks: since the error message is essentially the same for all of them, can you put them in a list and loop through it to make sure each isn't set?

tilne: Instead of saying `restoring file system from backup` can you say `restoring an FSx Lustre file system from backup`?

cli/tests/pcluster/config/test_validators.py
```diff
@@ -1626,3 +1628,93 @@ def test_instances_architecture_compatibility_validator(
...
+                    "LustreConfiguration": {
+                        "WeeklyMaintenanceStartTime": "1:03:00",
+                        "DeploymentType": "PERSISTENT_1",
+                        "PerUnitStorageThroughput": 200,
+                        "AutomaticBackupRetentionDays": 0,
+                    },
+                },
+            },
+        ]
+    }
```
tilne, line 1704: Can the fields that are not crucial for the test be removed from this?

PR #1885, opened 2020-07-16
FSx Section Validation Enhancements

tilne:
* Don't allow use of KMS key parameter in FSx section when restoring from backup
* Add validator that causes FSx validation to fail when parameters that would be ignored are used when using an existing Lustre file system (via the fsx_fs_id parameter)


siddharthsalot:
Reviewed PR. Changes looks good.

tilne:
Rebasing on feature branch

PR #1886, opened 2020-07-16
Fix EFS test in China

rexcsn:
* EFS test would fail in China because AZs are not explicitly provided so subnets would be set up in random AZs, and EFS test is not able to assert the networking assumptions

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -404,6 +404,8 @@ AVAILABILITY_ZONE_OVERRIDES = {
     "sa-east-1": ["sa-east-1a", "sa-east-1c"],
     # m6g.xlarge instances not available in eu-west-1c
     "eu-west-1": ["eu-west-1a", "eu-west-1b"],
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest.py
```diff
@@ -404,6 +404,8 @@ def cfn_stacks_factory(request):
     "sa-east-1": ["sa-east-1a", "sa-east-1c"],
     # m6g.xlarge instances not available in eu-west-1c
     "eu-west-1": ["eu-west-1a", "eu-west-1b"],
+    # provide AZs to enforce networking setup used for EFS test
```
demartinofra, line 407: Can't we simply perform a boto3 call to list the available AZs in a region and randomly select 2 AZs from there when the mapping is not present in `AVAILABILITY_ZONE_OVERRIDES`?  
I'd like to reduce the number of constraints encoded in `AVAILABILITY_ZONE_OVERRIDES` mappings and use it only as a workaround for capacity and resource unavailability.

PR #1891, opened 2020-07-17
Fix EBS snapshot test

rexcsn:
* Revised logic in snapshot factory to alway create new SG with unique name
* Old logic of using existing SG if possible will cause SG deletion failure when 2 tests are running in parallel in the same VPC

### DIFF
--- a/tests/integration-tests/tests/storage/snapshots_factory.py
+++ b/tests/integration-tests/tests/storage/snapshots_factory.py
@@ -17,6 +17,7 @@ import boto3
 from retrying import retry
 
 from fabric import Connection
+from utils import random_alphanumeric
 
 SnapshotConfig = namedtuple("ClusterConfig", ["ssh_key", "key_name", "vpc_id", "master_subnet_id"])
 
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/storage/snapshots_factory.py
```diff
@@ -184,7 +179,7 @@ def _launch_instance(self, ami_id, subnet):
                     "SubnetId": subnet.id,
                     "DeviceIndex": 0,
                     "AssociatePublicIpAddress": True,
-                    "Groups": [self._get_security_group_id()],
+                    "Groups": [self.security_group_id],
```
tilne, line 182: Should this be calling the function, as it was before?

rexcsn: I don't think so, `_launch_instance` is called after `self.security_group_id = self._get_security_group_id()`, and all the `_get_security_group_id()` function does before in context of `_launch_instance` is return the id of SG that was created.

Ideally we should refactor the snapshot factory slightly to move most of attribute initialization in `__init__`, but we will probably do this later

PR #1906, opened 2020-07-28
Unpin isort version and make needed changes

tilne:


cli/.isort.cfg
```diff
@@ -0,0 +1,13 @@
+[settings]
```
demartinofra, line 1: do we need to duplicate this file under the cli dir or the copy in the root dir is enough when running with tox as before?

tilne: I didn't realize we had one in the root directory to be honest.

When I try to move the changes into that file the config doesn't seem to have an effect. Placing it in the CLI dir seems to work.

PR #1911, opened 2020-07-29
Fix efs test in China

yuleiwan:
- previously, we add China region to AVAILABILITY_ZONE_OVERRIDES during subset setup to enable efs test in China. Instead of doing that, we use boto3 to retrieve available zones by region name, if the region is not in AVAILABILITY_ZONE_OVERRIDES dict, then set up subset using az.

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -21,6 +21,7 @@ import re
 from shutil import copyfile
 from traceback import format_tb
 
+import boto3
 import configparser
 import pytest
 from retrying import retry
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest.py
```diff
@@ -422,11 +422,44 @@ def vpc_stacks(cfn_stacks_factory, request):
...
-        availability_zones = random.sample(AVAILABILITY_ZONE_OVERRIDES.get(region, [None, None]), k=2)
+
+        # if region in AVAILABILITY_ZONE_OVERRIDES keys, find the availability_zones
+        if region in AVAILABILITY_ZONE_OVERRIDES:
+            availability_zones = random.sample(AVAILABILITY_ZONE_OVERRIDES.get(region), k=2)
+        # else if region is not in AVAILABILITY_ZONE_OVERRIDES keys, find available zones mapping to the region
+        else:
+            # get ec2 client
+            client = boto3.client('ec2', region_name=region)
+            response_az = client.describe_availability_zones(
```
rexcsn, line 438: Minor: It would be good if we can batch the boto3 call and get all the AZ with one call of `describe_availability_zones`. However I just checked and this is not possible. `describe_availability_zones` only returns AZs in the client's region.

So implementation here is good, no modification needed

PR #1915, opened 2020-07-30
Prevent writing architecture parameter to config

tilne:
Previously the architecture parameter was being written back to the
config files produced by `pcluster configure` when the original
`master_instance_type` supported a different architecture than the one
supported by the `master_instance_type` provided via the interactive
prompt. This change prevents that from occurring.

The same is done for the cluster_config_metadata parameter, although it
was not as critical in this case because its presence in a config file
does not cause it to fail validation.

A test case was added to the corresponding unit tests. To ensure the
test case fails when the fix is not in place, an extra check was added
to ensure that only the expected configuration keys are in the resulting
config file.

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -888,6 +888,15 @@ class ClusterConfigMetadataParam(JsonParam):
         LOGGER.debug("Automatic labels generated: {0}".format(str(section_labels)))
         return self.__section_resources.resources(section_key)
 
+    def to_file(self, config_parser, write_defaults=False):
+        """Ensure cluster_config_metadata_param is not exposed in the config file."""
+        LOGGER.debug("Ensuring cluster_config_metadata_param parameter is not written to configuration file.")
+        section_name = get_file_section_name(self.section_key, self.section_label)
+        try:
+            config_parser.remove_option(section_name, self.key)
+        except NoSectionError:
+            pass
+
 
 class ArchitectureParam(Param):
     """
@@ -910,6 +919,15 @@ class ArchitectureParam(Param):
             )
 
         return self
+
+    def to_file(self, config_parser, write_defaults=False):
+        """Do nothing because architecture is not exposed in the config file."""
+        LOGGER.debug("Ensuring architecture parameter is not written to configuration file.")
+        section_name = get_file_section_name(self.section_key, self.section_label)
+        try:
+            config_parser.remove_option(section_name, self.key)
+        except NoSectionError:
+            pass
 
     @staticmethod
     def get_master_instance_type_architecture(cluster_section):
--- a/cli/tests/pcluster/configure/test_pcluster_configure.py
+++ b/cli/tests/pcluster/configure/test_pcluster_configure.py
@@ -138,16 +138,22 @@ def _mock_create_network_configuration(mocker, public_subnet_id, private_subnet_
 
 
 def _mock_parallel_cluster_config(mocker):
-    supported_instance_types = ["t2.nano", "t2.micro", "t2.large", "c5.xlarge", "g3.8xlarge"]
+    supported_instance_types = ["t2.nano", "t2.micro", "t2.large", "c5.xlarge", "g3.8xlarge", "m6g.xlarge"]
     mocker.patch("pcluster.configure.easyconfig.get_supported_instance_types", return_value=supported_instance_types)
     mocker.patch(
         "pcluster.configure.easyconfig.get_supported_compute_instance_types", return_value=supported_instance_types
     )
     mocker.patch("pcluster.config.param_types.get_avail_zone", return_value="mocked_avail_zone")
-    mocker.patch("pcluster.config.param_types.get_supported_architectures_for_instance_type", return_value=["x86_64"])
+    mocker.patch(
+        "pcluster.config.param_types.get_supported_architectures_for_instance_type",
+        side_effect=lambda instance: ["arm64"] if instance == "m6g.xlarge" else ["x86_64"],
+    )
     # NOTE: the following shouldn't be needed given that easyconfig doesn't validate the config file,
     #       but it's being included in case that changes in the future.
-    mocker.patch("pcluster.config.validators.get_supported_architectures_for_instance_type", return_value=["x86_64"])
+    mocker.patch(
+        "pcluster.config.validators.get_supported_architectures_for_instance_type",
+        side_effect=lambda instance: ["arm64"] if instance == "m6g.xlarge" else ["x86_64"],
+    )
 
 
 def _launch_config(mocker, path, remove_path=True):
@@ -158,43 +164,16 @@ def _launch_config(mocker, path, remove_path=True):
     configure(args)
 
 
-def _are_configurations_equals(path_config_expected, path_config_after_input):
-    if not os.path.isfile(path_config_expected):
-        return False
-    if not os.path.isfile(path_config_after_input):
-        return False
+def _assert_configurations_are_equal(path_config_expected, path_config_after_input):
+    assert_that(path_config_expected).exists().is_file()
+    assert_that(path_config_after_input).exists().is_file()
     config_expected = ConfigParser()
     config_expected.read(path_config_expected)
-    dict1 = {s: dict(config_expected.items(s)) for s in config_expected.sections()}
-    config_temp = ConfigParser()
-    config_temp.read(path_config_after_input)
-    dict2 = {s: dict(config_temp.items(s)) for s in config_temp.sections()}
-    for section_name, section in dict1.items():
-        for key, value in section.items():
-            try:
-                if dict2[section_name][key] != value:
-                    print(
-                        (
-                            "Expected parameters are: {0}\n"
-                            "Actual parameters after running pcluster configure are: {1}"
-                        ).format(dict1, dict2)
-                    )
-                    print(
-                        "\nTest failed: Parameter '{0}' in section '{1}' is different from the expected one.".format(
-                            key, section_name
-                        )
-                    )
-                    print("The actual value is '{0}' but expected is '{1}'".format(dict2[section_name][key], value))
-                    return False
-            except KeyError:
-                print(
-                    "Expected parameters are: {0}\nActual parameters after running pcluster configure are: {1}".format(
-                        dict1, dict2
-                    )
-                )
-                print("\nTest failed: Parameter '{0}' doesn't exist in section '{1}'.".format(key, section_name))
-                return False
-    return True
+    config_expected_dict = {s: dict(config_expected.items(s)) for s in config_expected.sections()}
+    config_actual = ConfigParser()
+    config_actual.read(path_config_after_input)
+    config_actual_dict = {s: dict(config_actual.items(s)) for s in config_actual.sections()}
+    assert_that(config_actual_dict).is_equal_to(config_expected_dict)
 
 
 def _are_output_error_correct(capsys, output, error, config_path):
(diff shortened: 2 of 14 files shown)


cli/tests/pcluster/configure/test_pcluster_configure/test_no_available_no_input_no_automation_no_errors_with_config_file/pcluster.config.ini
```diff
@@ -10,6 +10,8 @@ compute_instance_type = t2.nano
 max_queue_size = 14
 initial_queue_size = 13
 maintain_initial_size = true
+base_os = alinux2
+additional_iam_policies = arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy
```
rexcsn, line 14: Why is this policy, or any `additional_iam_policies`, expected to be added by `pcluster configure`?

cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -352,7 +400,17 @@ def test_with_input_no_automation_no_errors_with_config_file(mocker, capsys, tes
...
-    _run_input_test_with_config(mocker, config, old_config_file, error, output, capsys, with_input=True)
+    _run_input_test_with_config(
+        mocker,
+        config,
+        old_config_file,
+        error,
+        output,
+        capsys,
+        with_input=True,
+        master_instance="m6g.xlarge",
```
rexcsn, line 411: What happens if user specifies a mix of ARM and X86 instances for head node and compute? I guess `pcluster configure` will succeed and config validation when creating cluster will fail right?

cli/tests/pcluster/configure/test_pcluster_configure/test_no_available_no_input_no_automation_no_errors_with_config_file/pcluster.config.ini
tilne: It's a bug with the current implementation. This the default policy that the config logic adds behind the scenes when cloudwatch logging is enabled. There's another one that's added when Batch is used as the scheduler.

tilne: I'll add an entry to the backlog for this.

cli/tests/pcluster/configure/test_pcluster_configure.py
tilne: Yeah. I've added a backlog item to eventually validate the config produced in the same way we do when creating a cluster, because as you said this validation logic is already defined there. Reasons it's not being enabled right now:
* The VPC automation doesn't take into account what instance types a user wants. This is fine for t2's, but for rarer instance types we probably need to query for availability as best we can and then create VPCs and subnets accordingly. Without that, the dry run in the config validation would fail frequently.
* We need to implement a way for the derived parameters to be updated when one of the parameters on which they're dependent changes during the `pcluster configure` process.

cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -158,42 +164,70 @@ def _launch_config(mocker, path, remove_path=True):
...
                 print(
                     "Expected parameters are: {0}\nActual parameters after running pcluster configure are: {1}".format(
-                        dict1, dict2
+                        config_expected_dict, config_actual_dict
                     )
                 )
                 print("\nTest failed: Parameter '{0}' doesn't exist in section '{1}'.".format(key, section_name))
                 return False
+        if not _do_sections_contain_same_keys(section_name, section, config_actual_dict.get(section_name, {})):
+            return False
```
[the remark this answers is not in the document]
tilne: Good call. I was making this harder than it needed to be.

> Not related to current issue: can we quickly check if we have a test case to make sure options/section that are not supposed to be modified by pcluster configure in user's existing config are not erased/overwrite after running pcluster configure? If not, would be nice to add those tests in this patch as well.

That's out of scope for this change in my opinion. That said, doesn't ensuring the expected and actual config files match exactly guarantee that for the specific sections it checks?

cli/tests/pcluster/configure/test_pcluster_configure/test_no_available_no_input_no_automation_no_errors_with_config_file/pcluster.config.ini
tilne: I looked more into this, and it's a bug related to the handling of loading a config that uses one region, and then changing it.

In this case, the region starts as eu-west-1. Since CloudWatch logging is enabled, the internal value of the parameter is set to ['arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy'].

When the region is updated to cn-north-1, the internal value is updated to ['arn:aws-cn:iam::aws:policy/CloudWatchAgentServerPolicy', 'arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy'].

Then when the value is converted to a string and the dynamically added parameters are removed, only the policy specific to the currently configured partition is removed, and thus arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy is written back.

rexcsn: Ok with leaving the fix for this out of this PR, but I think we should fix this ASAP.

Generally ok with the rest of this PR, waiting for another reviewer to give a second look.

cli/tests/pcluster/configure/test_pcluster_configure/test_subnet_automation_no_awsbatch_no_errors_empty_vpc/pcluster.config.ini
```diff
@@ -5,7 +5,7 @@ aws_region_name = eu-west-1
 key_name = key1
 vpc_settings = default
 # Implied value
```
ddeidda, line 7: Minor: maybe the comment could be removed since schedule is no longer an implied value.

cli/tests/pcluster/configure/test_pcluster_configure/test_subnet_automation_no_awsbatch_no_errors_with_config_file/pcluster.config.ini
```diff
@@ -5,14 +5,15 @@ aws_region_name = eu-west-1
...
-# scheduler = sge
+scheduler = sge
 base_os = centos6
 # Implied value
 # compute_instance_type = t2.micro
 master_instance_type = t2.nano
 max_queue_size = 14
 initial_queue_size = 13
 maintain_initial_size = true
+additional_iam_policies = arn:aws:iam::aws:policy/AWSBatchFullAccess
```
ddeidda, line 16: Why this policy with sge scheduler?

cli/tests/pcluster/configure/test_pcluster_configure/test_with_input_no_automation_no_errors_with_config_file/pcluster.config.ini
```diff
@@ -6,11 +6,12 @@ key_name = key2
...
 vpc_settings = default
 scheduler = slurm
-master_instance_type = c5.xlarge
-compute_instance_type = g3.8xlarge
+master_instance_type = m6g.xlarge
+compute_instance_type = m6g.xlarge
 max_queue_size = 18
 initial_queue_size = 7
 maintain_initial_size = true
+additional_iam_policies = arn:aws-cn:iam::aws:policy/CloudWatchAgentServerPolicy
```
ddeidda, line 14: Why adding a conditional policy?

cli/pcluster/config/param_types.py
```diff
@@ -888,6 +888,15 @@ def create_section_resources(self, section_key, num_labels, max_resources):
         LOGGER.debug("Automatic labels generated: {0}".format(str(section_labels)))
         return self.__section_resources.resources(section_key)
 
+    def to_file(self, config_parser, write_defaults=False):
+        """Ensure cluster_config_metadata_param is not exposed in the config file."""
+        LOGGER.debug("Ensuring cluster_config_metadata_param parameter is not written to configuration file.")
```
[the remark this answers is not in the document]
tilne: Here I was following the example set by the Param class. If it's in the config parser, it's going to be written back, right? This would seem to be the safest way to ensure it's never written back.

cli/pcluster/config/param_types.py
```diff
@@ -911,6 +920,15 @@ def from_file(self, config_parser):
 
         return self
 
+    def to_file(self, config_parser, write_defaults=False):
+        """Do nothing because architecture is not exposed in the config file."""
+        LOGGER.debug("Ensuring architecture parameter is not written to configuration file.")
```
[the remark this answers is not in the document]
tilne: Same answer.

cli/tests/pcluster/configure/test_pcluster_configure.py
```diff
@@ -158,43 +164,38 @@ def _launch_config(mocker, path, remove_path=True):
     configure(args)
 
 
-def _are_configurations_equals(path_config_expected, path_config_after_input):
-    if not os.path.isfile(path_config_expected):
-        return False
-    if not os.path.isfile(path_config_after_input):
-        return False
+def _do_sections_contain_same_keys(section_name, expected_config_section, actual_config_section):
```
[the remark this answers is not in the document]
tilne: nowhere. good catch. i had written it as part of an earlier change to the function that checks if configs are equal, and when i simplified that function i forgot to remove the helper i added.

cli/tests/pcluster/configure/test_pcluster_configure/test_with_input_no_automation_no_errors_with_config_file/pcluster.config.ini
tilne: It's the region change bug described here: https://github.com/aws/aws-parallelcluster/pull/1915#discussion_r463367524

cli/tests/pcluster/configure/test_pcluster_configure/test_subnet_automation_no_awsbatch_no_errors_with_config_file/pcluster.config.ini
tilne: It's like the region change bug, but this time it's due to the scheduler in the original config being awsbatch. This still applies: https://github.com/aws/aws-parallelcluster/pull/1915#discussion_r463367524

cli/tests/pcluster/configure/test_pcluster_configure/test_subnet_automation_no_awsbatch_no_errors_empty_vpc/pcluster.config.ini
tilne: Agreed.

cli/tests/pcluster/configure/test_pcluster_configure.py
tilne: https://github.com/aws/aws-parallelcluster/pull/1915#discussion_r463659994

cli/pcluster/config/param_types.py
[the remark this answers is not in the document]
ddeidda: OK, I think it can be good for now.

cli/tests/pcluster/configure/test_pcluster_configure/test_subnet_automation_no_awsbatch_no_errors_with_config_file/pcluster.config.ini
ddeidda: Thanks for the clarification. These tests need to be definitely improved in a separate PR.

PR #1923, opened 2020-08-04
Load JSON config file only for HIT enabled clusters.

ddeidda:
The code was always trying to retrieve and parse the JSON configuration file, which would be missing with traditional schedulers.

### DIFF
--- a/cli/pcluster/config/pcluster_config.py
+++ b/cli/pcluster/config/pcluster_config.py
@@ -16,7 +16,6 @@ import logging
 import os
 import stat
 import sys
-from tempfile import mkdtemp
 
 import boto3
 import configparser
@@ -32,6 +31,8 @@ from pcluster.utils import (
     get_stack,
     get_stack_name,
     get_stack_version,
+    is_hit_enabled_cluster,
+    read_remote_file,
 )
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


cli/pcluster/config/pcluster_config.py
```diff
@@ -439,28 +440,32 @@ def get_compute_availability_zone(self):
...
-                    return json.load(s3_config_file, object_pairs_hook=OrderedDict)
-            except ClientError as e:
-                self.error(
-                    "Unable to retrieve configuration from bucket '{0}'.\n{1}".format(
-                        s3_bucket_name, e.response.get("Error").get("Message")
+        json_config = None
+        scheduler = get_cfn_param(cfn_params, "Scheduler")
+        if is_hit_enabled_cluster(scheduler):
+            s3_bucket_name = get_cfn_param(cfn_params, "ResourcesS3Bucket")
+            if s3_bucket_name and s3_bucket_name != "NONE":
```
demartinofra, line 447: don't we need to also fail with error if s3_bucket_name is not found or NONE?

ddeidda: I don't think so. This would be the case of "sge" or "torque" schedulers, for which we do not create the resources bucket. Am I missing something?

ddeidda: As per our conversation, updated code to use read_remote_file and added unit tests

PR #1928, opened 2020-08-06
Split mappings definition for cluster section

ddeidda:
This commit splits the "cluster" section mappings in two different definitions, each one dedicated to its corresponding cluster model.

A "ClusterModel" class has been introduced to infer and map the "model" of a cluster configuration.
Supported cluster models are:
- SIT: Single Instance Type
- HIT: Heterogeneous Instance Types

When a pcluster_config instance is loaded from a config file or from cfn params, its target cluster model is first inferred; then the configuration is loaded and validated accordingly to this model.


cli/pcluster/models/hit/hit_cluster_model.py
```diff
@@ -0,0 +1,124 @@
...
+# limitations under the License.
+from botocore.exceptions import ClientError
+
+from pcluster.cluster_model import ClusterModel
+from pcluster.config import mappings
+from pcluster.utils import get_instance_type
+
+
+class HITClusterModel(ClusterModel):
+    """HIT cluster model."""
```
enrico-usai, line 19: In the comment I'd write: Heterogeneous Instance Type model, just to have a point in the code on which we're explaining the acronym :D (I'll do the same for SIT)

cli/pcluster/models/sit/sit_cluster_model.py
```diff
@@ -0,0 +1,109 @@
...
+# limitations under the License.
+from botocore.exceptions import ClientError
+
+from pcluster.cluster_model import ClusterModel
+from pcluster.config import mappings
+from pcluster.utils import get_instance_vcpus
+
+
+class SITClusterModel(ClusterModel):
+    """SIT cluster model."""
```
enrico-usai, line 19: See previous comment.

cli/pcluster/models/hit/hit_cluster_model.py
```diff
@@ -0,0 +1,124 @@
...
+    def get_cluster_section_definition(self):
+        """Get the cluster section definition used by the cluster model."""
+        return mappings.CLUSTER_HIT
+
+    def test_configuration(self, pcluster_config):
+        """Try to launch the requested instances (in dry-run mode) to verify configuration parameters."""
+        cluster_section = pcluster_config.get_section("cluster")
+        vpc_section = pcluster_config.get_section("vpc")
+
+        if not cluster_section or cluster_section.get_param_value("scheduler") == "awsbatch" or not vpc_section:
```
enrico-usai, line 33: Why there is no the condition `or cluster_section.get_param_value("cluster_type") == "spot"` we have in the SIT test?

cli/pcluster/models/sit/sit_cluster_model.py
```diff
@@ -0,0 +1,109 @@
...
+            compute_subnet = master_subnet
+        security_groups_ids = []
+        if vpc_security_group:
+            security_groups_ids.append(vpc_security_group)
+
+        # Initialize CpuOptions
+        disable_hyperthreading = cluster_section.get_param_value("disable_hyperthreading")
+        master_vcpus = get_instance_vcpus(master_instance_type)
+        compute_vcpus = get_instance_vcpus(compute_instance_type)
+        master_cpu_options = {"CoreCount": master_vcpus // 2, "ThreadsPerCore": 1} if disable_hyperthreading else {}
```
enrico-usai, line 63: This logic seems duplicated, can we move it into a common function in the ClusterModel class?

cli/pcluster/models/sit/sit_cluster_model.py
```diff
@@ -0,0 +1,109 @@
...
+        compute_cpu_options = {"CoreCount": compute_vcpus // 2, "ThreadsPerCore": 1} if disable_hyperthreading else {}
+
+        # Initialize Placement Group Logic
+        placement_group = cluster_section.get_param_value("placement_group")
+        placement = cluster_section.get_param_value("placement")
+        master_placement_group = (
+            {"GroupName": placement_group}
+            if placement_group not in [None, "NONE", "DYNAMIC"] and placement == "cluster"
```
enrico-usai, line 71: Also this too is replicated in both Models, can we put all the common functions in common methods?

cli/pcluster/utils.py
```diff
@@ -269,8 +269,9 @@ def get_instance_vcpus(region, instance_type):
     or the pricing file cannot be retrieved/parsed
```
enrico-usai, line 269: Comment out-to-date.

Then I'd put this important change in the commit message and in the PR description.

cli/pcluster/utils.py
```diff
@@ -269,8 +269,9 @@ def get_instance_vcpus(region, instance_type):
...
     """
     try:
-        instances = _get_json_from_s3(region, "instances/instances.json")
-        vcpus = int(instances[instance_type]["vcpus"])
+        instance_type_info = get_instance_type(instance_type)
+        vcpus_info = instance_type_info.get("VCpuInfo")
+        vcpus = vcpus_info.get("DefaultVCpus")
     except (KeyError, ValueError, ClientError):
```
enrico-usai, line 275: I think the ValueError is not more valid.

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -109,13 +112,89 @@ def test_cluster_section_from_cfn(mocker, cfn_params_dict, expected_section_dict
...
+            "'maintain_initial_size' is not allowed in the .* section",
+        ),
+        ({"cluster default": {"cluster_type": "ondemand"}}, {}, "'cluster_type' is not allowed in the .* section"),
+        ({"cluster default": {"spot_price": 0}}, {}, "'spot_price' is not allowed in the .* section"),
+    ],
+)
+def test_hit_cluster_section_from_file(mocker, config_parser_dict, expected_dict_params, expected_message):
+    config_parser_dict["cluster default"]["queue_settings"] = "queue1"
+    config_parser_dict["queue queue1"] = {
+        # "instance_type" : "t2.micro"
```
enrico-usai, line 191: Why is it commented out?

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -20,15 +20,18 @@
...
       "Description": "Desired number of compute EC2 instances / vCpus to launch within the cluster.",
-      "Type": "Number"
+      "Type": "Number",
+      "Default": 0
     },
     "MaxSize": {
       "Description": "Maximum number of compute EC2 instances / vCpus to launch within the cluster.",
-      "Type": "Number"
+      "Type": "Number",
+      "Default": 0
```
enrico-usai, line 34: Are we sure we want to put the default of MaxSize to 0? I'd put 10.
Then I'd add this change to the commit message and to the PR description.

cli/pcluster/config/mappings.py
```diff
@@ -84,6 +84,7 @@
     maintain_initial_size_validator,
     queue_settings_validator,
     queue_validator,
+    queue_xor_cluster_validator,
```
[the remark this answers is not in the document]
ddeidda: The validator checks that the parameter is used in cluster or in queue section (it cannot be used in both). Which name would you suggest to express this concept?

cli/pcluster/models/hit/hit_cluster_model.py
ddeidda: Because "cluster_type" is not present in HIT's cluster section.

cloudformation/aws-parallelcluster.cfn.json
ddeidda: This is for HIT clusters, which does not longer rely on this parameter. Default for SIT cluster is set in mappings and it's still 10.

cli/pcluster/config/json_param_types.py
```diff
@@ -251,6 +205,83 @@ def get_default_param_type(self):
...
+            # Set gpus according to instance features
+            gpu_info = instance_type.get("GpuInfo", None)
+            if gpu_info:
+                # Currently adding up all gpus. To be reviewed if the case of heterogeneous GPUs arises.
+                compute_resource_section.get_param("gpus").value = sum(
+                    [gpus.get("Count") for gpus in gpu_info.get("Gpus")]
+                )
+
+            # Set enable_efa according to queues' enable_efa and instance features
+            enable_efa = self.get_param_value("enable_efa")
```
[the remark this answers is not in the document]
ddeidda: The problem here is that ``compute_resource`` settings depend on parent ``queue`` settings, which in turn depend on ``cluster`` settings. Moving the logic into single parameters we would not have the call order guarantee.
Additionally, doing everything in a central place allows us to run a single boto3 call, and we do not need a class for each parameter.
About the validation warnings, I can work to move them to validators but this would require another boto3 call as well.

cli/pcluster/config/cfn_param_types.py
```diff
@@ -658,12 +658,17 @@ def to_cfn(self):
...
-
             master_cores = get_instance_vcpus(self.pcluster_config.region, master_instance_type) // 2
-            compute_cores = get_instance_vcpus(self.pcluster_config.region, compute_instance_type) // 2
+
+            if self.pcluster_config.cluster_model.name == "SIT":
+                # Compute instance type is valid only in SIT clusters
+                compute_instance_type = cluster_config.get_param_value("compute_instance_type")
+                compute_cores = get_instance_vcpus(self.pcluster_config.region, compute_instance_type) // 2
+            else:
+                compute_instance_type = ""
```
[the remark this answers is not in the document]
ddeidda: OK

cli/pcluster/config/hit_converter.py
```diff
@@ -24,60 +30,95 @@ def convert(self):
...
+                compute_resource_section = JsonSection(
+                    mappings.COMPUTE_RESOURCE,
+                    self.pcluster_config,
+                    section_label="default",
+                    parent_section=queue_section,
+                )
+                self.pcluster_config.add_section(compute_resource_section)
+                queue_section.get_param("compute_resource_settings").value = "default"
+
+                self._copy_param_value(
```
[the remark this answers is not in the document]
ddeidda: Do you think it is worth introducing this optimization? We are copying parameters from the SIT cluster section to the queue, to the compute_resource and to the HIT cluster section. Sometimes the name of the parameter is not the same in both sections and sometimes some conditional logic is needed. I'm worried the resulting code would result even harder to understand.

cli/pcluster/models/hit/hit_cluster_model.py
```diff
@@ -0,0 +1,124 @@
...
+            return
+
+        master_instance_type = cluster_section.get_param_value("master_instance_type")
+
+        # Retrieve network parameters
+        compute_subnet = vpc_section.get_param_value("compute_subnet_id")
```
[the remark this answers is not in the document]
ddeidda: Dryrun checks for HIT cluster model will very likely be modified as a separate commit.

cli/pcluster/config/json_param_types.py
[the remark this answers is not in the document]
enrico-usai: Let's  move at least the validator that doesn't require an additional boto3 call and add a comment in the "exceptional" validator we are keeping here, explaining the reason (that is a good one).

cli/pcluster/config/mappings.py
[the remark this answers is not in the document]
enrico-usai: WE can rename it as queue_xor_cluster_params_validator

cli/pcluster/models/hit/hit_cluster_model.py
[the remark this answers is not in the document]
enrico-usai: We could at least put in common the part related to the networking.

cli/pcluster/config/mappings.py
```diff
@@ -164,7 +165,7 @@
     "key": "global",
     "params": {
         "cluster_template": {
-            # TODO This could be a SettingsParam referring to a CLUSTER section
+            # TODO This could be a SettingsParam referring to a CLUSTER_SIT section
```
demartinofra, line 168: minor: why this renaming?

cli/pcluster/config/mappings.py
```diff
@@ -655,40 +656,258 @@
...
+    }),
+    # Compute fleet
+    ("compute_root_volume_size", {
+        "type": IntCfnParam,
+        "default": 25,
+        "allowed_values": ALLOWED_VALUES["greater_than_25"],
+        "cfn_param_mapping": "ComputeRootVolumeSize",
+        "update_policy": UpdatePolicy.COMPUTE_FLEET_STOP
+    }),
+    ("spot_bid_percentage", {
```
demartinofra, line 714: this is for awsbatch only, shall we move to SIT?

cli/pcluster/config/mappings.py
```diff
@@ -655,40 +656,258 @@
...
+        "update_policy": UpdatePolicy.SUPPORTED
+    }),
+    # Customization
+    ("template_url", {
+        # TODO add regex
+        "validators": [url_validator],
+        # Ignored during update since we force using previous template
+        "update_policy": UpdatePolicy.IGNORED
+    }),
+    ("hit_template_url", {
```
demartinofra, line 746: HIT?

cli/pcluster/config/json_param_types.py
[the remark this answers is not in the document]
ddeidda: Is it OK if I just add the comment and we address the proposed refactoring in another patch? I'd like to move both the warnings but we need to find a proper way for handling these cases, and in general to mitigate the exponential growth of boto3 calls.

cli/pcluster/config/mappings.py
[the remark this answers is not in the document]
ddeidda: Done

cli/pcluster/models/hit/hit_cluster_model.py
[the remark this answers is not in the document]
ddeidda: As per our conversation, leaving it to future improvements

cli/pcluster/models/sit/sit_cluster_model.py
ddeidda: I'd avoid spending much time optimizing this code since dryrun checks logic needs to be reviewed.

cli/pcluster/models/sit/sit_cluster_model.py
ddeidda: Please see my previous reply.

cli/pcluster/utils.py
ddeidda: Fixed

cli/pcluster/utils.py
ddeidda: Done

cli/tests/pcluster/config/test_section_cluster.py
ddeidda: Removed

cli/pcluster/config/mappings.py
ddeidda: Yes, it was wrong

cli/pcluster/config/mappings.py
ddeidda: Done

cli/pcluster/config/mappings.py
ddeidda: :+1:

cli/pcluster/config/json_param_types.py
```diff
@@ -251,6 +205,83 @@ def get_default_param_type(self):
...
+    def refresh_section(self):
+        """Take values of disable_hyperthreading and enable_efa from cluster section if not specified."""
+        if self.get_param_value("disable_hyperthreading") is None:
+            self.get_param("disable_hyperthreading").value = self.pcluster_config.get_section(
+                "cluster"
+            ).get_param_value("disable_hyperthreading")
+
+        if self.get_param_value("enable_efa") is None:
+            cluster_enable_efa = self.pcluster_config.get_section("cluster").get_param_value("enable_efa")
+            self.get_param("enable_efa").value = cluster_enable_efa == "compute"
```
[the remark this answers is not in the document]
ddeidda: Done

cli/pcluster/config/mappings.py
```diff
@@ -634,12 +635,12 @@
         ("enable_efa", {
             "type": BooleanJsonParam,
             "update_policy": UpdatePolicy.COMPUTE_FLEET_STOP,
-            "default": False
+            "validators": [queue_xor_cluster_params_validator]
```
enrico-usai, line 638: You removed the default, is it right?

cli/pcluster/config/mappings.py
```diff
@@ -655,40 +656,246 @@
...
+        "type": JsonCfnParam,
+        "update_policy": UpdatePolicy.UNSUPPORTED,
+    }),
+    ("custom_chef_cookbook", {
+        "cfn_param_mapping": "CustomChefCookbook",
+        # TODO add regex
+        "validators": [url_validator],
+        "update_policy": UpdatePolicy.UNSUPPORTED,
+    }),
+    ("custom_awsbatch_template_url", {
```
enrico-usai, line 812: We can move it in the SIT one (let's double check if we're skipping the installation of the cli in non-awsbatch cases, I think so)

cli/pcluster/config/mappings.py
```diff
@@ -655,40 +656,246 @@
...
+CLUSTER_COMMON_PARAMS = [
+    # OrderedDict due to conditional defaults values
+    ("cluster_config_metadata", {
+        # NOTE: this is not exposed as a configuration parameter
+        "type": ClusterConfigMetadataCfnParam,
+        "cfn_param_mapping": "ClusterConfigMetadata",
+        "update_policy": UpdatePolicy.IGNORED,
+        "visibility": Visibility.PRIVATE,
+    }),
+    ("key_name", {
```
enrico-usai, line 668: We can set it as required

cli/pcluster/config/mappings.py
```diff
@@ -655,40 +656,246 @@
...
+        "update_policy": UpdatePolicy.UNSUPPORTED,
+    }),
+    # Settings
+    ("scaling_settings", {
+        "type": SettingsCfnParam,
+        "default": "default",  # set a value to create always the internal structure for the scaling section
+        "referred_section": SCALING,
+        "update_policy": UpdatePolicy.UNSUPPORTED,
+    }),
+    ("vpc_settings", {
```
enrico-usai, line 832: We can set it as required

cli/pcluster/config/pcluster_config.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         # "From Stack" initialization parameters:
         :param cluster_name: the cluster name associated to a running Stack,
         if specified the initialization will start from the running Stack
+        :param auto_refresh: if set, automatic variables
```
enrico-usai, line 72: let's complete the comment, something like:
```
If set, the parameters will be refreshed when there is a section update. 
The add/remove_section methods always call the config_updated method
and this one will call the refresh of the parameters only if __autorefresh is set to true.
```

cli/pcluster/config/validators.py
```diff
@@ -404,11 +404,15 @@ def efa_validator(param_key, param_value, pcluster_config):
...
-        )
+    if pcluster_config.cluster_model.name == "SIT":
+        # Specific validations for SIT clusters
+        if cluster_section.get_param_value("compute_instance_type") not in allowed_instances:
+            errors.append(
+                "When using 'enable_efa = {0}' it is required to set the 'compute_instance_type' parameter "
+                "to one of the following values : {1}".format(param_value, allowed_instances)
+            )
+        if cluster_section.get_param_value("placement_group") is None:
+            warnings.append("You may see better performance using a cluster placement group.")
```
enrico-usai, line 415: What about a link to the official placement group docuemntation

cli/pcluster/config/validators.py
```diff
@@ -1200,3 +1209,17 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
     return errors, warnings
+
+
+def queue_xor_cluster_params_validator(param_key, param_value, pcluster_config):
+    """Return error if the same param is specified in the cluster section."""
+    errors = []
+    cluster_section_value = pcluster_config.get_section("cluster").get_param_value(param_key)
+    if param_value is not None:
+        if cluster_section_value is not None:
+            errors.append("Param '{0}' can be used only in 'cluster' or in 'queue' section".format(param_key))
```
enrico-usai, line 1220: I'd write: `"Parameter 'xxx' can be used in only one of the '[cluster ...]' or '[queue ...]' sections`. 
 @nstornetta do you have a better suggestion?

cli/pcluster/config/validators.py
```diff
@@ -1200,3 +1209,17 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    else:
+        if cluster_section_value is None:
+            errors.append("Param '{0}' must be specified in 'cluster' or in 'queue' section".format(param_key))
```
enrico-usai, line 1223: Param --> Parameter, then let's see previous comment

cli/pcluster/models/hit/hit_cluster_model.py
```diff
@@ -0,0 +1,124 @@
...
+                )
+
+                compute_resource_settings = queue_section.get_param_value("compute_resource_settings")
+                for label in compute_resource_settings.split(","):
+                    compute_resource_section = pcluster_config.get_section("compute_resource", label)
+
+                    self.__test_compute_resource(
+                        pcluster_config,
+                        compute_resource_section,
+                        disable_hyperthreading=queue_section.get_param_value("disable_hyperthreading"),
```
enrico-usai, line 89: should we use `or disable_hyperthreading` already taken from the cluster or the queue?

cli/tests/pcluster/config/defaults.py
```diff
@@ -136,6 +135,48 @@
     "fsx_settings": None,
     "dcv_settings": None,
     "cw_log_settings": None,
+    "cluster_config_metadata": {"sections": {}},
+    "architecture": "x86_64",
+}
+
+DEFAULT_CLUSTER_HIT_DICT = {
```
enrico-usai, line 142: We can define a "COMMON_CLUSTER_DICT" and create the other ones by concatenating the common and the custom stuff.

cli/tests/pcluster/config/defaults.py
```diff
@@ -282,6 +325,72 @@ class DefaultDict(Enum):
 }
 
 
+DEFAULT_CLUSTER_HIT_CFN_PARAMS = {
```
enrico-usai, line 328: should we have a COMMON? see previous comment

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -109,13 +112,87 @@ def test_cluster_section_from_cfn(mocker, cfn_params_dict, expected_section_dict
...
+            {},
+            "'placement_group' is not allowed in the .* section",
+        ),
+        ({"cluster default": {"placement": "ondemand"}}, {}, "'placement' is not allowed in the .* section"),
+        (
+            {"cluster default": {"compute_instance_type": "t2.micro"}},
+            {},
+            "'compute_instance_type' is not allowed in the .* section",
+        ),
+        ({"cluster default": {"initial_queue_size": 0}}, {}, "'initial_queue_size' is not allowed in the .* section"),
```
enrico-usai, line 177: Let's remember to adapt the documentation for each parameters, specifyin when it can be used or not.

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -397,8 +474,10 @@ def test_cluster_section_from_file(mocker, config_parser_dict, expected_dict_par
         ("fsx_settings", "test1", None, "Section .* not found in the config file"),
     ],
 )
-def test_cluster_param_from_file(mocker, param_key, param_value, expected_value, expected_message):
-    utils.assert_param_from_file(mocker, CLUSTER, param_key, param_value, expected_value, expected_message)
+def test_sit_cluster_param_from_file(
```
enrico-usai, line 477: We should add a similar test for hit_cluster, to be sure only the supported parameters are available.

We can move the common ones in another test and have two tests for the two specific models with the subset of specific parameters.

enrico-usai: Then we need to add test_section_queue.py and test_section_compute_resource.py 

They are very important to be sure to have defined the right parameters in the mapping, with the right validators.

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -408,9 +487,9 @@ def test_cluster_param_from_file(mocker, param_key, param_value, expected_value,
         ("base_os", None, None, "Configuration parameter 'base_os' must have a value"),
     ],
 )
-def test_cluster_param_from_file_with_validation(mocker, param_key, param_value, expected_value, expected_message):
+def test_sit_cluster_param_from_file_with_validation(mocker, param_key, param_value, expected_value, expected_message):
```
enrico-usai, line 490: see preiovus comment, we should have test_hit_cluster_param_from_file_with_validation

cli/pcluster/config/mappings.py
enrico-usai: What about having a unit test for required parameters?

cli/pcluster/config/mappings.py
ddeidda: Yes, there should be no default as the value now is taken from cluster section if not specified here. The ``queue_xor_cluster_params`` validator checks that the value is set here or there (but not in both sections).

cli/pcluster/models/hit/hit_cluster_model.py
ddeidda: No. During this phase all the parameters in ``compute_resource`` must already have been processed and initialized by refresh operations.

cli/tests/pcluster/config/test_section_cluster.py
ddeidda: `` test_sit_cluster_param_from_file`` is already implemented

cli/pcluster/config/mappings.py
ddeidda: Done

cli/pcluster/config/validators.py
ddeidda: I wouldn't change this in this patch, as the message comes frome previous versions.

cli/pcluster/config/validators.py
ddeidda: I wonder "only one" could be interpreted as "only in one queue section". Instead, if it is not specified in ``cluster`` section, it must be specified in *all* `queue` sections.

ddeidda: OK for Param -> Parameter

cli/pcluster/config/mappings.py
ddeidda: Could we leave this outside the patch? This change has a big impact on tests and the issue was not introduced here.

cli/pcluster/config/mappings.py
ddeidda: Same as before.

cli/pcluster/config/pcluster_config.py
ddeidda: Done

cli/pcluster/config/validators.py
ddeidda: Done

cli/tests/pcluster/config/test_section_cluster.py
ddeidda: `test_section_queue` and `test_section_compute_resource` are also already present.

cli/tests/pcluster/config/test_section_cluster.py
ddeidda: What is the purpose of this test? Looks like it's only testing required parameters, which are the same for HIT and SIT cluster sections.

cli/pcluster/config/mappings.py
enrico-usai: ok

cli/pcluster/config/validators.py
enrico-usai: ok

cli/tests/pcluster/config/test_section_cluster.py
enrico-usai: Sorry I missed them because they are not part of this patch.

cli/tests/pcluster/config/test_section_cluster.py
enrico-usai: You right

cli/tests/pcluster/config/defaults.py
ddeidda: What would be the advantages brought by this optimization? I'd keep the freedom of having completely separated dicts, since also common parameters could have a different default value, like `disable_hyperthreading`.

cli/tests/pcluster/config/test_section_cluster.py
ddeidda: + @HLeeHart

cli/tests/pcluster/config/defaults.py
ddeidda: Replied below

cli/tests/pcluster/config/test_validators.py
```diff
@@ -1484,34 +1486,72 @@ def test_queue_settings_validator(mocker, cluster_section_dict, expected_message
...
+            {"queue_settings": "default"},
+            {"compute_resource_settings": "cr1,cr2", "enable_efa": True, "disable_hyperthreading": True},
+            [
+                "Duplicate instance type 't2.micro' found in queue 'default'. "
+                "Compute resources in the same queue must use different instance types"
+            ],
+        ),
+        (
+            {"queue_settings": "default"},
+            {"compute_resource_settings": "cr3,cr4", "enable_efa": True, "disable_hyperthreading": True},
```
enrico-usai, line 1512: Great to see this new tests!

PR #1932, opened 2020-08-07
Add integration tests for `pcluster configure`

tilne:

### DIFF
--- a/tests/integration-tests/tests/configure/test_pcluster_configure.py
+++ b/tests/integration-tests/tests/configure/test_pcluster_configure.py
@@ -1,0 +1,158 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+import logging
+
+import pexpect
+import pytest
+from assertpy import assert_that
+
+from pcluster.config.pcluster_config import PclusterConfig
+
+
+@pytest.mark.regions(["us-east-1"])
+@pytest.mark.instances(["c5.xlarge", "m6g.xlarge"])
+@pytest.mark.schedulers(["awsbatch", "slurm"])
+def test_pcluster_configure(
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -0,0 +1,77 @@
...
+        key_name,
+        scheduler,
+        os,
+        instance,
+        vpc_stack.cfn_outputs["VpcId"],
+        vpc_stack.cfn_outputs["PublicSubnetId"],
+        vpc_stack.cfn_outputs["PrivateSubnetId"],
+        config_path,
+    )
+    clusters_factory(config_path)
```
demartinofra, line 37: shall we also compare the content of the config file against the expected config other than creating the cluster? This would have caught the issue with the metadata config option.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pexpect
+import pytest
+from assertpy import assert_that
+
+
+@pytest.mark.regions(["us-east-1"])
+@pytest.mark.instances(["c5.xlarge", "m6g.xlarge"])
+@pytest.mark.schedulers(["awsbatch", "slurm"])
+def test_pcluster_configure(vpc_stack, key_name, region, os, instance, scheduler, clusters_factory, test_datadir):
```
demartinofra, line 23: this test is not compatible with custom_ami and custom_url options, shall we skip it with an appropriate message in case such options are used with the tests rather than failing?

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -0,0 +1,77 @@
...
+        instance,
+        vpc_stack.cfn_outputs["VpcId"],
+        vpc_stack.cfn_outputs["PublicSubnetId"],
+        vpc_stack.cfn_outputs["PrivateSubnetId"],
+        config_path,
+    )
+    clusters_factory(config_path)
+
+
+def orchestrate_pcluster_configure(
```
demartinofra, line 40: nicely done :+1::+1::+1:

tests/integration-tests/tests/configure/test_pcluster_configure.py
tilne: I will add a check to ensure that the subset of parameters explicitly configured by the test have the expected values. As for checking for unexpected parameters: the unit tests were updated to ensure only the expected configs are being written to the file. That is sufficient in my opinion.

tests/integration-tests/tests/configure/test_pcluster_configure.py
tilne: Good point. I found six that looked like they should result in skips. Let me know if you disagree.

PR #1940, opened 2020-08-13
Restore default values for disable_hyperthreading and enable_efa in cluster section

ddeidda:
This commit fixes a regression introduced in https://github.com/aws/aws-parallelcluster/pull/1928, which made these parameters mandatory either in the cluster or in queue sections.

The new behavior allows configuration files to omit these parameter, with the meaning of setting them to false.
The validation checks that the parameters are specified only in cluster or in queue section, or that at least their valuea are the same in both sections.

### DIFF
--- a/cli/pcluster/config/param_types.py
+++ b/cli/pcluster/config/param_types.py
@@ -19,6 +19,7 @@ from configparser import NoSectionError
 
 from pcluster.config.update_policy import UpdatePolicy
 from pcluster.config.validators import settings_validator
+from pcluster.utils import get_file_section_name
 
 LOGGER = logging.getLogger(__name__)
 
@@ -589,10 +590,6 @@ class Section(ABC):
 
 
 # ---------------------- Common functions ---------------------- #
-def get_file_section_name(section_key, section_label=None):
-    return section_key + (" {0}".format(section_label) if section_label else "")
-
-
 def _ensure_section_existence(config_parser, section_name):
     """Add a section to the config_parser if not present."""
     if not config_parser.has_section(section_name):
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -23,6 +23,7 @@ from pcluster.utils import (
     ellipsize,
     get_base_additional_iam_policies,
     get_efs_mount_target_id,
+    get_file_section_name,
     get_instance_vcpus,
     get_partition,
     get_region,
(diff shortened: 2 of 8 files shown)


cli/pcluster/config/validators.py
```diff
@@ -1082,18 +1082,17 @@ def queue_validator(section_key, section_label, pcluster_config):
     queue_section = pcluster_config.get_section(section_key, section_label)
     compute_resource_labels = str(queue_section.get_param_value("compute_resource_settings") or "").split(",")
 
-    def check_queue_xor_cluster(param_key):
-        cluster_section_value = pcluster_config.get_section("cluster").get_param_value(param_key)
+    def check_queue_xor_cluster(param_key, cluster_enabled_value):
+        cluster_param_value = pcluster_config.get_section("cluster").get_param_value(param_key)
         param_value = queue_section.get_param_value(param_key)
```
enrico-usai, line 1087: What about renaming param_value to queue_param_value?

cli/pcluster/config/validators.py
```diff
@@ -1082,18 +1082,17 @@ def queue_validator(section_key, section_label, pcluster_config):
     queue_section = pcluster_config.get_section(section_key, section_label)
     compute_resource_labels = str(queue_section.get_param_value("compute_resource_settings") or "").split(",")
 
-    def check_queue_xor_cluster(param_key):
-        cluster_section_value = pcluster_config.get_section("cluster").get_param_value(param_key)
+    def check_queue_xor_cluster(param_key, cluster_enabled_value):
```
enrico-usai, line 1085: I'd add a comment in the function to describe what is "cluster_enabled_value"

cli/pcluster/config/param_types.py
```diff
@@ -19,6 +19,7 @@
 
 from pcluster.config.update_policy import UpdatePolicy
 from pcluster.config.validators import settings_validator
+from pcluster.utils import get_file_section_name
```
enrico-usai, line 22: Why are we moving this function?

I don't like too much our utils module, it is full of totally different functions, with different scopes.
I'm more in favour of trying to keep the utility functions in the related modules.

What do you think?

cli/pcluster/config/validators.py
```diff
@@ -1083,14 +1084,17 @@ def queue_validator(section_key, section_label, pcluster_config):
...
     def check_queue_xor_cluster(param_key):
-        cluster_section_value = pcluster_config.get_section("cluster").get_param_value(param_key)
-        param_value = queue_section.get_param_value(param_key)
-        if param_value is not None:
-            if cluster_section_value is not None:
+        """Check that the param is not used in both queue and cluster section."""
+        config_parser = pcluster_config.config_parser
+        if config_parser:
+            # This check is performed only if the configuration is loaded from file.
+            queue_param_in_config_file = config_parser.has_option(
```
enrico-usai, line 1091: Why are we searching in the config_parser and not in the pcluster_config object itself?

cli/pcluster/config/param_types.py
ddeidda: I had to move it because I needed to use it from the validator, and this caused a circular reference problem. I agree in the future we should organize better the utils function.

cli/pcluster/config/validators.py
ddeidda: As per our conversation, this validation must be performed at config_file level. I added a FIXME note to remind us to improve the validation mechanism by allowing validator to be explicitely attached to a specific validation phase.

PR #1945, opened 2020-08-17
Add DynamoDB table definition for HIT stacks

demartinofra:
- Define a DynamoDB table to be used with the new heterogeneous instance types logic. The table defines a GSI on the InstanceId attribute so that the look-up can be performed efficiently on that value.
- As part of this PR the DDB billing mode in GovCloud regions is being set to PAY_PER_REQUEST in order to be aligned to the other partitions. In the past this mode was not available in GovCloud regions.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -1359,8 +1359,12 @@
     "SQS": {
       "Type": "AWS::SQS::Queue",
       "Properties": {
-        "MessageRetentionPeriod": 1209600
-      }
+        "MessageRetentionPeriod": 1209600,
+        "QueueName": {
+          "Ref": "AWS::StackName"
+        }
+      },
+      "Condition": "CreateComputeFleet"
     },
     "FSXSubstack": {
       "Type": "AWS::CloudFormation::Stack",
@@ -1452,7 +1456,8 @@
             "Ref": "SQS"
           }
         ]
-      }
+      },
+      "Condition": "CreateComputeFleet"
     },
     "SNS": {
       "Type": "AWS::SNS::Topic",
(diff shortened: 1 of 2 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -2916,13 +2895,10 @@
             "Ref": "MaxSize"
           },
           "DynamoDBTable": {
-            "Ref": "DynamoDBTable"
+            "Ref": "AWS::StackName"
           },
           "SQSQueueName": {
```
enrico-usai, line 2900: Should we have a condition?

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -393,4 +393,26 @@ Resources:
...
   {%- endif %}
 {%- endfor %}
+  DynamoDBTable:
+    Type: AWS::DynamoDB::Table
+    UpdateReplacePolicy: Retain
+    DeletionPolicy: Delete
+    Properties:
+      TableName: !Ref 'MainStackName'
+      AttributeDefinitions:
+        - AttributeName: Id
```
enrico-usai, line 403: What about renaming it to Hostname? or something like this?

cloudformation/aws-parallelcluster.cfn.json
demartinofra: Not really needed. When `awsbatch` the table is not created nor it is used

cloudformation/compute-fleet-hit-substack.cfn.yaml
demartinofra: I used `id` only because this table also contains the compute fleet overall status with an item having key `COMPUTE_FLEET`

PR #1948, opened 2020-08-17
Create Route53 hosted zone within the cluster

enrico-usai:
Add Route53 hosted zone and Lambda cleanup function

A new private hosted zone is created within the cluster.
The domain name is: `<clustername>.parallelcluster`.

We extended the cleanup Lambda function to delete DNS record because the hosted zone cannot be deleted if there are DNS record on it.

Right now the DNS domain name is passed as input of the HITSubstack and propagated as output.
Both HostedZone Id and domain name are output of the substack.
The hosted zone is required by node daemons to add record sets for new nodes and by Lambda function to delete the record set before cluster deletion.

Other changes:
+ Added policy to permit secondary index queries in dynamodb and permissions for route53 updates
+ Added condition for HIT related policies: DescribeInstances and FleetTerminatePolicy

For all OSes:
+ Cluster creation
+ Job submission/completion
+ dig +search for short name and dig for fqdn in custom domain
+ ssh short name from master to compute and viceversa
+ verified resolv.conf in both master and compute
+ verified compute has the right hostname from the custom domain

Specific tests:
+ Cluster deletion with 1 record set
+ Cluster deletion without record sets

### DIFF
--- a/cli/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
+++ b/cli/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
@@ -20,6 +20,55 @@ logger = logging.getLogger(__name__)
 boto3_config = Config(retries={"max_attempts": 60})
 
 
+def _delete_dns_records(event):
+    """Delete all DNS entries from the private Route53 hosted zone created within the cluster."""
+    hosted_zone_id = event["ResourceProperties"]["ClusterHostedZone"]
+    if not hosted_zone_id:
+        logger.error("Hosted Zone ID is empty")
+        raise Exception("Hosted Zone ID is empty")
+
+    try:
+        logger.info("Deleting DNS records from %s", hosted_zone_id)
+        route53 = boto3.client("route53", config=boto3_config)
+
+        completed_successfully = False
+        while not completed_successfully:
+            completed_successfully = True
+            for changes in _list_resource_record_sets_iterator(hosted_zone_id):
+                if changes:
+                    try:
+                        route53.change_resource_record_sets(
+                            HostedZoneId=hosted_zone_id, ChangeBatch={"Changes": changes}
+                        )
+                    except Exception as e:
+                        logger.error("Failed when deleting DNS records from %s with error %s", hosted_zone_id, e)
+                        completed_successfully = False
+                        continue
+                else:
+                    logger.info("No DNS records to delete from %s.", hosted_zone_id)
+
+            logger.info("Sleeping for 5 seconds before retrying DNS records deletion.")
+            time.sleep(5)
+
+        logger.info("DNS records deletion from %s: COMPLETED", hosted_zone_id)
+    except Exception as e:
+        logger.error("Failed when listing DNS records from %s with error %s", hosted_zone_id, e)
+        raise
+
+
+def _list_resource_record_sets_iterator(hosted_zone_id):
+    route53 = boto3.client("route53", config=boto3_config)
+    pagination_config = {"PageSize": 100}
+
+    paginator = route53.get_paginator("list_resource_record_sets")
+    for page in paginator.paginate(HostedZoneId=hosted_zone_id, PaginationConfig=pagination_config):
+        changes = []
+        for record_set in page.get("ResourceRecordSets", []):
+            if record_set.get("Type") == "A":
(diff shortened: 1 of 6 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -2458,6 +2458,37 @@
                       }
                     }
                   }
+                },
+                {
+                  "Fn::If": [
+                    "CreateHITSubstack",
```
demartinofra, line 2464: nice! can you please add this conditional also to the ec2 policies in the 2 blocks above?

cli/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
```diff
@@ -20,6 +20,31 @@
...
 
 
+def _delete_dns_records(event):
+    """Delete all DNS entries from the private Route53 hosted zone created within the cluster."""
+    hosted_zone_id = event["ResourceProperties"]["ClusterHostedZone"]
+    logger.info("Deleting DNS records from %s", hosted_zone_id)
+
+    if hosted_zone_id:
+        try:
+            route53 = boto3.client("route53")
```
demartinofra, line 30: Can you apply the boto3_config here?

cli/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
```diff
@@ -20,6 +20,31 @@
...
+            pagination_config = {"PageSize": 100}
+            paginator = route53.get_paginator("list_resource_record_sets")
+            for page in paginator.paginate(HostedZoneId=hosted_zone_id, PaginationConfig=pagination_config):
```
demartinofra, line 33: to be sure everything gets deleted (since we cannot retry this lambda function) we could add a while on top of the for that checks again if there are no remaining record sets. By doing this we can avoid failing on exceptions in the inner loop. Something similar to what's done in `_terminate_cluster_nodes`

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1651,6 +1651,51 @@
                 }
               ]
             },
+            {
+              "Sid": "DynamoDBTableQuery",
```
demartinofra, line 1655: optional: if we want to start separating the policies that are only required by a HIT cluster you can use `ParallelClusterHITPolicies`

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -2439,27 +2484,74 @@
...
+                      "Sid": "Route53DeletePolicy",
+                      "Effect": "Allow",
+                      "Action": [
+                        "route53:ListResourceRecordSets",
+                        "route53:ChangeResourceRecordSets"
+                      ],
+                      "Resource": [
+                        {
+                          "Fn::Sub": [
+                            "arn:${AWS::Partition}:route53:::hostedzone/${ClusterHostedZone}",
```
demartinofra, line 2538: One problem here is that if we add a dependency on ComputeFleetHITSubstack, the lambda function performing the clean-up will only be created after ComputeFleetHITSubstack. This could be a problem in case the stack fails before ComputeFleetHITSubstack is created because the S3 bucket will be never removed. 
I'd say either we create a separate function or we get the name of the hosted zone from somewhere else that is available since the beginning

demartinofra (approved):
LGTM!

To address separately:
- Lambda function creation
- Optionally move HIT policies to HIT specific IAM policy resource
- Have a way to disable creation of Route53 HZ.

demartinofra (dismissed):

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -292,6 +299,7 @@ Resources:
                 echo "Reporting instance as unhealthy and dumping logs to ${!log_dir}/${!instance_id}.tar.gz"
                 tar -czf ${!log_dir}/${!instance_id}.tar.gz /var/log
                 # TODO: add possibility to disable this behavior
+                sleep 10
```
demartinofra, line 302: can we add this sleep also to the compute fleet substack and add a small comment saying we do this to allow logs to be sent to CloudWatch?

demartinofra (approved):

PR #1951, opened 2020-08-18
Implement support for start/stop commands in HIT cluster

demartinofra:
Known issue: currently the version check implemented for the update is preventing every command to be executed if there is a version mismatch

### DIFF
--- a/cli/pcluster/cli_commands/compute_fleet_status_manager.py
+++ b/cli/pcluster/cli_commands/compute_fleet_status_manager.py
@@ -9,6 +9,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import time
 from enum import Enum
 
 import boto3
@@ -40,6 +41,16 @@ class ComputeFleetStatus(Enum):
         """Return True if stop is requested or in progress."""
         return status in {ComputeFleetStatus.STOP_REQUESTED, ComputeFleetStatus.STOPPING}
 
+    @staticmethod
+    def is_stop_status(status):
+        """Return True if status is any of the stop ones."""
+        return status in {ComputeFleetStatus.STOP_REQUESTED, ComputeFleetStatus.STOPPING, ComputeFleetStatus.STOPPED}
+
+    @staticmethod
+    def is_start_status(status):
+        """Return True if status is any of the start ones."""
+        return status in {ComputeFleetStatus.START_REQUESTED, ComputeFleetStatus.STARTING, ComputeFleetStatus.RUNNING}
+
 
 class ComputeFleetStatusManager:
     """Implement functionalities to retrieve and update the compute fleet status."""
@@ -65,9 +76,7 @@ class ComputeFleetStatusManager:
                 raise Exception("COMPUTE_FLEET status not found in db table")
             return ComputeFleetStatus(compute_fleet_status["Item"][self.COMPUTE_FLEET_STATUS_ATTRIBUTE])
         except Exception as e:
-            LOGGER.error(
-                "Failed when retrieving fleet status from DynamoDB with error %s, using fallback value %s", e, fallback
-            )
+            LOGGER.error("Failed when retrieving fleet status from DynamoDB with error %s", e)
             return fallback
 
     def update_status(self, current_status, next_status):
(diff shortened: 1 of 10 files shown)


cli/pcluster/cli_commands/compute_fleet_status_manager.py
```diff
@@ -0,0 +1,141 @@
...
+    @staticmethod
+    def is_start_status(status):
+        """Return True if status is any of the start ones."""
+        return status in {ComputeFleetStatus.START_REQUESTED, ComputeFleetStatus.STARTING, ComputeFleetStatus.RUNNING}
+
+
+class ComputeFleetStatusManager:
+    """Implement functionalities to retrieve and update the compute fleet status."""
+
+    COMPUTE_FLEET_STATUS_KEY = "COMPUTE_FLEET"
```
enrico-usai, line 58: Why uppercase?

cli/pcluster/cli_commands/start.py
```diff
@@ -0,0 +1,129 @@
...
+def start(args):
+    """Start cluster compute fleet."""
+    pcluster_config = PclusterConfig(config_file=args.config_file, cluster_name=args.cluster_name, auto_refresh=False)
+    cluster_section = pcluster_config.get_section("cluster")
+    scheduler = cluster_section.get_param_value("scheduler")
+
+    SCHEDULER_TO_START_COMMAND_MAP[scheduler]().start(args, pcluster_config)
+
+
+class StartCommand(ABC):
```
enrico-usai, line 41: I like this idea.

It seems to me that in future this logic should be moved to the ClusterModel and have the start/stop command implementation that depends on the cluster model (awsbatch should became another Cluster Model).

I'm saying so because the initialization part is very similar to the one in the stop.py so we could put this logic in common.

cli/pcluster/cli_commands/compute_fleet_status_manager.py
```diff
@@ -0,0 +1,141 @@
...
+        if compute_fleet_status != final_status:
+            raise Exception(
+                "Unexpected final state {} probably due to a concurrent status update request.".format(
+                    compute_fleet_status
+                )
+            )
+        else:
+            LOGGER.info("Compute fleet status updated successfully.")
+
+    def _wait_for_status_transition(self, wait_on_status, timeout=300, retry_every_seconds=15):
```
enrico-usai, line 127: With this new method we are changing the behaviour of the start/stop commands.
I mean in past the commands were executed and returned without waiting for the transition.

Now we're waiting and checking for the status and the command could hang for a bit. I'd add a least a message before the sleep to notify the user that the command is waiting to the status change.

cli/pcluster/cli_commands/compute_fleet_status_manager.py
```diff
@@ -0,0 +1,141 @@
...
+        if compute_fleet_status not in {request_status, in_progress_status, final_status}:
+            self.update_status(current_status=compute_fleet_status, next_status=request_status)
+        LOGGER.info("Submitted compute fleet status transition request. Waiting for status update to start...")
+
+        compute_fleet_status = self._wait_for_status_transition(wait_on_status=request_status, timeout=180)
+        if compute_fleet_status == in_progress_status:
+            LOGGER.info(
+                "Compute fleet status transition is in progress. This operation might take a while to complete..."
+            )
+            compute_fleet_status = self._wait_for_status_transition(wait_on_status=in_progress_status, timeout=600)
```
enrico-usai, line 116: Timeout 600 means that the command will hang waiting for the status change for 10 minutes?
We need to add a message for the user.

enrico-usai (approved):

cli/pcluster/cli_commands/start.py
demartinofra: I agree on finding a way to not duplicate part of the logic. What I don't like about moving all commands to ClusterModel is that we are going to have a single class implementing all commands while I find it useful to have separate contained modules and classes for each command.

cli/pcluster/cli_commands/compute_fleet_status_manager.py
demartinofra: just to differentiate between this as a known constant label and a nodename? Since they both share the same Id "column"

cli/pcluster/cli_commands/compute_fleet_status_manager.py
demartinofra: If you look at `update_status_and_wait_transition` there are some messages shown before calling the wait function, in particular:
- Submitted compute fleet status transition request. Waiting for status update to start...
- Compute fleet status transition is in progress. This operation might take a while to complete...

Shall I reword them?

cli/pcluster/cli_commands/compute_fleet_status_manager.py
demartinofra: yes at most 10 mins. let me know if I need to reword the message above or if you just missed it.

enrico-usai (approved):

cli/pcluster/cli_commands/compute_fleet_status_manager.py
enrico-usai: I mean we should communicate the user we're performing some actions, e.g. a message each minute, or improve the previous message saying that it could take up to 10 minutes. 
Without a message it seems that the command is hanging.

cli/pcluster/cli_commands/compute_fleet_status_manager.py
enrico-usai: Probably we should give some information to the user for each loop or at every minute.

PR #1954, opened 2020-08-18
Choose availability zone(s) offering desired instance types

hanwen-pcluste:
This pull request is for pcluster configure, and will:
1. During automate subnets creation, automatically specify availability zone(s) offering desired instance types.
   If there is no single availability zone offering instance types both of master and compute node, stop pcluster configure
2. During manually subnets selection, only list subnets whose availability zone(s) offering desired instance types.
   If there is no qualified subnet to be selected, switch to automate subnets creation.
3. Change listed objects displayed in prompt_iterable from list of tuples to list of OrderedDict.
   Therefore change VPCs/Subnets' display format from list to table.
4. Call describe_instance_type_offerings() to validate input instance types.
   Keep original validator until further notice.

### DIFF
--- a/cli/pcluster/configure/easyconfig.py
+++ b/cli/pcluster/configure/easyconfig.py
@@ -20,6 +20,8 @@ standard_library.install_aliases()
 
 import logging
 import os
+import sys
+from collections import OrderedDict
 
 import boto3
 
@@ -34,6 +36,8 @@ from pcluster.configure.utils import get_regions, get_resource_tag, handle_clien
 from pcluster.utils import (
     error,
     get_region,
+    get_supported_az_for_multi_instance_types,
+    get_supported_az_for_one_instance_type,
     get_supported_compute_instance_types,
     get_supported_instance_types,
     get_supported_os_for_scheduler,
@@ -71,8 +75,10 @@ def _get_vpcs_and_subnets():
     Return a dictionary containing a list of vpc in the given region and the associated VPCs.
 
     Example:
-    {"vpc_list": list(tuple(vpc-id, name, number of subnets)) ,
-    "vpc_to_subnet" : {"vpc-id1": list(tuple(subnet-id1, name)), "vpc-id2": list(tuple(subnet-id1, name))}}
+    {"vpc_list": list({"id":vpc-id, "name":name, "number_of_subnets": 6}) ,
+    "vpc_to_subnet" :
+                   {"vpc-id1": list({"id":subnet-id, "name":name, "size":subnet-size, "availability_zone": subnet-az}),
+                    "vpc-id2": list({"id":subnet-id, "name":name, "size":subnet-size, "availability_zone": subnet-az})}}
     """
     ec2_client = boto3.client("ec2")
     vpcs = ec2_client.describe_vpcs()
@@ -84,11 +90,7 @@ def _get_vpcs_and_subnets():
         subnets = _get_subnets(ec2_client, vpc_id)
         vpc_name = get_resource_tag(vpc, tag_name="Name")
         vpc_subnets[vpc_id] = subnets
-        subnets_count = "{0} subnets inside".format(len(subnets))
-        if vpc_name:
-            vpc_options.append((vpc_id, vpc_name, subnets_count))
-        else:
-            vpc_options.append((vpc_id, subnets_count))
+        vpc_options.append(OrderedDict([("id", vpc_id), ("name", vpc_name), ("number_of_subnets", len(subnets))]))
 
     return {"vpc_list": vpc_options, "vpc_subnets": vpc_subnets}
 
@@ -97,13 +99,16 @@ def _get_subnets(conn, vpc_id):
     subnet_options = []
     subnet_list = conn.describe_subnets(Filters=[{"Name": "vpcId", "Values": [vpc_id]}]).get("Subnets")
     for subnet in subnet_list:
-        subnet_id = subnet.get("SubnetId")
-        subnet_size_description = "Subnet size: {0}".format(_extract_subnet_size(subnet.get("CidrBlock")))
-        name = get_resource_tag(subnet, tag_name="Name")
-        if name:
-            subnet_options.append((subnet_id, name, subnet_size_description))
-        else:
-            subnet_options.append((subnet_id, subnet_size_description))
+        subnet_options.append(
+            OrderedDict(
+                [
+                    ("id", subnet.get("SubnetId")),
+                    ("name", get_resource_tag(subnet, tag_name="Name")),
+                    ("size", _extract_subnet_size(subnet.get("CidrBlock"))),
+                    ("availability_zone", subnet.get("AvailabilityZone")),
+                ]
+            )
+        )
     return subnet_options
 
 
@@ -140,47 +145,31 @@ def configure(args):
     # Use built in boto regions as an available option
     available_regions = get_regions()
     default_region = pcluster_config.get_section("aws").get_param_value("aws_region_name")
-    aws_region_name = prompt_iterable(
-        "AWS Region ID",
-        available_regions,
-        default_value=default_region if default_region in available_regions else None,
-    )
+    aws_region_name = prompt_iterable("AWS Region ID", available_regions, default_value=default_region)
     # Set provided region into os environment for suggestions and validations from here on
     os.environ["AWS_DEFAULT_REGION"] = aws_region_name
 
     # Get the key name from the current region, if any
     available_keys = _get_keys()
     default_key = cluster_section.get_param_value("key_name")
-    key_name = prompt_iterable(
-        "EC2 Key Pair Name", available_keys, default_value=default_key if default_key in available_keys else None
-    )
+    key_name = prompt_iterable("EC2 Key Pair Name", available_keys, default_value=default_key)
 
     scheduler = prompt_iterable(
         "Scheduler", get_supported_schedulers(), default_value=cluster_section.get_param_value("scheduler")
     )
-    scheduler_handler = SchedulerHandler(cluster_section, scheduler)
+    cluster_config = ClusterConfigureHelper(cluster_section, scheduler)
+    cluster_config.prompt_os()
+    cluster_config.prompt_cluster_size()
+    cluster_config.prompt_instance_types()
 
-    scheduler_handler.prompt_os()
-    scheduler_handler.prompt_cluster_size()
+    vpc_parameters = _create_vpc_parameters(vpc_section, cluster_config)
+    # Here is the end of prompt. Code below assembles config and write to file
 
-    master_instance_type = prompt(
-        "Master instance type",
-        lambda x: x in get_supported_instance_types(),
-        default_value=cluster_section.get_param_value("master_instance_type"),
-    )
-
-    scheduler_handler.prompt_compute_instance_type()
-
-    automate_vpc = prompt("Automate VPC creation? (y/n)", lambda x: x in ("y", "n"), default_value="n") == "y"
-
-    vpc_parameters = _create_vpc_parameters(
-        vpc_section, scheduler, scheduler_handler.max_cluster_size, automate_vpc_creation=automate_vpc
-    )
-    cluster_parameters = {"key_name": key_name, "scheduler": scheduler, "master_instance_type": master_instance_type}
-    cluster_parameters.update(scheduler_handler.get_scheduler_parameters())
+    cluster_parameters = {"key_name": key_name, "scheduler": scheduler}
+    cluster_parameters.update(cluster_config.get_scheduler_parameters())
 
     # Remove parameters from the past configuration that can conflict with the user's choices.
-    _reset_config_params(cluster_section, scheduler_handler.get_parameters_to_reset())
+    _reset_config_params(cluster_section, cluster_config.get_parameters_to_reset())
     _reset_config_params(vpc_section, ("compute_subnet_id", "use_public_ips", "compute_subnet_cidr"))
 
     # Update configuration values according to user's choices
@@ -214,12 +203,13 @@ def _reset_config_params(section, parameters_to_remove):
         param.value = param.get_default_value()
 
 
-def _create_vpc_parameters(vpc_section, scheduler, min_subnet_size, automate_vpc_creation=True):
+def _create_vpc_parameters(vpc_section, cluster_config):
     vpc_parameters = {}
-    min_subnet_size = int(min_subnet_size)
+    min_subnet_size = int(cluster_config.max_cluster_size)
+    automate_vpc_creation = prompt("Automate VPC creation? (y/n)", lambda x: x in ("y", "n"), default_value="n") == "y"
     if automate_vpc_creation:
         vpc_parameters.update(
-            automate_vpc_with_subnet_creation(_choose_network_configuration(scheduler), min_subnet_size)
+            automate_vpc_with_subnet_creation(_choose_network_configuration(cluster_config), min_subnet_size)
         )
     else:
         vpc_and_subnets = _get_vpcs_and_subnets()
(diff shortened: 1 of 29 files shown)


cli/pcluster/configure/networking.py
```diff
@@ -70,29 +71,38 @@ def _create(self, vpc_id, vpc_cidr, subnet_cidrs, internet_gateway_id, compute_s
...
         ]
 
+    def _get_availability_zone(self):
+        if self.availability_zones is not None:
+            """To Do: is the first az in the azs list the optimal option?"""
+            return self.availability_zones.pop()
+        else:
+            return ""
+
+    # FIXME placeholder for a function that should decide the best availability zone for the current region
```
rexcsn, line 89: This should be a to-do comment. FIXME should not be present in stable code. Alternatively I would prefer to track this question and the previous optimal AZ comment together as "potential improvements" internally.

cli/pcluster/configure/utils.py
```diff
@@ -96,52 +107,73 @@ def prompt_iterable(message, options, default_value=None):
...
+    """
+    Return rows and header to print using tabulate.
+
+    :param items: list of dicts
+    :return: rows are all values in items. header is keys of dict
+    """
+    header = list(items[0].keys())
+    header.insert(0, "#")
+    rows = []
+    for iterator, item in enumerate(items, start=1):
```
rexcsn, line 170: Why `start=1`?

cli/pcluster/configure/utils.py
```diff
@@ -96,52 +107,73 @@ def prompt_iterable(message, options, default_value=None):
...
 
+def get_rows_and_header(items):
+    """
+    Return rows and header to print using tabulate.
+
+    :param items: list of dicts
+    :return: rows are all values in items. header is keys of dict
+    """
+    header = list(items[0].keys())
+    header.insert(0, "#")
```
rexcsn, line 168: Minor: for very specific logic like this it would be nice to have an example the the printed result as comment in this function so we can immediately see that we are adding this `#` for pretty print

cli/pcluster/configure/networking.py
```diff
@@ -109,16 +119,17 @@ def _create(self, vpc_id, vpc_cidr, subnet_cidrs, internet_gateway_id, compute_s
...
             config_type="Master in a public subnet and compute fleet in a private subnet",
             template_name="public-private",
             stack_name_prefix="pubpriv",
+            availability_zones=availability_zones,
         )
 
     def get_cfn_parameters(self, vpc_id, internet_gateway_id, public_cidr, private_cidr):
         """Create cloudformation-compatible stack parameter given the variables."""
-        parameters = super(PublicPrivateNetworkConfig, self)._get_cfn_parameters(vpc_id, internet_gateway_id)
+        parameters = self._get_cfn_parameters(vpc_id, internet_gateway_id)
```
rexcsn, line 132: I would prefer to leave the `super`, even if this works it looks like a recursion :p

cli/pcluster/configure/easyconfig.py
```diff
@@ -199,56 +198,80 @@ def _reset_config_params(section, parameters_to_remove):
...
     return vpc_parameters
 
 
-def _ask_for_subnets(subnet_list, vpc_section):
-    available_subnets = [subnet_entry[0] for subnet_entry in subnet_list]
+def _filter_subnets_by_azs(subnet_list, qualified_azs):
+    return [subnet_entry for subnet_entry in subnet_list if subnet_entry["availability_zone"] in qualified_azs]
+
+
+def _ask_for_subnets(subnet_list, vpc_section, qualified_master_subnets, qualified_compute_subnets):
```
rexcsn, line 256: code to prompt for master vs compute subnets looks almost identical. Let's see how we can abstract this functionality into a single function to reduce code duplication.

cli/pcluster/configure/easyconfig.py
```diff
@@ -258,18 +281,38 @@ def _ask_for_subnets(subnet_list, vpc_section):
...
         return PublicPrivateNetworkConfig()
+    common_availability_zones = set(qualified_azs["azs_for_master_type"]) & set(qualified_azs["azs_for_compute_type"])
+
+    if not common_availability_zones:
+        print(
+            "Error: There is no single availability zone offering master and compute in current region.\n",
+            "To create your cluster, run pcluster configure again and avoid using Automate VPC/Subnet creation.",
+        )
+        print("Exiting...")
+        sys.exit(0)
```
rexcsn, line 295: Since it's exit on error, should we do exit(1)?

cli/pcluster/configure/easyconfig.py
```diff
@@ -258,18 +281,38 @@ def _ask_for_subnets(subnet_list, vpc_section):
...
 
-def _choose_network_configuration(scheduler):
+def _choose_network_configuration(scheduler, qualified_azs):
     if scheduler == "awsbatch":
         return PublicPrivateNetworkConfig()
+    common_availability_zones = set(qualified_azs["azs_for_master_type"]) & set(qualified_azs["azs_for_compute_type"])
+
+    if not common_availability_zones:
+        print(
+            "Error: There is no single availability zone offering master and compute in current region.\n",
```
rexcsn, line 291: Nice-to-have: Let's try to make this error message more clear for the user.

The points we want user to know:
* There is no single availability zone offering master and compute instance type in current region
* `pcluster configure` network automation currently only support creating subnets in the same availability zone
* To get around this issue, user need to not use network automation and specify appropriate subnets in pcluster configuration file manually

Let's see how we can phrase these into our error message

cli/pcluster/configure/easyconfig.py
```diff
@@ -150,21 +154,16 @@ def configure(args):
...
 
     automate_vpc = prompt("Automate VPC creation? (y/n)", lambda x: x in ("y", "n"), default_value="n") == "y"
 
     vpc_parameters = _create_vpc_parameters(
-        vpc_section, scheduler, scheduler_handler.max_cluster_size, automate_vpc_creation=automate_vpc
+        vpc_section, scheduler, scheduler_handler.max_cluster_size, qualified_azs, automate_vpc_creation=automate_vpc
     )
-    cluster_parameters = {"key_name": key_name, "scheduler": scheduler, "master_instance_type": master_instance_type}
+    cluster_parameters = {"key_name": key_name, "scheduler": scheduler}
     cluster_parameters.update(scheduler_handler.get_scheduler_parameters())
```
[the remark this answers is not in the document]
hanwen-pcluste: Did some change, but was unable to achieve the ideal structure. Further restructure will require 2-3 hours.

cli/pcluster/utils.py
```diff
@@ -290,6 +290,87 @@ def get_supported_compute_instance_types(scheduler):
...
+    This function build above get_supported_az_for_instance_types,
+    but simplify the input to 1 instance type and result to a list
+
+    :param instance_type: the instance type for which the supporting AZs.
+    :return: a tuple of the supporting AZs
+    """
+    return get_supported_az_for_instance_types([instance_type])[instance_type]
+
+
+def get_supported_az_for_instance_types(instance_types):
```
rexcsn, line 306: Again, caching looks a bit complex here without adding any real benefit, let's remove it.

I don't think we need `get_supported_az_for_instance_type ` as well? Maybe just keep `get_supported_az_for_instance_types ` function as the only function to call boto3 and get list of qualified azs for instances. We can store the result in variables, say `qualified_master_azs` and `qualified_compute_azs`, and just pass them around

cli/pcluster/config/pcluster_config.py
```diff
@@ -527,6 +535,18 @@ def __ec2_run_instance(self, **kwargs):
                     "The specified subnet does not contain enough free private IP addresses "
                     "to fulfill your request.\n{0}".format(message)
                 )
+            elif code == "Unsupported" and get_availability_zone_of_subnet(
```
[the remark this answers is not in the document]
hanwen-pcluste: It is fine, I am open to change. 
Is it possible to do assignment in the conditional expression of if statement? If not, it is hard to reduce the number of the function calls here. Do you have any advice?

cli/pcluster/configure/easyconfig.py
```diff
@@ -199,56 +188,85 @@ def _reset_config_params(section, parameters_to_remove):
...
                 vpc_list,
-                default_value=default_vpc if default_vpc in [vpc_entry[0] for vpc_entry in vpc_list] else None,
+                default_value=default_vpc,
             )
             vpc_parameters["vpc_id"] = vpc_id
             subnet_list = vpc_and_subnets["vpc_subnets"][vpc_id]
-            if not subnet_list or (
-                prompt("Automate Subnet creation? (y/n)", lambda x: x in ("y", "n"), default_value="y") == "y"
+            qualified_master_subnets = _filter_subnets_offering_instance_type(
+                subnet_list, cluster_config.master_instance_type
```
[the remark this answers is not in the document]
hanwen-pcluste: You are right. Previous commit did pass qualified azs around. But I'd believe the new approach is more straightforward and has better scalability.
Straightforward: Code readers don't have to learn the az restriction unless they read `_filter_subnets_offering_instance_type()`. In other words, if viewers only need to read and change `configure(args)`, they don't have to care about the az restriction at all.
Scalability: if multiple instance types are allowed in easyconfig, it is possibly easier to make change upon this new code.

cli/pcluster/configure/easyconfig.py
```diff
@@ -349,3 +402,12 @@ def get_parameters_to_reset(self):
             return "max_queue_size", "initial_queue_size", "maintain_initial_size", "compute_instance_type"
         else:
             return "max_vcpus", "desired_vcpus", "min_vcpus", "compute_instance_type"
+
+    def cache_qualified_az(self):
```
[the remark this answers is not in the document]
hanwen-pcluste: Yes, this function is confusing. I will try to clarify it

cli/pcluster/utils.py
hanwen-pcluste: Right. Although the current approach add some complexity in these utility functions, it makes code invoking them less complex. `get_suppported_az_for_instance_type` is an attempt to make code elsewhere shorter

cli/pcluster/configure/easyconfig.py
```diff
@@ -128,47 +133,31 @@ def configure(args):
...
+    key_name = prompt_iterable("EC2 Key Pair Name", available_keys, default_value=default_key)
 
     scheduler = prompt_iterable(
         "Scheduler", get_supported_schedulers(), default_value=cluster_section.get_param_value("scheduler")
     )
-    scheduler_handler = SchedulerHandler(cluster_section, scheduler)
+    cluster_config = ClusterConfigure(cluster_section, scheduler)
+    cluster_config.prompt_os()
+    cluster_config.prompt_cluster_size()
+    cluster_config.prompt_instance_types()
```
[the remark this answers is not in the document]
hanwen-pcluste: Let me keep it. Current structure can remind code viewers very clearly about the workflow of `pcluster configure`

cli/pcluster/configure/easyconfig.py
```diff
@@ -128,47 +133,31 @@ def configure(args):
...
-    key_name = prompt_iterable(
-        "EC2 Key Pair Name", available_keys, default_value=default_key if default_key in available_keys else None
-    )
+    key_name = prompt_iterable("EC2 Key Pair Name", available_keys, default_value=default_key)
 
     scheduler = prompt_iterable(
         "Scheduler", get_supported_schedulers(), default_value=cluster_section.get_param_value("scheduler")
     )
-    scheduler_handler = SchedulerHandler(cluster_section, scheduler)
+    cluster_config = ClusterConfigure(cluster_section, scheduler)
```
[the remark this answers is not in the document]
hanwen-pcluste: I will change the code accordingly

cli/pcluster/configure/easyconfig.py
```diff
@@ -128,47 +133,31 @@ def configure(args):
...
     # Get the key name from the current region, if any
     available_keys = _get_keys()
     default_key = cluster_section.get_param_value("key_name")
-    key_name = prompt_iterable(
-        "EC2 Key Pair Name", available_keys, default_value=default_key if default_key in available_keys else None
-    )
+    key_name = prompt_iterable("EC2 Key Pair Name", available_keys, default_value=default_key)
 
     scheduler = prompt_iterable(
         "Scheduler", get_supported_schedulers(), default_value=cluster_section.get_param_value("scheduler")
```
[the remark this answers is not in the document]
hanwen-pcluste: Yes. I was trying to do it but failed. But it requires some effort (2-3 hours). I am also concerned that major change of the file will make it harder to merge from wip/hit branch I will invest the effort if necessary.

cli/pcluster/configure/utils.py
hanwen-pcluste: I will add it to the comment

cli/pcluster/configure/utils.py
```diff
@@ -96,52 +107,99 @@ def prompt_iterable(message, options, default_value=None):
...
         LOGGER.error("ERROR: No options found for {0}".format(message))
         sys.exit(1)
-    is_tuple = isinstance(options[0], (list, tuple))
-
-    if default_value is None:
+    is_dict = isinstance(options[0], dict)
+    if (is_dict and default_value not in [option["id"] for option in options]) or (
+        not is_dict and default_value not in options
+    ):
+        # If default value is not allowed, change default value to the first item of the options
```
rexcsn, line 121: Can we simplify these code like this?
```
valid_options = [option["id"] for option in options] if isinstance(options[0], dict) else options
if default_value not in valid_options:
```

cli/pcluster/configure/utils.py
```diff
@@ -49,16 +50,23 @@ def get_default_suggestion(parameter, options):
 
     if parameter in opinionated_suggestions:
         default = opinionated_suggestions.get(parameter)
-    elif isinstance(options, (list, tuple)) and isinstance(options[0], (list, tuple)):
-        default = options[0][0]
+    elif isinstance(options, (list, tuple)) and isinstance(options[0], dict):
+        default = options[0]["id"]
     elif isinstance(options, (list, tuple)):
         default = options[0]
```
rexcsn, line 56: Minor: can simplify to:
```
elif isinstance(options, (list, tuple)):
    default = options[0]["id"] if isinstance(options[0], dict) else options[0]
```

cli/pcluster/utils.py
```diff
@@ -329,6 +329,87 @@ def get_supported_compute_instance_types(scheduler):
...
+    Example:
+    If instance_types is:
+    ["t2.micro", "t2.large"]
+    Result can be:
+    {
+        "t2.micro": (us-east-1a, us-east-1b),
+        "t2.large": (us-east-1a, us-east-1b)
+    }
+    """
+    # Most of the fancy code in the func is to try using only one API call for all instance types
```
rexcsn, line 361: Let's change this comment to try to explain our "fancy code" rather than just calling it out. Or just remove this comment

cli/pcluster/utils.py
```diff
@@ -329,6 +329,87 @@ def get_supported_compute_instance_types(scheduler):
...
+    if not hasattr(get_supported_az_for_multi_instance_types, "cache"):
+        get_supported_az_for_multi_instance_types.cache = {}
+    cache = get_supported_az_for_multi_instance_types.cache
+    missing_instance_types = []
+    result = {}
+    for instance_type in instance_types:
```
rexcsn, line 367: Minor: let's try to avoid parameter or function names that differ only by a single character

cli/pcluster/utils.py
```diff
@@ -329,6 +329,87 @@ def get_supported_compute_instance_types(scheduler):
...
+        ec2_client = boto3.client("ec2")
+        paginator = ec2_client.get_paginator("describe_instance_type_offerings")
+        page_iterator = paginator.paginate(
+            LocationType="availability-zone", Filters=[{"Name": "instance-type", "Values": missing_instance_types}]
+        )
+        offerings = []
+        for page in page_iterator:
+            offerings.extend(page["InstanceTypeOfferings"])
+        for instance_type in missing_instance_types:
+            cache[instance_type] = tuple(
```
rexcsn, line 382: Minor: Let's try to use a list here instead of tuple?

cli/pcluster/utils.py
```diff
@@ -329,6 +329,87 @@ def get_supported_compute_instance_types(scheduler):
...
+        "t2.micro": (us-east-1a, us-east-1b),
+        "t2.large": (us-east-1a, us-east-1b)
+    }
+    """
+    # Most of the fancy code in the func is to try using only one API call for all instance types
+    if not hasattr(get_supported_az_for_multi_instance_types, "cache"):
+        get_supported_az_for_multi_instance_types.cache = {}
+    cache = get_supported_az_for_multi_instance_types.cache
+    missing_instance_types = []
+    result = {}
```
rexcsn, line 366: Minor: Let's think about how you can do things without copying cache into another dict. Our cache is small, so ideally this function can just always return the cache and the logic of this function will look much simpler.

We can discuss, but no need to change unless you want to.

cli/pcluster/utils.py
```diff
@@ -329,6 +329,87 @@ def get_supported_compute_instance_types(scheduler):
...
+    """
+    if not hasattr(get_availability_zone_of_subnet, "cache"):
+        get_availability_zone_of_subnet.cache = {}
+    cache = get_availability_zone_of_subnet.cache
+    if subnet_id not in cache:
+        try:
+            cache[subnet_id] = (
+                boto3.client("ec2").describe_subnets(SubnetIds=[subnet_id]).get("Subnets")[0].get("AvailabilityZone")
+            )
+        except ClientError as e:
```
rexcsn, line 404: You are catching the exception here, so let's think about what is the behavior we want here.

Do we want the program to exit or continue? In either case, should we handle ClientError consistently in `get_supported_az_for_multi_instance_types` as well?

cli/pcluster/configure/easyconfig.py
```diff
@@ -199,56 +188,85 @@ def _reset_config_params(section, parameters_to_remove):
         param.value = param.get_default_value()
 
 
-def _create_vpc_parameters(vpc_section, scheduler, min_subnet_size, automate_vpc_creation=True):
+def _create_vpc_parameters(vpc_section, scheduler, cluster_config):
     vpc_parameters = {}
-    min_subnet_size = int(min_subnet_size)
+    min_subnet_size = int(cluster_config.max_cluster_size)
+    cluster_config.cache_qualified_az()
```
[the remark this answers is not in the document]
hanwen-pcluste: Yep, this line is weird. Will change it

cli/pcluster/configure/easyconfig.py
```diff
@@ -273,18 +280,47 @@ def _ask_for_subnets(subnet_list, vpc_section):
...
+    azs_for_master_type = get_supported_az_for_one_instance_type(cluster_config.master_instance_type)
+    azs_for_compute_type = get_supported_az_for_one_instance_type(cluster_config.compute_instance_type)
+    common_availability_zones = set(azs_for_master_type) & set(azs_for_compute_type)
+
+    if not common_availability_zones:
+        print(
+            "Error: There is no single availability zone offering master and compute in current region.\n"
+            "To create your cluster, make sure you have a subnet for master node in {0}"
+            ", and a subnet for compute nodes in {1}. Then run pcluster configure again"
+            "and avoid using Automate VPC/Subnet creation.".format(azs_for_master_type, azs_for_compute_type)
```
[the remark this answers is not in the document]
hanwen-pcluste: Thank you for the rewording. I will change it

cli/pcluster/configure/utils.py
hanwen-pcluste: I will change it

cli/pcluster/configure/utils.py
hanwen-pcluste: I will change it

cli/pcluster/utils.py
hanwen-pcluste: I will just remove it

cli/pcluster/utils.py
hanwen-pcluste: Why list over tuple? I thought tuple is immutable, so better suited for the code here.

cli/pcluster/utils.py
hanwen-pcluste: Good catch. The code was adapted from the previous code. I will think more about the catch statement.

cli/pcluster/configure/easyconfig.py
```diff
@@ -214,57 +203,76 @@ def _reset_config_params(section, parameters_to_remove):
...
+                or not qualified_compute_subnets
+                or (prompt("Automate Subnet creation? (y/n)", lambda x: x in ("y", "n"), default_value="y") == "y")
             ):
+                # Start auto subnets creation in the absence of qualified subnets.
+                # Otherwise, user selects between manual and automate subnets creation
+                if not qualified_master_subnets or not qualified_compute_subnets:
+                    print("There are no qualified subnets. Starting automatic creation of subnets...")
                 vpc_parameters.update(
-                    automate_subnet_creation(vpc_id, _choose_network_configuration(scheduler), min_subnet_size)
+                    automate_subnet_creation(vpc_id, _choose_network_configuration(cluster_config), min_subnet_size)
```
rexcsn, line 251: Minor: I would have wrote it like this:
```
if not qualified_master_subnets or not qualified_compute_subnets:
    print("There are no qualified subnets. Starting automatic creation of subnets...")
    vpc_parameters.update ...
elif prompt("Automate Subnet creation? (y/n)", lambda x: x in ("y", "n"), default_value="y") == "y":
    vpc_parameters.update ...
else:
    ...
```

rexcsn (approved):
Nice work! 🚀

PR #1961, opened 2020-08-20
Add the some tests for torque-slurm-wrapper

yuleiwan:
* Test torque job submit command in slurm cluster(qsub)
* Show the computing nodes information by torque command in slurm cluster(pbsnodes)

### DIFF
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -428,9 +428,7 @@ class TorqueCommands(SchedulerCommands):
         assert_that(status).is_equal_to("0")
 
     def compute_nodes_count(self):  # noqa: D102
-        result = self._remote_command_executor.run_remote_command(
-            "echo $(( $(/opt/torque/bin/pbsnodes -l all | wc -l) - 1))"
-        )
+        result = self._remote_command_executor.run_remote_command("echo $(( $(pbsnodes -l all | wc -l) - 1))")
         # split()[-1] to extract last line and trim whitespaces
         return int(result.stdout.split()[-1])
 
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -434,6 +434,10 @@ def compute_nodes_count(self):  # noqa: D102
         # split()[-1] to extract last line and trim whitespaces
         return int(result.stdout.split()[-1])
 
+    def compute_nodes_count_wrapper(self):  # noqa: D102
+        result = self._remote_command_executor.run_remote_command("echo $(/opt/slurm/bin/pbsnodes | grep ip | wc -l)")
```
lukeseawalker, line 438: I don't like that we are using a `/opt/slurm/bin/` command path in the `TorqueCommands` class. When calling the function from outside you should not have to decide which function to call (`compute_nodes_count` or `compute_nodes_count_wrapper`) base on the running scheduler.
Is it possible to remove the full command path from the `compute_nodes_count` function and just leave the command `pbsnodes` so that the function can be called indifferently from Slurm and from Torque case?

tests/integration-tests/tests/schedulers/test_slurm/test_slurm/torque_job.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+echo "sleep 100"
```
lukeseawalker, line 2: Any particular reason why `100` and not `60 ` for example?

yuleiwan: The script purpose is to test if we can submit the job by torque command. So we can use any word in `echo` script. Both `echo "sleep 100"` or `echo "sleep 60" ` will be executed in less than 1 second, so I just randomly selected a number.

lukeseawalker (approved):

PR #1976, opened 2020-08-26
Split Lambda cleanup functions and add possibility to disable Route53 creation

enrico-usai:
# Lambda split
The lambda to cleanup the S3 Bucket must be executed as soon as possible instead the lambda functions to terminate the instances and to cleanup the route53 must be executed after the master is stopped to avoid other instances or record set to be added after the Lambda execution.

Moved the Lambda related code and the policy specific to HIT in the compute fleet hit substack.
We are creating a policy in the substack and attaching it to the RootRole if the customer is not using a custom EC2 IAM Role

The trick to have a dependency in the Master substack is to add the TerminateComputeFleetCustomResource and the CleanupRoute53CustomResource as custom parameter "DependsOnCustomResources".

# Disabling Route53 creation
From now it is possible to add `disable_cluster_dns = True` as configuration parameter to avoid the creation of the Route53 resource and related Lambda cleanup function.

To have a valid substack when Route53 resource is not created we  are using input parameters as metadata.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -1315,6 +1315,13 @@
               "false"
             ]
           },
+          "CreateCleanupRoute53FunctionLogGroup": {
+            "Fn::If": [
+              "CreateHITSubstack",
+              "true",
+              "false"
+            ]
+          },
           "CreateAWSBatchLogGroups": {
             "Fn::If": [
               "CreateAWSBatchStack",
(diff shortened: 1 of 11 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -3861,6 +3798,21 @@
...
+              "NONE",
+              {
+                "Ref": "RootRole"
+              }
+            ]
+          },
+          "ResourcesS3Bucket": {
+            "Ref": "ResourcesS3Bucket"
+          },
+          "CloudWatchLogsSubstack": {
```
demartinofra, line 3814: not needed

enrico-usai: Removed

demartinofra (approved):

PR #1977, opened 2020-08-26
 Add EC2 fleet clean-up to pcluster delete

demartinofra:
This is useful in case the stack fails to delete all EC2 instances.
This can happen in a HIT cluster where there is no ASG grouping the compute fleet anymore and the CustomResource performing the fleet cleanup could fail and leave orphaned nodes.


cli/pcluster/cli_commands/delete.py
```diff
@@ -33,6 +35,7 @@ def delete(args):
                 )
             )
         utils.warn("Cluster {0} has already been deleted.".format(args.cluster_name))
+        _terminate_cluster_nodes(stack_name)
```
ddeidda, line 38: Why are we terminating nodes here if the stack does not exist?

cli/pcluster/cli_commands/delete.py
```diff
@@ -117,5 +120,41 @@ def _delete_cluster(cluster_name, nowait):
         sys.stdout.flush()
         sys.exit(1)
     except KeyboardInterrupt:
+        terminate_compute_fleet = False
         LOGGER.info("\nExiting...")
         sys.exit(0)
+    finally:
+        if terminate_compute_fleet:
```
ddeidda, line 127: What happens if terminate_compute_fleet is False and there are node running? Will they just be running until their internal daemons shut them down?

cli/pcluster/cli_commands/delete.py
```diff
@@ -117,5 +120,41 @@ def _delete_cluster(cluster_name, nowait):
...
+
+
+def _describe_instance_ids_iterator(stack_name, instance_state=("pending", "running", "stopping", "stopped")):
+    ec2 = boto3.client("ec2")
+    filters = [
+        {"Name": "tag:Application", "Values": [stack_name]},
+        {"Name": "instance-state-name", "Values": list(instance_state)},
+        {"Name": "tag:aws-parallelcluster-node-type", "Values": [str(NodeType.compute)]},
+    ]
+    paginator = ec2.get_paginator("describe_instances")
```
ddeidda, line 154: Could have we used the utility `paginate_boto3` here?

cli/pcluster/cli_commands/delete.py
demartinofra: Consider this case:
- user runs a pcluster delete and the CustomResource fails for any reason to terminate some or all EC2 instances. Stack is set to DELETE_FAILED but the Lambda function for the cleanup is removed and this can't be avoided
- user runs again a pcluster delete or deletes the stack from the console and the stack completes successfully because there is no lambda function handling the cleanup

Because of this it is unlikely but possible to end up in situations where the stack does not exist but capacity for that cluster is running. pcluster delete at that point can be used to perform a clean-up

cli/pcluster/cli_commands/delete.py
demartinofra: There is anyway a CustomResource ran as part of the stack deletion that tries to delete all of them

cli/pcluster/cli_commands/delete.py
demartinofra: yes I can switch to that

cli/pcluster/cli_commands/delete.py
ddeidda: OK. I'm wondering if the user couldn't get confused by seeing the above message and the ones coming from `_terminate_cluster_nodes` when running the command against a wrong - non existent cluster name. It's a very minor issue I know; maybe we could just print an additional message informing that we are trying a clean up in any case, or only print messages only when we are really terminating instances.
Also a FIXME comment to address this would be OK if we are short in time.

demartinofra: Slightly rephrased and added a FIXME

cli/pcluster/cli_commands/delete.py
demartinofra: done

ddeidda (approved):
Let's merge it! :+1:

PR #1987, opened 2020-08-28
Add Cluster name and queue_settings validators and fix file_path regex

agobeaux:
Add Cluster name and queue_settings validators and fix file_path regex

Before this commit, the cluster name and queue_settings were not validated. Now, those are verified to ensure that they consist only of alphanumeric characters and hyphens (but begin only with letters) and that the lengths of the cluster name and of the each queue name in the queue_settings are respectively less than 60 and 30. Testing for those modifications has also been added.

The limitation on the cluster name length has been chosen based on two observed limitations:
- The LaunchTemplateName in compute-fleet-hit.substack.cfn.yaml has name ${ClusterName}-{{ queue }}-{{ instance_type }} and the LaunchTemplateName can have maximum 128 characters. Choosing 30 for the queue name, 21 for the instance type name, we still have 75 chars left for the queue name;
- The more limiting factor comes with HostedZone names which consist of a series of labels separated by dots. Each label can be up to 63 bytes long. The length of the cluster name can thus be at most 63 chars (as we do not want to introduce dots, which are subdomains in DNS).

Limitation on queue name: The hostname is queuename-st|dy-instancetype-N with N being a number and this hostname should be less than 64 chars long. That is why a length of 30 chars was chosen for the queue name.

Resources section in the main stack and in batch substack have been changed as the ClusterName/AWS::StackName gets truncated by CloudFormation (max 100 characters when concatenating (using hyphens) the main stack name with substack names with 13 random digits). The most limiting substack name is ComputeFleetHITSubstack.

The file path regex has also been fixed. Testing about this regex has also been improved (modified and created some).

A little typo has been fixed in the cloudformation README to provide the real command to render troposphere templates.


cli/tests/pcluster/test_commands.py
```diff
@@ -160,3 +161,28 @@ def test_update_failure_on_different_cluster_models(
         assert_that(caplog.text).contains(expected_message)
     else:
         assert_that(caplog.text).is_empty()
+
+
+@pytest.mark.parametrize(
+    "cluster_name, should_trigger_error",
```
enrico-usai, line 167: minor: should_trigger_error --> expected_error (to be aligned with other unit tests)

cli/tests/pcluster/config/test_section_cluster.py
```diff
@@ -501,6 +513,246 @@ def test_sit_cluster_param_from_file(
...
+                "ERROR: The configuration parameter 'queue_settings' has an invalid value '.*'\nAllowed values are: "
+                "\\^\\[a-zA-Z\\]\\[a-zA-Z0-9-\\]\\{0,29\\}\\(,\\\\s\\*\\[a-zA-Z\\]\\[a-zA-Z0-9-\\]\\{0,29\\}\\)\\*\\$"
+            ),
+        ),
+        (
+            "queue_settings",
+            "LongQueueNameJustOneCharTooLong",
+            None,
+            (
+                "ERROR: The configuration parameter 'queue_settings' has an invalid value '.*'\nAllowed values are: "
```
enrico-usai, line 733: minor you can avoid to put the regex, you can just write: 
`ERROR: The configuration parameter 'queue_settings' has an invalid value .*`
or `.*  'queue_settings' has an invalid value .*`

In this way we don't need to change this failing test when changing the regex

cli/pcluster/config/mappings.py
```diff
@@ -128,9 +128,10 @@
             r"([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])"
             r"(\/([0-9]|[1-2][0-9]|3[0-2]))$",
     "efs_fs_id": r"^fs-[0-9a-z]{8}$|^fs-[0-9a-z]{17}|NONE$",
-    "file_path": r"\/?[\w:]+",
+    "file_path": r"^\/?[^\/.\\][^\/\\]*(\/[^\/.\\][^\/]*)*$",
+    "queue_settings": r"^[a-zA-Z][a-zA-Z0-9-]{0,29}(,\s*[a-zA-Z][a-zA-Z0-9-]{0,29})*$",
     "fsx_fs_id": r"^fs-[0-9a-z]{17}|NONE$",
-    "greater_than_25": r"^([0-9]+[0-9]{2}|[3-9][0-9]|2[5-9])$",
+    "greater_than_25": r"^([1-9][0-9]*[0-9]{2}|[3-9][0-9]|2[5-9])$",
```
enrico-usai, line 134: if you fixed the regex let's add a test to cover the wrong case that we were supporting before.

cli/tests/pcluster/test_commands.py
agobeaux: I could do it, the only problem is that the expected_errors variables are usually not used for booleans but for strings containing the error. I could put the error instead of True and None instead of the error also, this would be strange as the error is the same every time but could be nice also.

cli/pcluster/config/mappings.py
agobeaux: From testing a bit further, my change did not fix anything as the number is apparently taken as an int (which erases the leading 0) and then compared using this regex. This means that my change is not useful since leading zeros will not be there. Should I delete it or just let it as is, as it seems more logical when reading?
This regex would still allow 025 as 025 would first be rendered to 25 and then passed to the regex. The regex would thus not match what is expected if we let my change.
Testing this with leading zeros would thus not be useful as they would not appear anyway.

cli/tests/pcluster/config/test_section_cluster.py
agobeaux: Done

cli/pcluster/commands.py
```diff
@@ -111,10 +112,24 @@ def _check_for_updates(pcluster_config):
         utils.check_if_latest_version()
 
 
+def _validate_cluster_name(cluster_name):
+    if not re.match(PCLUSTER_NAME_REGEX % (PCLUSTER_NAME_MAX_LENGTH - 1), cluster_name):
+        LOGGER.critical(
```
demartinofra, line 117: I'd not log this as critical, but rather as error. Also you could simply use the error function from utils.

As a side node: in general favor format over printf based formatting (%)- https://realpython.com/python-string-formatting/
With logger you don't need to use `%` but just pass the variables as arguments after the parametrized string

cli/tests/pcluster/test_commands.py
agobeaux: Spoke with Francesco, we leave as is now

cli/pcluster/config/mappings.py
agobeaux: As it is a useless change, we let as it was before

cli/pcluster/commands.py
agobeaux: used LOGGER.error instead, thank you for the link

demartinofra (approved):

PR #1997, opened 2020-09-01
 Add section label verifiers

agobeaux:
Before this commit, section labels were not verified. Now we verify that the section labels start with a letter, contain only alphanumeric characters, hyphens and underscores (except for the queue labels which should not contain underscores (it is filtered in the queue_settings validator). Concerning the queue label, another check has been made: it should not consist only of the string 'default' (either in lower case, upper case, or a mix of the two) but it can contain it (for example, my-default-queue is allowed). Some tests have also been added.


demartinofra (approved):

demartinofra:
Well done 🚀

PR #2007, opened 2020-09-03
Warn user to delete S3 bucket

agobeaux:
Before this commit, if the user tried to cancel the cluster creation, he would not be aware that his bucket would not be deleted and would end up deleting his cluster and having multiple unused S3 buckets on his account.

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -198,6 +198,11 @@ def create(args):  # noqa: C901 FIXME!!!
         sys.exit(1)
     except KeyboardInterrupt:
         LOGGER.info("\nExiting...")
+        if bucket_name:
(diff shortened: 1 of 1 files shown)


cli/pcluster/commands.py
```diff
@@ -198,6 +198,11 @@ def create(args):  # noqa: C901 FIXME!!!
         sys.exit(1)
     except KeyboardInterrupt:
         LOGGER.info("\nExiting...")
+        if bucket_name:
+            LOGGER.info(
+                "Your S3 bucket named %s was not deleted. Do not forget to delete it when deleting your cluster.",
```
demartinofra, line 203: the bucket is automatically deleted when the cluster is deleted, unless there is an very early failure

PR #2008, opened 2020-09-03
Fix sns subscription

ddeidda:
As reported in [this issue](https://github.com/serverless/serverless/issues/5149) SNS subscription resource is not deleted when defined "inline" inside the SNSTopic resource. Explicitly declaring a SNS Subscription resource fixes the problem.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -1460,18 +1460,21 @@
     },
     "SNS": {
       "Type": "AWS::SNS::Topic",
+      "Condition": "CreateComputeFleet"
+    },
+    "SNSSubscription": {
+      "Type": "AWS::SNS::Subscription",
       "Properties": {
-        "Subscription": [
-          {
-            "Endpoint": {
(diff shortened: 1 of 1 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -1459,19 +1459,21 @@
...
+            "SQS",
+            "Arn"
+          ]
+        },
+        "Protocol": "sqs",
+        "TopicArn": {
+          "Ref": "SNS"
+        }
       },
       "Condition": "CreateComputeFleet"
```
rexcsn, line 1478: Do we need this condition for "SNS" as well?

demartinofra: I think we don't need an SNSSubscription nor an SNS topic with awsbatch and slurm, which makes me think yes we need this condition and we need to also apply it to the SNS above

ddeidda: Done

demartinofra (approved):

PR #2014, opened 2020-09-03
Stop log error when getting availability zone for "NONE" subnet

hanwen-pcluste:
If master and compute nodes are in the same subnet, the function can be called with "NONE" subnet_id because compute subnet is not specified.

### DIFF
--- a/cli/pcluster/utils.py
+++ b/cli/pcluster/utils.py
@@ -393,6 +393,8 @@ def get_availability_zone_of_subnet(subnet_id):
     :param subnet_id: the id of the subnet.
     :return: a strings of availability zone name
     """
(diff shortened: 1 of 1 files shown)


cli/pcluster/utils.py
```diff
@@ -393,6 +393,8 @@ def get_availability_zone_of_subnet(subnet_id):
     :param subnet_id: the id of the subnet.
     :return: a strings of availability zone name
     """
+    if subnet_id == "NONE":
```
rexcsn, line 396: This could hide potential problems if say master_subnet_id is not read in correctly. I would say let's keep things as is and open an issue to improve this in our backlog.

PR #2025, opened 2020-09-08
Do not load json config on start/stop

demartinofra:
The main reason behind this change is that we don't want the start/stop command to require a downloadof the json config, which would imply a dependency on the cluster s3 bucket and cluster json configuration. This allows us to:
1. have a working start/stop also when the bucket is deleted
2. have a minimal policy for start/stop that does not require s3 access


enrico-usai (approved):
Let's add the reason in the commit message or in the PR description for which we're removing this configuration loading. 

We should also remember that when using this option we'll have an incomplete PclusterConfig object.

PR #2029, opened 2020-09-10
Add validator for EBS volumes IOPS

chenwany:
Signed-off-by: chenwany <wanyichen0521@gmail.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/pcluster/config/mappings.py
+++ b/cli/pcluster/config/mappings.py
@@ -56,6 +56,7 @@ from pcluster.config.validators import (
     disable_hyperthreading_architecture_validator,
     disable_hyperthreading_validator,
     ebs_settings_validator,
+    ebs_volume_iops_validator,
     ebs_volume_type_size_validator,
     ec2_ami_validator,
     ec2_ebs_snapshot_validator,
@@ -291,7 +292,7 @@ EBS = {
     "key": "ebs",
     "default_label": "default",
     "max_resources": 5,
-    "validators": [ebs_volume_type_size_validator],
+    "validators": [ebs_volume_type_size_validator, ebs_volume_iops_validator],
     "params": {
         "shared_dir": {
             "allowed_values": ALLOWED_VALUES["file_path"],
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -1304,3 +1304,31 @@ def ebs_volume_type_size_validator(section_key, section_label, pcluster_config):
             errors.append("The size of {0} volumes must be at least {1} GiB".format(volume_type, min_size))
 
     return errors, warnings
+
+
+def ebs_volume_iops_validator(section_key, section_label, pcluster_config):
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section(section_key, section_label)
+    volume_size = section.get_param_value("volume_size")
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    volume_iops = section.get_param_value("volume_iops")
+    min_iops = 100
+    max_iops = 64000
+    if volume_type in volume_type_to_iops_ratio:
+        if volume_iops and (volume_iops < min_iops or volume_iops > max_iops):
+            errors.append(
+                "IOPS rate must be between {min_iops} and {max_iops} when provisioning {volume_type} volumes.".format(
+                    min_iops=min_iops, max_iops=max_iops, volume_type=volume_type
+                )
+            )
+        if volume_iops and volume_iops > volume_size * volume_type_to_iops_ratio[volume_type]:
+            errors.append(
+                "IOPS to volume size ratio of {0} is too high; maximum is {1}.".format(
+                    float(volume_iops) / float(volume_size), volume_type_to_iops_ratio[volume_type]
+                )
+            )
+
+    return errors, warnings
--- a/cli/tests/pcluster/config/test_validators.py
+++ b/cli/tests/pcluster/config/test_validators.py
@@ -2095,3 +2095,35 @@ def test_ebs_allowed_values_all_have_volume_size_bounds():
         EBS_VOLUME_TYPE_TO_VOLUME_SIZE_BOUNDS.keys()
     )
     assert_that(allowed_values_all_have_volume_size_bounds).is_true()
+
+
+@pytest.mark.parametrize(
+    "section_dict, expected_message",
+    [
+        ({"volume_type": "io1", "volume_size": 20, "volume_iops": 120}, None),
+        (
+            {"volume_type": "io1", "volume_size": 20, "volume_iops": 90},
+            "IOPS rate must be between 100 and 64000 when provisioning io1 volumes.",
+        ),
+        (
+            {"volume_type": "io1", "volume_size": 20, "volume_iops": 64001},
+            "IOPS rate must be between 100 and 64000 when provisioning io1 volumes.",
+        ),
+        ({"volume_type": "io1", "volume_size": 20, "volume_iops": 1001}, "IOPS to volume size ratio of .* is too hig"),
+        # TODO Uncomment these lines after adding support for io2 volume types
+        #         ({"volume_type": "io2", "volume_size": 20, "volume_iops": 120}, None),
+        #         (
+        #             {"volume_type": "io2", "volume_size": 20, "volume_iops": 90},
+        #             "IOPS rate must be between 100 and 64000 when provisioning io2 volumes.",
+        #         ),
+        #         (
+        #             {"volume_type": "io2", "volume_size": 20, "volume_iops": 64001},
+        #             "IOPS rate must be between 100 and 64000 when provisioning io2 volumes.",
+        #         ),
+        #         ({"volume_type": "io2", "volume_size": 20, "volume_iops": 10001},
+        #         "IOPS to volume size ratio of .* is too hig"),
+    ],
+)
+def test_ebs_volume_iops_validator(mocker, section_dict, expected_message):
+    config_parser_dict = {"cluster default": {"ebs_settings": "default"}, "ebs default": section_dict}
+    utils.assert_param_validator(mocker, config_parser_dict, expected_message)


cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
 def _describe_ec2_key_pair(key_pair_name):
     """Return information about the provided ec2 key pair."""
     return boto3.client("ec2").describe_key_pairs(KeyNames=[key_pair_name])
+
+
+def ebs_volume_iops_validator(section_key, section_label, pcluster_config):
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section("ebs", section_label)
```
tilne, line 1262: use `section_key` instead of `"ebs"` here

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    volume_size = float(section.get_param_value("volume_size"))
```
tilne, line 1263: why cast this to a float?

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    if section.get_param_value("volume_iops"):
+        volume_iops = float(section.get_param_value("volume_iops"))
```
tilne, line 1267: same question here: why cast this to a float?

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+
+    if volume_type == "io1" or volume_type == "io2":
```
tilne, line 1269: instead of comparing `volume_type` against each of the accepted values, you could check if it's in the dictionary's keys.

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+        if volume_iops < 100 or volume_iops > 64000:
```
tilne, line 1270: i would assign 100 and 64000 to variables that describe what the values represent, maybe like `min_iops` and `max_iops`

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+            errors.append("The EBS volume can be provisioned with from 100 IOPS up to 64,000 IOPS per volume")
+        if volume_iops > volume_size * volume_type_to_iops_ratio[volume_type]:
+            errors.append(
+                "IOPS to volume size ratio of {0} is too high; maximum is 50.".format(volume_iops / volume_size)
```
tilne, line 1274: cast to float here since this is the only place you do division. also, use the max value from the dict instead of hard-coding 50 (which i think only applies to io1).

cli/pcluster/config/mappings.py
```diff
@@ -135,7 +136,7 @@
     "snapshot_id": r"^snap-[0-9a-z]{8}$|^snap-[0-9a-z]{17}$",
     "subnet_id": r"^subnet-[0-9a-z]{8}$|^subnet-[0-9a-z]{17}$",
     "volume_id": r"^vol-[0-9a-z]{8}$|^vol-[0-9a-z]{17}$",
-    "volume_types": ["standard", "io1", "gp2", "st1", "sc1"],
+    "volume_types": ["standard", "io1", "io2", "gp2", "st1", "sc1"],
```
tilne, line 139: let's remove io2 from this list for now. the io2 logic can remain in the validators, and the test cases can be commented out with a TODO comment explaining that they should be enabled when io2 support is added

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,25 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section("ebs", section_label)
+    volume_size = float(section.get_param_value("volume_size"))
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    if section.get_param_value("volume_iops"):
+        volume_iops = float(section.get_param_value("volume_iops"))
+
```
tilne, line 1268: I would add a check that, if volume_iops is set in the config and the volume_type does not support specifying a specific IOPS value, then the validator should add an error for that.

this would probably require changing the default value of volume_iops from an int literal to a function that returns 100 if the volume type supports provisioning a specific IOPS value, otherwise it returns None.

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,32 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+        if volume_iops < min_iops or volume_iops > max_iops:
+            errors.append(
+                "The EBS volume can be provisioned with from {0} IOPS up to {1} IOPS per volume.".format(
+                    min_iops, max_iops
+                )
+            )
+        if volume_iops > volume_size * volume_type_to_iops_ratio[volume_type]:
+            errors.append(
+                "IOPS to volume size ratio of {0} is too high; maximum is {1}.".format(
+                    float(volume_iops / volume_size), volume_type_to_iops_ratio[volume_type]
```
tilne, line 1280: The values each need to be cast to floats individually. For an example of why, use a python2  interpreter to print out the value of `float(9 / 2)` and `float(9) / float(2)`.

cli/pcluster/config/mappings.py
```diff
@@ -135,7 +136,7 @@
     "snapshot_id": r"^snap-[0-9a-z]{8}$|^snap-[0-9a-z]{17}$",
     "subnet_id": r"^subnet-[0-9a-z]{8}$|^subnet-[0-9a-z]{17}$",
     "volume_id": r"^vol-[0-9a-z]{8}$|^vol-[0-9a-z]{17}$",
-    "volume_types": ["standard", "io1", "gp2", "st1", "sc1"],
+    "volume_types": ["standard", "io1", "gp2", "st1", "sc1"],  # TODO Add io2 into the list after making it supported
```
tilne, line 139: you can remove this TODO

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,32 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+def ebs_volume_iops_validator(section_key, section_label, pcluster_config):
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section(section_key, section_label)
+    volume_size = section.get_param_value("volume_size")
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    if section.get_param_value("volume_iops"):
+        volume_iops = section.get_param_value("volume_iops")
```
tilne, line 1267: I would avoid this conditional. Always set `volume_iops` regardless of what `get_param_value` returns. Then instead of checking whether the parameter is set, you can check that in the conditional you use to decide whether or not to test the values.

```python
if volume_type in volume_type_to_iops_ration.keys() and volume_iops:
  # validation logic
```

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,32 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    min_iops = 100
+    max_iops = 64000
+    if volume_type in volume_type_to_iops_ratio.keys():
```
tilne, line 1270: you can omit the `.keys()` here

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,31 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+                )
+            )
+        if volume_iops > volume_size * volume_type_to_iops_ratio[volume_type]:
+            errors.append(
+                "IOPS to volume size ratio of {0} is too high; maximum is {1}.".format(
+                    float(volume_iops) / float(volume_size), volume_type_to_iops_ratio[volume_type]
+                )
+            )
+
+    return errors, warnings
```
tilne, line 1283: Even though `volume_size` and `volume_iops` have default values, I would still check both for values of `None` in case we get rid of the default values (as we have discussed doing as part of handling snapshots better).

If these checks aren't performed and the default values are removed, then it's possible that the arithmetic operations used in the validation logic will raise `TypeError`s due to trying to multiply or compare values with an int and `None`.

Thus:
* line 1270 would become `if volume_iops and (volume_iops < min_iops or volume_iops > max_iops):`
* line 1276 would become `if volume_iops and volume_size and volume_iops > volume_size * volume_type_to_iops_ratio[volume_type]:`

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,31 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    volume_size = section.get_param_value("volume_size")
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    volume_iops = section.get_param_value("volume_iops")
+    min_iops = 100
+    max_iops = 64000
+    if volume_type in volume_type_to_iops_ratio:
+        if volume_iops < min_iops or volume_iops > max_iops:
+            errors.append(
+                "The EBS volume can be provisioned with from {0} IOPS up to {1} IOPS per volume.".format(
```
tilne, line 1272: I would change this error message to something like "IOPS rate must be between {min_iops} to {max_iops} when provisioning {volume_type} volumes."

cli/pcluster/config/validators.py
chenwany: Yeah! That's right! It is better to check for value of None. Done!

cli/pcluster/config/validators.py
chenwany: yes! Done.

cli/pcluster/config/validators.py
chenwany: Done.

cli/pcluster/config/validators.py
chenwany: Done! Removed it here and added it in the line of division.

cli/pcluster/config/validators.py
chenwany: Done! Removed it here and added it in the line of division.

cli/pcluster/config/validators.py
chenwany: Done.

cli/pcluster/config/validators.py
chenwany: Good idea. Done!

cli/pcluster/config/validators.py
chenwany: Yeah! For io2 it should be 500. Done!

cli/pcluster/config/mappings.py
chenwany: Done

cli/pcluster/config/validators.py
chenwany: Haven't done this in this commit. Need another commit to add this check

cli/pcluster/config/validators.py
chenwany: Fixed.

cli/pcluster/config/mappings.py
chenwany: Done

cli/pcluster/config/validators.py
chenwany: Done

cli/pcluster/config/validators.py
chenwany: Done

cli/pcluster/config/validators.py
```diff
@@ -1253,3 +1253,31 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
...
+    volume_size = section.get_param_value("volume_size")
+    volume_type = section.get_param_value("volume_type")
+    volume_type_to_iops_ratio = {"io1": 50, "io2": 500}
+    volume_iops = section.get_param_value("volume_iops")
+    min_iops = 100
+    max_iops = 64000
+    if volume_type in volume_type_to_iops_ratio:
+        if volume_iops and (volume_iops < min_iops or volume_iops > max_iops):
+            errors.append(
+                "IOPS rate must be between {min_iops} to {max_iops} when provisioning {volume_type} volumes.".format(
```
tilne, line 1272: sorry if my previous comment had a typo in it: this should be "{min_iops} and {max_iops}" instead of "{min_iops} to {max_iops}"

chenwany: Done!

tilne (approved):

PR #2037, opened 2020-09-11
Add logic to cleanup pcluster_config into the convert utility

enrico-usai:
Before of this patch we were cleaning up the config_parser only when executin the pcluster configure utility.

By adding the prepare_to_file function we're sure to have the pcluster_config ready to be written in the file.

## Conversion logic
When a default is equal to default should not be written in the file.
It means: disable_hyperthreading = false, initial_count = min_count and min_count = 0
Instead enable_efa in the queue will be always written because the default is None

## Tests
Improved unit tests by covering all the enable_efa and dysable_hyperthreading cases.

### DIFF
--- a/cli/pcluster/config/hit_converter.py
+++ b/cli/pcluster/config/hit_converter.py
@@ -25,7 +25,7 @@ class HitConverter:
     def __init__(self, pcluster_config):
         self.pcluster_config = pcluster_config
 
-    def convert(self):
+    def convert(self, prepare_to_file=False):
         """
         Convert the pcluster_config instance from pre-HIT to HIT configuration model.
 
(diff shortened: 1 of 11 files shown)


cli/pcluster/config/hit_converter.py
```diff
@@ -147,6 +150,29 @@ def _copy_param_value(self, old_param, new_param, new_value=None):
...
+    @staticmethod
+    def _reset_config_params(section, parameters_to_remove):
+        for param_key in parameters_to_remove:
+            param = section.get_param(param_key)
+            param.value = param.get_default_value()
+
+    def _prepare_to_file(self):
+        """Reset some parameters from config file since their values can be inferred at runtime."""
+
+        # enable_efa and disable_hyperthreading will get their value from cluster section
```
ddeidda: remove enable_efa from comment

ddeidda (approved):

demartinofra (approved):

PR #2038, opened 2020-09-11
Disallow using AMIs created by other versions of pcluster by AMI_name…

hanwen-pcluste:
… if the name contains pcluster version

As a result, a AMI created by other versions of pcluster cannot be used as a BASE_AMI in AMI creation or a custom_AMI in cluster creation. However, if the AMI_name does not contain pcluster version, the code in CLI package has to allow the AMI and leave the validation to CloudFormation and Cookbook stage.

### DIFF
--- a/cli/pcluster/utils.py
+++ b/cli/pcluster/utils.py
@@ -771,6 +771,19 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
+def validate_pcluster_version_based_on_ami_name(ami_name):
+    match = re.match(r"(.*)aws-parallelcluster-(\d+\.\d+\.\d+)(.*)", ami_name)
+    if match:
+        if match.group(2) != get_installed_version():
+            error(
+                "This AMI was created with version {0} of ParallelCluster,"
+                " but is trying to be used with version {1}. Please either use an AMI created with "
+                "version {1} or change your ParallelCluster to version {0}".format(
+                    match.group(2), get_installed_version()
+                )
+            )
+
+
 def get_instance_types_info(instance_types, fail_on_error=True):
     """Return InstanceTypes list returned by EC2's DescribeInstanceTypes API."""
     try:
--- a/cli/tests/pcluster/createami/test_pcluster_createami.py
+++ b/cli/tests/pcluster/createami/test_pcluster_createami.py
@@ -84,8 +84,12 @@ def test_validate_createami_args_architecture_compatibility(
     mocker, base_ami_id, instance_type, base_ami_os, base_ami_architecture, supported_instance_archs, supported_os
 ):
     """Verify that parameter validation works as expected in the function that implements the createami command."""
-    mocker.patch("pcluster.createami.utils.get_info_for_amis").return_value = [{"Architecture": base_ami_architecture}]
+    base_ami_name = "ami-x"
+    mocker.patch("pcluster.createami.utils.get_info_for_amis").return_value = [
+        {"Architecture": base_ami_architecture, "Name": base_ami_name}
+    ]
     mocker.patch("pcluster.createami._get_default_createami_instance_type")
+    mocker.patch("pcluster.createami.utils.validate_pcluster_version_based_on_ami_name")
     mocker.patch(
         "pcluster.createami.utils.get_supported_architectures_for_instance_type"
     ).return_value = supported_instance_archs
@@ -100,12 +104,13 @@ def test_validate_createami_args_architecture_compatibility(
     )
     if error_expected:
         with pytest.raises(SystemExit) as sysexit:
-            createami._validate_createami_args_architecture_compatibility(args)
+            createami._validate_createami_args_ami_compatibility(args)
         assert_that(sysexit.value.code).is_not_equal_to(0)
     else:
-        assert_that(createami._validate_createami_args_architecture_compatibility(args)).is_equal_to(
-            base_ami_architecture
+        assert_that(createami._validate_createami_args_ami_compatibility(args)).is_equal_to(
+            {"Architecture": base_ami_architecture, "Name": base_ami_name}
         )
+        createami.utils.validate_pcluster_version_based_on_ami_name.assert_called_with(base_ami_name)
 
     createami.utils.get_info_for_amis.assert_called_with([base_ami_id])
 
--- a/cli/tests/pcluster/utils/test_pcluster_utils.py
+++ b/cli/tests/pcluster/utils/test_pcluster_utils.py
@@ -667,3 +667,29 @@ def test_is_hit_enabled_cluster(scheduler, expected_is_hit_enabled):
 )
 def test_get_bucket_url(region, expected_url):
     assert_that(get_bucket_url(region)).is_equal_to(expected_url)
+
+
+@pytest.mark.parametrize(
+    "ami_name, error_expected, expected_message",
+    [
+        # Compatible ami name
+        ("ami-xxxaws-parallelcluster-2.9.0xxx", False, ""),
+        ("ami-aws-parallelcluster-2.9.0", False, ""),
+        ("ami-name-one", False, ""),
+        ("aws-parallelcluster-2.9.0-ubuntu-1804-lts-hvm-x86_64-202009142226", False, ""),
+        # Incompatible ami name
+        (
+            "ami-aws-parallelcluster-0.0.0",
+            True,
+            "This AMI was created with version 0.0.0 of ParallelCluster, but is trying to be used with version 2.9.0."
+            " Please either use an AMI created with version 2.9.0 or change your ParallelCluster to version 0.0.0",
+        ),
+    ],
+)
+def test_validate_pcluster_version_based_on_ami_name(mocker, ami_name, error_expected, expected_message):
+    mocker.patch("pcluster.utils.get_installed_version", return_value="2.9.0")
+    if error_expected:
+        with pytest.raises(SystemExit, match=expected_message):
+            utils.validate_pcluster_version_based_on_ami_name(ami_name)
+    else:
+        utils.validate_pcluster_version_based_on_ami_name(ami_name)
--- a/cloudformation/compute-fleet-hit-substack.cfn.yaml
+++ b/cloudformation/compute-fleet-hit-substack.cfn.yaml
@@ -365,7 +365,7 @@ Resources:
               if [ -f /opt/parallelcluster/.bootstrapped ]; then
                 installed_version=$(cat /opt/parallelcluster/.bootstrapped)
                 if [ "${!parallelcluster_version}" != "${!installed_version}" ]; then
-                  bootstrap_instance
+                  error_exit "This AMI was created with ${!installed_version}, but is trying to be used with ${!parallelcluster_version}. Please either use an AMI created with ${!parallelcluster_version} or change your ParallelCluster to ${!installed_version}"
                 fi
               else
                 bootstrap_instance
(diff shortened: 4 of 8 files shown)


cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -667,3 +667,27 @@ def test_is_hit_enabled_cluster(scheduler, expected_is_hit_enabled):
...
+@pytest.mark.parametrize(
+    "ami_name, error_expected",
+    [
+        # Compatible ami name
+        ("ami-xxx{0}xxx".format(current_version), False),
+        ("ami-123{0}321".format(current_version), False),
+        ("ami-{0}".format(current_version), False),
+        ("ami-name-one", False),
+        # Incompatible ami name
+        ("ami-aws-parallelcluster-0.0.0", True),
```
rexcsn, line 684: Can we add a test case for something that looks like the official AMIs, i.e. just starts with `aws-parallelcluster-`, something like `aws-parallelcluster-2.8.0-ubuntu-1804-*`.

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -667,3 +667,27 @@ def test_is_hit_enabled_cluster(scheduler, expected_is_hit_enabled):
...
+    ],
+)
+def test_validate_pcluster_version_based_on_ami_name(ami_name, error_expected):
+    if error_expected:
+        with pytest.raises(SystemExit) as sysexit:
+            utils.validate_pcluster_version_based_on_ami_name(ami_name)
+        assert_that(sysexit.value.code).is_not_equal_to(0)
```
rexcsn, line 691: See if you can match the error message instead of checking the error code here, with something like`pytest.raises(SystemExit, match=expected_message)`

cli/pcluster/utils.py
```diff
@@ -771,6 +771,17 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
+def validate_pcluster_version_based_on_ami_name(ami_name):
+    match = re.match(r"(.*)aws-parallelcluster-(\d+\.\d+\.\d+)(.*)", ami_name)
+    if match:
+        if ami_name.find("aws-parallelcluster-{0}".format(get_installed_version())) < 0:
+            error(
```
rexcsn, line 778: I am not sure if using error with `fail_on_error=True` is the behavior we want here. The bottom line is, in the case of cluster creation, we want to be able to turn this "name version check" off when `sanity_check = false`. Please double check to make sure this is the case.

Another way to handle this is we print out a warning instead of an error, so user is not forced to skip other sanity checks.

cli/tests/pcluster/createami/test_pcluster_createami.py
```diff
@@ -100,11 +107,11 @@ def test_validate_createami_args_architecture_compatibility(
     )
     if error_expected:
         with pytest.raises(SystemExit) as sysexit:
-            createami._validate_createami_args_architecture_compatibility(args)
+            createami._validate_createami_args_ami_compatibility(args)
         assert_that(sysexit.value.code).is_not_equal_to(0)
```
rexcsn, line 111: Similar comment about asserting on error messages if possible

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -362,12 +362,7 @@ Resources:
               export cookbook_version=${CookbookVersion}
               export chef_version=${ChefVersion}
               export berkshelf_version=${BerkshelfVersion}
-              if [ -f /opt/parallelcluster/.bootstrapped ]; then
-                installed_version=$(cat /opt/parallelcluster/.bootstrapped)
-                if [ "${!parallelcluster_version}" != "${!installed_version}" ]; then
-                  bootstrap_instance
-                fi
-              else
+              if [ ! -f /opt/parallelcluster/.bootstrapped ]; then
```
rexcsn, line 365: Not sure if we can remove the old code here. I think what would happen here now is: say user is using AMI baked with old version of pcluster, the AMI already has an old version of cookbook on it, without bootstrapping(download and vendor) new version of cookbook onto this old AMI again, we will not actually run new check you might include in the new cookbook version because we will be running the old cookbook on the old AMI.

cloudformation/compute-fleet-substack.cfn.yaml
```diff
@@ -453,12 +453,7 @@ Resources:
               export cookbook_version=${CookbookVersion}
               export chef_version=${ChefVersion}
               export berkshelf_version=${BerkshelfVersion}
-              if [ -f /opt/parallelcluster/.bootstrapped ]; then
-                installed_version=$(cat /opt/parallelcluster/.bootstrapped)
-                if [ "${!parallelcluster_version}" != "${!installed_version}" ]; then
-                  bootstrap_instance
-                fi
-              else
+              if [ ! -f /opt/parallelcluster/.bootstrapped ]; then
```
rexcsn, line 456: Same comment as above about not removing this

cloudformation/master-server-substack.cfn.yaml
```diff
@@ -431,12 +431,7 @@ Resources:
               export cookbook_version=${CookbookVersion}
               export chef_version=${ChefVersion}
               export berkshelf_version=${BerkshelfVersion}
-              if [ -f /opt/parallelcluster/.bootstrapped ]; then
-                installed_version=$(cat /opt/parallelcluster/.bootstrapped)
-                if [ "${!parallelcluster_version}" != "${!installed_version}" ]; then
-                  bootstrap_instance
-                fi
-              else
+              if [ ! -f /opt/parallelcluster/.bootstrapped ]; then
```
rexcsn, line 434: Same comment as above about not removing version check and bootstrap call here

cli/pcluster/utils.py
rexcsn: On a second thought I think the current behavior is ok. I was worried that user with very customized AMI names might be affected, but I think the regex is fairly safe and they really should not have an older version in their AMI name.

cli/pcluster/utils.py
```diff
@@ -771,6 +771,17 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
+def validate_pcluster_version_based_on_ami_name(ami_name):
+    match = re.match(r"(.*)aws-parallelcluster-(\d+\.\d+\.\d+)(.*)", ami_name)
+    if match:
+        if ami_name.find("aws-parallelcluster-{0}".format(get_installed_version())) < 0:
```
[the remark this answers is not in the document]
hanwen-pcluste: If the version string is followed by digits, only current approach works properly.
For example the AMI name is ami-aws-parallelcluster-2.8.1123.

cli/tests/pcluster/utils/test_pcluster_utils.py
hanwen-pcluste: done

cli/pcluster/utils.py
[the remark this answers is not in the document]
rexcsn: It's good that you thought about this, but I think this is not a case we need to consider. Moreover, along the same lines, the current implementation would actually cause an issue in case of using `aws-parallelcluster-2.8.1` AMI with `aws-parallelcluster-2.8.11`, if we actually had an 2.8.11 release.

I would like us to err on the side of caution and only do comparison on the extracted version.

cli/tests/pcluster/createami/test_pcluster_createami.py
rexcsn: Still relevant, please fix

cli/tests/pcluster/createami/test_pcluster_createami.py
```diff
@@ -100,11 +107,11 @@ def test_validate_createami_args_architecture_compatibility(
     )
     if error_expected:
         with pytest.raises(SystemExit) as sysexit:
-            createami._validate_createami_args_architecture_compatibility(args)
+            createami._validate_createami_args_ami_compatibility(args)
```
rexcsn, line 110: On a second look, we are not testing the newly added version related logic in `_validate_createami_args_ami_compatibility`. Please modify this test or add new test to cover newly added logic.

hanwen-pcluste: I am not sure if a test is needed here. Because all logic about validating the version by name is covered by a test in `test_pcluster_utils.py`. Moreover adding tests here will expand the parametrize dramatically. There is one option to add another test in this file. However, the test will be very similar to the test in `test_pcluster_utils.py`

cli/tests/pcluster/createami/test_pcluster_createami.py
hanwen-pcluste: The error is written to logger, do you know how to check it? I would suggest to keep the current code to maintain a consistent style throughout the file.

cli/tests/pcluster/createami/test_pcluster_createami.py
hanwen-pcluste: Added an `assert_called_with` to ensure the validation by name is called in the function

cli/tests/pcluster/createami/test_pcluster_createami.py
```diff
@@ -100,12 +104,13 @@ def test_validate_createami_args_architecture_compatibility(
...
         assert_that(sysexit.value.code).is_not_equal_to(0)
     else:
-        assert_that(createami._validate_createami_args_architecture_compatibility(args)).is_equal_to(
-            base_ami_architecture
+        assert_that(createami._validate_createami_args_ami_compatibility(args)).is_equal_to(
+            {"Architecture": base_ami_architecture, "Name": base_ami_name}
         )
+        createami.utils.validate_pcluster_version_based_on_ami_name.assert_called_with(base_ami_name)
```
rexcsn, line 113: Minor: I would prefer not to call the mocked function like this, because from this code alone we can't really tell if this function is mocked.

Alternatively, we can do the following:
```
version_validation_spy = mocker.patch("pcluster.createami.utils.validate_pcluster_version_based_on_ami_name")
...
version_validation_spy.assert_called_with(base_ami_name)
```

rexcsn (approved):

cli/tests/pcluster/createami/test_pcluster_createami.py
hanwen-pcluste: Yeah. This is the first approach that came to my mind. I wrote the current code to keep a consistent style with the code surrounding

cli/pcluster/utils.py
```diff
@@ -771,6 +771,19 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
+def validate_pcluster_version_based_on_ami_name(ami_name):
+    match = re.match(r"(.*)aws-parallelcluster-(\d+\.\d+\.\d+)(.*)", ami_name)
```
demartinofra, line 775: what happens if the customer provides an AMI that is not one of the ones we publish but has a similar pattern in the name? Is there a way to make this check more robust?

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -365,7 +365,7 @@ Resources:
               if [ -f /opt/parallelcluster/.bootstrapped ]; then
                 installed_version=$(cat /opt/parallelcluster/.bootstrapped)
                 if [ "${!parallelcluster_version}" != "${!installed_version}" ]; then
```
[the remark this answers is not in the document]
hanwen-pcluste: Currently `.bootstrapped` file use pcluster version instead of cookbook version. You are right, maybe I should include cookbook version inside `.bootstrapped` file?

PR #2058, opened 2020-09-18
Add integration test for cluster creation in private subnet

ddeidda:

### DIFF
--- a/tests/integration-tests/tests/networking/test_cluster_networking.py
+++ b/tests/integration-tests/tests/networking/test_cluster_networking.py
@@ -1,0 +1,26 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import pytest
+from assertpy import assert_that
+from utils import get_compute_nodes_instance_ids
+
+
+@pytest.mark.dimensions("us-west-2", "c5.xlarge", "alinux2", "slurm")
+@pytest.mark.dimensions("eu-west-2", "c5.xlarge", "centos7", "sge")
+@pytest.mark.usefixtures("os", "scheduler", "instance")
+def test_cluster_in_private_subnet(region, pcluster_config_reader, clusters_factory):
+    # This test just creates a cluster in the private subnet and just checks that no failures occur
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    assert_that(cluster).is_not_none()
+
+    assert_that(len(get_compute_nodes_instance_ids(cluster.cfn_name, region))).is_equal_to(1)
--- a/tests/integration-tests/tests/networking/test_cluster_networking/test_cluster_in_private_subnet/pcluster.config.ini
+++ b/tests/integration-tests/tests/networking/test_cluster_networking/test_cluster_in_private_subnet/pcluster.config.ini
@@ -1,0 +1,23 @@
+[global]
+cluster_template = default
+
+[aws]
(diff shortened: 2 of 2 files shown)


tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -0,0 +1,23 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import pytest
+from assertpy import assert_that
+
+
+@pytest.mark.dimensions("eu-west-1", "c4.xlarge", "alinux2", "slurm")
```
demartinofra, line 16: let's use c5.xlarge only. I think we use c4 only in cn regions.

demartinofra: I think eu-west-1 regions is heavily used already. Shall we pick a region with fewer tests? 

also we could select centos7 with one of the schedulers.

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -0,0 +1,23 @@
...
+@pytest.mark.dimensions("eu-west-2", "c5.xlarge", "alinux2", "sge")
+@pytest.mark.usefixtures("scheduler, os, region, test_datadir")
+def test_cluster_in_private_subnet(pcluster_config_reader, clusters_factory):
+    # This test just creates a cluster in the private subnet and just checks that no failures occur
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    assert_that(cluster).is_not_none()
```
demartinofra, line 23: optional:
we could add a couple of assertions:
1. check that cloudformation stack is in create complete
2. check that the number of compute nodes is equal to 1. There is already a function that queries EC2 to get the cluster nodes.

tests/integration-tests/tests/networking/test_cluster_networking/test_cluster_in_private_subnet/pcluster.config.ini
```diff
@@ -0,0 +1,23 @@
...
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = {{ scheduler }}
+master_instance_type = {{ instance }}
+compute_instance_type = {{ instance }}
+initial_queue_size = 1
+max_queue_size = 1
+maintain_initial_size = true
```
demartinofra, line 16: let's set this to false so that in case of slurm we have the bootstrap logic submit a job to the scheduler (this happens only when initial_size > min_size ==> maintain_initial_size = false)

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -0,0 +1,23 @@
...
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import pytest
+from assertpy import assert_that
+
+
+@pytest.mark.dimensions("eu-west-1", "c4.xlarge", "alinux2", "slurm")
+@pytest.mark.dimensions("eu-west-2", "c5.xlarge", "alinux2", "sge")
+@pytest.mark.usefixtures("scheduler, os, region, test_datadir")
```
demartinofra, line 18: why test_datadir in the usefixtures? can't it be removed?

you also need to add instance to the fixtures.

tests/integration-tests/tests/networking/test_cluster_networking.py
ddeidda: Shouldn't #1 be already tested from clusters_factpry?

tests/integration-tests/tests/networking/test_cluster_networking.py
ddeidda: Test_datadir is needed from pcluster_config_reader, isn't it?

demartinofra: yes but pcluster_config will get the fixture regardless of it

demartinofra (approved):

PR #2061, opened 2020-09-22
Provide OS type specified by a user to CFN templates of compute nodes

hanwen-pcluste:
This commit will enable validation of OSs provided by the user is consistent with the actual OS of a custom AMI.

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -2867,7 +2867,7 @@
           "EFA": {
             "Ref": "EFA"
           },
-          "OS": {
+          "BaseOS": {
             "Ref": "BaseOS"
           },
           "OSUser": {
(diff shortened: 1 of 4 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -3321,6 +3321,9 @@
           "EFA": {
             "Ref": "EFA"
           },
+          "OS": {
```
rexcsn, line 3324: Minor: If we want to be consistent with other parameters that are passed from main stack into substack we should name this `BaseOS`. If we want to try to differentiate this parameter I would pick a more descriptive name like `ComputeOS` so it's easier to find

hanwen-pcluste: I will make it as BaseOS if it is Ok with you.

rexcsn (approved):

PR #2063, opened 2020-09-22
Add FSx S3 AutoImport option

gkao123:
*Issue #, if available:*

*Description of changes:*
Allow S3-linked FSx filesystems to enable AutoImport.

AutoImport is a new FSx feature that allows customers to configure their Amazon FSx for Lustre file system to automatically import metadata of objects that are added to or changed in the linked S3 bucket after file system creation. 

For more details, please see https://docs.aws.amazon.com/fsx/latest/LustreGuide/autoimport-data-repo.html.

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -767,7 +767,7 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
 
     if urlparse(param_value).scheme == "s3":
         try:
-            bucket = param_value.split("/")[2]
+            bucket = getBucketNameFromImportPath(param_value)
             boto3.client("s3").head_bucket(Bucket=bucket)
         except ClientError:
             warnings.append(
@@ -781,6 +781,27 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
 
     return errors, warnings
 
+def fsx_lustre_auto_import_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if param_value != "NONE":
+        try:
+            fsx_section = pcluster_config.get_section("fsx")
+            fsx_import_path = fsx_section.get_param_value("import_path")
+            bucket = getBucketNameFromImportPath(fsx_import_path)
+            s3BucketRegion = boto3.client("s3").get_bucket_location(Bucket=bucket)["LocationConstraint"]
+            # Buckets in Region us-east-1 have a LocationConstraint of null
+            if (s3BucketRegion is None):
+                s3BucketRegion = "us-east-1"
+            if s3BucketRegion != pcluster_config.region:
+                errors.append("AutoImport is not supported for cross-region buckets.")
+        except ClientError:
+            warnings.append(
+                "The S3 bucket '{0}' does not exist or you do not have access to it. "
+                "Please be sure the cluster nodes have access to it.".format(fsx_import_path)
+            )
+    return errors, warnings
 
 def ec2_ebs_snapshot_validator(param_key, param_value, pcluster_config):
     errors = []
@@ -1291,3 +1312,6 @@ def ebs_volume_type_size_validator(section_key, section_label, pcluster_config):
             errors.append("The size of {0} volumes must be at least {1} GiB".format(volume_type, min_size))
 
     return errors, warnings
+
+def getBucketNameFromImportPath(importPath):
+    return importPath.split("/")[2] 
--- a/cli/tests/pcluster/config/test_section_fsx.py
+++ b/cli/tests/pcluster/config/test_section_fsx.py
@@ -21,21 +21,21 @@ from tests.pcluster.config.defaults import DefaultCfnParams, DefaultDict
         (DefaultCfnParams["fsx"].value, DefaultDict["fsx"].value),
         ({}, DefaultDict["fsx"].value),
         (
-            {"FSXOptions": "NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE, NONE"},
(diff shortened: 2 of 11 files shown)


tests/integration-tests/tests/storage/test_fsx_lustre/test_fsx_lustre/pcluster.config.ini
```diff
@@ -38,3 +38,4 @@ deployment_type = {{ deployment_type }}
 {% if per_unit_storage_throughput %}
 per_unit_storage_throughput = {{ per_unit_storage_throughput }}
 {% endif %}
+auto_import_policy = NEW_CHANGED
```
tilne, line 41: Do you think it's worth testing `NEW` and `NONE` as well? There are already three test cases, so it wouldn't cost much extra time to add a third parameter and pass it into this config, and then add logic to `_test_auto_import` that asserts different conditions based on which policy was used.

cloudformation/fsx-substack.cfn.json
```diff
@@ -270,7 +287,7 @@
       "Type": "String"
     },
     "FSXOptions": {
-      "Description": "Comma separated list of fsx related options, 13 parameters in total, [shared_dir,fsx_fs_id,storage_capacity,fsx_kms_key_id,imported_file_chunk_size,export_path,import_path,weekly_maintenance_start_time,deployment_type,per_unit_storage_throughput,daily_automatic_backup_start_time,automatic_backup_retention_days,copy_tags_to_backups]",
+      "Description": "Comma separated list of fsx related options, 15 parameters in total, [shared_dir,fsx_fs_id,storage_capacity,fsx_kms_key_id,imported_file_chunk_size,export_path,import_path,weekly_maintenance_start_time,deployment_type,per_unit_storage_throughput,daily_automatic_backup_start_time,automatic_backup_retention_days,copy_tags_to_backups,fsx_backup_id,auto_import_policy]",
```
tilne, line 290: thank you for adding the missing parameter. i missed that one on the last round of review for this template.

cli/tests/pcluster/config/test_section_fsx.py
```diff
@@ -21,21 +21,21 @@
...
         (
-            {"FSXOptions": "test,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE"},
+            {"FSXOptions": "test,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE"},
             utils.merge_dicts(DefaultDict["fsx"].value, {"shared_dir": "test"}),
         ),
         (
             {
                 "FSXOptions": "test,test1,10,test2,20,test3,test4,test5,SCRATCH_1,"
-                "50,01:00,5,false,backup-0a1b2c3d4e5f6a7b8"
+                "50,01:00,5,false,backup-0a1b2c3d4e5f6a7b8, NEW_CHANGED"
```
tilne, line 38: minor: removing this space would be nice

cli/pcluster/config/validators.py
```diff
@@ -1304,3 +1325,6 @@ def ebs_volume_type_size_validator(section_key, section_label, pcluster_config):
             errors.append("The size of {0} volumes must be at least {1} GiB".format(volume_type, min_size))
 
     return errors, warnings
+
+def getBucketNameFromImportPath(importPath):
```
tilne, line 1329: since i believe `import_path` is an S3 URL, I'd prefer this be renamed to `get_bucket_name_from_s3_url` to reflect the more generic usage of the function in `s3_bucket_validator`.

cli/pcluster/config/validators.py
```diff
@@ -782,6 +782,27 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
 
     return errors, warnings
 
+def fsx_lustre_auto_import_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if param_value != "NONE":
```
tilne, line 789: my understanding is that if a parameter isn't passed in here, `param_value` would be `None` and not `"NONE"`.

cli/pcluster/config/validators.py
```diff
@@ -782,6 +782,27 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+            fsx_section = pcluster_config.get_section("fsx")
+            fsx_import_path = fsx_section.get_param_value("import_path")
+            bucket = getBucketNameFromImportPath(fsx_import_path)
+            s3BucketRegion = boto3.client("s3").get_bucket_location(Bucket=bucket)["LocationConstraint"]
+            # Buckets in Region us-east-1 have a LocationConstraint of null
+            if (s3BucketRegion is None):
+                s3BucketRegion = "us-east-1"
+            if s3BucketRegion != pcluster_config.region:
+                errors.append("AutoImport is not supported for cross-region buckets.")
+        except ClientError:
```
tilne, line 800: A couple of questions for how this exception is handled:
* Is there a set of a ClientError's attributes that could be checked to be more certain about whether the bucket doesn't exist or the user doesn't have access?
* warnings are used when we want to bring an issue to the user's attention, but we don't expect the issue to cause cluster creation failures. this seems like something that should be added to errors, because I would expect file system creation to fail.

cli/pcluster/config/validators.py
gkao123: This check here ensures that if a customer provides an AutoImport Policy of NONE that the s3 cross-region bucket check doesn't occur. Should I also add a "None" check here so the validation doesn't occur too in this case?

tilne: Yeah, I'd add the check for `None`: `if param_value and param_value != "NONE":`.

PR #2065, opened 2020-09-23
Handle 'optimal' when validating CCE error message parsing

tilne:
In order to make sure the supported instance types and families parsed
from the Batch's CCE are sane, they are compared to the list of all
known instance types and families returned by EC2's
DescribeInstanceTypeOfferings API.

This check was not accounting for the fact that Batch users can specify
an instance type/family of `optimal`. This change fixes that.

### DIFF
--- a/cli/pcluster/utils.py
+++ b/cli/pcluster/utils.py
@@ -335,7 +335,13 @@ def _parse_supported_instance_types_and_families_from_cce_emsg(emsg):
     """
     match = re.search(r"be\s+one\s+of\s*\[(.*[0-9a-z.\-]+.*,.*)\]", emsg)
     if match:
-        return [instance_type_token.strip() for instance_type_token in match.group(1).split(",")]
+        parsed_values = [instance_type_token.strip() for instance_type_token in match.group(1).split(",")]
+        LOGGER.debug(
+            "Parsed the following instance types and families from Batch CCE error message: {0}".format(
+                " ".join(parsed_values)
+            )
+        )
+        return parsed_values
     else:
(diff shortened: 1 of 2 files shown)


cli/pcluster/utils.py
```diff
@@ -359,9 +359,14 @@ def _get_instance_families_from_types(instance_types):
     return list(families)
 
 
-def _instance_types_and_families_are_supported(candidate_types_and_families, known_types_and_families):
-    """Return a boolean describing whether the given instance types and families are known to be supported."""
-    unknowns = [candidate for candidate in candidate_types_and_families if candidate not in known_types_and_families]
+def _batch_instance_types_and_families_are_supported(candidate_types_and_families, known_types_and_families):
+    """Return a boolean describing whether the instance types and families parsed from Batch API are known."""
+    known_exceptions = ["optimal"]
```
demartinofra, line 364: optimal can't be specified in a comma separated list but it needs to be the only value provided right?

tilne: that wasn't a constraint i was aware of. is that enforced by the Batch API?

i'll verify whether or not that's the case now.

tilne: I was able to successfully create a cluster using `compute_instance_type = optimal,m3.xlarge`, so at least the compute environment can be created using those multiple values. I seem to be able to submit jobs as well.

tilne:
Amended the commit to add a log message with the instance types/families that were actually parsed from the CCE error message.

demartinofra (approved):

PR #2086, opened 2020-09-28
Update FSx section to include AutoImport

gkao123:

*Description of changes:*

Allow S3-linked FSx filesystems to enable AutoImport.

AutoImport is a new FSx feature that allows customers to configure their Amazon FSx for Lustre file system to automatically import metadata of objects that are added to or changed in the linked S3 bucket after file system creation.

For more details, please see https://docs.aws.amazon.com/fsx/latest/LustreGuide/autoimport-data-repo.html.


cli/pcluster/config/validators.py
```diff
@@ -783,6 +761,29 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
     return errors, warnings
 
 
+def fsx_lustre_auto_import_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if param_value != "NONE":
```
tilne, line 768: Can you add a check for `None` here as well? `if param_value and param_value != "NONE"`

Strictly speaking, it shouldn't be necessary. The `auto_import_policy` parameter's entry in the `FSX` dict in mappings.py doesn't specify a `type`, so it defaults to the section's type's (`CfnSection`) default parameter type, `CfnParam`. This class uses the base `Param` class's validate implementation, which does check for `None` values.

There's a lot that could be changed there, which is the motivation for being more defensive in these validators with respect to uninitialized values.

cli/pcluster/config/validators.py
```diff
@@ -783,6 +761,29 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+            fsx_section = pcluster_config.get_section("fsx")
+            fsx_import_path = fsx_section.get_param_value("import_path")
+            bucket = get_bucket_name_from_s3_url(fsx_import_path)
+            s3BucketRegion = boto3.client("s3").get_bucket_location(Bucket=bucket)["ResponseMetadata"]["LocationConstraint"]
+            # Buckets in Region us-east-1 have a LocationConstraint of null
+            if s3BucketRegion is None:
+                s3BucketRegion = "us-east-1"
+            if s3BucketRegion != pcluster_config.region:
+                errors.append("AutoImport is not supported for cross-region buckets.")
+        except ClientError:
```
tilne, line 779: A couple of questions for how this exception is handled:

Is there a set of a ClientError's attributes that could be checked to be more certain about whether the bucket doesn't exist or the user doesn't have access?
warnings are used when we want to bring an issue to the user's attention, but we don't expect the issue to cause cluster creation failures. this seems like something that should be added to errors, because I would expect file system creation to fail.

cli/pcluster/config/validators.py
```diff
@@ -783,6 +772,35 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+    warnings = []
+
+    if param_value is not None and param_value != "NONE":
+        try:
+            fsx_section = pcluster_config.get_section("fsx")
+            fsx_import_path = fsx_section.get_param_value("import_path")
+            bucket = get_bucket_name_from_s3_url(fsx_import_path)
+            s3_bucket_region = boto3.client("s3").get_bucket_location(Bucket=bucket)["ResponseMetadata"][
+                "LocationConstraint"
+            ]
```
tilne, line 786: I think this will break for buckets in us-east-1. Use `.get` to avoid that:  `s3_bucket_region = boto3.client("s3").get_bucket_location(Bucket=bucket).get("ResponseMetadata", {}).get("LocationConstraint")`

tilne: Now that I look at the docs for `get_bucket_location`, it looks like "LocationConstraint" is a top-level field in the response dict, so this should be: `s3_bucket_region = boto3.client("s3").get_bucket_location(Bucket=bucket).get("LocationConstraint")`

cli/tests/pcluster/config/test_validators.py
```diff
@@ -404,6 +404,56 @@ def test_url_validator(mocker, boto3_stubber, capsys):
...
+                    "import_path": "s3://test/test1/test2",
+                    "export_path": "s3://test/test1/test2",
+                    "auto_import_policy": "NEW",
+                },
+            },
+            2,
+            {"Bucket": "test"},
+            "AutoImport is not supported for cross-region buckets.",
+        ),
+    ],
```
tilne, line 424: It would be nice to have test cases for the explicitly handled exception cases (e.g., the error code is "NoSuchBucket" or "AccessDenied").

You can force the boto3 stubber to generate client errors by adding a `generate_error` and `error_code` kwarg to the MockedBoto3Request constructor:
```python
MockedBoto3Request(
    method="get_bucket_location",
    response=get_bucket_location_response,
    expected_params=bucket,
    generate_error=error_code is not None,
    error_code=error_code,
)
```

You'll probably need to add an error_code parameter to your test cases for this as well.

cli/pcluster/config/validators.py
```diff
@@ -783,6 +772,35 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+            # Buckets in Region us-east-1 have a LocationConstraint of null
+            if s3_bucket_region is None:
+                s3_bucket_region = "us-east-1"
+            if s3_bucket_region != pcluster_config.region:
+                errors.append("AutoImport is not supported for cross-region buckets.")
+        except ClientError as client_error:
+            if client_error.response.get("Error").get("Code").is_equal_to("NoSuchBucket"):
```
tilne, line 793: use `== <VALUE>` instead of `is_equal_to(<VALUE>)`

cli/pcluster/config/validators.py
```diff
@@ -768,13 +755,15 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
-            bucket = param_value.split("/")[2]
+            bucket = get_bucket_name_from_s3_url(param_value)
             boto3.client("s3").head_bucket(Bucket=bucket)
-        except ClientError:
-            warnings.append(
-                "The S3 bucket '{0}' does not exist or you do not have access to it. "
-                "Please be sure the cluster nodes have access to it.".format(param_value)
-            )
+        except ClientError as client_error:
+            if client_error.response.get("Error").get("Code").is_equal_to("NoSuchBucket"):
```
tilne, line 761: Same thing here: `==` instead of `is_equal_to`

tilne (approved):

PR #2088, opened 2020-09-28
Allow specifying EBS volume size when using snapshot

chenwany:
Signed-off-by: chenwany <wanyichen0521@gmail.com>

*Issue #, if available:*

*Description of changes:*


cli/tests/pcluster/config/test_validators.py
```diff
@@ -2127,3 +2094,49 @@ def test_ebs_allowed_values_all_have_volume_size_bounds():
...
+        "Progress": "100%",
+        "OwnerId": "012345678910",
+        "SnapshotId": ebs_snapshot_id,
+    }
+    mocker.patch("pcluster.config.cfn_param_types.get_ebs_snapshots_response", return_value=describe_snapshots_response)
+    mocker.patch("pcluster.config.validators.get_ebs_snapshots_response", return_value=describe_snapshots_response)
+
+    config_parser_dict = {"cluster default": {"ebs_settings": "default"}, "ebs default": section_dict}
+    utils.assert_param_validator(
+        mocker, config_parser_dict, capsys=capsys, expected_warning=expected_warning, expected_error=expected_error
```
tilne, line 2141: minor: ideally kwargs are passed to the function in the same order that they appear in the function definition. that would make this line `mocker, config_parser_dict, expected_error=expected_error, capsys=capsys, expected_warning=expected_warning`

chenwany: Good suggestion for me! Done

cloudformation/ebs-substack.cfn.json
```diff
@@ -1141,24 +1141,16 @@
...
-                    }
-                  ]
-                },
-                20
+                  "Ref": "VolumeSize"
+                }
               ]
-            }
+            },
+            20
```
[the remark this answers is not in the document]
chenwany: I don't know whether these 20 are needed or not. Since they are here before my change, and there were default value for EBS volumes at that time. I think the CLI shouldn't ever not pass in a value for volume size before my change either. Maybe put it here is just to make the code safer?

cli/pcluster/config/mappings.py
```diff
@@ -292,7 +293,7 @@
     "key": "ebs",
     "default_label": "default",
     "max_resources": 5,
-    "validators": [ebs_volume_type_size_validator, ebs_volume_iops_validator],
+    "validators": [ebs_volume_type_size_validator, ebs_volume_iops_validator, ebs_volume_size_snapshot_validator],
```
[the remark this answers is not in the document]
chenwany: No. I didn't add test cases for to test when a snapshot id is used. Should I add those? That also require to mocker.patch the snapshot response there

cli/pcluster/utils.py
```diff
@@ -1112,3 +1112,26 @@ def get_bucket_url(region):
 def get_file_section_name(section_key, section_label=None):
     """Build a section name as in the config file, given section key and label."""
     return section_key + (" {0}".format(section_label) if section_label else "")
+
+
+def get_ebs_snapshots_response(ebs_snapshot_id):
```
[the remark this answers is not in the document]
chenwany: Right! We only use one snapshot here. Done

cli/pcluster/utils.py
```diff
@@ -1112,3 +1112,26 @@ def get_bucket_url(region):
...
+        "StartTime": "2014-02-28T21:28:32.000Z",
+        "Progress": "100%",
+        "OwnerId": "012345678910",
+        "SnapshotId": "snap-1234567890abcdef0",
+    }
+    """
+    try:
+        return boto3.client("ec2").describe_snapshots(SnapshotIds=[ebs_snapshot_id]).get("Snapshots")[0]
+    except ClientError as e:
+        error(e.response.get("Error").get("Message"))
```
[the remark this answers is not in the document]
chenwany: Good idea! Done!

cli/pcluster/config/mappings.py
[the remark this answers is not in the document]
tilne: i think we should handle it in the validators. for an ebs section like this:
```ini
[ebs custom]
shared_dir = /ebs_snapshot_dir
snapshot_id = X
volume_iops = Y
volume_type = Z
```

the IOPS and volume type values are both being passed to and used in the EBS substack's CFN template.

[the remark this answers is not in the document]
tilne: Actually after looking into this more, I don't think any changes are needed for the validators to work as expected.
* the first one make sure that the volume type and size are compatible. volume size is set by the time validation starts, and the volume type doesn't depend on whether or not a snapshot is used.
* the second one makes sure the volume type and IOPS values are compatible. these values are not dependent on whether a snapshot is used.

cli/pcluster/config/validators.py
```diff
@@ -1332,3 +1320,33 @@ def ebs_volume_iops_validator(section_key, section_label, pcluster_config):
...
+
+    The EBS snapshot is in "completed" state if it is specified
+    If user specified the volume size, the volume must be larger than then volume size of the EBS snapshot
+    """
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section(section_key, section_label)
+    if section.get_param_value("ebs_snapshot_id"):
+        ebs_snapshot_id = section.get_param_value("ebs_snapshot_id")
```
tilne, line 1337: Since we're covering with this validator the check for whether or not the snapshot exists, it would be nice to catch an exception for it here and handle that case explicitly:

```python
try:
    ebs_snapshot_id = section.get_param_value("ebs_snapshot_id")
    snapshot_response_dict = get_ebs_snapshot_info(ebs_snapshot_id, raise_exceptions=True)
except Exception as exception:
    if isinstance(exception, ClientError) and exception.response.get("Error").get("Code") in ["InvalidSnapshot.NotFound", "InvalidSnapshot.Malformed"]:
        pass  # append to errors some message about the snapshot not existing
    else:
        pass  # append to errors some generic message about being unable to get info for the snapshot
```

Note that this also requires implementing the optional `raise_exceptions` arg in `get_ebs_snapshot_info`.

cli/pcluster/utils.py
```diff
@@ -1112,3 +1112,30 @@ def get_bucket_url(region):
...
+    }
+    """
+    try:
+        return boto3.client("ec2").describe_snapshots(SnapshotIds=[ebs_snapshot_id]).get("Snapshots")[0]
+    except ClientError as e:
+        error(
+            "Failed when calling DescribeSnapshot for {0}: {1}".format(
+                ebs_snapshot_id, e.response.get("Error").get("Message")
+            )
+        )
```
tilne, line 1141: let's add tests for this function. they can be added to the files in cli/tests/pcluster/utils/test_pcluster_utils.py

chenwany: Yes! It is done

cli/pcluster/config/validators.py
chenwany: Good idea. Done

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -961,3 +961,42 @@ def test_get_instance_families_from_types(caplog, instance_types, error_expected
...
+            expected_params={"SnapshotIds": ["snap-1234567890abcdef0"]},
+            generate_error=error_message is not None,
+        )
+    ]
+    boto3_stubber("ec2", mocked_requests)
+    if error_message is None:
+        assert_that(utils.get_ebs_snapshot_info(snapshot_id)).is_equal_to(response)
+    else:
+        with pytest.raises(SystemExit, match=error_message) as sysexit:
+            utils.get_ebs_snapshot_info(snapshot_id)
```
tilne, line 1001: Add test a test cases that sets the `raise_exceptions=True` when calling this function. You can do this by adding another parameter to go along with `snapshot_id` and `error_message`.

cli/tests/pcluster/config/test_param_types.py
```diff
@@ -8,12 +8,21 @@
...
+from pcluster.config.cfn_param_types import CfnParam, CfnSection, VolumeSizeParam
+from pcluster.config.mappings import EBS
 from pcluster.config.param_types import Param
 
 
+@pytest.fixture()
+def boto3_stubber_path():
+    return "pcluster.utils.boto3"
+
+
```
tilne, line 25: You can get rid of this now that you're calling the function from utils instead of a boto3 client.

cli/tests/pcluster/config/test_param_types.py
```diff
@@ -40,3 +49,56 @@ def test_validate_section_label(self, section_label, should_trigger_error, mocke
...
+        section_label="default",
+        param_key="volume_size",
+        param_definition=EBS.get("params").get("volume_size"),
+        pcluster_config=mocked_pcluster_config,
+        owner_section=ebs_section,
+    ).from_file(config_parser)
+
+    if mocked_pcluster_config.get_section("ebs", "default").get_param_value(
+        "volume_size"
+    ) is None and mocked_pcluster_config.get_section("ebs", "default").get_param_value("ebs_snapshot_id"):
```
tilne, line 88: Unlike boto3_stubber, it won't cause errors if you call mocker.patch for a function that never gets called. For that reason, I would get rid of this `if` and always call mocker.patch.

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,42 @@ def validate_backup_options(
...
+                    "Snapshot {0} is in state '{1}' not 'completed'".format(
+                        ebs_snapshot_id, snapshot_response_dict.get("State")
+                    )
+                )
+        except Exception as exception:
+            if isinstance(exception, ClientError) and exception.response.get("Error").get("Code") in [
+                "InvalidSnapshot.NotFound",
+                "InvalidSnapshot.Malformed",
+            ]:
+                errors.append("The snapshot {0} does not exist".format(ebs_snapshot_id))
```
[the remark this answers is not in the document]
chenwany: Very good suggestion! Done

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,42 @@ def validate_backup_options(
...
+            else:
+                errors.append("Unable to get information for the snapshot {0}".format(ebs_snapshot_id))
```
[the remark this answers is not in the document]
chenwany: Done

cli/tests/pcluster/config/test_param_types.py
chenwany: Right! Done!

cli/tests/pcluster/config/test_param_types.py
chenwany: Right! Done

cli/tests/pcluster/utils/test_pcluster_utils.py
chenwany: Thanks! Done

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,58 @@ def validate_backup_options(
...
+                warnings.append(
+                    "Snapshot {0} is in state '{1}' not 'completed'".format(
+                        ebs_snapshot_id, snapshot_response_dict.get("State")
+                    )
+                )
+            if volume_size > snapshot_volume_size:
+                warnings.append(
+                    "In order to use the full capacity of the volume, you'll need to manually resize the partition "
+                    "according to this doc: "
+                    "https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/recognize-expanded-volume-linux.html"
```
tilne, line 1447: * Let's mention something about why the user might need to manually resize the partition (which is because the volume size is larger than the snapshot's size).
* Did we come to a conclusion on whether or not we should include the URL to the user guide?

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,58 @@ def validate_backup_options(
...
+    errors = []
+    warnings = []
+
+    section = pcluster_config.get_section(section_key, section_label)
+    if section.get_param_value("ebs_snapshot_id"):
+        try:
+            ebs_snapshot_id = section.get_param_value("ebs_snapshot_id")
+            snapshot_response_dict = get_ebs_snapshot_info(ebs_snapshot_id, raise_exceptions=True)
+            # validate that the input volume size is larger than the volume size of the EBS snapshot
+            snapshot_volume_size = snapshot_response_dict.get("VolumeSize")
```
tilne, line 1432: What happens if `snapshot_response_dict` doesn't have a `VolumeSize` attribute? I'd add a default value and handle that case explicitly.

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,58 @@ def validate_backup_options(
...
+            if volume_size < snapshot_volume_size:
+                errors.append("The EBS volume size must not be smaller than the volume size of EBS snapshot.")
+            # validate that the state of ebs snapshot
+            if snapshot_response_dict.get("State") != "completed":
+                warnings.append(
+                    "Snapshot {0} is in state '{1}' not 'completed'".format(
+                        ebs_snapshot_id, snapshot_response_dict.get("State")
+                    )
+                )
+            if volume_size > snapshot_volume_size:
```
tilne, line 1443: minor: i would move this check underneath the check for `volume_size < snapshot_volume_size`, and I would use an if-elif for it. The idea is to show that the checks are related (and both of them can't be true at the same time).

```python
if volume_size < snapshot_volume_size:
  # append error
elif volume_size > snapshot_volume_size:
  # append warning
```

cli/pcluster/config/validators.py
```diff
@@ -1423,3 +1411,58 @@ def validate_backup_options(
...
+                    )
+                )
+            else:
+                errors.append(
+                    "The snapshot {0} does not appear to exist: {1}".format(
+                        ebs_snapshot_id,
+                        exception.response.get("Error").get("Message")
+                        if isinstance(ClientError, exception)
+                        else exception,
+                    )
```
tilne, line 1466: Let's add test cases for the different error cases handled here.

cli/pcluster/config/validators.py
chenwany: Added

cli/pcluster/config/validators.py
chenwany: Done!

cli/pcluster/config/validators.py
chenwany: Added

tilne (approved):

PR #2094, opened 2020-09-30
integ-tests: test subnets selection in pcluster configure

hanwen-pcluste:
test the ability to avoid subnets without desired instance type


tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -59,6 +61,61 @@ def test_pcluster_configure(
     clusters_factory(config_path)
 
 
+@pytest.mark.regions(["us-east-1"])
```
yuleiwan, line 64: I'd suggest no choosing us-east-1 or eu-west-1 as test region due to integration test region balance, unless you need these regions for particular purpose.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -74,9 +131,26 @@ def get_unsupported_test_runner_options(request):
...
+    omitted_subnets_num=0,
 ):
     compute_units = "vcpus" if scheduler == "awsbatch" else "instances"
+    # Due to the order of creation,PublicSubnet is the first in the list of subnets. Therefore it is the default value.
+    vpc_stack_public_subnet_id = vpc_stack.cfn_outputs["PublicSubnetId"]
+    if omitted_subnets_num:
+        # When there are omitted subnets, a note should be printed
+        omitted_note = "Note:  {0} subnet\\(s\\) is/are not listed.+".format(omitted_subnets_num)
+    else:
+        omitted_note = ""
```
yuleiwan, line 153: could simplify the expression like 
`omitted_note = "Note:  {0} subnet\\(s\\) is/are not listed.+".format(omitted_subnets_num) if omitted_subnets_num else ""`

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -59,6 +61,61 @@ def test_pcluster_configure(
     clusters_factory(config_path)
 
 
+@pytest.mark.regions(["us-east-1"])
+@pytest.mark.instances(["c5.xlarge", "m6g.xlarge"])
+@pytest.mark.schedulers(["awsbatch", "slurm", "sge"])
+# Do not run on ARM + Batch
+# pcluster configure always picks optimal and Batch does not support ARM for optimal for now
+@pytest.mark.skip_dimensions("*", "m6g.xlarge", "*", "awsbatch")
```
rexcsn, line 69: We probably only need to run this test once with a particular os/scheduler/instance combination, because what we are testing do not depend on these 3 dimensions

tests/integration-tests/tests/configure/test_pcluster_configure.py
rexcsn: I think logic below might be specific to us-east-1, and I think we should only run this test on 1 dimension combination.
It would be ok to keep things as is, but general it is good to avoid "popular" regions in order to balance the tests. Or even better, make the logic generic so we could run on other regions as well.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +238,14 @@ def assert_config_contains_expected_values(
             validator.get("parameter_name")
         )
         assert_that(observed_value).is_equal_to(validator.get("expected_value"))
+
+
+@pytest.fixture()
+def subnet_in_use1_az3(vpc_stack):
+    # use1_az3 does not have c5.xlarge or m6g.xlarge
```
rexcsn, line 245: Can we do an assertion somewhere to assert this initial condition? It is possible that c5.xlarge will get added to use1_az3 in the future and we won't be aware of the issue

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -74,9 +131,26 @@ def get_unsupported_test_runner_options(request):
...
+    region,
+    key_name,
+    scheduler,
+    os,
+    instance,
+    vpc_id,
+    public_subnet_id,
+    private_subnet_id,
+    config_path,
+    vpc_stack,
```
rexcsn, line 143: `vpc_stack` is only used to get `vpc_stack_public_subnet_id`, but we are passing in `public_subnet_id` already, can we just use that?

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +238,14 @@ def assert_config_contains_expected_values(
...
+
+@pytest.fixture()
+def subnet_in_use1_az3(vpc_stack):
+    # use1_az3 does not have c5.xlarge or m6g.xlarge
+    ec2_client = boto3.client("ec2", region_name="us-east-1")
+    subnet_id = ec2_client.create_subnet(
+        AvailabilityZoneId="use1-az3", CidrBlock="192.168.16.0/20", VpcId=vpc_stack.cfn_outputs["VpcId"]
+    )["Subnet"]["SubnetId"]
+    yield subnet_id
+    ec2_client.delete_subnet(SubnetId=subnet_id)
```
rexcsn, line 251: When does this subnet get deleted? After each individual test finishes or after test _runner finishes?

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +238,14 @@ def assert_config_contains_expected_values(
             validator.get("parameter_name")
         )
         assert_that(observed_value).is_equal_to(validator.get("expected_value"))
+
+
+@pytest.fixture()
+def subnet_in_use1_az3(vpc_stack):
```
rexcsn, line 244: Let's only pass in `vpc_stack.cfn_outputs["VpcId"]` instead of `vpc_stack`

tests/integration-tests/tests/configure/test_pcluster_configure.py
hanwen-pcluste: When using fixture, we can use `@pytest.fixture(scope=???)` to specify the scope. The default is the subnet is deleted after each test. I will change it to `@pytest.fixture(scope="module")`

tests/integration-tests/tests/configure/test_pcluster_configure.py
rexcsn: I see. Then let's pass in `vpc_stack.cfn_outputs["PublicSubnetId"]` as a parameter and not pass in the entire `vpc_stack`.

Minor: Alternatively we can name things like `headnode_subnet_response` to make clear that the subnet_id is only used for response.

Please avoid using `master` in any new code that we write

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +230,31 @@ def assert_config_contains_expected_values(
...
+    )
+    offerings = []
+    for page in page_iterator:
+        offerings.extend(page["InstanceTypeOfferings"])
+    if offerings:
+        logging.fatal(
+            "c5.xlarge or m6g.xlarge has been added to use1-az3, change the test code to test other rare instance type"
+        )
+    subnet_id = ec2_client.create_subnet(
+        AvailabilityZoneId="use1-az3", CidrBlock="192.168.16.0/20", VpcId=vpc_stacks["us-east-1"].cfn_outputs["VpcId"]
```
rexcsn, line 257: Again, let's try to only pass in exactly what we need, in this case, `vpc_stacks["us-east-1"].cfn_outputs["VpcId"]`, but not the entire `vpc_stacks`
Please double check your code for this case

tests/integration-tests/tests/configure/test_pcluster_configure.py
rexcsn: Great, was just curious, but I think `module` makes sense here so the same subnet can be used by other tests in this testing file if needed

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -59,6 +61,54 @@ def test_pcluster_configure(
...
+        region,
+        key_name,
+        scheduler,
+        os,
+        instance,
+        vpc_stack.cfn_outputs["VpcId"],
+        # master/compute_subnet_ids are not provided.
+        # Therefore, pcluster configure uses the subnet specified by config file.
+        # However, the subnet in the config file does not contain the desired instance type.
+        # Eventually, pcluster configure uses the first subnet in the list of subnets
```
rexcsn, line 92: Please avoid using the word "master" in any newly added code/comment. Please double check rest of your code for this issue.

Minor: comment here is not that clear, can reword to something like
```
# Do not provide master/compute_subnet_ids inputs.
# Normally, pcluster configure should use the subnet specified in config file by default.
# However, for this test, the subnet in the config does not contain the desired instance type and will be omitted
# As a result, pcluster configure eventually uses the first subnet in the remaining list of subnets
```

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +191,73 @@ def assert_config_contains_expected_values(
...
+            "prompt": fr"{omitted_note}Master Subnet ID \[{vpc_stack_public_subnet_id}\]: ",
+            "response": headnode_subnet_id,
+        },
+        {
+            "prompt": fr"{omitted_note}Compute Subnet ID \[{vpc_stack_public_subnet_id}\]: ",
+            "response": compute_subnet_id,
+        },
+    ]
+    # When a user selects Batch as the scheduler, pcluster configure does not prompt for OS or compute instance type.
+    return [stage for stage in stage_list if scheduler != "awsbatch" or stage.get("skip_for_batch") is None]
```
rexcsn, line 235: Can change `stage.get("skip_for_batch") is None` to `not stage.get("skip_for_batch")`, because `skip_for_batch` could be False
```
return [stage for stage in stage_list if scheduler != "awsbatch" or not stage.get("skip_for_batch")]
```

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +191,73 @@ def assert_config_contains_expected_values(
...
+        Filters=[
+            {"Name": "instance-type", "Values": ["c5.xlarge", "m6g.xlarge"]},
+            {"Name": "location", "Values": ["use1-az3"]},
+        ],
+    )
+    offerings = []
+    for page in page_iterator:
+        offerings.extend(page["InstanceTypeOfferings"])
+    if offerings:
+        logging.fatal(
```
rexcsn, line 256: `logging.fatal` here does not actually stop or fail the test right?
Let's use a normal assertion and logging message to 1.) keep consistent with rest of testing framework, 2.) make sure this test fails if assertion fails.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +191,73 @@ def assert_config_contains_expected_values(
...
+            "c5.xlarge or m6g.xlarge has been added to use1-az3, change the test code to test other rare instance type"
```
rexcsn, line 257: Comment is not clear on what our goal is; shall we try changing to something like:
```
"c5.xlarge or m6g.xlarge has been added to use1-az3, change the test code so that we are testing the case that a specific instance type is not available in a chosen subnet"
```

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +191,73 @@ def assert_config_contains_expected_values(
...
+            "prompt": fr"{omitted_note}Compute Subnet ID \[{vpc_stack_public_subnet_id}\]: ",
+            "response": compute_subnet_id,
+        },
+    ]
+    # When a user selects Batch as the scheduler, pcluster configure does not prompt for OS or compute instance type.
+    return [stage for stage in stage_list if scheduler != "awsbatch" or stage.get("skip_for_batch") is None]
+
+
+@pytest.fixture()
+def subnet_in_use1_az3(vpc_stack):
```
rexcsn, line 239: Minor: hard coded implementation is ok for now, but we should try to make the function name and description more generic and add some To-do items.
So that if we choose to make this test generic or run it in another region in the future we can easily refactor the code.
For example we could do something like:
```
@pytest.fixture()
def subnet_without_instance_type(vpc_stack):
    # Subnet used to test the functionality of avoiding invalid subnet in AZs that do not have user specified instance types
    # Hard coding implementation to run on us-east-1 and create subnet in AZ without c5.xlarge or m6g.xlarge
    # Use use1_az3 as AZ, which does not have c5.xlarge or m6g.xlarge
    # To-do: extend code to support arbitrary region and arbitrary instance types
```

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +192,77 @@ def assert_config_contains_expected_values(
...
+        },
+    ]
+    # When a user selects Batch as the scheduler, pcluster configure does not prompt for OS or compute instance type.
+    return [stage for stage in stage_list if scheduler != "awsbatch" or not stage.get("skip_for_batch")]
+
+
+@pytest.fixture()
+def subnet_in_use1_az3(vpc_stack):
+    # Subnet used to test the functionality of avoiding subnet in AZs that do not have user specified instance types
+    # Hard coding implementation to run on us-east-1 and create subnet in use1_az3 without c5.xlarge or m6g.xlarge
```
rexcsn, line 242: Minor: remove `m6g.xlarge` from comment since it's not used anymore.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -161,3 +192,77 @@ def assert_config_contains_expected_values(
...
+            {"Name": "instance-type", "Values": ["c5.xlarge"]},
+            {"Name": "location", "Values": ["use1-az3"]},
+        ],
+    )
+    offerings = []
+    for page in page_iterator:
+        offerings.extend(page["InstanceTypeOfferings"])
+    logging.info(
+        "Asserting that c5.xlarge is not in use1-az3. If assertion fails, "
+        "c5.xlarge has been added to use1-az3, change the test code to test other rare instance type"
```
rexcsn, line 261: Minor: I still think this comment is not clear. Especially don't like "change the test code to test other rare instance type".
If we want to provide information, we should make clear the exact case we are testing here. But I don't think we should suggest something like `test other rare instance type` because we could also change AZ for example. In any case, we should let the future maintainers decide path forward and not mislead them into thinking we need to run this test on some expensive or rare instance type

rexcsn (dismissed):

rexcsn (approved):

PR #2097, opened 2020-09-30
Fix queue validation for enable_efa parameter

lukeseawalker:
* Fixed validator in validator.py that was comparing the same value for enable_efa taken from compute_resource

This change highlighted that there were two validators for enable_efa, one in validator.py and the other in json_param_types.py, producing the following outputs

```
WARNING: EFA was enabled on queue 'c5', but instance type 'c5.large' defined in compute resource settings c5 does not support EFA.
WARNING: The section [queue c5] is wrongly configured
EFA was enabled on queue 'c5', but instance type 'c5.large' defined in compute resource settings c5 does not support EFA.
```

* Removed the enable_efa validator in json_param_types.py

* Added unit test for an instance type efa enabled

*Issue #, if available:*

*Description of changes:*


enrico-usai (approved):

cli/tests/pcluster/config/test_json_param_types.py
```diff
@@ -94,26 +94,16 @@ def boto3_stubber_path():
...
-            "WARNING: EFA was enabled on queue 'queue1', but instance type 'm6g.xlarge' does not support EFA.\n",
-        ),
-        ([], ""),
-        (["queue3"], ""),
+        (["queue1"]),
+        (["queue2"]),
+        (["queue1", "queue2"]),
+        ([]),
+        (["queue3"]),
     ],
```
enrico-usai, line 104: Given the fact we moved these validation steps into the `test_queue_validator` probably it's now useless to have so many tests for test_config_to_json that are not triggering any validation path.

lukeseawalker: There is still the validation "Check that created json params match the expected ones"

PR #2110, opened 2020-10-06
Allow pcluster update if tags are specified but not changed

hanwen-pcluste:
This is a bug fix. The bug was that no matter if the tags were changed or not, pcluster update would report incompatible update as long as tags were specified


cli/pcluster/config/validators.py
```diff
@@ -1129,6 +1129,22 @@ def base_os_validator(param_key, param_value, pcluster_config):
     return [], warnings
 
 
+def tags_validator(param_key, param_value, pcluster_config):
+    warnings = []
+
+    for key in param_value.keys():
+        if key == "Version":
+            warnings.append(
+                'You are using "Version" as a key in your tags. Because "Version" tag is reserved to '
```
enrico-usai, line 1138: I'd set it as an error.

Then I'd simplify the message with something like:
`"The key 'Version' used in your 'tags' configuration parameter is a reserved one, please change it."`

cli/tests/pcluster/config/test_validators.py
```diff
@@ -185,6 +185,24 @@ def test_ec2_ami_validator(mocker, boto3_stubber, image_architecture, bad_ami_me
     utils.assert_param_validator(mocker, config_parser_dict, expected_message)
 
 
+@pytest.mark.parametrize(
+    "section_dict, expected_message",
+    [
+        ({"tags": {"key": "value", "key2": "value2"}}, None),
+        (
+            {"tags": {"key": "value", "Version": "value2"}},
+            'You are using "Version" as a key in your tags. Because "Version" tag is reserved to '
```
enrico-usai, line 194: In the test is not needed to use the entire message, you can use a substring. It simplify the work if we need to do minor changes in the message.

We just need to be sure that is failing.

enrico-usai (commented):
Great job! Almost ready to go!

enrico-usai (dismissed):
LGTM!

enrico-usai (approved):

PR #2113, opened 2020-10-08
Include user tags in HIT substack launch templates

tilne:
CFN does not currently support tag propagation to launch templates. See
https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/184
for details.

Work around this by including the tags when rendering the template.

### DIFF
--- a/cloudformation/compute-fleet-hit-substack.cfn.yaml
+++ b/cloudformation/compute-fleet-hit-substack.cfn.yaml
@@ -157,6 +157,28 @@ Resources:
                 Value: !Ref 'FilesystemTagValue'
               - Key: aws-parallelcluster-node-type
                 Value: Compute
+    {%- for tag in tags %}
+              - Key: {{ tag.Key }}
+                Value: {{ tag.Value }}
+    {%- endfor %}
+          - ResourceType: volume
+            Tags:
+              - Key: ClusterName
+                Value: !Select
+                  - '1'
+                  - !Split
+                    - parallelcluster-
+                    - !Ref 'MainStackName'
+              - Key: QueueName
+                Value: {{ queue }}
+              - Key: Application
+                Value: !Ref 'MainStackName'
+              - Key: aws-parallelcluster-node-type
+                Value: Compute
+    {%- for tag in tags %}
+              - Key: {{ tag.Key }}
+                Value: {{ tag.Value }}
+    {%- endfor %}
         BlockDeviceMappings:
           - DeviceName: /dev/xvdba
             VirtualName: ephemeral0
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -33,6 +33,8 @@ class Cluster:
         self.create_complete = False
         self.__cfn_outputs = None
         self.__cfn_resources = None
+        self.__head_node_substack_cfn_resources = None
+        self.__ebs_substack_cfn_resources = None
 
     def __repr__(self):
         attrs = ", ".join(["{key}={value}".format(key=key, value=repr(value)) for key, value in self.__dict__.items()])
@@ -63,8 +65,7 @@ class Cluster:
         logging.info("Cluster {0} updated successfully".format(self.name))
 
         # reset cached properties
-        self.__cfn_outputs = None
-        self.__cfn_resources = None
+        self._reset_cached_properties()
 
         return result
 
@@ -127,15 +128,19 @@ class Cluster:
             )
             raise
 
-    def instances(self):
+    def instances(self, desired_instance_role=None):
         """Run pcluster stop and return the result."""
+        if desired_instance_role and desired_instance_role not in ("MasterServer", "ComputeFleet"):
+            raise ValueError
         cmd_args = ["pcluster", "instances", "--config", self.config_file, self.name]
         try:
             result = run_command(cmd_args, log_error=False)
             logging.info("Get cluster {0} instances successfully".format(self.name))
             cluster_instances = []
             for entry in result.stdout.splitlines():
-                cluster_instances.append(entry.split()[1])
+                instance_role, instance_id = entry.split()
+                if not desired_instance_role or desired_instance_role == instance_role:
+                    cluster_instances.append(instance_id)
             return cluster_instances
         except subprocess.CalledProcessError as e:
             logging.error(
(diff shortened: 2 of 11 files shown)


cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -157,6 +157,10 @@ Resources:
                 Value: !Ref 'FilesystemTagValue'
               - Key: aws-parallelcluster-node-type
                 Value: Compute
+    {%- for tag in tags %}
```
demartinofra, line 160: While addressing this shall we also duplicate the Tags array and assign it to a ResourceType of type volume so that hopefully we also get EBS volumes tagged correctly?

demartinofra: Can you include tags in the input provided in test_hit_substack_rendering (`cloudformation/tests/test_cfn_template_rendering.py`)?

demartinofra (commented):
As you suggested let's add a quick integ test for this where we spin up a cluster with min size set to 1 and we check both tags on compute and head node (also on EBS volumes if we add those). The same test can then be extended to verify the fix for the tag update functionality.

cloudformation/compute-fleet-hit-substack.cfn.yaml
tilne: I believe both of these have been addressed.

tilne:
> As you suggested let's add a quick integ test for this where we spin up a cluster with min size set to 1 and we check both tags on compute and head node (also on EBS volumes if we add those). The same test can then be extended to verify the fix for the tag update functionality.

The integ test is a WIP. It's written but needs to be tested. I should be able to get it done tomorrow.

tests/integration-tests/tests/tags/test_tag_propagation.py
```diff
@@ -0,0 +1,154 @@
...
+import logging
+
+import boto3
+import pytest
+from assertpy import assert_that
+
+from pcluster.utils import get_installed_version
+
+
+@pytest.mark.regions(["us-east-1"])
```
tilne, line 24: any guidance on what region to put new tests in?

demartinofra: ap-southeast-1 or 2

tests/integration-tests/tests/tags/test_tag_propagation.py
```diff
@@ -0,0 +1,154 @@
...
+        {"resource": "Compute Node", "tag_getter": get_compute_node_tags, "skip": scheduler == "awsbatch"},
+        {"resource": "EBS Volume", "tag_getter": get_shared_volume_tags},
+    ]
+    for test_case in test_cases:
+        if test_case.get("skip"):
+            continue
+        logging.info("Verifying tags were propagated to %s", test_case.get("resource"))
+        tag_getter = test_case.get("tag_getter")
+        # Assume tag getters use lone cluster object arg if none explicitly given
+        tag_getter_args = test_case.get("tag_getter_args", (cluster,))
```
demartinofra, line 66: minor: wondering if we can use named args just so it's harder to mess up with the order

tests/integration-tests/tests/tags/test_tag_propagation.py
```diff
@@ -0,0 +1,154 @@
...
+        {"resource": "Main CloudFormation Stack", "tag_getter": get_main_stack_tags},
+        {"resource": "Head Node CloudFormation Stack", "tag_getter": get_head_node_substack_tags},
+        {
+            "resource": "ComputeFleet CloudFormation Stack",
+            "tag_getter": get_compute_fleet_substack_tags,
+            "tag_getter_args": (cluster, scheduler),
+        },
+        {"resource": "Head Node", "tag_getter": get_head_node_tags},
+        {"resource": "Compute Node", "tag_getter": get_compute_node_tags, "skip": scheduler == "awsbatch"},
+        {"resource": "EBS Volume", "tag_getter": get_shared_volume_tags},
```
demartinofra, line 58: I think the change you made to tag the EBS volume applies to the root volume of instances launched with that specific template.

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -135,7 +135,8 @@ Resources:
         Monitoring:
           Enabled: false
         TagSpecifications:
-          - ResourceType: instance
+    {%- for resource_type in ["instance", "volume"] %}
```
demartinofra, line 138: shall we do this also for the head node and the SIT fleet?

tests/integration-tests/tests/tags/test_tag_propagation.py
demartinofra: In any case it's good to check also the tags on shared EBS volumes

tests/integration-tests/clusters_factory.py
```diff
@@ -199,6 +205,42 @@ def cfn_resources(self):
...
+        Resources are retrieved only once and then cached.
+        """
+        if not self.__head_node_substack_cfn_resources:
+            self.__head_node_substack_cfn_resources = retrieve_cfn_resources(
+                self.cfn_resources.get("MasterServerSubstack"), self.region
+            )
+        return self.__head_node_substack_cfn_resources
+
+    @property
+    def compute_fleet_substack_cfn_resources(self):
```
demartinofra, line 221: nit: this only works with hit substack? shall we make it explicit in the function name or handle the sit case?

tests/integration-tests/clusters_factory.py
```diff
@@ -199,6 +205,42 @@ def cfn_resources(self):
...
+            )
+        return self.__compute_fleet_substack_cfn_resources
+
+    @property
+    def ebs_substack_cfn_resources(self):
+        """
+        Return the CloudFormation stack resources for the cluster's EBS substack.
+        Resources are retrieved only once and then cached.
+        """
+        if not self.__compute_fleet_substack_cfn_resources:
```
demartinofra, line 238: you forgot to rename `__compute_fleet_substack_cfn_resources`

tests/integration-tests/tests/tags/test_tag_propagation.py
tilne: > I think the change you made to tag the EBS volume applies to the root volume of instances launched with that specific template.

i thought this was the case as well, but it doesn't appear to be tagged given the changes made thus far. do you know what else i would need to change to propagate the tags to the root volume?

tests/integration-tests/clusters_factory.py
tilne: oof

tests/integration-tests/clusters_factory.py
tilne: i ended up not using this, so i'm going to get rid of it

tests/integration-tests/tests/tags/test_tag_propagation.py
tilne: i turned them into kwargs. i'm not sure if that's what you were looking for here though?

demartinofra (approved):
👍🏻

PR #2114, opened 2020-10-08
integ-tests: test cfn_scheduler_slots in pcluster update

yuleiwan:
* Add an integration test in test_update to check value of slots is correct after compute instance type updating.
* Fix an unit test on window os


tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
 def _test_min_vcpus(region, stack_name, vcpus):
     asg_min_size = get_batch_ce_min_size(stack_name, region)
     assert_that(asg_min_size).is_equal_to(vcpus)
+
+
+@pytest.mark.dimensions("us-west-1", "c5.xlarge", "centos7", "sge")
+@pytest.mark.usefixtures("os", "instance")
+def test_update_sit_update_compute_instance(
```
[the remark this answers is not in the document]
demartinofra: These test is very expensive in terms of time cause it needs to create 5 different clusters and also perform an update on all of them. 
On one side we should not aggregate tests that run on independent clusters in the same test function so that we better leverage the test parallelism. On the other side can we drop some of the combinations? Personally I'd test only one combination with disable_hyperthreading=true and one with cfn_scheduler_slots=cores so that we reduce the tests from 5 to 2.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+    # Create cluster with initial configuration
+    init_config_file = pcluster_config_reader(config_file=init_config_name)
+    cluster = clusters_factory(init_config_file)
+
+    # Read init configuration
+    init_config = configparser.ConfigParser()
+    init_config.read(init_config_file)
+
+    # Update cluster with new configuration and change the compute instance type to c5.9xlarge
+    new_compute_instance_type = "c5.9xlarge"
```
demartinofra, line 595: why such a big instance and not something like c5.large or at most c5.2xlarge?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+    cluster.update()
+
+
+def _save_config_file(config, test_datadir, file_name):
+    try:
+        with open(str(test_datadir / file_name), "w") as configfile:
+            config.write(configfile)
+        return file_name
+    except IOError as err:
+        logging.info("I/O error: {0}".format(err))
```
demartinofra, line 665: are we just swallowing the error? I'd just let this exception propagate to communicate the failure. probably you don't even need to catch it and log the info cause it'll be logged as an error automatically

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+    _update_config_compute_instance_type(
+        init_config, pcluster_config_reader, test_datadir, cluster, new_compute_instance_type, init_config_name
+    )
+
+    _check_cfn_scheduler_slots(
+        cluster, scheduler, region, new_compute_instance_type, cfn_scheduler_slots=cfn_scheduler_slots
+    )
+
+
+def _update_config_compute_instance_type(
```
demartinofra, line 650: optional: this is a personal preference but I'd rather have two files in test_update_sit rather than updating the config dynamically with a configparser.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+def _save_config_file(config, test_datadir, file_name):
+    try:
+        with open(str(test_datadir / file_name), "w") as configfile:
+            config.write(configfile)
+        return file_name
+    except IOError as err:
+        logging.info("I/O error: {0}".format(err))
+
+
+def _check_cfn_scheduler_slots(
```
demartinofra, line 668: nit: _check_compute_node_slots?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,159 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+    if disable_hyperthreading or cfn_scheduler_slots == "cores":
+        slots_per_instance = slots_per_instance // 2
+    elif cfn_scheduler_slots and cfn_scheduler_slots.isdigit():
+        slots_per_instance = int(cfn_scheduler_slots)
+
+    # submit a job to check cfn_scheduler_slot has been updated
+    result = scheduler_commands.submit_command("sleep 1", nodes=1, slots=slots_per_instance)
+    job_id = scheduler_commands.assert_job_submitted(result.stdout)
+    scheduler_commands.wait_job_completed(job_id)
+    assert_that(scheduler_commands.compute_nodes_count()).is_equal_to(1)
```
demartinofra, line 686: can we also assert the actual number of slots seen by the scheduler?

yuleiwan: what do you mean actual number of slots seen by the scheduler

tests/integration-tests/configs/develop.yaml
```diff
@@ -509,3 +509,9 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["centos7"]
           schedulers: ["sge"]
+    test_update.py::test_update_sit_update_compute_instance:
+      dimensions:
+        - regions: ["us-west-1"]
```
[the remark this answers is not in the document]
yuleiwan: us-west-1 has 22 tests, other lower tests commercial region could be eu-north-1(19), ap-northeast-2(19), what do you think?

tests/integration-tests/tests/update/test_update.py
yuleiwan: Done

tests/integration-tests/configs/develop.yaml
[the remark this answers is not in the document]
demartinofra: ok I thought we had more tests in us-west-1. In that case I'm fine with keeping it as is

tests/integration-tests/tests/update/test_update.py
demartinofra: no I mean running the qhost command and check how many slots are configured in the scheduler for the specific node

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,96 @@ def _test_max_vcpus(region, stack_name, vcpus):
 def _test_min_vcpus(region, stack_name, vcpus):
     asg_min_size = get_batch_ce_min_size(stack_name, region)
     assert_that(asg_min_size).is_equal_to(vcpus)
+
+
+@pytest.mark.dimensions("us-west-1", "c5.xlarge", "centos7", "sge")
+@pytest.mark.usefixtures("os", "instance")
+def test_sit_update_compute_instance(region, scheduler, pcluster_config_reader, clusters_factory, s3_bucket_factory):
```
demartinofra, line 536: can we have 2 different test functions or parametrize this one as done in lustre tests so that the two tests are ran in parallel whenever possible?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -529,3 +529,96 @@ def _test_max_vcpus(region, stack_name, vcpus):
...
+    )
+    cluster.stop()
+
+    # Update cluster with new configuration and change the compute instance type to c5.2xlarge
+    updated_config_file = pcluster_config_reader(update_config_name)
+    cluster.config_file = str(updated_config_file)
+    cluster.update()
+
+    # Check cfn_scheduler_slots changed by changing compute instance type
+    _check_compute_node_slots(
```
demartinofra, line 592: I'm wondering how this is working if you don't start the cluster.

tests/integration-tests/tests/update/test_update/test_sit_update_compute_instance/pcluster.config.disable_ht.ini
```diff
@@ -0,0 +1,25 @@
...
+scheduler = {{ scheduler }}
+master_instance_type = {{ instance }}
+compute_instance_type = {{ instance }}
+initial_queue_size = 0
+max_queue_size = 2
+vpc_settings = parallelcluster-vpc
+disable_hyperthreading = true
+
+
+[vpc parallelcluster-vpc]
```
demartinofra, line 20: can you set use_public_ips to false so that we don't need to use one elastic ip for each test?

tests/integration-tests/tests/update/test_update/test_sit_update_compute_instance/pcluster.config.disable_ht.ini
```diff
@@ -0,0 +1,25 @@
+[global]
+cluster_template = default
+
+[aws]
+aws_region_name = {{ region }}
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+
```
demartinofra, line 10: nit: why the new lines?

tests/integration-tests/tests/update/test_update.py
yuleiwan: Done

tests/integration-tests/tests/update/test_update.py
yuleiwan: Done

tests/integration-tests/tests/update/test_update.py
yuleiwan: After update, the scheduler will use autoscaling to initialize the compute nodes, we needn't start the cluster. I tested it and passed. Correct me if I am wrong.

tests/integration-tests/tests/update/test_update/test_sit_update_compute_instance/pcluster.config.disable_ht.ini
yuleiwan: Done

tests/integration-tests/tests/update/test_update/test_sit_update_compute_instance/pcluster.config.disable_ht.ini
yuleiwan: Removed

demartinofra (approved):

PR #2115, opened 2020-10-08
Add support to HDD storage type for FSx section

chenwany:

*Description of changes:*
Allow to create new FSx file systems that utilize the new HDD feature as part of the cluster creation process. 

FSx for Lustre offers HDD-backed filesystems, a new feature that allows to create cost-optimized FSx for Lustre filesystems for customers with throughput-intensive workloads.

For more details, please see https://aws.amazon.com/blogs/aws/new-high-performance-hdd-storage-for-amazon-fsx-for-lustre-file-systems/


cli/pcluster/config/mappings.py
```diff
@@ -141,9 +141,11 @@
     "volume_types": ["standard", "io1", "gp2", "st1", "sc1"],
     "vpc_id": r"^vpc-[0-9a-z]{8}$|^vpc-[0-9a-z]{17}$",
     "deployment_type": ["SCRATCH_1", "SCRATCH_2", "PERSISTENT_1"],
-    "per_unit_storage_throughput": [50, 100, 200],
+    "per_unit_storage_throughput": [12, 40, 50, 100, 200],
```
enrico-usai, line 144: I'd define two keys:
```python
"fsx_ssd_throughput": [50, 100, 200],
"fsx_hdd_throughput": [12, 40],
```

In the `per_unit_storage_throughput` param you can use:
```python
"allowed_values": ALLOWED_VALUES["fsx_ssd_throughput"] + ALLOWED_VALUES["fsx_hdd_throughput"],
```

and then you can also reuse these values in the validators:
```python
if fsx_per_unit_storage_throughput not in ALLOWED_VALUES["fsx_hdd_throughput"]:
    errors.append("For HDD filesystems, per_unit_storage_throughput can only have the following values: {0}".format(ALLOWED_VALUES["fsx_hdd_throughput"]:))
```

This help us to have [dry](https://realpython.com/lessons/dry-principle/) code and have these numbers here and there in the code.

enrico-usai (commented):
Great job!

As a general suggestions it is good to add a more detailed description in both the commit message and the PR description.
E.g. you can list the changes you're doing and add links to the official doc, if any.

I'd also add the FSx word in the title and in the commit message, it facilitates research for "fsx" related PRs.

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1430,19 @@ def validate_backup_options(
...
+def validate_storage_type_options(
+    errors, fsx_storage_type, fsx_deployment_type, fsx_per_unit_storage_throughput, fsx_drive_cache_type
+):
+    if fsx_storage_type == "HDD":
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, deployment_type must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in [12, 40]:
+            errors.append("For HDD filesystems, per_unit_storage_throughput can only have values of 12 or 40")
+        if not fsx_drive_cache_type:
+            errors.append("For HDD filesystems, the 'fsx_drive_cache_type' must be specified")
```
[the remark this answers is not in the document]
chenwany: Fixed

cli/pcluster/config/mappings.py
chenwany: Done with adding it in the mappings!
But for reusing these values in the validators, when I import ALLOWED_VALUE in mappings.py into validator.py, that cause a lots of unit tests in other test files failed.

cli/pcluster/config/validators.py
```diff
@@ -1393,6 +1399,10 @@ def validate_persistent_options(errors, fsx_deployment_type, fsx_kms_key_id, fsx
         if fsx_per_unit_storage_throughput:
             errors.append("'per_unit_storage_throughput' can only be used when 'deployment_type = PERSISTENT_1'")
 
+    if fsx_deployment_type == "PERSISTENT_1":
```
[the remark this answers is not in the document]
chenwany: Yeah! Done

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1430,19 @@ def validate_backup_options(
...
+    if fsx_storage_type == "HDD":
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, deployment_type must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in [12, 40]:
+            errors.append("For HDD filesystems, per_unit_storage_throughput can only have values of 12 or 40")
+        if not fsx_drive_cache_type:
+            errors.append("For HDD filesystems, the 'fsx_drive_cache_type' must be specified")
+
+    if fsx_storage_type == "SSD":
+        if fsx_per_unit_storage_throughput not in [None, 50, 100, 200]:
```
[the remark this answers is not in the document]
chenwany: Good idea! Done!

cli/pcluster/config/mappings.py
```diff
@@ -141,9 +141,11 @@
...
     "architectures": SUPPORTED_ARCHITECTURES,
-    "auto_import_policy": ["NONE", "NEW", "NEW_CHANGED"]
+    "auto_import_policy": ["NONE", "NEW", "NEW_CHANGED"],
```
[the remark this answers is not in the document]
chenwany: Done!

cloudformation/fsx-substack.cfn.json
```diff
@@ -254,6 +254,36 @@
...
+            "NONE"
+          ]
+        }
+      ]
+    },
+    "UseDriveCacheType": {
+      "Fn::Equals": [
+        {
+          "Fn::Select": [
+            "15",
```
enrico-usai, line 279: 15 --> should be 16

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -200,6 +216,19 @@ def get_fsx_fs_id(cluster, region):
...
 
 
+def get_storage_type(fsx_fs_id, region):
+    logging.info("Getting StorageType from DescribeFilesystem API.")
+    fsx = boto3.client("fsx", region_name=region)
+    return fsx.describe_file_systems(FileSystemIds=[fsx_fs_id]).get("FileSystems")[0].get("StorageType")
+
+
+def _test_correct_storage_type(storage_type, fsx_fs_id, region):
+    if storage_type == "HDD":
```
enrico-usai, line 226: you can rewrite this with: 
```python
assert_that(get_storage_type(fsx_fs_id, region)).is_equal_to(storage_type)
```

enrico-usai: and you can remote it to _test_storage_type, to be consistent with other _test functions.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -167,7 +183,7 @@ def _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os, r
     result = remote_command_executor.run_remote_command("df -h -t lustre | tail -n +2 | awk '{print $1, $2, $6}'")
     mount_name = get_mount_name(fsx_fs_id, region)
     assert_that(result.stdout).matches(
```
enrico-usai, line 185: Could you add here an example of the output we're expecting/parsing?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -33,7 +33,16 @@
...
 @pytest.mark.parametrize(
-    "deployment_type, per_unit_storage_throughput", [("PERSISTENT_1", 200), ("SCRATCH_1", None), ("SCRATCH_2", None)]
+    "deployment_type, per_unit_storage_throughput, storage_type, drive_cache_type, storage_capacity",
+    [
+        ("PERSISTENT_1", 200, None, None, 1200),
+        ("SCRATCH_1", None, None, None, 1200),
+        ("SCRATCH_2", None, None, None, 1200),
+        ("PERSISTENT_1", 200, "SSD", None, 1200),
+        ("PERSISTENT_1", 40, "HDD", "NONE", 1800),
+        ("PERSISTENT_1", 40, "HDD", "READ", 1800),
```
enrico-usai, line 43: I think we can avoid this test, it's very similar to the following one.

cli/pcluster/config/validators.py
```diff
@@ -286,7 +292,8 @@ def fsx_storage_capacity_validator(section_key, section_label, pcluster_config):
     fsx_section = pcluster_config.get_section(section_key, section_label)
     storage_capacity = fsx_section.get_param_value("storage_capacity")
     deployment_type = fsx_section.get_param_value("deployment_type")
-
+    storage_type = fsx_section.get_param_value("storage_type")
+    per_unit_storage_throughput = fsx_section.get_param_value("per_unit_storage_throughput")
```
enrico-usai, line 296: E.g. first check can be separate by the other if/else:
```python
    if fsx_section.get_param_value("fsx_fs_id") or fsx_section.get_param_value("fsx_backup_id"):
       # comment
       return errors, warnings
```

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1430,22 @@ def validate_backup_options(
...
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, deployment_type must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in [12, 40]:
+            errors.append("For HDD filesystems, per_unit_storage_throughput can only have values of 12 or 40")
+        if not fsx_drive_cache_type:
+            errors.append("For HDD filesystems, the 'drive_cache_type' must be specified")
+    if fsx_storage_type != "HDD" and fsx_drive_cache_type is not None:
+        errors.append("'drive_cache_type' features can be used only with HDD filesystems")
+    if not fsx_drive_cache_type and fsx_drive_cache_type is not None:
+        errors.append("'drive_cache_type' features can be used only with HDD filesystems")
```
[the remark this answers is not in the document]
chenwany: This condition is to avoid customer have an input for fsx_drive_cache_type = "NONE" without specifying the "Storage_type"

if not None: True
if None is not None: false

tests/integration-tests/tests/storage/test_fsx_lustre.py
chenwany: I am not sure. 
For per_unit_storage_throughput is 40, the storage_capacity should be the multiple of 1800
For per_unit_storage_throughput is 12, the storage_capacity should be the multiple of 6000
Should we delete one test case?

tests/integration-tests/tests/storage/test_fsx_lustre.py
chenwany: I don't know whether we can get storage_type from the remote, the df -h commands doesn't provide that info. I search for ways to do that and didn't found it. So instead I just make the boto3 call to check the storage type

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1430,22 @@ def validate_backup_options(
...
+    errors, fsx_storage_type, fsx_deployment_type, fsx_per_unit_storage_throughput, fsx_drive_cache_type
+):
+    if fsx_storage_type == "HDD":
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, deployment_type must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in [12, 40]:
+            errors.append("For HDD filesystems, per_unit_storage_throughput can only have values of 12 or 40")
+        if not fsx_drive_cache_type:
+            errors.append("For HDD filesystems, the 'drive_cache_type' must be specified")
+    if fsx_storage_type != "HDD" and fsx_drive_cache_type is not None:
```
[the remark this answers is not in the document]
chenwany: Yes! It is done!

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1430,22 @@ def validate_backup_options(
         fsx_imported_file_chunk_size or fsx_import_path or fsx_export_path or fsx_auto_import_policy
     ) and fsx_automatic_backup_retention_days:
         errors.append("Backups cannot be created on S3-linked file systems")
+
+
+def validate_storage_type_options(
+    errors, fsx_storage_type, fsx_deployment_type, fsx_per_unit_storage_throughput, fsx_drive_cache_type
+):
+    if fsx_storage_type == "HDD":
```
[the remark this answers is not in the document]
chenwany: Done!

cli/pcluster/config/validators.py
chenwany: Done!

tests/integration-tests/tests/storage/test_fsx_lustre.py
enrico-usai: Sorry I meant rename, not remote :D 
> you can rename it to _test_storage_type, to be consistent with other _test functions.

tests/integration-tests/tests/storage/test_fsx_lustre.py
enrico-usai: In our tests we should verify that the features are working correctly within ParallelCluster but we don't need to test all the combination of the FSx parameters, it is up to the FSx team to verify that the all the combinations work.

In the unit tests it is good to have all the combinations because we're testing our validators but for integration tests we can remove the `("PERSISTENT_1", 40, "HDD", "READ", 1800)` and just keep the `("PERSISTENT_1", 12, "HDD", "READ", 6000),` because we're already testing the 40 case in the pervious test.

tests/integration-tests/tests/storage/test_fsx_lustre.py
chenwany: Done! :D

tests/integration-tests/tests/storage/test_fsx_lustre.py
chenwany: Oh, I see! Done!

cloudformation/fsx-substack.cfn.json
chenwany: Fixed this logic 👌

cli/tests/pcluster/config/test_section_fsx.py
```diff
@@ -35,7 +35,7 @@
         (
             {
                 "FSXOptions": "test,test1,10,test2,20,test3,test4,test5,SCRATCH_1,"
-                "50,01:00,5,false,backup-0a1b2c3d4e5f6a7b8,NEW_CHANGED"
+                "50,01:00,5,false,backup-0a1b2c3d4e5f6a7b8,NEW_CHANGED,SSD,READ"
```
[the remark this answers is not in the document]
chenwany: Added!

tests/integration-tests/tests/storage/test_fsx_lustre.py
chenwany: Added

cli/pcluster/config/validators.py
```diff
@@ -1420,3 +1433,22 @@ def validate_backup_options(
...
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, 'deployment_type' must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in [12, 40]:
+            errors.append("For HDD filesystems, 'per_unit_storage_throughput' can only have values of 12 or 40")
+        if not fsx_drive_cache_type:
+            errors.append("For HDD filesystems, the 'drive_cache_type' must be specified")
+    elif fsx_storage_type != "HDD" and fsx_drive_cache_type:
+        errors.append("'drive_cache_type' features can be used only with HDD filesystems")
+    elif not fsx_drive_cache_type and fsx_drive_cache_type:
```
enrico-usai, line 1450: This can never happen, you're testing the opposite conditions:
`not fsx_drive_cache_type and fsx_drive_cache_type` --> `not A and A` is not possible :D

Probbaly you meant: `elif not fsx_storage_type and fsx_drive_cache_type:`

In any case this test is useless because the previous check `fsx_storage_type != "HDD"` already includes the `fsx_storage_type = None`.
You can verify it by differentiating the error messages and adding unit tests to address the different cases.

cli/pcluster/config/mappings.py
enrico-usai: The problem is that in `mappings` file we're including the `validators`, so you cannot include the `mappings` in the `validators` one, it is a circulary dependency.

You can move the definition of these lists in the `constants.py` module, we already have, by defining `FSX_SSD_THROUGHPUT_ALLOWED_VALUES` and `FSX_HDD_THROUGHPUT_ALLOWED_VALUES` and reusing them in both the mappings and validators modules.

It is important to avoid to replicate this information because if in future there are changes we'll have a single place to update.

cli/pcluster/config/mappings.py
```diff
@@ -524,8 +529,20 @@
...
+                "allowed_values": ALLOWED_VALUES["fsx_auto_import_policy"],
+                "update_policy": UpdatePolicy.UNSUPPORTED
+            }),
+            ("storage_type", {
+                "allowed_values": ALLOWED_VALUES["fsx_storage_type"],
+                "update_policy": UpdatePolicy.UNSUPPORTED
+            }),
+            ("drive_cache_type", {
+                "type": NoneCfnParam,
+                "NONE_value": "NONEDEFAULT",
```
enrico-usai, line 544: let's rename it to `"cfn_default" :  "NONEDEFAULT",`

cli/pcluster/config/mappings.py
```diff
@@ -457,6 +461,7 @@
     "type": CfnSection,
     "key": "fsx",
     "default_label": "default",
+    "dont_convert_to_NONE": True,
```
enrico-usai, line 464: Let's remove this one, it's very specific to the NONE case, instead we can generalize the CFN default setting approach as explained before.

cli/pcluster/config/cfn_param_types.py
```diff
@@ -1019,7 +1019,7 @@ def from_storage(self, storage_params):
                 except IndexError:
                     # This happen if the expected comma separated CFN param doesn't exist in the Stack,
                     # so it is set to a single NONE value
-                    cfn_value = "NONE"
+                    cfn_value = param_definition.get("NONE_value", "NONE")
```
enrico-usai, line 1022: let's use: `param_definition.get("cfn_default", "NONE")`, see explanation below.

cli/pcluster/config/cfn_param_types.py
```diff
@@ -1061,7 +1061,7 @@ def to_storage(self, storage_params=None):
                     param = param_type(self.key, self.label, param_key, param_definition, self.pcluster_config)
                     cfn_items.append(param.get_cfn_value())
 
-            if cfn_items[0] == "NONE":
+            if cfn_items[0] == "NONE" and not self.definition.get("dont_convert_to_NONE"):
```
enrico-usai, line 1064: Let's remove this old approach.

Here we were checking if the first item was NONE, and if so we were setting all the items to NONE, because we wanted to change all the default values to NONE.

Example in the RAIDSection case, by looking at the mappings you can see that some params have a default value, so the RAIDOptions CFN param would be:
`<NONE,NONE,2,gp2,20,100,false,NONE>`
but we want to convert it to `<NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE>` so we were executing the following code, to reset all the values to NONE: 
```
if cfn_items[0] == "NONE"
     cfn_items = ["NONE"] * len(self.definition.get("params"))
``` 

I'd remove this approach. We can get rid of both the adaptation code before and the extra attribute `dont_convert_to_NONE`.
We can instead define a new `cfn_default` attribute.

We need to set `cfn_default` for all the options in the mappings that have a "default" value (e.g. int, boolean or string) but must be converted to NONE during the `to_storage` conversion.

I mean for example:
```python
            ("num_of_raid_volumes", {
                "default": 2,   # the param has a default value but we want to convert to NONE
                "cfn_default": "NONE",  # we can explicitly say that this must be converted to none when used for CFN
```

In this way we'll have a more generic solution that can work for all the old parameters that have a default (on which we can add `"cfn_default": "NONE",`) and for the new parameters for FSx HDD, on which we can define "cfn_default": "NONEDEFAULT",.


So we can remove the adapation code:
```
if cfn_items[0] == "NONE"
     cfn_items = ["NONE"] * len(self.definition.get("params"))
```
because by using the `definition.get("cfn_default", "NONE")` approach all the values will be the correct ones or NONE, without the need to convert all of them to NONE.

cli/pcluster/config/cfn_param_types.py
```diff
@@ -1169,3 +1169,24 @@ def to_storage(self, storage_params=None):
 
         super(ClusterCfnSection, self).to_storage(storage_params)
         return storage_params
+
+
+class NoneCfnParam(CfnParam):
```
enrico-usai, line 1174: We can remove this class, by using the `cfn_default` approach mentioned before.

cli/pcluster/config/cfn_param_types.py
enrico-usai: Note, after using the cfn_default approach for al lthe parameter you can remove the `cfn_items = ["NONE"] * len(self.definition.get("params"))` from `EFSCfnSection` too.

cli/pcluster/config/mappings.py
enrico-usai: We also need to fix all the parameter in the sections that have a "default" value but we want to convert to NONE when passed to CFN. 
In details the parameter on which we need to add `"cfn_default" : "NONE"` are: 
* EFS --> performance_mode, encrypted, throughput_mode
* RAID --> num_of_raid_volumes, volume_type, volume_size, volume_iops, encrypted
* DCV --> port, access_from

cli/pcluster/config/mappings.py
chenwany: Done!

cli/pcluster/config/validators.py
chenwany: You are right! Done :D

cli/tests/pcluster/config/test_section_fsx.py
```diff
@@ -21,21 +21,21 @@
         (DefaultCfnParams["fsx"].value, DefaultDict["fsx"].value),
         ({}, DefaultDict["fsx"].value),
         (
-            {"FSXOptions": "NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE"},
+            {"FSXOptions": "NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE," "NONE"},
```
enrico-usai, line 24: minor: extra quotes, probably due to autoformat

cli/tests/pcluster/config/test_section_fsx.py
```diff
@@ -21,21 +21,21 @@
...
             DefaultDict["fsx"].value,
         ),
         (
-            {"FSXOptions": "NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE"},
+            {"FSXOptions": "NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE," "NONE"},
```
enrico-usai, line 28: same here

cli/tests/pcluster/config/test_section_fsx.py
```diff
@@ -21,21 +21,21 @@
...
             DefaultDict["fsx"].value,
         ),
         (
-            {"FSXOptions": "test,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE"},
+            {"FSXOptions": "test,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE," "NONE"},
```
enrico-usai, line 32: same here

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -203,6 +218,19 @@ def get_fsx_fs_id(cluster, region):
     return utils.retrieve_cfn_outputs(fsx_stack, region).get("FileSystemId")
 
 
+def get_storage_type(fsx_fs_id, region):
+    logging.info("Getting StorageType from DescribeFilesystem API.")
+    fsx = boto3.client("fsx", region_name=region)
+    return fsx.describe_file_systems(FileSystemIds=[fsx_fs_id]).get("FileSystems")[0].get("StorageType")
+
+
+def _test_storage_type(storage_type, fsx_fs_id, region):
```
enrico-usai, line 227: Let's simplify the function with:
```python
def _test_storage_type(storage_type, fsx_fs_id, region):
    assert_that(get_storage_type(fsx_fs_id, region)).is_equal_to(storage_type)
```

cli/pcluster/config/validators.py
```diff
@@ -1484,3 +1499,27 @@ def ebs_volume_size_snapshot_validator(section_key, section_label, pcluster_conf
...
+    if fsx_storage_type == "HDD":
+        if fsx_deployment_type != "PERSISTENT_1":
+            errors.append("For HDD filesystems, 'deployment_type' must be 'PERSISTENT_1'")
+        if fsx_per_unit_storage_throughput not in FSX_HDD_THROUGHPUT:
+            errors.append(
+                "For HDD filesystems, 'per_unit_storage_throughput' can only have the following values: {0}".format(
+                    FSX_HDD_THROUGHPUT
+                )
+            )
+    elif fsx_storage_type != "HDD" and fsx_drive_cache_type is not "NONE":
```
enrico-usai, line 1516: I'd limit this `if/else` to the two cases HDD and SSD and avoid the `!= HDD` one.

In this way in the SSD case, if both the conditions are not satisfied, we can provide both the errors to the users:
* _'drive_cache_type' features can be used only with HDD filesystems_ and
* _For SSD filesystems, 'per_unit_storage_throughput' can only have the following values_

Instead by having a different `elif` branch we are showing only one of the two messages.

We can write:
```python
if fsx_storage_type == "HDD":
    ...
else:  # SSD
    if fsx_drive_cache_type is not "NONE":
       errors.append("'drive_cache_type' features can be used only with HDD filesystems")
    if fsx_per_unit_storage_throughput and fsx_per_unit_storage_throughput not in FSX_SSD_THROUGHPUT:
        errors.append(
            "For SSD filesystems, 'per_unit_storage_throughput' can only have the following values: {0}".format(
                FSX_SSD_THROUGHPUT
            )
        )
```

cli/pcluster/config/validators.py
```diff
@@ -1304,11 +1314,13 @@ def fsx_ignored_parameters_validator(section_key, section_label, pcluster_config
 
     # If fsx_fs_id is specified, all parameters besides shared_dir are ignored.
     relevant_when_using_existing_fsx = ["fsx_fs_id", "shared_dir"]
+    PARAM_WITH_DEFAULT = {"drive_cache_type": "NONE"}
     if fsx_section.get_param_value("fsx_fs_id") is not None:
         for fsx_param in fsx_section.params:
-            if fsx_param not in relevant_when_using_existing_fsx and fsx_section.get_param_value(fsx_param) is not None:
+            if fsx_param not in relevant_when_using_existing_fsx and PARAM_WITH_DEFAULT.get(
```
enrico-usai, line 1320: We don't need to define the `PARAM_WITH_DEFAULT`. We can replace: 
`PARAM_WITH_DEFAULT.get(fsx_param, None) != fsx_section.get_param_value(fsx_param)`
with:
`fsx_section.get_param(fsx_param).get_default_value() != fsx_section.get_param_value(fsx_param)`

enrico-usai (commented):
The latest minor comments, thanks!

enrico-usai (dismissed):
Great job!

PR #2118, opened 2020-10-09
Enable `-r/--region` command line args for `pcluster dcv connect` and `pcluster ssh`

tilne:

### DIFF
--- a/cli/pcluster/createami.py
+++ b/cli/pcluster/createami.py
@@ -306,10 +306,6 @@ def _validate_createami_args_ami_compatibility(args):
 def create_ami(args):
     LOGGER.info("Building AWS ParallelCluster AMI. This could take a while...")
 
-    # Ensure AWS_DEFAULT_REGION env is set.
-    # Set it explicitly if the CLI arg was passed. Otherwise let the config object do it.
-    if args.region:
-        os.environ["AWS_DEFAULT_REGION"] = args.region
     pcluster_config = PclusterConfig(config_file=args.config_file, fail_on_file_absence=True)
 
     ami_info = _validate_createami_args_ami_compatibility(args)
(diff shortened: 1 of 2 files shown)


cli/pcluster/commands.py
```diff
@@ -418,6 +418,8 @@ def ssh(args, extra_args):  # noqa: C901 FIXME!!!
     :param extra_args: pcluster CLI extra_args
     """
     # FIXME it always search for the default config file
+    if args.region:
+        os.environ["AWS_DEFAULT_REGION"] = args.region
```
tilne, line 422: after looking at cli.py it seems this isn't needed.

cli/pcluster/createami.py
```diff
@@ -308,8 +308,6 @@ def create_ami(args):
 
     # Ensure AWS_DEFAULT_REGION env is set.
     # Set it explicitly if the CLI arg was passed. Otherwise let the config object do it.
-    if args.region:
-        os.environ["AWS_DEFAULT_REGION"] = args.region
```
rexcsn, line 312: I guess this code never worked, but createami command seems to accept a -r/--region option.
Should we remove that option from createami, or somehow make it work?

cli/pcluster/createami.py
```diff
@@ -308,8 +308,6 @@ def create_ami(args):
 
     # Ensure AWS_DEFAULT_REGION env is set.
     # Set it explicitly if the CLI arg was passed. Otherwise let the config object do it.
```
hanwen-pcluste, line 310: Shall we delete the comments?

hanwen-pcluste (approved):

cli/pcluster/createami.py
tilne: yes. good catch.

cli/pcluster/createami.py
tilne: i thought the same thing, bu this code was. redundant. there's a conditional in cli.py::main that does this for any subcommand that supports a region arg.

cli/pcluster/createami.py
tilne: done.

hanwen-pcluste (approved):

PR #2124, opened 2020-10-12
Fix pre/post install args double quotes issue

yuleiwan:
* Add json escape for pre/post install args in CLI
* Support single quote and double quotes args format
* Convert dna.json in master substack template to plain syntax string to avoid json escape twice
* Add unit test and integration test


cli/pcluster/config/cfn_param_types.py
```diff
@@ -816,6 +816,38 @@ def refresh(self):
...
+        return self
+
+    def to_cfn(self):
+        """Convert param to CFN representation, if "cfn_param_mapping" attribute is present in the Param definition."""
+        cfn_params = {}
+        cfn_converter = self.definition.get("cfn_param_mapping", None)
+
+        if cfn_converter:
+            cfn_value = self.get_cfn_value()
+            cfn_params[cfn_converter] = json.dumps(cfn_value)[1:-1]
```
rexcsn, line 846: I assume this is done to remove `""` from `cfn_value`, is there a way to do this not with indexing? I am just afraid we might remove something unexpected silently. If not, please add a comment in the code here clarifying why this `[1:-1]` logic is needed

cloudformation/master-server-substack.cfn.yaml
```diff
@@ -480,55 +480,62 @@ Resources:
...
-              owner: root
-              group: root
-              content:
-                cfncluster:
-                  stack_name: !Ref 'MainStackName'
-                  enable_efa: !Ref 'EFA'
-                  cfn_raid_vol_ids: !Ref 'RAIDVolIds'
-                  cfn_raid_parameters: !Ref 'RAIDOptions'
-                  cfn_scheduler_slots: !If
+              content: !Sub
```
rexcsn, line 483: Just curious, why is `Sub` needed here? I guess we can't deploy file with `""` using just userdata?

tests/integration-tests/tests/cfn-init/test_cfn_init/test_install_args_quotes/post_install.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "post-install script has $# arguments"
+for arg in "$@"
+do
+    echo "arg: ${arg}"
```
rexcsn, line 6: Should we assert on the actual args we see to make sure the args are `R curl wget` and not `"R curl wget"`

tests/integration-tests/tests/cfn-init/test_cfn_init.py
```diff
@@ -54,6 +54,28 @@ def test_replace_compute_on_failure(
...
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "pre_install.sh"), "scripts/pre_install.sh")
+    bucket.upload_file(str(test_datadir / "post_install.sh"), "scripts/post_install.sh")
+
+    # Create cluster with initial configuration
+    init_config_file = pcluster_config_reader(bucket_name=bucket_name)
+    cluster = clusters_factory(init_config_file)
+
+    # Check master and compute node status
+    _assert_server_status(cluster)
```
rexcsn, line 76: Should we assert on the actual args we see to make sure the args are `R curl wget` and not `"R curl wget"`.
Your pre/post install will not trigger an error even if `"R curl wget"` is passed into scripts as a single argument

cli/tests/pcluster/config/test_param_to_cfn.py
```diff
@@ -87,6 +87,11 @@ def test_param_to_cfn_value(mocker, section_definition, param_key, param_value,
         (CLUSTER_SIT, "shared_dir", "test", {"SharedDir": "test"}),
         # (CLUSTER_SIT, "shared_dir", {"ebs": [], "shared_dir": "test"}, {"SharedDir": "test"}),
         # (CLUSTER_SIT, "shared_dir", {"ebs": [{"label": "fake_ebs"}], "shared_dir": "unused_value"}, {}),
+        # ArgsParam
+        (CLUSTER_SIT, "pre_install_args", '"R wget"', {"PreInstallArgs": '\\"R wget\\"'}),
+        (CLUSTER_SIT, "pre_install_args", "'R wget'", {"PreInstallArgs": "'R wget'"}),
```
rexcsn, line 92: Do we need to add a test case for no `""`, so like `"pre_install_args", "single_argument_no_quotes"`

yuleiwan: Done

cloudformation/master-server-substack.cfn.yaml
yuleiwan: yes, the reason is "${SchedulerSlots}" and "${DisableHyperthreadingManually}" are two conditions. We can't deploy conditions directly and need !sub to define the value below. like
```
  !Sub
  - String
  - Var1Name: Var1Value
    Var2Name: Var2Value
```

tests/integration-tests/tests/cfn-init/test_cfn_init.py
yuleiwan: good point, done

tests/integration-tests/tests/cfn-init/test_cfn_init/test_install_args_quotes/post_install.sh
yuleiwan: good point, done

tests/integration-tests/tests/cfn-init/test_cfn_init/test_install_args_quotes/post_install.sh
```diff
@@ -0,0 +1,15 @@
...
+done
+
+case $# in
+    4)
```
rexcsn, line 10: Is this correct?

```
$ cat test_args.sh 
#!/bin/bash

echo "$#"

$ ./test_args.sh R curl wget
3
```

Alternative we can just assert each argument individually, i.e. `"$1" == "R"`, `"$2" == "curl"` ...
It's ok to hard code our assertions in tests, to avoid potential confusion and unexpected silent failures

tests/integration-tests/tests/cfn-init/test_cfn_init.py
rexcsn: Let's do some test runs locally to make sure this test fails for existing release version, i.e. it can catch the failure, and passes for your patch

tests/integration-tests/tests/cfn-init/test_cfn_init/test_install_args_quotes/post_install.sh
yuleiwan: Because in post_install_args, the first argument is s3 bucket url, so if we use "R curl wget", we will have 4 argument. I checked this in `cfn-init.log `.

rexcsn (approved):

PR #1632, opened 2020-02-06
Enable AWS Batch in GovCloud regions

demartinofra:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2131, opened 2020-10-14
Upgrade x86_64 CodeBuild image

tilne:
There are two reasons to upgrade this image. The x86_64 image we
currently use is deprecated, and this deprecated image is not supported
in govcloud. In order to support that partition we need to use a newer
image.


tilne:
This is hopefully the last thing needed to unblock #1632.

demartinofra (approved):

demartinofra (approved):

PR #2133, opened 2020-10-15
Validator for multiple "shared_dir"

chenwany:

*Description of changes:* 
Before this change, when user specify single EBS section, "shared_dir" can be specified together both in the cluster section and EBS section.  Only the "shared_dir" specified in the EBS section would be used without telling the customer about that. This change adds a validator to check the following cases:
When using single EBS section, "shared_dir" can not be specified both in cluster section and EBS section. 
When using multiple EBS sections, "shared_dir" can not be specified in the cluster section.

### DIFF
--- a/cli/pcluster/config/mappings.py
+++ b/cli/pcluster/config/mappings.py
@@ -56,6 +56,7 @@ from pcluster.config.validators import (
     dcv_enabled_validator,
     disable_hyperthreading_architecture_validator,
     disable_hyperthreading_validator,
+    duplicate_shared_dir_validator,
     ebs_settings_validator,
     ebs_volume_iops_validator,
     ebs_volume_size_snapshot_validator,
@@ -97,6 +98,7 @@ from pcluster.config.validators import (
 )
 from pcluster.constants import CIDR_ALL_IPS, FSX_HDD_THROUGHPUT, FSX_SSD_THROUGHPUT, SUPPORTED_ARCHITECTURES
 
+CLUSTER_COMMON_VALIDATORS = [duplicate_shared_dir_validator]
 # This file contains a definition of all the sections and the parameters configurable by the user
 # in the configuration file.
 
@@ -933,7 +935,7 @@ CLUSTER_SIT = {
     "key": "cluster",
     "default_label": "default",
     "cluster_model": "SIT",
-    "validators": [cluster_validator],
+    "validators": [cluster_validator] + CLUSTER_COMMON_VALIDATORS,
     "params": OrderedDict(
         CLUSTER_COMMON_PARAMS + [
             ("placement_group", {
@@ -1030,6 +1032,7 @@ CLUSTER_HIT = {
     "key": "cluster",
     "default_label": "default",
     "cluster_model": "HIT",
+    "validators": CLUSTER_COMMON_VALIDATORS,
     "params": OrderedDict(
         CLUSTER_COMMON_PARAMS + [
             ("default_queue", {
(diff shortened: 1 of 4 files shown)


cli/pcluster/config/validators.py
```diff
@@ -1523,3 +1523,24 @@ def validate_storage_type_options(
...
+    warnings = []
+    config_parser = pcluster_config.config_parser
+    if config_parser:
+        shared_dir_in_cluster = config_parser.has_option(get_file_section_name("cluster", section_label), "shared_dir")
+        ebs_settings_in_cluster = config_parser.has_option(
+            get_file_section_name("cluster", section_label), "ebs_settings"
+        )
+
+        if shared_dir_in_cluster and ebs_settings_in_cluster:
+            errors.append("'shared_dir' can not be specify in the Cluster section when specifying 'ebs_settings'.")
```
tilne, line 1544: typo: 'specify' should be 'specified'

tilne: i would also not capitalize 'Cluster'

cli/tests/pcluster/config/test_validators.py
```diff
@@ -2505,3 +2505,44 @@ def test_ebs_volume_size_snapshot_validator(
...
+            {"shared_dir": "shared_directory1", "volume_size": 30},
+            {"shared_dir": "shared_directory2", "volume_size": 30},
+            "'shared_dir' can not be specify in the Cluster section when specifying 'ebs_settings'.",
+        ),
+        (
+            {"ebs_settings": "vol1, vol2"},
+            {"shared_dir": "shared_directory1", "volume_size": 30},
+            {"shared_dir": "shared_directory2", "volume_size": 30},
+            None,
+        ),
```
tilne, line 2536: can you add a test case for when shared_dir is in the cluster and EBS settings aren't specified?

cli/pcluster/config/mappings.py
```diff
@@ -1030,6 +1031,7 @@
     "key": "cluster",
     "default_label": "default",
     "cluster_model": "HIT",
+    "validators": [shared_dir_and_ebs_settings_validator],
```
tilne, line 1034: Instead of adding this to two lists, we could make a list called CLUSTER_COMMON_VALIDATORS, add this to it, and specify it for both cluster model definitions.

Then in CLUSTER_SIT, the validators attribute could be set to `CLUSTER_COMMON_VALIDATORS + [cluster_validator]`, and for CLUSTER_HIT it woulds be just the common validators list.

chenwany: Good idea, done!

cli/tests/pcluster/config/test_validators.py
chenwany: Good idea, done!

cli/pcluster/config/validators.py
chenwany: Fixed!

cli/tests/pcluster/config/test_validators.py
tilne: What i meant by this was a test case where the cluster_section_dict has shared_dir and nothing else.

tilne: Probably that doesn't need to block this patch though.

tilne (dismissed):

cli/tests/pcluster/config/test_validators.py
chenwany: Fixed!

enrico-usai (approved):
LGTM!

PR #2141, opened 2020-10-15
integ-tests: test AMI validations (OS validation and pcluster version validation)

hanwen-pcluste:
1. If an AMI was baked by another version of pcluster and is trying to be used by current pcluster, pcluster should detect the difference, fail the creation and give a message
2. If OS provided by a user is different from the actual OS of an AMI, pcluster should detect the fail the creation and give a message

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -243,11 +243,12 @@ class ClustersFactory:
         self.__created_clusters = {}
         self._keep_logs_on_failure = keep_logs_on_failure
 
-    def create_cluster(self, cluster, extra_args=None):
+    def create_cluster(self, cluster, extra_args=None, raise_on_error=True):
         """
         Create a cluster with a given config.
         :param cluster: cluster to create.
         :param extra_args: list of strings; extra args to pass to `pcluster create`
+        :param raise_on_error: raise exception if cluster creation fails
         """
         name = cluster.name
         config = cluster.config_file
@@ -261,11 +262,16 @@ class ClustersFactory:
         if extra_args:
             create_cmd_args.extend(extra_args)
         create_cmd_args.append(name)
-        result = run_command(create_cmd_args, timeout=7200)
+        result = run_command(
+            create_cmd_args,
+            timeout=7200,
+            raise_on_error=raise_on_error,
+        )
         if "Status: {0} - CREATE_COMPLETE".format(cluster.cfn_name) not in result.stdout:
             error = "Cluster creation failed for {0} with output: {1}".format(name, result.stdout)
             logging.error(error)
-            raise Exception(error)
+            if raise_on_error:
+                raise Exception(error)
         elif "WARNING" in result.stdout:
             error = "Cluster creation for {0} generated a warning: {1}".format(name, result.stdout)
             logging.warning(error)
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -66,6 +66,17 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["alinux2"]
           schedulers: ["slurm"]
+  create:
+    test_create.py::test_create_wrong_os:
+      dimensions:
+        - regions: ["eu-central-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["ubuntu1804"]
+    test_create.py::test_create_wrong_pcluster_version:
+      dimensions:
+        - regions: ["ca-central-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["alinux"]
   createami:
     test_createami.py::test_createami:
       dimensions:
@@ -86,6 +97,16 @@ test-suites:
         - regions: ["eu-west-1"]
           instances: {{ common.INSTANCES_DEFAULT_ARM }}
           oss: ["alinux2"]
+    test_createami.py::test_createami_wrong_os:
+      dimensions:
+        - regions: ["eu-central-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["alinux"]
+    test_createami.py::test_createami_wrong_pcluster_version:
+      dimensions:
+        - regions: ["ca-central-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["alinux"]
   dashboard:
     test_dashboard.py::test_dashboard:
       dimensions:
--- a/tests/integration-tests/remote_command_executor.py
+++ b/tests/integration-tests/remote_command_executor.py
@@ -27,8 +27,9 @@ class RemoteCommandExecutionError(Exception):
 class RemoteCommandExecutor:
     """Execute remote commands on the cluster master node."""
 
-    def __init__(self, cluster):
-        username = get_username_for_os(cluster.os)
+    def __init__(self, cluster, username=None):
+        if not username:
+            username = get_username_for_os(cluster.os)
         self.__connection = Connection(
             host=cluster.master_ip,
             user=username,
--- a/tests/integration-tests/tests/common/utils.py
+++ b/tests/integration-tests/tests/common/utils.py
@@ -67,7 +67,7 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
     try:
         if ami_type == "pcluster":
             ami_name = "aws-parallelcluster-{version}-{ami_name}".format(
-                version=_get_installed_parallelcluster_version(),
+                version=get_installed_parallelcluster_version(),
                 ami_name=AMI_TYPE_DICT.get(ami_type).get(os).get("name"),
             )
         else:
@@ -90,11 +90,53 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
         sys.exit(1)
 
 
+def retrieve_pcluster_ami_without_standard_naming(region, os, version="2.8.1", architecture="x86_64"):
+    try:
+        client = boto3.client("ec2", region_name=region)
+        ami_name = f"ami-for-testing-pcluster-version-validation-without-standard-naming-{version}-{os}"
+        amis = client.describe_images(
+            Filters=[{"Name": "name", "Values": [ami_name]}, {"Name": "architecture", "Values": [architecture]}],
+        ).get("Images", [])
+        if amis:
+            # If the AMI is found, return the AMI
+            return amis[0]["ImageId"]
+        else:
+            # If no AMI is found, copy from an official AMI
+            official_ami_name = "aws-parallelcluster-{version}-{ami_name}".format(
+                version=version, ami_name=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("name")
+            )
+            official_amis = client.describe_images(
+                Filters=[
+                    {"Name": "name", "Values": [official_ami_name]},
+                    {"Name": "architecture", "Values": [architecture]},
+                ],
+                Owners=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("owners"),
+            ).get("Images", [])
+            return client.copy_image(
+                Description="This AMI is a copy from an official AMI but uses a different naming. "
+                "It is used to bypass the AMI's name validation of pcluster version "
+                "to test the validation in Cookbook.",
+                Name=ami_name,
+                SourceImageId=official_amis[0]["ImageId"],
+                SourceRegion=region,
+            ).get("ImageId")
+
+    except ClientError as e:
+        LOGGER.critical(e.response.get("Error").get("Message"))
+        sys.exit(1)
+    except AttributeError as e:
+        LOGGER.critical("Error no attribute {0} in dict: {1}".format(os, e))
+        sys.exit(1)
+    except IndexError as e:
+        LOGGER.critical("Error no ami retrieved: {0}".format(e))
+        sys.exit(1)
+
+
 @retry(stop_max_attempt_number=3, wait_fixed=5000)
 def fetch_instance_slots(region, instance_type):
     return get_instance_info(instance_type, region).get("VCpuInfo").get("DefaultVCpus")
 
 
-def _get_installed_parallelcluster_version():
+def get_installed_parallelcluster_version():
     """Get the version of the installed aws-parallelcluster package."""
     return pkg_resources.get_distribution("aws-parallelcluster").version
--- a/tests/integration-tests/tests/create/test_create.py
+++ b/tests/integration-tests/tests/create/test_create.py
@@ -1,0 +1,69 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import logging
+
+import pytest
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from utils import get_username_for_os
+
+from tests.common.assertions import assert_errors_in_logs
+from tests.common.utils import (
+    get_installed_parallelcluster_version,
+    retrieve_latest_ami,
+    retrieve_pcluster_ami_without_standard_naming,
+)
+
+
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "ubuntu1804", "*")
+def test_create_wrong_os(region, os, pcluster_config_reader, clusters_factory, architecture, instance):
+    """Test error message when os provide is different from the os of custom AMI"""
+    # ubuntu1804 is specified in the config file but an AMI of centos7 is provided
+    wrong_os = "centos7"
+    logging.info("Asserting os fixture is different from wrong_os variable")
+    assert_that(os != wrong_os).is_true()
+    custom_ami = retrieve_latest_ami(region, wrong_os, ami_type="remarkable", architecture=architecture)
+    cluster_config = pcluster_config_reader(custom_ami=custom_ami)
+    cluster = clusters_factory(cluster_config, raise_on_error=False)
+    username = get_username_for_os(wrong_os)
+    remote_command_executor = RemoteCommandExecutor(cluster, username=username)
+
+    logging.info("Verifying error in logs")
+    assert_errors_in_logs(
+        remote_command_executor,
+        ["/var/log/cfn-init.log"],
+        ["RuntimeError", fr"custom AMI.+{wrong_os}.+base.+os.+config file.+{os}"],
+    )
+
+
+@pytest.mark.dimensions("ca-central-1", "c5.xlarge", "alinux", "*")
+def test_create_wrong_pcluster_version(region, os, pcluster_config_reader, clusters_factory, architecture, instance):
+    """Test error message when AMI provided was baked by a pcluster whose version is different from current version"""
+    current_version = get_installed_parallelcluster_version()
+    wrong_version = "2.8.1"
+    logging.info("Asserting wrong_version is different from current_version")
+    assert_that(current_version != wrong_version).is_true()
+    # Retrieve an AMI without 'aws-parallelcluster-<version>' in its name.
+    # Therefore, we can bypass the version check in CLI and test version check of .bootstrapped file in Cookbook.
+    wrong_ami = retrieve_pcluster_ami_without_standard_naming(
+        region, os, version=wrong_version, architecture=architecture
+    )
+    cluster_config = pcluster_config_reader(custom_ami=wrong_ami)
+    cluster = clusters_factory(cluster_config, raise_on_error=False)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+
+    logging.info("Verifying error in logs")
+    assert_errors_in_logs(
+        remote_command_executor,
+        ["/var/log/cfn-init.log"],
+        ["RuntimeError", fr"AMI was created.+{wrong_version}.+is.+used.+{current_version}"],
+    )
--- a/tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
+++ b/tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
@@ -1,0 +1,19 @@
+[global]
+cluster_template = default
+
+[aws]
+aws_region_name = {{ region }}
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = slurm
+master_instance_type = {{ instance }}
+custom_ami = {{ custom_ami }}
+master_root_volume_size = 100
+compute_root_volume_size = 100
+
+[vpc parallelcluster-vpc]
+vpc_id = {{ vpc_id }}
+master_subnet_id = {{ public_subnet_id }}
--- a/tests/integration-tests/tests/createami/test_createami.py
+++ b/tests/integration-tests/tests/createami/test_createami.py
@@ -10,12 +10,19 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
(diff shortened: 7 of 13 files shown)


tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,34 @@
...
+from tests.common.assertions import assert_errors_in_logs
+from tests.common.utils import retrieve_latest_ami
+
+
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "ubuntu1804", "slurm")
+def test_create_wrong_os(region, os, pcluster_config_reader, clusters_factory, test_datadir, scheduler, instance):
+    """Test error message when os provide is different from the os of custom AMI"""
+    # ubuntu1804 is specified in the config file but an AMI of centos7 is provided
+    wrong_os = "centos7"
+    custom_ami = retrieve_latest_ami(region, wrong_os)
```
yuleiwan, line 26: should we use pcluster ami rather than vanilla ami to test pcluster create to ensure common cases?

tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,34 @@
...
+import logging
+
+import pytest
+from remote_command_executor import RemoteCommandExecutor
+
+from tests.common.assertions import assert_errors_in_logs
+from tests.common.utils import retrieve_latest_ami
+
+
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "ubuntu1804", "slurm")
```
yuleiwan, line 21: Do we need to pass the instance type and scheduler in dimensions, but we don't use it in function and config file?

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -55,29 +58,13 @@ def test_createami(region, os, instance, request, pcluster_config_reader, vpc_st
...
-    ).is_true()
-    assert_that(
-        any("Starting Cinc Client".lower() in s.lower() for s in pcluster_createami_result.stdout.split("\n"))
-    ).is_true()
-
-    assert_that(pcluster_createami_result.stdout).does_not_contain("No custom AMI created")
+    stdout_lower = pcluster_createami_result.stdout.lower()
+    assert_that(stdout_lower).contains("downloading https://{0}-aws-parallelcluster.s3".format(region).lower())
+    assert_that(stdout_lower).does_not_contain("chef.io/chef/install.sh".lower())
+    assert_that(stdout_lower).does_not_contain("packages.chef.io".lower())
```
yuleiwan, line 64: I am not sure do we need to lower the string that doesn't contain upper case?

tests/integration-tests/tests/create/test_create.py
hanwen-pcluste: If we want to pin the dimension to `@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "ubuntu1804", "slurm")`, passing in the parameters is the only way I know to let `test_runner` collect the test.
If we don't pass `scheduler` and `instance`, the dimension has to be changed to `@pytest.mark.dimensions("eu-central-1", "*", "ubuntu1804", "*")`

tests/integration-tests/tests/create/test_create.py
hanwen-pcluste: Common case should be an unofficial AMI. I might be wrong, official AMIs are usually passed correctly since it is the case when custom_ami is not specified in the config file

tests/integration-tests/tests/createami/test_createami.py
hanwen-pcluste: Done

tests/integration-tests/remote_command_executor.py
```diff
@@ -27,8 +27,10 @@ def __init__(self, result):
 class RemoteCommandExecutor:
     """Execute remote commands on the cluster master node."""
 
-    def __init__(self, cluster):
-        username = get_username_for_os(cluster.os)
+    def __init__(self, cluster, actual_os=None):
```
rexcsn, line 30: Would it be make more sense to change `actual_os` to something like `execute_as_user`? The caller of this function will take care of figuring out the correct user?
In this context what we really care about is executing command as the right user, but not executing different commands that are dependent on OS, so I would not pass OS in for this reason.

tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,34 @@
...
+def test_create_wrong_os(region, os, pcluster_config_reader, clusters_factory, test_datadir, scheduler, instance):
+    """Test error message when os provide is different from the os of custom AMI"""
+    # ubuntu1804 is specified in the config file but an AMI of centos7 is provided
+    wrong_os = "centos7"
```
rexcsn, line 25: We would want to do a quick assertion that wrong_os != provided os here, in case we change the dimensions by accident in future

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -116,8 +103,38 @@ def test_createami_post_install(
...
-    assert_that(pcluster_createami_result.stdout).does_not_contain("No custom AMI created")
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "alinux", "sge")
+def test_createami_wrong_os(
+    region, os, instance, test_datadir, request, pcluster_config_reader, vpc_stack, architecture, scheduler
+):
+    """Test error message when os provide is different from the os of custom AMI"""
+    cluster_config = pcluster_config_reader()
+
+    # alinux is specified in the config file but an AMI of alinux2 is provided
+    wrong_os = "alinux2"
```
rexcsn, line 119: Similar to comment above, let's do a quick assertion here to make sure wrong_os != os

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -116,8 +103,38 @@ def test_createami_post_install(
...
+
+    # Custom Cookbook
+    custom_cookbook = request.config.getoption("createami_custom_chef_cookbook")
+    custom_cookbook_args = "" if not custom_cookbook else f"-cc {custom_cookbook}"
+
+    command = (
+        f"pcluster createami -ai {base_ami} -os {os} -r {region} -c {cluster_config.as_posix()} "
+        f"{custom_cookbook_args} -i {instance} {networking_args}"
+    )
+    createami_process = pexpect.spawn(command, timeout=600)
```
rexcsn, line 135: Just curious, why did we choose `pexpect,spawn` instead of `run_command ` here?

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,56 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
         sys.exit(1)
 
 
+def retrieve_pcluster_ami_without_standard_naming(region, os, version="2.8.1", architecture="x86_64"):
+    try:
+        client = boto3.client("ec2", region_name=region)
+        ami_name = f"pcluster-without-standard-naming-{version}-{os}"
```
rexcsn, line 96: How about a more descriptive name, something like `ami-for-test_create_wrong_pcluster_version-DO-NOT-DELETE`?

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,56 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
...
+                        {"Name": "name", "Values": [official_ami_name]},
+                        {"Name": "architecture", "Values": [architecture]},
+                    ],
+                    Owners=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("owners"),
+                )
+                .get("Images", [])
+            )
+            return client.copy_image(
+                Description="This AMI is a copy from a official AMI but uses different naming. "
+                "It is used to bypass the name validation of pcluster version validation",
```
rexcsn, line 121: Minor: I would mention that this AMI is used by `test_create_wrong_pcluster_version` here in description, and include why we need this AMI in `test_create_wrong_pcluster_version`

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,56 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
         sys.exit(1)
 
 
+def retrieve_pcluster_ami_without_standard_naming(region, os, version="2.8.1", architecture="x86_64"):
```
rexcsn, line 93: This is very specific to `test_create_wrong_pcluster_version`, should we move this function into test create file rather than keeping in common util?

tests/integration-tests/tests/create/test_create.py
rexcsn: Just a note, I don't think run time bake is fully supported today. In other words, if we start with a vanilla AMI there is chance that cluster might fail because of some other issue and not OS validation.
Of course OS validation should be the first thing that errors, so we don't necessarily need to change this, but just something to keep in mind

I would consider using a strong assertion to make sure cluster fails because of OS validation, please think about if asserting on `fr"custom AMI.+{wrong_os}.+base.+os.+{os}"` is enough. If it is, I am ok with keeping things as is

tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,65 @@
...
+    retrieve_pcluster_ami_without_standard_naming,
+)
+
+
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "ubuntu1804", "slurm")
+def test_create_wrong_os(region, os, pcluster_config_reader, clusters_factory, test_datadir, scheduler, instance):
+    """Test error message when os provide is different from the os of custom AMI"""
+    # ubuntu1804 is specified in the config file but an AMI of centos7 is provided
+    wrong_os = "centos7"
+    logging.info("Asserting os fixture is different from wrong_os variable, so that we can test the validation code")
```
rexcsn, line 32: Minor: filler phrases like `so that we can test the validation code` does not add much info, and is probably not needed in assertion message. If you decide to remove, please also remove for other assertion message

tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,65 @@
...
+@pytest.mark.dimensions("ca-central-1", "c5.xlarge", "alinux", "slurm")
+def test_create_wrong_pcluster_version(
+    region, os, pcluster_config_reader, clusters_factory, test_datadir, scheduler, instance
+):
+    """Test error message when AMI provided was baked by a pcluster whose version is different from current version"""
+    current_version = get_installed_parallelcluster_version()
+    wrong_version = "2.8.1"
+    logging.info("Asserting wrong_version is different from current_version, so that we can test the validation code")
+    assert_that(current_version != wrong_version).is_true()
+    wrong_ami = retrieve_pcluster_ami_without_standard_naming(region, os, version=wrong_version, architecture="x86_64")
```
rexcsn, line 55: Should talk about why we need an AMI without standard naming here with a comment

tests/integration-tests/tests/create/test_create.py
yuleiwan: what about we choose remarkable ami like deep learning ami for os validation test. Even if we use unofficial ami here, I think is remarkable ami  better than vanilla to ensure common cases and avoid other failure like volume is not enough?

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,56 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
...
+        if amis:
+            # If the AMI is found, return the AMI
+            return amis[0]["ImageId"]
+        else:
+            # If no AMI is found, copy from an official AMI
+            official_ami_name = "aws-parallelcluster-{version}-{ami_name}".format(
+                version=version, ami_name=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("name")
+            )
+            official_amis = (
+                boto3.client("ec2", region_name=region)
```
yuleiwan, line 109: why do we call boto3 twice?

hanwen-pcluste: Thanks, I simplified the code. But I am still keeping the two boto3 calls for simplicity. If we want to use only one boto3 call, we need to use a [paginator](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html) to ensure the first call gets everything we need. However, using paginator leads to possible extra calls

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: I will try to do it within 255 characters

tests/integration-tests/tests/create/test_create.py
hanwen-pcluste: Done

tests/integration-tests/tests/create/test_create.py
hanwen-pcluste: Done

tests/integration-tests/clusters_factory.py
```diff
@@ -243,11 +243,12 @@ def __init__(self, keep_logs_on_failure=False):
         self.__created_clusters = {}
         self._keep_logs_on_failure = keep_logs_on_failure
 
-    def create_cluster(self, cluster, extra_args=None):
+    def create_cluster(self, cluster, extra_args=None, suppress_exception=False):
```
rexcsn, line 246: Minor: we usually name `suppress_exception` as `raise_on_error` in our code repo. Would be nice if we can try to keep consistency here, but not required.

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -116,8 +107,74 @@ def test_createami_post_install(
...
-        any("No post install script".lower() in s.lower() for s in pcluster_createami_result.stdout.split("\n"))
-    ).is_false()
+    stdout_lower = pcluster_createami_result.stdout.lower()
+    assert_that(stdout_lower).does_not_contain("no post install script")
+    assert_that(stdout_lower).does_not_contain("no custom ami created")
+
+
+@pytest.mark.dimensions("eu-central-1", "c5.xlarge", "alinux", "sge")
+def test_createami_wrong_os(
+    region, os, instance, test_datadir, request, pcluster_config_reader, vpc_stack, architecture, scheduler
```
rexcsn, line 117: `scheduler` is not used in this test, so probably don't need to pass it in?

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -116,8 +107,74 @@ def test_createami_post_install(
...
+    logging.info("Verifying errors in stdout")
+    createami_process.expect("RuntimeError")
+    createami_process.expect(fr"custom AMI.+{wrong_os}.+base.+os.+config file.+{os}")
+    createami_process.expect("No custom AMI created")
+    createami_process.close()
+
+
+@pytest.mark.dimensions("ca-central-1", "c5.xlarge", "alinux", "sge")
+def test_createami_wrong_pcluster_version(
+    region, os, instance, test_datadir, request, pcluster_config_reader, vpc_stack, architecture, scheduler
```
rexcsn, line 151: `scheduler` is not used in the test, so probably don't need to pass it in?
Please double check for all fixtures, I think `test_datadir ` and `architecture` are not used for this particular test as well?

tests/integration-tests/clusters_factory.py
hanwen-pcluste: Done

tests/integration-tests/tests/createami/test_createami.py
hanwen-pcluste: Done

tests/integration-tests/tests/createami/test_createami.py
hanwen-pcluste: Done

rexcsn (approved):

tests/integration-tests/configs/develop.yaml
```diff
@@ -66,6 +66,17 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["alinux2"]
           schedulers: ["slurm"]
+  create:
+    test_create.py::test_create_wrong_os:
+      dimensions:
+        - regions: ["eu-central-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
```
demartinofra, line 73: you are specifying an instance here but the test function is not using the fixture. Same below

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
```diff
@@ -0,0 +1,19 @@
...
+cluster_template = default
+
+[aws]
+aws_region_name = {{ region }}
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = slurm
```
demartinofra, line 11: Rather than hardcoding I would expose this in the config as done in the other tests. Same applies elsewhere

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
```diff
@@ -0,0 +1,19 @@
...
+master_instance_type = {{ instance }}
+custom_ami = {{ custom_ami }}
+master_root_volume_size = 100
+compute_root_volume_size = 100
+
+[vpc parallelcluster-vpc]
+vpc_id = {{ vpc_id }}
+master_subnet_id = {{ public_subnet_id }}
```
demartinofra, line 19: If you don't need the compute nodes in a public subnet, you can add them to the private subnet and set use_public_ips to false. This will reduce the number of EIPs needed by the integration tests. Same applies elsewhere

tests/integration-tests/configs/develop.yaml
```diff
@@ -66,6 +66,17 @@ test-suites:
...
+          oss: ["ubuntu1804"]
```
demartinofra, line 74: Can you add a comment in order to mention that this test works only with an os that is different than centos7? let's try to add this hints in the config file so that we know what we can change without having to look at every single test

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,53 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
...
+                "It is used to bypass the AMI's name validation of pcluster version "
+                "to test the validation in Cookbook.",
+                Name=ami_name,
+                SourceImageId=official_amis[0]["ImageId"],
+                SourceRegion=region,
+            ).get("ImageId")
+
+    except ClientError as e:
+        LOGGER.critical(e.response.get("Error").get("Message"))
+        sys.exit(1)
```
demartinofra, line 126: you don't want to exit the process in the tests but always raise exception so that pytest executes the cleanup handlers. you could also avoid catching an exception, the framework will log it automatically

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,53 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
         sys.exit(1)
 
 
+def retrieve_pcluster_ami_without_standard_naming(region, os, version="2.8.1", architecture="x86_64"):
```
demartinofra, line 93: why are we hardcoding the version as a default?

tests/integration-tests/tests/common/utils.py
```diff
@@ -90,11 +90,53 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
...
+                version=version, ami_name=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("name")
+            )
+            official_amis = client.describe_images(
+                Filters=[
+                    {"Name": "name", "Values": [official_ami_name]},
+                    {"Name": "architecture", "Values": [architecture]},
+                ],
+                Owners=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("owners"),
+            ).get("Images", [])
+            return client.copy_image(
```
demartinofra, line 115: are we going to leave these AMIs in the account forever?do we want to put in place some clean-up logic?

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
```diff
@@ -0,0 +1,19 @@
...
+aws_region_name = {{ region }}
+
+[cluster default]
+base_os = {{ os }}
+key_name = {{ key_name }}
+vpc_settings = parallelcluster-vpc
+scheduler = slurm
+master_instance_type = {{ instance }}
+custom_ami = {{ custom_ami }}
+master_root_volume_size = 100
```
demartinofra, line 14: do we need to explicitly set master and compute root volume size in these clusters?

tests/integration-tests/tests/create/test_create.py
```diff
@@ -0,0 +1,69 @@
...
+    wrong_version = "2.8.1"
+    logging.info("Asserting wrong_version is different from current_version")
+    assert_that(current_version != wrong_version).is_true()
+    # Retrieve an AMI without 'aws-parallelcluster-<version>' in its name.
+    # Therefore, we can bypass the version check in CLI and test version check of .bootstrapped file in Cookbook.
+    wrong_ami = retrieve_pcluster_ami_without_standard_naming(
+        region, os, version=wrong_version, architecture=architecture
+    )
+    cluster_config = pcluster_config_reader(custom_ami=wrong_ami)
+    cluster = clusters_factory(cluster_config, raise_on_error=False)
```
demartinofra, line 61: here cluster creation might fail for any reason not only the one you expect it to fail for. In some cases the cluster might not even be created and you won't be able to instantiate the RemoteCommandExecutor. Shall we make this more robust by asserting the reason of the failure?

tests/integration-tests/configs/develop.yaml
hanwen-pcluste: I am setting instance in the config files using this dimension. Should I use the fixture elsewhere in the test?

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: Yes, it is unnecessary now. I will remove them.

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: Should I call deregister explicitly? Or just wait for the "cleanup_old_private_ami" on Jenkins to run?

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
hanwen-pcluste: I am using deeping learning AMIs for runtime baking AMI. They require root volume at least 100GB

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
hanwen-pcluste: I will change it

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
hanwen-pcluste: I will change it

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: I will change it

tests/integration-tests/configs/develop.yaml
hanwen-pcluste: I will add it

tests/integration-tests/tests/create/test_create.py
hanwen-pcluste: Do you know how I can send failing reasons from Cookbook to CloudFormation?

tests/integration-tests/configs/develop.yaml
demartinofra: I think for some reason I just missed the fact that you were passing instance as an arg to the test function.

tests/integration-tests/tests/common/utils.py
demartinofra: it would be nice to have the test properly cleanup all resources that get created without relying on external cleanup projects. What you could is to transform this into a fixture so that the cleanup is automatically performed when the test case enters the tear down phase.

https://docs.pytest.org/en/stable/fixture.html#fixture-finalization-executing-teardown-code

hanwen-pcluste: Yea. Our concern was that deleting AMI in production account is 'dangerous'. Anyway, I will use fixture to remove the AMI

tests/integration-tests/tests/create/test_create.py
demartinofra: we could at least assert that the cluster exists before proceeding since the test requires that the create fails but the stack is there. it is just to make failures in tests easier to debug.

tests/integration-tests/tests/common/utils.py
demartinofra: that is a valid point. Maybe before triggering a deletion let's make sure the AMI we are trying to delete is a private one

PR #2143, opened 2020-10-15
Bug fixes: ebs settings from_cfn referred section labels, missing tags arg

tilne:
Two issues are addressed by this PR:
1. EBSSettingsCfnParam.from_storage was passing the cluster section's key and label when creating param objects to represent the referred EBS sections. Change that to pass those constructors the section key and label for the referred EBS section instead.
2. A call to `_upload_hit_resources` in update.py was missing the mandatory `tags` arg. Make that arg optional instead.

In an effort to avoid errors like 2. in the future, the last commit in this PR adds a pylint check to the code-linters tox environment that we require to pass for every PR. Right now that check only looks for that specific type of error, but over time we can add more.

### DIFF
--- a/cli/pcluster/commands.py
+++ b/cli/pcluster/commands.py
@@ -70,7 +70,9 @@ def _create_bucket_with_resources(pcluster_config, json_params, tags):
     return s3_bucket_name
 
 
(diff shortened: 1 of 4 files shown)


ddeidda (approved):
Good job!

PR #2166, opened 2020-10-22
Fix cfn_scheduler_slots in pcluster update

yuleiwan:
* pass cores rather than an integer to dna.json when disable_hyperthreading = true
* add validator for case extra json {"cfn_scheduler_slots" = number}, give a warning
* add unit test for new validator
* add extra json update policy 

### DIFF
--- a/cli/pcluster/config/update_policy.py
+++ b/cli/pcluster/config/update_policy.py
@@ -90,6 +90,11 @@ UpdatePolicy.FAIL_REASONS = {
     "ebs_volume_resize": "Updating the file system after a resize operation requires commands specific to your "
     "operating system.",
     "ebs_sections_change": "EBS sections cannot be added or removed during a 'pcluster update' operation",
+    "extra_json_update": lambda change, patch: "Updating the extra_json parameter is not supported because it only "
+    "applies updates to compute nodes. If you still want to proceed, first stop the cluster with the "
+    "following command:\n{0} -c {1} {2} and then run an update with the --force flag".format(
+        "pcluster stop", patch.config_file, patch.cluster_name
+    ),
 }
 
 # Common action_needed messages
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -352,6 +352,19 @@ def disable_hyperthreading_architecture_validator(param_key, param_value, pclust
     return errors, warnings
 
 
+def extra_json_validator(param_key, param_value, pcluster_config):
+    errors = []
+    warnings = []
+
+    if param_value and param_value.get("cluster") and param_value.get("cluster").get("cfn_scheduler_slots"):
+        warnings.append(
+            "It is highly recommended to use the disable_hyperthreading parameter in order to control the "
+            "hyper-threading configuration in the cluster rather than using cfn_scheduler_slots in extra_json"
+        )
+
+    return errors, warnings
+
+
 def dcv_enabled_validator(param_key, param_value, pcluster_config):
     errors = []
     warnings = []
--- a/cloudformation/master-server-substack.cfn.yaml
+++ b/cloudformation/master-server-substack.cfn.yaml
@@ -538,7 +538,7 @@ Resources:
                   }
                 - SchedulerSlots: !If
                     - DisableComputeHyperthreading
-                    - !Ref 'ComputeCoreCount'
+                    - cores
                     - vcpus
                   DisableHyperthreadingManually: !If
                     - DisableMasterHyperthreadingManually
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ CHANGELOG
 - Enable queue resizing on update without requiring to stop the compute fleet. Stopping the compute fleet is only
   necessary when existing instances risk to be terminated.
 
+**BUG FIXES**
+
+- Fix cfn_scheduler_slots don't update when updating compute instance type by ``pcluster update``. 
+
 2.9.1
 -----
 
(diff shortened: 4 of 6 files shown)


cli/pcluster/config/validators.py
```diff
@@ -352,6 +352,26 @@ def disable_hyperthreading_architecture_validator(param_key, param_value, pclust
...
+    warnings = []
+
+    if (
+        param_value
+        and param_value.get("cluster")
+        and param_value.get("cluster").get("cfn_scheduler_slots")
+        and str(param_value.get("cluster").get("cfn_scheduler_slots")).isdigit()
+    ):
+        warnings.append(
+            "you create a cluster with 'cfn_scheduler_slots' = {0}, you cannot update it using pcluster update. "
```
ddeidda, line 366: I'd ask a wording review from @nstornetta here.

ddeidda (commented):
LGTM; I'd just ask for a quick wording review on the warning messages.

cli/pcluster/config/update_policy.py
```diff
@@ -201,3 +201,18 @@ def _check_min_count(change, patch):
     )
     + ". If you need this change, please consider creating a new cluster instead of updating the existing one.",
 )
+
+# Update extra_json
+UpdatePolicy.EXTRA_JSON = UpdatePolicy(
+    level=1000,
+    fail_reason="Updating extra_json will only affect on computefleet and not affect on head node",
+    action_needed=lambda change, patch: (
```
demartinofra, line 209: Also I'd move the part on forcing the update to the fail_reason. At that point the action needed becomes the same as UNSUPPORTED which means you could put that code in common by creating a function

cli/pcluster/config/update_policy.py
```diff
@@ -201,3 +201,18 @@ def _check_min_count(change, patch):
     )
     + ". If you need this change, please consider creating a new cluster instead of updating the existing one.",
 )
+
+# Update extra_json
+UpdatePolicy.EXTRA_JSON = UpdatePolicy(
+    level=1000,
+    fail_reason="Updating extra_json will only affect on computefleet and not affect on head node",
```
demartinofra, line 208: What about something like? 
`Updating the extra_json parameter is not supported because changes would only take place on compute nodes. If you still want to proceed, first stop the cluster with ... and then run an update with the --force flag.`

cli/pcluster/config/validators.py
```diff
@@ -352,6 +352,19 @@ def disable_hyperthreading_architecture_validator(param_key, param_value, pclust
     return errors, warnings
 
 
+def extra_json_validator(param_key, param_value, pcluster_config):
```
demartinofra, line 355: I'm probably more in favor of removing this validator in order to be consistent with all other parameters that cannot be updated. Or what could be useful is to change the message and suggest to use the disable_hyperthreading parameter instead: `it is highly recommended to use the disable_hyperthreading parameter in order to control the hyper-threading configuration in the cluster`

cloudformation/master-server-substack.cfn.yaml
```diff
@@ -538,7 +538,7 @@ Resources:
                   }
                 - SchedulerSlots: !If
                     - DisableComputeHyperthreading
-                    - !Ref 'ComputeCoreCount'
+                    - cores
```
demartinofra, line 541: What about SchedulerSlots in compute-fleet-substack? do we need to change that part as well?

yuleiwan: We don't have to change SchedulerSlots in compute-fleet-substack, because when compute instance update, cfnconfig in compute node will update as well, the number of cfn_scheduler_slots is correct.
Another aspect,
for `node.default['cfncluster']['cfn_instance_slots']`, 
go to this branch
`elsif node['cfncluster']['cfn_scheduler_slots'] == 'cores' node['cpu']['cores']`
or go to this branch 
`else node['cfncluster']['cfn_scheduler_slots']`
will get the same result

cli/pcluster/config/validators.py
yuleiwan: Done

cli/pcluster/config/update_policy.py
yuleiwan: Done

cli/pcluster/config/update_policy.py
yuleiwan: Done

cli/pcluster/config/validators.py
```diff
@@ -352,6 +352,22 @@ def disable_hyperthreading_architecture_validator(param_key, param_value, pclust
...
+    errors = []
+    warnings = []
+
+    if param_value and param_value.get("cluster") and param_value.get("cluster").get("cfn_scheduler_slots"):
+        warnings.append(
+            "you create a cluster with 'cfn_scheduler_slots' = {0}, you cannot update this value using pcluster "
```
demartinofra, line 361: I'd drop the first sentence and say something like:
`It is highly recommended to use the disable_hyperthreading parameter in order to control the hyper-threading configuration in the cluster rather than using cfn_scheduler_slots in extra_json`

yuleiwan: Done

demartinofra (approved):

cli/pcluster/config/update_policy.py
```diff
@@ -90,6 +90,11 @@ def __eq__(self, other):
     "ebs_volume_resize": "Updating the file system after a resize operation requires commands specific to your "
     "operating system.",
     "ebs_sections_change": "EBS sections cannot be added or removed during a 'pcluster update' operation",
+    "extra_json_update": lambda change, patch: "Updating the extra_json parameter is not supported because it only "
```
yuleiwan, line 93: modified the message

demartinofra (approved):

PR #2167, opened 2020-10-22
Fix failing tests for validation of pcluster version

hanwen-pcluste:
This commit will
1. Amend a merged pull request: #2141.
2. Use private subnet for compute node and set use_public_ips to false to reduce usage of EIP for tests created in the merged PR
3. Use raise instead of sys.exit() if errors occur
4. Use a fixture for the AMI created during test for validation of pcluster version, so that it will be cleaned up after the tests end.
5. Since the tests are testing failing cases of cluster creation, assert the head node is running before asserting content of logs inside the node to facilitate debug in the future.

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -77,12 +77,14 @@ test-suites:
       dimensions:
         - regions: ["eu-central-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["ubuntu1804"]
+          oss: ["ubuntu1804"]  # os must be different from centos7 to test os validation logic when wrong os is provided
+          schedulers: ["slurm"]
     test_create.py::test_create_wrong_pcluster_version:
       dimensions:
         - regions: ["ca-central-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["alinux"]
+          schedulers: ["slurm"]
   createami:
     test_createami.py::test_createami:
       dimensions:
@@ -107,7 +109,7 @@ test-suites:
       dimensions:
         - regions: ["eu-central-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux"]
+          oss: ["alinux"]  # os must be different from alinux2 to test os validation logic when wrong os is provided
     test_createami.py::test_createami_wrong_pcluster_version:
       dimensions:
         - regions: ["ca-central-1"]
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -51,6 +51,8 @@ from utils import (
     unset_credentials,
 )
 
+from tests.common.utils import retrieve_pcluster_ami_without_standard_naming
+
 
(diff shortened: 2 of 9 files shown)


tests/integration-tests/conftest.py
```diff
@@ -691,3 +693,24 @@ def architecture(request, instance, region):
...
+    ami_id = None
+
+    def _pcluster_ami_without_standard_naming(version):
+        nonlocal ami_id
+        ami_id = retrieve_pcluster_ami_without_standard_naming(region, os, version, architecture)
+        return ami_id
+
+    yield _pcluster_ami_without_standard_naming
+
+    if ami_id and not request.config.getoption("no_delete"):
```
demartinofra, line 714: I don't think we want to apply the no_delete here since there is not a good way to reuse this resource in a next round of tests.

tests/integration-tests/configs/develop.yaml
```diff
@@ -77,7 +77,7 @@ test-suites:
       dimensions:
         - regions: ["eu-central-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["ubuntu1804"]
+          oss: ["ubuntu1804"]  # os must be different from centos7 to test os validation logic when wrong os is provided
```
demartinofra, line 80: don't forget to add the scheduler now that the test function uses it as fixture. same below.
was it working without having the scheduler specified here?

demartinofra (commented):
LGTM. Dropped just a couple of minor comments.

tests/integration-tests/conftest.py
hanwen-pcluste: With the AMI created, the test code can reuse the AMI in the next round of test and save 4 minutes of copying the AMI

tests/integration-tests/configs/develop.yaml
hanwen-pcluste: Done. I am not sure, because I was using test_runner to run tests locally

rexcsn (approved):

tests/integration-tests/conftest.py
```diff
@@ -691,3 +693,24 @@ def architecture(request, instance, region):
 def key_name(request):
     """Return the EC2 key pair name to be used."""
     return request.config.getoption("key_name")
+
+
+@pytest.fixture()
+def pcluster_ami_without_standard_naming(region, os, architecture):
```
rexcsn, line 699: Missing `version` parameter?

hanwen-pcluste: No. version is passed in in the yielded funtion

tests/integration-tests/conftest.py
demartinofra: that makes sense. still leaving undeleted resources that are defined outside of a CloudFormation stack, when specifying the --no-delete flag, makes it harder to perform a clean-up afterwards. For now I'd be in favor of only using the no-delete for vpc and cluster stacks and keep its behavior simple enough, with the downside of waiting 4 extra minutes when debugging this test. What do you think?

demartinofra (approved):
Good job!

PR #2173, opened 2020-10-24
integ-tests: test existing ebs volume

chenwany:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/integration-tests/tests/storage/test_ebs.py
+++ b/tests/integration-tests/tests/storage/test_ebs.py
@@ -139,6 +139,42 @@ def test_ebs_single_empty(scheduler, pcluster_config_reader, clusters_factory):
     _test_ebs_correctly_shared(remote_command_executor, mount_dir, scheduler_commands)
 
 
+@pytest.mark.dimensions("ap-northeast-2", "c5.xlarge", "centos7", "sge")
+@pytest.mark.usefixtures("os", "instance")
+def test_ebs_existing(
+    request, vpc_stacks, region, scheduler, pcluster_config_reader, clusters_factory, snapshots_factory
+):
+    logging.info("Testing ebs existing")
+    existing_mount_dir = "existing_mount_dir"
+
+    logging.info("Creating volume")
+
+    volume_id = snapshots_factory.create_existing_volume(
+        request, vpc_stacks[region].cfn_outputs["PublicSubnetId"], region
+    )
+
+    logging.info("Existing Volume id: %s" % volume_id)
+    cluster_config = pcluster_config_reader(
+        volume_id=volume_id,
+        existing_mount_dir=existing_mount_dir,
+    )
+
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    existing_mount_dir = "/" + existing_mount_dir
+    _test_ebs_correctly_mounted(remote_command_executor, existing_mount_dir, volume_size="9.8")
+    _test_ebs_correctly_shared(remote_command_executor, existing_mount_dir, scheduler_commands)
+    # Checks for test data
+    result = remote_command_executor.run_remote_command("cat {}/test.txt".format(existing_mount_dir))
+    assert_that(result.stdout.strip()).is_equal_to("hello world")
+
+    # delete the cluster before detaching the EBS volume
+    cluster.delete()
+    # check the volume still exists after deleting the cluster
+    _assert_volume_exist(volume_id, region)
+
+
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -112,6 +112,40 @@ def test_ebs_single_empty(scheduler, pcluster_config_reader, clusters_factory):
...
 
+@pytest.mark.dimensions("ap-northeast-2", "c5.xlarge", "alinux2", "sge")
+@pytest.mark.usefixtures("os", "instance")
+def test_ebs_existing(
+    request, vpc_stacks, region, scheduler, pcluster_config_reader, clusters_factory, snapshots_factory
+):
+    logging.info("Testing ebs existing")
+    existing_mount_dir = "existing_mount_dir"
+
+    logging.info("Creating volume id")
```
tilne, line 123: minor, terminology: you're creating a volume, and that volume happens to have an ID. i would say 'Creating volume' instead

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -112,6 +112,40 @@ def test_ebs_single_empty(scheduler, pcluster_config_reader, clusters_factory):
...
+
+    volume_id = snapshots_factory.create_existing_volume(
+        request, vpc_stacks[region].cfn_outputs["PublicSubnetId"], region
+    )
```
tilne, line 127: it doesn't look like this volume is deleted. can you do it via a fixture like the snapshots_factory fixture defined in test_ebs.py?

chenwany: It is deleted because I am using the snapshots_factory fixture here to create the volume

tests/integration-tests/tests/storage/test_ebs.py
chenwany: Right! Done

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -113,6 +113,40 @@ def test_ebs_single_empty(scheduler, pcluster_config_reader, clusters_factory):
...
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    existing_mount_dir = "/" + existing_mount_dir
+    _test_ebs_correctly_mounted(remote_command_executor, existing_mount_dir, volume_size="9.8")
+    _test_ebs_correctly_shared(remote_command_executor, existing_mount_dir, scheduler_commands)
+    # Checks for test data
+    result = remote_command_executor.run_remote_command("cat {}/test.txt".format(existing_mount_dir))
+    assert_that(result.stdout.strip()).is_equal_to("hello world")
+
+    # delete the cluster
+    cluster.delete()
```
tilne, line 147: why delete the cluster here?

tilne: i think the reason for deleting it is to ensure that the existing volume used when creating the cluster persists through deletion. if so, can you add a check after deleting the cluster that the volume still exists?

chenwany: The reason why we delete the cluster is because we have the fixture to clean up the existing ebs volume. The volume can not be detached if the cluster is still running.

tilne: Can you still add the check to ensure the pre-existing volume still exists after deleting the cluster? I think it's a good check to have in the test.

chenwany: yes! it is added

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -175,6 +212,13 @@ def _test_ebs_resize(remote_command_executor, mount_dir, volume_size):
     assert_that(result.stdout).matches(r"{size}G".format(size=volume_size))
 
 
+def _check_volume_exist(volume_id, region):
```
tilne, line 215: minor style point: a common pattern used in the integration tests is to name these types of functions `_assert_X_Y_Z` and then have these functions perform the assertions. so instead of naming this one `_check_volume_exist` you could have named it `_assert_volume_exists` and then performed the assertion here that you perform at the end `test_ebs_existing` after deleting the cluster rather than returning a value to be used in an assertion by the calling function.

This is a super minor point though. I don't think it's necessary to change this  before merging unless you really want to.

tilne (approved):

tests/integration-tests/tests/storage/test_ebs.py
chenwany: Great! Done!

tilne (approved):

PR #2180, opened 2020-10-28
integ-tests: test different volume types

chenwany:
add test cases for volume types sc1 and st1

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/integration-tests/tests/storage/test_ebs.py
+++ b/tests/integration-tests/tests/storage/test_ebs.py
@@ -76,13 +76,23 @@ def test_ebs_snapshot(
 def test_ebs_multiple(scheduler, pcluster_config_reader, clusters_factory):
     mount_dirs = ["/ebs_mount_dir_{0}".format(i) for i in range(0, 5)]
     volume_sizes = [15 + 5 * i for i in range(0, 5)]
+
+    # for volume type sc1 and st1, the minimum volume sizes are 500G
+    volume_sizes[3] = 500
+    volume_sizes[4] = 500
     cluster_config = pcluster_config_reader(mount_dirs=mount_dirs, volume_sizes=volume_sizes)
     cluster = clusters_factory(cluster_config)
     remote_command_executor = RemoteCommandExecutor(cluster)
 
     scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -76,13 +76,18 @@ def test_ebs_snapshot(
...
+    volume_sizes[4] = 500
     cluster_config = pcluster_config_reader(mount_dirs=mount_dirs, volume_sizes=volume_sizes)
     cluster = clusters_factory(cluster_config)
     remote_command_executor = RemoteCommandExecutor(cluster)
 
     scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
     for mount_dir, volume_size in zip(mount_dirs, volume_sizes):
-        _test_ebs_correctly_mounted(remote_command_executor, mount_dir, volume_size)
+        # for volume size equal to 500G, the filesystem size is only 492G
+        _test_ebs_correctly_mounted(remote_command_executor, mount_dir, volume_size if volume_size != 500 else 492)
```
tilne, line 90: what accounts for the difference here?

tilne (commented):
we support `volume_type`'s of standard, io1, gp2, st1, and sc1. in the test you modified we have two gp2 volumes and no standard volumes. should we change one of them to standard so that we're covering every supported case?

tests/integration-tests/tests/storage/test_ebs.py
chenwany: When I use command `lsblk` to list the partition size, it is equal to 500G, when I use `df -h`, the filesystem of that partition is 492G. I google the reason, it says that the file systems use some of the total space available on a device for storing internal structures and data (the file system's metadata). The overhead of the XFS filesystem is around 0.5%. Before, we only test with at most 40G, so the number is not large enough to show the gap between the partition size and the filesystem size. For sc1 and st1, the minimum size is 500G, so there will be a size difference.

tilne: can you add a comment that explains this?

chenwany: Yes, it is added

tilne (approved):

PR #2184, opened 2020-10-30
Extend RAID test to work with different mdadm versions

enrico-usai:
The output of `mdadm` utility in is a bit different, this could depend by the version
of the tool (4.1-13 on Centos8 vs 4.1-4 on Centos7) or by the fact we're using a
different superblock version (1.2 on Centos8 vs 0.90 on Centos7 and other OSes).

With the same `volume_size` we have:
* `Array Size : 41942912 (40.00 GiB 42.95 GB)` on Centos7 with `mdadm-4.1-4.el7`
* `Array Size : 41908224 (39.97 GiB 42.91 GB)` on Centos8 with `mdadm-4.1-13.el8`

# References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/789
* https://raid.wiki.kernel.org/index.php/RAID_superblock_formats


tests/integration-tests/tests/storage/test_raid.py
```diff
@@ -73,11 +74,17 @@ def _test_raid_correctly_shared(remote_command_executor, mount_dir, scheduler_co
...
-    assert_that(result.stdout).matches(r"Array Size : .*{0}\.00 GiB".format(volume_size))
     assert_that(result.stdout).contains("Raid Devices : {0}".format(raid_devices))
     assert_that(result.stdout).contains("Active Devices : {0}".format(raid_devices))
     assert_that(result.stdout).contains("Failed Devices : 0")
 
+    # Calculate rounded size to match output from different mdadm version
+    # Array Size : 41942912 (40.00 GiB 42.95 GB) --> on Centos7 with mdadm-4.1-4.el7
+    # Array Size : 41908224 (39.97 GiB 42.91 GB) --> on Centos8 with mdadm-4.1-13.el8
+    array_size = re.search(r"Array Size : .*\((.*) GiB", result.stdout).group(1)
+    rounded_size = round(float(array_size))
```
rexcsn, line 85: I am not sure rounding is the right approach here. For example, say we have an array that is supposed to be `8GiB` but we actually got `7.5GiB`, `round(7.5)=8` so the test will pass, but the volume is missing `6.25%` of its expected volume.

That is very different than the `39.97/40.00` case we want to accept here.
Of course, we are hard coding the volume size we want to test, so the test will not pass if more than `0.5/40 = 1.25%`(still a big variation IMO) of the volume is missing. However it's not clear how much volume we expect to be missing, and test could be invalidated if we change the size we are testing.

In any case, I think a percentage approach might be more reasonable here? We should set a strict bar on how much variation is allow, maybe 1%?

Also would be nice if we can understand more about this difference(not required, of course), there seems to be [some info here](https://unix.stackexchange.com/a/124948) but I am not sure if this is the exact case.

enrico-usai: I refactored the code to be sure to have `array_size` greater then or equals to `volume_size - 0.1`

rexcsn (approved):

PR #2201, opened 2020-11-05
Add Centos8 as supported OS for EFA

enrico-usai:
The `allowed_oses` condition is no longer needed because all the OSes are supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ CHANGELOG
 
 **CHANGES**
 
+- Upgrade EFA installer to version 1.10.1
+  - EFA configuration: ``efa-config-1.5`` (from efa-config-1.4)
+  - EFA profile: ``efa-profile-1.1`` (from efa-profile-1.0.0)
+  - EFA kernel module: ``efa-1.10.2`` (from efa-1.6.0)
+  - RDMA core: ``rdma-core-31.amzn0`` (from rdma-core-28.amzn0)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ CHANGELOG
 
 **CHANGES**
 
+- Upgrade EFA installer to version 1.10.1 to support CentOS 8:
```
demartinofra, line 15: the no change is with respect to 1.10.0 or with respect to what was deployed with pcluster 2.9.1? we should compare with the latter. Also I'd list the added support for centos 8 as a subitem

tests/integration-tests/configs/develop.yaml
```diff
@@ -198,8 +198,7 @@ test-suites:
       dimensions:
         - regions: ["us-east-1"]
           instances: ["c5n.18xlarge"]
-          # TODO: use INSTANCES_DEFAULT_X86 once there is centos8 support for efa
-          oss: ["alinux", "alinux2", "centos7", "ubuntu1604", "ubuntu1804"]
+          oss: {{ common.INSTANCES_DEFAULT_X86 }}
```
demartinofra, line 201: I think there is an error here cause you are using instances and not oss :)

enrico-usai: Fixed

demartinofra (approved):

PR #2204, opened 2020-11-06
Fix ec2_iam_role validator

rexcsn:
* Update ec2_iam_role validator to check for instance policies published in documentation

### DIFF
--- a/cli/pcluster/config/validators.py
+++ b/cli/pcluster/config/validators.py
@@ -8,6 +8,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+import json
 import logging
 import re
 import urllib.error
@@ -15,7 +16,9 @@ import urllib.request
 from urllib.parse import urlparse
 
 import boto3
+import pkg_resources
 from botocore.exceptions import ClientError
+from jinja2 import Environment, FileSystemLoader
 
 from pcluster.constants import CIDR_ALL_IPS, FSX_HDD_THROUGHPUT, FSX_SSD_THROUGHPUT
 from pcluster.dcv.utils import get_supported_dcv_os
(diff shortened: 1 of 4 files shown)


cli/pcluster/config/validators.py
```diff
@@ -508,12 +510,12 @@ def ec2_iam_role_validator(param_key, param_value, pcluster_config):
         iam = boto3.client("iam")
         arn = iam.get_role(RoleName=param_value).get("Role").get("Arn")
         account_id = boto3.client("sts", endpoint_url=_get_sts_endpoint()).get_caller_identity().get("Account")
-
-        iam_policy = _get_pcluster_user_policy(get_partition(), get_region(), account_id)
+        scheduler = pcluster_config.get_section("cluster").get_param_value("scheduler")
+        iam_policy = _get_pcluster_instance_policy(scheduler, get_partition(), get_region(), account_id)
```
tilne, line 514: this is just a naming change right? we were always validating instance policies, and the function that did so was just not accurately named?

i'm pretty certain this is true looking at the code, but wanted to double-check

cli/pcluster/config/validators.py
```diff
@@ -544,51 +546,18 @@ def ec2_iam_policies_validator(param_key, param_value, pcluster_config):
...
+    # Load and render instance policy file with Jinja
+    # Policy file should be consistent with what is published in documentation
+    file_loader = FileSystemLoader("./pcluster/iam_policy")
+    env = Environment(loader=file_loader, trim_blocks=True, lstrip_blocks=True)
+    policy_filename = "batch_instance_policy.json" if scheduler == "awsbatch" else "traditional_instance_policy.json"
+
+    rendered_policy = json.loads(
+        env.get_template(policy_filename).render(partition=partition, region=region, account_id=account_id)
+    )
+    return [(statement.get("Action"), statement.get("Resource")) for statement in rendered_policy.get("Statement")]
```
tilne, line 560: very clean 🐕 😎  (because i don't have the slack cool doge)

tilne (commented):
i think it's worth making the change where the JSON files go in another subdirectory of the resources directory, because i'm not sure how the relative path works for all installation scenarios.

tilne (approved):

PR #2213, opened 2020-11-10
integ-tests: test using existing EFS (efs_fs_id)

hanwen-pcluste:
This test uses `troposphere` to create cloudformation stacks for `efs`, `mount target`, and a instance to write an empty file with random name into the efs. Then the test verifies when the existing `efs` is provided through `efs_fs_id` in `pcluster` config file, the cluster created can read the randomly named file and share files between head node and compute node.

### DIFF
--- a/tests/integration-tests/tests/storage/test_efs.py
+++ b/tests/integration-tests/tests/storage/test_efs.py
@@ -14,10 +14,15 @@ import logging
 import boto3
 import pytest
 from assertpy import assert_that
+from cfn_stacks_factory import CfnStack
 from remote_command_executor import RemoteCommandExecutor
-from utils import get_vpc_snakecase_value
+from troposphere import Base64, Sub, Template
+from troposphere.ec2 import Instance
+from troposphere.efs import FileSystem, MountTarget
+from utils import generate_stack_name, get_vpc_snakecase_value, random_alphanumeric
 
 from tests.common.schedulers_common import get_scheduler_commands
+from tests.common.utils import retrieve_latest_ami
 from tests.storage.storage_common import verify_directory_correctly_shared
 
 
@@ -28,13 +33,13 @@ from tests.storage.storage_common import verify_directory_correctly_shared
 @pytest.mark.schedulers(["slurm", "awsbatch"])
 @pytest.mark.oss(["alinux2"])
 @pytest.mark.usefixtures("region", "os", "instance")
-def test_efs_compute_az(region, scheduler, pcluster_config_reader, clusters_factory, vpc_stacks):
+def test_efs_compute_az(region, scheduler, pcluster_config_reader, clusters_factory, vpc_stack):
     """
     Test when compute subnet is in a different AZ from master subnet.
 
     A compute mount target should be created and the efs correctly mounted on compute.
     """
-    _assert_subnet_az_relations(region, vpc_stacks, expected_in_same_az=False)
+    _assert_subnet_az_relations(region, vpc_stack, expected_in_same_az=False)
     mount_dir = "efs_mount_dir"
     cluster_config = pcluster_config_reader(mount_dir=mount_dir)
     cluster = clusters_factory(cluster_config)
@@ -50,13 +55,13 @@ def test_efs_compute_az(region, scheduler, pcluster_config_reader, clusters_fact
 @pytest.mark.instances(["c4.xlarge", "c5.xlarge"])
 @pytest.mark.schedulers(["slurm", "awsbatch"])
 @pytest.mark.usefixtures("region", "os", "instance")
-def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory, vpc_stacks):
+def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory, vpc_stack):
     """
     Test when compute subnet is in the same AZ as master subnet.
 
     No compute mount point needed and the efs correctly mounted on compute.
     """
-    _assert_subnet_az_relations(region, vpc_stacks, expected_in_same_az=True)
+    _assert_subnet_az_relations(region, vpc_stack, expected_in_same_az=True)
     mount_dir = "efs_mount_dir"
     cluster_config = pcluster_config_reader(mount_dir=mount_dir)
     cluster = clusters_factory(cluster_config)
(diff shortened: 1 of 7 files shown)


tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    - efs_mount_point_1=/mnt/efs/fs1
+    - mkdir -p "${efs_mount_point_1}"
+    - mount -t efs ${file_system_id_1}:/ ${efs_mount_point_1}
+    - touch ${efs_mount_point_1}/"""
+        + efs_stack.random_file_name
+    )
+    write_file_template.add_resource(
+        Instance(
+            "InstanceToWriteEFS",
+            ImageId="ami-0c2f25c1f66a1ff4d",
```
yuleiwan, line 172: I think it is a alinux2 ami. If we test alinux+awsbatch, it is better to keep consistency.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+            InstanceType="c5.xlarge",
+            SubnetId=vpc_stack.cfn_outputs["PublicSubnetId"],
+            UserData=Base64(user_data),
+            KeyName="pjenkins",
```
yuleiwan, line 176: Is this the keyname for personal jenkins? I think we should use jenkins-commercial keyname, otherwise, the stack for write_file_template will fail.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    )
+    write_file_template.add_resource(
+        MountTarget(
+            "MountTargetResource",
+            FileSystemId=efs_stack.cfn_resources["FileSystemResource"],
+            SubnetId=vpc_stack.cfn_outputs["PublicSubnetId"],
+            SecurityGroups=[default_security_group_id],
+        )
+    )
+    efs_stack.random_file_name = random_alphanumeric()
```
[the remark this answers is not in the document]
hanwen-pcluste: Is there any harm to call `random_alphanumeric()` multiple times?

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: Good catch. Done

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: Good catch. I shouldn't hardcode the `ImageId`. But I still prefer using `alinux2` , instead of choosing the `OS` according to the test. 
There are 2 reasons for doing so:
1. I am not sure if the userdata commands work for any `OS`
2. The primary purpose of this fixture is to create a efs with a file inside. The `OS` of the instance to write the file does not impair the validity of the purpose

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    write_file_template.set_description("Stack to write a file to the existing EFS")
+    default_security_group_id = (
+        boto3.client("ec2", region_name=region)
+        .describe_security_groups(
+            Filters=[
+                {"Name": "vpc-id", "Values": [vpc_stack.cfn_outputs["VpcId"]]},
+                {"Name": "group-name", "Values": ["default"]},
+            ]
+        )
+        .get("SecurityGroups")[0]
```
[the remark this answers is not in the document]
hanwen-pcluste: The default security group is created with a VPC. There is only one default security group in each VPC

tests/integration-tests/tests/storage/test_efs.py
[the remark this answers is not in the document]
yuleiwan: There is no harm and won't affect the test functionality using different random alphanumeric, but just for consistency. Moreover, If we can create all stacks in an integration stack as substacks, it would be better.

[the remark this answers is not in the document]
hanwen-pcluste: In this fixture, I am deleting `write_file_stack` while keeping `efs_stack`. I am not sure if putting them in the same stack or substacks achieve the same effect.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +74,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+            random_alphanumeric(),
+            "-" if request.config.getoption("stackname_suffix") else "",
+            request.config.getoption("stackname_suffix"),
+        ),
+        region=region,
+        template=write_file_template.to_json(),
+    )
+    cfn_stacks_factory.create_stack(write_file_stack)
+    logging.info("Sleeping for 30 seconds for the instance to execute user data")
+    time.sleep(30)
```
enrico-usai, line 189: Instead of sleeping what about checking the status of the stack and then of the instance? I don't think that 30 seconds are enough to create the stack, start the instance and execute the code.

Let's use something similar to `utils.verify_stack_creation`

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +74,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    - file_system_id_1="""
+        + efs_stack.cfn_resources["FileSystemResource"]
+        + """
+    - efs_mount_point_1=/mnt/efs/fs1
+    - mkdir -p "${efs_mount_point_1}"
+    - mount -t efs ${file_system_id_1}:/ ${efs_mount_point_1}
+    - touch ${efs_mount_point_1}/"""
+        + efs_stack.random_file_name
+    )
+    write_file_template.add_resource(
```
enrico-usai, line 167: I think the write_file_template should be defined out of the "efs_existing" fixture, to keep it self contained.
See similar comment: https://github.com/aws/aws-parallelcluster/pull/2232/files#r529363661

enrico-usai: The creation of this file is specific to the test but the "efs_existing" fixture could be more generic and can be reused for other use cases on which you need an efs.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +74,129 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    file_name = existing_efs.random_file_name
+    remote_command_executor.run_remote_command(f"cat {mount_dir}/{file_name}")
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    _test_efs_correctly_mounted(remote_command_executor, mount_dir)
+    _test_efs_correctly_shared(remote_command_executor, mount_dir, scheduler_commands)
+    remote_command_executor.run_remote_command(f"cat {mount_dir}/{file_name}")
+
+
+@pytest.fixture(scope="class")
+def existing_efs(vpc_stack, cfn_stacks_factory, request, region, key_name):
```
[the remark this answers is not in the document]
enrico-usai: Let's keep the two things aligned @yuleiwan

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: Solved by using `cfn-signal`
The 30 seconds were not for creating the instance. They were for running the `userdata`

tests/integration-tests/tests/storage/test_efs.py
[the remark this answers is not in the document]
enrico-usai: Let's use `efs_stack`

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,145 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
     _test_efs_correctly_shared(remote_command_executor, mount_dir, scheduler_commands)
 
 
+@pytest.mark.dimensions("ap-northeast-2", "c5.xlarge", "alinux", "awsbatch")
```
enrico-usai, line 76: let's use alinux2 for batch and centos8 for slurm, because we're going to deprecate alinux

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,145 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    cfn_stacks_factory.create_stack(efs_stack)
+
+    yield efs_stack
+
+    if not request.config.getoption("no_delete"):
+        cfn_stacks_factory.delete_stack(efs_stack.name, region)
+
+
+def _write_file_into_efs(region, vpc_stack, efs_stack, request, key_name, cfn_stacks_factory):
+    # write file stack contains a mount target and a instance to write a empty file with random name into the efs
```
enrico-usai, line 142: minor: let's use python triple quotes syntax for method comment.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,145 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+    remote_command_executor.run_remote_command(f"cat {mount_dir}/{random_file_name}")
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    _test_efs_correctly_mounted(remote_command_executor, mount_dir)
+    _test_efs_correctly_shared(remote_command_executor, mount_dir, scheduler_commands)
+    remote_command_executor.run_remote_command(f"cat {mount_dir}/{random_file_name}")
+
+
+@pytest.fixture(scope="class")
+def existing_efs(cfn_stacks_factory, request, region):
+    # efs stack contains a single efs resource
```
enrico-usai, line 119: minor: let's use python triple quotes syntax for method comment.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -68,6 +73,145 @@ def test_efs_same_az(region, scheduler, pcluster_config_reader, clusters_factory
...
+            ImageId=retrieve_latest_ami(region, "alinux2"),
+            InstanceType="c5.xlarge",
+            SubnetId=vpc_stack.cfn_outputs["PublicSubnetId"],
+            UserData=Base64(Sub(user_data)),
+            KeyName=key_name,
+            DependsOn=["MountTargetResource"],
+        )
+    )
+    write_file_stack = CfnStack(
+        name="integ-tests-efs-write-file-{0}{1}{2}".format(
```
enrico-usai, line 200: This code is very similar to the one in the `existing_fsx` and `existing_efs` methods. Can we put something in common, like a `_create_stack(feature_under_test, region, template)` function?

hanwen-pcluste: Done

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: Done

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: Done

enrico-usai (approved):
Great job!

PR #2216, opened 2020-11-11
Add a simple caching mechanism for functions

ddeidda:
This commit introduces a very simple caching mechanism based on a decorator that can be used to wrap functions that we want to cache.
The goal is to reduce the number of (duplicated) boto3 calls for actions like `describe_instance_types` that can be called multiple times for the same instance type.


enrico-usai:
Closed in favour of: https://github.com/aws/aws-parallelcluster/pull/2229

PR #2217, opened 2020-11-11
Support P4d instances on integration tests

ddeidda:
This commit contains some (temporary) adaptations to allow P4d instances being used in integration tests

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -462,11 +462,13 @@ AVAILABILITY_ZONE_OVERRIDES = {
     # c5.xlarge is not supported in use1-az3
     # FSx Lustre file system creation is currently not supported for use1-az3
     # m6g.xlarge is not supported in use1-az2 or use1-az3
-    "us-east-1": ["use1-az6", "use1-az1", "use1-az4", "use1-az5"],
+    # p4d.24xlarge is only available on use1-az6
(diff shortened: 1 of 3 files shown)


tests/integration-tests/conftest.py
```diff
@@ -462,11 +462,13 @@ def cfn_stacks_factory(request):
     # c5.xlarge is not supported in use1-az3
     # FSx Lustre file system creation is currently not supported for use1-az3
     # m6g.xlarge is not supported in use1-az2 or use1-az3
-    "us-east-1": ["use1-az6", "use1-az1", "use1-az4", "use1-az5"],
```
hanwen-pcluste, line 465: The [code here](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/conftest.py#L598) requires the list of availability zones to contain at least two items. Is there any easy fix?

ddeidda: Good point, but in the code above (https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/conftest.py#L569) we replicate the AZ if only one is provided. Additionally, the only test that explicitly requires two AZs is `test_efs_compute_az` which is set to run in us-west-1. Thanks anyway!

enrico-usai (approved):
LGTM

enrico-usai (approved):

PR #2224, opened 2020-11-12
Skip instance type info retrieval for compute nodes with awsbatch

demartinofra:
When scheduler is `awsbatch` we don't need to populate disable hyperthreading and network card info for compute nodes. Also compute nodes in Batch can assume the 'optimal' or an instance family value which can't be used in `describe_instance_types` calls.

This PR also addresses an issue with ParallelCluster version check in the CLI.


ddeidda (approved):
LGTM.

As per our conversation, I'd like to check together if we could handle this from a central point (`get_instance_type`) . But we can discuss it later.

PR #2226, opened 2020-11-12
Fix pcluster update with "cluster_resource_bucket"

chenwany:
Before this PR, when user does not specify bucket when create, but specifies bucket=[bucket we generate] when updating. In this case we will falsely detect that the bucket value has changed, when in reality it doesn’t.

After this PR,  we give user more visibility on what’s going on instead of just ignoring that parameter. When user provide the S3 bucket through "cluster_resource_bucket", the pcluster update will detect the bucket name, if user doesn't provide the bucket name, the pcluster update will not show the bucket name.

The following cases are tested:

-  Creation and update with same config: [cluster] not provide "cluster_resource_bucket"  --->The creation and update succeed.
- Creation and update with same config: [cluster] "cluster_resource_bucket"= "bucket1" --->The creation and update succeed.

- Creation with [cluster] not provide "cluster_resource_bucket", Update with [cluster] "cluster_resource_bucket"= "bucket1" ---> Creation succeed, update failed with:
```
#    parameter                old value    new value
---  -----------------------  -----------  ---------------
     [cluster default]
01*  cluster_resource_bucket  -            bucket1
```

- Creation with [cluster] "cluster_resource_bucket"= "bucket1" , Update with [cluster] "cluster_resource_bucket"= "bucket2"---> Creation succeed, update failed with:

```
#    parameter                old value       new value
---  -----------------------  --------------  ---------------
     [cluster default]
01*  cluster_resource_bucket  bucket1           bucket2
```

### DIFF
--- a/cli/pcluster/config/cfn_param_types.py
+++ b/cli/pcluster/config/cfn_param_types.py
@@ -1278,3 +1278,26 @@ class VolumeSizeParam(IntCfnParam):
             else:
                 default_volume_size = 20
             self.value = default_volume_size
+
+
+class ClusterResourceBucketCfnParam(CfnParam):
+    """Class to manage cluster_resource_bucket parameter."""
+
+    def from_cfn_params(self, cfn_params):
+        """
+        Initialize parameter value by parsing CFN input parameters.
+
+        :param cfn_params: list of all the CFN parameters, used if "cfn_param_mapping" is specified in the definition
+        """
+        cfn_converter = self.definition.get("cfn_param_mapping", None)
+        if cfn_params:
+            cfn_value = get_cfn_param(cfn_params, cfn_converter) if cfn_converter else "NONE"
+            self.value = (
+                None
+                if get_cfn_param(cfn_params, "RemoveBucketOnDeletion") == "True"
+                else self.get_value_from_string(cfn_value)
+            )
+            # When the value of "RemoveBucketOnDeletion" is "True",
+            # the "cluster_resource_bucket" is not created by the user.
+
+        return self
--- a/cli/pcluster/config/mappings.py
+++ b/cli/pcluster/config/mappings.py
@@ -20,6 +20,7 @@ from pcluster.config.cfn_param_types import (
     CfnSection,
     ClusterCfnSection,
     ClusterConfigMetadataCfnParam,
+    ClusterResourceBucketCfnParam,
     ComputeAvailabilityZoneCfnParam,
     DisableHyperThreadingCfnParam,
     EBSSettingsCfnParam,
@@ -996,9 +997,10 @@ CLUSTER_COMMON_PARAMS = [
         "visibility": Visibility.PRIVATE,
     }),
     ("cluster_resource_bucket", {
+        "type": ClusterResourceBucketCfnParam,
         "cfn_param_mapping": "ResourcesS3Bucket",
         "validators": [s3_bucket_validator],
-        "update_policy": UpdatePolicy.IGNORED,
+        "update_policy": UpdatePolicy.UNSUPPORTED,
     }),
 ]
 


cli/pcluster/config/cfn_param_types.py
```diff
@@ -55,7 +55,14 @@ def from_cfn_params(self, cfn_params):
         cfn_converter = self.definition.get("cfn_param_mapping", None)
         if cfn_params:
             cfn_value = get_cfn_param(cfn_params, cfn_converter) if cfn_converter else "NONE"
-            self.value = self.get_value_from_string(cfn_value)
+            self.value = (
+                None
+                if self.key == "cluster_resource_bucket"
+                and get_cfn_param(cfn_params, "RemoveBucketOnDeletion") == "True"
```
rexcsn, line 61: Minor: is there a relatively cleaner way to write this logic? For example we can extract this logic into a function `handle_special_param` or something?

cli/pcluster/config/cfn_param_types.py
```diff
@@ -55,7 +55,14 @@ def from_cfn_params(self, cfn_params):
...
+                else self.get_value_from_string(cfn_value)
+            )
+            # When the value of "RemoveBucketOnDeletion" is "True", the "cluster_resource_bucket" is not created by the
+            # user.
```
rexcsn, line 65: Minor: it's a bit odd having this comment breaking like this; this line only has 1 word. Try breaking at the comma, maybe?

rexcsn (commented):
Nice work, would be nice if you can add a brief description of the logic we are implementing in the commit message, and add the examples of update outputs like we discussed offline in the PR description.

cli/pcluster/config/cfn_param_types.py
chenwany: For creating a function `handle_special_param`, different special parameters will have different logic for this `if` and `else` here. I am not sure how to extract this logic.
I am thinking about another way: have another param class for 'cluster_resource_bucket', inherited from CfnParam, rewrite the method `def from_cfn_params`, but it is also not a clean way

cli/pcluster/config/cfn_param_types.py
chenwany: Yeah! Done!

enrico-usai (approved):

cli/pcluster/config/cfn_param_types.py
```diff
@@ -1278,3 +1278,26 @@ def refresh(self):
...
+        """
+        Initialize parameter value by parsing CFN input parameters.
+
+        :param cfn_params: list of all the CFN parameters, used if "cfn_param_mapping" is specified in the definition
+        """
+        cfn_converter = self.definition.get("cfn_param_mapping", None)
+        if cfn_params:
+            cfn_value = get_cfn_param(cfn_params, cfn_converter) if cfn_converter else "NONE"
+            self.value = (
+                None
```
enrico-usai, line 1296: The entire code could be simplified with something like:
```python
if cfn_params and get_cfn_param(cfn_params, "RemoveBucketOnDeletion") != "True":
    # When the value of "RemoveBucketOnDeletion" is not "True" the "cluster_resource_bucket" is created by the user.
    cfn_value = get_cfn_param(cfn_params, cfn_converter)
    self.value = self.get_value_from_string(cfn_value)
```

enrico-usai: We can avoid the check for the `cfn_converter` because we already know that this parameter has the `cfn_param_mapping` attribute in the mappings. --> we can update the method comment too.

enrico-usai (commented):
Great idea, I just added some minor comments.

rexcsn:
Closed in favor of https://github.com/aws/aws-parallelcluster/pull/2235

PR #2227, opened 2020-11-12
configure: Don't permit to use p4d instance type as head node

hanwen-pcluste:
Furthermore, this commit will reduce one boto3 call during pcluster configure. There were two validators for each of head and compute instance types. The reason to have the two validators was that one of them retrieved instance types from boto3 call, the other one retrieved instance types from instance.json. After this PR https://github.com/aws/aws-parallelcluster/pull/2052/, both of the two validators are retrieving instance types from boto3, so only one is needed.


rexcsn (commented):
Nicely done, would be nice if we can add a unit test for configure to test that p4d cannot be used.

rexcsn (dismissed):

cli/pcluster/configure/easyconfig.py
```diff
@@ -324,6 +325,13 @@ def _prompt_for_subnet(default_subnet, all_subnets, qualified_subnets, message):
     return prompt_iterable(message, qualified_subnets, default_value=default_subnet)
 
 
+def _is_instance_type_supported_for_head_node(instance_type):
+    if instance_type in HEADNODE_UNSUPPORTED_INSTANCE_TYPES:
+        print("The instance type '{0}' is not supported as head node.".format(instance_type))
```
enrico-usai, line 330: we should define a variable like HEAD_NODE_MESSAGES in the validators.py to be used in both the places

cli/pcluster/configure/easyconfig.py
```diff
@@ -360,21 +368,15 @@ def prompt_os(self):
...
-        ec2_client = boto3.client("ec2")
-        instance_type_offerings = [
-            offering["InstanceType"]
-            for offering in ec2_client.describe_instance_type_offerings()["InstanceTypeOfferings"]
-        ]
-
         self.master_instance_type = prompt(
             "Master instance type",
-            lambda x: x in get_supported_instance_types() and x in instance_type_offerings,
+            lambda x: x in get_supported_instance_types() and _is_instance_type_supported_for_head_node(x),
```
enrico-usai, line 373: I'd call the functions in the opposite order, because the `_is_instance_type_supported_for_head_node`is a static validation, the other one is a boto3 call, in this way we can fail fast.

cli/pcluster/configure/easyconfig.py
```diff
@@ -14,6 +14,7 @@
 
 from pcluster.cluster_model import ClusterModel
 from pcluster.config.hit_converter import HitConverter
+from pcluster.config.validators import HEADNODE_UNSUPPORTED_INSTANCE_TYPES
```
enrico-usai, line 17: Not related to your patch but could you please rename HEADNODE_UNSUPPORTED... to HEAD_NODE_UNSUPPORTED  to be aligned with the naming convention we're using elsewhere.

enrico-usai (approved):

PR #2228, opened 2020-11-13
Add integration test for cluster_resource_bucket

rexcsn:
* Add creation/deletion test for cluster_resource_bucket parameter
* Modified update test to use cluster_resource_bucket for test_update_hit

For reference, integration test coverage is as followed:
* `test_resource_bucket`: new test, test creation with specified bucket, update with different specified bucket. Has specific assertion to test that bucket outputs, parameters, and deletion behaviors are expected
* `test_update_sit`: no change to existing test, test creation and update with no specified bucket, and update is not forced, which covers that the default use case is always working
* `test_update_hit`: changed test, test creation with specified bucket and update with same specified bucket, update is forced, tests that HIT artifacts which heavily relies on the bucket are working
* Other cases of `resource_bucket` updates are covered by unit tests in this patch: https://github.com/aws/aws-parallelcluster/pull/2239

### DIFF
--- a/tests/integration-tests/tests/update/test_update.py
+++ b/tests/integration-tests/tests/update/test_update.py
@@ -42,7 +42,7 @@ def test_update_sit(
     bucket.upload_file(str(test_datadir / "postinstall.sh"), "scripts/postinstall.sh")
 
     # Create cluster with initial configuration
-    init_config_file = pcluster_config_reader()
+    init_config_file = pcluster_config_reader(resource_bucket=bucket_name)
     cluster = clusters_factory(init_config_file)
 
     # Update cluster with the same configuration, command should not result any error even if not using force update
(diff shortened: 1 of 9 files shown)


tests/integration-tests/utils.py
```diff
@@ -208,6 +210,7 @@ def delete_s3_bucket(bucket_name, region):
     try:
         bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
         bucket.objects.all().delete()
+        bucket.object_versions.all().delete()
```
tilne, line 213: Needed because we're turning on versioning in `create_s3_bucket` in the same file?

tilne (approved):
LGTM.

Do we have a test anywhere for the case where a bucket is created by the CLI when creating cluster and then is updated?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -148,7 +148,7 @@ def test_update_hit(region, scheduler, pcluster_config_reader, clusters_factory,
     bucket.upload_file(str(test_datadir / "postinstall.sh"), "scripts/postinstall.sh")
 
     # Create cluster with initial configuration
-    init_config_file = pcluster_config_reader()
+    init_config_file = pcluster_config_reader(resource_bucket=bucket_name)
```
ddeidda, line 151: Just out of curiosity: why are we passing the resource bucket here? What are you testing here that we didn't in the other test?

ddeidda (approved):
Everything looks clear to me :+1: 
I dropped a small question just to better understand the goal of a change but nothing blocking.

tests/integration-tests/tests/update/test_update.py
rexcsn: Updated PR description to explain test coverage

PR #2229, opened 2020-11-13
Optimize calls to describe_instance_types API

demartinofra:
Implement a caching mechanism to cache multiple calls to descrive_instance_types API


cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -1008,3 +967,129 @@ def test_get_ebs_snapshot_info(boto3_stubber, snapshot_id, raise_exceptions, err
...
+
+    def test_cached_method(self):
+        for _ in range(0, 2):
+            assert_that(self._cached_method_1(1, 2)).is_equal_to((1, 2))
+            assert_that(self._cached_method_2(1, 2)).is_equal_to((1, 2))
+            assert_that(self._cached_method_1(2, 1)).is_equal_to((2, 1))
+            assert_that(self._cached_method_1(1, arg2=2)).is_equal_to((1, 2))
+            assert_that(self._cached_method_1(arg1=1, arg2=2)).is_equal_to((1, 2))
+
+        assert_that(self.invocations).is_length(5)
```
enrico-usai, line 1003: what about checking the content or the size of the `Cache.caches` ?

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -1008,3 +967,129 @@ def test_get_ebs_snapshot_info(boto3_stubber, snapshot_id, raise_exceptions, err
...
+
+    def test_disabled_cache(self, disabled_cache):
+        assert_that(self._cached_method_1(1, 2)).is_equal_to((1, 2))
+        assert_that(self._cached_method_1(1, 2)).is_equal_to((1, 2))
+
+        assert_that(self.invocations).is_length(2)
```
enrico-usai, line 1009: what about checking the content or the size of the `Cache.caches` ?

cli/tests/pcluster/utils/test_pcluster_utils.py
```diff
@@ -1008,3 +967,129 @@ def test_get_ebs_snapshot_info(boto3_stubber, snapshot_id, raise_exceptions, err
...
+
+    def test_clear_all(self):
+        for _ in range(0, 2):
+            assert_that(self._cached_method_1(1, 2)).is_equal_to((1, 2))
+            assert_that(self._cached_method_2(1, 2)).is_equal_to((1, 2))
+
+        utils.Cache.clear_all()
```
enrico-usai, line 1016: Same here

cli/pcluster/utils.py
```diff
@@ -1225,25 +1170,122 @@ def get_ebs_snapshot_info(ebs_snapshot_id, raise_exceptions=False):
...
+        def wrapper(*args, **kwargs):
+            cache_key = Cache._make_key(args, kwargs)
 
-    # Until maximumNetworkCards is not available, 1 is a safe value for all instance types
-    needed_interfaces = int(instance_info.get("NetworkInfo").get("MaximumNetworkCards", 1))
+            if Cache.enabled() and cache_key in cache:
+                return cache[cache_key]
+            else:
+                rv = function(*args, **kwargs)
+                if Cache.enabled() and rv:
```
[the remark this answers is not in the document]
demartinofra: good catch! actually I should not be checking rv at all here.

cli/tests/pcluster/utils/test_pcluster_utils.py
demartinofra: Two reasons why I prefer not to do that:
1. caches is part of the internal implementation of the caching mechanism not something the class exposes (actually let me mark it as such). Having a unit test depend on the internal implementation of the caching mechanism makes the test invalid even if the class interface does not change.
2. new items are added to the caches array when the annotated function is loaded and not when it is executed for the first time, which means the number to assert depends on the number of imported functions that are annotated with `@Cache.cached`

cli/pcluster/utils.py
```diff
@@ -1225,25 +1170,122 @@ def get_ebs_snapshot_info(ebs_snapshot_id, raise_exceptions=False):
...
-        instance_info = get_instance_type(instance_type)
+class InstanceTypeInfo:
+    """Data object wrapping the result of a describe_instance_types call."""
+
+    def __init__(self, instance_type_data):
+        self.instance_type_data = instance_type_data
+
+    @staticmethod
+    @Cache.cached
+    def init_from_instance_type(instance_type, exit_on_error=True):
```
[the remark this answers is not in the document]
demartinofra: I would not generalize it to init cause it would be harder to add multiple factories if needed. I can try to pick a better name if that is not good enough.

demartinofra:
This is not ready because there are some issues with Python 2.7 that need to be addressed

demartinofra:
Ready for review!

enrico-usai (approved):
LGTM!

PR #2233, opened 2020-11-13
integ-tests: test FSx deployment types

yuleiwan:
* This test is to verify FSx file system launched by pcluster has the correct deployment type as user set in pcluster config file.
* FSx file system has three deployment types in commercial regions SCRATCH_1(dafault), SCRATCH_2, PERSISTENT_1 and one deployment type in China PERSISTENT_1

### DIFF
--- a/tests/integration-tests/tests/storage/test_fsx_lustre.py
+++ b/tests/integration-tests/tests/storage/test_fsx_lustre.py
@@ -86,14 +86,16 @@ def test_fsx_lustre_configuration_options(
         storage_capacity=storage_capacity,
     )
     cluster = clusters_factory(cluster_config)
-    _test_fsx_lustre(cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy)
+    _test_fsx_lustre(
+        cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy, deployment_type
+    )
 
 
 @pytest.mark.regions(["eu-west-1"])
 @pytest.mark.instances(["c5.xlarge", "m6g.xlarge"])
 @pytest.mark.schedulers(["slurm"])
 @pytest.mark.usefixtures("instance")
-# FSx is only supported on ARM instances for Ubuntu 18.04 and Amazon Linux 2
+# FSx is only supported on ARM instances for Ubuntu 18.04, Amazon Linux 2 and CentOS 8
 @pytest.mark.skip_dimensions("*", "m6g.xlarge", "alinux", "*")
 @pytest.mark.skip_dimensions("*", "m6g.xlarge", "centos7", "*")
 @pytest.mark.skip_dimensions("*", "m6g.xlarge", "ubuntu1604", "*")
@@ -121,10 +123,22 @@ def test_fsx_lustre(
         storage_capacity=1200,
     )
     cluster = clusters_factory(cluster_config)
-    _test_fsx_lustre(cluster, region, scheduler, os, mount_dir, bucket_name, storage_type=None, auto_import_policy=None)
+    _test_fsx_lustre(
+        cluster,
+        region,
+        scheduler,
+        os,
+        mount_dir,
+        bucket_name,
+        storage_type=None,
+        auto_import_policy=None,
+        deployment_type=None,
+    )
 
 
-def _test_fsx_lustre(cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy):
+def _test_fsx_lustre(
+    cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy, deployment_type
+):
     remote_command_executor = RemoteCommandExecutor(cluster)
     scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
     fsx_fs_id = get_fsx_fs_id(cluster, region)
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -266,6 +281,26 @@ def _test_storage_type(storage_type, fsx_fs_id, region):
...
+    )
+
+
+def _test_deployment_type(deployment_type, fsx_fs_id, region):
+    if deployment_type == "PERSISTENT_1":
+        assert_that(get_deployment_type(fsx_fs_id, region)).is_equal_to("PERSISTENT_1")
+    elif deployment_type == "SCRATCH_2":
+        assert_that(get_deployment_type(fsx_fs_id, region)).is_equal_to("SCRATCH_2")
+    else:
+        assert_that(get_deployment_type(fsx_fs_id, region)).is_equal_to("SCRATCH_1")
```
hanwen-pcluste, line 301: minor: the code could be simplified as
```
if not deployment_type:
    deployment_type = "SCRATCH_1"
assert_that(get_deployment_type(fsx_fs_id, region)).is_equal_to(deployment_type)
```

chenwany: minor: it can  also be simplified as one sentence:
assert_that(get_deployment_type(fsx_fs_id, region)).is_equal_to(deployment_type if deployment_type else "SCRATCH_1")
Maybe we can also add a comment to explain that the default deployment_type for the cloudformation is "SCRATCH_1"

yuleiwan: Done!

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -266,6 +281,22 @@ def _test_storage_type(storage_type, fsx_fs_id, region):
...
+    return (
+        fsx.describe_file_systems(FileSystemIds=[fsx_fs_id])
+        .get("FileSystems")[0]
+        .get("LustreConfiguration")
+        .get("DeploymentType")
+    )
+
+
+def _test_deployment_type(deployment_type, fsx_fs_id, region):
+    # The default deployment type is SCRATCH_1
```
enrico-usai, line 296: I don't think it's always true. 
In china AFAIK the only deployment_type supported in China regions is PERSISTENT_1.

If we don't know what deployment_type will be used, because empty, we can skip the check.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -266,6 +281,22 @@ def _test_storage_type(storage_type, fsx_fs_id, region):
         assert_that(get_storage_type(fsx_fs_id, region)).is_equal_to("SSD")
 
 
+def get_deployment_type(fsx_fs_id, region):
```
enrico-usai, line 284: internal convention: when a function is "private" for the module we're adding an underscore as prefix.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -266,6 +281,22 @@ def _test_storage_type(storage_type, fsx_fs_id, region):
...
+    logging.info("Getting DeploymentType from DescribeFilesystem API.")
```
enrico-usai, line 285: What about logging this message when we have the value and log the deployment type too?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -86,7 +86,9 @@ def test_fsx_lustre_configuration_options(
         storage_capacity=storage_capacity,
     )
     cluster = clusters_factory(cluster_config)
-    _test_fsx_lustre(cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy)
+    _test_fsx_lustre(
+        cluster, region, scheduler, os, mount_dir, bucket_name, storage_type, auto_import_policy, deployment_type
+    )
```
enrico-usai, line 91: not related to your change but could you please update the comment in this file (it's in 2 places):
`# FSx is only supported on ARM instances for Ubuntu 18.04 and Amazon Linux 2`
--> 
`# FSx is only supported on ARM instances for Ubuntu 18.04, Amazon Linux 2 and CentOS 8`

enrico-usai (commented):
From what we understand we were already testing all the deployment_types (through test_fsx_lustre_configuration_options) but we weren't checking if the value was correct. Right?

If so let's add this description in the PR.

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done, but as I know, we only test deployment type in commercial region, and the default type is SCRATCH_1, should we test it?

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

enrico-usai (approved):
Great job! Thanks!

If tested feel free to merge it.

PR #2239, opened 2020-11-16
 Make update behavior of cluster_resource_bucket consistent

ddeidda:
* Changes to `cluster_resource_bucket` when running update will never be applied. i.e. value of cfn param `ResourcesS3Bucket` will never change when updating. This behavior is checked by integration test
* If user does not specify `cluster_resource_bucket` both when create and update(using default pcluster behaviors), do not block update or print diff
* Otherwise, block update if `cluster_resource_bucket` is changed and print message to inform user value of `cluster_resource_bucket` will not be changed even if force update
* Add unit test to check the above behavior

Update behavior after this patch:
* No `cluster_resource_bucket` when create, no `cluster_resource_bucket` when update:
```
Retrieving configuration from CloudFormation for cluster test-no-bucket...
Validating configuration file ...
No changes found in your cluster configuration.
Do you want to proceed with the update? - Y/N: 
```
* No `cluster_resource_bucket` when create, specifies different `cluster_resource_bucket` when update:
```
Retrieving configuration from CloudFormation for cluster no-s3...
Validating configuration file ...
Found Configuration Changes:

#    parameter                old value                       new value
---  -----------------------  ------------------------------  -----------------
     [cluster no-s3]
01*  cluster_resource_bucket  parallelcluster-nwh0pdfdbbi...  new-bucket-name

Validating configuration update...
The requested update cannot be performed. Line numbers with an asterisk indicate updates requiring additional actions. Please review the details below:

#01
'cluster_resource_bucket' parameter is a read_only parameter that cannot be updated. New value 'new-bucket-name' will be ignored and old value 'parallelcluster-nwh0pdfdbbi1pavp' will be used if you force the update.
How to fix:
Restore the value of parameter 'cluster_resource_bucket' to 'parallelcluster-nwh0pdfdbbi1pavp'
```
* Specifies `cluster_resource_bucket` when create, different `cluster_resource_bucket` when update:
```
Retrieving configuration from CloudFormation for cluster s3...
Validating configuration file ...
Found Configuration Changes:

#    parameter                old value          new value
---  -----------------------  -----------------  -----------
     [cluster s3]
01*  cluster_resource_bucket  old-bucket-name  -

Validating configuration update...
The requested update cannot be performed. Line numbers with an asterisk indicate updates requiring additional actions. Please review the details below:

#01
'cluster_resource_bucket' parameter is a read_only parameter that cannot be updated. New value 'None' will be ignored and old value 'old-bucket-name' will be used if you force the update.
How to fix:
Restore the value of parameter 'cluster_resource_bucket' to 'old-bucket-name'
```

### DIFF
--- a/cli/pcluster/config/config_patch.py
+++ b/cli/pcluster/config/config_patch.py
@@ -232,7 +232,10 @@ class ConfigPatch(object):
 
                 if check_result != UpdatePolicy.CheckResult.SUCCEEDED:
                     patch_allowed = False
-
+                if change.update_policy == UpdatePolicy.READ_ONLY_RESOURCE_BUCKET and patch_allowed:
+                    # Special case for cluster_resource_bucket/ResourcesS3Bucket
+                    # Do not append diff if patch is allowed
+                    continue
                 rows.append(
                     [
                         section_name,
--- a/cli/pcluster/config/config_patch.py
+++ b/cli/pcluster/config/config_patch.py
@@ -225,17 +225,12 @@ class ConfigPatch(object):
         patch_allowed = True
 
         for change in self.changes:
-            if change.update_policy != UpdatePolicy.IGNORED:
+            check_result, reason, action_needed, print_change = change.update_policy.check(change, self)
+            if print_change:
                 section_name = get_file_section_name(change.section_key, change.section_label)
-
-                check_result, reason, action_needed = change.update_policy.check(change, self)
 
                 if check_result != UpdatePolicy.CheckResult.SUCCEEDED:
                     patch_allowed = False
-                if change.update_policy == UpdatePolicy.READ_ONLY_RESOURCE_BUCKET and patch_allowed:
-                    # Special case for cluster_resource_bucket/ResourcesS3Bucket
-                    # Do not append diff if patch is allowed
-                    continue
                 rows.append(
                     [
                         section_name,
--- a/cli/pcluster/config/config_patch.py
+++ b/cli/pcluster/config/config_patch.py
@@ -226,11 +226,12 @@ class ConfigPatch(object):
 
         for change in self.changes:
             check_result, reason, action_needed, print_change = change.update_policy.check(change, self)
+
+            if check_result != UpdatePolicy.CheckResult.SUCCEEDED:
+                patch_allowed = False
+
             if print_change:
                 section_name = get_file_section_name(change.section_key, change.section_label)
-
-                if check_result != UpdatePolicy.CheckResult.SUCCEEDED:
-                    patch_allowed = False
                 rows.append(
                     [
                         section_name,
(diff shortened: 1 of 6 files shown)


cli/pcluster/config/config_patch.py
```diff
@@ -225,17 +225,12 @@ def check(self):
         patch_allowed = True
 
         for change in self.changes:
-            if change.update_policy != UpdatePolicy.IGNORED:
+            check_result, reason, action_needed, print_change = change.update_policy.check(change, self)
+            if print_change:
```
rexcsn, line 229: Should we put this change right before `rows.append(`? As it is I think this would imply if we do not print change the `patch_allow` is always True?

cli/pcluster/config/update_policy.py
```diff
@@ -120,10 +132,41 @@ def _check_min_count(change, patch):
...
+
+
+def _check_generated_bucket(change, patch):
+
+    # If bucket is generated (no cluster_resource_bucket specified when creating) and no change in config
+    # Old value is retrieve from CFN's ResourcesS3Bucket param, and new value is from config(not specified)
+    # Diff based on config should be:
+    #
+    # old value                         new value
+    # ---------------------------------------------
```
enrico-usai, line 147: minor: I'd remove all these comments related to the visualization layer in this function.

cli/pcluster/config/update_policy.py
```diff
@@ -120,10 +132,41 @@ def _check_min_count(change, patch):
...
+    # {ResourcesS3Bucket}               -
+    #
+    # Actual ResourcesS3Bucket will not change when updating, so no actual change when update is applied
+    # Since there is no change, user should not be informed of a change/diff
+    # Print no diff and proceed with updating other parameters
+    if _is_bucket_pcluster_generated(patch.stack_name) and (not change.new_value or change.new_value == "NONE"):
```
enrico-usai, line 153: Why are we allowing NONE as value? --> we should prevent this value in the s3_bucket_validator

enrico-usai, line 153: minor: you can return without the if
```python
return  _is_bucket_pcluster_generated(patch.stack_name) and (not ...)
```

cli/pcluster/config/config_patch.py
ddeidda: Yes, if no changes are found to be displayed the patch should be allowed. In that case we'd have only ALLOWED, IGNORED and/or READ_ONLY not blocking changes. Do you think I am missing anything?

rexcsn: Ok, so I think this logic makes sense with what we are implementing today, but it is a bit weird that we would couple `print_change` with `patch_allowed`, when I think `patch_allowed ` should only be associated with `check_result`. It's a minor thing, but may be worth doing just to avoid confusions in the future

enrico-usai (approved):

PR #2248, opened 2020-11-17
Fix pcluster configure internal refresh flow

enrico-usai:
Before this patch the code was loading the configuration file by initializing `PclusterConfig` with `auto_refresh=True`. 
This setting means that the internal parameters must be recalculated every time there is a change in one of the internal sections.

The issue was that the calculation of internal parameters was using the values from the initial config file and after that, the `pcluster configure` workflow wasn't changing internal sections (parameters only) so the auto-refresh wasn't triggered anymore.

For this reason there was a piece of code to explicitly reload the config file again after changing the region. 
We don't need it anymore because I'm instead explicitly calling the `refresh` method.

Then, in the `no_awsbatch_*_with_config_file` test we were wrongly expecting the `AWSBatchFullAccess` policy with `sge` scheduler(!). The test was clearly wrong and was hiding a wrong behaviour.

Now we're executing the `refresh` command right after collecting all the inputs from the users, so the `additional_iam_policies` will contain the values related to the right scheduler (`sge`) and not the original one (`awsbatch`).

We're also setting the `auto_refresh=True` before calling the `HitConverter.convert` because this method can change some internal sections so the internal parameters must be updated accordingly.

This commit also fixes cases on which `refresh` method was called with default/initial values (e.g. `t2.micro` as instance type in regions on which this instance doesn't exist).
Right now we're instead validating the values coming from the user.

## Tests

Before:
```
$ pcluster configure
ERROR: Error when calling DescribeInstanceTypes for instances p4d.24xlarge: The following supplied instance types do not exist: [p4d.24xlarge]
```

After:
```
$ pcluster configure

Allowed values for AWS Region ID:
1. ...
AWS Region ID [eu-west-1]:
Allowed values for EC2 Key Pair Name:
1. ...
EC2 Key Pair Name [test-key]:
Allowed values for Scheduler:
1. ...
Scheduler [sge]:
Allowed values for Operating System:
1. ...
Operating System [centos7]:
Minimum cluster size (instances) [1]:
Maximum cluster size (instances) [10]:
Master instance type [p4d.24xlarge]:
The instance type 'p4d.24xlarge' is not supported as head node.
ERROR: p4d.24xlarge is not an acceptable value for Master instance type
Master instance type [p4d.24xlarge]:
```

I also added a unit test to verify that we're not using default/initial values for internal calls when executing `pcluster configure`.

### DIFF
--- a/cli/pcluster/configure/easyconfig.py
+++ b/cli/pcluster/configure/easyconfig.py
@@ -118,7 +118,7 @@ def configure(args):
     if args.config_file and os.path.exists(args.config_file) and not os.path.isfile(args.config_file):
         error("Invalid configuration file path: {0}".format(args.config_file))
 
-    pcluster_config = PclusterConfig(config_file=args.config_file, fail_on_error=False)
+    pcluster_config = PclusterConfig(config_file=args.config_file, fail_on_error=False, auto_refresh=False)
 
     # FIXME: Overriding HIT config files is currently not supported.
     if pcluster_config.cluster_model == ClusterModel.HIT:
@@ -141,10 +141,7 @@ def configure(args):
         default_region = pcluster_config.get_section("aws").get_param_value("aws_region_name")
         aws_region_name = prompt_iterable("AWS Region ID", available_regions, default_value=default_region)
         # Set provided region into os environment for suggestions and validations from here on
-        if os.environ["AWS_DEFAULT_REGION"] != aws_region_name:
-            os.environ["AWS_DEFAULT_REGION"] = aws_region_name
-            # Read config file again, because region change can cause change of the initial values in PclusterConfig
-            pcluster_config = PclusterConfig(config_file=args.config_file, fail_on_error=False)
+        os.environ["AWS_DEFAULT_REGION"] = aws_region_name
     else:
         aws_region_name = args.region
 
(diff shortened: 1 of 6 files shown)


cli/pcluster/configure/easyconfig.py
```diff
@@ -143,8 +143,6 @@ def configure(args):
         # Set provided region into os environment for suggestions and validations from here on
         if os.environ["AWS_DEFAULT_REGION"] != aws_region_name:
             os.environ["AWS_DEFAULT_REGION"] = aws_region_name
```
hanwen-pcluste, line 145: minor: this conditional statement could be simplified, because we are not reading the config file again.
Only `os.environ["AWS_DEFAULT_REGION"] = aws_region_name` would work

hanwen-pcluste (commented):
I like this patch.
One more question: in `mappings.py`, should we change the default of instance type to t3.micro, which is another free tier instance type and available in all regions (including govcloud and china) that I know.

cli/pcluster/configure/easyconfig.py
enrico-usai: Done, thanks!

enrico-usai:
> I like this patch.
> One more question: in `mappings.py`, should we change the default of instance type to t3.micro, which is another free tier instance type and available in all regions (including govcloud and china) that I know.

It's a good suggestion, we could change it but it's an important change for the customers that requires a discussion.

hanwen-pcluste (approved):

rexcsn (approved):
Really clean, love it!

PR #2254, opened 2020-11-20
Set default instance type to free tier instance type

hanwen-pcluste:
When running `pcluster` in a region with free tier, default instance type is set to the free tier instance type. When running `pcluster` in the China (BJS) region or AWS GovCloud (US) regions, default instance type is t3.micro.
Free tier is not available in the China (BJS) region and AWS GovCloud (US) regions.
For more information about free tier, please see https://aws.amazon.com/free/free-tier-faqs/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ CHANGELOG
 
 - Pull Amazon Linux Docker images from ECR when building docker image for `awsbatch` scheduler. This only applies to
   images built for `x86` architecture.
+- Change the default of instance types from the hardcoded `t2.micro` to the free tier instance type 
+  (`t2.micro` or `t3.micro` dependent on region). In regions without free tier, the default is `t3.micro`.
+
 
 **BUG FIXES**
 
--- a/cli/src/pcluster/config/cfn_param_types.py
+++ b/cli/src/pcluster/config/cfn_param_types.py
@@ -24,6 +24,7 @@ from pcluster.utils import (
     error,
     get_availability_zone_of_subnet,
     get_cfn_param,
+    get_default_instance_type,
     get_ebs_snapshot_info,
     get_efs_mount_target_id,
     get_file_section_name,
@@ -852,6 +853,34 @@ class ArgsCfnParam(CfnParam):
             cfn_params[cfn_converter] = json.dumps(cfn_value)[1:-1]
 
         return cfn_params
+
+
+class ComputeInstanceTypeCfnParam(CfnParam):
+    """
+    Class to manage the compute instance type parameter.
+
+    We need this class in order to set the default instance type from a boto3 call.
+    """
+
+    def refresh(self):
+        """Get default value from a boto3 call for free tier instance type."""
+        if not self.value:
+            scheduler = self.pcluster_config.get_section("cluster").get_param_value("scheduler")
+            if scheduler:
+                self.value = "optimal" if scheduler == "awsbatch" else get_default_instance_type()
+
+
+class MasterInstanceTypeCfnParam(CfnParam):
+    """
+    Class to manage the head node instance type parameter.
+
+    We need this class in order to set the default instance type from a boto3 call.
+    """
+
+    def refresh(self):
+        """Get default value from a boto3 call for free tier instance type."""
+        if not self.value:
+            self.value = get_default_instance_type()
 
 
 class TagsParam(JsonCfnParam):
--- a/cli/src/pcluster/config/mappings.py
+++ b/cli/src/pcluster/config/mappings.py
@@ -21,6 +21,7 @@ from pcluster.config.cfn_param_types import (
     ClusterCfnSection,
     ClusterConfigMetadataCfnParam,
     ComputeAvailabilityZoneCfnParam,
+    ComputeInstanceTypeCfnParam,
     DisableHyperThreadingCfnParam,
     EBSSettingsCfnParam,
     EFSCfnSection,
@@ -29,6 +30,7 @@ from pcluster.config.cfn_param_types import (
     IntCfnParam,
     MaintainInitialSizeCfnParam,
     MasterAvailabilityZoneCfnParam,
+    MasterInstanceTypeCfnParam,
     NetworkInterfacesCountCfnParam,
     QueueSizeCfnParam,
     SettingsCfnParam,
@@ -753,14 +755,6 @@ CLUSTER_COMMON_PARAMS = [
         "validators": [ec2_key_pair_validator],
         "update_policy": UpdatePolicy.UNSUPPORTED
     }),
-    ("base_os", {
-        "type": BaseOSCfnParam,
-        "cfn_param_mapping": "BaseOS",
-        "allowed_values": ["alinux", "alinux2", "ubuntu1604", "ubuntu1804", "centos7", "centos8"],
-        "validators": [base_os_validator, architecture_os_validator],
-        "required": True,
-        "update_policy": UpdatePolicy.UNSUPPORTED
-    }),
     ("scheduler", {
         "cfn_param_mapping": "Scheduler",
         "allowed_values": ["awsbatch", "sge", "slurm", "torque"],
@@ -770,7 +764,7 @@ CLUSTER_COMMON_PARAMS = [
     }),
     # Master
     ("master_instance_type", {
-        "default": "t2.micro",
+        "type": MasterInstanceTypeCfnParam,
         "cfn_param_mapping": "MasterInstanceType",
         "validators": [head_node_instance_type_validator, ec2_instance_type_validator],
         "update_policy": UpdatePolicy.UNSUPPORTED,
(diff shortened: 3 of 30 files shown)


cli/src/pcluster/config/mappings.py
```diff
@@ -1025,9 +1026,10 @@
             }),
             # Compute fleet
             ("compute_instance_type", {
+                "type": InstanceTypeCfnParam,
                 "default":
                     lambda section:
-                    "optimal" if section and section.get_param_value("scheduler") == "awsbatch" else "t2.micro",
+                    "optimal" if section and section.get_param_value("scheduler") == "awsbatch" else None,
```
ddeidda, line 1032: I think also the logic for the "optimal" value could be moved into the CfnParam. Of course we'd need a `ComputeInstanceTypeCfnParam` and a `MasterInstanceTypeCfnParam`.

cli/src/pcluster/config/param_types.py
```diff
@@ -441,7 +442,7 @@ def __init__(self, section_definition, pcluster_config, section_label=None, pare
         self.parent_section = parent_section
 
         # initialize section parameters with default values
-        self.params = {}
+        self.params = OrderedDict({})
```
ddeidda, line 445: Good one.

cli/src/pcluster/configure/easyconfig.py
```diff
@@ -367,16 +368,22 @@ def prompt_os(self):
 
     def prompt_instance_types(self):
         """Ask for master_instance_type and compute_instance_type (if necessary)."""
+        default_master_instance_type = self.cluster_section.get_param_value("master_instance_type")
+        if not default_master_instance_type:
+            default_master_instance_type = get_default_instance_type()
```
ddeidda, line 373: This shouldn't be needed if `pcluster_config` is loaded with `auto_refresh=True` (default value)

cli/src/pcluster/configure/easyconfig.py
```diff
@@ -367,16 +368,22 @@ def prompt_os(self):
...
         self.master_instance_type = prompt(
             "Master instance type",
             lambda x: _is_instance_type_supported_for_head_node(x) and x in get_supported_instance_types(),
-            default_value=self.cluster_section.get_param_value("master_instance_type"),
+            default_value=default_master_instance_type,
```
ddeidda, line 377: If pcluster_config is loaded with autorefresh we could directly put `self.cluster_section.get_param_value("master_instance_type")`

cli/src/pcluster/configure/easyconfig.py
```diff
@@ -367,16 +368,22 @@ def prompt_os(self):
...
         )
         if not self.is_aws_batch:
+            default_compute_instance_type = self.cluster_section.get_param_value("compute_instance_type")
+            if not default_compute_instance_type:
+                default_compute_instance_type = get_default_instance_type()
```
ddeidda, line 382: Same as above

cli/src/pcluster/utils.py
```diff
@@ -1247,3 +1247,25 @@ def get_instance_gpus(instance_type, instance_info=None):
...
+        ec2_client = boto3.client("ec2")
+        paginator = ec2_client.get_paginator("describe_instance_types")
+        page_iterator = paginator.paginate(
+            Filters=[
+                {"Name": "free-tier-eligible", "Values": ["true"]},
+                {"Name": "current-generation", "Values": ["true"]},
+            ]
+        )
+        free_tier_instance_type = []
+        for page in page_iterator:
```
ddeidda, line 1268: Why are we iterating all the results if we need just the first one?

cli/tests/pcluster/config/test_utils.py
```diff
@@ -34,3 +36,35 @@ def test_get_instance_vcpus(boto3_stubber, valid_instance_type, expected_vcpus):
 
     boto3_stubber("ec2", mocked_requests)
     assert_that(get_instance_vcpus(instance_type)).is_equal_to(expected_vcpus)
+
+
+@pytest.mark.parametrize(
+    "region, free_tier_instance_type, default_instance_type",
+    [
+        ("us-east-1", "t2.micro", "t2.micro"),
+        ("eu-north-1", "t3.micro", "t3.micro"),
```
ddeidda, line 45: Could we repeat one set of dimensions to check the internal cache? On the second call there should be no boto3 calls triggered.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -41,7 +41,6 @@ def test_pcluster_configure(
         vpc_stack.cfn_outputs["VpcId"],
         vpc_stack.cfn_outputs["PublicSubnetId"],
         vpc_stack.cfn_outputs["PrivateSubnetId"],
-        vpc_stack,
```
ddeidda, line 44: Why are we removing this arg?

cli/src/pcluster/utils.py
hanwen-pcluste: Yes. Even if we are just expecting a single result, boto3 will sometimes leave the first page blank. If we call boto3 directly, we will get an empty result.

tests/integration-tests/tests/configure/test_pcluster_configure.py
hanwen-pcluste: Just because it is unused. So I cleaned this up.

cli/src/pcluster/utils.py
ddeidda: Got it. I would suggest the utility function [`paginate_boto_3`](https://github.com/aws/aws-parallelcluster/blob/3a66c3e2250fca6138f5ffd420bbdc298083f003/cli/src/pcluster/utils.py#L120) to simplify the code.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ CHANGELOG
 2.10.1
 ------
 
+**CHANGES**
+
+- Change the default of instance types from the hardcoded `t2.micro` to the free tier instance type 
+  (`t2.micro` or `t3.micro` dependent on region) retrieved from `boto3`. In regions without free tier, the default is `t3.micro`.
```
enrico-usai, line 10: I think we can remove `retrieved from boto3`

cli/src/pcluster/configure/easyconfig.py
enrico-usai: We adde auto_refresh=False as part of this patch: https://github.com/aws/aws-parallelcluster/pull/2248 
Let's double check if we can move back to True.

cli/src/pcluster/utils.py
```diff
@@ -1247,3 +1247,25 @@ def get_instance_gpus(instance_type, instance_info=None):
...
+
+def get_default_instance_type():
+    """If current region support free tier, return the free tier instance type. Otherwise, return t3.micro ."""
+    if not hasattr(get_default_instance_type, "cache"):
+        get_default_instance_type.cache = {}
+    cache = get_default_instance_type.cache
+    region = os.environ.get("AWS_DEFAULT_REGION")
+    if region not in cache:
+        ec2_client = boto3.client("ec2")
+        paginator = ec2_client.get_paginator("describe_instance_types")
```
enrico-usai, line 1260: we have a `paginate_boto3` utility method in the `utils.py` module. We can reuse it.

cli/src/pcluster/configure/easyconfig.py
hanwen-pcluste: After a closer experiment, we still can't do auto_refresh. The problem is not about instance type this time, but about additional_iam_policy. As the code [here](https://github.com/aws/aws-parallelcluster/pull/2248/files#diff-4a0da8aa43fd2a763f436760124df48544177a813b12e02bc0c8dc87e33d8e8fL16) shown, with auto_refresh, the unnecessary policy for AWSBatch will be generated.

cli/src/pcluster/configure/easyconfig.py
hanwen-pcluste: related to this [thread](https://github.com/aws/aws-parallelcluster/pull/2254/files#r530417330)

cli/src/pcluster/configure/easyconfig.py
hanwen-pcluste: related to this [thread](https://github.com/aws/aws-parallelcluster/pull/2254/files#r530417330)

cli/src/pcluster/config/cfn_param_types.py
```diff
@@ -855,6 +856,39 @@ def to_cfn(self):
...
+class ComputeInstanceTypeCfnParam(CfnParam):
+    """
+    Class to manage the compute instance type parameter.
+
+    We need this class in order to set the default instance type from a boto3 call.
+    """
+
+    def refresh(self):
+        """Get default value from a boto3 call for free tier instance type."""
+        if not self.value:
```
ddeidda, line 868: What about a shorter syntax like this:

```
self.value = get_default_instance_type() if pcluster_config.get_section("cluster").get_param_value("scheduler") != "awsbatch" else "optimal"
```

cli/src/pcluster/configure/easyconfig.py
ddeidda: OK, let's talk about this :+1:

cli/tests/conftest.py
```diff
@@ -19,6 +19,14 @@ def clear_env():
         del os.environ["AWS_DEFAULT_REGION"]
 
 
+@pytest.fixture(autouse=True)
+def mock_default_instance(mocker, request):
+    # Mock get_default_instance_type for all tests.
+    if "nomockdefaultinstance" in request.keywords:
```
ddeidda, line 25: Could you breefly explain this skip logic in a comment?

ddeidda (commented):
We are close! Let's talk about the few open points :+1:

cli/src/pcluster/config/cfn_param_types.py
hanwen-pcluste: Done

cli/tests/conftest.py
hanwen-pcluste: Done

ddeidda (approved):

PR #2264, opened 2020-11-24
awsbatch: download AmazonLinux image from ECR rather than Docker Hub

demartinofra:
### Description of changes

* The change from `amazonlinux:latest` to `amazonlinux:2` is done to make sure we alway pull the latest for AL 2
* `docker: 18` removed from install/runtime-versions in CodeBuild buildspec because it is already available in the used container image and not needed anymore
* AL docker image is now pulled from ECR when the architecture is x86. ECR repos do not contain yet multi-arch images that support ARM. 
* Each partition is mapped to a single ECR repo hosted in a specific region. Currently there is no API to retrieve the registry id for AL images in the various regions, hence we keep this static mapping in the template for now.
* A new policy has been added to the CodeBuild role to allow access to the public ECR repo. This needs to be added to our docs.
* A fallback mechanism is put in place so that if the pull from ECR fails there is always a fallback to Docker Hub
* Integration tests have been extended to verify that the image is correctly pulled from ECR whenever possible

### Testing

I ran the following tests:
```
{%- import 'common.jinja2' as common -%}
test-suites:
  schedulers:
    test_awsbatch.py::test_awsbatch:
      dimensions:
        - regions: ["us-east-2"]
          instances: ["m6g.xlarge"]
          oss: ["alinux2"]
          schedulers: ["awsbatch"]
        - regions: {{ common.REGIONS_COMMERCIAL }}
          oss: ["alinux2"]
          schedulers: ["awsbatch"]
        - regions: {{ common.REGIONS_CHINA }}
          oss: ["alinux2"]
          schedulers: ["awsbatch"]
        - regions: {{ common.REGIONS_GOVCLOUD }}
          oss: ["alinux2"]
          schedulers: ["awsbatch"]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ CHANGELOG
 - Remove CloudFormation DescribeStacks API call from AWS Batch Docker entrypoint. This removes the possibility of job
   failures due to CloudFormation throttling.
 
+**CHANGES**
+
+- Pull Amazon Linux Docker images from ECR when building docker image for `awsbatch` scheduler. This only applies to
+  images built for `x86` architecture.
+
+**BUG FIXES**
+
 2.10.0
 ------
 
--- a/cli/src/pcluster/resources/batch/docker/pull-alinux-image.sh
+++ b/cli/src/pcluster/resources/batch/docker/pull-alinux-image.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+pull_docker_image_from_ecr() {
+    echo "Pulling amazonlinux:2 image from ECR"
+    aws ecr get-login-password --region "${ALINUX_ECR_REGISTRY_REGION}" | docker login --username AWS --password-stdin "${ALINUX_ECR_REGISTRY}" || return 1
+    docker pull "${ALINUX_ECR_REGISTRY}/amazonlinux:2" || return 1
+    docker tag "${ALINUX_ECR_REGISTRY}/amazonlinux:2" amazonlinux:2
+}
+
+if [ "${IMAGE}" = "alinux" ] || [ "${IMAGE}" = "alinux2" ]; then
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ CHANGELOG
 - Remove CloudFormation DescribeStacks API call from AWS Batch Docker entrypoint. This removes the possibility of job
   failures due to CloudFormation throttling.
 
+**CHANGES**
+
+- Pull Amazon Linux Docker images from ECR when building docker image for `awsbatch` scheduler. This only applies to
+  images built for `x86` architecture.
```
lukeseawalker, line 15: worth adding about the fallback mechanism?

lukeseawalker (approved):

cli/src/pcluster/resources/batch/docker/pull-alinux-image.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -eux
```
enrico-usai, line 2: let's add pipefail too: `set -euxo pipefail`

cli/src/pcluster/resources/batch/docker/upload-docker-images.sh
```diff
@@ -1,17 +1,20 @@
 #!/usr/bin/env bash
-set -eu
+set -eux
```
enrico-usai, line 2: same here.

cli/src/pcluster/resources/batch/docker/pull-alinux-image.sh
demartinofra: done

cli/src/pcluster/resources/batch/docker/upload-docker-images.sh
demartinofra: done

enrico-usai (approved):

PR #2277, opened 2020-11-27
Fix createami test to not fail when custom_node is not specified

enrico-usai:
if `custom_node` is not specified, the `env` variable was referenced before assignment.


lukeseawalker:
can we avoid to pass `env` to the `run_command` when `custom_node` is not specified?

lukeseawalker (approved):

PR #2282, opened 2020-12-01
Enable rollback data generation for arbitrary S3 files

tilne:
Previously generating rollback data was only supported by the modified
code for S3 keys related to the config files used to allow list certain
features/instance types. This enables the same code to be used for
generating rollback data for other S3 files. The motivation for this is
to enable the code to be used to generate rollback data for dependencies
to be hosted in S3.


tilne:
closing this because the script that was going to use this to generate rollback data is going to use a different rollback mechanism than originally planned

tilne:
Re-opening because it's still necessary to generate rollback data for dependencies being uploaded to s3

yuleiwan (approved):

PR #2284, opened 2020-12-03
Restore default values in cfn template for ComputeInstanceType

ddeidda:
Compute instance type parameter is not rendered if scheduler is Slurm. This caused the error `Parameters: [ComputeInstanceType] must have values` in CloudFormation because a value was still expected.
With this commit we set "NONE" as default to prevent this value being silently used as if set by the user.  

### DIFF
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -14,6 +14,7 @@
     "ComputeInstanceType": {
       "Description": "ComputeFleet EC2 instance type",
       "Type": "String",
+      "Default": "NONE",
       "ConstraintDescription": "Must be a valid EC2 instance type, with support for HVM."
     },
     "MinSize": {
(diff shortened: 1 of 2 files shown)


cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -9,11 +9,13 @@
     "MasterInstanceType": {
       "Description": "Head node EC2 instance type",
       "Type": "String",
+      "Default": "t2.micro",
```
enrico-usai, line 12: let's remove the default for the head node.

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -9,11 +9,13 @@
...
       "ConstraintDescription": "Must be a valid EC2 instance type, with support for HVM."
     },
     "ComputeInstanceType": {
       "Description": "ComputeFleet EC2 instance type",
       "Type": "String",
+      "Default": "t2.micro",
```
enrico-usai, line 18: We can improve the description by highlighting the fact that this parameter is useless for Slurm but needed as CFN param.

We can use NONE instead of t2.micro to avoid hiding other errors. In this way we'll be sure that a value is always passed.

enrico-usai (approved):

PR #2339, opened 2020-12-28
integ-tests: test cluster in subnet with no internet access

demartinofra:
Initial test to verify the possibility to run ParallelCluster in a subnet with no internet access.
Currently the test only creates a cluster with a min queue size of 1 node but does not run any job or perform additional testing.

Known limitations:
* The lack of a VPC endpoint for route53 prevents us from registering nodes in the DNS server. Currently I'm working around this by setting disable_cluster_dns = true.
* The lack of a VPC endpoint for FSx Lustre prevents us from querying the FSx API and retrieve the mount point of the Lustre file system. 
* We are currently installing packages from the OS repo when enable_intel_hpc_platform = true
* On Centos 8 we are apparently installing nfs-utils package at runtime despite we install it at build time. Probably something wrong with the package resource guard when using dnf? This comes from the nfs cookbook.


ddeidda (approved):
LGTM :+1:

ddeidda (approved):

PR #2342, opened 2020-12-29
Add NCCL benchmarks to integration tests

ddeidda:

### DIFF
--- a/tests/integration-tests/tests/nccl/test_nccl/test_nccl/init_nccl_benchmarks.sh
+++ b/tests/integration-tests/tests/nccl/test_nccl/test_nccl/init_nccl_benchmarks.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+set -e
+
+rm -rf /shared/${1}
+
+module load ${1}
+NCCL_BENCHMARKS_VERSION='2.0.0'
+NCCL_VERSION='2.7.8-1'
+ML_REPO_PKG='nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb'
+CUDA_VERSION='11.0'
+OFI_NCCL_VERSION='1.1.1'
+MPI_HOME=$(which mpirun | awk -F '/bin' '{print $1}')
+NVCC_GENCODE="-gencode=arch=compute_80,code=sm_80" # Arch for NVIDIA A100
+
+# Install nccl (binaries)
+#wget https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/${ML_REPO_PKG}
+#sudo dpkg -i ${ML_REPO_PKG}
+#sudo apt install libnccl2=${NCCL_VERSION}+cuda${CUDA_VERSION} libnccl-dev=${NCCL_VERSION}+cuda${CUDA_VERSION}
+
+mkdir -p /shared/${1}
+
+# Install and build nccl from sources
+cd /shared/${1}
+wget https://github.com/NVIDIA/nccl/archive/v${NCCL_VERSION}.tar.gz
+tar zxvf "v${NCCL_VERSION}.tar.gz"
+cd nccl-${NCCL_VERSION}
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/nccl/test_nccl/test_nccl/nccl_tests_submit_openmpi.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#SBATCH --nodes=2
+#SBATCH --exclusive
+#SBATCH --ntasks-per-node=4
```
rexcsn, line 4: I think this is the place we are forcing the use of 4 GPUs, by not allocating all CPUs for the test?

In any case, it would be nice to try to use all 8 GPUs

tests/integration-tests/tests/nccl/test_nccl/test_nccl/nccl_tests_submit_openmpi.sh
```diff
@@ -0,0 +1,14 @@
...
+
+module load openmpi
+mpirun \
+-x FI_PROVIDER="efa" \
+-x FI_EFA_USE_DEVICE_RDMA=1 \
+-x RDMAV_FORK_SAFE=1 \
+-x NCCL_ALGO=ring \
+-x LD_LIBRARY_PATH=/shared/openmpi/nccl-2.8.3-1/build/lib/:$LD_LIBRARY_PATH \
+--mca pml ^cm --mca btl tcp,self --mca btl_tcp_if_exclude lo,docker0 --bind-to none \
+/shared/openmpi/nccl-tests-2.0.0/build/all_reduce_perf -b 8 -e 1G -f 2 -g 1 -c 1 -n 100 > /shared/nccl_tests.out
```
rexcsn, line 14: Nice-to-have: we would want to understand what these options mean and ideally drop a comment about why we have chosen these set of options to run the test. Doc for options is [here](https://github.com/NVIDIA/nccl-tests/blob/master/README.md), but I not sure if we understand why we chose these options.

tests/integration-tests/tests/nccl/test_nccl/test_nccl/nccl_tests_submit_openmpi.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#SBATCH --nodes=2
+#SBATCH --exclusive
+#SBATCH --ntasks-per-node=8
+
+
+module load openmpi
+NCCL_VERSION='2.7.8-1'
```
rashikakheria, line 8: You should change this version whenever NCCL is upgraded in pcluster AMIs.

tests/integration-tests/tests/nccl/test_nccl/test_nccl/init_nccl_benchmarks.sh
```diff
@@ -0,0 +1,42 @@
...
+set -e
+
+rm -rf /shared/${1}
+
+module load ${1}
+NCCL_BENCHMARKS_VERSION='2.0.0'
+NCCL_VERSION='2.7.8-1'
+ML_REPO_PKG='nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb'
+CUDA_VERSION='11.0'
+OFI_NCCL_VERSION='1.1.1'
```
rashikakheria, line 11: This should also change with new releases of OFI NCCL Plugin

tests/integration-tests/tests/nccl/test_nccl/test_nccl/nccl_tests_submit_openmpi.sh
ddeidda: We are currently not installing NCCL in pcluster AMIs; that's why we do it in this test. Anyway, I agree we should find a way to always use the latest version. For now, versions are parameterised exactly for this reason.

rexcsn (approved):
LGTM!

PR #2358, opened 2021-01-11
3.0: Initialize the imagebuilder configuration and cdk

yuleiwan:
## Commits
* Add imagebuilder schema and config based on 3.0 new create ami configuration
* Add imagebuilder cdk to get parameter from imagebuilder config
* Add official ami build component and custom ami build component

#### imagebuilder_schema
`cli/src/pcluster/schemas/imagebuilder_schema.py`: contains three main sections schema(`image_schema`, `build_schema` and `dev_settings_schema`). For each section contains attributes and nested schema.
#### imagebuilder_config
`cli/src/pcluster/config/imagebuilder_config.py`: contains three main configs corresponding to three main schema sections. Give an example for parent AMI validation in this layer.
#### cdk_build
`cli/src/pcluster/templates/cdk_builder.py`: use cdk with imagebuilder object to retrieve the parameters defined by imagebuilder configuration to generate cloudformation template
#### imagebuilder components
`cloudformation/imagebuilder`: define official ami build component and custom ami build component. All the AMI with different OSes use the same component. Use condition `RebootAndUpdate` in dev setting  to choose the update_and_reboot  component for cdk template generation

## Tests
add basic test for imagebuilder_schema and imagebuilder cdk_builder

## Current Block Issues
* Data property in imagebuilder component with literal string isn't parsed correctly. Create an issue for cdk repo to track this issue: https://github.com/aws/aws-cdk/issues/12405 (resolved)
* centos7/centos8 ami creation fails after rebooting, instance status check fails

### DIFF
--- a/cli/src/pcluster/models/imagebuilder.py
+++ b/cli/src/pcluster/models/imagebuilder.py
@@ -1,0 +1,167 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This module contains all the classes representing the Resources objects.
+# These objects are obtained from the configuration file through a conversion based on the Schema classes.
+#
+
+from typing import List
+
+from pcluster import utils
+from pcluster.config.extended_builtin_class import MarkedBool
+from pcluster.models.cluster import Resource, Tag
+from pcluster.validators.ec2_validators import BaseAMIValidator
+
+# ---------------------- Image ---------------------- #
+
+
+class Volume(Resource):
+    """Represent the volume configuration for the ImageBuilder."""
+
+    def __init__(self, size: int = None, encrypted: bool = None, kms_key_id: str = None):
+        super().__init__()
+        self.size = size
+        self.encrypted = encrypted
+        self.kms_key_id = kms_key_id
+        self._set_defaults()
+        # TODO: add validator
+
+    def _set_defaults(self):
+        if self.encrypted is None:
+            self.encrypted = MarkedBool(False)
+
+
+class Image(Resource):
+    """Represent the image configuration for the ImageBuilder."""
+
+    def __init__(
+        self,
+        name: str,
+        description: str = None,
+        tags: List[Tag] = None,
+        root_volume: Volume = None,
+    ):
+        super().__init__()
+        self.name = name
+        self.description = description
+        self.tags = tags
+        self.root_volume = root_volume
+        self._set_default()
+        # TODO: add validator
+
+    def _set_default(self):
+        if self.tags is None:
+            self.tags = []
+        default_tag = Tag("PclusterVersion", utils.get_installed_version())
+        default_tag.implied = True
+        self.tags.append(default_tag)
+
+
+# ---------------------- Build ---------------------- #
+
+
+class Component(Resource):
+    """Represent the components configuration for the ImageBuilder."""
+
+    def __init__(self, type: str = None, value: str = None):
+        super().__init__()
+        self.type = type
+        self.value = value
+        # TODO: add validator
+
+
+class Build(Resource):
+    """Represent the build configuration for the ImageBuilder."""
+
+    def __init__(
+        self,
+        instance_type: str,
+        parent_image: str,
+        instance_role: str = None,  # TODO: auto generate if not assigned
+        subnet_id: str = None,  # TODO: auto generate if not assigned
+        tags: List[Tag] = None,
+        security_group_ids: List[str] = None,
+        components: List[Component] = None,
+    ):
+        super().__init__()
+        self.instance_type = instance_type
+        self.parent_image = parent_image
+        self.instance_role = instance_role
+        self.tags = tags
+        self.subnet_id = subnet_id
+        self.security_group_ids = security_group_ids
+        self.components = components
+        # TODO: add validator
+
+
+# ---------------------- Dev Settings ---------------------- #
+
+
+class ChefCookbook(Resource):
+    """Represent the chef cookbook configuration for the ImageBuilder."""
+
+    def __init__(self, url: str, json: str):
+        super().__init__()
+        self.url = url
+        self.json = json
+        # TODO: add validator
+
+
+class DevSettings(Resource):
+    """Represent the dev settings configuration for the ImageBuilder."""
+
+    def __init__(
+        self,
+        update_os_and_reboot: bool = None,
+        disable_pcluster_component: bool = None,
+        chef_cookbook: ChefCookbook = None,
+        node_url: str = None,
+        aws_batch_cli_url: str = None,
+        distribution_configuration_arn: str = None,
+        terminate_instance_on_failure: bool = None,
+    ):
+        super().__init__()
+        self.update_os_and_reboot = update_os_and_reboot
+        self.disable_pcluster_component = disable_pcluster_component
+        self.chef_cookbook = chef_cookbook
+        self.node_url = node_url
+        self.aws_batch_cli_url = aws_batch_cli_url
+        self.distribution_configuration_arn = distribution_configuration_arn
+        self.terminate_instance_on_failure = terminate_instance_on_failure
+        self._set_default()
+        # TODO: add validator
+
+    def _set_default(self):
+        if self.update_os_and_reboot is None:
+            self.update_os_and_reboot = MarkedBool(False)
+        if self.disable_pcluster_component is None:
+            self.disable_pcluster_component = MarkedBool(False)
+        if self.terminate_instance_on_failure is None:
+            self.terminate_instance_on_failure = MarkedBool(True)
+
+
+# ---------------------- ImageBuilder ---------------------- #
+
+
+class ImageBuilder(Resource):
+    """Represent the configuration of an ImageBuilder."""
+
+    def __init__(
+        self,
+        image: Image,
+        build: Build,
+        dev_settings: DevSettings = None,
+    ):
+        super().__init__()
+        self.image = image
+        self.build = build
+        self.dev_settings = dev_settings
+        self._add_validator(BaseAMIValidator, ami_id=self.build.parent_image)
--- a/cli/src/pcluster/schemas/imagebuilder_schema.py
+++ b/cli/src/pcluster/schemas/imagebuilder_schema.py
@@ -1,0 +1,128 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# This module contains all the classes representing the Schema of the configuration file.
+# These classes are created by following marshmallow syntax.
+#
+from marshmallow import fields, post_load
+
+from pcluster.models.imagebuilder import Build, ChefCookbook, Component, DevSettings, Image, ImageBuilder, Volume
+from pcluster.schemas.cluster_schema import BaseSchema, TagSchema
+
+# ---------------------- Image Schema---------------------- #
+
+
+class VolumeSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder Volume."""
+
+    size = fields.Int(data_key="Size")
+    encrypted = fields.Bool(data_key="Encrypted")
+    kms_key_id = fields.Str()
+
+    @post_load()
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return Volume(**data)
+
+
+class ImageSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder Image."""
+
+    name = fields.Str(data_key="Name", required=True)
+    description = fields.Str(data_key="Description")
+    tags = fields.List(fields.Nested(TagSchema), data_key="Tags")
+    root_volume = fields.Nested(VolumeSchema)
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return Image(**data)
+
+
+# ---------------------- Build Schema---------------------- #
+
+
+class ComponentSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder component."""
+
+    type = fields.Str(data_key="Type")
+    value = fields.Str(data_key="Value")
+
+    @post_load()
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return Component(**data)
+
+
+class BuildSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder Build."""
+
+    instance_role = fields.Str()
+    instance_type = fields.Str(required=True)
+    components = fields.List(fields.Nested(ComponentSchema), data_key="Components")
+    parent_image = fields.Str(required=True)
+    tags = fields.List(fields.Nested(TagSchema), data_key="Tags")
+    security_group_ids = fields.List(fields.Str)
+    subnet_id = fields.Str()
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return Build(**data)
+
+
+# ---------------------- Dev Settings Schema ---------------------- #
+
+
+class ChefCookbookSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder chef cookbook."""
+
+    url = fields.Str(data_key="Url")
+    json = fields.Str(data_key="Json")
+
+    @post_load()
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return ChefCookbook(**data)
+
+
+class DevSettingsSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder Dev Setting."""
+
+    update_os_and_reboot = fields.Bool()
+    disable_pcluster_component = fields.Bool()
+    chef_cookbook = fields.Nested(ChefCookbookSchema)
+    node_url = fields.Str()
+    aws_batch_cli_url = fields.Str(data_key="AWSBatchCliUrl")
+    distribution_configuration_arn = fields.Str()
+    terminate_instance_on_failure = fields.Bool()
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return DevSettings(**data)
+
+
+# ---------------------- ImageBuilder Schema ---------------------- #
+
+
+class ImageBuilderSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder."""
+
+    image = fields.Nested(ImageSchema, data_key="Image", required=True)
+    build = fields.Nested(BuildSchema, data_key="Build", required=True)
+    dev_settings = fields.Nested(DevSettingsSchema)
+
+    @post_load()
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return ImageBuilder(**data)
--- a/cli/src/pcluster/templates/cdk_builder.py
+++ b/cli/src/pcluster/templates/cdk_builder.py
@@ -12,15 +12,16 @@
 #
 # This module contains all the classes required to convert a Cluster into a CFN template by using CDK.
 #
-
 import os
 import tempfile
 
 from aws_cdk import core
 
-from common.utils import load_yaml
+from common.utils import load_yaml_dict
 from pcluster.models.cluster import Cluster
+from pcluster.models.imagebuilder import ImageBuilder
 from pcluster.templates.cluster_stack import ClusterStack
+from pcluster.templates.imagebuilder_stack import ImageBuilderStack
 
 
 class CDKTemplateBuilder:
@@ -33,6 +34,17 @@ class CDKTemplateBuilder:
             app = core.App(outdir=str(tempdir))
             ClusterStack(app, output_file, cluster=cluster)
             app.synth()
-            generated_template = load_yaml(os.path.join(tempdir, f"{output_file}.template.json"))
+            generated_template = load_yaml_dict(os.path.join(tempdir, f"{output_file}.template.json"))
+
+        return generated_template
+
+    def build_ami(self, imagebuild: ImageBuilder):
+        """Build template for the given imagebuilder and return as output in Yaml format."""
+        with tempfile.TemporaryDirectory() as tempdir:
+            output_file = "imagebuilder"
+            app = core.App(outdir=str(tempdir))
+            ImageBuilderStack(app, output_file, imagebuild)
+            app.synth()
+            generated_template = load_yaml_dict(os.path.join(tempdir, f"{output_file}.template.json"))
 
         return generated_template
--- a/cli/src/pcluster/templates/imagebuilder_stack.py
+++ b/cli/src/pcluster/templates/imagebuilder_stack.py
@@ -1,0 +1,186 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# This module contains all the classes required to convert a ImageBuilder into a CFN template by using CDK.
+#
+
+import os
+
+from aws_cdk import aws_iam as iam
+from aws_cdk import aws_imagebuilder as imagebuilder
+from aws_cdk import aws_ssm as ssm
+from aws_cdk import core
+
+import pcluster.utils as utils
+from common.utils import load_yaml
+from pcluster.models.imagebuilder import ImageBuilder
+
+
+class ImageBuilderStack(core.Stack):
+    """Create the Stack for imagebuilder."""
+
+    def __init__(self, scope: core.Construct, construct_id: str, imagebuild: ImageBuilder, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        image = imagebuild.image
+        build = imagebuild.build
+        dev_settings = imagebuild.dev_settings
+
+        # TODO: use attributes from imagebuilder config instead of using these static variables.
+        core.CfnParameter(self, "EnableNvidia", type="String", default="false", description="EnableNvidia")
+        core.CfnParameter(self, "EnableDCV", type="String", default="false", description="EnableDCV")
+        default_node_url = dev_settings.node_url if dev_settings and dev_settings.node_url else ""
+        core.CfnParameter(
+            self,
+            "CustomNodePackage",
+            type="String",
+            default=default_node_url,
+            description="CustomNodePackage",
+        )
+        core.CfnParameter(
+            self,
+            "UpdateAndReboot",
+            type="String",
+            default=str.lower(str(bool(dev_settings.update_os_and_reboot))),
+            description="UpdateAndReboot",
+        )
+        ami_info = utils.get_info_for_amis([build.parent_image])
+        architecture = ami_info[0].get("Architecture")
+        if build.instance_type:
+            instance_types = [build.instance_type]
+        else:
+            instance_types = ["p2.xlarge", "c5.xlarge"] if architecture == "x86_64" else ["m6g.xlarge"]
+
+        # Setup ImageBuilder Resources
+        resources_prefix = utils.generate_random_prefix()
+
+        # InstanceRole
+        managed_policy_arns = [
+            core.Fn.sub("arn:${AWS::Partition}:iam::aws:policy/AmazonSSMManagedInstanceCore"),
+            core.Fn.sub("arn:${AWS::Partition}:iam::aws:policy/EC2InstanceProfileForImageBuilder"),
+        ]
+        if build.instance_role:
+            managed_policy_arns.extend([build.instance_role])
+
+        instancerole = iam.CfnRole(
+            self,
+            id="InstanceRole",
+            managed_policy_arns=managed_policy_arns,
+            assume_role_policy_document={
+                "Statement": {
+                    "Action": "sts:AssumeRole",
+                    "Effect": "Allow",
+                    "Principal": {"Service": "ec2.amazonaws.com"},
+                },
+                "Version": "2012-10-17",
+            },
+            path="/executionServiceEC2Role/",
+        )
+        instancerole.add_metadata("Comment", "Role to be used by instance during image build.")
+
+        # InstanceProfile
+        iam.CfnInstanceProfile(
+            self, id="InstanceProfile", path="/executionServiceEC2Role/", roles=[core.Fn.ref("InstanceRole")]
+        )
+
+        # InfrastructureConfiguration
+        imagebuilder.CfnInfrastructureConfiguration(
+            self,
+            id="PClusterImageInfrastructureConfiguration",
+            name="-".join(["PCluster-Image-Infrastructure-Configuration", resources_prefix]),
+            instance_profile_name=core.Fn.ref("InstanceProfile"),
+            terminate_instance_on_failure=bool(dev_settings.terminate_instance_on_failure),
+            instance_types=instance_types,
+        )
+
+        # Define ami build instance ebs
+        increase_volume_size = 15
+        if image.root_volume:
+            ebs = imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                volume_size=image.root_volume.size
+                or ami_info[0].get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize") + increase_volume_size,
+                volume_type="gp2",
+            )
+        else:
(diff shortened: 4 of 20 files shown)


cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+                version="0.0.1",
+                description="Bake PCluster custom AMI",
+                change_description="First version",
+                platform="Linux",
+                # TODO: Fix the bug using data property or uri with core.Fn.sub
+                uri="s3://test-slurm/custom_ami_build_component.yml",
+            )
+
+        # ImageRecipe
+        imagebuilder.CfnImageRecipe(
```
enrico-usai, line 111: You could think to define a `ImageRecipeConstruct` to isolate the code.

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+            key_pair=imagebuild.config.build_config.key_pair_name,
+        )
+
+        # Component
+        if imagebuild.config.dev_settings_config and imagebuild.config.dev_settings_config.update_and_reboot:
+            imagebuilder.CfnComponent(
+                self,
+                id="PClusterComponent",
+                name="PCluster",
+                version="0.0.1",
```
enrico-usai, line 90: Are these attributes required?
If so let's use a constant for the version and use it  everywhere

cli/tests/pcluster/schemas/test_imagebuilder_schema/test_imagebuilder_schema/imagebuilder_schema_simple.yaml
```diff
@@ -0,0 +1,6 @@
+Image:
```
enrico-usai, line 1: Is this a test with only the required parameters? If so let's rename the test to imagebuilder_schema_required.yaml

cli/tests/pcluster/schemas/test_imagebuilder_schema.py
```diff
@@ -0,0 +1,24 @@
...
+    input_yaml = load_yaml(test_datadir / config_file_name)
+    print(input_yaml)
+    imagebuilder_config = ImageBuilderSchema().load(input_yaml)
+    print(imagebuilder_config)
+
+    # Re-create Yaml file from model and compare content
+    output_json = ImageBuilderSchema().dump(imagebuilder_config)
+
+    # TODO: don't compare the default value in imagebuilder config (e.g. delete_on_termination=true for volume config)
+    # assert_that(json.dumps(input_yaml, sort_keys=True)).is_equal_to(json.dumps(output_json, sort_keys=True))
```
enrico-usai, line 20: to be re-enabled?

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+            )
+
+        # ImageRecipe
+        imagebuilder.CfnImageRecipe(
+            self,
+            id="PClusterImageRecipe",
+            name="PCluster-2-10-1",
+            version="0.0.1",
+            parent_image=core.Fn.sub(
+                "arn:${AWS::Partition}:imagebuilder:${AWS::Region}:aws:image/amazon-linux-2-x86/x.x.x"
```
enrico-usai, line 117: If I'm not wrong you can simplify it by writing: 
```python
imagebuild.config.build_config.parent_image or "arn:${AWS::Partition}:imagebuilder:${AWS::Region}:aws:image/amazon-linux-2-x86/x.x.x"
```

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+            ),
+            components=[
+                imagebuilder.CfnImageRecipe.ComponentConfigurationProperty(
+                    component_arn=core.Fn.ref("PClusterComponent")
+                )
+            ],
+            block_device_mappings=[
+                imagebuilder.CfnImageRecipe.InstanceBlockDeviceMappingProperty(
+                    device_name="/dev/xvda",
+                    ebs=imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
```
enrico-usai, line 129: let's define the ebs outside the block_device_mappings list

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+                        volume_size=imagebuild.config.image_config.root_volume.size,
+                        volume_type="gp2",
+                    )
+                    if imagebuild.config.image_config.root_volume
+                    else imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                        delete_on_termination=True, volume_size=25, volume_type="gp2"
+                    ),
+                )
+            ],
+            tags={"ParallelClusterVersion": "2.10.1"},
```
enrico-usai, line 140: We already have an utility function to retrieve the version of the running CLI: `utils.get_installed_version()`

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+                imagebuilder.CfnImageRecipe.InstanceBlockDeviceMappingProperty(
+                    device_name="/dev/xvda",
+                    ebs=imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                        delete_on_termination=imagebuild.config.image_config.root_volume.delete_on_termination,
+                        volume_size=imagebuild.config.image_config.root_volume.size,
+                        volume_type="gp2",
+                    )
+                    if imagebuild.config.image_config.root_volume
+                    else imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                        delete_on_termination=True, volume_size=25, volume_type="gp2"
```
enrico-usai, line 136: let's move all these constant values on top of the file to be more clear and explicit.

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+        # Image
+        imagebuilder.CfnImage(
+            self,
+            id="PClusterImage",
+            image_recipe_arn=core.Fn.ref("PClusterImageRecipe"),
+            infrastructure_configuration_arn=core.Fn.ref("PClusterImageInfrastructureConfiguration"),
+        )
+
+        # AWS Systems Manager
+        ssm.CfnParameter(
```
enrico-usai, line 152: Why ssm.CfnParameter and not core.CfnParameter ?

cloudformation/imagebuilder/custom_ami_build_component.yml
```diff
@@ -0,0 +1,231 @@
+name: PCluster
```
enrico-usai, line 1: Should it be yaml instead of yml?

enrico-usai (commented):
Great job, I reviewed the cli part but I didn't look a the yaml components.

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+from aws_cdk import aws_imagebuilder as imagebuilder
+from aws_cdk import aws_ssm as ssm
 from aws_cdk import core
 
 from common.utils import load_yaml
 from pcluster.cluster import Cluster, Fsx, HeadNode
+from pcluster.imagebuilder import ImageBuilder
+
+
+class ImageBuilderStack(core.Stack):
```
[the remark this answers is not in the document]
yuleiwan: Done

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+        # TODO add S3 bucket
+
+        # InfrastructureConfiguration
+        imagebuilder.CfnInfrastructureConfiguration(
+            self,
+            id="PClusterImageInfrastructureConfiguration",
+            name="PCluster-Image-Infrastructure-Configuration",
+            instance_profile_name=core.Fn.ref("InstanceProfile"),
+            terminate_instance_on_failure=imagebuild.config.build_config.terminate_on_failure,
+            instance_types=["p2.xlarge", "c5.xlarge"] if archecture == "x86_64" else ["m6g.xlarge"],
```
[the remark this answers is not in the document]
yuleiwan: Done

cli/src/pcluster/templates/cdk_builder.py
```diff
@@ -12,15 +12,151 @@
...
+            key_pair=imagebuild.config.build_config.key_pair_name,
+        )
+
+        # Component
+        if imagebuild.config.dev_settings_config and imagebuild.config.dev_settings_config.update_and_reboot:
+            imagebuilder.CfnComponent(
```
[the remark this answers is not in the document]
yuleiwan: Done

cli/src/pcluster/templates/cdk_builder.py
yuleiwan: I don't fully understand the Construct, the ImageRecipe relies on components defined above, so should I move both components and ImageRecipe to  `ImageRecipeConstruct`? I notice HeadNode is an object attribute of Cluster, so we create a construct for it. In this case, should we create an object like Recipe for ImageBuilder?

cli/src/pcluster/templates/cdk_builder.py
yuleiwan: Done

cli/src/pcluster/templates/cdk_builder.py
yuleiwan: use `utils.get_installed_version()`, thanks

cli/src/pcluster/templates/cdk_builder.py
yuleiwan: replace constant values here with default values from volume config

cli/src/pcluster/templates/cdk_builder.py
yuleiwan: we defined a SSM rather than cfn parameter here, so we are able to store the AMI ID in a SSM parameter store and retrieve this from SSM parameter store.

cloudformation/imagebuilder/custom_ami_build_component.yml
yuleiwan: Yes, it should be yaml. Replace `custom_ami_build_component.yml` and `official_ami_build_component.yml` with `update_and_reboot.yaml` and `pcluster_install.yaml` to isolate internal and external build.

cli/tests/pcluster/schemas/test_imagebuilder_schema.py
yuleiwan: It Will be enabled when I implement the implied default values in config layer.

cli/src/pcluster/config/imagebuilder_config.py
```diff
@@ -0,0 +1,149 @@
...
+
+class ImageBuilderChefCookbookConfig(Config):
+    """Represent the chef cookbook configuration for the ImageBuilder."""
+
+    def __init__(self, url: str, json: str):
+        super().__init__()
+        self.url = url
+        self.json = json
+        # TODO: add validator
+        self._validators = []
```
enrico-usai, line 58: you don't need to inizialize the validators array because it's already done in the parent (the Config class), you only need to register the new validators.

cli/src/pcluster/config/imagebuilder_config.py
```diff
@@ -0,0 +1,149 @@
...
+        self,
+        image_config: ImageBuilderImageConfig,
+        build_config: ImageBuilderBuildConfig,
+        dev_settings_config: ImageBuilderDevSettingsConfig = None,
+    ):
+        super().__init__()
+        self.image_config = image_config
+        self.build_config = build_config
+        self.dev_settings_config = dev_settings_config
+        self._register_validator(BaseAMIValidator, priority=100, ami_id=self.build_config.parent_image)
```
enrico-usai, line 149: If you don't need a specific order for the validators you can avoid to set the priority.

cli/src/pcluster/schemas/imagebuilder_schema.py
```diff
@@ -0,0 +1,128 @@
+from marshmallow import fields, post_load
+
+from pcluster.config.imagebuilder_config import (
+    ImageBuilderBucketLogConfig,
+    ImageBuilderBuildConfig,
+    ImageBuilderChefCookbookConfig,
+    ImageBuilderComponentsConfig,
```
enrico-usai, line 7: I see there is the ImageBuilder prefix in all the components
I think you can avoid it and have something like:
```
BucketLogConfig,
BuildConfig,
...
ImageBuilderConfig
```

cli/src/pcluster/schemas/imagebuilder_schema.py
```diff
@@ -0,0 +1,128 @@
...
+    ImageBuilderConfig,
+    ImageBuilderDevSettingsConfig,
+    ImageBuilderImageConfig,
+    ImageBuilderVolumeConfig,
+)
+from pcluster.schemas.cluster_schema import BaseSchema
+
+
+class ImageBuilderVolumeSchema(BaseSchema):
```
enrico-usai, line 16: You can avoid the ImageBuilder prefix in the classes too. It's clear that we're talking about image builder components because we're in the imagebuilder_schema.py module.

cli/src/pcluster/schemas/imagebuilder_schema.py
```diff
@@ -0,0 +1,128 @@
...
+    @post_load()
+    def make_config(self, data, **kwargs):
+        """Generate config object."""
+        return ImageBuilderVolumeConfig(**data)
+
+
+class ImageBuilderComponentsSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder components."""
+
+    component_arns = fields.List(fields.Str, data_key="ComponentArns", required=True)
```
enrico-usai, line 33: Once the @hanwen-pcluste patch will be merged you can avoid to specify the data_key for all the attributes that are aligned to the configuration file attribute and the only difference is CamelCase vs snake_case

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+from pcluster.imagebuilder import ImageBuilder
+
+
+class ImageBuilderStack(core.Stack):
+    """Create the Stack for imagebuilder."""
+
+    def __init__(self, scope: core.Construct, construct_id: str, imagebuild: ImageBuilder, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        self.imagebuild = imagebuild
+        image_config = imagebuild.config.image_config
```
enrico-usai, line 19: By looking at this code it seems to me that we should review the design of this part.
I don't like accessing to config then image_config, we should instead use something like:

imagebuild.image.root_volume.size directly without the need to navigate the config items.
Feel free to go ahead but I think we'll refactor this part.

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+                or "arn:${AWS::Partition}:imagebuilder:${AWS::Region}:aws:image/amazon-linux-2-x86/x.x.x"
+            ),
+            components=components,
+            block_device_mappings=[
+                imagebuilder.CfnImageRecipe.InstanceBlockDeviceMappingProperty(
+                    device_name="/dev/xvda",
+                    ebs=ebs,
+                )
+            ],
+            tags={"ParallelClusterVersion": "2.10.1"},
```
enrico-usai, line 146: remember the utils.get_pclsuter_version

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+                imagebuilder.CfnImageRecipe.ComponentConfigurationProperty(
+                    component_arn=core.Fn.ref("PClusterComponent")
+                )
+            ]
+
+        # ImageRecipe
+        imagebuilder.CfnImageRecipe(
+            self,
+            id="PClusterImageRecipe",
+            name="PCluster-2-10-1" + str(hash(imagebuild)),
```
enrico-usai, line 133: if you need the version let's use the utils.get_pclsuter_version and convert the value. We should have a single place in the code with the 2.10.1 string, to be easier bump the version for the new releases.

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+            self,
+            id="PClusterParameter",
+            description="Image Id for PCluster",
+            name="/Test/Images/PCluster",
+            type="String",
+            value=core.Fn.get_att("PClusterImage", "ImageId").to_string(),
+        )
+
+
+def get_component_data(source_dir, file_name):
```
enrico-usai, line 168: minor: let's add an underscore prefix for private functions (functions used in this module only).

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        # InfrastructureConfiguration
+        imagebuilder.CfnInfrastructureConfiguration(
+            self,
+            id="PClusterImageInfrastructureConfiguration",
+            name="PCluster-Image-Infrastructure-Configuration",
+            instance_profile_name=core.Fn.ref("InstanceProfile"),
+            terminate_instance_on_failure=build_config.terminate_on_failure,
+            instance_types=instance_types,
```
lukeseawalker, line 68: get instance type from build_config

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+        image_config = imagebuild.config.image_config
+        build_config = imagebuild.config.build_config
+        dev_settings_config = imagebuild.config.dev_settings_config
+
+        # TODO: use attributes from imagebuilder config instead of using these static variables.
+        core.CfnParameter(self, "EnableNvidia", type="String", default="false", description="EnableNvidia")
+        core.CfnParameter(self, "EnableDCV", type="String", default="false", description="EnableDCV")
+
+        # TODO: get archecture by ami id
+        architecture = "x86_64"
```
lukeseawalker, line 28: instance type should be get from build_config and a validator should be implemented if the given parent image can be build using the given instance type

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+                version="0.0.1",
+                description="Bake PCluster AMI",
+                change_description="First version",
+                platform="Linux",
+                data=core.Fn.sub(get_component_data(cloudformation_dir, "pcluster_install.yaml")),
+            )
+
+            components = [
+                imagebuilder.CfnImageRecipe.ComponentConfigurationProperty(
+                    component_arn=core.Fn.ref("UpdateAndReboot")
```
lukeseawalker, line 105: reference is UpdateAndRebootComponent

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+                ),
+                imagebuilder.CfnImageRecipe.ComponentConfigurationProperty(
+                    component_arn=core.Fn.ref("PClusterComponent")
+                ),
+            ]
+        else:
```
lukeseawalker, line 111: can we factor out the PClusterComponent, removing the else?
something like
```
components = []

if update_and_reboot:
    create UpdateAndRebootComponent
    components += [UpdateAndRebootComponent]

create PClusterComponent
components += [PClusterComponent]
```

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+        # InstanceProfile
+        iam.CfnInstanceProfile(
+            self, id="InstanceProfile", path="/executionServiceEC2Role/", roles=[core.Fn.ref("InstanceRole")]
+        )
+
+        # InfrastructureConfiguration
+        imagebuilder.CfnInfrastructureConfiguration(
+            self,
+            id="PClusterImageInfrastructureConfiguration",
+            name="PCluster-Image-Infrastructure-Configuration",
```
lukeseawalker, line 65: name should be unique, we can use the utility function `generate_random_name_with_prefix`

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        cloudformation_dir = "../../../../cloudformation/imagebuilder/"
+
+        # Component
+        if dev_settings_config and dev_settings_config.update_and_reboot:
+            imagebuilder.CfnComponent(
+                self,
+                id="UpdateAndRebootComponent",
+                name="UpdateAndReboot",
```
lukeseawalker, line 85: components names should be unique as well, we can use `generate_random_name_with_prefix`

cli/src/pcluster/templates/imagebuilder_stack.py
lukeseawalker, line 146: not sure if it was intended to tag the imageRecipe as well, but to tag the output image this should be set on the `imagebuilder.CfnImage`

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Yes, as far as I know, we will refactor the config and core layer to ImageBuilder and ImageBuilderManager, I just keep the logic now and will modify it after the new design of this part is released

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Deleted here, will add it if we need

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done, I refactor `generate_random_name_with_prefix` and create a new function `generate_random_prefix`. So in the same stack, we can use same suffix/prefix for resource name.

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: yes, and this validator should be implemented in config(core) layer, I leave the validator part blank and think we should have a validation plan before implementation.

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Deleted this, will add if we need

cli/src/pcluster/schemas/imagebuilder_schema.py
yuleiwan: Done

cli/src/pcluster/schemas/imagebuilder_schema.py
yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, size: int = None, encrypted: bool = None, kms_key_id: str = None):
+        super().__init__()
+        self.size = size
+        self.encrypted = encrypted
+        self.kms_key_id = kms_key_id
+        self._default_value()
+        # TODO: add validator
+
+    def _default_value(self):
```
enrico-usai, line 37: The idea to group the set of the defaults together it's a nice one.
I'd rename the function to `_set_defaults()`

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,175 @@
...
+from common.utils import load_yaml
+from pcluster.models.imagebuilder import ImageBuilder
+
+
+class ImageBuilderStack(core.Stack):
+    """Create the Stack for imagebuilder."""
+
+    def __init__(self, scope: core.Construct, construct_id: str, imagebuild: ImageBuilder, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        self.imagebuild = imagebuild
```
enrico-usai, line 19: I don't think you need to set the imagebuild in the stack object.

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,175 @@
...
+        image_config = imagebuild.image
+        build_config = imagebuild.build
+        dev_settings_config = imagebuild.dev_settings
```
enrico-usai, line 22: you can avoid to define local variables for image_config, build_config and dev_settings_config.

If you watn to keep them let's rename to:
image, build and dev_settings, you can remove the _config suffix

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,175 @@
...
+            description="Image Id for PCluster",
+            name="-".join(["/Test/Images/PCluster", resources_prefix]),
+            type="String",
+            value=core.Fn.get_att("PClusterImage", "ImageId").to_string(),
+        )
+
+
+def _get_component_data(source_dir, file_name):
+    """Get component string data from yaml file."""
+    os.path.join(source_dir, file_name)
```
enrico-usai, line 173: useless

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,175 @@
...
+    data = load_yaml(os.path.join(source_dir, file_name))
+    return yaml.dump(data)
```
enrico-usai, line 175: let's rename the load_yaml in common to load_yaml_dict.
and you can move this function in common by calling it load_yaml

cli/src/pcluster/utils.py
```diff
@@ -141,11 +141,21 @@ def generate_random_name_with_prefix(name_prefix):
 
     Example: <name_prefix>-4htvo26lchkqeho1
     """
-    random_string = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(16))
+    random_string = generate_random_prefix()
     output_name = "-".join([name_prefix.lower()[: 63 - len(random_string) - 1], random_string])
     return output_name
 
 
+def generate_random_prefix():
```
enrico-usai, line 149: Let's add a unit test for this new function.

cli/src/pcluster/utils.py
```diff
@@ -141,11 +141,21 @@ def generate_random_name_with_prefix(name_prefix):
...
+    """
+    Generate a random prefix that is 16 characters long.
+
+    Example: 4htvo26lchkqeho1
+    """
+    random_string = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(16))
```
enrico-usai, line 155: minor you can avoid the defintion of the internal random_string variable and just have the return call.

cli/src/pcluster/config/extended_builtin_class.py
```diff
@@ -51,3 +51,13 @@ def __init__(self, b: bool, implied: bool = True):
...
+
+
+class MarkedTag(object):
+    """Custom class to add more attributes to the tag."""
+
+    def __init__(self, key: str, value: str, implied: bool = True):
+        """Initialize extended tag."""
+        self.key = key
+        self.value = value
+        self.implied = implied
```
hanwen-pcluste, line 63: I am not sure about this. If in the future we add more attributes and methods to `Tag` class, there will be inconsistencies between the `Tag` and `MarkedTag`.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -0,0 +1,166 @@
...
+        self.parameter_store_name_output = parameter_store_name_output
+        self.root_volume = root_volume
+        self.components = components
+        self._default_value()
+        # TODO: add validator
+
+    def _default_value(self):
+        if not self.tags:
+            self.tags = []
+        self.tags.append(MarkedTag("PclusterVersion", "".join(utils.get_installed_version())))
```
hanwen-pcluste, line 77: Another option is:
`default_tag = Tag("PclusterVersion", "".join(utils.get_installed_version())`
`default_tag.implied = True`
`self.tags.append(default_tag)`
I couldn't do it with built-in classes because Python doesn't allow adding more attribute to build-in classes. But here it is reasonable to utilize the flexibility of Python to just add the `implied` attribute here.
`

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -0,0 +1,166 @@
...
+        self.terminate_instance_on_failure = terminate_instance_on_failure
+        # TODO: add validator
+
+    def _default_value(self):
+        if not self.update_os_and_reboot:
+            self.update_os_and_reboot = MarkedBool(False)
+        if not self.disable_pcluster_component:
+            self.disable_pcluster_component = MarkedBool(False)
+        if not self.terminate_instance_on_failure:
+            self.terminate_instance_on_failure = MarkedBool(True)
```
hanwen-pcluste, line 147: Is it a bug? What if `self.terminate_instance_on_failure` is False? I assume in your code, the attribute will always be `True`?

yuleiwan: yes, it is a bug, not `self.terminate_instance_on_failure` should be `self.terminate_instance_on_failure` is `None`

yuleiwan: Changed

cli/src/pcluster/models/imagebuilder.py
yuleiwan: good idea, done

cli/src/pcluster/config/extended_builtin_class.py
yuleiwan: use Tag with extra attribute `implied` rather than MarkedTag

cli/src/pcluster/utils.py
yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Removed

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Removed

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: Done

cli/src/pcluster/utils.py
yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -0,0 +1,169 @@
...
+        self.parameter_store_name_output = parameter_store_name_output
+        self.root_volume = root_volume
+        self.components = components
+        self._set_default()
+        # TODO: add validator
+
+    def _set_default(self):
+        if self.tags is None:
+            self.tags = []
+        default_tag = Tag("PclusterVersion", "".join(utils.get_installed_version()))
```
enrico-usai, line 77: Why do you need to join the installed version. Is it not a string?

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,177 @@
...
+        # Define ami build instance ebs
+        if image.root_volume:
+            ebs = imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                volume_size=image.root_volume.size
+                or ami_info[0].get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize") + 15,
+                volume_type="gp2",
+            )
+        else:
+            ebs = imagebuilder.CfnImageRecipe.EbsInstanceBlockDeviceSpecificationProperty(
+                volume_size=ami_info[0].get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize") + 15,
```
enrico-usai, line 106: Let's define a constant for this [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)).

yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Mistake, it should be a string, changed

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -0,0 +1,167 @@
...
+        self,
+        image: Image,
+        build: Build,
+        dev_settings: DevSettings = None,
+    ):
+        super().__init__()
+        self.image = image
+        self.build = build
+        self.dev_settings = dev_settings
+        self._add_validator(BaseAMIValidator, ami_id=self.build.parent_image)
```
lukeseawalker, line 167: The validators should be defined next to where the property is stored, i.e. in this case we are validating the `build.parent_image`, so the validator should be placed in the Build class.
Instead if the validator is checking a combination of properties from different sections, this is the right place where to define it

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -0,0 +1,186 @@
...
+            "UpdateAndReboot",
+            type="String",
+            default=str.lower(str(bool(dev_settings.update_os_and_reboot))),
+            description="UpdateAndReboot",
+        )
+        ami_info = utils.get_info_for_amis([build.parent_image])
+        architecture = ami_info[0].get("Architecture")
+        if build.instance_type:
+            instance_types = [build.instance_type]
+        else:
```
lukeseawalker, line 59: The `build.instance_type` is a required param. There should be no default. Instead there should be a validator to check if the instance type used is compatible with the parent image

lukeseawalker (approved):

PR #2290, opened 2020-12-07
Enable EFA on all supported OSs but Centos8

ddeidda:
The new EFA installer provides the EFA kmod for all supported OSs except for Centos8. This commit adds a validator to prevent EFA from being enabled on ARM architectures with Centos8.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2359, opened 2021-01-12
Fix EFA OS architecture validator

enrico-usai:
The `efa_os_arch_validator` has been introduced in the 2.10.1 version, with patch #2290 

The configuration files with the following combination of settings are affected by the issue: `scheduler=slurm`, `base_os=centos8`, `queue_settings` specified in the file, arm instances and `enable_efa` not set in the `cluster` section or set to `false` in the `queue` section.

The error message is coming from the validation of the `enable_efa` parameter in the `queue` section, because this parameter has a default value so the validator is always executed.

The code of the validator is not checking if the value is set to true so it is failing in any case.

## Tests

Added two unit test to verify the cases on which there is no value for `enable_efa` in the `cluster` section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+2.10.2
+------
+
+**BUG FIXES**
+
+- Fix `enable_efa` parameter validation when using Centos8 and Slurm or ARM instances.
+
+
 2.10.1
 ------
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+2.10.2
+------
+
+**BUG FIXES**
+
+- Fix `enable_efa` parameter validation when using Centos8 and Slurm.
```
ddeidda: On ARM instances.

ddeidda (approved):

PR #2363, opened 2021-01-13
integ-tests: FSx Storage Capacity/File Chunk Size/Maintenance Start Time

yuleiwan:
* Ensure FSx File Chunk Size/Maintenance Start Time retrieved from cluster are same as config file setting
* Ensure deviation of FSx storage capacity is less than 0.2 between the storage size in cluster and config file

### DIFF
--- a/tests/integration-tests/tests/storage/test_fsx_lustre.py
+++ b/tests/integration-tests/tests/storage/test_fsx_lustre.py
@@ -43,14 +43,15 @@ MAX_MINUTES_TO_WAIT_FOR_BACKUP_COMPLETION = 7
         "storage_type",
         "drive_cache_type",
         "storage_capacity",
+        "imported_file_chunk_size",
     ),
     [
-        ("PERSISTENT_1", 200, "NEW_CHANGED", None, None, 1200),
-        ("SCRATCH_1", None, "NEW", None, None, 1200),
-        ("SCRATCH_2", None, None, None, None, 1200),
-        ("PERSISTENT_1", 200, None, "SSD", None, 1200),
-        ("PERSISTENT_1", 40, None, "HDD", None, 1800),
-        ("PERSISTENT_1", 12, None, "HDD", "READ", 6000),
+        ("PERSISTENT_1", 200, "NEW_CHANGED", None, None, 1200, 1024),
+        ("SCRATCH_1", None, "NEW", None, None, 1200, 1024),
+        ("SCRATCH_2", None, None, None, None, 1200, 1024),
+        ("PERSISTENT_1", 200, None, "SSD", None, 1200, 2048),
+        ("PERSISTENT_1", 40, None, "HDD", None, 1800, 512),
+        ("PERSISTENT_1", 12, None, "HDD", "READ", 6000, 1024),
     ],
 )
 @pytest.mark.regions(["eu-west-1"])
@@ -71,11 +72,13 @@ def test_fsx_lustre_configuration_options(
     storage_type,
     drive_cache_type,
     storage_capacity,
+    imported_file_chunk_size,
 ):
     mount_dir = "/fsx_mount_dir"
     bucket_name = s3_bucket_factory()
     bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
     bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
+    weekly_maintenance_start_time = (datetime.datetime.utcnow() + datetime.timedelta(minutes=60)).strftime("%w:%H:%M")
     cluster_config = pcluster_config_reader(
         bucket_name=bucket_name,
         mount_dir=mount_dir,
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -520,6 +542,50 @@ def _test_data_repository_task(remote_command_executor, mount_dir, bucket_name,
     assert_that(file_permissions).is_equal_to("0100777")
 
 
+def _test_storage_capacity(remote_command_executor, mount_dir, storage_capacity):
+    logging.info("Test FSx storage capacity")
+    if storage_capacity:
+        result = remote_command_executor.run_remote_command(
+            f"df -BG | grep {mount_dir}" + "| awk '{print$2}' | tr -d -c 0-9"
+        )
+        assert_that((storage_capacity - int(result.stdout)) // storage_capacity).is_less_than(0.2)
```
rexcsn, line 551: Might be worth adding some explanation in comments whenever we have hardcoded tolerances like this. Any documentation explaining this? Will this tolerance be the same for all FSx storage type?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -144,6 +160,9 @@ def _test_fsx_lustre(
     storage_type=None,
     auto_import_policy=None,
     deployment_type=None,
+    weekly_maintenance_start_time=None,
+    imported_file_chunk_size=None,
+    storage_capacity=None,
```
rexcsn, line 165: This `_test_fsx_lustre` function is getting very complicated and these default parameters are getting a bit out of hand here :sweat_smile:

IMO we should extract the common assertions also used by other test into a separate function like `_assert_basic_fsx_functions` and rename this function to something like `_assert_fsx_configuration_options` and get rid of all the default `=None`

The issue I see with `=None` here is we don't really need them here, and having them here will not produce an obvious failure if we decreased test coverage, because no error will be shown if we remove any of these parameters by accident

This way you can also get rid of all the conditions within your assertion functions like `if storage_capacity:`

The bottom line is we should always be enforcing our assertions strictly in our tests. If things change, we would want to see an error rather than not being aware of the issue.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -520,6 +542,50 @@ def _test_data_repository_task(remote_command_executor, mount_dir, bucket_name,
     assert_that(file_permissions).is_equal_to("0100777")
 
 
+def _test_storage_capacity(remote_command_executor, mount_dir, storage_capacity):
+    logging.info("Test FSx storage capacity")
+    if storage_capacity:
```
rexcsn, line 547: Can we remove these? These conditions will fail silently if we decreased coverage by accident

Please double check and change for all other assertions as well

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -520,6 +542,50 @@ def _test_data_repository_task(remote_command_executor, mount_dir, bucket_name,
...
+def _test_weekly_maintenance_start_time(fsx_fs_id, weekly_maintenance_start_time, region):
+    logging.info("Test FSx weekly maintenance start time setting")
+    if weekly_maintenance_start_time:
+        assert_that(get_weekly_maintenance_start_time(fsx_fs_id, region)).is_equal_to(weekly_maintenance_start_time)
+
+
+def get_imported_chunch_size(fsx_fs_id, region):
+    logging.info("Getting ImportedFileChunkSize from DescribeFilesystem API.")
+    return (
+        boto3.client("fsx", region_name=region)
```
rexcsn, line 569: Can we just describe the file system once and parse different info out of the result in different tests?

Please double check for other assertions/test cases that may need to describe the file system as well

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -157,6 +176,9 @@ def _test_fsx_lustre(
     _test_export_path(remote_command_executor, mount_dir, bucket_name, region)
     _test_auto_import(auto_import_policy, remote_command_executor, mount_dir, bucket_name, region)
     _test_data_repository_task(remote_command_executor, mount_dir, bucket_name, fsx_fs_id, region)
+    _test_storage_capacity(remote_command_executor, mount_dir, storage_capacity)
```
rexcsn, line 179: We will want to include `_test_storage_capacity` for our tests that are specifying, but currently not testing storage capacity, like `test_fsx_lustre`.

Please double check for all existing tests

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
yuleiwan: add a comment to explain it

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -85,13 +88,58 @@ def test_fsx_lustre_configuration_options(
...
+    storage_capacity,
+):
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    fsx_fs_id = get_fsx_fs_id(cluster, region)
+    fsx = boto3.client("fsx", region_name=region).describe_file_systems(FileSystemIds=[fsx_fs_id])
+
+    _test_fsx_lustre_correctly_mounted(remote_command_executor, mount_dir, os, region, fsx_fs_id)
+    _test_import_path(remote_command_executor, mount_dir)
+    _test_fsx_lustre_correctly_shared(scheduler_commands, remote_command_executor, mount_dir)
```
rexcsn, line 132: Minor: instead of rewriting these common test cases can we just call `_test_fsx_lustre`?

yuleiwan: Done

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -520,6 +550,41 @@ def _test_data_repository_task(remote_command_executor, mount_dir, bucket_name,
...
 
 
+def _test_storage_capacity(remote_command_executor, mount_dir, storage_capacity):
+    logging.info("Test FSx storage capacity")
+    # Get the storage size of mount_dir with GB, the storage size is slightly less than storage capacity defined,
+    # here we set 0.2 as the difference ratio threshold
+    result = remote_command_executor.run_remote_command(
+        f"df -BG | grep {mount_dir}" + "| awk '{print$2}' | tr -d -c 0-9"
+    )
+    assert_that((storage_capacity - int(result.stdout)) // storage_capacity).is_less_than(0.2)
```
rexcsn, line 560: Incorrect use of `//` here? `1//2 = 0`

yuleiwan: Done

rexcsn (approved):
LGTM! Feel free to merge after successful test run.

PR #2364, opened 2021-01-15
Support cluster creation without EC2 key pair

ddeidda:
This commit introduces the support for a cluster configuration where no EC2 key pair
is used. Such configuration is possible only if  different authentication methods are
setup in the provided AMI, so we raise a warning when this condition is detected.

### DIFF
--- a/cli/src/pcluster/config/validators.py
+++ b/cli/src/pcluster/config/validators.py
@@ -960,6 +960,12 @@ def cluster_validator(section_key, section_label, pcluster_config):
         if max_size < min_size:
             errors.append("max_queue_size must be greater than or equal to initial_queue_size")
 
+    if not section.get_param_value("key_name"):
+        warnings.append(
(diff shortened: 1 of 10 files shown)


cli/src/pcluster/config/validators.py
```diff
@@ -960,6 +960,12 @@ def cluster_validator(section_key, section_label, pcluster_config):
         if max_size < min_size:
             errors.append("max_queue_size must be greater than or equal to initial_queue_size")
 
+    if not section.get_param_value("key_name"):
```
enrico-usai, line 963: I'd define a key_pair validator, it's a self contained validation. To avoid having too many validators in the section.

ddeidda: The reason I did it here is because validators are not called on None values.

enrico-usai: ah, good reason

enrico-usai (approved):

enrico-usai (approved):

PR #2369, opened 2021-01-19
Fix update policy for max_queue_size parameter

ddeidda:
When the value of the max_queue_size parameter is updated we must check that no running nodes will be shut down.


rexcsn (approved):
LGTM!

PR #2376, opened 2021-01-20
integ-tests: test maintain_initial_queue_size

chenwany:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/integration-tests/tests/scaling/test_scaling.py
+++ b/tests/integration-tests/tests/scaling/test_scaling.py
@@ -36,7 +36,11 @@ from tests.common.hit_common import (
     submit_initial_job,
     wait_for_num_nodes_in_scheduler,
 )
-from tests.common.scaling_common import get_compute_nodes_allocation, get_desired_asg_capacity
+from tests.common.scaling_common import (
+    get_compute_nodes_allocation,
+    get_desired_asg_capacity,
+    test_maintain_initial_size,
+)
 from tests.common.schedulers_common import get_scheduler_commands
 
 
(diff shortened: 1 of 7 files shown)


tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -121,6 +125,8 @@ def test_nodewatcher_terminates_failing_node(scheduler, region, pcluster_config_
     )
 
     assert_no_errors_in_logs(remote_command_executor, scheduler)
+    maintain_initial_size = cluster.config.get("cluster default", "maintain_initial_size")
```
rexcsn, line 128: I mentioned this in a previous PR as well, don't like the general approach of how we are retrieving parameter values like `maintain_initial_size` from the config and using them in our test, because we never really enforce the coverage of the test. We should only do this if the parameter value is not related to test coverage.

Right now `test_scaling` has `maintain_initial_size = true` in the config and `test_sge` and `test_torque` has `maintain_initial_size = false` in the configs, so we are able to cover both cases today.
However, if we changed `maintain_initial_size = false` in `test_scaling` tomorrow, all the test will still pass, but we essentially have less test coverage because none of our test will cover the case of `maintain_initial_size = true`, and it would be difficult for us to notice this.

In this case it would be preferred to hard code things, or declare parameter values in the test explicitly, because then we would know exactly what we are testing for

rexcsn (commented):
Looks like we are only testing `maintain_initial_queue_size` because we are never asserting that `initial_queue_size` specified is consistent with what's in ASG.
I would change the description of this PR/commit to test `maintain_initial_queue_size` only, or add additional assertions to make sure `initial_queue_size` corresponds to the desired capacity in ASG

chenwany:
> Looks like we are only testing `maintain_initial_queue_size` because we are never asserting that `initial_queue_size` specified is consistent with what's in ASG.
> I would change the description of this PR/commit to test `maintain_initial_queue_size` only, or add additional assertions to make sure `initial_queue_size` corresponds to the desired capacity in ASG

Change it into test `maintain_initial_queue_size`, the `initial_queue_size` is already tested elsewhere.

tests/integration-tests/tests/scaling/test_scaling.py
chenwany: Oh, I see! Done

rexcsn (approved):

PR #2379, opened 2021-01-21
3.0: Collapse "config" layer into the "model" one

enrico-usai:
All the `*Config` classes will be renamed to `*Resource` classes and placed in the `models` folder rather than the `config` one.

The idea is that by parsing the items from the configuration file, through the schema, we'll obtain directly the core items of the model. We don't need a separate layer for the config.

We're instead moving the logic to manage the model classes (e.g. Cluster) in the ClusterManager class that will permit to execute all the actions on the model classes.

## Other changes
* Moved Cluster Stack CDK code in a separate `cluster_stack` module
* Aligned tests to the previous changes
* Removed useless storage module. We'll read add it when/if needed.


cli/src/pcluster/models/cluster.py
```diff
@@ -453,36 +454,37 @@ def __init__(
...
+class Logs(Resource):
+    """Represent the CloudWatch Logs configuration."""
 
     def __init__(
         self,
-        cloud_watch: CloudWatchLogsConfig = None,
+        cloud_watch: CloudWatchLogs = None,
     ):
         super().__init__()
+        # FIXME
```
hanwen-pcluste, line 465: Minor: what is to be fixed?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -140,8 +142,8 @@ class BaseEbsSchema(BaseSchema):
 
     @post_load
     def make_config(self, data, **kwargs):
```
hanwen-pcluste, line 144: Minor: rename `make_config` to `make_resource`?

hanwen-pcluste (approved):
LGTM

PR #2380, opened 2021-01-21
3.0: Refactor storage classes hierarchy and make it independent by schema

enrico-usai:
Before this patch we had a 1:1 conversion between schema and model layers but some Schema classes don't make sense in the model.

For example we don't need `SharedStorage` class containing EBS or EFS or FSx because EBS, EFS and FSx _are_ shared storage items.

## New classes hierarchy

* `Ebs` representing a generic ebs volume (to be used for root volume or shared ebs)
* `SharedStorage` representing a generic shared storage that can be of different types and just have the `mount_dir` as attribute.
* `SharedEfs`, `SharedFsx` and `SharedEbs` are inheriting from `SharedStorage` and adding specific attributes.
* `SharedEbs` inherits from both `SharedStorage` and `Ebs`.
* Defined `SharedStorage.Type` enum.

## Schema changes

* In the `SharedStorageSchema` we're now instantiating a specific `SharedStorage` according to the type found in the config file rather than creating a `SharedStorage` class that was totally useless from the model point of view.
  As an example we're collapsing the `SharedStorage` and `EFS` information into a single object.
* Clearly when dumping the `SharedStorage` we need to restore the fields with a `pre_dump` action.
* For this reason the Efs|Ebs|Fsx Schemas don't need to generate any object. It will be done by the `SharedStorageSchema`.

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -31,7 +31,6 @@ from marshmallow import (
 from pcluster.constants import FSX_HDD_THROUGHPUT, FSX_SSD_THROUGHPUT, SUPPORTED_ARCHITECTURES
 from pcluster.models.cluster import (
     AdditionalIamPolicy,
-    BaseEbs,
     CloudWatchDashboards,
     CloudWatchLogs,
     Cluster,
@@ -40,9 +39,7 @@ from pcluster.models.cluster import (
     Dashboards,
     Dcv,
     Ebs,
-    Efs,
     EphemeralVolume,
-    Fsx,
     HeadNode,
     HeadNodeNetworking,
     Iam,
@@ -58,7 +55,9 @@ from pcluster.models.cluster import (
     S3Access,
     Scheduling,
     SchedulingSettings,
-    SharedStorage,
+    SharedEbs,
+    SharedEfs,
+    SharedFsx,
     Ssh,
     Storage,
     Tag,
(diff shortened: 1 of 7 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -140,10 +139,14 @@ class BaseEbsSchema(BaseSchema):
     throughput = fields.Int()
     encrypted = fields.Bool()
 
+
+class RootVolumeSchema(_BaseEbsSchema):
+    """Represent the RootVolume schema."""
+
```
hanwen-pcluste, line 145: Why creating a new class if the class is exactly the same as BaseEbsSchema?

hanwen-pcluste (approved):
General questions: Could you explain any obvious benefit we will enjoy after this change? I am not sure because the new approach increases the amount of code. And by dropping the 1:1 mapping from schema to object, we are writing more `@post_load` and `@pre_dump`.
Anyway, I am fine with applying this change.

cli/src/pcluster/schemas/cluster_schema.py
enrico-usai: The two classes have the same attributes but different methods. I.e. `_BaseEbsSchema` doesn't have the `post_load` method.

enrico-usai:
The main benefit is to simplify the model and match to the resources we really need rather than the configuration file.
In future we can add more levels/items in the configuration file tree but we don't want to change the model for every change in the file, the two layers must be decoupled.

We're adding complexity in the Schema layer but it will help to remove complexity from the model.

PR #2383, opened 2021-01-22
Add test for status check failures

rexcsn:
* Add new integration test for EC2 status check failures to make sure nodes with failure are replaced


tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -184,6 +185,33 @@ def test_hit_scaling(scheduler, region, instance, pcluster_config_reader, cluste
...
+    assert_no_errors_in_logs(remote_command_executor, scheduler)
+
+
+@pytest.mark.regions(["us-west-1"])
+@pytest.mark.instances(["c5.xlarge"])
+@pytest.mark.schedulers(["slurm"])
+@pytest.mark.oss(["alinux2"])
+@pytest.mark.usefixtures("region", "os")
+@pytest.mark.hit_rare_failure
+def test_hit_rare_failure(scheduler, region, instance, pcluster_config_reader, clusters_factory, test_datadir):
```
demartinofra, line 197: shall we move these 2 slurm tests (test_hit_scaling, test_hit_rare_failure) to the slurm specific module and rename them to something like: test_slurm_scaling and test_error_handling?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -184,6 +185,33 @@ def test_hit_scaling(scheduler, region, instance, pcluster_config_reader, cluste
...
+    """Test that slurm-specific scaling logic can handle rare failures."""
+    cluster_config = pcluster_config_reader(scaledown_idletime=3)
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+
+    _assert_cluster_initial_conditions(scheduler_commands, instance, 10, 10, 1, 1)
```
demartinofra, line 204: why are we asserting this here and also in _test_status_check_replacement after having submitted a job?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -355,6 +390,57 @@ def _test_keep_or_replace_suspended_nodes(
...
+    cluster_name,
+    region,
+    partition,
+    num_static_nodes,
+    num_dynamic_nodes,
+    dynamic_instance_type,
+):
+    """Test nodes failing static check are correctly replaced."""
+    logging.info("Testing that nodes failing static check are correctly replaced")
+    job_id = submit_initial_job(
```
demartinofra, line 405: do we need this job?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -355,6 +390,57 @@ def _test_keep_or_replace_suspended_nodes(
...
+    )
+    static_nodes, dynamic_nodes = assert_initial_conditions(
+        scheduler_commands, num_static_nodes, num_dynamic_nodes, partition, job_id
+    )
+    # Get network interface name from Head node, assuming Head node and Compute are of the same instance type
+    interface_name = remote_command_executor.run_remote_command(
+        "nmcli device status | grep ether | awk '{print $1}'"
+    ).stdout
+    logging.info("Detaching network interface {} on Compute nodes".format(interface_name))
+    # Submit job that will detach network interface on all static nodes, this will cause EC2 status check to fail
```
demartinofra, line 421: is this true? it looks like we are only detaching the dynamic one

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -355,6 +390,57 @@ def _test_keep_or_replace_suspended_nodes(
...
+    kill_job_id = scheduler_commands.submit_command_and_assert_job_accepted(
+        submit_command_args={
+            "command": "sudo ifconfig {} down && sleep 600".format(interface_name),
+            "partition": partition,
+            "constraint": "dynamic",
+            # Run array job that requires 3 nodes
```
demartinofra, line 427: probably needs to be updated

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -184,6 +185,33 @@ def test_hit_scaling(scheduler, region, instance, pcluster_config_reader, cluste
...
+@pytest.mark.hit_rare_failure
+def test_hit_rare_failure(scheduler, region, instance, pcluster_config_reader, clusters_factory, test_datadir):
+    """Test that slurm-specific scaling logic can handle rare failures."""
+    cluster_config = pcluster_config_reader(scaledown_idletime=3)
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+
+    _assert_cluster_initial_conditions(scheduler_commands, instance, 10, 10, 1, 1)
+    _test_status_check_replacement(
```
demartinofra, line 205: this test probably is not using the static node. Is the node used by the next one or we can remove it?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -184,6 +185,33 @@ def test_hit_scaling(scheduler, region, instance, pcluster_config_reader, cluste
         num_dynamic_nodes=3,
         dynamic_instance_type=instance,
     )
+    assert_no_errors_in_logs(remote_command_executor, scheduler)
+
+
+@pytest.mark.regions(["us-west-1"])
```
demartinofra, line 191: we can get rid of all these annotations

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: Done

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: This is to assert that cluster creation behaves as expected, and all min/initial/max are configured as expected in queue.
We can remove if you think this is not necessary, but I don't think we have dedicated test for HIT creation?

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: Static instance is used in `_test_clustermgtd_down_logic`

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: Probably not if we just use the exact setup at cluster creation. However, having this here can help us easily re-organize different test cases, and can also be useful when running the test on an existing cluster.

If setup at cluster creation is identical to the nodes requested here(which it is currently) this step should be cheap.

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: Done

tests/integration-tests/tests/scaling/test_scaling.py
rexcsn: Done

demartinofra (approved):

PR #2394, opened 2021-01-26
3.0: ImageBuilder Validators and CLI refactor

yuleiwan:
### Validators
-syntactic validator
* image::name
* image::tags
* build::components:component
* build::parent_image
* build::security_group_ids
* build::subnet_id
* dev_settings::chef_cookbook:json
* dev_settings::distribution_configuration_arn

-semantic validator
* build::BaseAMIValidator
* build::InstanceTypeBaseAMICompatibleValidator
* build::InstanceTypeValidator(share with CLI)
* image::volume::EBSVolumeKmsKeyIdValidator
* imagebuilder::EbsVolumeTypeSizeValidator(share with CLI)
* dev_settings::UrlValidator(share with CLI)

### ImageBuilder CLI refactor
* Initialize the default root volume inside imagebuilder config
* parent_ami supports both ami id and image arn as input value, implement the get_info_for_ami_from_arn and get_ami_id_from_parent_image in utils
* Remove default instance type, instance type is a required config value
* Add more unit tests for imagebuilder cli and cdk
* Use Param to wrap values in config


cli/src/pcluster/models/imagebuilder.py
```diff
@@ -150,4 +170,23 @@ def __init__(
         self.image = image
         self.build = build
         self.dev_settings = dev_settings
-        self._add_validator(BaseAMIValidator, ami_id=self.build.parent_image)
+        self._set_default()
+        self._add_validator(
```
enrico-usai, line 174: This validator should stay in the root volume given that we're validating a root volume's field.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -150,4 +170,23 @@ def __init__(
...
+    def _set_default(self):
+        # set default root volume
+        if self.image.root_volume is None or self.image.root_volume.size.value is None:
+            increase_volume_size = 15
+            ami_id = utils.get_ami_id_from_parent_image(self.build.parent_image.value)
+            ami_info = utils.get_info_for_amis([ami_id])
+            default_root_volume_size = (
+                ami_info[0].get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize") + increase_volume_size
+            )
+            if self.image.root_volume is None:
```
enrico-usai, line 187: this case `if self.image.root_volume is None` it's in the first if block too.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -150,4 +170,23 @@ def __init__(
...
         self.dev_settings = dev_settings
-        self._add_validator(BaseAMIValidator, ami_id=self.build.parent_image)
+        self._set_default()
+        self._add_validator(
+            EbsVolumeTypeSizeValidator, priority=10, volume_type=Param("gp2"), volume_size=self.image.root_volume.size
+        )
+
+    def _set_default(self):
+        # set default root volume
+        if self.image.root_volume is None or self.image.root_volume.size.value is None:
```
enrico-usai, line 180: This use case is very interesting because we're setting a default by using the values coming from other fields. 

I think this logic should stay in the Volume because it's the responsible to set the default but it's not yet clear to me how to propagate the required info from the build.parent_image.

Let's add a FIXME for now and let's think to some alternatives.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -150,4 +170,23 @@ def __init__(
...
+            increase_volume_size = 15
+            ami_id = utils.get_ami_id_from_parent_image(self.build.parent_image.value)
+            ami_info = utils.get_info_for_amis([ami_id])
+            default_root_volume_size = (
+                ami_info[0].get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize") + increase_volume_size
+            )
+            if self.image.root_volume is None:
+                default_root_volume = Volume(size=default_root_volume_size)
+                default_root_volume.implied = True
```
enrico-usai, line 189: We don't need to use the implied for objects, it's only useful for the params, so the size in this case. You should do: `volume.size.implied = True` instead. 
If we're able to move this intiialization in the Volume as mentioned before the set of the implied would be automatic.

cli/tests/pcluster/validators/test_ebs.py
```diff
@@ -123,3 +124,27 @@ def test_ebs_size_iops_validators(section_dict, expected_message):
...
+                "KmsKeyId": "arn:aws:kms:us-east-1:012345678910:1234abcd-12ab-34cd-56ef-1234567890ab",
+                "Encrypted": False,
+            },
+            "KmsKeyId arn:aws:kms:us-east-1:012345678910:1234abcd-12ab-34cd-56ef-1234567890ab "
+            "is specified, the encrypted state can't be False.",
+        ),
+    ],
+)
+def test_ebs_volume_kms_key_id_validator(volume_dict, expected_message):
+    actual_failures = VolumeSchema().load(volume_dict).validate()
```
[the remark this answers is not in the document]
hanwen-pcluste: I would have more discussion on this. Since Marshmallow allows partial load, we don't have to load all required fields when testing. Moreover, because we are handling multiple error messages, we would like to make sure all validators in one schema/model work well together. I am trying to prove that with just a little sacrifice of the granularity of the unit tests, we are getting much more test coverage and better simulation of the real use case.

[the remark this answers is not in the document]
yuleiwan: Here I agree with @enrico-usai,  we still need to test Validator directly. For @hanwen-pcluste  case, can we put this in model unit tests for  handling multiple error messages? It would make sense that we test each unit in corresponding module.

cli/tests/pcluster/validators/test_s3.py
```diff
@@ -0,0 +1,32 @@
...
+        (
+            "fake://test/cookbook.tgz",
+            "ChefCookbook",
+            False,
+            "The value 'fake://test/cookbook.tgz' used for the parameter 'ChefCookbook' is not a valid URL, "
+            "choose URL with 'https', 's3' or 'file' prefix.",
+        ),
+    ],
+)
+def test_url_validator(mocker, url, name, response, expected_message):
```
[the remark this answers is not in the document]
yuleiwan: Done

cli/src/pcluster/utils.py
```diff
@@ -926,6 +930,20 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
+def get_ami_id(parent_image):
+    """Get ami id from parent image."""
+    if parent_image.startswith("arn"):
+        ami_id = get_info_for_ami_from_arn(parent_image).get("image").get("outputResources").get("amis")[0].get("image")
```
enrico-usai, line 936: I think we should hide this series of getters in the ImageBuilderClient() by creating a method like:
`ImageBuilderClient().get_image_id(image_arn)`

because with the current code the caller (get_ami_id function) need to know all the structure of the boto3 answer, instead this information should be encapsulated in the ImageBuilderClient.

If needed in future we can have other methods exposed by the Client to have other information.

The other reason is that you should check for the fields in the response.
by doing `.get("image").get("outputResources").get("amis")[0].get("image")` you can get some KeyError if the ami is not found or the output format is not the expected one.

In the ImageBuilderClient().get_image_id you can do something like:
`.get("image", []).get("outputResources", []).get("amis", [])` by managing and checking all the errors and then return an empty string if the ami is not found.

cli/src/pcluster/validators/s3_validators.py
```diff
@@ -0,0 +1,68 @@
...
+                        FailureLevel.WARNING,
+                    )
+                except ValueError:
+                    self._add_failure(
+                        ("The value '{0}' is not a valid URL").format(url.value),
+                        FailureLevel.ERROR,
+                    )
+        else:
+            self._add_failure(
+                ("The value '{0}' is not a valid URL, " "choose URL with 'https', 's3' or 'file' prefix.").format(
```
enrico-usai, line 44: minor, there are etra quotes here: `, " "choose` probably due to autoformat.

cli/tests/pcluster/validators/test_ebs.py
```diff
@@ -123,3 +125,29 @@ def test_ebs_size_iops_validators(section_dict, expected_message):
 def test_ebs_volume_type_size_validator(section_dict, expected_message):
     actual_failures = SharedStorageSchema().load({"MountDir": "/my/mount/point", "EBS": section_dict}).validate()
     assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
+    "volume_dict, expected_message",
```
enrico-usai, line 131: instead of defining the volume_dict you can directly use kms_key_id and encrypted as test parameters.
To test the validator you don't need the entire volume structure, you just need the two parameters to be tested.

enrico-usai: So you can avoid to define the Size.

enrico-usai (commented):
You're doing a great job.

cli/tests/pcluster/validators/test_ebs.py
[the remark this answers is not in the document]
enrico-usai: In unit tests individual _units or components_ of a software must be tested. The purpose is to validate that each unit of the code performs as expected, in this case we need to verify that the logic of the validator is the right one, according with the different values of the parameters.

If we put in the game marshmallow we are not just testing if the logic of the validator but we're also testing marshmallow loading, the schema definition, the partial loading of marshmallo, we're putting in the game the logic for implied and default values etc.

The simulation of the real case can be done by adding less granularity unit test like you're suggesting (or integration tests) but we cannot avoid the unit tests for the single validators.

cli/src/pcluster/utils.py
yuleiwan: Done

cli/src/pcluster/validators/s3_validators.py
yuleiwan: Removed

cli/tests/pcluster/validators/test_ebs.py
yuleiwan: Done

cli/tests/pcluster/validators/test_ebs.py
```diff
@@ -201,3 +202,27 @@ def test_ebs_volume_size_snapshot_validator(
...
+
+
+@pytest.mark.parametrize(
+    "kms_key_id, encrypted, expected_message",
+    [
+        (
+            "arn:aws:kms:us-east-1:012345678910:1234abcd-12ab-34cd-56ef-1234567890ab",
+            None,
+            "Kms Key Id arn:aws:kms:us-east-1:012345678910:1234abcd-12ab-34cd-56ef-1234567890ab "
+            "is specified, the encrypted state can't be None.",
```
enrico-usai, line 214: minor: I'd improve the message saying: "the encrypted state must be True."

enrico-usai (approved):
I just left a minor comment but the patch is approved for me. Thanks for addressing all the comments.

cli/tests/pcluster/validators/test_ebs.py
yuleiwan: Done

PR #2395, opened 2021-01-26
3.0: Apply Param and finish migrating allowed values from mappings.py

hanwen-pcluste:
Commit 1:
1. To append more info to basic classes such as str, int , bool, float, we create a class Param to wrap the value and add `implied` and `valid` fields. The reason of not wrapping object of `Resource` to avoid writing lnng sentence to access a value, e.g `cluster.head_node.value.networking.value.subnet.value`.
2. Utilize `valid` field to build the dependencies of ebs validators. Although the approach is still under discussion, I'd include it in the PR to show what can we do with marked values.
3. Delete old ebs validators after migration.
Major Concern: This approach adds noticeable overhead when when accessing value (`.value` has to be used).
Ambiguity: For List[str], should we wrap every string in the list using Param?
The reason of not using a method in the base Resource class to require validity of all input args of a validator is that we are still dealing with the ambiguity between separated EBS validators vs centrally orchestrated FSx validators.

Commit 2:
Finish migrating validations that were done by `allowed_values` in `mappings.py`


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -161,6 +170,12 @@ def make_resource(self, data, **kwargs):
         """Generate resource."""
         return Ebs(**data)
 
+    @validates("size")
+    def validate_size(self, value):
```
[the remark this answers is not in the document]
hanwen-pcluste: The limitation only applies to root volume but not shared EBS. Therefore I have to write the validator here.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -129,6 +130,14 @@ def remove_implied_values(self, data, **kwargs):
                 value[:] = [v for v in value if not _is_implied(v)]
         return data
 
+    @pre_dump
+    def unwrap_marked_class(self, data, **kwargs):
+        """Remove value implied by the code. i.e., only keep parameters that were specified in the yaml file."""
+        for key, value in vars(data).items():
+            if isinstance(value, Param):
+                setattr(data, key, value.value)
```
enrico-usai, line 138: value.value is not so nice.

let's use param_name, param_value vs key, value

cli/src/pcluster/validators/common.py
```diff
@@ -51,14 +53,24 @@ def __init__(self, raise_on_error=False):
...
         result = ValidationResult(message, level)
+        for field in fields:
+            field.valid = False
         if self._raise_on_error:
             raise ConfigValidationError(result)
         else:
             self._failures.append(result)
 
+    def __call__(self, *arg, **kwargs):
+        """Must be implemented with specific validation logic."""
```
enrico-usai, line 66: minor: comment out-date

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: Yea, so weird. But before `isinstance(value, Param)`, we can't be sure if we can call it `param`.
In other words, I would prefer wording weirdness than logic weirdness

cli/src/pcluster/validators/common.py
hanwen-pcluste: Done

enrico-usai (approved):

PR #2398, opened 2021-01-28
3.0: Split, move, clean up FSx validators (lustre and architecture related are not touched)

hanwen-pcluste:
## Split FsxValidator in fsx_validators.py
The original approach was to have to a central method to orchestrate all validations of FSx section. We decided to change the approach to split the validator into smaller ones to be consistent with other validators and facilitate unit testing.

## Remove moved FSx validators from mappings.py
Starting from #2405, we decided to delete the migrated validators in the old code. This commit will delete `fsx_id_validator`, `fsx_ignored_parameters_validator`, `fsx_imported_file_chunk_size_validator`, `fsx_storage_capacity_validator`, `fsx_validator` in mappings.py

## Minor changes
1. Prevent validation of iops/volume_size ratio if validation of absolute iops value fails.
2. In the base Validator class, return the `_failures` explicitly so that the return of `_failures` at the end of each `validate()` is not required

### DIFF
--- a/cli/src/pcluster/validators/ebs_validators.py
+++ b/cli/src/pcluster/validators/ebs_validators.py
@@ -121,7 +121,7 @@ class EbsVolumeIopsValidator(Validator):
                     FailureLevel.ERROR,
                     [volume_iops],
                 )
-            if (
+            elif (
                 volume_iops_value
                 and volume_iops_value > volume_size_value * EBS_VOLUME_TYPE_TO_IOPS_RATIO[volume_type_value]
             ):
--- a/cli/src/pcluster/config/mappings.py
+++ b/cli/src/pcluster/config/mappings.py
@@ -81,13 +81,8 @@ from pcluster.config.validators import (
     efs_validator,
     extra_json_validator,
     fsx_architecture_os_validator,
-    fsx_id_validator,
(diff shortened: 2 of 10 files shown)


cli/src/pcluster/validators/ebs_validators.py
```diff
@@ -122,6 +116,7 @@ def validate(self, volume_type, volume_size, volume_iops):
                     FailureLevel.ERROR,
                     [volume_iops],
                 )
+                return
```
enrico-usai, line 119: Instead of returning here I think we can convert the next if into an elsif

cli/src/pcluster/models/param.py
```diff
@@ -27,3 +27,21 @@ def __init__(self, value, default=None, valid: bool = True):
             self.value = value
             self.implied = False
         self.valid = valid
+
+    def __bool__(self):
```
enrico-usai, line 31: I'm not sure about these facilities because sometimes you need to call the param.value but for comparison you don't need it, I'd prefer an unambiguous behaviour.

hanwen-pcluste: Done

enrico-usai (approved):

PR #2399, opened 2021-01-28
3.0: Add model and schema for AWS Batch

enrico-usai:
* Added two different hierarchies for `ComputeResource` and `SchedulingSettings`
* Restructured `Scheduling` section in the file to have the scheduler name as child.
  This change permits to easily validate the different schemas and build different objects.

## Other changes

* Renamed `type` to `raid_type` in `Raid` resource and define it as `int`.
* Renamed `id` to `file_system_id` in `EphemeralVolume`.
* Removed `image` field from `QueueSchema`, it is not supported yet.
* Renamed everywhere `batch` to `awsbatch`.
* Added missing requirements for image-builder code.

## Tests

* Added tests for `awsbatch.full.yaml` and `awsbatch.simple.yaml` config files


hanwen-pcluste:
In model layer, you are diverging two classes from the cluster level. Why did you choose to do so instead of diverging awsbatch and slurm from the scheduling level?

cli/src/pcluster/models/cluster.py
```diff
@@ -372,58 +372,46 @@ def __init__(
...
         self.networking = networking
-        self.compute_resources = compute_resources
+        self.storage = storage
+        self.compute_type = Param(compute_type, default="ONDEMAND")
 
 
-class SchedulingSettings(Resource):
-    """Represent the Scheduling configuration."""
+class CommonSchedulingSettings(Resource):
+    """Represent the common scheduler settings."""
```
hanwen-pcluste, line 408: Minor: when should we use Common* or Base* ? And there is another case we don't use prefix at all: _ComputeResourceSchema

hanwen-pcluste (approved):

enrico-usai:
> In model layer, you are diverging two classes from the cluster level. Why did you choose to do so instead of diverging awsbatch and slurm from the scheduling level?

For now the scheduling level seems enough but in future we need to add more responsibility to the Cluster (e.g. status(), stop(), start()) and this really depend by the scheduler.

PR #2404, opened 2021-01-29
3.0: Support CDK conversion for FSX, EBS and EFS storages

ddeidda:

### DIFF
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -9,6 +9,8 @@ jinja2>=2.11.0
 marshmallow
 aws-cdk.core
 aws-cdk.aws-ec2
+aws-cdk.aws-fsx
 aws-cdk.aws-imagebuilder
 aws-cdk.aws-iam
 aws-cdk.aws-ssm
+aws-cdk.aws-efs
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -305,8 +305,8 @@ class _BaseNetworking(Resource):
     ):
         super().__init__()
         self.assign_public_ip = Param(assign_public_ip)
-        self.security_groups = security_groups
-        self.additional_security_groups = additional_security_groups
+        self.security_groups = Param(security_groups)
+        self.additional_security_groups = Param(additional_security_groups)
         self.proxy = proxy
 
     def _register_validators(self):
@@ -345,7 +345,7 @@ class QueueNetworking(_BaseNetworking):
 
     def __init__(self, subnet_ids: List[str], placement_group: PlacementGroup = None, **kwargs):
         super().__init__(**kwargs)
-        self.subnet_ids = subnet_ids
+        self.subnet_ids = Param(subnet_ids)
         self.placement_group = placement_group
 
 
--- a/cli/src/pcluster/models/cluster_slurm.py
+++ b/cli/src/pcluster/models/cluster_slurm.py
@@ -152,3 +152,28 @@ class SlurmCluster(BaseCluster):
                         os=self.image.os,
                         architecture=self.head_node.architecture,
                     )
+
+    @property
+    def compute_subnet_ids(self):
+        """Return the list of all compute subnet ids in the cluster."""
+        return list(
+            {
+                subnet_id
+                for queue in self.scheduling.queues
+                if queue.networking.subnet_ids and queue.networking.subnet_ids.value
+                for subnet_id in queue.networking.subnet_ids.value
+                if queue.networking.subnet_ids
+            }
+        )
+
+    @property
+    def compute_security_groups(self):
+        """Return the list of all compute security groups in the cluster."""
+        return list(
+            {
+                security_group
+                for queue in self.scheduling.queues
+                if queue.networking.security_groups and queue.networking.security_groups.value
+                for security_group in queue.networking.security_groups.value
+            }
+        )
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -14,9 +14,12 @@
 #
 
 from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_efs as efs
+from aws_cdk import aws_fsx as fsx
 from aws_cdk import core
 
-from pcluster.models.cluster import HeadNode, SharedFsx
+from common.aws.aws_api import AWSApi
+from pcluster.models.cluster import HeadNode, SharedEbs, SharedEfs, SharedFsx, SharedStorage
 from pcluster.models.cluster_slurm import SlurmCluster
 
 
@@ -179,17 +182,6 @@ class HeadNodeConstruct(core.Construct):
         # )
 
 
-class FsxConstruct(core.Construct):
-    """Create the resources related to the FSX."""
-
-    # https://cdkworkshop.com/30-python/40-hit-counter/100-api.html
-    def __init__(self, scope: core.Construct, id: str, fsx: SharedFsx, **kwargs):
-        super().__init__(scope, id)
-        self.fsx = fsx
-        # TODO add all the other required info other than fsx object and generate template
-        # TODO Verify if there are building blocks ready to be used
-
-
 class ClusterStack(core.Stack):
     """Create the Stack and delegate to specific Construct for the creation of all the resources for the Cluster."""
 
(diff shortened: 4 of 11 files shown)


cli/tests/pcluster/templates/test_cdk_builder.py
```diff
@@ -42,7 +43,30 @@ def dummy_cluster():
     queue_networking = QueueNetworking(subnet_ids=["test"])
     queues = [SlurmQueue(name="test", networking=queue_networking, compute_resources=compute_resources)]
     scheduling = SlurmScheduling(queues=queues)
-    return SlurmCluster(image=image, head_node=head_node, scheduling=scheduling)
+    # shared storage
+    shared_storage: List[SharedStorage] = []
+    shared_storage.append(
+        SharedFsx(
```
yuleiwan, line 49: I'd like to use a function to generate the dummy fsx, so we can use different parameters to test template generation in future.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -183,12 +184,36 @@ class FsxConstruct(core.Construct):
...
     # https://cdkworkshop.com/30-python/40-hit-counter/100-api.html
-    def __init__(self, scope: core.Construct, id: str, fsx: SharedFsx, **kwargs):
+    def __init__(self, scope: core.Construct, id: str, shared_fsx: SharedFsx, subnet_id: str, **kwargs):
         super().__init__(scope, id)
-        self.fsx = fsx
+        self.shared_fsx = shared_fsx
         # TODO add all the other required info other than fsx object and generate template
         # TODO Verify if there are building blocks ready to be used
 
+        fsx.CfnFileSystem(
```
yuleiwan, line 193: Here, we should consider security group ids if using existing fsx-id. As far as I know, If using an existing file system, it must be associated to a security group that allows inbound TCP traffic to port 988. Setting the source to 0.0.0.0/0 on a security group rule provides client access from all the IP ranges within your VPC security group for the protocol and port range for that rule. https://docs.aws.amazon.com/parallelcluster/latest/ug/fsx-section.html.
So, should we add a condition to check fsx_id exists or not?

cli/requirements.txt
```diff
@@ -9,6 +9,7 @@ jinja2>=2.11.0
 marshmallow
 aws-cdk.core
 aws-cdk.aws-ec2
+aws-cdk.aws-fsx
```
yuleiwan, line 12: we could add dependency to `cli/test/requirements.txt` as well

yuleiwan (commented):
Looks good for the initial fsx cdk template!

cli/src/pcluster/templates/cluster_stack.py
ddeidda: I would demand this responsibility to validators; CDK conversion should only take care of translating the model. Anyway, security groups and subnet ids were missing, so thanks for pointing to this code.

cli/requirements.txt
ddeidda: Let's see what others think about this :+1:

cli/src/pcluster/models/cluster.py
```diff
@@ -281,8 +281,8 @@ def __init__(
     ):
         super().__init__()
         self.assign_public_ip = Param(assign_public_ip)
-        self.security_groups = security_groups
-        self.additional_security_groups = additional_security_groups
+        self.security_groups = Param(security_groups)
```
yuleiwan, line 284: security_groups and additional_security_groups are List[str] type. Instead of wrapping them to Param, we could wrap the element in list to Param. Sometimes, elements in list is a combination of default values and customer specified values. 
In BaseSchema, `remove_implied_values ` will iterate the list and remove implied values.

But if all the elements are default values or customer specified values, just ignore my comment 😄

cli/src/pcluster/models/cluster.py
```diff
@@ -309,8 +309,8 @@ class QueueNetworking(_BaseNetworking):
 
     def __init__(self, subnet_ids: List[str], placement_group: PlacementGroup = None, **kwargs):
         super().__init__(**kwargs)
-        self.subnet_ids = subnet_ids
-        self.placement_group = placement_group
+        self.subnet_ids = Param(subnet_ids)
```
yuleiwan, line 312: subnet_ids is the same case as security_groups and additional_security_groups. Type of placement_group is PlacementGroup. As I know, we only wrap raw types to Param, for this case. If we want to set default place_group, we could do this
```
default_placement_group = PlacementGroup(id)
default_placement_group.implied = True
self.placement_group = default_placement_group
``` 
If it is not default value, use it directly.

cli/src/pcluster/models/cluster_slurm.py
```diff
@@ -123,3 +123,28 @@ def __init__(
                         os=self.image.os,
                         architecture=self.architecture,
                     )
+
+    @property
+    def compute_subnet_ids(self):
+        """Return the list of all compute subnet ids in the cluster."""
+        return list(
+            {
+                subnet_id
```
yuleiwan, line 132: If we wrap the element in list, we could do `subnet_id.value ...for subnet_id in queue.networking.subnet_ids`

cli/src/pcluster/models/cluster.py
ddeidda: TBH I'm not a big fan of this `Param` class in general. Anyway I think that a configuration property is either entirely computed or not (a "partial" computation does not seem to me to make sense). I mean, even if some items in the list are implied and others not, it's always the entire list that must be computed, for instance to remove duplicates. 
Does it make sense?

cli/src/pcluster/models/cluster.py
ddeidda: Agreed on reverting the change for `PlacementGroup`; however I would avoid using the `implied` property in that way, as "implied" is by definition a property of a `Param` instance. Let's talk about this point if you want.

cli/src/pcluster/models/cluster_slurm.py
ddeidda: See my comment above and of course let's follow up if you think I am missing something.

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Yes. We need more discussion on if we want to use `Param` or not. There are several questions to be answered:
1. Do we want to keep `implied` for each field?
2. Besides using `Param`, is there a better way to keep the `implied`?
3. If `Param` has to be used, where and how should we use it?

enrico-usai: I just want to confirm that the Param represent a parameter in the configuration file, and in this case the param is AdditionalSecurityGroups or SecurityGroups. It's a single Param that contains a list of string and not a list of Param.

Instead we can discuss about the Param usage. We can consider to insert a map in the Resource to store all the information related to the Params instead of attaching them to the param itself.
For now let's go ahead with this Param approach that works as expected altough not the optimal one, then we can review it. Thanks for all the feedbacks!

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -179,30 +180,98 @@ def __init__(self, scope: core.Construct, id: str, head_node: HeadNode, **kwargs
...
+                id="{0}Ids".format(storage_type.name),
+                description="{0} Filesystem IDs".format(storage_type.name),
+                value=",".join(storage_ids),
+            )
+
+    def _add_fsx_storage(self, id: str, shared_fsx: SharedFsx, subnet_ids: [str], security_group_ids: [str]):
+        """Add specific Cfn Resources to map the FSX storage."""
+        fsx_id = shared_fsx.file_system_id.value
+
+        if not fsx_id and shared_fsx.mount_dir:
```
yuleiwan, line 230: `shared_fsx.mount_dir` is Param, so should we check it value?

ddeidda: Right! Fixed now

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -179,30 +182,155 @@ def __init__(self, scope: core.Construct, id: str, head_node: HeadNode, **kwargs
...
+        for subnet_id in compute_subnet_ids:
+            self._add_efs_mount_target(
+                id, efs_id, self._cluster.compute_security_groups, subnet_id, checked_availability_zones
+            )
+
+        # Mount Target for Head Node
+        self._add_efs_mount_target(
+            id,
+            efs_id,
+            self._cluster.head_node.networking.security_groups.value,
```
enrico-usai, line 294: security_groups parameter corresponds to the old vpc_security_group parameter that has a value only if the user is setting it.
By default security_groups is None so you should take it from the one created by us in the main template.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -179,30 +182,155 @@ def __init__(self, scope: core.Construct, id: str, head_node: HeadNode, **kwargs
...
+            ebs_resource = ec2.CfnVolume(
+                scope=self,
+                id=id,
+                availability_zone=self._cluster.head_node.networking.availability_zone,
+                encrypted=shared_ebs.encrypted.value,
+                iops=shared_ebs.iops.value,
+                throughput=shared_ebs.throughput.value,
+                kms_key_id=shared_ebs.kms_key_id.value,
+                size=shared_ebs.size.value,
+                snapshot_id=shared_ebs.snapshot_id.value,
```
enrico-usai, line 331: Does it works correctly if passing a None value?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -179,30 +182,155 @@ def __init__(self, scope: core.Construct, id: str, head_node: HeadNode, **kwargs
...
+            self._cluster.head_node.networking.subnet_id.value,
+            checked_availability_zones,
+        )
+        return efs_id
+
+    def _add_efs_mount_target(
+        self, efs_cfn_resource_id, file_system_id, security_groups, subnet_id, checked_availability_zones
+    ):
+        """Create a EFS Mount Point for the file system, if not already available on the same AZ."""
+        availability_zone = AWSApi.instance().ec2.get_availability_zone_of_subnet(subnet_id)
```
enrico-usai, line 304: We might create a callable class for AWSApi and have:  AWSApi() without the instance method.

cli/src/pcluster/models/cluster.py
```diff
@@ -345,8 +345,10 @@ class QueueNetworking(_BaseNetworking):
 
     def __init__(self, subnet_ids: List[str], placement_group: PlacementGroup = None, **kwargs):
         super().__init__(**kwargs)
-        self.subnet_ids = subnet_ids
+        self.subnet_ids = Param(subnet_ids)
         self.placement_group = placement_group
+        # TODO: retrieve AZs from subnet_id
+        self.availability_zones = None
```
enrico-usai, line 351: to be removed

ddeidda: Done

enrico-usai (approved):
LGTM!

cli/src/pcluster/templates/cluster_stack.py
ddeidda: Fixed by adding a collection `_compute_security_groups` to ClusterStack. In this collection we will add both security groups created in the main substack and the ones explicitly set by users. All storage resources will refer to this list.

cli/src/pcluster/templates/cluster_stack.py
ddeidda: As discussed, None value should be managed from the lib.

cli/src/pcluster/templates/cluster_stack.py
ddeidda: Yes, let's discuss this in a further patch.

PR #2405, opened 2021-01-29
3.0: Migrate EBS validators

enrico-usai:
## Migrate EBS volume type size validator
* Removed from mappings
* Tests were not present in the old code
* Improved unit tests to call the Validator and not the Schema.load().
  In this way the test will be more focused on the validation logic rather than calling other methods.

## Migrate EBS volume IOPS validator

* Removed from mappings
* Tests were not present in the old code
* Improved unit tests to call the Validator and not the Schema.load().
  In this way the test will be more focused on the validation logic rather than calling other methods.

## Migrate EBS volume throughput validator
* Removed from mappings.
* Tests were not present in the old code, now added.
* The old ebs_volume_throughput_validator has been splitted into two classes:
  EbsVolumeThroughputValidator and EbsVolumeThroughputIopsValidator
* Improved unit tests to call the Validator and not the Schema.load().
  In this way the test will be more focused on the validation logic rather than calling other methods.

## Migrate EBS volume size snapshot validator
* Removed from mappings.
* Migrated old tests. Removed the case on which the volume_size was None.


cli/src/pcluster/validators/common.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, raise_on_error=False):
     def _fail(self, message, level):
         raise ConfigValidationError
 
-    def _add_failure(self, message, level, fields=()):
+    def _add_failure(self, message: str, level: FailureLevel, fields=()):
```
hanwen-pcluste, line 56: minor: why not also add type hint for fields? For example: `fields:List[Param]=()`

cli/tests/pcluster/validators/test_ebs.py
```diff
@@ -10,116 +10,194 @@
...
-            {"VolumeType": "gp3", "Throughput": 760, "Iops": 3000},
-            "Throughput to IOPS ratio of .* is too high",
-        ),
-        ({"VolumeType": "gp3", "Throughput": 760, "Iops": 10000}, None),
-        (
-            {"VolumeType": "gp3", "Throughput": 1001, "Iops": 2900},
-            [
-                "IOPS rate must be between 3000 and 16000 when provisioning gp3 volumes.",
-                "Throughput must be between 125 MB/s and 1000 MB/s when provisioning gp3 volumes.",
-            ],
```
hanwen-pcluste, line 40: Are we going to test multiple failure messages? My concern was that now we are trying to report multiple failures at once. By testing validators individually,  can we make sure these wrong inputs do not trigger any exceptions during the execution of the chain of the validators? If not, are we going to test it somewhere else?

hanwen-pcluste (approved):

cli/src/pcluster/validators/common.py
enrico-usai: Fixed, thanks

cli/tests/pcluster/validators/test_ebs.py
enrico-usai: For multiple failures from the same validator we should extend these unit tests.
For multiple failures not related to the same validator we can create more unit tests loading the section as before.
The test for the "chain of validators" should be validated by the unit tests related to the priority feature.

cli/src/pcluster/validators/common.py
```diff
@@ -53,20 +53,21 @@ def __init__(self, raise_on_error=False):
     def _fail(self, message, level):
         raise ConfigValidationError
 
-    def _add_failure(self, message: str, level: FailureLevel, fields=()):
+    def _add_failure(self, message: str, level: FailureLevel, failed_params: List[Param] = None):
```
hanwen-pcluste, line 56: `List[Param]=()` works, doesn't it? 
I am curious why you have to add another line of code for `if failed_params:`

enrico-usai: Fixed

PR #2412, opened 2021-02-01
3.0: Move architecture and hyperthreading related validators

enrico-usai:
# DynamicParam
* Defined new `DynamicParam` class to cope with parameters that are not defined in the config and   must be evaluated at runtime (e.g. `architecture`).
* We might use properties for them but if a validator requires the property it will evaluate it  when calling `_add_validator`. With the `DynamicParam` the value will be evaluated when needed.
* Added tests for `dynamic_property_behaviour`.

# Disable Hyperthreading
* Renamed to `SimultaneousMultithreading` (Hyperthreading is Intel's proprietary)
* Added attribute to Head node and enabled in schema tests
* Moved `disable_hyperthreading_architecture_validator` to `SimultaneousMultithreadingArchitectureValidator`
  and moved tests

# Architecture
* Defined new DynamicParam for architecture attribute.
* Moved `architecture_os_validator` to `ArchitectureOsValidator` and moved tests.
* Moved `efa_os_arch_validator` to `EfaOsArchitectureValidator`. It will be executed only if efa is defined
  in the class.
* Moved `instances_architecture_compatibility_validator` and splitted into two different validators:
  `InstanceArchitectureCompatibilityValidator` and `AwsbatchInstancesArchitectureCompatibilityValidator`,
  splitted the tests too.

# Extra json and cfn_scheduler_slots
* Removed `extra_json` and `cfn_scheduler_slots` validators
* Extra json parameter and the possibility to configure `cfn_scheduler_slots` is deprecated. It was already deprecated for Slurm.

# Other changes
+ Fixed `FsxNetworkingValidator` by adding `Param` type hint.
+ Split image builder and cluster stack generation tests and move dummy model classes into new tests modules.
+ Reset validators failures when calling validate function
+ Changed mock in old tests because some methods are no longer included in the validators module.
+ Other minor cleanup changes.

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -59,6 +59,7 @@ class Resource(ABC):
     def validate(self, raise_on_error=False):
         """Execute registered validators, ordered by priority (high prio --> executed first)."""
         # order validators by priority
+        self._validation_failures.clear()
         self.__validators = sorted(self.__validators, key=operator.attrgetter("priority"), reverse=True)
 
         # execute validators and add results in validation_failures array
@@ -415,7 +416,7 @@ class HeadNode(Resource):
         self.storage = storage
         self.dcv = dcv
         self.efa = efa
-        self._add_validator(InstanceTypeValidator, priority=1, instance_type=self.instance_type)
+        self._add_validator(InstanceTypeValidator, instance_type=self.instance_type)
 
 
 class BaseComputeResource(Resource):
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -19,8 +19,14 @@ from enum import Enum
 from typing import List
 
 from pcluster.constants import CIDR_ALL_IPS, EBS_VOLUME_TYPE_IOPS_DEFAULT
-from pcluster.models.param import Param
-from pcluster.validators.cluster_validators import FsxNetworkingValidator
+from pcluster.models.param import DynamicParam, Param
+from pcluster.utils import error, get_supported_architectures_for_instance_type
+from pcluster.validators.cluster_validators import (
+    ArchitectureOsValidator,
+    EfaOsArchitectureValidator,
+    FsxNetworkingValidator,
+    SimultaneousMultithreadingArchitectureValidator,
+)
 from pcluster.validators.common import ValidationResult, Validator
 from pcluster.validators.ebs_validators import (
     EbsVolumeIopsValidator,
@@ -403,6 +409,7 @@ class HeadNode(Resource):
         instance_type: str,
         networking: HeadNodeNetworking,
         ssh: Ssh,
+        simultaneous_multithreading: bool = None,
         image: Image = None,
         storage: Storage = None,
         dcv: Dcv = None,
@@ -410,6 +417,7 @@ class HeadNode(Resource):
     ):
         super().__init__()
         self.instance_type = Param(instance_type)
+        self.simultaneous_multithreading = Param(simultaneous_multithreading, default=True)
         self.image = image
         self.networking = networking
         self.ssh = ssh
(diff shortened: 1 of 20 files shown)


cli/src/pcluster/models/cluster_slurm.py
```diff
@@ -104,3 +107,19 @@ def __init__(
...
+                    architecture=self.architecture,
+                )
+                if compute_resource.efa:
+                    self._add_validator(
+                        EfaOsArchitectureValidator,
+                        priority=9,
+                        efa_enabled=compute_resource.efa.enabled,
+                        os=self.image.os,
+                        architecture=self.architecture,
+                    )
```
hanwen-pcluste, line 125: Do we really want to replicate the code for Awsbatchcluster?

cli/src/pcluster/models/cluster_slurm.py
```diff
@@ -104,3 +107,19 @@ def __init__(
...
         self.scheduling = scheduling
+
+        for queue in self.scheduling.queues:
+            for compute_resource in queue.compute_resources:
+                self._add_validator(
+                    InstanceArchitectureCompatibilityValidator,
+                    instance_type=compute_resource.instance_type,
+                    architecture=self.architecture,
+                )
+                if compute_resource.efa:
```
hanwen-pcluste, line 118: When seeing this conditional statement, it feels like the validator should be inside compute_resource.efa. And here comes to the discussion about if we should allow access parent information such as os and architecture inside efa section.
Anyway, I am fine with your current code.

hanwen-pcluste (approved):
Awesome work. Thank you!

hanwen-pcluste:
I didn't read CDK unit tests and their dummy models.

cli/src/pcluster/models/cluster.py
```diff
@@ -59,6 +65,7 @@ def __init__(self):
     def validate(self, raise_on_error=False):
         """Execute registered validators, ordered by priority (high prio --> executed first)."""
         # order validators by priority
+        self._validation_failures.clear()
```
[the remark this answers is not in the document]
enrico-usai: To be sure that every time we call the `validate()` function we don't have old `validation_failures` in the list.
See as an example of the expected behaviour the `test_dynamic_property_validate`

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -127,3 +137,105 @@ def _check_sg_rules_for_port(self, rule, port_to_check):
...
+        return self._failures
+
+
+class InstanceArchitectureCompatibilityValidator(Validator):
+    """Validate instance type and architecture combination."""
+
+    def validate(self, instance_type: Param, architecture: DynamicParam):
+        """Verify that head node and compute instance types imply compatible architectures."""
+        head_node_architecture = architecture.value
+        compute_architectures = get_supported_architectures_for_instance_type(instance_type.value)
```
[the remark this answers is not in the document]
enrico-usai: It's a good point. The difference I can see is that the head node architecture is a relevant information for a lot of validators, the architecture of the compute nodes is only required here. I think for now we can avoid the DynamicParam but we might add it if needed in future.

cli/src/pcluster/models/cluster_slurm.py
enrico-usai: I agree, for now let's go ahead with this approach, that works as expected even if we're not sure it's the optimal one.

cli/src/pcluster/models/cluster_slurm.py
enrico-usai: Another good point. I wasn't sure either. 
The decision here is whether to define a generic interface for the ComputeResource and Scheduling classes.
For now let's go ahead and we can go back to this decision.

cli/src/pcluster/models/param.py
```diff
@@ -27,3 +28,16 @@ def __init__(self, value, default=None, valid: bool = True):
             self.value = value
             self.implied = False
         self.valid = valid
+
+
+class DynamicParam:
+    """Class to manage dynamic params that must be calculated at usage time."""
+
+    def __init__(self, value_calculator):
+        """Create MarkedValue object and add attributes."""
```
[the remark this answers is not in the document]
enrico-usai: Done

PR #2414, opened 2021-02-02
3.0: Move common model classes into a common module and rename validators modules

enrico-usai:
## Move all the common model classes in the same module

This refactoring permits to have in a single place the classes to be used by both `Cluster` and `ImageBuilder` modules (e.g. `Param`, `Resources`, `Tags`)

The models/common.py now includes:
* `Param` and `DynamicParam` moved from param.py
* `Resource` and related classes from cluster.py
* `Tag` class (in common between `Cluster` and `ImageBuilder`)
* Generic `Validator` class and validation related classes

## Rename test validators modules
In this way it's easier to understand what we're testing from the module name.

## Improve Validator classes 

* Use `execute` method as entry point in place of callable classes, this change should clarify that the method to be used to execute the validator is the `execute` one, the only one public.
* The `validate` method is now private and in this way nobody can wrongly call it from outside.
* Removed return `self._failures` from all the validators. It's done by the execute method.


cli/src/pcluster/models/common.py
```diff
@@ -0,0 +1,180 @@
...
+class FailureLevel(Enum):
+    """Validation failure level."""
+
+    CRITICAL = 50
+    ERROR = 40
+    WARNING = 30
+    INFO = 20
+
+
+# TODO disable specific validator, according with the name
```
ddeidda, line 67: What is this comment referred to?

ddeidda (approved):

cli/src/pcluster/models/common.py
enrico-usai: Removed

PR #2415, opened 2021-02-02
3.0: Migrate compute resource validator and other refactorings

enrico-usai:
## Move compute resource size validators

* initial_queue_size has been deprecated
* min_size and max_size have been renamed to MinCount and MaxCount
* We had a single validator for both awsbatch and slurm, now we have two different validators associated to the different ComputeResources
* migrated and updated related tests

## Other changes
* Add missing type hints for Fsx validators 
* Remove os validator. It was used to alert the users when using alinux but it will be deprecated within the next version.
* Move AWS Batch validators and tests in separate modules
* Remove useless CRITICAL FailureLevel
* Add missing information about the failed parameters in validators
* Align validators code by using .value everywhere

### DIFF
--- a/cli/src/pcluster/validators/awsbatch_validators.py
+++ b/cli/src/pcluster/validators/awsbatch_validators.py
@@ -1,0 +1,45 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pcluster.models.common import DynamicParam, FailureLevel, Param, Validator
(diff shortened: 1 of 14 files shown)


cli/src/pcluster/validators/awsbatch_validators.py
```diff
@@ -0,0 +1,65 @@
...
+class AwsbatchComputeResourceValidator(Validator):
+    """Awsbatch compute resource validator."""
+
+    def _validate(self, min_vcpus: Param, desired_vcpus: Param, max_vcpus: Param):
+        """Validate min, desired and max vcpus combination."""
+
+        if desired_vcpus.value < min_vcpus.value:
+            self._add_failure(
+                "The number of desired vcpus must be greater than or equal to min vcpus", FailureLevel.ERROR
+            )
```
hanwen-pcluste, line 25: Minor: You are not marking the any param as invalid. If you want, you can write
```
self._add_failure(
                "The number of desired vcpus must be greater than or equal to min vcpus", FailureLevel.ERROR, [desired_vcpus, min_vcpus]
            )
```

hanwen-pcluste: But if the validators are not as intertwined as EBS or FSx validator, keeping the current code is fine.

hanwen-pcluste (approved):

cli/src/pcluster/validators/awsbatch_validators.py
enrico-usai: Fixed

PR #2417, opened 2021-02-03
3.0: Move compute resource and instance type validators

enrico-usai:
## slurm compute resource validator
* The checks for min values have been moved into Schema Ranges validators.
* initial_count has been deprecated so related checks have been removed.
* The combination between min/max/desired is already checked in the (Awsbatch)ComputeResourceValidator
* Defined DUMMY_COMPUTE_RESOURCE for the parameters required by the schema.

## instance type validators
* `ec2_instance_type_validator` was already moved and already used for slurm model
* `compute_instance_type_validator` moved to `AwsbatchComputeInstanceTypeValidator`. Now the code is AWS Batch specific.
* Removed from mappings
* Moved tests

I deleted the `get_supported_compute_instance_types` utility method.
In the new `AwsbatchComputeInstanceTypeValidator` we can use `get_supported_batch_instance_types` and in easyconfig.py we're in the no-awsbatch case so we can use `get_supported_instance_types`. Removed related tests.

## Other changes:
* Rename `*ComputeResourceValidator` to `*ComputeResourceSizeValidator`
* Add instance type validator for Slurm compute instances
* Move validators doc from private method to class definition
* Move architecture attribute in the head node
* Remove draft code

### DIFF
--- a/cli/src/pcluster/config/mappings.py
+++ b/cli/src/pcluster/config/mappings.py
@@ -56,7 +56,6 @@ from pcluster.config.param_types import Visibility
 from pcluster.config.update_policy import UpdatePolicy
 from pcluster.config.validators import (
     compute_instance_type_validator,
-    compute_resource_validator,
     dcv_enabled_validator,
     duplicate_shared_dir_validator,
     ebs_settings_validator,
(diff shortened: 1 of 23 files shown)


enrico-usai:
Closed in favour of: #2418

PR #2419, opened 2021-02-03
3.0: Delete migrated allowed values from mappings.py

hanwen-pcluste:
1. After deleting the allowed values, relevant old unit tests were failing. This commit will adapt all the failing tests to the test_schema_validators.py.
2. Change `_load_and_assert_error()` to `_validate_and_assert_error()` to avoid the loading the data into model layer when unit testing. However, `validate()` in Marshmallow returns a dict of error messages instead of raising ValidationError, the code of the utility function becomes more complicated (It is Ok).
3. Change the minimal provisioned throughput from 0 to 1 and disallow 0 bid percentage.
4. Skip validating existencies when partial loading
5. This commit won't remove validations that doen't apply to pcluster 3.0, e.g. compute_subnet_cidr, vpc_id. I'd remove them at the last stage of the refactoring, so we will be able to double check everything necessary is moved.
6. There are still unmoved unit tests. I only focused on moving the failing unit tests because of the deletion of the allowed values


enrico-usai (approved):
Great job!

PR #2420, opened 2021-02-04
3.0: Move DCV validator and other storage validators

enrico-usai:
## Migration:
* Move DCV validator
* Move duplicate mount dir validator
* Migrate number of storage validator
    * Created a generic `NumberOfStorageValidator` to be used for all the storage types. Migrated and adapted tests

## Other changes:
* Remove image from `HeadNode` schema. The `Image` section is only supported in the root of the config file not in the `HeadNode` section.
* Minor moving of storage class and validators definition


cli/src/pcluster/models/cluster.py
```diff
@@ -45,6 +46,8 @@
 
 # ---------------------- Storage ---------------------- #
 
+MAX_STORAGE_COUNT = {"ebs": 5, "efs": 1, "fsx": 1}
```
hanwen-pcluste, line 49: Minor: how about moving this into `NumOfStorageValidator`

hanwen-pcluste (approved):
LGTM, thank you!

cli/src/pcluster/models/cluster.py
enrico-usai: Let's think about this, thanks. I'll go ahead for now.

PR #2421, opened 2021-02-04
3.0: Register validators right before resource validation

enrico-usai:
The previous approach of registering the validators in the initialization of the Resource wasn't considering the possibility to change the model attributes.
The registered validators were validating only the attributes present at init time.

With the new approach the validator are registered just before calling the validation so we'll have the validators associated to the current status of the model.
For this reason we don't need anymore the Dynamic param because we don't need to pass a reference to the validator but we can simply pass the property with the value in that specific moment.

By removing the DynamicParam concept we're simplifying the tests and the usage of the model.

### DIFF
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -16,7 +16,7 @@ from botocore.exceptions import ClientError
 
 from pcluster.constants import CIDR_ALL_IPS
 from pcluster.dcv.utils import get_supported_dcv_os
-from pcluster.models.common import DynamicParam, FailureLevel, Param, Validator
+from pcluster.models.common import FailureLevel, Param, Validator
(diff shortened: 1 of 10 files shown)


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -76,12 +76,12 @@ class ArchitectureOsValidator(Validator):
     ARM AMIs are only available for a subset of the supported OSes.
     """
 
-    def _validate(self, os: Param, architecture: DynamicParam):
-        allowed_oses = get_supported_os_for_architecture(architecture.value)
+    def _validate(self, os: Param, architecture):
```
hanwen-pcluste, line 79: Minor: missing type hint

hanwen-pcluste (approved):

cli/src/pcluster/validators/cluster_validators.py
enrico-usai: added thanks

PR #2422, opened 2021-02-04
3.0 ImageBuilder add and amend validators and unit tests

yuleiwan:
* Add a new validator AMIVolumeSizeValidator
* Amend validators and corresponding unit tests, combine BaseAMIValidator, IntsanceTypeValidator and InstanceTypeBaseAMICompatibleValidator together.
* Add imagebuilder model unit tests

### DIFF
--- a/cli/src/common/boto3/ec2.py
+++ b/cli/src/common/boto3/ec2.py
@@ -11,7 +11,7 @@
 
 # TODO move s3_factory.py and awsbatch/utils.py here
 
-from common.boto3.common import AWSExceptionHandler, Boto3Client
+from common.boto3.common import AWSClientError, AWSExceptionHandler, Boto3Client
 
 
 class Ec2Client(Boto3Client):
@@ -29,15 +29,12 @@ class Ec2Client(Boto3Client):
         ]
 
     @AWSExceptionHandler.handle_client_exception
-    def describe_ami_id_offering(self, ami_id):
-        """Return a boolean to determine ami id is valid or not."""
-        return self._paginate_results(
-            self._client.describe_images(
-                ImageIds=[
-                    ami_id,
-                ]
-            )
-        )
+    def describe_image(self, ami_id):
+        """Return a dict of ami info."""
+        result = self._client.describe_images(ImageIds=[ami_id])
+        if result.get("Images"):
+            return result.get("Images")[0]
+        raise AWSClientError(function_name="describe_image", message=f"Image {ami_id} not found")
 
     @AWSExceptionHandler.handle_client_exception
     def describe_key_pair(self, key_name):
--- a/cli/src/pcluster/models/imagebuilder.py
+++ b/cli/src/pcluster/models/imagebuilder.py
@@ -15,15 +15,16 @@
 
 from typing import List
 
+from common import imagebuilder_utils
+from common.boto3.common import AWSClientError
+from common.boto3.ec2 import Ec2Client
 from pcluster import utils
 from pcluster.models.common import BaseDevSettings, BaseTag, Resource
 from pcluster.validators.ebs_validators import EBSVolumeKmsKeyIdValidator, EbsVolumeTypeSizeValidator
-from pcluster.validators.ec2_validators import (
-    BaseAMIValidator,
-    InstanceTypeBaseAMICompatibleValidator,
-    InstanceTypeValidator,
-)
+from pcluster.validators.ec2_validators import InstanceTypeBaseAMICompatibleValidator
+from pcluster.validators.imagebuilder_validators import AMIVolumeSizeValidator
 
+PCLUSTER_RESERVED_VOLUME_SIZE = 15
 # ---------------------- Image ---------------------- #
 
 
@@ -75,7 +76,7 @@ class Image(Resource):
 class Component(Resource):
     """Represent the components configuration for the ImageBuilder."""
 
-    def __init__(self, type: str = None, value: str = None):
+    def __init__(self, type: str, value: str):
         super().__init__()
         self.type = Resource.init_param(type)
         self.value = Resource.init_param(value)
@@ -105,13 +106,10 @@ class Build(Resource):
         self.components = components
 
     def _register_validators(self):
-        self._add_validator(BaseAMIValidator, priority=15, parent_image=self.parent_image)
-        self._add_validator(InstanceTypeValidator, priority=15, instance_type=self.instance_type)
         self._add_validator(
             InstanceTypeBaseAMICompatibleValidator,
-            priority=14,
             instance_type=self.instance_type,
-            parent_image=self.parent_image,
+            image=self.parent_image,
         )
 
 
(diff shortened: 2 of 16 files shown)


cli/src/common/boto3/ec2.py
```diff
@@ -30,11 +30,9 @@ def describe_instance_type_offerings(self):
...
     @AWSExceptionHandler.handle_client_exception
     def describe_ami_id_offering(self, ami_id):
-        """Return a boolean to determine ami id is valid or not."""
-        return self._paginate_results(
-            self._client.describe_images(
-                ImageIds=[
-                    ami_id,
-                ]
-            )
+        """Return a dict of ami info."""
```
enrico-usai, line 33: Thanks for fixing this method.

1. let's rename the method to describe_image
2. let's return the dictionary or raise an AWSClientError(..., "image xxx not found") if the ami is not found, then the caller will manage the exception.

cli/src/common/boto3/ec2.py
```diff
@@ -30,11 +30,9 @@ def describe_instance_type_offerings(self):
...
+        return self._client.describe_images(
```
enrico-usai, line 34: let's return the Image dict without the entire Images list, in this way all the callers dont' need to manage the KeyError or navigate the Images.

You can do something like:
```python
result = self._client.describe_images(...)
if result.get("Images"):
    return  result.get("Images")[0]
else:
    raise AWSClientError(....)
```

cli/src/common/boto3/ec2.py
```diff
@@ -30,11 +30,9 @@ def describe_instance_type_offerings(self):
...
+            ImageIds=[
```
enrico-usai, line 35: minor: you can compact everything in a single line.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -21,9 +22,11 @@ class BaseAMIValidator(Validator):
     """
 
     def _validate(self, image: Param):
-        ami_id = utils.get_ami_id(image.value)
-        if not Ec2Client().describe_ami_id_offering(ami_id=ami_id):
-            self._add_failure(f"The ami id '{ami_id}' is not supported.", FailureLevel.ERROR, [image])
+        try:
+            ami_id = utils.get_ami_id(image.value)
+            Ec2Client().describe_ami_id_offering(ami_id=ami_id)
+        except AWSClientError:
```
enrico-usai, line 28: It's the right approach, thanks for fixing this.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -173,15 +178,21 @@ def __init__(
...
+            volume_size=self.image.root_volume.size,
+            image=self.build.parent_image,
+        )
 
     def _set_default(self):
         # set default root volume
         if self.image.root_volume is None or self.image.root_volume.size.value is None:
-            increase_volume_size = 15
+            reserved_size = 15
             ami_id = utils.get_ami_id(self.build.parent_image.value)
```
enrico-usai, line 192: not strictly related to this patch but by looking at this get_ami_id method I have some comments:
1. It's  ImageBuilder specific, in fact it's using ImageBuilder boto3 client. I'll keep it in the imagebuilder.py model or in a imagebuilder_utils.py to avoid that someone start using this utility method for the base cli (we don't want it because permission for calling ImageBuilder are only required for the createami command).
2. Let's rename to evaluate_ami_id(ami) to be clear that the method is evaluating the id from the given param.
3. Let's rename the parameter to ami, rather than parent_image that is too strictly related to the image builder stuff.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -47,13 +50,13 @@ class InstanceTypeBaseAMICompatibleValidator(Validator):
 
     def _validate(self, instance_type: Param, parent_image: Param):
```
enrico-usai, line 51: It seems to me that we're mixing the validators for imagebuilder and cluster.

If they are very imagebuilder specific let's create a imagebuilder_validator module (e.g. if they are using ImageBuilder boto3 calls inside).

yuleiwan: I am thinking we can share this validator with Cluster and Imagebuilder. I just need to rename parent_image to image, and `get_ami_id` won't call ImageBuilder boto3 calls if `image` is an ami_id instead of image_arn.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: I'd like to rename the function name to `evaluate_ami_id(ami)` and rename parameter to ami to make it more generic, and I still keep this in utils, because if we need to evaluate_ami_id by other info (e.g. ami name) in future, we can extend this function.

cli/src/common/boto3/ec2.py
yuleiwan: done

cli/src/common/boto3/ec2.py
yuleiwan: done

cli/src/common/boto3/ec2.py
yuleiwan: done, thanks for clear example

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -179,15 +185,25 @@ def _register_validators(self):
...
+        if self.dev_settings:
+            self.dev_settings.validate()
 
     def _set_default(self):
         # set default root volume
         if self.image.root_volume is None or self.image.root_volume.size.value is None:
-            increase_volume_size = 15
-            ami_id = utils.get_ami_id(self.build.parent_image.value)
-            ami_info = utils.get_info_for_amis([ami_id])
+            reserved_size = 15
```
lukeseawalker, line 202: let's put this in a constant in the `imagebuilder.py` for now, so it can be used by the validator

cli/src/pcluster/utils.py
```diff
@@ -920,12 +920,12 @@ def get_info_for_amis(ami_ids):
         error(e.response.get("Error").get("Message"))
 
 
-def get_ami_id(parent_image):
-    """Get ami id from parent image, parent image could be image id or image arn."""
-    if parent_image.startswith("arn"):
-        ami_id = ImageBuilderClient().get_image_id(parent_image)
+def evaluate_ami_id(ami):
```
lukeseawalker, line 923: -minor: I like more the previous naming `get_ami_id(parent_image)`

cli/src/pcluster/validators/imagebuilder_validators.py
```diff
@@ -0,0 +1,43 @@
...
+    """AMI root volume size validator."""
+
+    def _validate(self, volume_size: Param, image: Param):
+        """Validate the volume size is larger than base ami volume size."""
+        ami_id = utils.evaluate_ami_id(image.value)
+        ami_info = Ec2Client().describe_image(ami_id)
+        reserved_size = 15
+        base_ami_volume_size = ami_info.get("BlockDeviceMappings")[0].get("Ebs").get("VolumeSize")
+        min_volume_size = base_ami_volume_size + reserved_size
+        if volume_size.value < base_ami_volume_size:
```
lukeseawalker, line 27: It's possible to have one failure with `volume_size.value < min_volume_size`
"Root volume size (size) is lesser than the minimum required size of base ami (size) plus size (15) to allow PCluster software stack installation"?

In this way we are clear from the first error on what value we are expecting.

cli/src/pcluster/utils.py
enrico-usai: This change has been done after my comment :D 

The _get_ word isn't clear when you're already passing an ami id to the method. I mean, getting an ami id from an ami id?

The _get_ word let me think we need to obtain the value from somewhere, _evaluate_ (or _determine_, or another word) let me think we're evaluating it from the given value.

lukeseawalker: we are getting the ami id from the `parent_image` string, that could be either an `image arn` or an `ami id` already. In case it's an `image arn` we are calling an api to retrieve the `ami id`.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Done

cli/src/pcluster/utils.py
yuleiwan: I create a `imagebuilder_utils` module in `common` dir to put imagebuilder related utility functions, and move `get_ami_id` to that module

cli/src/pcluster/validators/imagebuilder_validators.py
yuleiwan: Done

cli/tests/pcluster/models/test_imagebuilder_model.py
```diff
@@ -0,0 +1,214 @@
...
+            None,
+            ["x86_64"],
+            [
+                "Kms Key Id key_id is specified, the encrypted state must be True.",
+                "Root volume size 25 GB is less than the minimum required size 65 GB that equals "
+                "base ami 50 GB plus size 15 GB to allow PCluster software stack installation.",
+            ],
+            [FailureLevel.ERROR, FailureLevel.ERROR],
+        ),
+        # test InstanceTypeBaseAMICompatibleValidator
```
enrico-usai, line 69: Let's split the tests into small ones.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -23,8 +26,10 @@
     InstanceTypeBaseAMICompatibleValidator,
     InstanceTypeValidator,
 )
+from pcluster.validators.imagebuilder_validators import AMIVolumeSizeValidator
 from pcluster.validators.s3_validators import UrlValidator
 
+PCLUSTER_RESERVED_SIZE = 15
```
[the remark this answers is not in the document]
yuleiwan: changed

cli/tests/pcluster/models/imagebuilder_dummy_model.py
```diff
@@ -8,10 +8,22 @@
...
+from pcluster.models.imagebuilder import ImageBuilder, Volume
+
+CLASS_DICT = {
+    "imagebuilder": ImageBuilder,
+    "image": Image,
+    "build": Build,
+    "dev_settings": DevSettings,
+    "root_volume": Volume,
+    "tag": BaseTag,
+    "compnent": Component,
```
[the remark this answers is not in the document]
yuleiwan: fixed

cli/tests/pcluster/models/test_imagebuilder_model.py
yuleiwan: re-organize the unit tests

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -122,7 +127,7 @@ def _register_validators(self):
 class ChefCookbook(Resource):
     """Represent the chef cookbook configuration for the ImageBuilder."""
 
-    def __init__(self, url: str, json: str):
+    def __init__(self, url: str = None, json: str = None):
```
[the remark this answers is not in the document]
yuleiwan: yes, only json is optional, fixed

hanwen-pcluste (approved):

PR #2424, opened 2021-02-04
3.0: Migrate additional IAM policies and key pair validators

enrico-usai:
* Migrate additional Iam policies validator
  * With the new model we can validate a single policy rather than a list.
  * There were no unit tests for it, still to do.
*  Migrate key pair validator
   * There was a unit test but it was useless, just testing the working case, still to do.
* Remove already migrated validator from validator module

### DIFF
--- a/cli/src/common/boto3/ec2.py
+++ b/cli/src/common/boto3/ec2.py
@@ -38,3 +38,8 @@ class Ec2Client(Boto3Client):
                 ]
             )
         )
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_key_pair(self, key_name):
+        """Return the given key, if exists."""
+        return self._client.describe_key_pairs(KeyNames=[key_name])
--- a/cli/src/common/boto3/iam.py
+++ b/cli/src/common/boto3/iam.py
@@ -1,0 +1,24 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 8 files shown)


cli/src/common/boto3/ec2.py
```diff
@@ -38,3 +38,8 @@ def describe_ami_id_offering(self, ami_id):
                 ]
             )
         )
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_key_pair(self, key_name):
+        """Return the given key, if exists."""
+        return self._client.describe_key_pairs(KeyNames=[key_name])
```
hanwen-pcluste, line 45: I would suggest to not bother adding anything to boto3 layer before we reach a consensus on what is the responsibility of this layer.

hanwen-pcluste (approved):
LGTM

cli/src/common/boto3/ec2.py
enrico-usai: The boto3 layer is still in draft state, but the responsibilities are:
* **Decouple** boto3 specific code from our model
  * It means that if in future boto3 deprecate `ClientError` or change is format or change the client initialization we don't need to change our code everywhere but we only need to change the classes in the boto3 layer
  * If in future there is boto4 we only need to change that layer
  * If in future we want to drop boto3 and use aws cli to have the same information we only need to change that layer
* **Abstract** the api calls
  * Instead of using the boto3 specific parameters everywhere (`KeyNames`, `ImageIds`, etc) we can have more abstract methods, accepting `key_name`, `image_ids` or more generic parameters more related to our model
  * The methods in that layer might return more abstract objects (e.g. `ImageInfo`) rather than a boto3 specific dictionary.
  * The methods can navigate the boto3 answer to retrieve the requested info (e.g. `get_ami_info` can do `result.get("Images")[0]`
* **Manage common functionalities**
  * Catch exceptions (e.g. `ClientError`) by returning an exception specific for our model (e.g. `AWSClientError`)
  * Have a single place on which add common functionalities (e.g. pagination and exception handler) and easily attach them to the different methods
  * If you need to add a proxy or config params to all the boto3 calls you only need to change the common `Boto3Client` class rather than changing the code everywhere. 
* **Simplify** unit tests and mocking
  * We have classes and methods that can be easily mocked rather than having to mock boto3 calls with expected params, expected results, etc
  * Old validators unit tests were like integration tests because they were testing too many things, with this new layer we can focus the test on the validator logic, then if we want to test the boto3 layer we can add unit tests specific for it.
* ...

btw it's still in draft state.

PR #2430, opened 2021-02-08
3.0: Migrate EFA, Queue and PlacementGroup validators 

enrico-usai:
## Migrate EFA and Queue validators 
* efa_validator splitted into EfaValidator, EfaPlacementGroupValidator and EfaSecurityGroupValidator
* efa_gdr_validator migrated into EfaValidator
* queue_validator migrated into EfaValidator and DuplicateInstanceTypeValidator
* Remove efa attribute from Batch model (not supported)
* Create utility method to find duplicate Param in a list and   use it for both mount_dirs and instance types

## Create property for mount dir list 
By using a property we can retrieve the list of mount dirs at any time and not only when registering the validators.

## Migrate Placement Group id validator 
Existing test was useless, still todo


hanwen-pcluste (approved):
Great! Can be merged after fixing code check errors

PR #2433, opened 2021-02-10
3.0: Add support for DevSettings for cluster and uniform namings with imagebuilder

hanwen-pcluste:
To import validators inside `models/common.oy`, this commit also separate the module into `models/common.py`, `models/param.py`, `validators/common.py` to avoid circular import.

Fix an infinite reference (`restore_child` in cluster_schema.py) crashing Python when debug. The code ran finely without debug. However, when debugging, the original code would fail.

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -16,7 +16,7 @@ from enum import Enum
 from typing import List
 
 from pcluster.constants import CIDR_ALL_IPS, EBS_VOLUME_TYPE_IOPS_DEFAULT
-from pcluster.models.common import BaseTag, Resource
+from pcluster.models.common import BaseDevSettings, BaseTag, Resource
 from pcluster.utils import (
     error,
     get_availability_zone_of_subnet,
@@ -651,6 +651,18 @@ class AdditionalPackages(Resource):
         self.intel_select_solutions = intel_select_solutions
 
 
+class ClusterDevSettings(BaseDevSettings):
+    """Represent the dev settings configuration."""
+
+    def __init__(self, cluster_template: str = None, **kwargs):
+        super().__init__(**kwargs)
+        self.cluster_template = Resource.init_param(cluster_template)
+
+    def _register_validators(self):
+        super()._register_validators()
+        self._add_validator(UrlValidator, url=self.cluster_template)
+
+
 # ---------------------- Root resource ---------------------- #
 
 
@@ -669,6 +681,7 @@ class BaseCluster(Resource):
         custom_actions: CustomAction = None,
         cluster_s3_bucket: str = None,
         additional_resources: str = None,
+        dev_settings: ClusterDevSettings = None,
     ):
         super().__init__()
         self.image = image
@@ -681,6 +694,7 @@ class BaseCluster(Resource):
         self.custom_actions = custom_actions
         self.cluster_s3_bucket = cluster_s3_bucket
         self.additional_resources = additional_resources
+        self.dev_settings = dev_settings
 
     def _register_validators(self):
         self._add_validator(
--- a/cli/src/pcluster/models/common.py
+++ b/cli/src/pcluster/models/common.py
@@ -14,65 +14,11 @@
 #
 
 import operator
-from abc import ABC, abstractmethod
-from enum import Enum
+from abc import ABC
 from typing import List
 
-# ----------------- Validators ----------------- #
-
-
-class FailureLevel(Enum):
-    """Validation failure level."""
-
-    ERROR = 40
-    WARNING = 30
-    INFO = 20
-
-
-class ValidationResult:
-    """Represent the result of the validation."""
-
-    def __init__(self, message: str, level: FailureLevel):
-        self.message = message
-        self.level = level
-
-
-class ConfigValidationError(Exception):
-    """Configuration file validation error."""
-
-    def __init__(self, validation_result: ValidationResult):
-        message = f"{validation_result.level.name}: {validation_result.message}"
-        super().__init__(message)
-
-
-class Validator(ABC):
-    """Abstract validator. The children must implement the validate method."""
-
-    def __init__(self, raise_on_error=False):
-        self._failures = []
-        self._raise_on_error = raise_on_error
-
-    def _fail(self, message, level):
-        raise ConfigValidationError
-
-    def _add_failure(self, message: str, level: FailureLevel):
-        result = ValidationResult(message, level)
-        if self._raise_on_error:
-            raise ConfigValidationError(result)
-        self._failures.append(result)
-
-    def execute(self, *arg, **kwargs):
-        """Entry point of all validators to verify all input params are valid."""
-        self._validate(*arg, **kwargs)
-        return self._failures
-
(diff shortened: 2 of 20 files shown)


cli/src/pcluster/models/cluster.py
```diff
@@ -652,6 +652,14 @@ def __init__(
         self.intel_select_solutions = intel_select_solutions
 
 
+class DevSettings(BaseDevSettings):
+    """Represent the dev settings configuration for the ImageBuilder."""
```
enrico-usai, line 656: wrong comment

cli/src/pcluster/models/common.py
```diff
@@ -183,3 +105,35 @@ def __init__(
         super().__init__()
         self.key = Param(key)
         self.value = Param(value)
+
+
+class Cookbook(Resource):
+    """Represent the chef cookbook configuration for the ImageBuilder."""
```
enrico-usai, line 111: ImageBuilder --> wrong

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -119,45 +118,23 @@ def _register_validators(self):
...
     def __init__(
         self,
         update_os_and_reboot: bool = None,
         disable_pcluster_component: bool = None,
-        chef_cookbook: ChefCookbook = None,
-        node_url: str = None,
-        aws_batch_cli_url: str = None,
         distribution_configuration_arn: str = None,
         terminate_instance_on_failure: bool = None,
+        **kwargs
```
enrico-usai, line 130: let's call them explicitly. in this way it's clear which parameters can be passed to the constructor of this class.

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -111,3 +112,29 @@ class TagSchema(BaseSchema):
     def make_resource(self, data, **kwargs):
         """Generate resource."""
         return BaseTag(**data)
+
+
+class CookbookSchema(BaseSchema):
+    """Represent the schema of the ImageBuilder chef cookbook."""
```
enrico-usai, line 118: ImageBuilder  --> wrong

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -111,3 +112,29 @@ class TagSchema(BaseSchema):
...
+        if value and not validate_json_format(value):
+            raise ValidationError(message="'{0}' is invalid".format(value))
+
+
+class BaseDevSettingsSchema(BaseSchema):
+    """Represent the common schema of Dev Setting for ImageBuilder and Cluster."""
+
+    cookbook = fields.Nested(CookbookSchema)
+    node_package = fields.Str()
+    aws_batch_cli_package = fields.Str(data_key="AWSBatchCLIPackage")
```
enrico-usai, line 140: We should align the convention, so use AwsBatchCliPackage

cli/src/pcluster/schemas/imagebuilder_schema.py
```diff
@@ -131,34 +136,11 @@ def validate_security_group_ids(self, value):
...
-    def validate_json(self, value):
-        """Validate json."""
-        if value and not validate_json_format(value):
-            raise ValidationError(
-                message="The Json in ChefCookbook '{0}' is invalid, check the json format.".format(value)
-            )
-
-
-class DevSettingsSchema(BaseSchema):
+class DevSettingsSchema(BaseDevSettingsSchema):
```
enrico-usai, line 139: let's avoid the same class name, let's call them ImagebuilderDevSettingsSchema and ClusterDevSettingsSchema

cli/src/pcluster/models/common.py
```diff
@@ -183,3 +105,35 @@ def __init__(
...
+
+    def __init__(self, chef_cookbook: str, extra_chef_attributes: str = None):
+        super().__init__()
+        self.chef_cookbook = Param(chef_cookbook)
+        self.extra_chef_attributes = Param(extra_chef_attributes)
+        # TODO: add validator
+
+    def _register_validators(self):
+        self._add_validator(UrlValidator, url=self.url)
```
yuleiwan, line 120: self.url -> self.chef_cookbook

cli/src/pcluster/models/param.py
```diff
@@ -0,0 +1,32 @@
...
+#
+# This module contains all the classes representing the Resources objects.
+# These objects are obtained from the configuration file through a conversion based on the Schema classes.
+#
+
+
+class Param:
+    """Custom class to wrap a value and add more attributes."""
+
+    def __init__(self, value, default=None, valid: bool = True):
```
yuleiwan, line 20: here we could set `value=None` as default arg, so when we initialize it, we needn't pass None to the value in some cases. But just ignore this comment, since we will remove the Param class.

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Done

cli/src/pcluster/models/common.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/common_schema.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -671,6 +675,17 @@ def make_resource(self, data, **kwargs):
         return AdditionalPackages(**data)
 
 
+class DevSettingsSchema(BaseDevSettingsSchema):
+    """Represent the schema of the ImageBuilder Dev Setting."""
```
yuleiwan, line 679: wrong comment

cli/src/pcluster/schemas/common_schema.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -111,3 +112,29 @@ class TagSchema(BaseSchema):
...
+
+    chef_cookbook = fields.Str()
+    extra_chef_attributes = fields.Str()
+
+    @post_load()
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return Cookbook(**data)
+
+    @validates("extra_chef_attributes")
```
yuleiwan, line 128: I am not sure the extra_chef_attributes only supports json format or not, we can add a TODO mark for this validation

yuleiwan (commented):
looks good

cli/src/pcluster/models/imagebuilder.py
hanwen-pcluste: I'd kept it unchanged for now. Because the caller of these initialization methods is only Marshmallow, instead of any manual calls. More over, it is easier to add another field to the BaseDevSettings with **kwargs

cli/src/pcluster/schemas/imagebuilder_schema.py
hanwen-pcluste: Done

yuleiwan (approved):

PR #2434, opened 2021-02-10
3.0: Encapsulate Params metadata in Resource classes

ddeidda:
This commit replaces the usage of Param instances in resource attributes with a new internal logic inside the base resource class.
With this change resource attributes can be accessed directly instead of through the `value` property, while the related configuration information can always be retrieved through the new `get_param` method.

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -12,12 +12,11 @@
 # This module contains all the classes representing the Resources objects.
 # These objects are obtained from the configuration file through a conversion based on the Schema classes.
 #
-
 from enum import Enum
 from typing import List
 
 from pcluster.constants import CIDR_ALL_IPS, EBS_VOLUME_TYPE_IOPS_DEFAULT
-from pcluster.models.common import BaseTag, Param, Resource
+from pcluster.models.common import BaseTag, Resource
 from pcluster.utils import (
     error,
     get_availability_zone_of_subnet,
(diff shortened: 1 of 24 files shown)


cli/src/pcluster/validators/ebs_validators.py
```diff
@@ -31,7 +31,7 @@ class EbsVolumeTypeSizeValidator(Validator):
     The volume sizes of st1 and sc1 range from 500 GiB - 16 TiB(16384 GiB)
     """
 
-    def _validate(self, volume_type: Param, volume_size: Param):
+    def _validate(self, volume_type: Resource.Param, volume_size: Resource.Param):
```
hanwen-pcluste, line 34: Same question as above. When executing `ebs.validate()`, the arguments are `str`/`int` instead of `Resource.Param`. This problem is not dected because of the absence of unit tests to test the validation of a whole Resource. 
I am fine with either of the fixings below: 
1. Change the type hint to `str`/`int`, remove all '.value's in validators and remove all `Resource.Param(xxx)` in unit tests for validators
2. Change every `_register_validators()` to pass in `Resource.Param` instead of the `str`/`int`

cli/src/pcluster/models/common.py
```diff
@@ -115,10 +91,101 @@
...
+                # If other type, check if it is backed by a param; if yes, sync the param
+                param = self.__params.get(key, None)
+                if param:
+                    param._Param__value = value
+                    param._Param__implied = False
+
+        super().__setattr__(key, value)
+
+    @staticmethod
+    def init_param(value, default=None, update_policy=None):
```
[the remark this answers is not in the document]
ddeidda: It is mainly a semantic issue. 
When you assign a `Resource.Param` instance to a `Resource` attribute, the final value that will be assigned to the attribute is the *value* of the `Param`, not the `Param` itself. The purpose of this function call is basically to "mask" this behavior. Other than that, `Param` objects are strictly related to the resource they belong to, and calling its constructor should be avoided from any other place.

cli/src/pcluster/models/cluster.py
```diff
@@ -220,23 +219,23 @@
...
+        self.per_unit_storage_throughput = Resource.init_param(per_unit_storage_throughput)
+        self.backup_id = Resource.init_param(backup_id)
+        self.kms_key_id = Resource.init_param(kms_key_id)
+        self.file_system_id = Resource.init_param(file_system_id)
+        self.auto_import_policy = Resource.init_param(auto_import_policy)
+        self.drive_cache_type = Resource.init_param(drive_cache_type)
+        self.storage_type = Resource.init_param(storage_type)
 
     def _register_validators(self):
         self._add_validator(
```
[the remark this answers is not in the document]
ddeidda: We talked about this and we reach the agreement that validators should keep receiving Parameter instances rather than just values in order to be able to reference them in the list of failed params.

cli/src/pcluster/validators/ebs_validators.py
ddeidda: You are totally right; see my answer above.
We decided to adopt solution 2.

ddeidda: Actually, on a second thought we switched to solution 1. The reason is that this option goes in the same direction of this patch, i.e. removing the explicit use of the `Param` class.

hanwen-pcluste (approved):
LGTM! Thank you!

PR #2435, opened 2021-02-11
3.0: Use new model for create, delete and status command

enrico-usai:
## PclusterApi
* Define PclusterApi, proxy class for all Pcluster API commands used in the CLI.
* Add create_cluster, delete_cluster and describe_cluster in the api.
* Define ClusterInfo class to store minimal cluster information returned by the API for a running cluster.
* Region taken from the argument will be set as environment variable and it will be used by subsequent calls.

## Cluster model
* Renamed BaseCluster to BaseClusterConfig and cluster.py to cluster_config.py
* Define ClusterBucket class to store additional bucket information like artifact directory and remove on deletion.
* Define ClusterActionError as a generic error to be raised during cluster actions.
* Move awsbatch and slurm config classes into cluster_config.py.
* Define Cluster class to contain both ClusterStack, ClusterConfig and ClusterBucket
  Cluster is the controller of all the other model classes, PclusterApi should only interact with it.

## Create
* Move awsbatch specific logic to upload batch resources in batch model.
* Deprecate cluster-template, extra-parameters and tags parameters for create command.
* Add disable-update-check parameter for create command.
* Remove evaluate_tags and _check_for_updates functions.
* Parse yaml dict rather than ini file and PclusterConfig structure.
* Removed region from create_s3_bucket utility function, using the one from the env.

## Delete
* Move _terminate_cluster_nodes in the Cluster.
* Move _persist_cloudwatch_log_groups, _get_unretained_cw_log_group_resource_keys,
  _persist_stack_resources in the ClusterStack
* persist_cloudwatch_log_groups logic has been changed because we no longer have substacks.
* Deprecate config parameter for delete command.
* Remove loading of configuration file.
* Adapt existing unit test.
* Removed get_cluster_substacks utility function.
* Unified delete and _delete functions.

## Status
* Deprecate config parameter for status command.
* Remove loading of configuration file.

## AWS Api
* Add dynamodb client to AWSApi and some actions to existing clients.
* Define a Stack resource to interact with CFN stack.
* Reorder catching in AWSExceptionHandler because ParamValidationError is a subclass of BotoCoreError
* Improve describe_subnets and list_instances_ids to return a complete list rather than a paginator.

## CDK
* Added bucket class as a separate information required to build the template.
* Renamed ClusterStack to ClusterCdkStack and ImageBuilderStack to ImageBuilderCdkStack.


cli/src/common/aws/aws_resources.py
```diff
@@ -0,0 +1,69 @@
...
+        return AWSApi().instance().cfn.describe_stack(self.name).get("StackStatus")
+
+    def delete(self):
+        """Delete stack."""
+        try:
+            # delete_stack does not raise an exception if stack does not exist
+            # Use describe_stacks to explicitly check if the stack exists
+            AWSApi().instance().cfn.delete_stack(self.name)
+
+            if self.updated_status() == "DELETE_FAILED":
```
hanwen-pcluste, line 65: Deletion of stack takes time. Is doing the check here too early. In other words, the stack status here is always "DELETE_IN_PROGRESS"?

cli/src/common/pcluster_api.py
```diff
@@ -0,0 +1,165 @@
...
+
+            if validation_failures:
+                # TODO skip validation errors
+                return ApiFailure(message="Configuration is invalid", validation_failures=validation_failures)
+            else:
+                # Add tags information to the cluster
+                version = get_installed_version()
+                tags = cluster.tags or []
+                tags.append(Tag(key="Version", value=version))
+                tags = [{"Key": tag.key, "Value": tag.value} for tag in tags]
```
hanwen-pcluste, line 80: Why append the default tag here, instead of doing it in model layer?

cli/src/common/pcluster_api.py
```diff
@@ -0,0 +1,165 @@
...
+                        stack_name=stack_name,
+                        template_url=cluster.template_url,
+                        disable_rollback=disable_rollback,
+                        tags=tags,
+                    )
+
+                    stack = ClusterStack(cluster_name)
+                    LOGGER.debug("StackId: %s", stack.id)
+                    LOGGER.info("Status: %s", stack.status)
+                    return ClusterInfo(stack, cluster_name)
```
hanwen-pcluste, line 106: Are we waiting for the stack creation complete before return the cluster info?

cli/src/common/pcluster_api.py
```diff
@@ -0,0 +1,165 @@
...
+
+                except (ClusterActionError, Exception) as e:
```
hanwen-pcluste, line 108: ?Why except ClusterActionError with Exception? I believe ClusterActionError is a subclass of Exception?

cli/src/common/pcluster_api.py
enrico-usai: Nope, the goal is having a stateless command. It means the create command starts the creation and that's all, then the user need to execute the status command and eventually create a loop to see the progress.

cli/src/common/pcluster_api.py
enrico-usai: It's a good point. 
I don't think that the version tag should be added when parsing the config because if you are running version 3.1 and you parse a configuration of an existing cluster (created with version 3.0) you'll add 3.1 in the config and it's wrong.
I think this particular tag should be added only when creating the cluster, we only need it at cluster creation time, but let's talk about it. This can be interesting for @yuleiwan too.

cli/src/common/aws/aws_resources.py
enrico-usai: Good point.

cli/src/common/pcluster_api.py
hanwen-pcluste: Very true, the default tag should be here

cli/src/pcluster/validators/awsbatch_validators.py
```diff
@@ -113,23 +116,24 @@ class AwsbatchInstancesArchitectureCompatibilityValidator(Validator):
...
-            compute_architectures = get_supported_architectures_for_instance_type(instance_type)
-            if architecture not in compute_architectures:
-                self._add_failure(
-                    "The specified compute instance type ({0}) supports the architectures {1}, none of which are "
-                    "compatible with the architecture supported by the head node instance type ({2}).".format(
-                        instance_type, compute_architectures, architecture
-                    ),
-                    FailureLevel.ERROR,
-                )
+        if instance_types:
```
hanwen-pcluste, line 119: `instance_types` is required parameter, isn't it? If yes, do we need this conditional statement?

cli/src/pcluster/commands.py
```diff
@@ -196,96 +193,66 @@ def _validate_cluster_name(cluster_name):
...
-        stack_name = utils.get_stack_name(args.cluster_name)
+def create(args):
+    """Create cluster."""
+    LOGGER.info("Beginning cluster creation for cluster: %s", args.cluster_name)
+    LOGGER.debug("CLI args: %s", str(args))
 
-        # merge tags from configuration, command-line and internal ones
-        tags = _evaluate_tags(pcluster_config, preferred_tags=args.tags)
+    if not args.disable_update_check:
+        utils.check_if_latest_version()
```
hanwen-pcluste, line 240: Why `check_if_latest_version` is only in create_cluster command?

enrico-usai: We don't want to show a warning message "there is a new version of parallelcluster" for every command execution, but only when creating new clusters.

cli/src/common/aws/aws_resources.py
enrico-usai: Changed.

cli/src/common/pcluster_api.py
enrico-usai: Fixed

cli/src/pcluster/validators/awsbatch_validators.py
enrico-usai: The validator should be safe by itself, if we change something in the model the validator shouldn't fail.

cli/src/pcluster/models/cluster_config.py
```diff
@@ -0,0 +1,1188 @@
...
+        for queue in self.scheduling.queues:
+            for compute_resource in queue.compute_resources:
+                self._execute_validator(
+                    InstanceArchitectureCompatibilityValidator,
+                    instance_type=compute_resource.instance_type,
+                    architecture=self.head_node.architecture,
+                )
+
+
+class ClusterBucket:
```
hanwen-pcluste, line 1163: Why put the class here instead of cluster.py? Or should it inheritate from Resource?

cli/src/pcluster/models/cluster.py
```diff
@@ -12,939 +12,339 @@
...
+                {"Name": "tag:Application", "Values": [self.stack_name]},
+                {"Name": "instance-state-name", "Values": ["pending", "running", "stopping", "stopped"]},
+                {"Name": "tag:aws-parallelcluster-node-type", "Values": [str(NodeType.compute)]},
+            ]
+            result = AWSApi().instance().ec2.describe_instances(filters)
+            instances = []
+            for instance in result.get("Instances", []):
+                instances.append(instance.get("InstanceId"))
+
+            for instance_ids in grouper(instances, 100):
```
hanwen-pcluste, line 343: Why using grouper?

cli/src/common/aws/aws_resources.py
```diff
@@ -0,0 +1,76 @@
...
+
+        If the stack doesn't exist it raises an exception.
+        """
+        self.name = stack_name
+        self.region = get_region()
+        try:
+            self._stack_data = AWSApi().instance().cfn.describe_stack(stack_name)
+        except AWSClientError as e:
+            if "does not exist" in str(e):
+                raise StackActionError(f"The stack {stack_name} does not exist.")
```
hanwen-pcluste, line 38: Why creating another Exception class, instead of using AWSClientError?

cli/src/common/boto3/ec2.py
```diff
@@ -31,7 +36,15 @@
     @AWSExceptionHandler.handle_client_exception
     def describe_subnets(self, subnet_ids):
         """Return a list of subnets."""
-        return self._paginate_results(self._client.describe_subnets, SubnetIds=subnet_ids)
+        return self._client.describe_subnets(SubnetIds=subnet_ids).get("Subnets", [])
```
hanwen-pcluste, line 39: Why not paginate the result?

cli/src/pcluster/utils.py
```diff
@@ -118,7 +120,7 @@
 
 def get_region():
     """Get AWS_DEFAULT_REGION from the environment."""
-    return os.environ.get("AWS_DEFAULT_REGION")
+    return os.environ.get("AWS_DEFAULT_REGION") or boto3.session.Session().region_name
```
hanwen-pcluste, line 123: You could write `return boto3.session.Session().region_name`. The `boto3.session...` will cover the environ var

cli/src/common/boto3/ec2.py
enrico-usai: One reason is that the paginate is useful only when the result contains more than [1000 items](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html).
This describe subnets method is accepting a list of subnets (that is a limited one) so we don't need to paginate.

The other reason is that `@AWSExceptionHandler.handle_client_exception` is not able to correctly manage the handling the exception iwhen there is a generator from the pagination. Still to be investigated.

cli/src/pcluster/utils.py
enrico-usai: Fixed

cli/src/common/boto3/ec2.py
hanwen-pcluste: boto3 is not behaving as neatly as AWS CLI. When I tried to [retrieve free tier instance type](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/utils.py#L1173), I had to use pagination, or I might get empty result

cli/src/common/aws/aws_resources.py
enrico-usai: Because I want to manage StackActionError differently from AWSClientError.
When getting a StackActionError "the **stack** pcluster-yyy doesn't exist" I can catch it and show: "the **cluster** yyy doesn't exist".

cli/src/api/pcluster_api.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class PclusterApi:
+    """Proxy class for all Pcluster API commands used in the CLI."""
+
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def create_cluster(cluster_config: dict, cluster_name: str, region: str = None, disable_rollback: bool = False):
```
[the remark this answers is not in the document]
enrico-usai: It's a good question, it's still not supported, all the internal boto3 calls will use the default region from boto3.
We'll need to set it in the environment to be used by all the subsequent code.

cli/src/common/boto3/cfn.py
```diff
@@ -28,3 +29,33 @@ def create_stack(self, stack_name, template_url, disable_rollback, tags):
...
+    def delete_stack(self, stack_name: str):
+        """Delete CFN stack."""
+        return self._client.delete_stack(StackName=stack_name)
+
+    @AWSExceptionHandler.handle_client_exception
+    def update_stack(self, stack_name: str, updated_template: str, params: list):
+        """Update CFN stack."""
+        return self._client.update_stack(
+            StackName=stack_name,
+            TemplateBody=json.dumps(updated_template, indent=2),  # Indent so it looks nice in the console
```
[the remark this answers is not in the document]
enrico-usai: I just inserted the function I need for now, but we can extend/change it.

cli/src/common/pcluster_api.py
enrico-usai: I resterored the previous behaviour from the CLI, by calling multiple times the describe_cluster api call.

cli/src/pcluster/models/cluster.py
enrico-usai: Because AWSApi automatically perform a pagination and return the full list.
Here we need to group the instances in small_group because it's suggested by boto3 doc: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.terminate_instances

cli/src/common/boto3/ec2.py
enrico-usai: Fixed, re-added pagination.

cli/src/pcluster/utils.py
```diff
@@ -117,14 +119,22 @@ def update_stack_template(stack_name, updated_template, cfn_parameters):
...
+    Get region used internally for all the AWS calls.
+
+    The region from the env has higher priority because it can be explicitly set from the code (e.g. unit test).
+    """
+    region = os.environ.get("AWS_DEFAULT_REGION")
+    if not region:
+        bot3_session = boto3.session.Session()
+        region = bot3_session.region_name
+
+    return region
```
hanwen-pcluste, line 132: Minor: why not return `boto3.session.Session().region_name` directly?

enrico-usai: Fixed

cli/src/pcluster/commands.py
```diff
@@ -196,96 +193,88 @@ def _validate_cluster_name(cluster_name):
...
-    HitConverter(pcluster_config).convert()
+def _parse_config_file(config_file, fail_on_config_file_absence=True):
+    """
+    Parse the config file and initialize config_file and config_parser attributes.
 
-    # get CFN parameters, template url and tags from config
-    storage_data = pcluster_config.to_storage()
-    cfn_params = storage_data.cfn_params
+    :param config_file: The config file to parse
+    :param fail_on_config_file_absence: set to true to raise SystemExit if config file doesn't exist
```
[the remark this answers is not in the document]
enrico-usai: Yes, it's useless, but it's still used by old code (pcluster_config.py). We'll remove it once cleaned up all the old code. Thanks for the comment!

cli/src/pcluster/commands.py
```diff
@@ -196,96 +193,88 @@ def _validate_cluster_name(cluster_name):
...
-        sys.exit(0)
-    except KeyError as e:
-        LOGGER.critical("ERROR: KeyError - reason:\n%s", e)
-        utils.cleanup_s3_resources(bucket_name, artifact_directory, cleanup_bucket)
-        sys.exit(1)
-    except Exception as e:
-        LOGGER.critical(e)
-        utils.cleanup_s3_resources(bucket_name, artifact_directory, cleanup_bucket)
-        sys.exit(1)
+            result = PclusterApi().describe_cluster(cluster_name=args.cluster_name, region=utils.get_region())
```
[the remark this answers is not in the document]
enrico-usai: Good catch! I'll address it.

cli/src/pcluster/models/cluster.py
```diff
@@ -12,939 +12,340 @@
...
+        tags.append(Tag(key="Version", value=version))
+        tags = [{"Key": tag.key, "Value": tag.value} for tag in tags]
+
+        # Create bucket if needed
+        self._setup_cluster_bucket()
+
+        creation_result = None
+        try:
+            # Create template if not provided by the user
+            if not (self.config.dev_settings and self.config.dev_settings.cluster_template):
```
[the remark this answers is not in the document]
enrico-usai: The custom cluster_template is specified in the dev_settings section of the configuration file :D

BTW I understand the question. If cluster_template is provided we don't need to generate the template but the config file is required to know that you have a custom template and it might be useful because the custom template can have an user-data that uses the data from the configurafion file.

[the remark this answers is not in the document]
enrico-usai: The custom cluster_template is specified in the dev_settings section of the configuration file :D

BTW I understand the question. If cluster_template is provided we don't need to generate the template but the config file is required to know that you have a custom template and it might be useful because the custom template can have an user-data that uses the data from the configurafion file.

cli/src/pcluster/commands.py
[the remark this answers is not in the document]
enrico-usai: Yes, it's useless, but it's still used by old code (pcluster_config.py). We'll remove it once cleaned up all the old code. Thanks for the comment!

cli/src/pcluster/commands.py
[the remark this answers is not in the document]
enrico-usai: Good catch! I'll address it.

PR #2437, opened 2021-02-11
3.0: Migrate validators: volume id, intel, S3

hanwen-pcluste:
Migrate SharedEBSVolumeIdValidator, IntelHpcArchitectureValidator, IntelHpcOsValidator, S3BucketUriValidator, S3BucketValidator

Fix a bug in SecurityGroupsValidator, this bug shows that our unit test coverage is low

Remove old url validators (already moved by yuleiwan)

Split BaseCluster._register_validators to pass code-linters check for complex code

### DIFF
--- a/cli/src/pcluster/validators/s3_validators.py
+++ b/cli/src/pcluster/validators/s3_validators.py
@@ -3,9 +3,11 @@ from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
+from common.aws.aws_api import AWSApi
 from common.boto3.common import AWSClientError
 from common.boto3.s3 import S3Client
 from pcluster.validators.common import FailureLevel, Validator
+from pcluster.validators.utils import get_bucket_name_from_s3_url
 
 
 class UrlValidator(Validator):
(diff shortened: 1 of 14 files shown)


cli/src/pcluster/validators/s3_validators.py
```diff
@@ -64,3 +66,62 @@ def _validate_s3_uri(self, s3_url):
...
+            message = _process_generic_s3_bucket_error(client_error, bucket)
+            self._add_failure(message, FailureLevel.ERROR)
+        except ParamValidationError as validation_error:
+            self._add_failure(
+                "Error validating parameter '{0}'. Failed with exception: {1}".format(bucket, str(validation_error)),
+                FailureLevel.ERROR,
+            )
+
+
+def _process_generic_s3_bucket_error(client_error, bucket_name):
```
enrico-usai, line 116: You're using S3Client, it means that the ClientError will be captured by the boto3 layer and you'll have instead a AWSClientError.

You should move this logic in the boto3 layer.

cli/src/pcluster/validators/s3_validators.py
```diff
@@ -64,3 +66,62 @@ def _validate_s3_uri(self, s3_url):
...
+        else:
+            self._add_failure(f"The value '{url}' is not a valid S3 URI.", FailureLevel.ERROR)
+
+
+class S3BucketValidator(Validator):
+    """S3 Bucket Validator."""
+
+    def _validate(self, bucket):
+
+        s3_client = S3Client()
```
enrico-usai, line 92: You should use AWSApi().instance().s3 rather than calling the client directly.

cli/src/pcluster/validators/s3_validators.py
hanwen-pcluste: Done

cli/src/pcluster/validators/s3_validators.py
hanwen-pcluste: Done

enrico-usai (approved):

PR #2459, opened 2021-02-23
Upgrade to Slurm 20.11.4

rexcsn:
* Add gpu_type to internal config to specify instance GPU model as GRES GPU Type in gres.conf
* Updated integ test to submit test submitting job with specific model of GPU
* Add test for checking slurm health check behavior for CLOUD nodes. Previously slurm will not perform node health check before ResumeTimeout expires. This is fixed in 20.11.4
* Modify status check test to avoid slurm health check and use static capacity
* Add additional test to make sure nodeaddr and nodehostname are reset on power_down. This should verify the behavior of relying on cloud_reg_addrs option


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
+    )
+    time.sleep(15)
+    scheduler_commands.cancel_job(kill_job_id)
+    # Assert dynamic nodes are in DOWN
+    _wait_for_compute_nodes_states(scheduler_commands, dynamic_nodes, expected_states=["down*", "down"])
+    # Assert dynamic nodes are reset
+    _wait_for_node_reset(scheduler_commands, static_nodes=[], dynamic_nodes=dynamic_nodes)
+    assert_num_instances_in_cluster(cluster_name, region, len(static_nodes))
+    # Assert ec2_status_check code path is not triggered
+    assert_no_msg_in_logs(
```
demartinofra, line 429: this is a weak assertion. If we change the log message this becomes useless and we won't even realize. Is there any more robust check we can perform? for example assert for the presence of an error message? if not we could leave it as is

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
         ["Setting nodes failing health check type ec2_health_check to DRAIN"],
     )
     scheduler_commands.cancel_job(kill_job_id)
-    # Assert dynamic nodes are reset
-    _wait_for_node_reset(scheduler_commands, static_nodes=[], dynamic_nodes=dynamic_nodes)
+    # Assert static nodes are reset
+    _wait_for_node_reset(scheduler_commands, static_nodes=static_nodes, dynamic_nodes=[])
     assert_num_instances_in_cluster(cluster_name, region, len(static_nodes))
+    # Reset SlurmdTimeout to 180s
+    _set_slurmd_timeout(remote_command_executor, timeout=180)
```
demartinofra, line 464: if we change this value in the cookbook the test will be misaligned. Can we restore the previous value without hardcoding the value?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
+    assert_no_msg_in_logs(
+        remote_command_executor,
+        ["/var/log/parallelcluster/clustermgtd"],
+        ["Setting nodes failing health check type ec2_health_check to DRAIN"],
     )
-    # Assert that dynamic nodes with failing status check are put into DRAIN
-    # Can take up to 15 mins for status check to show
+
+
+def _test_status_check_replacement(
```
demartinofra, line 436: nit: test_ec2_status_check_replacement?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
-            "other_options": "-a 1-{} --exclusive --no-requeue".format(num_dynamic_nodes),
-        }
+    # Assert that the default SlurmdTimeout=180 is in effect
+    _assert_slurmd_timeout(remote_command_executor, timeout=180)
+    # Nodes with networking failures should fail slurm health check before failing ec2_status_check
+    # Test on freshly launched dynamic nodes
+    kill_job_id = _submit_kill_networking_job(
+        remote_command_executor, scheduler_commands, partition, node_type="dynamic", num_nodes=num_dynamic_nodes
+    )
+    time.sleep(15)
```
demartinofra, line 421: nit: let's add a comment every time we need to use an arbitrary sleep

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
+    scheduler_commands.cancel_job(kill_job_id)
```
demartinofra, line 422: why do we need to cancel the job?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -393,33 +411,57 @@ def _test_status_check_replacement(
...
         remote_command_executor,
         ["/var/log/parallelcluster/clustermgtd"],
         ["Setting nodes failing health check type ec2_health_check to DRAIN"],
     )
     scheduler_commands.cancel_job(kill_job_id)
-    # Assert dynamic nodes are reset
-    _wait_for_node_reset(scheduler_commands, static_nodes=[], dynamic_nodes=dynamic_nodes)
+    # Assert static nodes are reset
+    _wait_for_node_reset(scheduler_commands, static_nodes=static_nodes, dynamic_nodes=[])
     assert_num_instances_in_cluster(cluster_name, region, len(static_nodes))
```
demartinofra, line 462: do we need to check logs to make sure the replacement happened due to failing health checks?

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ CHANGELOG
 - Make `key_name` parameter optional to support cluster configurations without a key pair. 
 - Remove support for Python 3.4
 - Root volume size increased from 25GB to 35GB on all AMIs. Minimum root volume size is now 35GB.
+- Upgrade Slurm to version 20.11.4.
```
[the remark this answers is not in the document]
rexcsn: Done

cli/src/pcluster/config/json_param_types.py
```diff
@@ -303,6 +303,7 @@ def refresh_compute_resource(self, compute_resource_section):
             # Set gpus according to instance features
             gpus = instance_type_info.gpu_count()
             compute_resource_section.get_param("gpus").value = gpus
+            compute_resource_section.get_param("gpu_type").value = instance_type_info.gpu_type()
```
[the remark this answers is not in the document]
rexcsn: The parameter will still be there in the Json because it is defined in mappings.py? Are you saying we should use an empty string "" instead of "no_gpu_type" as the default?

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: Thanks for the callout, done.

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: Because the job will hang, so we can cancel it and avoid waiting for the job to fail

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: Great point, adding assertion to assert messages in slurmctld log to check that slurm took action to mark the node as DOWN

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: Done

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: Yes, we check that `Setting nodes failing health check type ec2_health_check to DRAIN` is in clustermgtd log above

tests/integration-tests/tests/schedulers/test_slurm.py
rexcsn: I think that hardcoding the value would be more clear. Since we are changing the slurm configuration across test cases, the time could fail if value is not reset correctly, or set to some value we are not aware of. The hardcoded value will make updating value in cookbook a bit more complicated, but IMO it might help us to spot issues with our tests more easily

cli/src/pcluster/config/json_param_types.py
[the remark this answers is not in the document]
demartinofra: either empty "" or None so that it corresponds to a false value in Python. But feel free to leave it as is if you prefer.

demartinofra (approved):

PR #2460, opened 2021-02-23
3.0: minor changes

enrico-usai:
* Add CFN boto3 client and some methods
* Save original config into cluster model
* Add guard for empty list in validators and other cleanup


cli/src/common/utils.py
```diff
@@ -21,9 +21,9 @@ def download_file(url):
     return response.read().decode("utf-8")
 
 
-def load_yaml_dict(config_file):
+def load_yaml_dict(file_path):
```
lukeseawalker, line 24: load_yaml_file?

PR #2466, opened 2021-02-23
Add a timeout for dcv connect command

enrico-usai:

### DIFF
--- a/cli/src/pcluster/dcv/connect.py
+++ b/cli/src/pcluster/dcv/connect.py
@@ -28,7 +28,7 @@ class DCVConnectionError(Exception):
 
 
 def _check_command_output(cmd):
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/dcv/connect.py
```diff
@@ -28,7 +28,7 @@ class DCVConnectionError(Exception):
 
 
 def _check_command_output(cmd):
-    return sub.check_output(cmd, shell=True, universal_newlines=True, stderr=sub.STDOUT).strip()
+    return sub.check_output(cmd, shell=True, timeout=5000, universal_newlines=True, stderr=sub.STDOUT).strip()
```
rexcsn, line 31: I believe the `timeout` parameter is in seconds, are we trying to do 5 second timeout here?
I think we can also do a longer timeout like 10 or 15 second, if commands are subject to slow networkings in China

PR #2474, opened 2021-02-25
Fix instance type sanity checks

ddeidda:
This commmit replaces the latest linux image id with the actual cluster ami in dryrun run-instances calls performed during sanity checks.

This new behaviour fixes in particular the issue occurring when the architecture of the instance types is different from the one of the latest linux AMI, and provides a more reliable check when custom amis are used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 2.10.2
 ------
+**ENHANCEMENTS**
+
+- Improve cluster config validation by using cluster target AMI when invoking RunInstances in dryrun mode
 
 **CHANGES**
 
@@ -13,6 +16,7 @@ CHANGELOG
 **BUG FIXES**
 
 - Fix `enable_efa` parameter validation when using Centos8 and Slurm or ARM instances.
+- Fix sanity checks with ARM instance types by using cluster AMIs when performing validation      
 
 2.10.1
 ------
--- a/cli/src/pcluster/cluster_model.py
+++ b/cli/src/pcluster/cluster_model.py
@@ -18,6 +18,7 @@ from botocore.exceptions import ClientError
 from pcluster.utils import (
     error,
     get_availability_zone_of_subnet,
+    get_installed_version,
     get_supported_az_for_one_instance_type,
     is_hit_enabled_cluster,
 )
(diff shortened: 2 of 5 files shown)


cli/src/pcluster/cluster_model.py
```diff
@@ -115,6 +116,53 @@ def _ec2_run_instance(self, pcluster_config, **kwargs):  # noqa: C901 FIXME!!!
...
+                        },
+                        {"Name": "owner-alias", "Values": ["amazon"]},
+                        {"Name": "is-public", "Values": ["true"]},
+                    ],
+                )
+                .get("Images")
+            )
+        except ClientError as e:
+            error(
+                "Unable to retrieve official image id for base_os='{0}' and architecture='{1}': {3}",
```
demartinofra, line 138: I think you missed {2}

cli/src/pcluster/cluster_model.py
```diff
@@ -115,6 +116,53 @@ def _ec2_run_instance(self, pcluster_config, **kwargs):  # noqa: C901 FIXME!!!
...
+            images = (
+                boto3.client("ec2")
+                .describe_images(
+                    Filters=[
+                        {
+                            "Name": "name",
+                            "Values": ["{0}*".format(self._get_official_image_name_prefix(os, architecture))],
+                        },
+                        {"Name": "owner-alias", "Values": ["amazon"]},
+                        {"Name": "is-public", "Values": ["true"]},
```
demartinofra, line 131: this can be problematic between the time we bump version and the time we release the AMIs. probably for now we just want to remove this and rely on the canaries to make sure the AMIs are public after we release.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
 **BUG FIXES**
 
 - Fix `enable_efa` parameter validation when using Centos8 and Slurm or ARM instances.
+- Fix AMI/instance types by using the actual cluster AMI for the target architecture
```
demartinofra, line 16: I think there is something missing in this sentence. To make it more meaningful for the user shall we mention that we are fixing sanity check for clusters with ARM architectures by using cluster AMI when performing validations?

cli/src/pcluster/models/hit/hit_cluster_model.py
```diff
@@ -72,7 +72,7 @@ def test_configuration(self, pcluster_config):
             else {}
         )
         try:
-            latest_alinux_ami_id = self._get_latest_alinux_ami_id()
+            cluster_ami_id = self._get_cluster_ami_id(pcluster_config)
```
demartinofra, line 75: cluster_ami_id can be None if the query does not return any result. Shall we handle this?

cli/src/pcluster/models/sit/sit_cluster_model.py
```diff
@@ -111,7 +111,7 @@ def test_configuration(self, pcluster_config):
         )
 
         try:
-            latest_alinux_ami_id = self._get_latest_alinux_ami_id()
+            cluster_ami_id = self._get_cluster_ami_id(pcluster_config)
```
demartinofra, line 114: same here

cli/src/pcluster/cluster_model.py
ddeidda: Fixed

cli/src/pcluster/cluster_model.py
ddeidda: Done

CHANGELOG.md
ddeidda: Fixed

cli/src/pcluster/models/hit/hit_cluster_model.py
ddeidda: It should never happen in theory, but I agree. Fixed.

cli/src/pcluster/models/sit/sit_cluster_model.py
ddeidda: Fixed

cli/src/pcluster/cluster_model.py
```diff
@@ -115,6 +116,62 @@ def _ec2_run_instance(self, pcluster_config, **kwargs):  # noqa: C901 FIXME!!!
...
+
+    def _get_cluster_ami_id(self, pcluster_config):
+        """Get the image id of the cluster."""
+        os = pcluster_config.get_section("cluster").get_param_value("base_os")
+        architecture = pcluster_config.get_section("cluster").get_param_value("architecture")
+        custom_ami = pcluster_config.get_section("cluster").get_param_value("custom_ami")
+        try:
+            return custom_ami if custom_ami else self._get_official_image_id(os, architecture)
+        except Exception as e:
+            pcluster_config.error("Error when resolving cluster ami id: {0}".format(e))
```
demartinofra, line 173: In this case there is no return value. What happens in the caller?

demartinofra (approved):

cli/src/pcluster/cluster_model.py
ddeidda: Added return `None` in that case

rexcsn (approved):

PR #2479, opened 2021-02-26
3.0: Pcluster build image command

yuleiwan:
# Use the new model for build-image command

## PclusterApi
* Created an ImageInfo class to contain the information of API response
* Add build_image API method 

## Imagebuilder
* Add a sub class ImagebuilderStack, that contains imagebuilder stack info
* Add an Imagebuilder class contains config and stack info. This class manages methods from API call, add `create` function to create imagebuilder stack.

## Imagebuilder_config
* rename imagebuilder resource class to imagebuilder_config
* Remove name attribute from image config

## imagebuilder cdk
* Use image_name from cli rather than config file to generate ImagebuilderImage AWS resource.

## tests 
manual test for pcluster `build-image` command.
config.yaml
```
Build:
  InstanceType: c5.xlarge
  ParentImage: ami-0185634c5a8a37250
```
command
```
pcluster build-image -i test-pcluster -c config.yaml -r us-east-1
```
output
```
Building AWS ParallelCluster AMI. This could take a while...
Creating stack named: test-pcluster
Status: CREATE_IN_PROGRESS
Response:
{'image': '{"image_name": "test-pcluster", "imagebuild_status": "BUILD_IN_PROGRESS", "stack_status": "CREATE_IN_PROGRESS", "stack_arn": "arn:aws:cloudformation:us-east-1:xxxxxxxxxxxxx:stack/test-pcluster/4d532880-7aef-11eb-b853-0e71a2c850a7", "region": "us-east-1", "version": "2.10.1"}'}
```

### DIFF
--- a/cli/src/api/pcluster_api.py
+++ b/cli/src/api/pcluster_api.py
@@ -17,6 +17,7 @@ from packaging import version
 from common.aws.aws_api import AWSApi
 from pcluster.cli_commands.compute_fleet_status_manager import ComputeFleetStatus
 from pcluster.models.cluster import Cluster, ClusterActionError, ClusterStack
+from pcluster.models.imagebuilder import ImageBuilder, ImageBuilderActionError
 from pcluster.utils import get_installed_version, get_region
 
 LOGGER = logging.getLogger(__name__)
@@ -67,6 +68,21 @@ class FullClusterInfo(ClusterInfo):
         self.compute_instances = cluster.compute_instances
         # Config info
         self.scheduler = cluster.config.scheduling.scheduler
+
+
+class ImageInfo:
+    """Minimal representation of a building image."""
+
+    def __init__(self, imagebuilder: ImageBuilder):
+        self.image_name = imagebuilder.stack.name
+        self.imagebuild_status = imagebuilder.imagebuild_status
+        self.stack_status = imagebuilder.stack.status
+        self.stack_arn = imagebuilder.stack.id
+        self.region = get_region()
+        self.version = imagebuilder.stack.version
+
+    def __repr__(self):
+        return json.dumps(self.__dict__)
 
 
 class PclusterApi:
@@ -188,3 +204,25 @@ class PclusterApi:
     @staticmethod
     def _is_version_2(cluster):
         return version.parse(cluster.stack.version) < version.parse("3.0.0")
+
+    @staticmethod
+    def build_image(imagebuilder_config: dict, image_name: str, region: str, disable_rollback: bool = True):
+        """
+        Load imagebuilder model from imagebuilder_config and create stack.
+
+        :param imagebuilder_config: imagebuilder configuration (yaml dict)
+        :param image_name: the image name(the same as cfn stack name)
+        :param region: AWS region
+        :param disable_rollback: Disable rollback in case of failures
+        """
+        try:
+            # Generate model from imagebuilder config dict
+            if region:
+                os.environ["AWS_DEFAULT_REGION"] = region
+            imagebuilder = ImageBuilder(image_name, imagebuilder_config)
+            imagebuilder.create(disable_rollback)
+            return ImageInfo(imagebuilder)
+        except ImageBuilderActionError as e:
+            return ApiFailure(str(e), e.validation_failures)
+        except Exception as e:
+            return ApiFailure(str(e))
--- a/cli/src/common/boto3/cfn.py
+++ b/cli/src/common/boto3/cfn.py
@@ -21,8 +21,19 @@ class CfnClient(Boto3Client):
         super().__init__("cloudformation")
 
     @AWSExceptionHandler.handle_client_exception
-    def create_stack(self, stack_name: str, template_url: str, disable_rollback: bool, tags: list):
+    def create_stack(self, stack_name: str, disable_rollback: bool, tags: list, template_body: str):
         """Create CFN stack by using the given template."""
+        return self._client.create_stack(
+            StackName=stack_name,
+            TemplateBody=template_body,
+            Capabilities=["CAPABILITY_IAM"],
+            DisableRollback=disable_rollback,
+            Tags=tags,
+        )
+
+    @AWSExceptionHandler.handle_client_exception
+    def create_stack_from_url(self, stack_name: str, disable_rollback: bool, tags: list, template_url: str):
+        """Create CFN stack by using the given template url."""
         return self._client.create_stack(
             StackName=stack_name,
             TemplateURL=template_url,
@@ -74,3 +85,8 @@ class CfnClient(Boto3Client):
             for stack in self._paginate_results(self._client.describe_stacks)
             if stack.get("ParentId") is None and stack.get("StackName").startswith(PCLUSTER_STACK_PREFIX)
         ]
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_stack_resource(self, stack_name: str, logic_resource_id: str):
+        """Get stack resource information."""
+        self._client.describe_stack_resource(StackName=stack_name, LogicalResourceId=logic_resource_id)
--- a/cli/src/pcluster/cli.py
+++ b/cli/src/pcluster/cli.py
@@ -24,7 +24,6 @@ from botocore.exceptions import NoCredentialsError
 import pcluster.cli_commands.update as pcluster_update
 import pcluster.commands as pcluster
 import pcluster.configure.easyconfig as easyconfig
-import pcluster.createami as createami
 import pcluster.utils as utils
 from pcluster.dcv.connect import dcv_connect
 
@@ -79,8 +78,8 @@ def stop(args):
     pcluster.stop(args)
 
 
-def create_ami(args):
-    createami.create_ami(args)
+def build_image(args):
+    pcluster.build_image(args)
 
 
 def config_logger():
@@ -297,69 +296,17 @@ Returns an ssh command with the cluster username and IP address pre-populated::
     pssh.set_defaults(func=ssh)
 
     # createami command subparser
-    pami = subparsers.add_parser(
-        "createami", help="(Linux/macOS) Creates a custom AMI to use with AWS ParallelCluster."
-    )
-    pami.add_argument(
-        "-ai",
-        "--ami-id",
-        dest="base_ami_id",
-        required=True,
-        help="Specifies the base AMI to use for building the AWS ParallelCluster AMI.",
-    )
-    pami.add_argument(
-        "-os",
-        "--os",
-        dest="base_ami_os",
-        required=True,
-        help="Specifies the OS of the base AMI. "
-        "Valid options are: alinux, ubuntu1604, ubuntu1804, centos7, centos8.",
-    )
+    pami = subparsers.add_parser("build-image", help="Creates a custom AMI to use with AWS ParallelCluster.")
     pami.add_argument(
         "-i",
-        "--instance-type",
-        dest="instance_type",
-        default="t2.xlarge",
-        help="Sets instance type to build the ami on. Defaults to t2.xlarge.",
-    )
-    pami.add_argument(
-        "-ap",
-        "--ami-name-prefix",
-        dest="custom_ami_name_prefix",
-        default="custom-ami-",
-        help="Specifies the prefix name of the resulting AWS ParallelCluster AMI.",
-    )
-    pami.add_argument(
-        "-cc",
-        "--custom-cookbook",
-        dest="custom_ami_cookbook",
-        help="Specifies the cookbook to use to build the AWS ParallelCluster AMI.",
-    )
-    pami.add_argument(
-        "--post-install",
-        dest="post_install_script",
-        help="Specifies the post install script to use to build the AWS ParallelCluster AMI.",
-    )
-    pami.add_argument(
-        "--no-public-ip",
-        dest="associate_public_ip",
-        action="store_false",
-        default=True,
-        help="Do not associate public IP to the Packer instance. Defaults to associate public ip",
+        "--image-name",
+        dest="image_name",
+        required=True,
+        help="Specifies the image name to use for building the AWS ParallelCluster AMI.",
     )
     _addarg_config(pami)
-    pami_group1 = pami.add_argument_group("Build AMI by using VPC settings from configuration file")
-    pami_group1.add_argument(
-        "-t",
-        "--cluster-template",
-        help="Specifies the 'cluster' section of the configuration file to retrieve VPC settings.",
-    )
-    pami_group2 = pami.add_argument_group("Build AMI in a custom VPC and Subnet")
-    pami_group2.add_argument("--vpc-id", help="Specifies the VPC to use to build the AWS ParallelCluster AMI.")
-    pami_group2.add_argument("--subnet-id", help="Specifies the Subnet to use to build the AWS ParallelCluster AMI.")
     _addarg_region(pami)
-    pami.set_defaults(template_url=None)
-    pami.set_defaults(func=create_ami)
+    pami.set_defaults(func=build_image)
 
     # configure command subparser
     pconfigure = subparsers.add_parser("configure", help="Start the AWS ParallelCluster configuration.")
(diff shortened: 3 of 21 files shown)


cli/src/api/pcluster_api.py
```diff
@@ -43,6 +44,21 @@ def __repr__(self):
         return json.dumps(self.__dict__)
 
 
+class ImageInfo:
+    """Minimal representation of a building image."""
+
+    def __init__(self, imagebuilder_manager: ImageBuilderManager):
```
enrico-usai, line 50: If all the information are from the stack you can pass the imagebuilder_manager.stack rather than passing the entire manager. This helps for mocking and for testing.

cli/src/common/boto3/cfn.py
```diff
@@ -20,8 +20,18 @@ def __init__(self):
         super().__init__("cloudformation")
 
     @AWSExceptionHandler.handle_client_exception
-    def create_stack(self, stack_name: str, template_url: str, disable_rollback: bool, tags: list):
+    def create_stack(
+        self, stack_name: str, disable_rollback: bool, tags: list, template_url: str = None, template_body: str = None
+    ):
         """Create CFN stack by using the given template."""
+        if template_url is None:
```
enrico-usai, line 27: let's add a check to be sure that at least one from template_body and template_url is set. or we can create two different utilities:
* create_stack_from_url
* create_stack

cli/src/common/imagebuilder_utils.py
```diff
@@ -20,6 +20,32 @@
 PCLUSTER_RESERVED_VOLUME_SIZE = 15
 AMI_NAME_REQUIRED_SUBSTRING = " {{ imagebuilder:buildDate }}"
 InstanceRole = Enum("InstanceRole", ("ROLE", "INSTANCE_PROFILE"))
+ImageBuilderStatusMapping = {
```
enrico-usai, line 23: Why do you need them in the utils?
Let's avoid the "utils" module if possible, I'm also trying to cleanup our old utility module by moving the methods in the right layer/module.

If they are constant lets move to imagebuilder_constants module.
if they are just used by the model we can have this definition there.

cli/src/pcluster/cli.py
```diff
@@ -331,68 +336,21 @@ def _get_parser():
...
-        help="Specifies the base AMI to use for building the AWS ParallelCluster AMI.",
-    )
-    pami.add_argument(
-        "-os",
-        "--os",
-        dest="base_ami_os",
-        required=True,
-        help="Specifies the OS of the base AMI. "
-        "Valid options are: alinux, ubuntu1604, ubuntu1804, centos7, centos8.",
+        "build-image", help="(Linux/macOS) Creates a custom AMI to use with AWS ParallelCluster."
```
enrico-usai, line 339: I think now it can work for Windows too, because we removed the execution of the bash script.

cli/src/pcluster/cli.py
```diff
@@ -82,6 +82,11 @@ def stop(args):
     pcluster_stop.stop(args)
 
 
+def build_image(args):
+    pcluster.build_image(args)
+
+
+# moved
```
enrico-usai, line 89: Feel free to delete this code if useless

cli/src/pcluster/cli.py
```diff
@@ -331,68 +336,21 @@ def _get_parser():
...
-    pami.add_argument(
-        "--no-public-ip",
-        dest="associate_public_ip",
-        action="store_false",
-        default=True,
-        help="Do not associate public IP to the Packer instance. Defaults to associate public ip",
+        "--image-name",
+        dest="image_name",
+        required=True,
+        help="Specifies the image stack name to use for building the AWS ParallelCluster AMI.",
```
enrico-usai, line 346: image stack name?  do you mean the stack name? or is it the image name?

cli/src/pcluster/commands.py
```diff
@@ -544,3 +544,16 @@ def _get_default_template_url(region):
             REGION=region, SUFFIX=".cn" if region.startswith("cn") else "", VERSION=utils.get_installed_version()
         )
     )
+
+
+def build_image(args):
+    """Build AWS ParallelCluster AMI."""
+    LOGGER.info("Building AWS ParallelCluster AMI. This could take a while...")
+    response = PclusterApi().build_image(
+        imagebuilder_config=load_yaml_dict(args.config_file),
```
enrico-usai, line 553: let's add a try catch because the load_yaml_dict can fail.
In the clsuter create we are catching it in the _parse_config_file function.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+        self.validation_failures = validation_failures or []
+
+
+class ImageBuilderStack(Stack):
+    """Class representing a running stack associated to a building image."""
+
+    def __init__(self, name: str):
+        """Init stack info."""
+        super().__init__(name)
+        try:
```
enrico-usai, line 44: Please take a look at https://github.com/aws/aws-parallelcluster/pull/2483

I changed this initialization, by moving the describe_stack in a property.
This helps when mocking the object and when istantiating it directly.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+            self._image_resource = AWSApi().instance().cfn.describe_stack_resouce(name, "ParallelClusterImage")
+        except AWSClientError:
+            self._image_resource = None
+
+    @property
+    def version(self):
+        """Return the version of ParallelCluster used to create the stack."""
+        return self._get_tag("pcluster_version")
```
enrico-usai, line 52: What about using the same tag name used by the Cluster?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+        for key, value in ImageBuilderStatusMapping.items():
+            if cfn_status in value:
+                return key
+        return None
+
+    @property
+    def image_id(self):
+        """Return output image id."""
+        if self._image_resource:
+            image_build_version_arn = self._image_resource.get("StackResourceDetail").get("PhysicalResourceId")
```
enrico-usai, line 67: I think you can create an ImageInfo AWS resources (take a look at InstanceInfo in the https://github.com/aws/aws-parallelcluster/pull/2483 ) that give you all the attribute you need.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+            try:
+                resource = AWSApi.instance().imagebuilder.get_image_resources(image_build_version_arn)
+            except AWSClientError:
+                return None
+            image_id = resource.get("image").get("outputResources").get("amis")[0].get("image")
```
enrico-usai, line 72: You can create an utility in the ImageBuilderClient: get_image_id
that performs all the steps: to retrieve the image from the resources.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+            return image_id
+        return None
+
+
+class ImageBuilderManager:
```
enrico-usai, line 77: I think you can rename it as ImageBuilder and rename the old ImageBuilder like ImageBuilderConfig , to keep the things aligned with the cluster one.

Let's talk about it, if we want to call them "managers" I can rename the cluster one.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,112 @@
...
+    """Represent a building image, composed by an ImageBuilder config and an ImageBuilderStack."""
+
+    def __init__(self, name: str, config: dict = None, stack: ImageBuilderStack = None):
+        self.name = name
+        self.stack = stack
+        self.template_body = None
+        self.config_version = None
+        self.config = ImageBuilderSchema().load(config)
```
enrico-usai, line 85: Take a look at the already mentioned patch.
I'm saving __source_config and I have self.config as a property, in this way we don't need to load in the initializator, but only when/if using the config.

In this way the loading and validation of syntactic will be done at the same time of the self.config.validate() call, because it first load the config validationg semantically then calls our model validators.

cli/src/pcluster/models/imagebuilder_config.py
```diff
@@ -0,0 +1,164 @@
...
+        self.update_os_and_reboot = Resource.init_param(update_os_and_reboot, default=False)
+        self.disable_pcluster_component = Resource.init_param(disable_pcluster_component, default=False)
+        self.distribution_configuration = distribution_configuration
+        self.terminate_instance_on_failure = Resource.init_param(terminate_instance_on_failure, default=True)
+
+
+# ---------------------- ImageBuilder ---------------------- #
+
+
+class ImageBuilder(Resource):
```
enrico-usai, line 138: This can be rneamed to ImageBuilderConfig

enrico-usai (commented):
Great job!

cli/src/api/pcluster_api.py
yuleiwan: yes, after thinking a while, here I still pass the imagebuilder_manager object, because I decide to move imagebuild_status to manager rather than stack, imagebuild_status should be imagebuilder_manager's attribute.

cli/src/common/boto3/cfn.py
yuleiwan: Done, create two funcs.

cli/src/common/imagebuilder_utils.py
yuleiwan: Moved mapping to imagebuilder module, should we drop utils? Is there any benefit for doing that?

cli/src/pcluster/cli.py
yuleiwan: sorry, I miss something, which bash script you mentioned here?

cli/src/pcluster/cli.py
yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: sorry, `get_image_id` function is there, I should use it rather than `get_image_resources`

cli/src/pcluster/models/imagebuilder.py
yuleiwan: I struggle with the name. I feel we have two options:
1. use name xxxmanager and rename module name as well to manage the controllers and keep resources name as current name.
2. change all the resources name to xxxConfig, use imagebuilder/cluster here.
For now, we just change the root resource name to xxxConfig, I think it's a little bit confusing.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: good idea, followed

cli/src/pcluster/cli.py
yuleiwan: it is stack name, rename it to --name

cli/src/pcluster/models/imagebuilder.py
yuleiwan: I need to discuss with @lukeseawalker, because all the tags we use in ami start with `pcluster_` prefix.

cli/src/pcluster/cli.py
enrico-usai: I mean, the previous create-ami command didn't work on windows because it executed the build-ami.sh, the new build-ami can work on windows, so we can remove the `(Linux/macOS)` from the help.

cli/src/pcluster/models/imagebuilder.py
enrico-usai: Good point.
I think we should rename all the resources by adding the Config suffix. 
They don't represent real/live objects but just the configuration from the file.

cli/src/common/imagebuilder_utils.py
enrico-usai: Yes, I think we should get rid of the utils module or try to move most of the things in the appropriate module.
See some interesting discussions about this:
https://stackoverflow.com/questions/3340032/utility-classes-are-evil
https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

cli/src/pcluster/commands.py
```diff
@@ -544,3 +544,23 @@ def _get_default_template_url(region):
             REGION=region, SUFFIX=".cn" if region.startswith("cn") else "", VERSION=utils.get_installed_version()
         )
     )
+
+
+def build_image(args):
+    """Build AWS ParallelCluster AMI."""
+    LOGGER.info("Building AWS ParallelCluster AMI. This could take a while...")
+    response = {}
```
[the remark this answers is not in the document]
enrico-usai: minor: you can use fstring,

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,172 @@
...
+            except (AWSClientError, KeyError):
+                return None
+        return None
+
+    @property
+    def get_source_config(self):
+        """Get source config from metadata."""
+        try:
+            source_config = AWSApi.instance().cfn.get_template(self.name).get("Metadata").get("Config")
+            return source_config
```
enrico-usai, line 98: minor: you can avoid defining the source_config

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -12,160 +12,172 @@
...
+        """Return the version of ParallelCluster used to create the stack."""
+        return self._get_tag("pcluster_version")
+
+    @property
+    def image_id(self):
+        """Return output image id."""
+        if self._image_resource:
+            try:
+                image_build_version_arn = self._image_resource.get("StackResourceDetail").get("PhysicalResourceId")
+                image_id = AWSApi.instance().imagebuilder.get_image_id(image_build_version_arn)
```
enrico-usai, line 87: minor: you can avoid defining the image_id

enrico-usai (approved):
I just left some minor comments, but the patch is good for me.

cli/src/api/pcluster_api.py
```diff
@@ -111,3 +127,25 @@ def update_cluster(cluster_name: str, region: str):
                 "version used to create the cluster."
             )
         return ClusterInfo(cluster)
+
+    @staticmethod
+    def build_image(imagebuilder_config: dict, name: str, region: str = None, disable_rollback: bool = False):
```
lukeseawalker, line 132: `disable_rollback` default is `True`

cli/src/pcluster/cli.py
```diff
@@ -331,68 +330,21 @@ def _get_parser():
...
-        "--cluster-template",
-        help="Specifies the 'cluster' section of the configuration file to retrieve VPC settings.",
-    )
-    pami_group2 = pami.add_argument_group("Build AMI in a custom VPC and Subnet")
-    pami_group2.add_argument("--vpc-id", help="Specifies the VPC to use to build the AWS ParallelCluster AMI.")
-    pami_group2.add_argument("--subnet-id", help="Specifies the Subnet to use to build the AWS ParallelCluster AMI.")
     _addarg_region(pami)
-    pami.set_defaults(template_url=None)
-    pami.set_defaults(func=create_ami)
+    pami.add_argument(
```
lukeseawalker, line 344: rollback should be disabled by default, let's remove this for now.

cli/src/pcluster/commands.py
```diff
@@ -544,3 +544,23 @@ def _get_default_template_url(region):
...
+    try:
+        response = PclusterApi().build_image(
+            imagebuilder_config=load_yaml_dict(args.config_file),
+            image_name=args.name,
```
lukeseawalker, line 556: `build_image` parameter takes `name` and not `image_name`

cli/src/pcluster/commands.py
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/api/pcluster_api.py
yuleiwan: done

lukeseawalker (approved):

PR #2485, opened 2021-02-26
Restore instance type sanity check using cluster ami

ddeidda:
This commit restores the fix for instance type sanity checks using the actual cluster AMI, including new logic to handle the case where the AMIs are still under development.

### DIFF
--- a/cli/src/pcluster/cluster_model.py
+++ b/cli/src/pcluster/cluster_model.py
@@ -16,8 +16,9 @@ import boto3
 from botocore.exceptions import ClientError
 
 from pcluster.utils import (
-    error,
+    Cache,
     get_availability_zone_of_subnet,
+    get_installed_version,
     get_supported_az_for_one_instance_type,
     is_hit_enabled_cluster,
 )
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/cluster_model.py
```diff
@@ -115,17 +117,82 @@ def _ec2_run_instance(self, pcluster_config, **kwargs):  # noqa: C901 FIXME!!!
...
+        architecture = pcluster_config.get_section("cluster").get_param_value("architecture")
+        custom_ami = pcluster_config.get_section("cluster").get_param_value("custom_ami")
+        try:
+            cluster_ami = custom_ami if custom_ami else self._get_official_image_id(os, architecture)
+        except Exception as e:
+            pcluster_config.error("Error when resolving cluster ami id: {0}".format(e))
+
+        return cluster_ami
+
+    def _get_latest_alinux_ami_id(self):
```
demartinofra, line 185: shall we get rid of this unused function

ddeidda: :+1:

demartinofra (approved):

PR #2493, opened 2021-03-01
Backport: Only require account ID when git refs are used

rexcsn:
When --json-regions and --json-template are used (and the git ref args
are not) the account ID argument is unused.


tilne (approved):

rexcsn:
Ignoring py3.4 CI

PR #2501, opened 2021-03-02
Add validation for s3 bucket to be in same region of cluster stack

chenwany:
Add a validator to validate that `cluster_resource_bucket` is in the same region with the cluster stack.
Tests:
When specify `cluster_resource_bucket` in a different region with where the cluster create, the validator will generate the following error:
ERROR: The configuration parameter 'cluster_resource_bucket' generated the following errors:
cluster_resource_bucket must be in the same region of the cluster.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG
 ------
 **ENHANCEMENTS**
 
+- Add validation to prevent using a `cluster_resource_bucket` that is in a different region than the cluster.
+
 **CHANGES**
 
 - Upgrade Slurm to version 20.11.4.
--- a/cli/src/pcluster/config/validators.py
+++ b/cli/src/pcluster/config/validators.py
@@ -797,6 +797,26 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
     return errors, warnings
 
 
+def s3_bucket_region_validator(param_key, param_value, pcluster_config):
+    """Validate S3 bucket is in the same region with the cloudformation stack."""
+    errors = []
+    warnings = []
+    s3_client = boto3.client("s3")
+    try:
+        bucket_region = s3_client.get_bucket_location(Bucket=param_value).get("LocationConstraint")
+        # Buckets in Region us-east-1 have a LocationConstraint of null
+        # Example output from get_bucket_location for us-east-1:
+        #   {'ResponseMetadata': {...}, 'LocationConstraint': None}
+        if bucket_region is None:
+            bucket_region = "us-east-1"
+        if bucket_region != pcluster_config.region:
+            errors.append("cluster_resource_bucket must be in the same region of the cluster.")
+    except ClientError as client_error:
+        _process_generic_s3_bucket_error(client_error, param_value, warnings, errors)
+
+    return errors, warnings
+
+
 def _process_generic_s3_bucket_error(client_error, bucket_name, warnings, errors):
     if client_error.response.get("Error").get("Code") == "NoSuchBucket":
         errors.append(
@@ -812,7 +832,7 @@ def _process_generic_s3_bucket_error(client_error, bucket_name, warnings, errors
         )
     else:
         errors.append(
-            "Unexpected error when calling get_bucket_location on S3 bucket '{0}': '{1}'".format(
+            "Unexpected error for S3 bucket '{0}': '{1}'".format(
                 bucket_name, client_error.response.get("Error").get("Message")
             )
         )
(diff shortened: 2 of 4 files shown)


cli/src/pcluster/config/validators.py
```diff
@@ -769,6 +769,41 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+    """Validate S3 bucket  is in the same region with the cloudformation stack."""
+    errors = []
+    warnings = []
+    s3_client = boto3.client("s3")
+    try:
+        bucket_region = s3_client.get_bucket_location(Bucket=param_value).get("LocationConstraint")
+        # Buckets in Region us-east-1 have a LocationConstraint of null
+        if bucket_region is None:
+            bucket_region = "us-east-1"
+        if bucket_region != pcluster_config.region:
```
rexcsn, line 782: We need to consider the case that the region is provided through the CLI, i.e. `pcluster create <some_cluster> -r us-east-1`. We might need to validate this in `commands.py` or somehow pass in the region argument

cli/tests/pcluster/config/test_validators.py
```diff
@@ -545,6 +546,69 @@ def test_s3_validator(mocker, boto3_stubber, config, num_calls, bucket, expected
...
+            "AccessDenied",
+            "You do not have access to the S3 bucket 'access_denied_bucket'",
+        ),
+    ],
+)
+def test_s3_bucket_region_validator(mocker, boto3_stubber, error_code, bucket, expected_message):
+    os.environ["AWS_DEFAULT_REGION"] = "eu-west-1"
+    get_bucket_location_response = {
+        "ResponseMetadata": {},
+        "LocationConstraint": "af-south1",
```
rexcsn, line 573: `us-east-1` is an edge case, so we should add a test case for that.
We also need to add a test case where the validation passes for full coverage.

cli/tests/pcluster/config/test_validators.py
```diff
@@ -545,6 +546,69 @@ def test_s3_validator(mocker, boto3_stubber, config, num_calls, bucket, expected
...
+        ),
+        (
+            "access_denied_bucket",
+            "AccessDenied",
+            "You do not have access to the S3 bucket 'access_denied_bucket'",
+        ),
+    ],
+)
+def test_s3_bucket_region_validator(mocker, boto3_stubber, error_code, bucket, expected_message):
+    os.environ["AWS_DEFAULT_REGION"] = "eu-west-1"
```
rexcsn, line 570: While this is ok, would be nice if we can test out the different cases when region is provided with CLI vs. provided in original config.
If that logic is already tested, you can also leave a comment in the code to point to an existing test case here

cli/src/pcluster/config/validators.py
chenwany: The case `-r us-east-1` is already considered here, since the pcluster_config in the validator stage already convert all the input arguments into the config file.

cli/src/pcluster/config/validators.py
```diff
@@ -769,6 +769,41 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
     return errors, warnings
 
 
+def s3_bucket_region_validator(param_key, param_value, pcluster_config):
+    """Validate S3 bucket  is in the same region with the cloudformation stack."""
```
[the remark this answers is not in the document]
chenwany: Yes, fixed!

cli/src/pcluster/config/validators.py
chenwany: The case -r us-east-1 is already considered here, since the pcluster_config in the validator stage already convert all the input arguments into the config file.

cli/tests/pcluster/config/test_validators.py
chenwany: Done! Here we provide the region by using os.environ["AWS_DEFAULT_REGION"], it should be passed to the config, although we use it in many places, it seems that we don't have test for this, we may need to add this test later

cli/tests/pcluster/config/test_validators.py
chenwany: Added. Also added explainations in the comments of the code

cli/src/pcluster/config/validators.py
```diff
@@ -797,6 +797,43 @@ def s3_bucket_validator(param_key, param_value, pcluster_config):
...
+                "You do not have access to the S3 bucket '{0}': '{1}'".format(
+                    param_value, client_error.response.get("Error").get("Message")
+                )
+            )
+        else:
+            errors.append(
+                "Unexpected error when calling get_bucket_location on S3 bucket '{0}': '{1}'".format(
+                    param_value, client_error.response.get("Error").get("Message")
+                )
+            )
```
rexcsn, line 832: You can probably call `_process_generic_s3_bucket_error` to avoid code duplication.
We will need to fix the error message `Unexpected error when calling get_bucket_location on S3 bucket` in `_process_generic_s3_bucket_error`, because it should be a generic error.

nit: we can also fix similar code duplications in `fsx_lustre_auto_import_validator
` below

cli/tests/pcluster/config/test_validators.py
```diff
@@ -546,6 +547,98 @@ def test_s3_validator(mocker, boto3_stubber, config, num_calls, bucket, expected
     utils.assert_param_validator(mocker, config, expected_message)
 
 
+@pytest.mark.parametrize(
+    "bucket, region, error_code, expected_message, client_error",
+    [
+        (
+            "bucket",
+            "us-east-1",
```
rexcsn, line 555: I would add another test case for a region that passes the validator and is not `us-east-1`, since `us-east-1` is a special case, we also want to test that other "normal" region can pass the validator.

cli/src/pcluster/config/validators.py
chenwany: Done! Aslo fixed that for `fsx_lustre_auto_import_validator`

cli/tests/pcluster/config/test_validators.py
chenwany: Added

rexcsn (dismissed):

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG
 ------
 **ENHANCEMENTS**
 
+- Add validation to prevent `cluster_resource_bucket` creating not in same region as cluster.
```
rexcsn, line 8: How about "Add validation to prevent using a `cluster_resource_bucket` that is in a different region than the cluster"

chenwany: Good idea, done!

rexcsn (approved):

PR #2503, opened 2021-03-03
3.0: Implement head node creation

enrico-usai:
###  Model
* Add CustomActionEvent to distinguish between pre and post install custom actions
  and use it for both schema and config.
* Defined disable_simultaneous_multithreading_via_cpu_options property
* Defined get_custom_action in the HeadNode to be able to retrieve pre/post install actions

###  Stack
* Defined _storage_resource_options to store shared storage options for Raid, EFS and FSx.
  They are required by the cookbook.
  TODO: we should save in the cookbook only the required options.
* Defined head_node_iam_role that will be created if head-node instance role is not specified
  TODO: we should use different roles for the queues.
* I defined CloudFormation::Init as a raw metadata because CfnInstance doesn't have "init" attribute,
  Instance construct has the "init" attribute to specify a CloudFormationInit object but it doesn't support
  launch template and we need it for cpu_options and network interfaces.
* Move head node creation as latest, because we need storage information.
* Removed Ganglia outputs and added head node  private ip/dns and public ip

### Tests
* The patch permits to create the head node but the recipes execution is failing for multiple reasons.
1. It fails if there are no EBS shared storages specified, because in cookbook code it is a requirement.
   We need to adapt the cookbook to avoid accepting a shared folder by default, but we can do it later,
   for now we can add a SharedStorage in the config file.
2. It's failing when trying to insert same data in the DynamoDB table because this part is still missing.

## Define OS_MAPPING in constants module

This mapping object can be reused to store all the information
related to a specific os (e.g. supported features, architectures, etc)

## Remove image from HeadNode and Queue

It's not supported, this can only be defined at root level.

### DIFF
--- a/cli/src/pcluster/models/cluster_config.py
+++ b/cli/src/pcluster/models/cluster_config.py
@@ -30,8 +30,10 @@ from pcluster.constants import (
 )
 from pcluster.models.common import BaseDevSettings, BaseTag, Resource
 from pcluster.utils import (
+    InstanceTypeInfo,
     delete_s3_artifacts,
     delete_s3_bucket,
+    disable_ht_via_cpu_options,
     get_availability_zone_of_subnet,
     get_partition,
     get_region,
@@ -120,7 +122,6 @@ class Ebs(Resource):
         self.throughput = Resource.init_param(throughput, default=125 if self.volume_type == "gp3" else None)
 
     def _validate(self):
-        # FIXME This method is not executed because subclass override the method.
         self._execute_validator(EbsVolumeTypeSizeValidator, volume_type=self.volume_type, volume_size=self.size)
         self._execute_validator(
             EbsVolumeIopsValidator,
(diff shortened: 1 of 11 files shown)


cli/src/pcluster/models/cluster_config.py
```diff
@@ -695,15 +701,39 @@ def vcpus(self):
...
+    def max_network_interface_count(self) -> int:
+        """Return max number of NICs for the instance."""
+        return self._instance_type_info.max_network_interface_count()
+
+    @property
+    def _instance_type_info(self) -> InstanceTypeInfo:
+        """Return head node instance type information as returned from aws ec2 describe-instance-types."""
+        if not self.__instance_type_info:
+            self.__instance_type_info = AWSApi.instance().ec2.get_instance_type_info(self.instance_type)
+        return self.__instance_type_info
```
hanwen-pcluste, line 723: Why store the info in an internal attribute. Shall we do all the caching in AWSApi layer?

hanwen-pcluste (approved):
LGTM.
I didn't look very closely in cluster_stack.py

cli/src/pcluster/models/cluster_config.py
enrico-usai: Good point, yes I know it's cached on aws api layer but given that I'm using this a lot of times for a lot of properties I want to be sure we're caching this value even if we remove the cache from aws api layer.

PR #2506, opened 2021-03-03
3.0: Fix PolicyDocument syntax errors

hanwen-pcluste:
1.There is an inconsistency between CDK and IAM. `condition` in IAM is called `conditions`(plural form) in CDK. However the `conditions` only accepts a dict instead of a list. This commit will pass dict to `conditions`.
2. `aws_cdk.core.Stack.format_arn()` always has the format `arn:{partition}:{service}:{region}:{account}:{resource}{sep}}{resource-name}`. However, for some services (s3, iam), `region` and/or `account` are not needed. We can bypass the issue by setting `region=""` and/or `account=""` or assembling arn by ourselves.

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -410,10 +410,12 @@ class ClusterCdkStack(core.Stack):
                                 actions=s3_policy_actions,
                                 effect=Effect.ALLOW,
                                 resources=[
-                                    self.format_arn(service="s3", resource=self._bucket.name),
+                                    self.format_arn(service="s3", resource=self._bucket.name, region="", account=""),
                                     self.format_arn(
                                         service="s3",
                                         resource="{0}/{1}/*".format(self._bucket.name, self._bucket.artifact_directory),
+                                        region="",
+                                        account="",
                                     ),
                                 ],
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -410,11 +410,8 @@ def _add_iam_lambda_role(self):
                                 actions=s3_policy_actions,
                                 effect=Effect.ALLOW,
                                 resources=[
-                                    self.format_arn(service="s3", resource=self._bucket.name),
-                                    self.format_arn(
-                                        service="s3",
-                                        resource="{0}/{1}/*".format(self._bucket.name, self._bucket.artifact_directory),
-                                    ),
+                                    f"arn:aws:s3:::{self._bucket.name}",
```
enrico-usai, line 413: The "aws" partition must be changed to "aws-cn" when we're in china. I'd keep format_arn because it automatically manages it.

enrico-usai (commented):
Thanks for fixing this syntax error.
Let's restore format_arn.

cli/src/pcluster/templates/cluster_stack.py

enrico-usai (approved):
LGTM

PR #2507, opened 2021-03-04
3.0: Pcluster delete image and describe image command

yuleiwan:
# Use the new model for delete-image, describe-image command

## CLI
* Add delete-image command
* Add `--name`, `--region`, `--force` to `delete-image` command
* Add describe-image command
* Add `--name`, `--region` to `describe-image` command

## AWSResource
* Add ImageInfo represents the information of a live image. It will be initialized in describe images boto3 call.

## PclusterApi
* Add delete_image and describe_image API methods

## Imagebuilder
* Add image property to imagebuilder to store the result of describe images call
* Add delete function to imagebuilder

### describe-image
building:
command
```
pcluster describe-image -n pcluster-test -r us-east-1
```
output
```
Response:
{'image': '{"stack_exist": true, "image_exist": false, "region": "us-east-1", "image_configurtion": "https://parallelcluster-378a6715c688c0ce-v1-do-not-delete.s3.us-east-1.amazonaws.com/parallelcluster/images/pcluster-test-9l60rt41g4dzkbty/configs/image-config.yaml", "stack_name": "pcluster-test", "stack_status": "CREATE_IN_PROGRESS", "stack_arn": "arn:aws:cloudformation:us-east-1:xxxxxxxxxxxx:stack/pcluster-test/7b596b20-9c70-11eb-ba32-0e1c3468127d", "tags": [{"Key": "parallelcluster:version", "Value": "2.10.3"}, {"Key": "parallelcluster:image_name", "Value": "pcluster-test"}, {"Key": "parallelcluster:s3_bucket", "Value": "parallelcluster-378a6715c688c0ce-v1-do-not-delete"}, {"Key": "parallelcluster:s3_image_dir", "Value": "parallelcluster/images/pcluster-test-9l60rt41g4dzkbty"}], "version": "2.10.3", "creation_time": "2021-04-13 15:54:08.788000+00:00", "imagebuild_status": "BUILD_IN_PROGRESS", "build_log": "arn:aws:logs:us-east-1:xxxxxxxxxxxx:log-group:/aws/imagebuilder/ParallelClusterImage-pcluster-test"}'}
```
build complete:
command
```
pcluster describe-image -n pcluster-test-2 -r us-east-1
```
output
```
Response:
{'image': '{"stack_exist": false, "image_exist": true, "region": "us-east-1", "image_configurtion": "https://parallelcluster-378a6715c688c0ce-v1-do-not-delete.s3.us-east-1.amazonaws.com/parallelcluster/images/test-pcluster-2-pa1ptkn02txtw0n9/configs/image-config.yaml", "image_name": "test-pcluster-2 2021-04-13T04-50-20.756Z", "image_id": "ami-02ad439c060fc9a2b", "image_state": "available", "image_architecture": "x86_64", "image_tags": [{"Key": "parallelcluster:dcv_server", "Value": "nice-dcv-server-2020.2.9662-1.el7.x86_64"}, {"Key": "parallelcluster:munge", "Value": "munge-0.5.14"}, {"Key": "parallelcluster:dcv_xdcv", "Value": "nice-xdcv-2020.2.359-1.el7.x86_64"}, {"Key": "parallelcluster:efa_profile", "Value": "efa-profile-1.4-1.amzn2.noarch"}, {"Key": "parallelcluster:sudo", "Value": "sudo-1.8.23-10.amzn2.1.x86_64"}, {"Key": "parallelcluster:kernel", "Value": "4.14.225-169.362.amzn2.x86_64"}, {"Key": "Ec2ImageBuilderArn", "Value": "arn:aws:imagebuilder:us-east-1:xxxxxxxxxxxx:image/parallelclusterimage-652c5d10-9c13-11eb-9dbb-12936a84a87d/2.10.3/1"}, {"Key": "parallelcluster:image_name", "Value": "test-pcluster-2"}, {"Key": "parallelcluster:version", "Value": "2.10.3"}, {"Key": "parallelcluster:os", "Value": "alinux2"}, {"Key": "parallelcluster:efa", "Value": "efa-1.10.2-1.amzn2.x86_64"}, {"Key": "parallelcluster:slurm", "Value": "slurm-20.02.4"}, {"Key": "parallelcluster:efa_config", "Value": "efa-config-1.7-1.amzn2.noarch"}, {"Key": "parallelcluster:efa_rdma_core", "Value": "rdma-core-31.2amzn-1.amzn2.x86_64"}, {"Key": "parallelcluster:lustre", "Value": "lustre-client-2.10.8-5.amzn2.x86_64"}, {"Key": "parallelcluster:bootstrap_file", "Value": "aws-parallelcluster-cookbook-2.10.3"}, {"Key": "parallelcluster:efa_openmpi40_aws", "Value": "openmpi40-aws-4.1.0-2.amzn2.x86_64"}, {"Key": "parallelcluster:s3_image_dir", "Value": "parallelcluster/images/alinux2-3-pa1ptkn02txtw0n9"}, {"Key": "parallelcluster:pmix", "Value": "pmix-3.1.5"}, {"Key": "parallelcluster:s3_bucket", "Value": "parallelcluster-378a6715c688c0ce-v1-do-not-delete"}, {"Key": "CreatedBy", "Value": "EC2 Image Builder"}], "imagebuild_status": "BUILD_COMPLETE", "creation_time": "2021-04-13T05:24:17.000Z", "build_log": "arn:aws:logs:us-east-1:xxxxxxxxxxxx9:log-group:/aws/imagebuilder/ParallelClusterImage-pcluster-test"}'}
```
### delete-image
command
```
pcluster delete-image -n pcluster-test -r us-east-1
```
output
```
Deleting: pcluster-test
Image deletion started correctly. {"stack_name": "pcluster-test", "stack_status": "DELETE_IN_PROGRESS", "stack_arn": "arn:aws:cloudformation:us-east-1:xxxxxxxxxxxx:stack/pcluster-test/8e2cce70-6e8c-11eb-8b84-0aedc5aecc23", "region": "us-east-1", "version": 2.10.1, "imagebuild_status": "DELETE_IN_PROGRESS"}
Status: PClusterComponent - DELETE_COMPLETE
Image deleted successfully.
``` 

*Issue #, if available:*

*Description of changes:*


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ CHANGELOG
   `--ami-name-prefix`, `--custom-cookbook`, `--post-install`, `--no-public-ip`, `--cluster-template`, `--vpc-id`,
    `--subnet-id`.
 - Add `--image-name`, `--config`, `--region` to `build-image` command
+- Add delete-image command
```
enrico-usai, line 17: minor: Add `delete-image` command, with --image-name, --region, --nowait
 parameters.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ CHANGELOG
...
+- Add --image-name, --region, --nowait to deletd-image command
```
enrico-usai, line 18: typo

cli/src/api/pcluster_api.py
```diff
@@ -74,13 +74,24 @@ class ImageInfo:
     """Minimal representation of a building image."""
 
     def __init__(self, imagebuilder: ImageBuilder):
-        self.image_name = imagebuilder.stack.name
-        self.imagebuild_status = imagebuilder.imagebuild_status
+        # imagebuilder stack info
```
enrico-usai, line 77: I'd rename this class from ImageInfo to PclusterImageInfo to avoid name overlapping with ImageInfo in aws_resources.

Or ImageBuilderInfo given that you're returing both info about the ami and the image-builder stack

cli/src/common/boto3/cfn.py
```diff
@@ -89,4 +89,4 @@ def list_pcluster_stacks(self):
     @AWSExceptionHandler.handle_client_exception
     def describe_stack_resource(self, stack_name: str, logic_resource_id: str):
         """Get stack resource information."""
-        self._client.describe_stack_resource(StackName=stack_name, LogicalResourceId=logic_resource_id)
+        return self._client.describe_stack_resource(StackName=stack_name, LogicalResourceId=logic_resource_id)
```
enrico-usai, line 92: good catch, thanks!

cli/src/pcluster/cli.py
```diff
@@ -308,6 +316,28 @@ def _get_parser():
     _addarg_region(pami)
     pami.set_defaults(func=build_image)
 
+    # delete image command subparser
+    pdeleteami = subparsers.add_parser(
+        "delete-image",
+        help="Deletes an image.",
```
enrico-usai, line 322: "Deletes an image and related image builder stack".?

cli/src/pcluster/cli.py
```diff
@@ -308,6 +316,28 @@ def _get_parser():
...
+    )
+    pdeleteami.add_argument(
+        "-i", "--image-name", dest="image_name", required=True, help="Name of the AWS ParallelCluster AMI to delete."
+    )
+    _addarg_region(pdeleteami)
+    _addarg_nowait(pdeleteami)
+    pdeleteami.set_defaults(func=delete_image)
+
+    # describe image command subparser
+    pdescribeami = subparsers.add_parser("describe-image", help="Lists an AWS ParallelCluster AMI info.")
```
enrico-usai, line 334: help="Describes the specified ParallelCluster image."

cli/src/pcluster/commands.py
```diff
@@ -506,6 +506,14 @@ def delete(args):
                     ).ljust(80)
                     sys.stdout.write("\r%s" % resource_status)
                     sys.stdout.flush()
+                elif isinstance(result, ApiFailure):
+                    # If stack is already deleted
+                    if f"Cluster {args.cluster_name} does not exist." in result.message:
+                        LOGGER.info("\nCluster deleted successfully.")
```
enrico-usai, line 512: Let's do `LOGGER.warning(f"\nCluster {args.cluster_name}
 has already been deleted or does not exist.")`

cli/src/pcluster/commands.py
```diff
@@ -517,6 +525,13 @@ def delete(args):
             sys.stdout.flush()
         if result.stack_status == "DELETE_FAILED":
             LOGGER.info("Cluster did not delete successfully. Run 'pcluster delete %s' again", args.cluster_name)
+    except ClientError as e:
+        if e.response.get("Error").get("Message").endswith("does not exist"):
+            LOGGER.info("\nCluster deleted successfully.")
```
enrico-usai, line 530: same here.

cli/src/pcluster/commands.py
```diff
@@ -569,9 +584,83 @@ def build_image(args):
...
+                    events = utils.get_stack_events(result.stack_name, raise_on_error=True)[0]
+                    resource_status = (
+                        "Status: %s - %s" % (events.get("LogicalResourceId"), events.get("ResourceStatus"))
+                    ).ljust(80)
+                    sys.stdout.write("\r%s" % resource_status)
+                    sys.stdout.flush()
+                elif isinstance(result, ApiFailure):
+                    # If stack is already deleted
+                    if f"ImageBuilder stack {args.image_name} does not exist." in result.message:
+                        LOGGER.info("\nImage deleted successfully.")
```
enrico-usai, line 628: LOGGER.warning(f"\nImage {args.image_name} has already been deleted or does not exist.")

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -80,6 +80,28 @@ def version(self):
...
         return self._get_tag("pcluster_version")
 
+    @property
+    def image(self):
+        """Return created image by imagebuilder stack."""
+        try:
+            image_id = self.image_id
+            if image_id:
+                image_info = AWSApi.instance().ec2.describe_image(image_id)
+                return ImageInfo(
```
enrico-usai, line 90: You don't need to pass a restricted dictionary to ImageInfo, you can pass the full one, it will be stored in _image_data and then you already have the properties in ImageInfo to return the right attributees.

The goal of having ImageInfo class is that you don't need to do get("Name"), get("Description") etc everywhere.

You can even think to modify the describe_image method in ec2_client to directly return an ImageInfo object.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -99,6 +121,10 @@ def get_source_config(self):
         except (AWSClientError, KeyError):
             return None
 
+    def update_stack(self):
```
enrico-usai, line 124: update_stack_info?

because update_stack let me think that we're updating the stack in cfn.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -187,6 +213,24 @@ def create(self, disable_rollback: bool = False):
             LOGGER.critical(e)
             raise ImageBuilderActionError(f"ImageBuilder stack creation failed.\n{e}")
 
+    def delete(self):
+        """Delete CFN Stack and associate resources and deregister the image."""
+        # validate image name
+        self._validate_image_name()
```
enrico-usai, line 219: not clear to me why we need to validate the image name at deletion time.
If it's an important step we should do it in the api or in the creation of the image object.

We might add validators for the syntax of the parameters ar first step of the api.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -187,6 +213,24 @@ def create(self, disable_rollback: bool = False):
...
+
+        # check imagebuilder stack existence
+        if not AWSApi.instance().cfn.stack_exists(self.image_name):
```
enrico-usai, line 222: I think you can directly call deregister_image then delete_stack by catching exceptions, rather than checking if the stack exist, it permits to save one boto3 call and have the same result.

CHANGELOG.md
yuleiwan: done

CHANGELOG.md
yuleiwan: fixed

cli/src/api/pcluster_api.py
yuleiwan: good point, rename to ImageBuilderInfo

cli/src/pcluster/cli.py
yuleiwan: Done

cli/src/pcluster/commands.py
yuleiwan: Done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: correct, I will modify the method `describe_image` in ec2_client later with a separate commit to make it clear

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Done.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: validate_image_name idea here is for the different regex patterns between image name and cfn stack name.
Although it is a corner case, for example, if a name is suitable for cfn stack name, without validation, we will pass this name to generate cdk template and create cfn stack. We will meet error during cfn stack creation due to invalid image name. But if we validate before cfn stack creation, we will avoid that failure.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: good point, done

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -112,47 +128,53 @@ def __init__(self, image_name: str, config: dict = None, stack: ImageBuilderStac
...
-                self.__source_config = self.bucket.get_config(config_name=self._s3_artifacts_dict.get("config_name"))
-            except Exception as e:
-                raise ImageBuilderActionError(
-                    "Unable to load configuration from bucket " f"'{self.bucket.name}/{self.s3_artifacts_dir}'.\n{e}"
-                )
-        return self.__source_config
+                self.__image = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
+            except AWSClientError as e:
+                if isinstance(e, ImageNotFoundError):
+                    raise ImageBuilderActionError(f"Image {self.image_name} does not exist.")
```
enrico-usai, line 174: I think you can use a different exception here: e.g. NonExistingImageError, given that you are catching it internally.

You can catch this new exception type in a different way, rather than looking at the error message.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -185,24 +207,23 @@ def bucket(self):
...
         """Try to get custom bucket name from image tag or stack tag."""
         custom_bucket_name = None
         try:
-            image_info = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
-            custom_bucket_name = search_tag(image_info, PCLUSTER_S3_BUCKET_TAG)
-        except AWSClientError as e:
-            if not isinstance(e, ImageNotFoundError):
+            custom_bucket_name = self.image.s3_bucket_name
+        except ImageBuilderActionError as e:
+            if f"Image {self.image_name} does not exist." not in str(e):
```
enrico-usai, line 220: Here you can catch NonExistingImageError

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -185,24 +207,23 @@ def bucket(self):
...
                 raise ImageBuilderActionError(f"Unable to get S3 bucket name from image {self.image_name} tag. {e}")
 
         if custom_bucket_name is None:
             try:
                 custom_bucket_name = self.stack.s3_bucket_name
-            except AWSClientError as e:
+            except ImageBuilderActionError as e:
```
enrico-usai, line 226: Same here, you can define a NonExistingStackError.

The basic idea is that the model should raise an ImageBuilderAction to the "api" layer for all the known issues and this should be the only known exception to be catched by the api layer.

But for internal errors/exception we can define new exceptions and manage them in a different way.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -213,29 +234,28 @@ def _get_custom_bucket(self):
...
-            if self.__s3_artifact_dir is None:
-                raise AWSClientError(
-                    function_name="_get_artifact_dir",
-                    message="No artifact directory found in image tag and cloudformation stack tag.",
+                if s3_artifact_dir is None:
+                    raise ImageBuilderActionError(
+                        "No artifact directory found in image tag and cloudformation stack tag."
+                    )
+            except ImageBuilderActionError as e:
+                raise ImageBuilderActionError(
```
enrico-usai, line 253: As you can see there is something strange here, you are raising ImageBuilderActionError in the 249 line then you're caching it here.
If you have NonExistingStackError you can distinguish the two error cases in a simpler way.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -285,9 +305,10 @@ def create(
 
         # Add tags information to the stack
         cfn_tags = copy.deepcopy(self.config.build.tags) or []
-        cfn_tags.append(BaseTag(key="pcluster_build_image", value=get_installed_version()))
+        cfn_tags.append(BaseTag(key="parallelcluster:version", value=get_installed_version()))
```
enrico-usai, line 308: PCLUSTER_VERSION_TAG

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -332,12 +354,9 @@ def _upload_config(self):
         """Upload source config to S3 bucket."""
         try:
             if self.config:
-                # Upload config with default values
-                self.bucket.upload_config(config=self.config, config_name=self._s3_artifacts_dict.get("config_name"))
-
                 # Upload original config
                 self.bucket.upload_config(
-                    config=self.config.source_config, config_name=self._s3_artifacts_dict.get("source_config_name")
+                    config=self.__source_config, config_name=self._s3_artifacts_dict.get("config_name")
```
enrico-usai, line 359: You shouldn't access private attributes, it was the reason why we had source_config property.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -368,6 +387,52 @@ def _upload_artifacts(self):
...
+
+                # Delete s3 image directory
+                try:
+                    self.bucket.check_bucket_exists()
+                    self.bucket.delete_s3_artifacts()
+                except AWSClientError:
+                    logging.warning("S3 bucket %s does not exist, skip image s3 artifacts deletion.", self.bucket.name)
+
+            except AWSClientError as e:
+                raise ImageBuilderActionError(f"Unable to delete image and stack, due to {str(e)}")
```
enrico-usai, line 417: Let's double think to error cases, e.g. you don't have permission to deregister_image but you have permissions to delete_snapshots or viceversa. Is it correct that we're failing at the first action, or should we proceed with the actions we can do?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -368,6 +387,52 @@ def _upload_artifacts(self):
...
+
+    def _check_instance_using_image(self):
```
enrico-usai, line 419: Could you add a comment to this method?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -368,6 +387,52 @@ def _upload_artifacts(self):
...
+                logging.error(
+                    "Image %s is used by instances %s. In case you want to delete the image, "
+                    "please use the --force flag.",
+                    self.image_name,
+                    str(result),
+                )
+                raise ImageBuilderActionError("Unable to delete image and stack")
+            return False
+        except (AWSClientError, ImageBuilderActionError) as e:
+            if f"Image {self.image_name} does not exist." in str(e):
```
enrico-usai, line 432: Same here: we're raising ImageBuilderActionError in 429 and we're re-catching it, let's use NonExistingImageError.

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -133,15 +133,16 @@ def _add_cfn_parameters(self):
...
         tags = copy.deepcopy(self.config.build.tags) or []
-        tags.append(BaseTag(key="pcluster_build_image", value=utils.get_installed_version()))
+        tags.append(BaseTag(key=PCLUSTER_IMAGE_NAME_TAG, value=self.image_name))
         build_tags_map = {tag.key: tag.value for tag in tags}
         build_tags_list = [core.CfnTag(key=tag.key, value=tag.value) for tag in tags]
 
         # Add default ami tags information
         tags = copy.deepcopy(self.config.image.tags) if self.config.image and self.config.image.tags else []
-        tags.append(BaseTag(key="pcluster_version", value=utils.get_installed_version()))
+        tags.append(BaseTag(key="parallelcluster:version", value=utils.get_installed_version()))
```
enrico-usai, line 142: PCLUSTER_VERSION_TAG

enrico-usai: Then I think we should create a function to add all these tags, rather than having this duplicate code to appending all the tags.

cli/src/api/pcluster_api.py
```diff
@@ -65,16 +65,38 @@ def __repr__(self):
...
-        self.stack_status = imagebuilder.stack.status
-        self.stack_arn = imagebuilder.stack.id
+        self.stack_exist = False
+        self.image_exist = False
         self.region = get_region()
-        self.version = imagebuilder.stack.version
+        # image config file url
+        self.image_configurtion = imagebuilder.config_url
+
+        if AWSApi.instance().cfn.stack_exists(imagebuilder.image_name):
```
enrico-usai, line 78: I'm not sure about this call in the ImageBuilderInfo, it seems to me that it's up of the caller, using ImageBuilderInfo to pass the known information.

You can do it by passing the stack to the ImageBuilder constructor:
```python
def __init__(self, imagebuilder: ImageBuilder, stack: StackInfo = None):
  ...
if stack:
    # imagebuilder stack info
    self.stack_exist = True
    ...

```

cli/src/common/boto3/ec2.py
```diff
@@ -113,6 +113,25 @@ def describe_image_by_name_tag(self, image_name):
...
+            for result in self._paginate_results(
+                self._client.describe_instances,
+                Filters=[
+                    {
+                        "Name": "image-id",
+                        "Values": [
+                            image_id,
+                        ],
+                    },
+                ],
```
enrico-usai, line 130: nit: you can save some lines by removing the ending commas, the auto-format should compact it in a single line.

cli/src/pcluster/commands.py
```diff
@@ -445,3 +446,99 @@ def update(args):
...
+        else:
+            utils.error(f"Image deletion failed. {result.message}")
+
+        if result.stack_exist:
+            sys.stdout.write("\rImageBuilderStackStatus: %s" % result.stack_status)
+            sys.stdout.flush()
+            LOGGER.debug("ImageBuilderStackStatus: %s", result.stack_status)
+
+            if not args.nowait:
+                while result.stack_status == "DELETE_IN_PROGRESS":
```
enrico-usai, line 469: Try to look at utils.verify_stack_status I think you can reuse it or do something similar to simplify this funciton and remove the `# noqa: C901`

cli/src/pcluster/commands.py
```diff
@@ -445,3 +446,99 @@ def update(args):
...
+
+        else:
+            sys.stdout.write("\rImageStatus: %s" % result.image_state)
+            sys.stdout.flush()
+            LOGGER.debug("ImageStatus: %s", result.image_state)
+
+            if not args.nowait:
+                image_name = result.image_name
+                while result.image_exist:
+                    time.sleep(0.1)
```
enrico-usai, line 508: Is 0.1 the right value?

cli/src/pcluster/commands.py
enrico-usai: Then I'm not sure we want to introduce this while/loop and nowait feature for image commands. I kept them for pcluster commands because they were already there and integ tests framework was relying on them but I think we can avoid this loop for these new commands.  
I think we'll remove the loop from the pcluster commands too.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Added ImageError and subclass NonExistingImageError to imagebuilder.py. For all the places calling image property, we will catch ImageError and then check if it is instance of NonExistingImageError or not. Therefore, calling image property won't raise ImageBuilderActionError, instead raise ImageBuilderActionError in the caller.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Added StackError and subclass NonExistingStackError to imagebuilder.py. For all the places calling stack property, we will catch StackError and then check if it is instance of NonExistingStackError or not. Therefore, calling stack property won't raise ImageBuilderActionError, instead raise ImageBuilderActionError in the caller.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -78,20 +78,36 @@ def __init__(self, stack_data: dict):
...
+    @property
+    def s3_bucket_name(self):
+        """Return the name of the bucket used to store image information."""
+        return self._get_tag(PCLUSTER_S3_BUCKET_TAG)
+
     @property
     def version(self):
         """Return the version of ParallelCluster used to create the stack."""
-        return self._get_tag("pcluster_version")
+        return self._get_tag("parallelcluster:version")
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: It depends on what kind of user experience we want to deliver.
case 1. A boto3 client operation failed due to  permission, we raise the error and let user manually add the permission and run command again.
case 2. A boto3 client operation failed, we log a warning message and continue executing the command.

For now, we are developing on case 1, raise error when there is a boto3 client operation failure.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: done

cli/src/common/boto3/ec2.py
yuleiwan: done

cli/src/pcluster/commands.py
yuleiwan: deregistering image and deleting snapshot is super quick, I think nowait is meanless if it is the case of image deletion not stack deletion. I'd remove nowait logic if it is image deletion.

cli/src/api/pcluster_api.py
yuleiwan: I don't fully understand this comment, you mean in API methods to pass imagebuilder and stack to ImageBuilderInfo constructor? like
```
    @staticmethod
    def delete_image(image_name: str, region: str, force: bool = False):
            ...
            imagebuilder = ImageBuilder(image_name)
            try:
               stack = imagebuilder.stack
            except StackError:
               stack = None
            imagebuilder.delete(force=force)
            return ImageBuilderInfo(imagebuilder=imagebuilder, stack=stack)
```
?

enrico-usai: Yes when the api method has the stack information is up to him to pass these information to ImageBuilderInfo, we don't need to do another call into the ImageBuilderInfo to retrieve stack information.

In this way the ImageBuilderInfo is just a collector of "static" information, we can consider it as an object representing the Json dictionary of the Api response, it shouldn't have methods or do api calls.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -67,6 +72,34 @@ def __init__(self, message: str, validation_failures: list = None):
...
+
+class NonExistingImageError(ImageError):
+    """Represent an error if image does not exist."""
+
+    def __init__(self, image_name):
+        super().__init__(f"Image {image_name} does not exist.")
+
+
+class NonExistingStackError(StackError):
+    """Repesent an error if stack doesn't exist."""
```
enrico-usai, line 97: minor: typo

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -112,62 +161,78 @@ def __init__(self, image_name: str, config: dict = None, stack: ImageBuilderStac
...
     def stack(self):
         """Return the ImageBuilderStack object."""
         if not self.__stack:
             try:
-                self.__stack = ImageBuilderStack(AWSApi().instance().cfn.describe_stack(self.image_name))
-            except AWSClientError:
-                raise ImageBuilderActionError(f"ImageBuilder stack {self.image_name} does not exist.")
+                self.__stack = ImageBuilderStack(AWSApi.instance().cfn.describe_stack(self.image_name))
+            except AWSClientError as e:
+                if f"Stack with id {self.image_name} does not exist" in str(e):
```
enrico-usai, line 196: What about StackNotFoundError ? I like the approch you're using with ImageNotFoundError

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -112,62 +161,78 @@ def __init__(self, image_name: str, config: dict = None, stack: ImageBuilderStac
...
+        if not self.__image:
             try:
-                self.__source_config = self.bucket.get_config(config_name=self._s3_artifacts_dict.get("config_name"))
-            except Exception as e:
-                raise ImageBuilderActionError(
-                    "Unable to load configuration from bucket " f"'{self.bucket.name}/{self.s3_artifacts_dir}'.\n{e}"
-                )
-        return self.__source_config
+                self.__image = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
+            except AWSClientError as e:
```
enrico-usai, line 207: Alternative way:
```python
except ImageNotFoundError:
    raise NonExistingImageError(self.image_name)
except AWSClientError as e:
    raise ImageError(f"Unable to get Image {self.image_name} info, due to {e}.")
```

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -168,6 +169,19 @@ def _add_resources(self):
         lambda_cleanup = self._add_lambda_cleanup(lambda_cleanup_policy_statements, build_tags_list)
         self._add_sns_topic(lambda_cleanup, build_tags_list)
 
+    def _get_image_tags(self):
+        tags = copy.deepcopy(self.config.image.tags) if self.config.image and self.config.image.tags else []
+        # TODO add tags for build log
+        tag_list = [
```
enrico-usai, line 175: This code is duplicated, we should have a common function for preparing cfn tags list

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -112,62 +161,78 @@ def __init__(self, image_name: str, config: dict = None, stack: ImageBuilderStac
...
+                if isinstance(e, ImageNotFoundError):
+                    raise NonExistingImageError(self.image_name)
+                raise ImageError(f"Unable to get Image {self.image_name} info, due to {e}.")
```
enrico-usai, line 210: I'm not sure we really need ImageError and StackError, maybe we can raise ImageBuilderActionError directly but let's talk about it.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -368,6 +422,53 @@ def _upload_artifacts(self):
                 f"Unable to upload imagebuilder cfn template to the S3 bucket {self.bucket.name} due to exception: {e}"
             )
 
+    def delete(self, force=False):
+        """Delete CFN Stack and associate resources and deregister the image."""
+        # validate image name
+        self._validate_image_name()
```
enrico-usai, line 428: We shouldn't validate image name at deletion time. If the image is created with a long(wrong?) name we need a way to delete it.
I think image name validation at creation time is enough.

cli/src/pcluster/models/imagebuilder.py
enrico-usai: What about `_is_image_in_use` or something like this?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -368,6 +422,53 @@ def _upload_artifacts(self):
...
+            if result:
+                logging.error(
+                    "Image %s is used by instances %s. In case you want to delete the image, "
+                    "please use the --force flag.",
+                    self.image_name,
+                    str(result),
+                )
+                raise ImageBuilderActionError("Unable to delete image and stack")
+            return False
+        except (AWSClientError, ImageError) as e:
```
enrico-usai, line 467:
```python
except NonExistingImageError:
    return False
except (AWSClientError, ImageError) as e:
    raise ImageBuilderActionError(f"Unable to delete image and stack, due to {str(e)}")
```

cli/src/pcluster/commands.py
yuleiwan: temporarily remove --no-wait

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done, it's fine to raise StackNotFoundError, but need to remind that the difference is error message of describe_image is defined by ourselves, but the error message of describe stack is define by boto3 client. So if boto3 client change the error message, StackNotFoundError is still not reliable.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: you mean it is duplicate comparing to imagebuilder_stack.py module or cluster.py module?

cli/src/pcluster/models/imagebuilder.py
yuleiwan: we can raise the ImageBuilderActionError in stack/image property rather than StackError/ImageError, but my concern is if we do that, will we lose the error trace, for example:
```
    @property
    def imagebuild_status(self):
        """Return the status of the stack of build image process."""
        try:
            cfn_status = self.stack.status
            for key, value in ImageBuilderStatusMapping.items():
                if cfn_status in value:
                    return key
            return None
        except StackError as e:
            raise ImageBuilderActionError(f"Unable to get imagebuilder {self.image_name} status, due to {e}")
```
If we raise error in self.stack, we don't know the error is caused by imagebuild_status

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

CHANGELOG.md
```diff
@@ -18,8 +18,8 @@ CHANGELOG
   `--ami-name-prefix`, `--custom-cookbook`, `--post-install`, `--no-public-ip`, `--cluster-template`, `--vpc-id`,
    `--subnet-id`.
 - Add `--image-name`, `--config`, `--region` parameters to `build-image` command.
-- Create S3 bucket per region shared with cluster and image if custom bucket isn't specified instead creating bucket
```
yuleiwan, line 21: duplicate with line 33

cli/src/api/pcluster_api.py
```diff
@@ -65,16 +65,40 @@ def __repr__(self):
...
-        self.imagebuild_status = imagebuilder.imagebuild_status
-        self.stack_status = imagebuilder.stack.status
-        self.stack_arn = imagebuilder.stack.id
+    def __init__(self, imagebuilder: ImageBuilder, stack=None):
+        self.stack_exist = False
+        self.image_exist = False
         self.region = get_region()
-        self.version = imagebuilder.stack.version
+        # image config file url
+        self.image_configurtion = imagebuilder.config_url
```
lukeseawalker, line 76: typo `configurtion`

cli/src/pcluster/models/imagebuilder.py
lukeseawalker: In this case snapshots cannot be deleted if image is not deregistered, so it's ok to fail fast

cli/src/common/boto3/ec2.py
```diff
@@ -112,6 +112,17 @@ def describe_image_by_name_tag(self, image_name):
         owners = ["self"]
         return self.describe_images(ami_ids=[], filters=filters, owners=owners)[0]
 
+    @AWSExceptionHandler.handle_client_exception
+    def describe_instances_by_ami_id(self, image_id):
```
lukeseawalker, line 116: method is returning just instance ids, not all the instance infos of the describe. Would "get_instance_ids_by_ami_id" fit better?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -381,3 +486,24 @@ def _validate_image_name(self):
                     str(1024 - len(AMI_NAME_REQUIRED_SUBSTRING))
                 )
             )
+
+    def _get_cfn_tags(self):
+        """Add cfn tags information to the stack."""
```
lukeseawalker, line 491: wrong doc

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -78,20 +113,41 @@ def __init__(self, stack_data: dict):
...
-
     @property
     def s3_artifact_directory(self):
         """Return the artifact directory of the bucket used to store image information."""
         return self._get_tag(PCLUSTER_S3_IMAGE_DIR_TAG)
 
+    @property
+    def tags(self):
+        """Return stack tags."""
+        return self._tags
```
enrico-usai, line 124: if we need tags outside we can avoid the underscore prefix and make the attribute private rather than having a property to expose the attribute.

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -133,16 +154,12 @@ def _add_cfn_parameters(self):
     def _add_resources(self):
         # Add default build tags information
         tags = copy.deepcopy(self.config.build.tags) or []
-        tags.append(BaseTag(key="pcluster_build_image", value=utils.get_installed_version()))
+        tags.append(BaseTag(key=PCLUSTER_IMAGE_NAME_TAG, value=self.image_name))
```
lukeseawalker, line 157: I think the `PCLUSTER_IMAGE_NAME_TAG` tag should be added as last step of the parallelcluster_tag component execution, because if the image is created during the build stage (with the tag PCLUSTER_IMAGE_NAME_TAG) and then the test stage will fail, we'll end up in having an Image with PCLUSTER_IMAGE_NAME_TAG that is not valid.

And the PCLUSTER_IMAGE_NAME_TAG is used by the build-image to check if an Image is already present.

An alternative to add the PCLUSTER_IMAGE_NAME_TAG in the parallelcluster_tag component execution is that we use another tag to check if AMI has passed the test, and we add a check on this new tag on the build-image  command

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -86,6 +92,21 @@ def _build_resource_name(self, name, to_lower=False):
...
+        if to_lower:
+            name = name.lower()
+        return name
+
+    def _get_log_group_arn(self):
+        log_group_arn = self.format_arn(
+            service="logs",
+            resource="log-group",
+            region=utils.get_region(),
+            resource_name=":{0}/{1}/{2}".format("aws", "imagebuilder", self._build_image_recipe_name()),
```
enrico-usai, line 106: let's use f-strings when possible. 
Here you can simplify a lot the resource name `f":aws/imagebuilder/{self._build_image_recipe_name()}"`

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -488,7 +518,9 @@ def _add_lambda_cleanup(self, policy_statements, build_tags):
                         service="iam",
                         resource="role",
                         region="",
-                        resource_name="{0}/{1}-{2}-*".format(RESOURCE_NAME_PREFIX, self.image_name, "InstanceRole"),
+                        resource_name="{0}/{1}-*".format(
+                            RESOURCE_NAME_PREFIX, "-".join([self.image_name, "InstanceRole"])[0:50]
```
enrico-usai, line 522: We should avoid this 0:50 everywhere and create a function/methdo like `_build_resource_name`

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: yes, I'd add this tag in tag component

cli/src/pcluster/templates/imagebuilder_stack.py
yuleiwan: done

cli/src/api/pcluster_api.py
yuleiwan: done

cli/src/common/boto3/ec2.py
yuleiwan: done

enrico-usai (approved):
LGTM

lukeseawalker (approved):

rexcsn (approved):

rexcsn (approved):

PR #2510, opened 2021-03-05
3.0: Add Compute Fleet launch template resources and other refactoring

enrico-usai:
# Add compute fleet Launch template resources 

### Model
* Added instance_type info and instance type related methods to BaseComputeResource
* Added get_custom_action to SlurmQueue to retrieve pre/post install scripts

### Stack generator
* Defined _get_block_device_mappings utility method to build device mapping for head node and queues.
* A root volume will be created for each Launch template if not defined in the config and added to the block_device_mappings.
* Created _get_custom_tags method and used it to add in the template custom tags from the users (they were missing in the head node too).
* Created _create_hash_suffix to generate random strings to be used as suffix for queue resources.
* Defined a generic _condition_disable_hyperthreading_manually that can check the setting for both head_node and compute_resource
* Added compute_node/user_data.sh file.

### Tests
* Small improvements to dummy_model

# Associate public ip to head node according to config param

I removed the default from the config because the cases are the following:
* assign_public_ip == true => set associate_public_ip_address to true in the LaunchTemplate
* assign_public_ip ==  false => set associate_public_ip_address to false in the LaunchTemplate
* unset => we don't set associate_public_ip_address e use the default from the subnet

# Add missing info in the dna.json

We're storing in the _storage_resource_options different values according to the
different storage type.

* EBS: List of mount_dirs that were previously mapped in the SharedDir param
* RAID: List of raid and volume info
* EFS: all storage info (but only a subset of them are really required by cookbook)
* FSX: all storage info (but only a subset of them are really required by cookbook)

# Other changes
* Define ComputeType enum and use it for validation …
* Reorder Launch template attributes and tags.
* Moved the user_data as latest attribute.
* Restored extra.json file.
* Append tags in the original configuration. We need them available in the config to be added to all the resources.

### DIFF
--- a/cli/src/pcluster/models/cluster_config.py
+++ b/cli/src/pcluster/models/cluster_config.py
@@ -360,7 +360,7 @@ class _BaseNetworking(Resource):
         proxy: Proxy = None,
     ):
         super().__init__()
-        self.assign_public_ip = Resource.init_param(assign_public_ip, default=False)
+        self.assign_public_ip = Resource.init_param(assign_public_ip)
         self.security_groups = Resource.init_param(security_groups)
         self.additional_security_groups = Resource.init_param(additional_security_groups)
         self.proxy = proxy
--- a/cli/src/pcluster/models/cluster_config.py
+++ b/cli/src/pcluster/models/cluster_config.py
@@ -401,7 +401,7 @@ class QueueNetworking(_BaseNetworking):
 
     def __init__(self, subnet_ids: List[str], placement_group: PlacementGroup = None, **kwargs):
         super().__init__(**kwargs)
-        self.subnet_ids = Resource.init_param(subnet_ids)
+        self.subnet_ids = Resource.init_param(subnet_ids)  # FIXME slurm support a single subnet id
         self.placement_group = placement_group
 
 
@@ -685,12 +685,12 @@ class HeadNode(Resource):
         )
 
     @property
-    def architecture(self):
+    def architecture(self) -> str:
         """Compute cluster's architecture based on its head node instance type."""
         return self._instance_type_info.supported_architecture()[0]
 
     @property
-    def vcpus(self):
+    def vcpus(self) -> int:
         """Get the number of vcpus for the instance according to disable_hyperthreading and instance features."""
         instance_type_info = self._instance_type_info
         default_threads_per_core = instance_type_info.default_threads_per_core()
@@ -723,11 +723,11 @@ class HeadNode(Resource):
         return self.__instance_type_info
 
     @property
-    def disable_simultaneous_multithreading_via_cpu_options(self):
+    def disable_simultaneous_multithreading_via_cpu_options(self) -> bool:
         """Return true if simultaneous multithreading must be disabled through cpu options."""
         return self.disable_simultaneous_multithreading and disable_ht_via_cpu_options(self.instance_type)
 
-    def get_custom_action(self, event: CustomActionEvent):
+    def get_custom_action(self, event: CustomActionEvent) -> CustomAction:
         """Return the first CustomAction corresponding to the specified event."""
         return (
             next((action for action in self.custom_actions if action.event == event), None)
@@ -753,6 +753,7 @@ class BaseComputeResource(Resource):
         self.disable_simultaneous_multithreading = Resource.init_param(
             disable_simultaneous_multithreading, default=False
         )
+        self.__instance_type_info = None
 
     def _validate(self):
         self._execute_validator(NameValidator, name=self.name)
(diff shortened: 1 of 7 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -959,14 +959,35 @@ def _add_head_node(self):
...
+                iam_instance_profile=CfnLaunchTemplate.IamInstanceProfileProperty(name=self.root_instance_profile.ref),
+                user_data=Fn.base64(
+                    Fn.sub(
+                        head_node_lt_user_data,
+                        {
+                            "YumProxy": head_node.networking.proxy if head_node.networking.proxy else "_none_",
+                            "DnfProxy": head_node.networking.proxy if head_node.networking.proxy else "",
+                            "AptProxy": head_node.networking.proxy if head_node.networking.proxy else "false",
+                            "ProxyServer": head_node.networking.proxy if head_node.networking.proxy else "NONE",
+                            "CustomChefCookbook": self._custom_chef_cookbook(),
```
yuleiwan, line 1022: check `_custom_chef_cookbook` func, it seems returning a cookbook object, not a cookbook url.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1052,20 +1046,16 @@ def _add_head_node(self):
...
                     else "NONE",
-                    # "cfn_shared_dir": self._cluster_config.head_node.storage.root_volume.mount_dir  # TODO
-                    # if self._cluster_config.head_node.storage and self._cluster_config.head_node.storage.root_volume
+                    # "cfn_shared_dir": head_node.storage.root_volume.mount_dir  # TODO
+                    # if head_node.storage and head_node.storage.root_volume
                     # else "NONE",
-                    "cfn_proxy": self._cluster_config.head_node.networking.proxy
-                    if self._cluster_config.head_node.networking.proxy
-                    else "NONE",
+                    "cfn_proxy": head_node.networking.proxy if head_node.networking.proxy else "NONE",
```
yuleiwan, line 1112: I am not sure "None" value works for chef client attributes json or not, did you test it? I use "" empty string instead of "NONE".

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -889,26 +890,28 @@ def _add_dynamo_db_table(self):
         self._dynamo_db_table = table
 
     def _add_head_node(self):
+        head_node = self._cluster_config.head_node
```
yuleiwan, line 965: for the _cluster_config field, I am concern about we store config object both in Cluster and ClusterCdkStack as an instance field. But Cluster also has ClusterCdkStack instance as a field. Is it ok for this dependency relationship? Or we should pass config to these private functions rather than store config as an instance field.

cli/src/pcluster/models/cluster_config.py
```diff
@@ -765,8 +766,45 @@ def _validate(self):
     @property
     def architecture(self):
         """Compute cluster's architecture based on its head node instance type."""
-        instance_type_info = AWSApi.instance().ec2.get_instance_type_info(self.instance_type)
-        return instance_type_info.supported_architecture()[0]
+        return self._instance_type_info.supported_architecture()[0]
```
yuleiwan, line 769: we use functions in utils including some boto3 call, I think we should refactor them later to align with the AWSApi.

cli/src/pcluster/models/cluster_config.py
```diff
@@ -765,8 +766,52 @@ def _validate(self):
...
+
+    @property
+    def vcpus(self):
```
yuleiwan, line 772: add hint?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -765,8 +766,52 @@ def _validate(self):
...
+
+    @property
+    def _instance_type_info(self) -> InstanceTypeInfo:
+        """Return instance type information as returned from aws ec2 describe-instance-types."""
+        if not self.__instance_type_info:
+            self.__instance_type_info = AWSApi.instance().ec2.get_instance_type_info(self.instance_type)
+        return self.__instance_type_info
+
+    @property
+    def disable_simultaneous_multithreading_via_cpu_options(self):
```
yuleiwan, line 805: add hint?

cli/src/pcluster/templates/cluster_stack.py
hanwen-pcluste: Yep, should be `self._custom_chef_cookbook().chef_cookbook`

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1258,8 +1299,203 @@ def _add_head_node(self):
                 properties={"ConfigVersion": self._cluster_config.config_version, "DynamoDBTable": ""},  # TODO
             )
 
-    def _head_node_efa_enabled(self):
-        return self._cluster_config.head_node.efa.enabled if self._cluster_config.head_node.efa else "NONE"
+    def _add_compute_fleet(self):
```
hanwen-pcluste, line 1302: Any thoughts on sharing some code with `_add_head_node()`?

cli/src/pcluster/templates/cluster_stack.py
enrico-usai: Thanks!

cli/src/pcluster/templates/cluster_stack.py
enrico-usai: It depends by the cookbook logic. For some of them in the recipes we have `if param != "NONE"`, I'm trying to keep the things aligned to old behavior to avoid changing the recipes for now.

cli/src/pcluster/models/cluster_config.py
enrico-usai: Yes is in the TODO list, I'm currently focused on CDK migration to enable end-to-end tests, then I'll go back to fix/align all these remaining things. Thanks for pointing that out.

cli/src/pcluster/templates/cluster_stack.py
enrico-usai: `Cluster` doesn't have `ClusterCdkStack` instance as a field.
`Cluster` just calls `CDKTemplateBuilder().build_cluster_template()` static method.

`_cluster_config` is just an internal reference to the the configuration we're passing to the ClusterCdkStack.
Let's talk about it.

cli/src/pcluster/models/cluster_config.py
enrico-usai: Done

cli/src/pcluster/models/cluster_config.py
enrico-usai: Done

cli/src/pcluster/templates/cluster_stack.py
enrico-usai: Yes, but it's not so easy because compute-fleet merges information from both queue and compute.
For sure we have a lot of things that can be refactored here, but I'm trying to reach an end-to-end flow, then we'll refactor and fix all the small things. Thanks for pointing this out.

hanwen-pcluste:
LGTM

PR #2516, opened 2021-03-10
3.0: Migrate Batch and CWLogs substacks to CDK

enrico-usai:
# Add possibility to suppress validators and filter by validation level to create command

* Rename ClusterInfo to ClusterStackInfo
* Rename FullClusterInfo to ClusterInfo
* Created an internal dictionary to pass compute_security_group information to slurm construct.

# Move CDK code related to slurm in a SlurmConstruct

* Added utility method in HeadNode e BaseQueue to retrieve instance_role.
* Added utility method in HeadNode and ComputeResource to check if simultaneous_multithreading
  must be disabled manually with a cookbook script.
* Added is_intel_hpc_platform_enabled and is_cw_logging_enabled properties in ClusterConfig.
* Added extra_chef_attributes and custom_chef_cookbook properties in ClusterConfig.
* Moved Slurm specific code into a SlurmConstruct in the slurm_builder module.
  Size of cluster_stack reduced from 1800+ rows to 1200+.
* Splitted HeadNode and ComputeFleet role, the code now will create a Role for each queue.
* Moved all the methods used by both SlurmConstruct and ClusterCdkStack in the cdk_builder_utils
* Moved packages_versions mappings in constants module and added it to bump-version.sh script.
* Fixed _add_compute_security_group to not use head_security_group if not created by us.
  (In the old template there was a single condition to create all the security groups,
  in pcluster3 there are different conditions).
* Moved role and policy creation in the same method because they are strictly related.
* Refactored roles and policies creation to use the same code for head node and queues.
* Pass the stack_name to use it as a string rather than using the internal self.stack_name from the Stack
  that is a reference, so it cannot be used for utility methods.
* Removed underscore prefix for stack variables passed to Slurm construct.
* Moved  head_ip and head_eni creation in the same method because they are strictly related.
* Moved _add_cleanup_resources_lambda and _add_cleanup_resources_bucket_custom_resource in the same method.
* Created an internal structure to store head_node/queue roles and the information if they are created by us
  or provided by the user. It is used to pass information to Slurm construct.

# Add possibility to configure different security groups for each queue
* We're creating a single new security group for compute fleet and associating it
  to all the queues for which there is no security group specified by the user.
* Defined some utility methods to return the complete list of security groups
  to be used for head node and compute.
* Fixed EFS mount target creation to use all the compute security groups,
  both created by us and provided by the user.
* Fixed FSx security groups list by using all the compute security groups,
  both created by us and provided by the user and not just the ones from the config.
* Created an internal dictionary to pass compute_security_group information to slurm construct.
* Define PclusterLambdaConstruct and refactor code to avoid code repetition
* Create utility method to create Lambda role to avoid code duplication

# Migrate AWS Batch substack to CDK Construct

## Model
* Move instance_type info and properties from BaseComputeResource to SlurmComputeResource
  because we can't get instance type information from Batch, on which the user can provide
  a list of instance types or instance families.
* Move validators using architecture property from BaseComputeResource to SlurmComputeResource
* Move EFA os architecture validator from BaseClusterConfig to SlurmClusterConfig
* Align validators tests/mocking to these changes.

## CDK builder
* Move get_queue_security_groups_full to cdk_builder_utils because it's required by both
  scheduler construct.
* Add raw_dict parameter to utility functions to generate tags, because Batch is accepting
  a raw dictionary.
* Migrated CloudWatch log groups creation into Awsbatch construct.

## Tests
* Generate two different dummy cluster config and test template creation for both the schedulers.

## Note about CfnRule
We're using a raw event_pattern rather than using events.EventPattern because of: aws/aws-cdk#5703

# Migrate ClodWatch logs substack to CDK

* Move log group creation for lambda function into PclusterLambdaConstruct
* Added a timestamp to CloudWatch Log group name for the cluster


cli/src/pcluster/templates/slurm_builder.py
```diff
@@ -0,0 +1,590 @@
...
+        core.CfnOutput(
+            scope=self.stack_scope,
+            id="ConfigVersion",
+            description="Version of the config used to generate the stack",
+            value=self.config.config_version,
+        )
+
+    def _add_private_hosted_zone(self):
+        cluster_hosted_zone = route53.CfnHostedZone(
+            self.stack_scope,
```
hanwen-pcluste, line 240: scope=self,.stack_scope

enrico-usai: Fixed

cli/src/pcluster/constants.py
```diff
@@ -44,3 +44,11 @@
 DEFAULT_MIN_COUNT = 0
 
 MAX_STORAGE_COUNT = {"ebs": 5, "efs": 1, "fsx": 1, "raid": 1}
+
+COOKBOOK_PACKAGES_VERSIONS = {
+    "parallelcluster": "2.10.1",
```
hanwen-pcluste, line 49: is this version used by `pcluster version`

cli/src/api/pcluster_api.py
```diff
@@ -47,27 +44,25 @@
...
         self.version = stack.version
+        self.scheduler = stack.scheduler
 
     def __repr__(self):
         return json.dumps(self.__dict__)
 
 
-class FullClusterInfo(ClusterInfo):
-    """Full representation of a running cluster."""
+class ClusterInfo(ClusterStackInfo):
```
hanwen-pcluste, line 58: Minor: I am not sure about the necessity of this class. Only tow more attributes and no additional method are added upon the base class.

hanwen-pcluste (approved):

hanwen-pcluste:
Awesome and speedy work. Thank you!

cli/src/common/aws/aws_resources.py
```diff
@@ -39,6 +39,11 @@ def status(self):
         """Return the status of the stack."""
         return self._stack_data.get("StackStatus")
 
+    @property
+    def is_working_status(self):
```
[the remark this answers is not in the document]
enrico-usai: It's a good question.
I tend to use properties when:
* there are no parameters.
* the method doesn't do any complex job.
* the method returns a value that is a "conceptual property" of the object.

We should avoid properties when there are parameters or the method is expensive (probably we'll have to change some of them).

cli/src/pcluster/commands.py
```diff
@@ -392,12 +394,13 @@ def ssh(args, extra_args):
...
+        except ImportError:
+            from pipes import quote as cmd_quote
+
         result = PclusterApi().describe_cluster(cluster_name=args.cluster_name, region=utils.get_region())
-        if isinstance(result, FullClusterInfo):
-            try:
-                from shlex import quote as cmd_quote
-            except ImportError:
-                from pipes import quote as cmd_quote
+        if isinstance(result, (ClusterStackInfo, ClusterInfo)):
```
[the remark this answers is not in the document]
enrico-usai: Thanks fixed

cli/src/pcluster/models/cluster.py
```diff
@@ -235,16 +236,23 @@ def status(self):
...
+        suppress_validators: bool = False,
+        validation_failure_level: FailureLevel = FailureLevel.ERROR,
+    ):
         """Create cluster."""
         # check cluster existence
         if AWSApi.instance().cfn.stack_exists(self.stack_name):
             raise ClusterActionError(f"Cluster {self.name} already exists")
 
-        validation_failures = self.config.validate()
+        validation_failures = [] if suppress_validators else self.config.validate()
```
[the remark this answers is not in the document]
enrico-usai: Thanks fixed

cli/src/api/pcluster_api.py
enrico-usai: Merged with ClusterStackInfo, thanks!

enrico-usai: Thanks, fixed

PR #2534, opened 2021-03-16
Fix disable_hyperthreading tests

ddeidda:
Some instance types may have fixed hyperthreading settings; instead this test was assuming that disabling hyperthreading is always possible. Fixed by comparing against instance data instead of fixed number of cores.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2535, opened 2021-03-16
Fix disable_hyperthreading tests

ddeidda:
Some instance types may have fixed hyperthreading settings; instead this test was assuming that disabling hyperthreading is always possible. Fixed by comparing against instance data instead of fixed number of cores.


tilne:
this looks like a backport of #2534 ?

tilne (approved):

PR #2537, opened 2021-03-17
Dump slurm job failure output into log message

yuleiwan:
Dump slurm job failure output into log message


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -405,9 +405,28 @@ def _submit_batch_job(
...
+
+        return ""
+
     def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
         result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id))
-        assert_that(result.stdout).contains("JobState=COMPLETED")
+        try:
+            assert_that(result.stdout).contains("JobState=COMPLETED")
+        except AssertionError as error:
+            error.args += (self._dump_job_output(result.stdout),)
```
demartinofra, line 428: optional: we could also simply logging.error

yuleiwan: done, use logging instead

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -405,9 +406,29 @@ def _submit_batch_job(
         else:
             return self._remote_command_executor.run_remote_command(submission_command, raise_on_error=raise_on_error)
 
+    def _dump_job_output(self, job_info):
+        match = re.match(r".*StdErr=(.*)?\sStdIn=(.*)?\sStdOut=(.*)?\sPower.*", job_info)
```
tilne, line 410: There's a couple ways I'd simplify this.

1. This regex assumes that `StdErr` will always be before `StdIn`, which is always before `StdOut`, which is always before `Power`. Instead of relying on those assumptions I would loop through a list of only the strings we care about ('StdOut' and 'StdErr') and do the regex matching for each of them.
2. Use `re.search` instead of `re.match` so that you don't have to worry about not matching at the beginning of the string.

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -405,9 +406,29 @@ def _submit_batch_job(
...
+            stderr = match.group(1)
+            stdout = match.group(3)
+            if stdout == stderr:
+                command = f'echo "stderr/stdout:" && cat {stderr}'
+            else:
+                command = f'echo "stderr:" && cat {stderr} && echo "stdout:" && cat {stdout}'
+            result = self._remote_command_executor.run_remote_command(command)
+
+            return result.stdout
+
```
tilne, line 421: It would be nice to log something when the function is unable to parse the StdOut or StdErr paths from the job info string.

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -405,9 +406,29 @@ def _submit_batch_job(
...
+        return ""
+
     def assert_job_succeeded(self, job_id, children_number=0):  # noqa: D102
         result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id))
-        assert_that(result.stdout).contains("JobState=COMPLETED")
+        try:
+            assert_that(result.stdout).contains("JobState=COMPLETED")
+        except AssertionError:
+            logging.info("Job output")
+            logging.error(self._dump_job_output(result.stdout))
```
tilne, line 430: minor: I would move the logging calls into _dump_job_output. That way it doesn't have to return anything.

tests/integration-tests/tests/common/schedulers_common.py
yuleiwan: Done

tests/integration-tests/tests/common/schedulers_common.py
yuleiwan: Done

tests/integration-tests/tests/common/schedulers_common.py
yuleiwan: Done

tilne (approved):

PR #2540, opened 2021-03-17
3.0: Initial migration of integration test framework and integration tests

hanwen-pcluste:
See each commit description for details.


tests/integration-tests/conftest.py
```diff
@@ -363,24 +368,44 @@ def _config_renderer(config_file="pcluster.config.ini", **kwargs):
...
+    if config_content["Image"].get("CustomAmi") is None:
+        additional_filters = []
+        if request.config.getoption("createami_cookbook_git_ref"):
+            additional_filters.append(
+                {
+                    "Name": "tag:parallelcluster_cookbook_ref",
+                    "Values": [request.config.getoption("createami_cookbook_git_ref")],
+                }
+            )
+        config_content["Image"]["CustomAmi"] = retrieve_pcluster_ami_with_additional_filters(
```
demartinofra, line 397: I wouldn't do this otherwise we end up testing only the custom ami path. can we change the ec2 query in the CLI so that it also fetches the private AMIs in the account?Look at the implementation of `_get_official_image_id` in develop.

tests/integration-tests/conftest.py
```diff
@@ -363,24 +368,44 @@ def _config_renderer(config_file="pcluster.config.ini", **kwargs):
...
+            architecture=get_architecture_supported_by_instance_type(config_content["HeadNode"]["InstanceType"]),
+            additional_filters=additional_filters,
+        )
 
+    with open(cluster_config, "w") as conf_file:
+        yaml.dump(config_content, conf_file)
+
+
+"""
     for custom_option in [
```
demartinofra, line 408: what about the rest of the method? do you plan to fix it in a separate PR?

tests/integration-tests/conftest.py
```diff
@@ -363,24 +368,44 @@ def _config_renderer(config_file="pcluster.config.ini", **kwargs):
         env = Environment(loader=file_loader)
         rendered_template = env.get_template(config_file).render(**{**kwargs, **default_values})
         config_file_path.write_text(rendered_template)
-        add_custom_packages_configs(config_file_path, request, region)
-        _enable_sanity_check_if_unset(config_file_path)
```
demartinofra, line 367: you removed this but you are not enabling sanity check anywhere else. Shall we do it when we call the CLI commands that accept it?

tests/integration-tests/test_runner.py
```diff
@@ -213,6 +214,11 @@ def _init_argparser():
         default=TEST_DEFAULTS.get("createami_custom_cookbook_url"),
         type=_is_url,
     )
+    custom_group.add_argument(
+        "--createami-cookbook-git-ref",
```
demartinofra, line 218: probably we can get rid of this if we change the way we retrieve the AMI Ids

tests/integration-tests/tests/cfn-init/test_cfn_init/test_replace_compute_on_failure/pcluster.config.yaml
```diff
@@ -0,0 +1,27 @@
+Image:
+  Os: {{ os }}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  DisableSimultaneousMultithreading: false
```
demartinofra, line 9: please add a FIXME next to all unneeded defaults we will have to remove

tests/integration-tests/tests/cfn-init/test_cfn_init/test_replace_compute_on_failure/pcluster.config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    Settings:
+      ScaledownIdletime: 10
+    Queues:
+      - Name: queue-0
+        CustomActions:
+          - Script: s3://{{ bucket_name }}/post_install.sh
+        ComputeResources:
+          - Name: compute-resource-0
+            InstanceType: {{ instance }}
+            MinCount: 10
```
demartinofra, line 23: why MinCount 10?

tests/integration-tests/tests/common/utils.py
```diff
@@ -94,6 +94,29 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
         raise
 
 
+def retrieve_pcluster_ami_with_additional_filters(os, architecture, additional_filters):
```
demartinofra, line 97: if this does the same as `retrieve_latest_ami` why not adding the `filters` parameter to it and set a default to None?

util/upload-cfn-templates.py
```diff
@@ -1,236 +0,0 @@
-#!/usr/bin/python
```
demartinofra, line 1: nit: can we remove this in a separate commit and eventually PR? It doesn't have much to do with the integration testing framework right?

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: Done

cli/src/pcluster/models/cluster.py
```diff
@@ -355,7 +355,7 @@ def _upload_config(self):
             if self.config.source_config:
                 result = AWSApi.instance().s3.put_object(
                     bucket_name=self.bucket.name,
-                    body=yaml.dump(self.config.source_config),
+                    body=yaml.dump(ClusterSchema().dump(self.config)),
```
enrico-usai, line 358: I'd upload two different files, the original one and the dumped one (with defaults).

The first one is useful if we want to retrieve the original config written by the customer.

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -66,6 +66,8 @@ def only_one_field(self, data, field_list, **kwargs):
             return True
         return len([data.get(field_name) for field_name in field_list if data.get(field_name)]) == 1
 
+
+'''
```
enrico-usai, line 70: TODO

tests/integration-tests/tests/cfn-init/test_cfn_init/test_replace_compute_on_failure/pcluster.config.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - Name: queue-0
+        CustomActions:
+          - Script: s3://{{ bucket_name }}/post_install.sh
+        ComputeResources:
+          - Name: compute-resource-0
+            InstanceType: {{ instance }}
+        Networking:
+          SubnetIds:
+            - {{ public_subnet_id }}
+        Iam:
```
enrico-usai, line 28: nit: you can use [Yaml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to avoid repeating some blocks.

tests/integration-tests/tests/common/utils.py
```diff
@@ -105,9 +108,8 @@ def retrieve_pcluster_ami_without_standard_naming(region, os, version, architect
             Filters=[
                 {"Name": "name", "Values": [official_ami_name]},
                 {"Name": "architecture", "Values": [architecture]},
-                {"Name": "is-public", "Values": ["true"]},
             ],
-            Owners=OS_TO_PCLUSTER_AMI_NAME_OWNER_MAP.get(os).get("owners"),
+            Owners=["self"],
```
enrico-usai, line 112: Should we also include the official ones?

tests/integration-tests/tests/create/test_create/test_create_wrong_pcluster_version/pcluster.config.ini
```diff
@@ -1,22 +0,0 @@
...
-
-[cluster default]
-base_os = {{ os }}
-key_name = {{ key_name }}
-vpc_settings = parallelcluster-vpc
-scheduler = {{ scheduler }}
-master_instance_type = {{ instance }}
-compute_instance_type = {{ instance }}
-custom_ami = {{ custom_ami }}
-master_root_volume_size = 100
```
enrico-usai, line 15: missing volume size and use public ips options

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
```diff
@@ -1,21 +0,0 @@
...
-aws_region_name = {{ region }}
-
-[cluster default]
-base_os = {{ os }}
-key_name = {{ key_name }}
-vpc_settings = parallelcluster-vpc
-scheduler =  {{ scheduler }}
-master_instance_type = {{ instance }}
-custom_ami = {{ custom_ami }}
-master_root_volume_size = 100
```
enrico-usai, line 14: missing volume size and use public ips options

tests/integration-tests/tests/dns/test_dns/test_hit_no_cluster_dns_mpi/print_hostname.sh
```diff
@@ -1,2 +0,0 @@
-#!/bin/bash
-echo `hostname` > /shared/compute_hostname
```
enrico-usai, line 2: Why did you remove it?

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Agreed, I will add a ToDo. I chose to upload one file to avoid change in cookbook

tests/integration-tests/tests/common/utils.py
hanwen-pcluste: No. We need to call `copy_image`, which leads to permission error if the source AMI is official

tests/integration-tests/tests/create/test_create/test_create_wrong_os/pcluster.config.ini
hanwen-pcluste: Default volume size was increased to 35, so the specification is no longer needed.
A False default of ElasticIp will be added to the config layer.
So neither of the parameter is necessary now

tests/integration-tests/tests/dns/test_dns/test_hit_no_cluster_dns_mpi/print_hostname.sh
hanwen-pcluste: https://github.com/aws/aws-parallelcluster/pull/2018/files#diff-72b5de0831d42ea5780a01dcadb05167c21810c8c1416993f31a43eaff22eb83R43
This PR forgot to remove it.

tests/integration-tests/tests/common/utils.py
```diff
@@ -94,6 +97,29 @@ def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
...
+        )
+        return response["Images"][0]["ImageId"]
+    except ClientError as e:
+        LOGGER.critical(e.response.get("Error").get("Message"))
+        raise
+    except AttributeError as e:
+        LOGGER.critical("Error no attribute {0} in dict: {1}".format(os, e))
+        raise
+    except IndexError as e:
+        LOGGER.critical("Error no ami retrieved: {0}".format(e))
```
chenwany: I think in pcluster3 we use fstrings. We can use fstrings everywhere rather than format. You can see here [why.](https://realpython.com/python-f-strings/)

tests/integration-tests/requirements.txt
```diff
@@ -17,3 +17,19 @@ PyYAML
...
+aws-cdk.aws-efs
+aws-cdk.aws-events
+aws-cdk.aws-fsx
+aws-cdk.aws-imagebuilder
+aws-cdk.aws-iam
+aws-cdk.aws-logs
+aws-cdk.aws-route53
+aws-cdk.aws-ssm
+aws-cdk.aws-sqs
+aws_cdk.aws_lambda
```
chenwany, line 35: minor: aws_cdk.aws_lambda here should be aws_cdk.aws-lambda. 
minor2: let's put the imports in alphabetical order, in this way it's simpler to find them.

tests/integration-tests/conftest.py
demartinofra: We might want to introduce something like an AmiSearchTags param in DevSettings where we specify the node and cookbook git ref when testing develop?

tests/integration-tests/conftest.py
demartinofra: there are some cases where we pass to the `test_runner` global `pre-install` or `post-install` through the --pre-install or --post-install cli argument so we need to preserve this behavior. Let's follow-up on this.

tests/integration-tests/conftest.py
demartinofra: you are right, just disregard my comment.

tests/integration-tests/tests/common/utils.py
demartinofra: additional best practice: when using a logging object it is better to use lazy logging evaluation so that no concatenation or string formatting is performed if the log level is lower than the configured one. This means that rather than using format or f-strings with the logger object you want to default to the logging specific formatting capabilities.

https://pycodequ.al/docs/pylint-messages/w1201-logging-not-lazy.html

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Fixed

cli/src/pcluster/schemas/common_schema.py
hanwen-pcluste: Fixed

tests/integration-tests/requirements.txt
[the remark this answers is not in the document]
hanwen-pcluste: Fixed by adding these dependencies to setup.py

tests/integration-tests/conftest.py
```diff
@@ -521,6 +530,14 @@ def setup_sts_credentials(region, request):
     unset_credentials()
 
 
+@pytest.fixture(scope="class", autouse=True)
+def setup_env_variable(region):
```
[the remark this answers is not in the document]
demartinofra: can't we have those fixtures depend on the region one so that we don't rely on this env variable. For example when creating an object of type Cluster we could pass the region and then use the `--region` argument when executing commands.

My concern is that by setting the env variable we create an implicit dependency. During the teardown of fixtures since Cluster does not explicitly depend on region we could have this fixture being deleted before Cluster and thus breaking the cleanup of the Cluster fixture

tests/integration-tests/clusters_factory.py
```diff
@@ -156,15 +155,15 @@ def cfn_name(self):
     @property
     def region(self):
```
demartinofra, line 156: Shall we pass region as an arg to the Cluster __init__?

tests/integration-tests/conftest.py
```diff
@@ -74,6 +74,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--createami-custom-chef-cookbook", help="url to a custom cookbook package for the createami command"
     )
+    parser.addoption("--createami-cookbook-git-ref", help="Git ref of the cookbook used to bake the AMI")
```
demartinofra, line 77: can be done in a separate PR but let's name this cookbook-git-ref only and add a similar argument for the node too

tests/integration-tests/conftest.py
```diff
@@ -436,24 +461,13 @@ def _get_arn_partition(region):
...
-        config.write(f)
-
-
 def _get_default_template_values(vpc_stack, request):
     """Build a dictionary of default values to inject in the jinja templated cluster configs."""
     default_values = get_vpc_snakecase_value(vpc_stack)
     default_values.update({dimension: request.node.funcargs.get(dimension) for dimension in DIMENSIONS_MARKER_ARGS})
+    scheduler = default_values.get("scheduler")
+    if scheduler is not None:
+        default_values["scheduler"] = scheduler.title()
```
demartinofra, line 470: I would do this in the cluster config by using jinja builtin filter: `{{ scheduler | title }}`

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.yaml
```diff
@@ -0,0 +1,24 @@
...
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+Scheduling:
+  {{ scheduler }}:
+    Queues:
+      - Name: queue-0
```
demartinofra, line 12: I believe the default name for a SIT queue was `compute`. I'd preserve this in case some tests rely on it,

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -74,25 +74,29 @@ def fields_coexist(self, data, field_list, one_required=False, **kwargs):
     @pre_dump
     def remove_implied_values(self, data, **kwargs):
         """Remove value implied by the code. i.e., only keep parameters that were specified in the yaml file."""
-        for key, value in vars(data).copy().items():
-            if _is_implied(data, key, value):
-                delattr(data, key)
-            if isinstance(value, list):
-                value[:] = [v for v in value if not _is_implied(data, key, v)]
+        if self.context.get("delete_defaults_when_dump"):
```
enrico-usai, line 77: I'd rename the context to: "no_defaults"

cli/src/pcluster/models/cluster_config.py
```diff
@@ -433,24 +433,26 @@ def __init__(self, enabled: bool, port: int = None, allowed_ips: str = None):
 class Efa(Resource):
     """Represent the EFA configuration."""
 
-    def __init__(self, enabled: bool = None, gdr_support: bool = None):
-        super().__init__()
-        self.enabled = Resource.init_param(enabled, default=True)
+    def __init__(self, enabled: bool = None, gdr_support: bool = None, **kwargs):
```
enrico-usai, line 436: We can extend `Efa` to accept instance_type and initialize the enabled value according to this param if passed.
```python
def __init__(self, enabled: bool = None, gdr_support: bool = None, instance_type: str = None, **kwargs):
        super().__init__(**kwargs)

        default_enabled = AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported() if instance_type else True
        self.enabled = Resource.init_param(gdr_support, default=default_enabled)
        self.gdr_support = Resource.init_param(gdr_support, default=False)
```

then initializing it with:
```python
self.efa = efa or Efa(instance_type=self.instance_type, implied=True)
```

This should simplify tests and mocking.

tests/integration-tests/conftest.py
hanwen-pcluste: Why would you prefer the code change in multiple config files than this single change?

tests/integration-tests/conftest.py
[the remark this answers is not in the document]
hanwen-pcluste: The env var is only for cluster commands. Here `scope="class", autouse=True` guarantees the correct execution order with `clusters_factory` which has `scope="class"`. It is also feasible to add --region to each of commands needed. For now, I was wondering if I could just use this minimal change and add a FixMe comment?

cli/src/pcluster/models/common.py
```diff
@@ -71,10 +71,11 @@ def update_policy(self):
             """Return the update policy."""
             return self.__update_policy()
 
-    def __init__(self):
+    def __init__(self, implied: bool = False):
```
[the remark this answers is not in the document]
hanwen-pcluste: This renaming requires change in files that are unrelated to this PR. I would address it in another PR if needed

tests/integration-tests/conftest.py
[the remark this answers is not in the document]
demartinofra: I agree that Cluster is not affected now. I'm afraid that future fixtures will start to rely on this without being aware of this side effect. ok with marking as fixme and handling separately

tests/integration-tests/clusters_factory.py
hanwen-pcluste: Done

tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging/test_cloudwatch_logging/pcluster.config.yaml
```diff
@@ -0,0 +1,24 @@
...
+        ComputeResources:
+          - Name: compute-resource-0
+            InstanceType: {{ instance }}
```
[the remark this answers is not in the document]
hanwen-pcluste: Fixed

enrico-usai (approved):

PR #2556, opened 2021-03-19
integ-tests: Enable runtime baking test

yuleiwan:
Enable runtime baking test

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -5,9 +5,9 @@ test-suites:
 {% include 'common/common.yaml' %}
 {% endfilter %}
   runtime_bake:
-    test_runtime_bake.py::test_runtime_bake:  # These are currently skipped
+    test_runtime_bake.py::test_runtime_bake: # arm instance tests are temporarily skipped
       dimensions:
-        {%- for os, scheduler in [("alinux2", "slurm"), ("ubuntu2004", "torque"), ("centos7", "sge")] %}
+        {%- for os, scheduler in [("centos8", "torque"), ("centos7", "sge"), ("alinux2", "slurm"), ("ubuntu2004", "torque")] %}
         - regions: ["eu-west-2"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["{{ os }}"]
@@ -19,12 +19,12 @@ test-suites:
           oss: ["{{ os }}"]
           schedulers: ["{{ scheduler }}"]
         {%- endfor %}
-        {%- for os, scheduler in [("ubuntu1804", "sge"), ("alinux2", "slurm")] %}
-        - regions: ["us-east-1"]
-          instances: {{ common.INSTANCES_DEFAULT_ARM }}
-          oss: ["{{ os }}"]
-          schedulers: ["{{ scheduler }}"]
-        {%- endfor %}
+#        {%- for os, scheduler in [("ubuntu1804", "sge"), ("alinux2", "slurm")] %}
+#        - regions: ["us-east-1"]
+#          instances: {{ common.INSTANCES_DEFAULT_ARM }}
+#          oss: ["{{ os }}"]
+#          schedulers: ["{{ scheduler }}"]
+#        {%- endfor %}
   scaling:
     test_scaling.py::test_nodewatcher_terminates_failing_node:
       dimensions:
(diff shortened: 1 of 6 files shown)


tests/integration-tests/configs/develop.yaml
```diff
@@ -5,9 +5,9 @@ test-suites:
 {% include 'common/common.yaml' %}
 {% endfilter %}
   runtime_bake:
-    test_runtime_bake.py::test_runtime_bake:  # These are currently skipped
+    test_runtime_bake.py::test_runtime_bake: # arm instance tests are temporarily skipped
       dimensions:
-        {%- for os, scheduler in [("alinux2", "slurm"), ("ubuntu2004", "torque"), ("centos7", "sge")] %}
+        {%- for os, scheduler in [("alinux2", "torque"), ("centos7", "sge"), ("centos8", "slurm"), ("ubuntu2004", "torque")] %}
```
tilne, line 10: was there a reason for changing this so that AL2 runs with torque instead of slurm?

tests/integration-tests/configs/develop.yaml
```diff
@@ -19,12 +19,6 @@ test-suites:
           oss: ["{{ os }}"]
           schedulers: ["{{ scheduler }}"]
         {%- endfor %}
-        {%- for os, scheduler in [("ubuntu1804", "sge"), ("alinux2", "slurm")] %}
```
tilne, line 22: Is it possible to comment this out rather than removing it so it's easier to restore later?

tests/integration-tests/tests/runtime_bake/test_runtime_bake.py
```diff
@@ -13,35 +13,30 @@
...
+@pytest.mark.dimensions("eu-west-2", "c5.xlarge", "centos7", "sge")
+@pytest.mark.dimensions("eu-west-2", "c5.xlarge", "ubuntu2004", "torque")
+# @pytest.mark.dimensions("us-east-1", "m6g.xlarge", "alinux2", "slurm")
+# @pytest.mark.dimensions("us-east-1", "m6g.xlarge", "ubuntu1804", "sge")
 @pytest.mark.usefixtures("instance", "scheduler")
 def test_runtime_bake(scheduler, os, region, pcluster_config_reader, clusters_factory, test_datadir, architecture):
     """Test cluster creation with runtime bake."""
-    # remarkable AMIs are not available for ARM yet
-    ami_type = "remarkable" if architecture == "x86_64" else "official"
+    # remarkable AMIs are not available for ARM yet and disable centos7,8 remarkable AMIs
```
tilne, line 30: From what you said offline we're not using remarkable AMIs for centos 7 or 8 for x86_64 for two reasons:
* FPGA ami’s volume size is not enough, so we disable testing it
* FPGA ami installs epel, it is not suitable for our case to test it

Can you add this information to the comment?

tests/integration-tests/tests/runtime_bake/test_runtime_bake/test_runtime_bake/verify_cinc_download.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -x
 
+# get cloudformation log if it doesn't exist
+if [ ! -d "/var/log/cloud-init-output.log" ]; then
```
tilne, line 5: `-d` tests if something is a directory. i don't think that's what you want here.

tests/integration-tests/configs/develop.yaml
tilne: No need to change it. I was just wondering.

tests/integration-tests/tests/runtime_bake/test_runtime_bake/test_runtime_bake/pre-install.sh
```diff
@@ -0,0 +1,25 @@
...
+
+. /etc/os-release
+release="${ID}${VERSION_ID:+.${VERSION_ID}}"
+
+if [ `echo ${release} | grep '^centos\.7'` ]; then
+  os='centos7'
+elif [ `echo "${release}" | grep '^centos\.8'` ]; then
+  os='centos8'
+else
+  os='other'
```
tilne, line 12: I think it would be good to fail here, because it would be much easier to debug an explicit failure here and change the OS detection logic instead of having to trace a failure during the runtime baking process back to a silent failure. This would require not running this script for OSs other than centos7 and centos8.

tests/integration-tests/tests/runtime_bake/test_runtime_bake.py
```diff
@@ -9,39 +9,45 @@
...
-    ami_type = "remarkable" if architecture == "x86_64" else "official"
+    # Remarkable AMIs are not available for ARM yet
+    # Disable centos7,8 remarkable AMIs, because FPGA AMI volume size is not enough and we need to test epel
+    # installation during runtime baking, but FPGA AMI is baked with epel
+    if architecture == "x86_64" and os not in ["centos7", "centos8"]:
+        ami_type = "remarkable"
+    else:
+        ami_type = "official"
+
+    # Create S3 bucket for pre install scripts
```
tilne, line 40: Can you expand this comment to include an explanation of why we're running the pre-install script?

PR #2560, opened 2021-03-22
Drop support for python versions < 3.6

rexcsn:
* Remove dependency on `future` package and `__future__` module. `absolute_import`, `print_function`, `unicode_literals` are available natively with python3
* Remove comments about older versions of python

### DIFF
--- a/cli/src/awsbatch/utils.py
+++ b/cli/src/awsbatch/utils.py
@@ -11,7 +11,6 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
-from __future__ import print_function
 
 import pipes
 import re
@@ -115,9 +114,6 @@ def is_mnp_job(job):
 def get_job_type(job):
     """
     Get the type of the job.
-
-    Job type is of type string and not enum since enums have been introduced
-    since Python 3.4.
 
     :param job: the job dictionary returned by AWS Batch api
     :return: one of ["SIMPLE", "ARRAY", "MNP"]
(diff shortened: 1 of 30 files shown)


yuleiwan (commented):
should we remove all the comments about python versions < 3.6?
for version 2.7, we have comments in `/pcluster/configure/networking.py` and `/pcluster/configure/subnet_computation.py`
for version 3.4, we have comments in `/awsbatch/utils.py`
for version 3.5, we have comments in `/pcluster/config/mappings.py`

cli/src/awsbatch/utils.py
```diff
@@ -116,9 +115,6 @@ def get_job_type(job):
     """
     Get the type of the job.
 
-    Job type is of type string and not enum since enums have been introduced
```
yuleiwan, line 119: optional: Should we convert these string to enum? Since we already saitisfy the python version requirement.

cli/src/pcluster/config/cfn_param_types.py
```diff
@@ -268,8 +268,6 @@ def get_value_from_string(self, string_value):
         """Return internal representation starting from CFN/user-input value."""
         param_value = self.get_default_value()
         try:
-            # Do not convert empty string and use format and yaml.load in place of json.loads
-            # for Python 2.7 compatibility because it returns unicode chars
             if string_value:
```
yuleiwan, line 271: Minor: should we also refactor this method since  Python 2.7 has been removed, but it's fine if we keep this logic, because it doesn't affect functionality.

yuleiwan (approved):

cli/src/pcluster/config/cfn_param_types.py
rexcsn: Yep I am for keeping the current implementation

cli/src/awsbatch/utils.py
rexcsn: I am for keeping implementation as is, because converting from string to enum does not bring significant benefit to functionality

PR #2567, opened 2021-03-24
3.0: Remove old/useless code and tests

enrico-usai:
## Removed Code
- cfn_param_types, param_types and json_param_types (Resources and Param in the new model)
- hit_converter and pcluster_config CLI utility
- iam_policy_rules (now are added by CDK when needed)
- mappings (we're using schema and config classes)
- pcluster_config (redesigned)
- resource map
- hit and sit cluster model
- commands: _setup_bucket_with_resources, _upload_cluster_artifacts, upload_hit_resources,
  upload_dashboard_resource, _validate_cluster_name, evaluate_pcluster_template_url,
  _get_pcluster_version_from_stack, _get_default_template_url (moved from commands to config/cluster classes)
- createami (already replaced with build-image command)
- utils: get_cluster_name, get_stack_version, cleanup_s3_resources, get_batch_ce, get_batch_ce_capacity (moved to new classes)
- utils: get_cloudformation_directory, get_asg_name, get_asg_settings, (SGE and Torque are deprecated, ASG is unused)
- utils: is_hit_enabled_scheduler, is_hit_enabled_cluster, read_remote_file (we no longer have sit logic)
- utils: get_bucket_url, get_file_section_name (useless)
- utils: render_template (we don't need to render any template)

## Removed Tests
- all the "from_cfn" tests (we're retriving the source config from s3)
- cluster model and hit converter tests (sit model no longer exists, we just have hit model)
- json param types tests (we changed cookbook to retrieve the info from source config and instance types data)
- pcluster config and sections tests (internal classes have been redesigned)
- aliases, global and aws sections tests (no longer available in the config file)
- source consistency (mappings file removed)
- validators (migrated)
- createami (moved to build-ami)

## Additional changes
- Remove examples config file 
- Remove old CloudFormation templates and related tests
- Execute InstanceTypeBaseAMICompatibleValidator for Slurm only. The validator doesn't work for AWS Batch because it can contain a list of instance families. Fix awsbatch attribute to be instance_types and not instance_type.
- Improve configure command to use AWSApi
- Move AWS Batch validators code from utils to awsbatch_validators 
- Move get_stack_name as cluster method


demartinofra (commented):
Awesome! Thanks for organizing this in separate commits.

demartinofra (approved):

PR #2571, opened 2021-03-25
3.0: Fix pcluster delete issues

yuleiwan:
* Exit no wait if stack has been deleted
* Return template in yaml format instead of string. 
* Update stack with updated template
* Check updated status is UPDATE_COMPLETE_CLEANUP_IN_PROGRESS
*  Update stack data after executing stack deletion


cli/src/pcluster/commands.py
```diff
@@ -520,6 +520,14 @@ def delete(args):
                     ).ljust(80)
                     sys.stdout.write("\r%s" % resource_status)
                     sys.stdout.flush()
+                elif isinstance(result, ApiFailure):
+                    # If stack is already deleted
+                    if f"Cluster {args.cluster_name} doesn't exist." in result.message:
+                        LOGGER.warning(f"\nCluster {args.cluster_name} has already been deleted or does not exist.")
```
hanwen-pcluste, line 526: Why treating `doesn't exist` error differently?

cli/src/pcluster/models/cluster.py
```diff
@@ -175,17 +175,18 @@ def _get_unretained_cw_log_group_resource_keys(self):
 
     def _persist_stack_resources(self, keys):
         """Set the resources in template identified by keys to have a DeletionPolicy of 'Retain'."""
+        template = self.template
```
hanwen-pcluste, line 178: Why not using self.template directly?

cli/src/pcluster/commands.py
hanwen-pcluste: Is this path reachable? I guess all `doesn't exist` exception goes to line 542?

cli/src/pcluster/models/cluster.py
yuleiwan: Because self.template returns `yaml.safe_load(AWSApi.instance().cfn.get_stack_template(self.name))`. It will always requery the exisiting stack info, although we modify the DeletionPocliy here, we will get original template by using self.template. But we actually need modified template to update the stack.

cli/src/pcluster/commands.py
yuleiwan: Because `result = PclusterApi().describe_cluster(cluster_name=args.cluster_name, region=utils.get_region())` in while loop will get APIFailure at the end of deletion, the stack has been deleted. It's not the ClientError, so will go to this branch.

enrico-usai (approved):

PR #2573, opened 2021-03-25
Add linters to CI

tilne:
* Enable dependency updates via dependabot PRs
* Add semgrep, bandit, and dependency-check to CI
* Add a [pre-commit](https://pre-commit.com/) config file that can be used to automatically run linters pre-commit on modified source


.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
+repos:
```
enrico-usai, line 1: Why there are two .pre-commit-config files?

tilne: Good catch. Didn't mean to add one under the github dir

.github/workflows/dep_check.yml
```diff
@@ -0,0 +1,19 @@
+# See https://semgrep.dev/explore to learn about what the checks do.
+name: DepCheck
+on: [pull_request]
+jobs:
+  semgrep:
```
rexcsn, line 5: Minor: Since dependency-check is unrelated to semgrep, we might want to rename this.

rexcsn (dismissed):

.github/workflows/dep_check.yml
tilne: Done.

.github/workflows/dep_check.yml
```diff
@@ -0,0 +1,18 @@
+name: DepCheck
+on: [pull_request]
+jobs:
+  depcheck:
+    runs-on: ubuntu-latest
+    name: DepCheck
+    steps:
+      - uses: actions/checkout@v1
```
demartinofra, line 8: can we use actions/checkout@v2?

.github/workflows/dep_check.yml
```diff
@@ -0,0 +1,18 @@
+name: DepCheck
```
demartinofra, line 1: I'm wondering if we should have these new workflows as new jobs of our CI workflow rather than creating new workflows.

demartinofra: also I'm wondering if it makes sense to implement depcheck and semgrep as tox actions so that it's easier to run such linters locally as part of the tox -e code-linters

.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - id: requirements-txt-fixer
+      - id: mixed-line-ending
+        args: ['--fix=no']
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.0
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-docstrings, flake8-bugbear, flake8-mypy, flake8-colors, pep8-naming]
+        entry: bash -c 'cd cli && flake8 $(for arg in $@; do echo -n "../$arg " ; done)'
```
demartinofra, line 27: what is this for?

cli/src/pcluster/resources/custom_resources/custom_resources_code/crhelper/utils.py
```diff
@@ -17,17 +17,17 @@ def _send_response(response_url, response_body):
     except Exception as e:
         msg = "Failed to convert response to json: {}".format(str(e))
         logger.error(msg, exc_info=True)
-        response_body = {'Status': 'FAILED', 'Data': {}, 'Reason': msg}
+        response_body = {"Status": "FAILED", "Data": {}, "Reason": msg}
```
demartinofra, line 20: I was keeping the formatting checks on this file disabled since we are importing it from https://github.com/aws-cloudformation/custom-resource-helper.

cli/tox.ini
```diff
@@ -128,6 +128,7 @@ deps =
     bandit
 commands =
     bandit -r \
+        --skip B105,B404 \
```
demartinofra, line 131: can you add a comment on why we disable these 2 rules globally?

cli/tox.ini
```diff
@@ -128,6 +128,7 @@ deps =
...
         setup.py \
```
demartinofra, line 132: can you use the code_dirs var?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+    rev: 20.8b1
+    hooks:
+      - id: black
+        args: ['-l 120']
+
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.0
+    hooks:
+      - id: bandit
+        args: ['-r', '--skip', 'B105,B404', '--exclude', 'tests,cli/tests']
```
demartinofra, line 46: I'm wondering if we can have the config in a global file that is shared also with tox (like we do for flake and isort)

demartinofra (commented):
Nice! Thanks for also adding a pre-commit config

.github/workflows/dep_check.yml
tilne: > I'm wondering if we should have these new workflows as new jobs of our CI workflow rather than creating new workflows.

Is there a downside to using new workflows?


> also I'm wondering if it makes sense to implement depcheck and semgrep as tox actions so that it's easier to run such linters locally as part of the tox -e code-linters

I opted not to do this because for [dependency check specifically](https://github.com/jeremylong/DependencyCheck) the installation is platform-specific.

.pre-commit-config.yaml
tilne: If flake8 isn't run from the CLI directory it doesn't read in its configuration in the same way. Paths are relative to the top-level of the repo so the nested shell handles that.

cli/src/pcluster/resources/custom_resources/custom_resources_code/crhelper/utils.py
tilne: Is it an issue to make the changes here? That enables it to play more nicely with pre-commit.

.pre-commit-config.yaml
demartinofra: it might be worth to drop a comment

cli/tox.ini
tilne: Yes. I'm adding inline comments to the bandit config file.

rexcsn (approved):

PR #2579, opened 2021-03-26
3.0: Create S3 bucket per region

yuleiwan:
* Define S3Bucket model for common bucket utilities and object key generation.
* Define S3BucketFactory to check the bucket existence and manage bucket creation, configuration.
* Add custom bucket schema to imagebuilder config file
* Complete S3 bucket per region logic in `pcluster create`, `pcluster delete`, `pcluster update`, `pcluster build-image` commands.

### DIFF
--- a/cli/src/common/boto3/common.py
+++ b/cli/src/common/boto3/common.py
@@ -24,6 +24,14 @@ class AWSClientError(Exception):
         if error_code:
             message += f" Error code: {error_code}"
         super().__init__(message)
+        self.error_code = error_code
+
+
+class ImageNotFoundError(AWSClientError):
+    """Error during describe image if image is not found."""
+
+    def __init__(self, function_name: str):
+        super().__init__(function_name=function_name, message="No image matching the search criteria found")
 
 
 class AWSExceptionHandler:
@@ -69,3 +77,10 @@ class Boto3Client(ABC):
         for page in paginator.paginate(**kwargs).result_key_iters():
             for result in page:
                 yield result
+
+
+class Boto3Resource(ABC):
+    """Abstract Boto3 resource."""
+
+    def __init__(self, resource_name: str):
+        self._resource = boto3.resource(resource_name)
--- a/cli/src/common/boto3/ec2.py
+++ b/cli/src/common/boto3/ec2.py
@@ -11,7 +11,7 @@
 from typing import List
 
 from common.aws.aws_resources import InstanceInfo, InstanceTypeInfo
-from common.boto3.common import AWSClientError, AWSExceptionHandler, Boto3Client
+from common.boto3.common import AWSClientError, AWSExceptionHandler, Boto3Client, ImageNotFoundError
 from pcluster import utils
 from pcluster.constants import PCLUSTER_IMAGE_NAME_TAG, SUPPORTED_ARCHITECTURES
 from pcluster.utils import Cache
@@ -99,19 +99,22 @@ class Ec2Client(Boto3Client):
         result = self._client.describe_images(ImageIds=ami_ids, Filters=filters, Owners=owners)
         if result.get("Images"):
             return result.get("Images")
-        raise AWSClientError(function_name="describe_images", message="No image matching the search criteria found")
+        raise ImageNotFoundError(function_name="describe_images")
 
     def image_exists(self, image_name):
         """Return a boolean describing whether or not an image with the given search criteria exists."""
+        try:
+            self.describe_image_by_name_tag(image_name)
+            return True
+        except ImageNotFoundError:
+            return False
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_image_by_name_tag(self, image_name):
+        """Return a dict of image info by searching image name tag as filter."""
         filters = [{"Name": "tag:" + PCLUSTER_IMAGE_NAME_TAG, "Values": [image_name]}]
         owners = ["self"]
-        try:
-            self.describe_images(ami_ids=[], filters=filters, owners=owners)
-            return True
-        except AWSClientError as e:
-            if "No image matching the search criteria found" in str(e):
-                return False
-            raise e
+        return self.describe_images(ami_ids=[], filters=filters, owners=owners)[0]
 
     @AWSExceptionHandler.handle_client_exception
     def describe_key_pair(self, key_name):
--- a/cli/src/common/boto3/s3.py
+++ b/cli/src/common/boto3/s3.py
@@ -80,6 +80,42 @@ class S3Client(Boto3Client):
                 error_code=client_error.response["Error"]["Code"],
             )
 
+    @AWSExceptionHandler.handle_client_exception
+    def create_bucket(self, bucket_name, region):
+        """Create S3 bucket."""
+        if region != "us-east-1":
+            self._client.create_bucket(Bucket=bucket_name, CreateBucketConfiguration={"LocationConstraint": region})
+        else:
+            self._client.create_bucket(Bucket=bucket_name)
+
+    @AWSExceptionHandler.handle_client_exception
+    def put_bucket_versioning(self, bucket_name, configuration):
+        """Set bucket versioning property."""
+        self._client.put_bucket_versioning(Bucket=bucket_name, VersioningConfiguration=configuration)
+
+    @AWSExceptionHandler.handle_client_exception
+    def put_bucket_encryption(self, bucket_name, configuration):
+        """Set bucket encryption property."""
+        self._client.put_bucket_encryption(
+            Bucket=bucket_name,
+            ServerSideEncryptionConfiguration=configuration,
+        )
+
+    @AWSExceptionHandler.handle_client_exception
+    def put_bucket_policy(self, bucket_name, policy):
+        """Set bucket policy property."""
+        self._client.put_bucket_policy(Bucket=bucket_name, Policy=policy)
+
+    @AWSExceptionHandler.handle_client_exception
+    def upload_fileobj(self, bucket_name, file_obj, key):
+        """Upload file-like object to S3 bucket."""
+        self._client.upload_fileobj(Fileobj=file_obj, Bucket=bucket_name, Key=key)
+
+    @AWSExceptionHandler.handle_client_exception
+    def upload_file(self, bucket_name, file_path, key):
+        """Upload file to S3 bucket."""
+        self._client.upload_file(Filename=file_path, Bucket=bucket_name, Key=key)
+
 
 def _process_s3_bucket_error(client_error, bucket_name):
     error_message = client_error.response.get("Error").get("Message")
--- a/cli/src/common/boto3/s3_resource.py
+++ b/cli/src/common/boto3/s3_resource.py
@@ -1,0 +1,45 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+from common.boto3.common import AWSExceptionHandler, Boto3Resource
+from pcluster.utils import Cache
+
+
+class S3Resource(Boto3Resource):
+    """S3 Boto3 resource."""
+
+    def __init__(self):
+        super().__init__("s3")
+
+    @AWSExceptionHandler.handle_client_exception
+    @Cache.cached
+    def get_bucket(self, bucket_name):
+        """Get bucket object by bucket name."""
+        return self._resource.Bucket(bucket_name)
+
+    @AWSExceptionHandler.handle_client_exception
+    def delete_object(self, bucket_name, prefix=None):
+        """Delete objects by filter."""
+        self.get_bucket(bucket_name).objects.filter(Prefix=prefix).delete()
+
+    @AWSExceptionHandler.handle_client_exception
+    def delete_all_objects(self, bucket_name):
+        """Delete all objects."""
+        self.get_bucket(bucket_name).objects.all().delete()
+
+    @AWSExceptionHandler.handle_client_exception
+    def delete_object_versions(self, bucket_name, prefix=None):
+        """Delete object versions by filter."""
+        self.get_bucket(bucket_name).object_versions.filter(Prefix=prefix).delete()
+
+    @AWSExceptionHandler.handle_client_exception
+    def delete_all_object_versions(self, bucket_name):
+        """Delete all object versions."""
+        self.get_bucket(bucket_name).object_versions.delete()
--- a/cli/src/common/boto3/sts.py
+++ b/cli/src/common/boto3/sts.py
@@ -1,0 +1,25 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+from common.boto3.common import AWSExceptionHandler, Boto3Client
+from pcluster.utils import Cache
+
+
+class StsClient(Boto3Client):
+    """STS Boto3 client."""
+
+    def __init__(self):
+        super().__init__("sts")
+
+    @AWSExceptionHandler.handle_client_exception
+    @Cache.cached
+    def get_account_id(self):
+        """Get account id by get_caller_identity."""
+        return self._client.get_caller_identity().get("Account")
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -62,6 +62,10 @@ CW_DASHBOARD_ENABLED_DEFAULT = True
 CW_LOGS_ENABLED_DEFAULT = True
 
 PCLUSTER_IMAGE_NAME_TAG = "parallelcluster:image_name"
+PCLUSTER_S3_IMAGE_DIR_TAG = "parallelcluster:s3_image_dir"
+PCLUSTER_S3_BUCKET_TAG = "parallelcluster:s3_bucket"
+
+PCLUSTER_S3_BUCKET_VERSION = "v1"
 
 SUPPORTED_REGIONS = [
     "af-south-1",
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -12,7 +12,6 @@
 # This module contains all the classes representing the Resources objects.
 # These objects are obtained from the configuration file through a conversion based on the Schema classes.
 #
-import json
 import logging
 import time
 from copy import deepcopy
@@ -29,17 +28,11 @@ from common.boto3.common import AWSClientError
 from pcluster.cli_commands.compute_fleet_status_manager import ComputeFleetStatus, ComputeFleetStatusManager
 from pcluster.config.config_patch import ConfigPatch
 from pcluster.constants import OS_MAPPING, PCLUSTER_STACK_PREFIX
-from pcluster.models.cluster_config import BaseClusterConfig, ClusterBucket, SlurmScheduling, Tag
+from pcluster.models.cluster_config import BaseClusterConfig, SlurmScheduling, Tag
+from pcluster.models.common import S3Bucket, S3BucketFactory, S3FileFormat
 from pcluster.schemas.cluster_schema import ClusterSchema
 from pcluster.templates.cdk_builder import CDKTemplateBuilder
-from pcluster.utils import (
-    check_s3_bucket_exists,
-    create_s3_bucket,
-    generate_random_name_with_prefix,
-    get_installed_version,
-    grouper,
-    upload_resources_artifacts,
-)
+from pcluster.utils import generate_random_name_with_prefix, get_installed_version, get_region, grouper
 from pcluster.validators.cluster_validators import ClusterNameValidator
 from pcluster.validators.common import FailureLevel, ValidationResult
 
@@ -126,10 +119,10 @@ class ClusterStack(StackInfo):
         """Delete stack."""
         AWSApi.instance().cfn.delete_stack(self.name)
 
-    def update_template(self, template):
+    def update_template(self, template_url):
         """Update template of the running stack according to updated template."""
         try:
-            AWSApi.instance().cfn.update_stack(self.name, template, self._params)
+            AWSApi.instance().cfn.update_stack_from_url(self.name, template_url)
             self._wait_for_update()
         except AWSClientError as e:
             if "no updates are to be performed" in str(e).lower():
@@ -160,9 +153,20 @@ class Cluster:
         self.name = name
         self.__source_config = config
         self.__stack = stack
-        self.bucket = None
+        self.__bucket = None
         self.template_body = None
         self.__config = None
+        self._s3_artifacts_dict = {
+            "root_directory": "parallelcluster",
+            "root_cluster_directory": "clusters",
+            "source_config_name": "cluster-config-original.yaml",
+            "config_name": "cluster-config.yaml",
+            "template_name": "aws-parallelcluster.cfn.yaml",
+            "instance_types_data_name": "instance-types-data.json",
+            "custom_artifacts_name": "artifacts.zip",
+            "scheduler_resources_name": "scheduler_resources.zip",
+        }
+        self.__s3_artifact_dir = None
 
         self.__has_running_capacity = None
         self.__running_capacity = None
@@ -186,11 +190,44 @@ class Cluster:
             self.__source_config = self._get_cluster_config_dict()
         return self.__source_config
 
+    @property
+    # pylint: disable=R0801
+    def s3_artifacts_dir(self):
+        """Get s3 artifacts dir."""
+        if self.__s3_artifact_dir is None:
+            self._get_artifact_dir()
+        return self.__s3_artifact_dir
+
+    def _get_artifact_dir(self):
+        """Get artifact directory in S3 bucket by stack output."""
+        try:
+            self.__s3_artifact_dir = self.stack.s3_artifact_directory
+            if self.__s3_artifact_dir is None:
+                raise AWSClientError(
+                    function_name="_get_artifact_dir", message="No artifact dir found in cluster stack output."
+                )
+        except AWSClientError as e:
+            LOGGER.error("No artifact dir found in cluster stack output.")
+            raise ClusterActionError(f"Unable to find artifact dir in cluster stack {self.stack_name} output. {e}")
+
+    def _generate_artifact_dir(self):
+        """
+        Generate artifact directory in S3 bucket.
+
+        cluster artifact dir is generated before cfn stack creation and only generate once.
+        artifact_directory: e.g. parallelcluster/clusters/{cluster_name}-jfr4odbeonwb1w5k
+        """
+        service_directory = generate_random_name_with_prefix(self.name)
+        self.__s3_artifact_dir = "/".join(
+            [
+                self._s3_artifacts_dict.get("root_directory"),
+                self._s3_artifacts_dict.get("root_cluster_directory"),
+                service_directory,
+            ]
+        )
+
     def _get_cluster_config_dict(self):
         """Retrieve cluster config content."""
-        if not self.bucket:
-            raise ClusterActionError("Unable to retrieve S3 Bucket name")
-
         table_name = self.stack.name
         config_version = None
         try:
@@ -202,16 +239,12 @@ class Cluster:
             pass
 
         try:
-            s3_object = AWSApi.instance().s3.get_object(
-                bucket_name=self.bucket.name,
-                key=f"{self.bucket.artifact_directory}/configs/cluster-config.yaml",
-                version_id=config_version,
+            return self.bucket.get_config(
+                version_id=config_version, config_name=self._s3_artifacts_dict.get("config_name")
             )
-            config_content = s3_object["Body"].read().decode("utf-8")
-            return yaml.safe_load(config_content)
         except Exception as e:
             raise ClusterActionError(
-                f"Unable to load configuration from bucket '{self.bucket.name}/{self.bucket.artifact_directory}'.\n{e}"
+                f"Unable to load configuration from bucket '{self.bucket.name}/{self.s3_artifacts_dir}'.\n{e}"
             )
 
     @property
@@ -238,6 +271,34 @@ class Cluster:
         """Return the cluster status."""
         return self.stack.status
 
+    @property
+    # pylint: disable=R0801
+    def bucket(self):
+        """Return a bucket configuration."""
+        if self.__bucket:
+            return self.__bucket
+
+        if self.__source_config:
+            # get custom_s3_bucket in create command
+            custom_bucket_name = self.config.custom_s3_bucket
+        else:
+            # get custom_s3_bucket in delete, update commands
+            custom_bucket_name = self.stack.s3_bucket_name
+            if custom_bucket_name == S3Bucket.get_bucket_name(AWSApi.instance().sts.get_account_id(), get_region()):
+                custom_bucket_name = None
+
+        try:
+            self.__bucket = S3BucketFactory.init_s3_bucket(
+                service_name=self.name,
+                stack_name=self.stack_name,
+                custom_s3_bucket=custom_bucket_name,
+                artifact_directory=self.s3_artifacts_dir,
+            )
+        except AWSClientError as e:
+            raise ClusterActionError(f"Unable to initialize s3 bucket. {e}")
+
+        return self.__bucket
+
     def create(
         self,
         disable_rollback: bool = False,
@@ -252,9 +313,7 @@ class Cluster:
                 raise ClusterActionError(f"Cluster {self.name} already exists")
             self.config = self._validate_and_parse_config(suppress_validators, validation_failure_level)
 
-            # Create bucket if needed
-            self.bucket = self._setup_cluster_bucket()
-
+            self._generate_artifact_dir()
             self._add_version_tag()
             self._upload_config()
 
@@ -270,7 +329,9 @@ class Cluster:
             LOGGER.info("Creating stack named: %s", self.stack_name)
             creation_result = AWSApi.instance().cfn.create_stack_from_url(
                 stack_name=self.stack_name,
-                template_url=self.template_url,
+                template_url=self.bucket.get_cfn_template_url(
+                    template_name=self._s3_artifacts_dict.get("template_name")
+                ),
                 disable_rollback=disable_rollback,
                 tags=self._get_cfn_tags(),
             )
@@ -280,12 +341,10 @@ class Cluster:
             LOGGER.info("Status: %s", self.stack.status)
 
         except Exception as e:
-            if not creation_result and self.bucket:
+            if not creation_result:
                 # Cleanup S3 artifacts if stack is not created yet
-                self.bucket.delete()
-            if isinstance(e, ClusterActionError):
-                raise e
-            raise ClusterActionError(str(e))
+                self.bucket.delete_s3_artifacts()
+            raise ClusterActionError(f"Cluster creation failed.\n{e}")
 
     def _validate_and_parse_config(self, suppress_validators, validation_failure_level, config_dict=None):
         """
@@ -316,63 +375,23 @@ class Cluster:
 
         return config
 
-    def _setup_cluster_bucket(self) -> ClusterBucket:
-        """
-        Create pcluster bucket, if needed.
-
-        If no bucket specified, create a bucket associated to the given stack.
-        Created bucket needs to be removed on cluster deletion.
-        """
-        if self.config.custom_s3_bucket:
-            # Use user-provided bucket
-            # Do not remove this bucket on deletion, but cleanup artifact directory
-            name = self.config.custom_s3_bucket
-            remove_on_deletion = False
-            try:
-                check_s3_bucket_exists(name)
-            except Exception as e:
-                LOGGER.error("Unable to access config-specified S3 bucket %s: %s", name, e)
-                raise e
-        else:
-            # Create 1 bucket per cluster named "parallelcluster-{random_string}" if bucket is not provided
-            # This bucket needs to be removed on cluster deletion
-            name = generate_random_name_with_prefix("parallelcluster")
-            # self.custom_s3_bucket = name
-            remove_on_deletion = True
-            LOGGER.debug("Creating S3 bucket for cluster resources, named %s", name)
-            try:
-                create_s3_bucket(name)
-            except Exception as e:
-                LOGGER.error("Unable to create S3 bucket %s.", name)
-                raise e
-
-        # Use "{stack_name}-{random_string}" as directory in bucket
-        artifact_directory = generate_random_name_with_prefix(self.stack_name)
-        return ClusterBucket(name, artifact_directory, remove_on_deletion)
-
     def _upload_config(self):
         """Upload source config and save config version."""
-        if not self.bucket:
-            ClusterActionError("S3 bucket must be created before uploading artifacts.")
-
         try:
-            # Upload original config
-            if self.config.source_config:
-                AWSApi.instance().s3.put_object(
-                    bucket_name=self.bucket.name,
-                    body=yaml.dump(self.config.source_config),
-                    key=self._get_original_config_key(),
-                )
             # Upload config with default values and sections
             if self.config:
-                config_copy = deepcopy(self.config)
-                result = AWSApi.instance().s3.put_object(
-                    bucket_name=self.bucket.name,
-                    body=yaml.dump(ClusterSchema().dump(config_copy)),
-                    key=self._get_config_key(),
+                result = self.bucket.upload_config(
+                    config=ClusterSchema().dump(deepcopy(self.config)),
+                    config_name=self._s3_artifacts_dict.get("config_name"),
                 )
+
                 # config version will be stored in DB by the cookbook at the first update
                 self.config.config_version = result.get("VersionId")
+
+                # Upload original config
+                self.bucket.upload_config(
+                    config=self.config.source_config, config_name=self._s3_artifacts_dict.get("source_config_name")
+                )
 
         except Exception as e:
             raise ClusterActionError(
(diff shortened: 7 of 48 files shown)


cli/src/pcluster/constants.py
```diff
@@ -58,3 +58,7 @@
 CW_LOGS_ENABLED_DEFAULT = True
 
 PCLUSTER_IMAGE_NAME_TAG = "parallelcluster:image_name"
+PCLUSTER_S3_IMAGE_DIR_TAG = "parallelcluster:image_dir"
+PCLUSTER_BUCKET_TAG = "parallelcluster:bucket"
```
demartinofra, line 62: s3_bucket?

cli/src/pcluster/constants.py
```diff
@@ -58,3 +58,7 @@
 CW_LOGS_ENABLED_DEFAULT = True
 
 PCLUSTER_IMAGE_NAME_TAG = "parallelcluster:image_name"
+PCLUSTER_S3_IMAGE_DIR_TAG = "parallelcluster:image_dir"
```
demartinofra, line 61: s3_image_dir?

cli/src/common/boto3/ec2.py
```diff
@@ -98,6 +98,15 @@ def image_exists(self, image_name):
                 return False
             raise e
 
+    def describe_image_by_tag(self, image_name):
```
demartinofra, line 101: describe_image_by_name_tag?

does it make sense to cache this call?

cli/src/common/boto3/ec2.py
```diff
@@ -98,6 +98,15 @@ def image_exists(self, image_name):
...
+        """Return a dict of image info by searching image name tag as filter."""
+        filters = [{"Name": "tag:" + PCLUSTER_IMAGE_NAME_TAG, "Values": [image_name]}]
+        owners = ["self"]
+        try:
+            return self.describe_images(ami_ids=[], filters=filters, owners=owners)[0]
+        except AWSClientError:
+            return None
```
demartinofra, line 108: are we ok to handle all the errors the same way: e.g. unexisting ami, missing IAM permissions, throttling?

demartinofra: I think there is a similar issue with the image_exists method above

cli/src/common/boto3/sts.py
```diff
@@ -0,0 +1,13 @@
+from common.boto3.common import AWSExceptionHandler, Boto3Client
+
+
+class StsClient(Boto3Client):
+    """STS Boto3 client."""
+
+    def __init__(self):
+        super().__init__("sts")
+
+    @AWSExceptionHandler.handle_client_exception
```
demartinofra, line 10: can this be cached?

cli/src/common/utils.py
```diff
@@ -69,3 +70,12 @@ def parse_bucket_url(url):
         raise URLError("Invalid s3 url: {0}".format(url))
 
     return {"bucket_name": bucket_name, "object_key": object_key, "object_name": object_name}
+
+
+def generate_s3_bucket_hash_suffix(account_id, region):
```
demartinofra, line 75: let's try to minimize the usage of the common/utils module. probably this can even be a private function for now since only used in a single place

cli/src/pcluster/constants.py
yuleiwan: Done

cli/src/pcluster/constants.py
yuleiwan: Done

cli/src/common/boto3/ec2.py
yuleiwan: Changed the function name, but I am concern about caching this call, because I am not sure we may have a case that will modify the image after first call. For example, the image was described before deletion, and if we want to get updated information of image after deletion, we can't do that if we cache the image info.

cli/src/common/boto3/ec2.py
yuleiwan: Updated, the idea here is we always throw an AWSClientError for describe_image or describe_stack function if boto3 call fails, and catch this AWSClientError in logic function which calls these boto3 methods, and check the error message. If the error message contains `stack id doesn't exist` or `No image matching the search criteria found`, which means it is the case of unexisting ami/stack. Otherwise, we raise the error for other errors like missing IAM permissions, throttling.

In image_exists and stack_exists functions, we only return False if the case is unexisting ami/stack.

cli/src/common/boto3/sts.py
yuleiwan: yes, it is. Done

cli/src/common/utils.py
yuleiwan: move into S3Bucket class as a static method.

cli/src/pcluster/schemas/imagebuilder_schema.py
```diff
@@ -174,8 +174,11 @@ class ImageBuilderSchema(BaseSchema):
...
     dev_settings = fields.Nested(ImagebuilderDevSettingsSchema)
+    cluster_s3_bucket = fields.Str()
 
-    @post_load()
-    def make_resource(self, data, **kwargs):
+    @post_load(pass_original=True)
+    def make_resource(self, data, original_data, **kwargs):
         """Generate resource."""
-        return ImageBuilderConfig(**data)
+        imagebuilder = ImageBuilderConfig(**data)
```
demartinofra, line 182: is there a reason why we are not passing the source_config in the constructor?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
 
     def __init__(
         self,
-        name: str,
-        artifact_directory: str,
-        remove_on_deletion: bool,
+        cluster_name: str,
+        stack_name: str,
+        name: str = None,
+        artifact_remove_on_deletion: bool = True,
```
demartinofra, line 1366: nit: cleanup_on_deletion?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+        self.cluster_name = cluster_name
+        self.stack_name = stack_name
+        self.root_cluster_directory = "clusters"
+        self.source_config_name = "cluster-config-original.yaml"
+        self.config_name = "cluster-config.yaml"
+        self.template_name = "aws-parallelcluster.cfn.yaml"
+        self.artifacts_name = "artifacts.zip"
+        self.instance_types_data_name = "instance-types-data.json"
+        self.cluster_directory = None
+        self.full_cluster_dir = None
```
demartinofra, line 1378: are these all supposed to be "public" properties the caller can access? it might be good to encapsulate the structure of the bucket in this class and for example make dir and file names some local constants that can also be defined at the cluster level with a leading underscore to mark them as "private"

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+                stack_data = AWSApi.instance().cfn.describe_stack(self.stack_name)
+                self.full_cluster_dir = next(
+                    (
+                        out["OutputValue"]
+                        for out in stack_data.get("Outputs", [])
+                        if out["OutputKey"] == "ArtifactS3RootDirectory"
+                    ),
+                    None,
+                )
+                self.cluster_directory = self.full_cluster_dir.split("/")[-1]
```
demartinofra, line 1398: A method named get_* should not have a side effect of re-initializing the class attributes. can the caller do this?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+        self.template_name = "aws-parallelcluster.cfn.yaml"
+        self.artifacts_name = "artifacts.zip"
+        self.instance_types_data_name = "instance-types-data.json"
+        self.cluster_directory = None
+        self.full_cluster_dir = None
+        self._get_cluster_dir()
+
+    def _get_object_key(self, object_type, object_name):
+        """Get object key of an artifact."""
+        return os_lib.path.join(self._get_cluster_dir(), object_type, object_name)
```
demartinofra, line 1383: Can we have this function access a local cluster dir parameter rather than calling the _get_cluster_dir, otherwise it seems that it always tries to fetch from cfn unless you dive deep and realize there is the `self._generate_cluster_dir()` which is eventually setting the full_cluster_dir. it took me some time to get there. Can we do something to make this sequence more explicit?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+                    None,
+                )
+                self.cluster_directory = self.full_cluster_dir.split("/")[-1]
+                return self.full_cluster_dir
+            except (AWSClientError, KeyError) as e:
+                if f"Stack with id {self.stack_name} does not exist" in str(e):
+                    self._generate_cluster_dir()
+                else:
+                    LOGGER.error("No cluster dir found in stack output.")
+                    raise
```
demartinofra, line 1405: This causes a KeyError to be propagated that is very hard to debug or handle in the caller. Shall we remove the KeyError from the except and handle this case with some validation + raising the proper exception if needed?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+    def _generate_cluster_dir(self):
+        """
+        Generate full cluster directory path in S3 bucket.
+
+        Cluster dir is generated before cfn stack creation and only generate once.
+        cluster_directory: e.g.{cluster_name}-jfr4odbeonwb1w5k
+        full_cluster_dir: e.g. parallelcluster/clusters/{cluster_name}-jfr4odbeonwb1w5k
+        """
+        self.cluster_directory = generate_random_name_with_prefix(self.cluster_name)
+        self.full_cluster_dir = os_lib.path.join(
```
demartinofra, line 1418: does this work on windows where the path separator is not /?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+
+    def _get_object_key(self, object_type, object_name):
+        """Get object key of an artifact."""
+        return os_lib.path.join(self._get_cluster_dir(), object_type, object_name)
+
+    def _get_cluster_dir(self):
+        """Get cluster directory in S3 bucket by stack output."""
+        if self.full_cluster_dir is None:
+            try:
+                stack_data = AWSApi.instance().cfn.describe_stack(self.stack_name)
```
demartinofra, line 1389: Can you review the boto3/AWSApi calls added as part of this PR and see if all of them are necessary or some optimizations can be done? 
Also is there a way to avoid calling the describe_stack if we already know the cluster does not exist because we are in a create action?

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
-            self.remove_on_deletion,
+            self.cluster_name,
+            self.artifact_remove_on_deletion,
         )
-        if self.artifact_directory:
-            delete_s3_artifacts(self.name, self.artifact_directory)
-        if self.remove_on_deletion:
-            delete_s3_bucket(self.name)
+        if self.full_cluster_dir and self.artifact_remove_on_deletion:
+            delete_s3_artifacts(self.name, self.full_cluster_dir)
```
demartinofra, line 1456: shall we move this function in here from the utils class?

cli/src/pcluster/models/imagebuilder_config.py
```diff
@@ -185,3 +193,85 @@ def _set_default(self, dev_settings: ImagebuilderDevSettings):
         for key, value in self.__dict__.items():
             if not key.startswith("_") and key not in self._cfncluster_attributes:
                 self._cfncluster_attributes.update({key: value})
+
+
+class ImageBuilderBucket(S3Bucket):
```
demartinofra, line 198: do we really need two different classes to manage the cluster and the image builder bucket? from a quick look it looks like they perform the same operations but on different directories. probably I'm oversimplifying but aren't these classes different only on the value of the attributes that can be assigned when instantiated?

cli/src/pcluster/utils.py
```diff
@@ -187,7 +172,7 @@ def _add_file_to_zip(zip_file, path, arcname):
         zip_file.writestr(zinfo, input_file.read())
 
 
-def _zip_dir(path):
+def zip_dir(path):
```
demartinofra, line 175: let's move this if needed only in the new S3 class

cli/tests/pcluster/models/cluster_dummy_model.py
```diff
@@ -237,3 +251,29 @@ def dummy_efs(mount_dir, name="name", file_system_id=None):
         provisioned_throughput=500,
         file_system_id=file_system_id,
     )
+
+
+class DummyClusterBucket(ClusterBucket):
```
demartinofra, line 256: can we avoid creating Dummy subclasses? I'm just afraid that we end up writing unit tests that test Dummy code. Is it possible to instantiate a ClusterBucket with the necessary mocking?

cli/tests/pcluster/models/test_cluster_stack.py
```diff
@@ -110,24 +111,30 @@ def test_wait_for_stack_update(mocker, stack_statuses):
 def test_update_stack_template(mocker, error_message):
```
demartinofra, line 111: Adding this here but it is more of a general comment not only related to your changes:
some of these unit tests are getting very complex and it's hard to understand what we are testing and what not. also if we write unit tests that are more complex than the code they are testing we end-up in a situation where we need to test the tests. We are also doing a lot of mocking which probably is necessary but at the same time it makes very hard to understand what is the real code under test. Usually when a lot of mocking is required this is a symptom that the function under test is complex. Also we could consider having multiple simple test function if a single parametrized one risks to become too complex.

demartinofra (commented):
Overall looks good.

Let's try to reduce duplication and see if we can handle the S3 logic in a unified class and avoid duplicating all components for Cluster and ImageBuilder logic. What if we need to add a new command that requires this bucket?We do not want to have to copy paste the entire logic but ideally instantiate a single class that manages the bucket.

I quickly scanned through unit tests and to be honest I had a hard time understanding what we are testing and what not, what is mocked and what is real code. Do you have any idea on how we could simplify them?

cli/src/pcluster/models/cluster.py
```diff
@@ -345,63 +356,25 @@ def _validate_and_parse_config(self, suppress_validators, validation_failure_lev
...
-                    key=self._get_config_key(),
+                result = self.bucket.upload_config(
+                    config=ClusterSchema().dump(self.config), config_name=self.bucket.config_name
                 )
+
                 # config version will be stored in DB by the cookbook at the first update
                 self.config.config_version = result.get("VersionId")
 
+            # Upload original config
+            if self.config.source_config:
```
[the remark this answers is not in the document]
demartinofra: yes I'd avoid all the extra conditional checks that are not really needed so that we simplify the code. Otherwise we end up having this if conditionals everywhere in the code. It is good to use a defensive approach and check preconditions on methods that represent a library interface or functions that are used in modules we have poor control over but in the other cases we can avoid it. In this case if source_config is None it is more of a bug I'd say.

cli/src/pcluster/schemas/imagebuilder_schema.py
demartinofra: But we could do ImageBuilderConfig(source_config=original_data, **data)

cli/src/pcluster/models/cluster_config.py
demartinofra: I'm afraid that pat.join uses the system separator that it's not always "/" while here for the S3 path we always want to use "/". I'd rather go for a "/".join

cli/src/pcluster/models/imagebuilder_config.py
demartinofra: but you could use a factory that is responsible to instantiate such classes in the proper way. Also _get_cluster_dir and _get_image_dir in the end are just returning the base path where to upload the files and they don't have to contain cluster or image in the name.
my concern is that there is some good amount of duplicated code. At the very least if we don't want to get rid of this separation let's try to generalize all the functinalities and put them in the base class and then have the subclasses only assign parameters about the various dirs.

In the end we want to have a generic object that manages the S3 bucket and exposes a set of operations to upload different types of files

cli/src/pcluster/models/cluster_config.py
demartinofra: It can also be done in the __init__ method. if the class is responsible to instantiate the bucket when initialized then that is the right place to do, unless of course we need to do some lazy initialization.

In the init rather than just calling `self._get_cluster_dir()` you can do something like `self.cluster_directory = self._get_cluster_dir()` and in _get_object_key you can simply use self.cluster_directory.

cli/src/pcluster/models/cluster_config.py
demartinofra: btw I think there are other places where we do the path.join to build the S3 path. Let's check all of them

cli/src/pcluster/models/cluster_config.py
```diff
@@ -1350,29 +1352,105 @@ def _validate(self):
...
+        Cluster dir is generated before cfn stack creation and only generate once.
+        cluster_directory: e.g.{cluster_name}-jfr4odbeonwb1w5k
+        full_cluster_dir: e.g. parallelcluster/clusters/{cluster_name}-jfr4odbeonwb1w5k
+        """
+        self.cluster_directory = generate_random_name_with_prefix(self.cluster_name)
+        self.full_cluster_dir = os_lib.path.join(
+            self.root_directory, self.root_cluster_directory, self.cluster_directory
+        )
+
+    def upload_cutsom_resources(self, resource_dir):
```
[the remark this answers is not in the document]
yuleiwan: rename to upload_resources

cli/src/pcluster/models/cluster_config.py
demartinofra: I meant at the class level not cluster, sorry :)

cli/tests/pcluster/models/cluster_dummy_model.py
yuleiwan: remove subclass.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+                        bucket.create_bucket()
+                    else:
+                        raise
+            except Exception as e:
+                LOGGER.error("Unable to create S3 bucket %s.", bucket.name)
+                raise AWSClientError("create_bucket", str(e))
+        return bucket
+
+    @staticmethod
+    def get_cluster_bucket(service_name, stack_name, bucket_name):
```
demartinofra, line 647: I'm not sure I understand why we need these extra get_*_bucket methods. I would avoid the dependency on the S3BucketFactory from the ClusterStack. isn't the bucket already part of the Cluster object?

I believe the error is that we are misusing the ClusterStack class (not as part of this PR) and this is causing this unneeded coupling. ClusterStack should not contain the logic to fetch the cluster configuration nor the logic to handle the keep logs on deletion but needs to simply reflect the CloudFormation stack status with a series of properties to retrieve the stack info and outputs. On top of that it can expose simple create, update  and delete operations.
The retrieval of the config from the bucket and the logic to preserve logs need to be moved into the Cluster class.
I know it's not directly part of this PR but I'm afraid that the best path forward is to fix that and then apply your changes.

cli/src/pcluster/models/cluster.py
```diff
@@ -186,7 +179,13 @@ def _persist_stack_resources(self, keys):
     def _update_template(self, template):
         """Update template of the running stack according to updated template."""
         try:
-            AWSApi.instance().cfn.update_stack(self.name, template, self._params)
+            bucket = S3BucketFactory.get_cluster_bucket(
```
enrico-usai, line 182: After merging #2605 let's move this code specific for the bucket management in the Cluster, before calling the `stack.update_template()`

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+                    None,
+                )
+                return self.__artifact_directory
+            except AWSClientError as e:
+                if "No image matching the search criteria found" not in str(e):
+                    LOGGER.error("No artifact dir found in image tag.")
+                    raise
+
+            try:
+                stack_info = AWSApi.instance().cfn.describe_stack(self._stack_name)
```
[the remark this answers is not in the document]
yuleiwan: If the image does not exist we won't raise, previous step only raises other cases.
```
  except AWSClientError as e:
      if "No image matching the search criteria found" not in str(e):
          LOGGER.error("No artifact dir found in image tag.")
          raise
```

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+            configuration={"Rules": [{"ApplyServerSideEncryptionByDefault": {"SSEAlgorithm": "AES256"}}]},
+        )
+        deny_http_policy = (
+            '{{"Id":"DenyHTTP","Version":"2012-10-17","Statement":[{{"Sid":"AllowSSLRequestsOnly","Action":"s3:*",'
+            '"Effect":"Deny","Resource":["arn:{partition}:s3:::{bucket_name}","arn:{partition}:s3:::{bucket_name}/*"],'
+            '"Condition":{{"Bool":{{"aws:SecureTransport":"false"}}}},"Principal":"*"}}]}}'
+        ).format(bucket_name=self.name, partition=self.partition)
+        AWSApi.instance().s3.put_bucket_policy(bucket_name=self.name, policy=deny_http_policy)
+
+    def _upload_file(self, content, file_name, file_type, format="yaml"):
```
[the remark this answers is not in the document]
yuleiwan: forgot, I will add.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+        if custom_s3_bucket:
+            bucket = S3Bucket(
+                service_name=service_name,
+                name=custom_s3_bucket,
+                bucket_type=bucket_type,
+                stack_name=stack_name,
+                is_custom_bucket=True,
+            )
+            try:
+                bucket.check_bucket_exists()
```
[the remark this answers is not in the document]
yuleiwan: yes, we can reuse S3BucketValidators for custom bucket to default bucket for config validation.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+                    (
+                        out["OutputValue"]
+                        for out in stack_data.get("Outputs", [])
+                        if out["OutputKey"] == "ArtifactS3RootDirectory"
+                    ),
+                    None,
+                )
+                return self.__artifact_directory
+            except AWSClientError:
+                LOGGER.error("No artifact dir found in cluster stack output.")
```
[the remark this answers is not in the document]
yuleiwan: describe_stack will throw ClientError is stack doesn't exist, describe_images will return an empty list if image doesn't exist.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+            try:
+                # Get artifact directory by image tag or imagebuilder stack tag
+                image_info = AWSApi.instance().ec2.describe_image_by_name_tag(self._service_name)
+                self.__artifact_directory = next(
+                    (tag["Value"] for tag in image_info.get("Tags", []) if tag["Key"] == PCLUSTER_S3_IMAGE_DIR_TAG),
+                    None,
+                )
+                return self.__artifact_directory
+            except AWSClientError as e:
+                if "No image matching the search criteria found" not in str(e):
```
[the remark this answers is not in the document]
yuleiwan: good idea

cli/src/common/boto3/s3_resource.py
```diff
@@ -0,0 +1,50 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+from common.boto3.common import AWSExceptionHandler, Boto3Resource
+from pcluster.utils import Cache
+
+
+class S3Resource(Boto3Resource):
```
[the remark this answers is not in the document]
yuleiwan: I'd keep the s3_resource in a class, because some of function names in s3 resource and s3 client are similiar, it may cause confusion.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
         attribute_json = {"cfncluster": self._cfncluster_attributes}
         attribute_json.update(self._extra_attributes)
         return json.dumps(attribute_json, sort_keys=True)
+
+
+class S3File(Enum):
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+        self._cleanup_on_deletion = cleanup_on_deletion
+        self._root_directory = "parallelcluster"
+        self._custom_artifacts_name = "artifacts.zip"
+        self._is_custom_bucket = is_custom_bucket
+        self.__partition = None
+        self.__region = None
+        self.__account_id = None
+        self.__artifact_directory = None
+        self.__name = name
+        if self._bucket_type == BucketType.CLUSTER:
```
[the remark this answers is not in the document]
yuleiwan: pass artifact_directory to S3Bucket class, so S3Bucket object needn't be intialized in different way by bucket type.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+        stack_name: str,
+        cleanup_on_deletion: bool = True,
+        name: str = None,
+        is_custom_bucket: bool = False,
+    ):
+        super().__init__()
+        self._bucket_type = bucket_type
+        self._service_name = service_name
+        self._stack_name = stack_name
+        self._cleanup_on_deletion = cleanup_on_deletion
```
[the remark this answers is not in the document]
yuleiwan: I put this parameter as input argument with default value because if you want to add a new feature like "keep artifacts", it will be easy.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+            AWSApi.instance().s3.create_bucket(bucket_name=self.name)
+        try:
+            self._configure_s3_bucket()
+        except AWSClientError as e:
+            if "NoSuchBucket" not in str(e):
+                LOGGER.error("Failed when configuring cluster S3 bucket.")
+                self.delete_bucket()
+                raise
+
+    def delete_bucket(self):
```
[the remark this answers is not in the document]
yuleiwan: addressed in design doc.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+                )
+                raise
+
+    # --------------------------------------- S3 objects utils --------------------------------------- #
+
+    def get_object_key(self, object_type, object_name):
+        """Get object key of an artifact."""
+        return "/".join([self.artifact_directory, object_type, object_name])
+
+    def generate_artifact_dir(self):
```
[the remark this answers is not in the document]
yuleiwan: I think generate_artifact_dir can't be called with constructor, because it calls random string generation function, this function should only be called in create command. Otherwise, we will generate different artifacts_dir for same image/cluster.

The logic here is:
If the consumer is pcluster command or pcluster build-image, call generate_artifact_dir method, and assign to the parameter artifact_directory.
If the consumer is other commands, call get_artifact_dir to retrieve the artifact_directory from stack or image, and assign to the parameter artifact_directory.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+
+    def get_config_url(self, config_name):
+        """Get config file http url from S3 bucket."""
+        return self._get_file_url(file_name=config_name, file_type=S3File.CONFIGS.value)
+
+    def get_cfn_template(self, template_name, version_id=None):
+        """Get cfn template from S3 bucket."""
+        return self._get_file(file_type=S3File.TEMPLATES.value, file_name=template_name, version_id=version_id)
+
+    def get_cfn_template_url(self, template_name):
```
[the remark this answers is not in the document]
yuleiwan: moved all artifacts_file_name to cluster/imagebuilder class

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+
+    def _generate_s3_bucket_hash_suffix(self):
+        """
+        Generate 16 characters hash suffix for ParallelCluster s3 bucket.
+
+        :return: 16 chars string e.g. 2238a84ac8a74529
+        """
+        return hashlib.sha256((self.account_id + self.region).encode()).hexdigest()[0:16]
+
+    def _get_artifact_dir(self):
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/common.py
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/common.py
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/common.py
yuleiwan: remove

cli/src/pcluster/models/cluster.py
yuleiwan: done

cli/src/common/boto3/s3.py
```diff
@@ -80,6 +80,42 @@ def get_bucket_region(self, bucket_name):
...
+
+    @AWSExceptionHandler.handle_client_exception
+    def upload_fileobj(self, bucket_name, file_obj, key):
+        """Upload file-like object to S3 bucket."""
+        return self._client.upload_fileobj(Fileobj=file_obj, Bucket=bucket_name, Key=key)
+
+    @AWSExceptionHandler.handle_client_exception
+    def upload_file(self, bucket_name, file_path, key):
+        """Upload file to S3 bucket."""
+        return self._client.upload_fileobj(Filename=file_path, Bucket=bucket_name, Key=key)
```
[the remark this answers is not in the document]
yuleiwan: remove return

cli/src/common/boto3/s3.py
```diff
@@ -80,6 +80,42 @@ def get_bucket_region(self, bucket_name):
...
+
+    @AWSExceptionHandler.handle_client_exception
+    def put_bucket_policy(self, bucket_name, policy):
+        """Set bucket policy property."""
+        self._client.put_bucket_policy(Bucket=bucket_name, Policy=policy)
+
+    @AWSExceptionHandler.handle_client_exception
+    def upload_fileobj(self, bucket_name, file_obj, key):
+        """Upload file-like object to S3 bucket."""
+        return self._client.upload_fileobj(Fileobj=file_obj, Bucket=bucket_name, Key=key)
```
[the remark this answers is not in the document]
yuleiwan: remove return

cli/src/pcluster/models/common.py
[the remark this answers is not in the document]
demartinofra: yes but in the error message we write: "No artifact dir found in cluster stack output." while we get there only if the cluster is not found. In case the value is not found in the Outputs the logic returns None

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +252,413 @@ def dump_json(self):
...
+        """Get artifact directory in S3 bucket by stack output or stack/image tag."""
+        if self._bucket_type == BucketType.CLUSTER:
+            try:
+                # Get artifact directory by cluster stack output
+                stack_data = AWSApi.instance().cfn.describe_stack(self._stack_name)
+                self.__artifact_directory = next(
```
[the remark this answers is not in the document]
enrico-usai: I'd like to get rid of get_stack_output_value and other similar utilities.
You can instead return a StackInfo from describe_stack and use StackInfo.get_ouput method.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -146,6 +157,91 @@ def config(self):
...
+        If image and stack don't exist, get custom bucket name from config.
+        """
+        try:
+            image_info = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
+            custom_bucket_name = search_tag(image_info, PCLUSTER_S3_BUCKET_TAG)
+            if custom_bucket_name == S3Bucket.get_bucket_name(AWSApi.instance().sts.get_account_id(), get_region()):
+                custom_bucket_name = None
+        except AWSClientError as e:
+            if not isinstance(e, ImageNotFoundError):
+                raise ImageBuilderActionError(f"Unable to get bucket name from image {self.image_name} tag. {e}")
```
demartinofra, line 185: nit: S3 bucket

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -146,6 +157,91 @@ def config(self):
...
+        if self.__bucket is None:
+            self.__bucket = S3BucketFactory.init_s3_bucket(
+                service_name=self.image_name,
+                stack_name=self.image_name,
+                custom_s3_bucket=self._get_custom_bucket(),
+                artifact_directory=self.s3_artifacts_dir,
+            )
+        return self.__bucket
+
+    def _get_custom_bucket(self):
```
demartinofra, line 172: optional: we can transform this into a get_bucket_name and return whatever is in the tag and then have the caller check if the returned name is == to the S3Bucket.get_bucket_name so that we simplify all this articulated logic

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -146,6 +157,91 @@ def config(self):
...
+    def _get_artifact_dir(self):
+        """Get artifact directory from image tag or stack tag."""
+        try:
+            image_info = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
+            self.__s3_artifact_dir = search_tag(image_info, PCLUSTER_S3_IMAGE_DIR_TAG)
+            if self.__s3_artifact_dir is None:
+                raise ImageNotFoundError
+        except AWSClientError as e:
+            if not isinstance(e, ImageNotFoundError):
+                LOGGER.error("Error occurs during searching the artifact directory in image tag.")
```
demartinofra, line 211: nit: let's find a more meaningful error message

cli/src/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
```diff
@@ -76,35 +76,23 @@ def _delete_s3_artifacts(event):
...
-                bucket.object_versions.delete()
-                bucket.delete()
-                logger.info("S3 bucket %s deletion: COMPLETED", bucket_name)
-            else:
-                logger.info("Cluster S3 artifact under %s/%s deletion: STARTED", bucket_name, artifact_directory)
-                bucket.objects.filter(Prefix="%s/" % artifact_directory).delete()
-                bucket.object_versions.filter(Prefix="%s/" % artifact_directory).delete()
-                logger.info("Cluster S3 artifact under %s/%s deletion: COMPLETED", bucket_name, artifact_directory)
+            logger.info("Cluster S3 artifact under %s/%s deletion: STARTED", bucket_name, artifact_directory)
+            bucket.objects.filter(Prefix="%s/" % artifact_directory).delete()
```
demartinofra, line 86: nit: let's be consistent in string formatting and use f-strings everywhere?

cli/src/common/boto3/ec2.py
```diff
@@ -91,20 +91,25 @@ def describe_images(self, ami_ids, filters, owners):
...
 
     def image_exists(self, image_name):
         """Return a boolean describing whether or not an image with the given search criteria exists."""
-        filters = [{"Name": "tag:" + PCLUSTER_IMAGE_NAME_TAG, "Values": [image_name]}]
-        owners = ["self"]
         try:
-            self.describe_images(ami_ids=[], filters=filters, owners=owners)
+            self.describe_image_by_name_tag(image_name)
             return True
         except AWSClientError as e:
```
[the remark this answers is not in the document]
yuleiwan: Isn't there a case describe_images call failed due to policy or other reasons and raise AWSClientError not ImageNotFoundError?

cli/src/pcluster/models/cluster.py
```diff
@@ -186,11 +202,47 @@ def source_config(self):
...
+    def s3_artifacts_dir(self):
+        """Get s3 artifacts dir."""
+        if self.__s3_artifact_dir is None:
+            self._get_artifact_dir()
+        return self.__s3_artifact_dir
+
+    def _get_artifact_dir(self):
+        """Get artifact directory in S3 bucket by stack output."""
+        try:
+            stack_data = AWSApi.instance().cfn.describe_stack(self.stack_name)
```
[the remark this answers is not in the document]
yuleiwan: I don't think we can use self.stack, it raises ClusterActionError if describe_stack fails, should we handle ClusterActionError and AWSClientError in this function?

cli/src/pcluster/models/cluster.py
```diff
@@ -238,6 +286,33 @@ def status(self):
...
+            custom_bucket_name = self.stack.s3_bucket_name
+            if custom_bucket_name == S3Bucket.get_bucket_name(AWSApi.instance().sts.get_account_id(), get_region()):
+                custom_bucket_name = None
+        except AWSClientError as e:
+            if f"Stack with id {self.stack_name} does not exist" in str(e):
+                custom_bucket_name = self.config.custom_s3_bucket
+            else:
+                raise ClusterActionError(f"Unable to get custom bucket name. {e}")
+
+        if self.__bucket is None:
```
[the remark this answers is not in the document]
yuleiwan: good catch, done

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+            try:
+                bucket.check_bucket_exists()
+            except AWSClientError as e:
+                raise AWSClientError(
+                    "check_bucket_exists", f"Unable to access config-specified S3 bucket {bucket.name}. Due to {str(e)}"
+                )
+        else:
+            bucket = S3Bucket(service_name=service_name, artifact_directory=artifact_directory, stack_name=stack_name)
+            try:
+                bucket.check_bucket_exists()
```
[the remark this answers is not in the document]
yuleiwan: I don't think we can use head_object failure to determine the bucket existence, because the error message for head_object if bucket doesn't exist is:
```
Error during execution of head_object. Unexpected error when calling get_bucket_location on S3 bucket 'hello': 'Forbidden'. Error code: 403
```
it could be permission deny.
Instead, head_bucket gives error `Not Found` is clearer.

cli/src/pcluster/models/cluster.py
```diff
@@ -270,22 +343,24 @@ def create(
...
             LOGGER.debug("StackId: %s", self.stack.id)
             LOGGER.info("Status: %s", self.stack.status)
 
         except Exception as e:
-            if not creation_result and self.bucket:
+            if not creation_result:
                 # Cleanup S3 artifacts if stack is not created yet
-                self.bucket.delete()
+                self.bucket.delete_s3_artifacts()
             if isinstance(e, ClusterActionError):
```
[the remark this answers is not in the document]
yuleiwan: yes, done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: What about "Unable to get artifact directory from image {self.image_name} tag."?

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Done

cli/src/pcluster/models/cluster.py
```diff
@@ -152,6 +151,13 @@ def batch_compute_environment(self):
         """Return Batch compute environment."""
         return self._get_output("BatchComputeEnvironmentArn")
 
+    def update_stack_data(self):
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/cluster.py
```diff
@@ -238,6 +286,33 @@ def status(self):
         """Return the cluster status."""
         return self.stack.status
 
+    @property
+    def bucket(self):
+        """Return a bucket configuration."""
+        # try to get custom bucket name from stack.
+        try:
+            AWSApi.instance().cfn.describe_stack(self.stack_name)
```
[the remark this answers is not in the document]
yuleiwan: I don't think we can use self.stack to replace this check, the cases are different for `_get_artifact_dir` and `bucket` property, because `_get_artifact_dir` is executed in pcluster delete/update commands, but `bucket` will be executed in pcluster create/delete/update commands, in pcluster create we don't have stack, in that case we still need to initailize the bucket rather than raising a ClusterActionError.

cli/src/pcluster/models/common.py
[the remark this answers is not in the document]
yuleiwan: use self.stack.s3_artifact_directory

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
         attribute_json = {"cfncluster": self._cfncluster_attributes}
         attribute_json.update(self._extra_attributes)
         return json.dumps(attribute_json, sort_keys=True)
+
+
+S3FileFormat = Enum("S3FileFormat", ("YAML", "JSON", "TEXT"))
```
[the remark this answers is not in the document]
yuleiwan: yes, create a class for S3FileFormat

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+        AWSApi.instance().s3.head_bucket(bucket_name=self.name)
+
+    def create_bucket(self):
+        """
+        Create a new S3 bucket.
+
+        :param bucket_name: name of the S3 bucket to create
+        :param region: aws region
+        """
+        if self.region != "us-east-1":
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+                LOGGER.info("Deleting artifacts under %s/%s", self.name, self.artifact_directory)
+                AWSApi.instance().s3_resource.delete_object(
+                    bucket_name=self.name, prefix="%s/" % self.artifact_directory
+                )
+                AWSApi.instance().s3_resource.delete_object_versions(
+                    bucket_name=self.name, prefix="%s/" % self.artifact_directory
+                )
+            except AWSClientError as e:
+                LOGGER.warning(
+                    "Failed to delete S3 artifact under %s/%s with error %s. Please delete them manually.",
```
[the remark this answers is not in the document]
yuleiwan: `%s/%s` includes bucket name and artifact directory. But I'd avoid raising error here, because it's not neccessary to raise an error when artifacts directory deletion fails.

cli/src/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
yuleiwan: done

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+            except AWSClientError as e:
+                if "Error code: 404" in str(e):
+                    try:
+                        bucket.configure_s3_bucket()
+                        bucket.upload_bootstrapped_file()
+                    except AWSClientError:
+                        LOGGER.error("Configure S3 bucket %s failed.", bucket.name)
+                        raise
+                else:
+                    LOGGER.error("Unable to check S3 bucket %s bootstrapped file.", bucket.name)
```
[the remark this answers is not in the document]
yuleiwan: What about `"Unable to check S3 bucket %s is configured properly.", bucket_name`

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+                        LOGGER.error("Unable to create S3 bucket %s.", bucket.name)
+                        raise
+                else:
+                    LOGGER.error("Unable to check S3 bucket %s existence.", bucket.name)
+                    raise
+
+            try:
+                bucket.check_bucket_is_bootstrapped()
+            except AWSClientError as e:
+                if "Error code: 404" in str(e):
```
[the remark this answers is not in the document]
yuleiwan: done, add error_code as attribute to AWSClientError and use it.

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,330 @@ def dump_json(self):
...
+        return result
+
+    def _get_file(self, file_name, file_type, version_id=None):
+        """Get file from S3 bucket."""
+        result = AWSApi.instance().s3.get_object(
+            bucket_name=self.name, key=self.get_object_key(file_type, file_name), version_id=version_id
+        )
+
+        config_content = result["Body"].read().decode("utf-8")
+        return yaml.safe_load(config_content)
```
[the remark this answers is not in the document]
yuleiwan: add different formats

cli/src/pcluster/models/imagebuilder.py
yuleiwan: what if stack doesn't exist? custom_bucket_name only needs during initalization, we can use name property to get bucket name in other cases, so I think it's more reasonable to put custom bucket name retrieving in imagebuilder/cluster.

cli/src/common/boto3/ec2.py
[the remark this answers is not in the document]
demartinofra: yes but in the current code if the exception is not of type ImageNotFoundError you are just raising which is the same thing that happens if you do not try to catch the broader AWSClientError.

The proposed code should have the same behavior as the current one

cli/src/pcluster/models/common.py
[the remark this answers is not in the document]
demartinofra: ok then let's leave it as is. I wish we could spare the extra call.

cli/src/pcluster/models/imagebuilder.py
demartinofra: or 'Unable to find tag X in image Y'?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -146,6 +167,89 @@ def config(self):
...
+                    custom_bucket_name = None
+            except AWSClientError as e:
+                raise ImageBuilderActionError(f"Unable to get S3 bucket name from stack {self.stack.name} tag. {e}")
+
+        return custom_bucket_name
+
+    def _get_artifact_dir(self):
+        """Get artifact directory from image tag or stack tag."""
+        try:
+            image_info = AWSApi.instance().ec2.describe_image_by_name_tag(self.image_name)
```
demartinofra, line 214: we can optimize this in a separate PR but describe_image_by_name_tag is called multiple times while it could be reused (e.g. _get_custom_bucket is calling it)

demartinofra (commented):
LGTM. Just left a few minor comments.

one additional advice: if you haven't done it already it is useful to look at the codecoverage report form [here](https://app.codecov.io/gh/aws/aws-parallelcluster/compare/2579/diff) so that you can make sure unit tests are covering the most important code paths.

cli/src/pcluster/models/imagebuilder.py
yuleiwan: Done

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,338 @@ def dump_json(self):
...
+            except AWSClientError as e:
+                raise AWSClientError(
+                    "check_bucket_exists", f"Unable to access config-specified S3 bucket {bucket.name}. Due to {str(e)}"
+                )
+        else:
+            bucket = S3Bucket(service_name=service_name, artifact_directory=artifact_directory, stack_name=stack_name)
+            try:
+                bucket.check_bucket_exists()
+            except AWSClientError as e:
+                if "Not Found" in str(e):
```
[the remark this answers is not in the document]
yuleiwan: it has error code, use 404 instead of Not Found

cli/src/pcluster/models/common.py
```diff
@@ -239,3 +251,338 @@ def dump_json(self):
...
+            config_key=self.get_object_key(file_type, file_name),
+        )
+        return url
+
+
+class S3BucketFactory:
+    """S3 bucket factory to return a bucket object with existence check and creation."""
+
+    @staticmethod
+    # flake8: noqa: C901
```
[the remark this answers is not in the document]
yuleiwan: Simplified

cli/src/common/boto3/common.py
```diff
@@ -24,6 +24,15 @@ def __init__(self, function_name: str, message: str, error_code: str = None):
         if error_code:
             message += f" Error code: {error_code}"
         super().__init__(message)
+        if error_code:
```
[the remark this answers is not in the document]
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
yuleiwan: done

cli/src/pcluster/models/imagebuilder.py
enrico-usai: You can even add Cached annotation to describe_image_by_name_tag call in AWS Api layer.

yuleiwan: I prefer using an ImageInfo class to store the describe image result like StackInfo, and add a property in ImageBuilder, it has been implemented in pcluster delete PR. https://github.com/aws/aws-parallelcluster/pull/2507

demartinofra (approved):
🚀

PR #2574, opened 2021-03-25
3.0: Cherry pick changes from develop

enrico-usai:
# Included changes

* from: https://github.com/aws/aws-parallelcluster/commit/44b9cc25c2b36bc56b864d979551a1d92e696cf8 - Starting date: Commits on Dec 24, 2020 - Update badges in README
* to: https://github.com/aws/aws-parallelcluster/commit/32749e40ef40534396e936481190d42b9b0fd3cb - Drop support for python versions < 3.6

## Impacting changes:
* ef3393fe Support cluster creation without EC2 key pair - This required a change in the KeyPair validator and I added tests for it.

# Excluded changes 

Run instance sanity check:
* Restore instance type sanity check using cluster ami 9eabda48 ddeidda on 26/02/2021 at 16:03
* Fix run_instance dryrun by using latest AL2 AMI with correct architecture 57dccaee Rex on 26/02/2021 at 03:17

Validators:
* Add cluster region validator 11a81be3 ddeidda on 02/03/2021 at 16:41
* Add validation for s3 bucket to be in same region of cluster stack f34e67b0 chenwany on 02/03/2021 at 22:09
* Remove ap-northeast-3 from list of supported regions fb3146fd Francesco De Martino on 23/03/2021 at 11:28
* Add sanity check for instance usage classes 40e36584 ddeidda on 17/03/2021 at 13:30

Slurm upgrade:
* Upgrade to Slurm 20.11.4 a17dbdf7 Rex <shuningc@amazon.com> on 22/02/2021 at 23:27

Instance types data param:
* Introduce new configuration parameter for development 5a6250fc ddeidda on 10/03/2021 at 18:20
* Add integration tests support for `instance_types_data` parameter e717cb36 ddeidda on 11/03/2021 at 19:23
* Fix additional_instance_types data injection in integration tests 796a865d ddeidda on 12/03/2021 at 21:06
* Return supported instance types as list 5cedde67 ddeidda on 13/03/2021 at 08:46
* Move instance types data into cfncluster section 186098ba ddeidda on 15/03/2021 at 10:03
* Fix pcluster update behavior with additional instance types data b1bf396c ddeidda on 17/03/2021 at 23:09
* Replace completeley instance types data when loading config 34bab301 ddeidda on 18/03/2021 at 11:05

PR #2584, opened 2021-03-29
3.0: Migrate launch templates validators

enrico-usai:
Split code into two different validators to tests head node and queue configuration.

+ Move get_supported_az_for_instance_type and get_supported_az_for_instance_types into AWSApi layer.
+ Differentiate list_instance_types from describe_instance_type_offerings but use the same
  cached local method to save boto3 calls.

This includes cherry pick of:
* Restore instance type sanity check using cluster ami 9eabda4 ddeidda on 26/02/2021 at 16:03
* Fix run_instance dryrun by using latest AL2 AMI with correct architecture 57dccae Rex on 26/02/2021 at 03:17

Left behind from: #2574


hanwen-pcluste (approved):
LGTM! 
Sorry for not being to look into each line of the change. Because the change is huge, and requires much time to review. Let's merge and run integration tests.

PR #2601, opened 2021-04-01
Fix import of packaging.version

demartinofra:
`from packaging import version` was breaking when installing the CLI in a clean env (e.g. after creating the wheel dist).

### DIFF
--- a/cli/src/api/pcluster_api.py
+++ b/cli/src/api/pcluster_api.py
@@ -12,7 +12,7 @@ import json
 import logging
 import os
 
-from packaging import version
+from pkg_resources import packaging
 
 from common.aws.aws_api import AWSApi
 from pcluster.cli_commands.compute_fleet_status_manager import ComputeFleetStatus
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -806,7 +806,7 @@ def make_resource(self, data, **kwargs):
 class QueueCustomActionSchema(BaseSchema):
     """Represent the schema of the custom action for the queue."""
 
-    script = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
+    script = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
```
enrico-usai, line 809: I decided to use the script as "update_key", it means that the script field cannot be updated.

If you're changing the script it means that you're inserting a new CustomAction and you will not see the differences in the update output. You'll see a deleted custom action and a new one.

The update of the CustomActions is already supported (see where the list of QueueCustomActionSchema is defined)

demartinofra: Removed the commit for now

enrico-usai (approved):

PR #2615, opened 2021-04-07
3.0: Try to fix test tag propagation

enrico-usai:
* Remove substack refences
* Add a shared storage to check if tags are correctly propagated


hanwen-pcluste (approved):
LGTM!

PR #2617, opened 2021-04-07
FSx in isolated subnet

rexcsn:
* Do not query FSx API for MountName and DNSName
* For existing filesystems, MountName and DNSName are retrieved by CLI and passed into cookbook using attributes
* For filesystem created with cluster, MountName is retrieved from CFN filesystem resource attribute. DNSName is generated in hardcoded format
* Fix unit tests
* Add integration test for FSx in private subnet

### DIFF
--- a/cli/src/pcluster/config/mappings.py
+++ b/cli/src/pcluster/config/mappings.py
@@ -27,6 +27,8 @@ from pcluster.config.cfn_param_types import (
     EFSCfnSection,
     ExtraJsonCfnParam,
     FloatCfnParam,
+    FSxDNSNameParam,
+    FSxMountNameParam,
     HeadNodeAvailabilityZoneCfnParam,
     HeadNodeInstanceTypeCfnParam,
     IntCfnParam,
@@ -575,7 +577,19 @@ FSX = {
                 "default": "NONE",
                 "allowed_values": ALLOWED_VALUES["fsx_drive_cache_type"],
                 "update_policy": UpdatePolicy.UNSUPPORTED
-            })
+            }),
+            ("existing_mount_name", {
+                "type": FSxMountNameParam,
+                "default": "NONE",
+                "update_policy": UpdatePolicy.IGNORED,
+                "visibility": Visibility.PRIVATE,
+            }),
+            ("existing_dns_name", {
+                "type": FSxDNSNameParam,
+                "default": "NONE",
+                "update_policy": UpdatePolicy.IGNORED,
+                "visibility": Visibility.PRIVATE,
+            }),
         ]
     )
 }
--- a/cloudformation/aws-parallelcluster.cfn.json
+++ b/cloudformation/aws-parallelcluster.cfn.json
@@ -336,9 +336,9 @@
       "Default": "1"
     },
     "FSXOptions": {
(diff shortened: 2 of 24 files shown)


cli/src/pcluster/config/mappings.py
```diff
@@ -575,7 +577,19 @@
                 "default": "NONE",
                 "allowed_values": ALLOWED_VALUES["fsx_drive_cache_type"],
                 "update_policy": UpdatePolicy.UNSUPPORTED
-            })
+            }),
+            ("existing_mount_name", {
+                "type": FSxMountNameParam,
+                "default": "NONE",
+                "update_policy": UpdatePolicy.UNSUPPORTED,
```
demartinofra, line 584: probably for private parameters we want to have IGNORED as UpdatePolicy and rely on the UNSUPPORTED update for fsx_fs_id?otherwise I don't know if this causes problems such as reporting them in the update diffs

cloudformation/aws-parallelcluster.cfn.json
```diff
@@ -336,9 +336,9 @@
       "Default": "1"
     },
     "FSXOptions": {
-      "Description": "Comma separated list of FSx related options, 17 parameters in total, [shared_dir,fsx_fs_id,storage_capacity,fsx_kms_key_id,imported_file_chunk_size,export_path,import_path,weekly_maintenance_start_time,deployment_type,per_unit_storage_throughput,daily_automatic_backup_start_time,automatic_backup_retention_days,copy_tags_to_backups,fsx_backup_id,auto_import_policy,storage_type,drive_cache_type]",
+      "Description": "Comma separated list of FSx related options, 19 parameters in total, [shared_dir,fsx_fs_id,storage_capacity,fsx_kms_key_id,imported_file_chunk_size,export_path,import_path,weekly_maintenance_start_time,deployment_type,per_unit_storage_throughput,daily_automatic_backup_start_time,automatic_backup_retention_days,copy_tags_to_backups,fsx_backup_id,auto_import_policy,storage_type,drive_cache_type,existing_mount_name,existing_dns_name]",
```
demartinofra, line 339: not part of this PR but I'm wondering how close we are to reach the param value length limit when several fsx settings are used

cloudformation/tests/test_cfn_template_rendering.py
```diff
@@ -199,7 +199,7 @@ def test_hit_substack_rendering(tmp_path, test_config):
             "vol1,vol2,vol3,vol4,vol5",
             "gp2,io1,sc1,st1,gp2",  # No io1 metrics
             "raid,1,2,gp2,20,100,false,NONE",
-            "/fsx,NONE,1200,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE",
+            "/fsx,NONE,1200,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE,NONE",
```
demartinofra, line 202: was this test even running?cause I can see you had to add NONE 4 times

demartinofra:
let's add a note to the changelog about supporting FSX in subnets with no internet access

cloudformation/tests/test_cfn_template_rendering.py
rexcsn: Test is running, but logic only depends on the first shared_dir parameter, so other FSx params do not matter.
Added a comment in the code to explain this

demartinofra (approved):

PR #2631, opened 2021-04-12
3.0: Fix some issues in config validation and bucket deletion

enrico-usai:
* Do not modify source data during pre_dump/pre_load actions. This was causing errors when trying to reuse the same source dict.
* Cleanup S3 bucket only if artifacts are uploaded 
* Remove useless ConfigValidationError and raise_on_error option from Validator
* Use different exceptions for validation and update failures. In this way it's easier to distinguish the failure type and re-raise these exception that contain additional info (i.e. validation failures and update changes).

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -307,6 +307,7 @@ class Cluster:
     ):
         """Create cluster."""
         creation_result = None
+        artifacts_uploaded = False
         try:
             # check cluster existence
             if AWSApi.instance().cfn.stack_exists(self.stack_name):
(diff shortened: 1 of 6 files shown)


cli/src/pcluster/models/cluster.py
```diff
@@ -324,7 +344,7 @@ def create(
                 )
 
             # upload cluster artifacts and generated template
-            self._upload_artifacts()
+            artifacts_uploaded = self._upload_artifacts()
```
demartinofra, line 347: rather than having this return a bool we could just set artifacts_uploaded to True right after calling the function?

enrico-usai: Changed

demartinofra (approved):

PR #2634, opened 2021-04-12
Centos7 ARM

sean-smith:

### DIFF
--- a/cli/src/pcluster/config/validators.py
+++ b/cli/src/pcluster/config/validators.py
@@ -90,14 +90,14 @@ FSX_MESSAGES = {
 
 FSX_SUPPORTED_ARCHITECTURES_OSES = {
     "x86_64": SUPPORTED_OSS,
-    "arm64": ["ubuntu1804", "ubuntu2004", "alinux2", "centos8"],
+    "arm64": ["ubuntu1804", "ubuntu2004", "alinux2", "centos7", "centos8"],
 }
 
 FSX_PARAM_WITH_DEFAULT = {"drive_cache_type": "NONE"}
 
 EFA_UNSUPPORTED_ARCHITECTURES_OSES = {
     "x86_64": [],
-    "arm64": ["centos8"],
+    "arm64": ["centos7", "centos8"],
 }
 
 EBS_VOLUME_TYPE_TO_VOLUME_SIZE_BOUNDS = {
--- a/cli/src/pcluster/config/validators.py
+++ b/cli/src/pcluster/config/validators.py
@@ -31,7 +31,6 @@ from pcluster.utils import (
     get_supported_architectures_for_instance_type,
     get_supported_compute_instance_types,
     get_supported_instance_types,
-    get_supported_os_for_architecture,
     get_supported_os_for_scheduler,
     is_instance_type_format,
     paginate_boto3,
@@ -90,7 +89,7 @@ FSX_MESSAGES = {
 
 FSX_SUPPORTED_ARCHITECTURES_OSES = {
     "x86_64": SUPPORTED_OSS,
-    "arm64": ["ubuntu1804", "ubuntu2004", "alinux2", "centos7", "centos8"],
+    "arm64": SUPPORTED_OSS,
 }
 
 FSX_PARAM_WITH_DEFAULT = {"drive_cache_type": "NONE"}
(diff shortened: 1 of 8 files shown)


cli/src/pcluster/config/validators.py
```diff
@@ -90,14 +90,14 @@
 
 FSX_SUPPORTED_ARCHITECTURES_OSES = {
     "x86_64": SUPPORTED_OSS,
-    "arm64": ["ubuntu1804", "ubuntu2004", "alinux2", "centos8"],
+    "arm64": ["ubuntu1804", "ubuntu2004", "alinux2", "centos7", "centos8"],
```
rexcsn, line 93: We can just use `SUPPORTED_OSS` here.

cli/src/pcluster/utils.py
```diff
@@ -563,10 +563,7 @@ def get_supported_os_for_scheduler(scheduler):
 
 def get_supported_os_for_architecture(architecture):
     """Return list of supported OSes for the specified architecture."""
-    oses = ["alinux2", "ubuntu1804", "ubuntu2004", "centos8"]
-    if architecture == "x86_64":
-        oses.extend(["centos7"])
-    return oses
+    return ["alinux2", "ubuntu1804", "ubuntu2004", "centos7", "centos8"]
```
rexcsn, line 566: We can just remove `get_supported_os_for_architecture` function and use `SUPPORTED_OSS` everywhere the function is used.

`architecture_os_validator` and `_validate_createami_args_ami_compatibility` logic might need to be modified to account for the warning message when using centos7+arm.

cli/tests/pcluster/config/test_validators.py
```diff
@@ -1827,13 +1827,6 @@ def test_dcv_enabled_validator(
                 supported_architectures=list(FSX_SUPPORTED_ARCHITECTURES_OSES.keys())
             ),
         ),
-        (
-            "arm64",
```
rexcsn, line 1831: We need to add this in the Supported combinations above

cli/src/pcluster/utils.py
sean-smith: Ok fixed!

cli/tests/pcluster/config/test_validators.py
sean-smith: It's there by nature, it's in `SUPPORTED_OSS`

cli/src/pcluster/config/validators.py
```diff
@@ -1125,11 +1124,10 @@ def architecture_os_validator(param_key, param_value, pcluster_config):
     warnings = []
 
     architecture = pcluster_config.get_section("cluster").get_param_value("architecture")
-    allowed_oses = get_supported_os_for_architecture(architecture)
-    if param_value not in allowed_oses:
+    if param_value not in SUPPORTED_OSS:
```
rexcsn, line 1127: This is a redundant logic that is already covered by `"allowed_values": SUPPORTED_OSS,` in `mappings.py`. I would remove this validator normally, but I know we would need to add a warning for centos7+arm in this validator later.

Can we either add a `TO-DO` comment in the code or a placeholder warning message to track why we are still keeping this validator for now, and make sure we can revisit this code later

cli/tests/pcluster/config/test_validators.py
rexcsn: Sorry should have been clear, I meant we would want to add a new test case `("arm64", "centos7", None),` above

cli/src/pcluster/config/validators.py
sean-smith: Add validation message and refactored run_architecture_validator_test()

cli/tests/pcluster/config/test_validators.py
```diff
@@ -2285,47 +2281,68 @@ def test_intel_hpc_architecture_validator(mocker, enabled, architecture, expecte
...
     )
 
 
 @pytest.mark.parametrize(
-    "disable_hyperthreading, architecture, expected_message",
+    "disable_hyperthreading, architecture, expected_errors, expected_warnings",
     [
-        (True, "x86_64", []),
-        (False, "x86_64", []),
+        (True, "x86_64", [], []),
```
rexcsn, line 2332: Minor: Since these is no test for warning in this test case, we could have not added the extra `expected_warnings` parameter and use the hardcoded `[]` when calling `run_architecture_validator_test`, similar to the way you did it in `test_instances_architecture_compatibility_validator`.

rexcsn (approved):
LGTM!

cli/tests/pcluster/config/test_validators.py
sean-smith: fixed

rexcsn (approved):

PR #2637, opened 2021-04-13
3.0: FSx in isolated subnet

rexcsn:
* Do not query FSx API for MountName and DNSName
* For existing filesystems, MountName and DNSName are retrieved by CLI and passed into cookbook using attributes
* For filesystem created with cluster, MountName is retrieved from CFN filesystem resource attribute. DNSName is generated in hardcoded format
* Add integration test for FSx in private subnet

### DIFF
--- a/cli/src/common/boto3/fsx.py
+++ b/cli/src/common/boto3/fsx.py
@@ -1,0 +1,31 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+from common.aws.aws_resources import FsxFileSystemInfo
+from common.boto3.common import AWSExceptionHandler, Boto3Client
+from pcluster.utils import Cache
+
+
+class FSxClient(Boto3Client):
+    """S3 Boto3 client."""
+
+    def __init__(self):
+        super().__init__("fsx")
+
+    @AWSExceptionHandler.handle_client_exception
+    @Cache.cached
+    def get_filesystem_info(self, fsx_fs_id):
+        """
+        Return FSx filesystem info.
+
+        :param fsx_fs_id: FSx file system Id
+        :return: filesystem info
+        """
+        return FsxFileSystemInfo(self._client.describe_file_systems(FileSystemIds=[fsx_fs_id]).get("FileSystems")[0])
--- a/cli/src/pcluster/models/cluster_config.py
+++ b/cli/src/pcluster/models/cluster_config.py
@@ -356,16 +356,12 @@ class SharedFsx(Resource):
     @property
     def existing_mount_name(self):
         """Return MountName if using existing FSx filesystem."""
-        if self.file_system_id:
-            return self.file_system_data.mount_name
-        return ""
+        return self.file_system_data.mount_name if self.file_system_id else ""
 
     @property
     def existing_dns_name(self):
         """Return DNSName if using existing FSx filesystem."""
-        if self.file_system_id:
-            return self.file_system_data.dns_name
-        return ""
+        return self.file_system_data.dns_name if self.file_system_id else ""
 
 
 # ---------------------- Networking ---------------------- #
--- a/cli/src/pcluster/models/cluster_config.py
+++ b/cli/src/pcluster/models/cluster_config.py
@@ -289,6 +289,7 @@ class SharedFsx(Resource):
         self.auto_import_policy = Resource.init_param(auto_import_policy)
         self.drive_cache_type = Resource.init_param(drive_cache_type)
         self.fsx_storage_type = Resource.init_param(fsx_storage_type)
+        self.__file_system_data = None
 
     def _validate(self):
         self._execute_validator(NameValidator, name=self.name)
@@ -344,6 +345,27 @@ class SharedFsx(Resource):
             self._execute_validator(
                 FsxAutoImportValidator, auto_import_policy=self.auto_import_policy, import_path=self.import_path
             )
+
+    @property
+    def file_system_data(self):
+        """Return filesystem information if using existing FSx."""
+        if not self.__file_system_data and self.file_system_id:
+            self.__file_system_data = AWSApi.instance().fsx.get_filesystem_info(self.file_system_id)
+        return self.__file_system_data
+
+    @property
+    def existing_mount_name(self):
+        """Return MountName if using existing FSx filesystem."""
+        if self.file_system_id:
+            return self.file_system_data.mount_name
+        return ""
+
+    @property
+    def existing_dns_name(self):
+        """Return DNSName if using existing FSx filesystem."""
+        if self.file_system_id:
+            return self.file_system_data.dns_name
+        return ""
 
 
 # ---------------------- Networking ---------------------- #
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -86,6 +86,7 @@ class ClusterCdkStack(core.Stack):
         self.compute_security_groups = {}
         self.shared_storage_mappings = {storage_type: [] for storage_type in SharedStorageType}
         self.shared_storage_options = {storage_type: "" for storage_type in SharedStorageType}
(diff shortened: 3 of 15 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -607,6 +608,20 @@ def _add_shared_storage(self, storage):
     def _add_fsx_storage(self, id: str, shared_fsx: SharedFsx):
         """Add specific Cfn Resources to map the FSX storage."""
         fsx_id = shared_fsx.file_system_id
+        self.shared_storage_attributes[shared_fsx.shared_storage_type]["MountName"] = ",".join(
```
enrico-usai, line 611: `_add_fsx_storage` is a function to add a single fsx storage. We shouldn't loop for all the Fsx in `self.config.shared_storage`.

If you want to keep the "multi-fsx" behavior you can move these loops at the end, after finishing `_add_shared_storage` loop, but multiple FSx is not supported yet so you can simply do:
```
self.shared_storage_options[shared_fsx.shared_storage_type]["MountName"] = fs.existing_mount_name
self.shared_storage_options[shared_fsx.shared_storage_type]["DNSName"] = fs.existing_dns_name
```

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -641,6 +656,9 @@ def _add_fsx_storage(self, id: str, shared_fsx: SharedFsx):
                 security_group_ids=self._get_compute_security_groups(),
             )
             fsx_id = fsx_resource.ref
+            self.shared_storage_attributes[shared_fsx.shared_storage_type][
```
enrico-usai, line 659: If you have this code, what about adding the set of DNSName here and remove the loop below?

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -9,18 +9,100 @@
...
     assert_that(cluster).is_not_none()
 
     assert_that(len(get_compute_nodes_instance_ids(cluster.cfn_name, region))).is_equal_to(1)
+    _test_fsx_in_private_subnet(cluster, os, region, scheduler, fsx_mount_dir, bastion_factory)
+
+
+def _test_fsx_in_private_subnet(cluster, os, region, scheduler, fsx_mount_dir, bastion_factory):
+    """Test FSx can be mounted in private subnet."""
+    bastion_ip = bastion_factory()
+    logging.info("Sleeping for 60 sec to wait for bastion ssh to become ready.")
```
enrico-usai, line 49: Is there a way to check bastion/ec2 instance status rather than waiting 60 seconds?

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -9,18 +9,100 @@
...
+    time.sleep(60)
+    logging.info("Bastion_ip: {}".format(bastion_ip))
```
enrico-usai, line 51: nit: what about f-strings or pass the bastion_ip as logging argument?

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -9,18 +9,100 @@
...
+    remote_command_executor = RemoteCommandExecutor(cluster, bastion="ec2-user@{}".format(bastion_ip))
```
enrico-usai, line 52: same here

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -424,14 +424,10 @@ def _test_import_path(remote_command_executor, mount_dir):
     assert_that(result.stdout).is_equal_to("Downloaded by FSx Lustre")
 
 
-def _test_fsx_lustre_correctly_shared(scheduler_commands, remote_command_executor, mount_dir):
+def test_fsx_lustre_correctly_shared(scheduler_commands, remote_command_executor, mount_dir):
```
enrico-usai, line 427: Why did you remove the underscore prefix?
The underscore was a way to avoid considering this function as a "test" function, I think we should restore it because it's an utility function called by other real tests.

We can rename it to "assert_*" to avoid confusion.
The same is true for test_fsx_lustre_correctly_mounted.

tests/integration-tests/remote_command_executor.py
```diff
@@ -27,15 +27,29 @@ def __init__(self, result):
...
-            connect_kwargs={"key_filename": [cluster.ssh_key]},
-        )
+        connection_kwargs = {
+            "host": cluster.head_node_ip,
+            "user": username,
+            "forward_agent": False,
+            "connect_kwargs": {"key_filename": [cluster.ssh_key]},
+        }
+        if bastion:
+            # Need to execute simple ssh command before using Connection to avoid Paramiko _check_banner error
```
enrico-usai, line 40: Can you add a link if it's a known issue?

tests/integration-tests/remote_command_executor.py
```diff
@@ -27,15 +27,29 @@ def __init__(self, result):
...
+            run_command(
+                "ssh -tt -i {key_path} -o StrictHostKeyChecking=no "
+                '-o ProxyCommand="ssh -tt -o StrictHostKeyChecking=no -W %h:%p -A {bastion}" '
+                "-A {user}@{head_node} hostname".format(
+                    key_path=cluster.ssh_key, bastion=bastion, user=username, head_node=cluster.head_node_ip
+                ),
+                timeout=30,
+                shell=True,
+            )
+            connection_kwargs["gateway"] = "ssh -W %h:%p -A {}".format(bastion)
```
enrico-usai, line 50: nit: you can use f-strings everywhere

tests/integration-tests/remote_command_executor.py
rexcsn: Was not able to figure out the root cause of the issue. Most issue online about `Paramiko _check_banner` are about adjusting Paramiko `connect` parameters. However we use fabric connection in integration test, which depends on Paramiko, but do not support all Paramiko `connect` parameters to adjust `_check_banner` behaviors.

Didn't experiment with using Paramiko directly, since we do not want to change existing behavior of tests; I just came up with the work around.

tests/integration-tests/remote_command_executor.py
rexcsn: Done

cli/src/pcluster/models/cluster_config.py
```diff
@@ -345,6 +346,27 @@ def _validate(self):
...
+    def file_system_data(self):
+        """Return filesystem information if using existing FSx."""
+        if not self.__file_system_data and self.file_system_id:
+            self.__file_system_data = AWSApi.instance().fsx.get_filesystem_info(self.file_system_id)
+        return self.__file_system_data
+
+    @property
+    def existing_mount_name(self):
+        """Return MountName if using existing FSx filesystem."""
+        if self.file_system_id:
```
enrico-usai, line 359: `return self.file_system_data.mount_name if self.file_system_id else ""`

cli/src/pcluster/models/cluster_config.py
```diff
@@ -345,6 +346,27 @@ def _validate(self):
...
+            return self.file_system_data.mount_name
+        return ""
+
+    @property
+    def existing_dns_name(self):
+        """Return DNSName if using existing FSx filesystem."""
+        if self.file_system_id:
```
enrico-usai, line 366: same here

cli/src/common/boto3/fsx.py
```diff
@@ -0,0 +1,31 @@
...
+    @AWSExceptionHandler.handle_client_exception
+    @Cache.cached
+    def get_filesystem_info(self, fsx_fs_id):
+        """
+        Return FSx filesystem info.
+
+        :param fsx_fs_id: FSx file system Id
+        :return: filesystem info
+        """
+        return FsxFileSystemInfo(self._client.describe_file_systems(FileSystemIds=[fsx_fs_id]).get("FileSystems")[0])
```
enrico-usai, line 31: Just double check if the file system doesn't exist if it fails with a clear message or raises a KeyError

An alternative is to raise:
`raise AWSClientError(function_name="describe_file_systems", message=f"File System {fsx_fs_id} not found")`

cli/src/pcluster/models/cluster_config.py
rexcsn: Done

cli/src/pcluster/models/cluster_config.py
rexcsn: Done

cli/src/common/boto3/fsx.py
rexcsn: Current behavior:
```
Beginning cluster creation for cluster: pl3
Validating cluster configuration...
ERROR: Cluster creation failed. Configuration is invalid.
Validation failures:
ERROR: File system 'fs-0b3ae5b7380d8ab06' does not exist.
```

enrico-usai (approved):

PR #2644, opened 2021-04-15
3.0: Initial import of ParallelCluster API code

demartinofra:
### Initial import of ParallelCluster API code

The API server stub has been generated by using OpenAPI generator and running the following commands:
```
cd api
./gradlew generatePythonServer
```

The code generated under `api/generated/python-server` directory has been imported into the cli source code and modified to obtain a first working [Connexion](https://connexion.readthedocs.io/en/latest/)+[Flask](https://flask.palletsprojects.com/en/1.1.x/) application.

Here is how the code is currently organized:
```
cli/src/api
├── __init__.py
├── controllers: defines the handlers for the various API operations. This handlers are automatically invoked by connexion following the routing defined in the openapi definition.
├── encoder.py: auto generated encoder to serialize Model objects into JSON 
├── flask_app.py: entry-point of the ParallelClusterApi Application. Implements the code to initialize the Connexion app and attach the proper error handling. It also allows to run a local development server.
├── models: dir containing all model classes that have been automatically generated by OpenAPI Generator. These classes did not require any change.
├── openapi/openapi.yaml: OpenAPI model for the ParallelCluster API. This is the same as the one generated from the Smithy model in the api/spec dir with the addition of a series of tags to integrate with Connexion routing.
├── typing_utils.py: auto generated file for typing utils
└── util.py: auto generated file
```

Currently API dependencies have been defined as `extras_require` and can be installed with `pip install -e '.[api]'`. When integrating the CLI with the API code we will evaluate whether to include these deps as part of the default ones.

### Implement entrypoint for AWS Lambda
This commit implements the entrypoint to run the ParallelCluster API as an AWS Lambda function. In order to run the WSGI Flask application in the Lambda environment we make use of a WSGI wrapper written for the Serverless framework: https://github.com/logandk/serverless-wsgi/blob/master/serverless_wsgi.py. The wrapper maps the API Gateway event sent as input to the AWS Lambda function into an WSGI invocation of the Flask app.

The ParallelClusterFlaskApp is initialized as a global variable in order to avoid to pay the inizialization overhead when the Lambda runtime is reused. 

### Add APIGateway definition to sample SAM template
Enhance the SAM sample template in order to facilitate local testing and deployment of the API. The `api/README.md` file contains instructions on how to use it.

### Clear cache when reusing objects across Lambda invocations
Since objects are reused in order to keep the Lambda environment warm and avoid needless initializations we need to cleanup the local cache when the Lambda handler is invoked.

### Add handler for ParallelCluster specific API exceptions
Registers an error handler to properly render as an HTTP response all ParallelClusterApiException that are raised from controllers.

It also implements a series of exceptions as they are defined in the API spec.

### Enable XRay tracing in AWS Lambda
Enables AWS XRay tracing and instruments tracing of the ParallelCluster API when invoked as a Lambda function: https://docs.aws.amazon.com/xray/latest/devguide/aws-xray.html. This enables us to easily trace the time spent by the various application components to handle a request.

***


cli/src/api/controllers/cluster_operations_controller.py
```diff
@@ -0,0 +1,206 @@
...
+    dryrun=None,
+    rollback_on_failure=None,
+    client_token=None,
+):
+    """
+    Create a ParallelCluster managed cluster in a given region.
+
+    :param create_cluster_request_content:
+    :param suppress_validators: Identifies one or more config validators to suppress. Format:
+    sALL|id:$value|level:(info|error|warning)|type:$value
```
enrico-usai, line 48: typo? `sALL`

cli/.pylintrc
```diff
@@ -158,6 +157,7 @@ disable=print-statement,
         import-error,
         no-self-use,
         no-member,
+        duplicate-code,
```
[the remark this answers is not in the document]
demartinofra: Added duplicate-code back for now. Main problem is that it detects duplicates in comments and this check can't be even disabled on a file basis.

cli/src/api/controllers/cluster_operations_controller.py
demartinofra: fixed

enrico-usai (approved):

PR #2646, opened 2021-04-16
Update armpl integration tests

ddeidda:
- Code changes to make test_arm_pl work correctly with armpl 21.0.0 integration.
- Added new check to verify that EULA docs are really available at the path specified in the output  when loading the module

### DIFF
--- a/tests/integration-tests/tests/arm_pl/test_arm_pl.py
+++ b/tests/integration-tests/tests/arm_pl/test_arm_pl.py
@@ -10,6 +10,7 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
+import re
 
 import pytest
 from assertpy import assert_that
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/arm_pl/test_arm_pl.py
```diff
@@ -58,13 +61,22 @@ def _test_armpl_examples(
...
+    # Clean up ANSI escape sequences from the output
+    eula_path = re.compile(r"\x1b[^m]*m| \x08").sub("", eula_path)
+    eula_path_result = remote_command_executor.run_remote_command("ls {0}".format(eula_path)).stdout
+    assert_that(eula_path_result).contains("license_agreement.txt")
+    assert_that(eula_path_result).contains("third_party_licenses.txt")
+
     # Assert pass the example tests
     remote_command_executor.run_remote_command(
-        f"sudo chmod 777 /opt/arm/armpl/{armpl_version}/armpl_{armpl_version}_gcc-{gcc_version}/examples"
+        f"sudo chmod 777 /opt/arm/armpl/{armpl_version}/armpl_{armpl_major_minor_version}_gcc-{gcc_version}/examples"
```
yuleiwan, line 74: Is there a way to determine we are using major_minor version or full version? So we needn't change test again in next time. For example, check the patch version of full version. If it is 0, then we use major_minor_version, otherwise we use full version

yuleiwan: but it is fine since we need to change module file as well.

yuleiwan (approved):

PR #2663, opened 2021-04-22
3.0: Add build_status tag during AMI build process

lukeseawalker:
Add parallelcluster:build_status tag at the end of the build process and fail if not able to set the tag
When checking for instance by AMI id, exclude instances that are terminated or shutting-down

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -13,7 +13,7 @@ from typing import List
 from pcluster import utils
 from pcluster.aws.aws_resources import ImageInfo, InstanceTypeInfo
 from pcluster.aws.common import AWSClientError, AWSExceptionHandler, Boto3Client, ImageNotFoundError
-from pcluster.constants import PCLUSTER_IMAGE_NAME_TAG, SUPPORTED_ARCHITECTURES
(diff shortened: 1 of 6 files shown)


cli/src/pcluster/aws/ec2.py
```diff
@@ -97,28 +97,35 @@ def describe_images(self, ami_ids, filters, owners):
...
         try:
-            self.describe_image_by_name_tag(image_name)
+            self.describe_image_by_name_tag(image_name, build_status_avaliable)
             return True
         except ImageNotFoundError:
             return False
 
     @AWSExceptionHandler.handle_client_exception
-    def describe_image_by_name_tag(self, image_name):
+    def describe_image_by_name_tag(self, image_name: str, build_status_avaliable: bool = False):
```
yuleiwan, line 109: I am concern if we don't pass `build_status_avaliable=true` in image property of imagebuilder.py. Because describe-image and list-image depend on that property to return ImageBuilderInfo. I notice we should delete image even if it doesn't have build_status_avaliable tag, one possible solution is to add a boolean is_available in ImageInfo, and use this boolean to determine image has build_status_avaliable or not.

yuleiwan (approved):
LGTM, fix the format before merge

cli/src/pcluster/aws/ec2.py
lukeseawalker: list-image fixed taking into account the new `build_status` tag

yuleiwan (dismissed):

yuleiwan (approved):

PR #2669, opened 2021-04-23
3.0: Modify generate-ami-list utility to work with simple json and not with CFN template

enrico-usai:
In ParallelCluster 3.x the CFN template no longer contains ami list. We're saving these information in a generic json file.

With this patch I'm also restoring the upload-cfn-templates.py that is required to upload networking templates, used by configure command.

### DIFF
--- a/util/generate-ami-list.py
+++ b/util/generate-ami-list.py
@@ -16,8 +16,6 @@
 #
 # Search for AWS ParallelCluster AMIs and generate a list in json and txt format
 #
-# usage: ./generate-ami-list.py --version <aws-parallelcluster-version> --date <release-date>
-
 import json
 import re
 from collections import OrderedDict
@@ -40,11 +38,7 @@ ARCHITECTURES_TO_MAPPING_NAME = {"x86_64": "AWSRegionOS2AMIx86", "arm64": "AWSRe
 
 
 def get_initialized_mappings_dicts():
-    """
-    Get dict with two keys initialized to empty dicts.
-
-    This is the same structure as the portion of the CFN template's Mappings that contains default AMIs.
-    """
+    """Get dict with two keys initialized to empty dicts."""
     return {mapping_name: {} for _, mapping_name in ARCHITECTURES_TO_MAPPING_NAME.items()}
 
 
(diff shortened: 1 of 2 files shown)


util/generate-ami-list.py
```diff
@@ -240,60 +250,35 @@ def get_all_aws_regions_from_ec2(region):
...
-
-
-def update_amis_txt(amis_txt_file, cfn_template_file):
-    """Write amis_txt_file using the updated information contained in cfn_template_file."""
-    cfn_data = read_cfn_template(cfn_template_file)
-    amis_txt = convert_json_to_txt(cfn_data.get("Mappings"))
+        json_file.write("\n")
+
+
+def write_amis_txt(amis_txt_file, json_file):
```
lukeseawalker, line 269: I think this needs to be remover as well. If mapping was removed from template also the ami.txt must be removed

util/generate-ami-list.py
```diff
@@ -310,25 +295,23 @@ def parse_args():
...
     local_file_group.add_argument(
-        "--json-template", type=str, help="path to input json cloudformation template", required=False
+        "--json-amis", type=str, help="path to input json file containing the AMIs", required=False
     )
     local_file_group.add_argument(
         "--json-regions", type=str, help="path to input json file containing the regions", required=False
     )
+
+    # General inputs and outputs of the utility
     parser.add_argument("--txt-file", type=str, help="txt output file path", required=False, default="amis.txt")
```
lukeseawalker, line 309: not needed if txt file is removed

util/generate-ami-list.py
```diff
@@ -363,15 +346,15 @@ def main():
...
-        # into a single CFN template.
+        # into a single json file.
         regions = get_aws_regions_from_file(args.json_regions)
-        amis_dict = get_ami_list_from_file(regions, args.json_template)
+        amis_dict = get_ami_list_from_file(regions, args.json_amis)
 
-    update_cfn_template(cfn_template_file=args.cloudformation_template, amis_to_update=amis_dict)
-    update_amis_txt(amis_txt_file=args.txt_file, cfn_template_file=args.cloudformation_template)
+    write_json_file(json_file_path=args.json_file, json_data=amis_dict)
+    write_amis_txt(amis_txt_file=args.txt_file, json_file=args.json_file)
```
lukeseawalker, line 357: not needed if txt file is removed

lukeseawalker (approved):
we agreed offline to remove the amis.txt in another PR

PR #2679, opened 2021-04-27
3.0: Improved API error handling, logging and validation

demartinofra:
* Move errors.py outside of the models dir: leaving only auto-generated classes in models dir
* Add validation for supported AWS regions in CLI
* Add region validator in API controllers
* Setting AWS_DEFAULT_REGION in lambda entrypoint
* Mask fatal error exception message from Lambda response
* Improve logging and error handling in flask app
* Add API dependencies to default requirements: this will be required by pcluster cli


cli/src/pcluster/api/flask_app.py
```diff
@@ -46,51 +48,96 @@ class ParallelClusterFlaskApp:
     def __init__(self, swagger_ui: bool = False, validate_responses=False):
         options = {"swagger_ui": swagger_ui}
 
-        app = connexion.FlaskApp(__name__, specification_dir="openapi/", skip_error_handlers=True)
-        app.app.json_encoder = encoder.JSONEncoder
-        app.add_api(
+        self.app = connexion.FlaskApp(__name__, specification_dir="openapi/", skip_error_handlers=True)
```
enrico-usai, line 51: minor: what about adding another name for this self.app? To avoid having self.app.app

enrico-usai (approved):

cli/src/pcluster/api/flask_app.py
demartinofra: makes sense. let me do this as part of the next PR

PR #2712, opened 2021-05-06
[3.0] Copy build_status tag in distribution regions

lukeseawalker:

*Issue #, if available:*

*Description of changes:*


cli/src/pcluster/resources/custom_resources/custom_resources_code/delete_image_stack.py
```diff
@@ -43,27 +69,7 @@ def handler(event, context):
...
-                    Tags={
-                        "parallelcluster:build_status": "available",
-                        "parallelcluster:parent_image": parent_image,
-                    },
-                )
-            except KeyError as e:
-                logger.error("Failed to parse message with exception: %s", e)
-            except ClientError as e:
-                logging.error("Tagging EC2 AMI %s failed with exception: %s", image_arn, e)
+            _tag_ami(message_json)
```
yuleiwan, line 72: Should we only tag build_status=available to the ami with stack including latest test component(TagComponent), or with stack including PclusterComponent. Because I think we shouldn't tag  build_status=available to output AMI of first stage, or AMI built with DisablePclusterComponent: True in DevSettings.

lukeseawalker: we should always tag with `build_status=available` otherwise the commands logic that rely on this tag (build, describe, list) will not going to work properly

yuleiwan: why we need to display AMIs with DisablePclusterComponent: True in describe, list commands? To retrieve first stage AMIs, we can use awscli with name and tag filter rather than describe command.

cli/src/pcluster/resources/custom_resources/custom_resources_code/delete_image_stack.py
```diff
@@ -21,6 +21,32 @@
 boto3_config = Config(retries={"max_attempts": 60})
 
 
+def _tag_ami(message_json):
+    # tag EC2 AMI
+    for ami in message_json["outputResources"]["amis"]:
```
yuleiwan, line 26: why not put this line into try ... except? So we can handle the key error of message_json["outputResources"]["amis"]

lukeseawalker: done

yuleiwan (approved):

PR #2714, opened 2021-05-06
3.0: Add SAM template to create the API infrastructure

ermanno:
### Notes
This patch adds the SAM template that creates the ParallelCluster API infrastructure. The current template supports:
1. Creating the Lambda function with Administrator permissions (to be restricted in the future)
2. Creating an APIGateway API, eventually configuring a custom domain name
3. Eventually creating a copy of the public Docker image for the Lambda function using ImageBuilder (at the moment we still do not support automatic deletion of the image, will be added in the future in a custom resource)

### Tests
Performed the following manual tests:
1. API Gateway API without custom domain name
2. API Gateway API with custom domain name
3. Using a provided image URI
4. Using the "public image URI" (creation of new repository and copy of the "public image", for the purposes of this test simply an image in a private ECR repository, to be substituted with the public image once we publish it to the public ECR repository)

### DIFF
--- a/api/infrastructure/parallelcluster-api.yaml
+++ b/api/infrastructure/parallelcluster-api.yaml
@@ -1,0 +1,250 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: 'Template for the ParallelCluster API'
+
+Parameters:
+  EcrImageUri:
+    Description: When specified use this image for the Lambda function and skip the import phase
+    Type: String
+    Default: ''
+
+  ApiDefinitionS3Uri:
+    Description: S3 URI of the ParallelCluster API spec
+    Type: String
+    Default: '' # TODO change with actual URI
+
+  CustomDomainName:
+    Description: When specified, the custom domain name of the ParallelCluster API. Requires specifying a custom domain certificate
+    Type: String
+    Default: ''
+
+  CustomDomainCertificate:
+    Description: When specified, the ARN of the certificate for the custom domain name of the ParallelCluster API. Required when specifying a custom domain name
+    Type: String
+    Default: ''
+
+  CustomDomainHostedZoneId:
+    Description: When specified, the id of the Hosted Zone where the custom domain record of the ParallelCluster API is registered
+    Type: String
+    Default: ''
+
+  PublicEcrImageUri:
+    Description: When specified, the URI of the Docker image for the Lambda of the ParallelCluster API
+    Type: String
+    Default: '' # TODO change with actual URI
+
+
+Mappings:
+  ParallelCluster:
+    Constants:
+      Version: 3.0.0
+      Stage: prod
+
+
+# More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
+Globals:
+  Function:
+    Timeout: 30
+    MemorySize: 256
+    Tags:
+      'parallelcluster:version': !FindInMap [ParallelCluster, Constants, Version]
+
+  Api:
+    Auth:
+      DefaultAuthorizer: AWS_IAM
+    TracingEnabled: True
+    EndpointConfiguration:
+      Type: REGIONAL
+
+
+Conditions:
+  UseCustomEcrImageUri: !Not [!Equals [!Ref EcrImageUri, '']]
+  DoNotUseCustomEcrImageUri: !Not [!Condition UseCustomEcrImageUri]
+  UseCustomDomain: !Not [!Equals [!Ref CustomDomainName, '']]
+  UseRoute53Configuration: !Not [!Equals [!Ref CustomDomainHostedZoneId, '']]
+  UseCustomDomainAndRoute53Configuration: !And
+    - !Condition UseCustomDomain
+    - !Condition UseRoute53Configuration
+  DoNotUseCustomDomain: !Not [!Condition UseCustomDomain]
+
+
+Resources:
+
+  # We need to define two AWS::Serverless::Api due to an issue with the handling of AWS::NoValue
+  # See related GitHub issue: https://github.com/aws/serverless-application-model/issues/1435
+  ApiGatewayApiWithCustomDomainAndRoute53Configuration:
+    Condition: UseCustomDomainAndRoute53Configuration
+    Type: AWS::Serverless::Api
+    Properties:
+      Tags:
+        'parallelcluster:version': !FindInMap [ParallelCluster, Constants, Version]
+      StageName: !FindInMap [ParallelCluster, Constants, Stage]
+      DefinitionBody:
+        Fn::Transform:
+          Name: AWS::Include
+          Parameters:
+            Location: !Ref ApiDefinitionS3Uri
+      Domain:
+        DomainName: !Ref CustomDomainName
+        CertificateArn: !Ref CustomDomainCertificate
+        Route53:
+          HostedZoneId: !Ref CustomDomainHostedZoneId
+
+  ApiGatewayApiWithoutCustomDomain:
+    Condition: DoNotUseCustomDomain
+    Type: AWS::Serverless::Api
+    Properties:
+      Tags:
+        'parallelcluster:version': !FindInMap [ParallelCluster, Constants, Version]
+      StageName: !FindInMap [ParallelCluster, Constants, Stage]
+      DefinitionBody:
+        Fn::Transform:
+          Name: AWS::Include
+          Parameters:
+            Location: !Ref ApiDefinitionS3Uri
+
+  # Note that even for Chinese regions here we need to use apigateway.amazonaws.com instead of apigateway.amazonaws.com.cn
+  APIGatewayExecutionRole:
+   Type: AWS::IAM::Role
+   Properties:
+     AssumeRolePolicyDocument:
+       Version: 2012-10-17
+       Statement:
+         - Effect: Allow
+           Principal:
+             Service:
+               - apigateway.amazonaws.com
+           Action:
+             - 'sts:AssumeRole'
+     Policies:
+       - PolicyName: lambda-invoke
+         PolicyDocument:
+           Version: '2012-10-17'
+           Statement:
+             - Effect: Allow
+               Action: lambda:InvokeFunction
+               Resource: !GetAtt ParallelClusterFunction.Arn
+
+  ParallelClusterFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Policies:
+        - AdministratorAccess
+      Tracing: Active
+      PackageType: Image
+      ImageUri: !If
+        - UseCustomEcrImageUri
+        - !Ref EcrImageUri
+        - !Sub
+          - ${AWS::AccountId}.dkr.ecr.${AWS::Region}.${AWS::URLSuffix}/${Repository}:${Version}
+          - Repository: !Ref PrivateEcrRepository
+            Version: !Join
+              - '-'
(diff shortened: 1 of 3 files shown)


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+
+  Api:
+    Auth:
+      DefaultAuthorizer: AWS_IAM
+    TracingEnabled: True
+    EndpointConfiguration:
+      Type: REGIONAL
+
+
+Conditions:
```
demartinofra, line 64: nit: I would use an imperative form for the conditions name: e.g. UseCustomEcrImageUri rather than Uses

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+
+  # We need to define three AWS::Serverless::Api due to an issue with the handling of AWS::NoValue
+  # See related GitHub issue: https://github.com/aws/serverless-application-model/issues/1435
+
+  ApiGatewayApiWithOnlyCustomDomain:
+    Condition: UsesOnlyCustomDomain
+    Type: AWS::Serverless::Api
+    Properties:
+      Tags:
+        parallelcluster: !Ref ParallelClusterVersion
```
demartinofra, line 88: let's use `parallelcluster:version` as the tag key. Also we might want to add this to the globals

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Properties:
+      PackageType: Image
+      Tracing: Active
+      Tags:
+        parallelcluster: !Ref ParallelClusterVersion
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSXRayDaemonWriteAccess
+        - AdministratorAccess
+      ImageUri: !If [UsesCustomEcrImageUri, !Ref EcrImageUri, !Ref PublicEcrImageUri]
```
demartinofra, line 163: In the else clause we probably want to use the imported image not the public one

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+
+  LambdaLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      RetentionInDays: 30
+
+  InstanceRole:
```
demartinofra, line 170: let's be more specific with something like `ImageBuilderInstanceRole`. same for InstanceProfile

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Type: AWS::ImageBuilder::Image
+    Properties:
+      ContainerRecipeArn: !Ref EcrImageRecipe
+      EnhancedImageMetadataEnabled: true
+      InfrastructureConfigurationArn: !Ref InfrastructureConfiguration
+      ImageTestsConfiguration:
+        ImageTestsEnabled: false
+
+
+Outputs:
```
demartinofra, line 240: Shall we add as an output the uri of the pushed ECR image?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Properties:
+      Name: ParallelClusterImageBuilderInfrastructureConfiguration
+      InstanceProfileName: !Ref InstanceProfile
+      TerminateInstanceOnFailure: true
+
+  PrivateEcrRepository:
+    Condition: DoesNotUseCustomEcrImageUri
+    Type: AWS::ECR::Repository
+    Properties:
+      RepositoryName: !Sub
```
demartinofra, line 208: I know I was the one suggesting this in the first place but probably better to remove this name and just tag the repo so to avoid conflicts if creating multiple API stacks in the same region

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+        - 'aws-parallelcluster-${Version}'
+        - { Version: !Ref ParallelClusterVersion }
+
+  EcrImageRecipe:
+    Condition: DoesNotUseCustomEcrImageUri
+    Type: AWS::ImageBuilder::ContainerRecipe
+    Properties:
+      Components:
+        - ComponentArn: !Sub arn:aws:imagebuilder:${AWS::Region}:aws:component/update-linux/1.0.0
```
demartinofra, line 217: what do we need this for? is it to have at least 1 component?I'm wondering what is this component updating and if we can have a no-op component instead

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Description: When specified, the id of the Hosted Zone of the custom domain name of the ParallelCluster API
+    Type: String
+    Default: ''
+
+  PublicEcrImageUri:
+    Description: When specified, the URI of the Docker image for the Lambda of the ParallelCluster API
+    Type: String
+    Default: uri_of_public_ecr_image # TODO change with actual URI
+
+  ParallelClusterVersion:
```
demartinofra, line 44: I'm wondering if we should move this to a Mapping rather than a Parameter so that the user cannot change this

demartinofra (commented):
LVGTM! left a few minor comments

api/infrastructure/parallelcluster-api.yaml
ermanno: Agreed, the `ParallelClusterVersion` should be a "constant", not a parameter that the customer should specify. Question, should we also make the following a Mapping?
* `ApiDefinitionS3Uri`
* `PublicEcrImageUri`

api/infrastructure/parallelcluster-api.yaml
ermanno: Afaik yes, we need to specify at least one component and there is no no-op component. Also, in practice this should not really be used since in the template section I don't specify `{{{ imagebuilder:components }}}`. Ideally I would like to remove this section, I have reached out to the ImageBuilder team to ask for clarification

api/infrastructure/parallelcluster-api.yaml
ermanno: The `Tags` property can be specified in the `Globals` section for `AWS::Serverless::Function`, but not for `AWS::Serverless::Api` (see [documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html)), I will change the name and put it in the `Globals` section for the `Function` resource type

api/infrastructure/parallelcluster-api.yaml
demartinofra: I believe that those will be needed as parameter to test the develop version

api/infrastructure/parallelcluster-api.yaml
ermanno: In light of our discussion regarding the names of `AWS::ImageBuilder::InfrastructureConfiguration` and `AWS::ImageBuilder::ContainerRecipe`, what about following the same pattern and putting `aws-parallelcluster-${Version}-${StackIdSuffix}`?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Properties:
+      Path: /executionServiceEC2Role/
+      Roles:
+        - !Ref InstanceRole
+
+  InfrastructureConfiguration:
+    Condition: DoesNotUseCustomEcrImageUri
+    Type: AWS::ImageBuilder::InfrastructureConfiguration
+    Properties:
+      Name: ParallelClusterImageBuilderInfrastructureConfiguration
```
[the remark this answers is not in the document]
ermanno: As agreed, we will adopt the convention of appending the Stack id suffix (part after the last /)

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,246 @@
...
+        - { Version: !Ref ParallelClusterVersion }
+
+  EcrImageRecipe:
+    Condition: DoesNotUseCustomEcrImageUri
+    Type: AWS::ImageBuilder::ContainerRecipe
+    Properties:
+      Components:
+        - ComponentArn: !Sub arn:aws:imagebuilder:${AWS::Region}:aws:component/update-linux/1.0.0
+      ContainerType: DOCKER
+      Name: ImportPublicEcrImage
```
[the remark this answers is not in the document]
ermanno: As agreed, we will adopt the convention of appending the Stack id suffix (part after the last /)

api/infrastructure/parallelcluster-api.yaml
ermanno: I have validated with the ImageBuilder team that the behavior corresponds to what I have detailed above

ermanno:
### Updates
1. Tested creation of multiple stacks
2. Return APIGateway API URI (if template not created with custom domain name)
3. Return URI of ECR image copied from the "public" image (if a custom ECR image uri was not passed in as a paremeter)

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,248 @@
...
+      Tracing: Active
+      Policies:
+        - arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess
+        - AWSLambdaBasicExecutionRole
+        - AdministratorAccess
+      ImageUri: !If
+        - UseCustomEcrImageUri
+        - !Ref EcrImageUri
+        - !Sub
+          - ${AWS::AccountId}.dkr.ecr.${AWS::Region}.amazonaws.com/${Repository}:${Version}
```
demartinofra, line 140: Can you check if this needs the `.cn` suffix in cn regions?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,248 @@
...
+    Value: !GetAtt ParallelClusterFunction.Arn
+  ParallelClusterApiInvokeUrl:
+    Description: 'Url to reach the API endpoint'
+    Value: !If
+      - UseCustomDomain
+      - !Sub
+        - https://${CustomDomain}
+        - { CustomDomain: !Ref CustomDomainName }
+      - !Sub
+        - https://${Api}.execute-api.${AWS::Region}.amazonaws.com/${StageName}
```
demartinofra, line 238: check domain in cn regions

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -0,0 +1,248 @@
...
+        - { Api: !Ref ApiGatewayApiWithoutCustomDomain, StageName: !FindInMap [ParallelCluster, Constants, Stage] }
+  UriOfCopyOfPublicEcrImage:
+    Condition: DoNotUseCustomEcrImageUri
+    Description: 'Uri of the copy of the Public ParallelCluster API Lambda Container image'
+    Value: !Sub
+      - ${AWS::AccountId}.dkr.ecr.${AWS::Region}.amazonaws.com/${Repository}:${Version}
```
demartinofra, line 244: same for this one

demartinofra (approved):
Well done! 🚀

demartinofra (approved):

ermanno:
### Updates
Following updates for Chinese regions
1. `apigateway.amazonaws.com` instead of `apigateway.amazonaws.com.cn` https://github.com/serverless/serverless/issues/5365
2. No Lambda and XRay policies in APIGateway execution role, because they are already added by SAM (based on the presence of tracing=true, etc.) and in Chinese regions CFN complains that the policies are duplicated

PR #2737, opened 2021-05-12
3.0: Add custom resource to delete the Docker image on stack deletion

ermanno:
### Notes
The `AWS::ImageBuilder::Image` resource adds an image to the ECR repository, but when the stack is deleted this image is not removed, and this causes a failure due to the fact that we cannot delete an ECR repository that is not empty. This patch introduces a custom resource that removes the created image, thus allowing the successful deletion of the CloudFormation stack

### Tests
Tested creation and deletion of stack in personal account

### Notes for reviewer
I would like to solicit comments regarding error handling for the Lambda of the custom resource. I see that in the `manage_docker_images.py` code we don't do anything in particular on that side (for example we don't catch the response of the delete, which could contain a failure in case the image was not present in the repository -which could happen if the customer has manually removed it, for example-), I was wondering if this is what we want


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -221,6 +221,93 @@ Resources:
...
+        - PolicyName: S3Policy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - ecr:BatchDeleteImage
+                Resource: !GetAtt PrivateEcrRepository.Arn
+
+  EcrImagesDeletor:
```
demartinofra, line 303: nit: not sure "deletor" is a correct word.

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -221,6 +221,93 @@ Resources:
...
+            Action:
+              - 'sts:AssumeRole'
+      Policies:
+        - PolicyName: LoggingPolicy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - logs:CreateLogGroup
```
demartinofra, line 290: granting CreateLogGroup permissions will cause this lambda to create a log group that will not be deleted. Shall we explicitly create a log group in the template?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -221,6 +221,93 @@ Resources:
...
+
+              Arn format: arn:aws:imagebuilder:${AWS::Region}:${AWS::AccountId}:image/.../${Version}/${BuildNumber}
+              Tag format: ${Version}-${BuildNumber}
+              """
+              parts = image_arn.split('/')
+              version = parts[-2]
+              build_number = parts[-1]
+              return '{}-{}'.format(version, build_number)
+
+          def handler(event, context):
```
demartinofra, line 244: Regarding error handling:
1. I agree with you that we can better handle the case where the ECR image has been deleted already without failing.
2. We should probably handle failures and send a failure response to CFN otherwise CFN will wait until the Lambda times-out
3. when sending a failure event to CFN can we also send the reason of the failure (e.g. the failure from the boto3 call)
4. Let's male sure to have the proper boto3 retry in place (the default might be fine but let's double-check it) cause once the lambda fails this cannot be retried by CFN.

ermanno: Regarding the retries, the client does 3 retries by default. Regrading the code, I have wrapped the handler in a try/catch and failed fast by returning a failure response and putting the exception in the status reason. I checked what was the Boto3 version, and verified that there is an exception that we might want to catch and "ignore" (i.e. succeeding anyway), namely [ECR.Client.exceptions.RepositoryNotFoundException](https://boto3.amazonaws.com/v1/documentation/api/1.17.42/reference/services/ecr.html#ECR.Client.batch_delete_image). Not sure about this, my current reasoning is:
1. In general if the customer deletes resources created by the stack it is correct to give an error, so we might want to fail deletion if we catch this exception
2. In the case of trying to remove an image that does not exist in the repository, the API does not give an exception, so we don't fail deletion: this is consistent with the current PC behavior elsewere, plus the Docker image is not a resource in the CFN template (it is "created indirectly" through the definition of the ImageBuilder image, which is the [AWS::ImageBuilder::Image
](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-imagebuilder-image.html) resource in the CFN template)

demartinofra: I think it's resonable never to fail deletion if resources are not present

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -221,6 +221,103 @@ Resources:
...
+                  response_data['Message'] = 'Resource creation successful!'
+              elif event['RequestType'] == 'Update':
+                  response_data['Message'] = 'Resource update successful!'
+              elif event['RequestType'] == 'Delete':
+                  try:
+                      handle_deletion_event(event)
+                      response_data['Message'] = 'Resource deletion successful!'
+                  except Exception as exception:
+                      response_status = cfnresponse.FAILURE
+                      response_data['Message'] = exception
```
demartinofra, line 269: Can we prepend something to the exception message such as "Failed when deleting bla bla with error bla"?

demartinofra (approved):

demartinofra (approved):

PR #2745, opened 2021-05-14
Add protected mode and integration tests

chenwany:
* Add new compute fleet status `PROTECTED` and unit tests.
* Add integration test for Slurm protected mode, test static and dynamic nodes with bootstrap failures will turn cluster into protected mode, test jobs can be run in healthy queue under the protected mode:
  * Create a healthy cluster with three queues.
  * Update the cluster with a pre-install script, with the script, the queue `queue broken` contains all nodes are broken,`queue half-broken` contains one compute resource is broken and one compute resource work as expected, `queue normal` contains dynamic and static node work as expected. 
 
 Tests cover the following logics:
* Bootstrap failures have no affect on cluster when protected mode is disabled.
* Cluster is not placed into protected mode when there is an active job running
* Cluster will be placed into protected mode when there's no active job running in problematic queue
* Under protected mode, job can run successfully in normal queue 
* After recover from protected mode, pending job that is in the queue when cluster entered protected mode will be run again when cluster is taken out of protected mode
* Jobs can be run successfully in all queues after recover from protected mode.

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -226,6 +226,34 @@ def test_error_handling(scheduler, region, instance, pcluster_config_reader, clu
     )
 
 
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+@pytest.mark.slurm_protected_mode
+def test_slurm_protected_mode(
+    scheduler, region, pcluster_config_reader, clusters_factory, test_datadir, s3_bucket_factory
+):
+    """Test that slurm protected mode logic can handle bootstrap failure nodes."""
+    # Create S3 bucket for pre-install scripts
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "preinstall.sh"), "scripts/preinstall.sh")
+    cluster_config = pcluster_config_reader(bucket=bucket_name)
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    # Disable protected_mode by setting protected_failure_count to -1
+    _set_protected_failure_count(remote_command_executor, -1)
+    _inject_bootstrap_failures(cluster, bucket_name, pcluster_config_reader)
+    _test_disable_protected_mode(remote_command_executor)
+    # Re-enable protected mode
+    _enable_protected_mode(remote_command_executor)
+    # Decrease protected failure count for quicker enter protected mode.
+    _set_protected_failure_count(remote_command_executor, 2)
+    pending_job_id = _test_active_job_running(scheduler_commands, remote_command_executor)
+    _test_protected_mode(scheduler_commands, remote_command_executor, cluster)
+    _test_job_run_in_working_queue(scheduler_commands)
+    _test_recover_from_protected_mode(pending_job_id, pcluster_config_reader, bucket_name, cluster, scheduler_commands)
+
+
 def _assert_cluster_initial_conditions(
     scheduler_commands,
     instance,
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -939,3 +963,152 @@ def _get_num_gpus_on_instance(instance_type_info):
...
+    )
+    remote_command_executor.run_remote_script(str(test_datadir / "slurm_restart_clustermgtd.sh"), run_as_root=True)
+
+
+def _test_disable_protected_mode(remote_command_executor, test_datadir):
+    """Test Bootstrap failures have no affect on cluster when protected mode is disabled."""
+    # Disable protected mode by setting protected failure count to -1
+    _set_protected_failure_count(remote_command_executor, test_datadir, -1)
+    # wait till the node failed
+    retry(wait_fixed=seconds(20), stop_max_delay=minutes(15))(assert_errors_in_logs)(
```
rexcsn, line 1015: Do we need to wait 15 mins? If it's just a pre-install failure that causes self-termination it should happen quickly, maybe we can try 5 or 7 mins instead?
Same comment applies to everywhere that we are using the 15 mins timeout

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -939,3 +963,152 @@ def _get_num_gpus_on_instance(instance_type_info):
...
+
+def _test_active_job_running(scheduler_commands, remote_command_executor):
+    """Test cluster is not placed into protected mode when there is an active job running even reach threshold."""
+    # Submit a job to the queue contains broken nodes and normal node, submit the job to the normal node to test
+    # the queue will not be disabled if there's active job running.
+    cancel_job_id = scheduler_commands.submit_command_and_assert_job_accepted(
+        submit_command_args={"command": "sleep 3000", "nodes": 1, "partition": "half-broken", "constraint": "c5.xlarge"}
+    )
+    # Sleep for a bit for the job to run, also for some iteration of clustermgtd to change node states
+    time.sleep(120)
```
rexcsn, line 1040: Do we need to do this? I would imagine the problematic job would take a bit of time to fail anyway, so maybe we can just get rid of this sleep time?

rexcsn, line 1040: Do we need to sleep here? I would imagine the bootstrap failure job would take some mins to fail anyway?

tests/integration-tests/tests/schedulers/test_awsbatch.py
```diff
@@ -24,7 +24,7 @@
 @pytest.mark.instances(["c5.xlarge", "c4.xlarge", "m6g.xlarge"])
 @pytest.mark.dimensions("*", "c5.xlarge", "alinux2", "awsbatch")
 @pytest.mark.dimensions("cn-north-1", "c4.xlarge", "alinux2", "awsbatch")
-@pytest.mark.dimensions("ap-northeast-1", "m6g.xlarge", "alinux2", "awsbatch")
+@pytest.mark.dimensions("us-east-1", "c5.xlarge", "alinux2", "awsbatch")
```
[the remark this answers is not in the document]
chenwany: No, removed

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -226,6 +226,30 @@ def test_error_handling(scheduler, region, instance, pcluster_config_reader, clu
...
+    # Create S3 bucket for pre-install scripts
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "preinstall.sh"), "scripts/preinstall.sh")
+    cluster_config = pcluster_config_reader(bucket=bucket_name)
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    _inject_bootstrap_failures(cluster, bucket_name, pcluster_config_reader)
+    _test_disable_protected_mode(remote_command_executor, test_datadir)
```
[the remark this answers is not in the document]
chenwany: Changed

tests/integration-tests/tests/schedulers/test_slurm.py
chenwany: Set them to 7 minutes

tests/integration-tests/tests/schedulers/test_slurm.py
chenwany: If two jobs submit together, the normal node maybe not ready to start before the pre-install bootstrap failure happen, have been reduce it to 15, fail, now reduced the sleep time here to 60.

rexcsn: Sounds like we should assert that the normal node is running the job before submitting the job with bootstrap failure to enforce the sequence of events, instead of just using some arbitrary sleep time here.

chenwany: Added a function to to wait for job running

rexcsn (approved):
Good job!

PR #2750, opened 2021-05-14
Fix issues with retrieving pcluster AMIs for createami tests

rexcsn:


tests/integration-tests/tests/common/utils.py
```diff
@@ -55,17 +55,13 @@
...
 
-def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64"):
+def retrieve_latest_ami(region, os, amis_dict, ami_type="official", architecture="x86_64"):
     try:
         if ami_type == "pcluster":
-            ami_name = "aws-parallelcluster-{version}-{ami_name}".format(
-                version=get_installed_parallelcluster_version(),
-                ami_name=AMI_TYPE_DICT.get(ami_type).get(os).get("name"),
-            )
+            return amis_dict[os]
```
yuleiwan, line 64: amis_dict.get(os) would be better, since we catch index error here

yuleiwan (approved):

PR #2775, opened 2021-05-24
3.0: Add Lambda to push AWS credentials to the head node for the cluster admin user

gmarciani:
### Changes
1. Add Lambda to push AWS credentials to the head node for the cluster admin user

#### Important Notes:
1. This PR requires [Cookbook:PR-968](https://github.com/aws/aws-parallelcluster-cookbook/pull/968) to be merged first.
2. Pre-commit checks reformatted some changelog lines.

### Tests
1. Tested on personal account.
2. Jenkins tests pending. Going to start them after code review.

### DIFF
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -15,6 +15,7 @@ from typing import List, Union
 
 import pkg_resources
 from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_events as events
 from aws_cdk import aws_iam as iam
 from aws_cdk import aws_lambda as awslambda
 from aws_cdk import aws_logs as logs
@@ -281,10 +282,18 @@ class PclusterLambdaConstruct(Construct):
         execution_role: iam.CfnRole,
         handler_func: str,
         timeout: int = 900,
+        memory_size: int = 128,
+        environment=None,
+        event_pattern=None,
+        schedule_expression: str = None,
+        schedule_enabled: bool = False,
     ):
         super().__init__(scope, id)
 
         function_name = f"pcluster-{function_id}-{self._stack_unique_id()}"
+
+        if environment is None:
+            environment = {}
 
         self.log_group = logs.CfnLogGroup(
             scope,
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -354,6 +385,113 @@ def _add_cleanup_resources_lambda(self):
...
+            environment={
+                "CLUSTER_NAME": _cluster_name,
+                "ROLE_ARN": self.cluster_admin_role.attr_arn,
+                "SSM_CW_LOG_GROUP_ENABLED": str(self.config.is_cw_logging_enabled),
+                "SSM_CW_LOG_GROUP_NAME": f"/aws/ssm/{_cluster_name}",
+            },
+            event_pattern={
+                "source": ["aws.ec2"],
+                "detail-type": ["EC2 Instance State-change Notification"],
+                "detail": {"state": ["running"], "instance-id": ["i-0fd6e84f8762a1675"]},
```
tilne, line 489: should there be a ref to the head node rather than a hard-coded instance ID?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -302,12 +311,38 @@ def __init__(
...
-            memory_size=128,
+            memory_size=memory_size,
             role=execution_role,
             runtime="python3.8",
             timeout=timeout,
+            environment=awslambda.CfnFunction.EnvironmentProperty(variables=environment),
         )
 
+        if schedule_expression is not None:
+            self.lambda_rule = events.CfnRule(
```
tilne, line 322: is there a reason `CfnRule` has to be used instead of the `Rule` construct? probably there is but i'm admittedly not very familiar with this part of the 3.0 code.

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I'm sorry: I forgot the hard wired testing value.

cli/src/pcluster/templates/cdk_builder_utils.py
gmarciani: Generally speaking, Cfn* constructs are preferable over higher level constructs when you want more control over the final template. On the other hand, higher level constructs are more comfortable because they exposes handy signatures and take care of some underlying resources. I think that, when it is possible Cfn* constructs are preferable aklso because, at the end, higher level constructs are translated to low level ones.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -354,6 +385,113 @@ def _add_cleanup_resources_lambda(self):
...
+            environment={
+                "CLUSTER_NAME": _cluster_name,
+                "ROLE_ARN": self.cluster_admin_role.attr_arn,
+                "SSM_CW_LOG_GROUP_ENABLED": str(self.config.is_cw_logging_enabled),
+                "SSM_CW_LOG_GROUP_NAME": f"/aws/ssm/{_cluster_name}",
+            },
+            event_pattern={
+                "source": ["aws.ec2"],
+                "detail-type": ["EC2 Instance State-change Notification"],
+                "detail": {"state": ["running"], "instance-id": [self.head_node_instance.ref]},
```
gmarciani, line 489: **Problem**
Since `event_pattern` uses `head_node_instance.ref`, the event implicitly depends on the head node instance. 
So, head node instance is created before the rule and it may happen that the rule is not triggered because the instance could have been gone running before rule creation.

tilne: would a custom resource that waits for the head node to come up solve this?

gmarciani: Yes

PR #2776, opened 2021-05-24
3.0: Implement the CreateCluster API

demartinofra:
This PR implements the new CreateCluster API as defined in the API spec file: https://github.com/aws/aws-parallelcluster/blob/wip/pcluster3/api/spec/openapi/ParallelCluster.openapi.yaml

### Commits in this PR
* **Move usage of dynamodb client to aws package**: move boto3 logic present in ComputeFleetStatusManager to interact with DynamoDB to the AWSApi class.
* **Use version id when retrieving config in DescribeCluster API**: change DescribeCluster API to return a presigned S3 url for the versioned config file.
* **Add error code to handle dryrun operations in API**: respond with HTTP 412 status code when a request would have succeeded if the dryrun flag was not set.
* **Check nodejs is installed and use jsonify to encode errors**: Using jsonify from Flask enables the usage of the custom JSON encoder when encoding Model objects to Json.
* **Add logging for API request/response body**: dump request/response body in API logs.
* **Update suppressValidators in API model**: update suppressValidators format to match `"^(ALL|type:[A-Za-z0-9]+)$"`
* **Implement CreateCluster API**

### API Example
```
curl --location --request POST 'http://{{endpoint}}/v3/clusters' \
--header 'Content-Type: application/json' \
--data-raw '{
    "clusterConfiguration": "{{Base64EncodedConfig}}",
    "name": "clustername",
    "region": "eu-west-1"
}'
```

```
{
    "cluster": {
        "cloudformationStackArn": "arn:aws:cloudformation:eu-west-1:***",
        "clusterName": "clustername",
        "clusterStatus": "CREATE_IN_PROGRESS",
        "region": "eu-west-1",
        "version": "3.0.0"
    },
    "validationMessages": [
        {
            "level": "WARNING",
            "message": "Instance type 'c5.2xlarge' does not support EFA.",
            "type": "EfaValidator"
        }
    ]
}
```

***


cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -69,22 +79,54 @@ def create_cluster(
...
+    rollback_on_failure = rollback_on_failure or True
+    validation_failure_level = validation_failure_level or ValidationLevel.ERROR
+    dryrun = dryrun or False
     create_cluster_request_content = CreateClusterRequestContent.from_dict(create_cluster_request_content)
-    if create_cluster_request_content.cluster_configuration == "invalid":
-        raise CreateClusterBadRequestException(
-            CreateClusterBadRequestExceptionResponseContent(configuration_validation_errors=[], message="invalid")
+
+    # Validate inputs
+    if client_token:
```
ermanno, line 89: I see that we also have this parameter for the `create_image` and `delete_image` APIs, I suppose that we want to raise an exception there as well. How are we planning to use this in the future? Which part of the workflow would need it? I notice that the implementation in the `pcluster_api.py` file does not make use of it

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -49,7 +59,7 @@
 def create_cluster(
```
ermanno, line 59: In general, why do we re-imlement this logic here rather than reuse (eventually with some changes) the code in the `Cluster` class defined in `cli/src/pcluster/models/cluster.py`? This is the code that is used by the `pcluster_api.py` implementation of `create_cluster`

cli/src/pcluster/api/controllers/cluster_operations_controller.py
demartinofra: we could also remove it from the spec for now. Propagating this token to the CFN create_stack does not seem to have the desired behavior

cli/src/pcluster/api/controllers/cluster_operations_controller.py
demartinofra: I tried to move to the controller the logic required to handle API specific errors. Also I did not want to heavily modify the cluster.py class at the moment to avoid impacting current CLI functionalities that are not hooked yet to the API call

cli/src/pcluster/api/flask_app.py
```diff
@@ -78,61 +96,73 @@ def _clear_cache():
...
-            LOGGER.info("Handling request: %s %s", request.method, request.full_path)
+            LOGGER.info(
+                "Handling request: %s %s - Body: %s",
+                request.method,
+                request.full_path,
+                request.get_json() if request.data else "EMPTY",
+            )
 
         @self.flask_app.after_request
         def _log_response(response: Response):  # pylint: disable=unused-variable
```
enrico-usai, line 107: minor: you can remove pylint comment

cli/src/pcluster/api/flask_app.py
```diff
@@ -78,61 +96,73 @@ def _clear_cache():
 
         @self.flask_app.before_request
         def _log_request():  # pylint: disable=unused-variable
```
enrico-usai, line 98: minor: useless pylint comment

cli/src/pcluster/config/common.py
```diff
@@ -125,38 +167,47 @@ def init_param(value, default=None, update_policy=None):
...
+                        self._validation_failures.extend(item.validate(validator_suppressor_list))
 
         # Update validators to be executed according to current status of the model and order by priority
-        self._validate()
+        self._validators.clear()
+        self._register_validators()
+        for validator_class, validator_args in self._validators:
+            validator = validator_class()
+            if validator_suppressor_list and any(
+                suppressor.suppress_validator(validator) for suppressor in validator_suppressor_list
```
enrico-usai, line 192: what about `is_enabled`, or `is_suppressed`, or `is_validator_suppressed`? 
`suppressor.suppress_validator(...)` seems more an action to be performed than a check.

enrico-usai (approved):
I just left minor comments, great job!

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -69,22 +79,54 @@ def create_cluster(
...
+                cloudformation_stack_arn=stack_id,
+                region=os.environ.get("AWS_DEFAULT_REGION"),
+                version=get_installed_version(),
+                cluster_status=cloud_formation_status_to_cluster_status(CloudFormationStatus.CREATE_IN_PROGRESS),
+            ),
+            validation_messages=_build_validation_messages_list(ignored_validation_failures) or None,
+        )
+    except ConfigValidationError as e:
+        raise _handle_config_validation_error(e)
+    except ClusterActionError as e:
```
ermanno, line 124: Asking because this is also relevant to my work on the `DeleteImage`. The `Cluster` class wraps exceptions in a `ClusterActionError`, which is surfaced as a 500, however I note that this might also be caused by `AWSClientError`, for example if the `AWSApi.instance().cfn.create_stack_from_url` or `AWSApi.instance().cfn.describe_stack` have a `REQUEST_LIMIT_EXCEEDED`, which I think we would like to surface as a 429 `LimitExceededException`

demartinofra:
Close in favor of https://github.com/aws/aws-parallelcluster/pull/2855

PR #2777, opened 2021-05-24
integ-tests: restore old creds in env when unsetting temporary STS creds

demartinofra:
When unsetting STS credentials in the integ-tests framework we are not restoring the creds that were previously available in the env. This is preventing us from running tests where creds are set in the `AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN` variables. This patch addresses this problem.


tests/integration-tests/utils.py
```diff
@@ -330,6 +330,14 @@ def set_credentials(region, credential_arg):
...
 
             # Set credential for all cli command e.g. pcluster create
+            credentials_to_backup = {
+                "AWS_ACCESS_KEY_ID": os.environ.get("AWS_ACCESS_KEY_ID"),
+                "AWS_SECRET_ACCESS_KEY": os.environ.get("AWS_SECRET_ACCESS_KEY"),
+                "AWS_SESSION_TOKEN": os.environ.get("AWS_SESSION_TOKEN"),
+            }
+            for key, value in credentials_to_backup.items():
+                if value:
+                    os.environ[f"BACKUP_{key}"] = value
```
tilne, line 340: will using envs for this be a problem if credentials are set by assuming a role multiple times in a given process?

example:
* test runs in opt-in region, so credentials are set by assuming that role
* user policy role is assumed; new creds are assumed and old creds are written to these envs
* integ test needs to assume another role for whatever reason; user policy creds are written to these envs and original opt-in region creds are clobbered

i'm not sure it the last bullet point ever happens, so maybe it's not worth handling.

PR #2778, opened 2021-05-24
3.0: Add integration test for Infrastructure API CloudFormation template

ermanno:
### Notes
This patch adds an infrastructure tests validating that when creating a stack with the Infrastructure API CloudFormation template using the default parameters the APIGateway API backed by Lambda is correctly set up. The test requires inputting two parameters, the URI of the published Docker image for the ParallelCluster Lambda and the S3 URI of the API definition.

### Tests
Test run in personal account

### Notes for the reviewer
1. This test exercises creates the stack with default parameters. There are two notable alternative scenarios: one where we specify a custom domain name, another where we create a PRIVATE API (instead of a REGIONAL one) associating it to a VPC endpoint. Both these scenarios require the customer to run the test in an account where they have created those resources, eventually we could add appropriate parameters (custom domain name and HostedZone, VPC endpoint id)
2. I wrote the test with multiple assertions to be coherent with the rest of the integration tests. To my knowledge `assertpy`'s default behavior does not support multiple failures like other tools (such as [pytest-check](https://github.com/okken/pytest-check)): an assertion failure will halt test execution immediately by raising an error. I have wrapped the assertions inside a [soft_assertions()](https://github.com/assertpy/assertpy#soft-assertions) context manager to to collect assertion failures together, to be raise all at once at the end, without halting the test

### DIFF
--- a/tests/integration-tests/test_runner.py
+++ b/tests/integration-tests/test_runner.py
@@ -65,6 +65,8 @@ TEST_DEFAULTS = {
     "post_install": None,
     "vpc_stack": None,
     "cluster": None,
+    "api_definition_s3_uri": None,
+    "public_ecr_image_uri": None,
     "no_delete": False,
     "benchmarks": False,
     "benchmarks_target_capacity": 200,
@@ -282,6 +284,16 @@ def _init_argparser():
         "--cluster", help="Use an existing cluster instead of creating one.", default=TEST_DEFAULTS.get("cluster")
     )
     debug_group.add_argument(
+        "--api-definition-s3-uri",
+        help="URI of the Docker image for the Lambda of the ParallelCluster API",
+        default=TEST_DEFAULTS.get("api_definition_s3_uri"),
+    )
+    debug_group.add_argument(
+        "--public-ecr-image-uri",
+        help="S3 URI of the ParallelCluster API spec",
+        default=TEST_DEFAULTS.get("public_ecr_image_uri"),
+    )
+    debug_group.add_argument(
         "--no-delete",
         action="store_true",
         help="Don't delete stacks after tests are complete.",
@@ -481,6 +493,12 @@ def _set_custom_stack_args(args, pytest_args):
 
     if args.cluster:
         pytest_args.extend(["--cluster", args.cluster])
+
+    if args.api_definition_s3_uri:
+        pytest_args.extend(["--api-definition-s3-uri", args.api_definition_s3_uri])
+
+    if args.public_ecr_image_uri:
+        pytest_args.extend(["--public-ecr-image-uri", args.public_ecr_image_uri])
 
     if args.no_delete:
         pytest_args.append("--no-delete")
--- a/tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
+++ b/tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
@@ -1,0 +1,152 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import os
+import re
+
+import boto3
+import botocore
+import pytest
+import requests
+from assertpy import assert_that, soft_assertions
+from cfn_stacks_factory import CfnStack, CfnStacksFactory
+from utils import generate_stack_name
+
+
+@pytest.fixture()
+@pytest.mark.usefixtures("setup_sts_credentials")
+def api_infrastructure_stack_factory(request):
+    """Define a fixture to manage the creation and destruction of the API Infrastructure CloudFormation stack."""
+    factory = CfnStacksFactory(request.config.getoption("credential"))
+
+    def _create_api_infrastructure(region, parameters):
+        template_path = os.path.join("..", "..", "api", "infrastructure", "parallelcluster-api.yaml")
+        file_content = extract_template(template_path)
+        stack = CfnStack(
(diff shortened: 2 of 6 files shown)


tests/integration-tests/test_runner.py
```diff
@@ -65,6 +65,8 @@
     "post_install": None,
     "vpc_stack": None,
     "cluster": None,
+    "api_definition_s3_uri": "",
```
demartinofra, line 68: nit: shall we default to None like done for the other params?

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
```diff
@@ -0,0 +1,131 @@
...
+import boto3
+import pytest
+from assertpy import assert_that, soft_assertions
+from cfn_stacks_factory import CfnStack, CfnStacksFactory
+from utils import generate_stack_name
+
+
+@pytest.fixture()
+@pytest.mark.usefixtures("setup_sts_credentials")
+def api_infrastructure_stack_factory(request):
```
demartinofra, line 24: shall we move this fixture to conftest and have it in common with the networking test? In the end what we need is a cfn_stacks_factory that has a class scope rather than session (that already exists in conftest). we could rename the other one to session_cfn_stacks_factory or find a meaningful name for the new one.

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
```diff
@@ -0,0 +1,131 @@
...
+        with open(template_path) as cfn_file:
+            file_content = cfn_file.read()
+        return file_content
+
+    yield _create_api_infrastructure
+    factory.delete_all_stacks()
+
+
+@pytest.mark.skip_regions(["cn-north-1", "cn-northwest-1"])  # TODO No Lambda container support in China regions
+@pytest.mark.usefixtures("os", "instance", "scheduler")
```
demartinofra, line 51: I think you can get rid of this one if you don't need to parametrize on these dimensions

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
```diff
@@ -0,0 +1,131 @@
...
+            lambda_arn=parallelcluster_lambda_arn,
+            lambda_image_uri=parallelcluster_api_copied_image_uri,
+        )
+
+        _assert_parallelcluster_api(
+            client=apigateway_client, api_id=parallelcluster_api_id, api_url=parallelcluster_api_url
+        )
+
+
+def _assert_parallelcluster_lambda(client, lambda_name, lambda_arn, lambda_image_uri):
```
demartinofra, line 91: shall we also assert the timeout, memory size and tracing config?

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
```diff
@@ -0,0 +1,131 @@
...
+    :param lambda_image_uri: the URI of the local copy of the ParallelCluster Lambda Docker image
+    """
+    lambda_resource = client.get_function(FunctionName=lambda_name)
+
+    assert_that(lambda_resource["Configuration"]["FunctionArn"]).is_equal_to(lambda_arn)
+    assert_that(lambda_resource["Tags"]).contains("parallelcluster:version")
+    assert_that(lambda_resource["Code"]["ImageUri"]).is_equal_to(lambda_image_uri)
+
+
+def _assert_parallelcluster_api(client, api_id, api_url):
```
demartinofra, line 106: can we also assert authorizer config and tracing?
I'm also wondering how hard would it be to perform a simple end-to-end request to the simplest api operation such as ListClusters and just assert the code is 200.

demartinofra (commented):
Very well done. just dropped some minor comments

tests/integration-tests/test_runner.py
ermanno: Agreed

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
ermanno: Are you referring to `@pytest.mark.usefixtures("os", "instance", "scheduler")`? I don't know if I have interpreted wrongly [what's written in the README](https://github.com/ermanno/aws-parallelcluster/blob/wip/pcluster3/tests/integration-tests/README.md#define-parametrized-test-cases), but I was under the impression that the fixture was a way to "comply" with the test framework in case we didn't really need to access the listed dimensions

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
ermanno: Agreed, I was thinking of moving it to utils, but conftest is better since it's where I have put the other two fixtures

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
ermanno: Confirmed that the test can be run without the fixture, will proceed to remove it

tests/integration-tests/tests/api_infrastructure/test_api_infrastructure.py
ermanno: 1. I have tried the library suggested in [this GitHub issue](https://github.com/boto/boto3/issues/1246) unsuccessfully. I have implemented the call with Botocore/Requests
1. Regarding the authorizer config, I am unsure about what to assert. Specifically, I tried [get_deployment](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#APIGateway.Client.get_deployment), and [get_resource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#APIGateway.Client.get_resource)/[get_resources](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#APIGateway.Client.get_resources) methods and I don't see an `authorizationType` in the response. Looking at the processed template, I see the following in the [body of the REST API resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-body), which is the OpenAPI specification defining the RESTful API:

```
"security": [
    {
        "AWS_IAM": []
    }
]
```

demartinofra (approved):

PR #2793, opened 2021-05-27
Move all boto3 call from utils.py to boto layer

hanwen-pcluste:
One call (boto3.session.Session().region_name) is kept in the utils.py. This call does not require internet connection, but is about retrieve information from environment variables and aws config file.

There are boto3 calls in Lambda code, integration tests, and the root util folder. None of them needs to be changed.


cli/tests/pcluster/aws/test_cfn.py
```diff
@@ -20,6 +23,12 @@ def boto3_stubber_path():
     return "pcluster.aws.common.boto3"
 
 
+@pytest.fixture(autouse=True)
+def reset_aws_api():
+    """Reset AWSApi singleton to remove dependencies between tests."""
+    AWSApi._instance = None
```
enrico-usai, line 29: Let's move it to conftest.py, to be used by all the tests.

enrico-usai (approved):
Great job! I just added a minor comment.

cli/tests/pcluster/aws/test_cfn.py
hanwen-pcluste: I was blocked by circular imports. As discussed, we decided to keep it here instead of spending more time investigating

PR #2796, opened 2021-05-28
3.0: Execute integration tests using customer role

gmarciani:
### Changes
1. Add fixture to configure aws credentials to be used when executing tests, according to the region and the specified scheduler.
2. Add assumed role arn into a log line within utility function that sets credentials. Since we are going to switch credentials more often, it is useful to specify not only the region where we are setting credentials, but also the assumed role.
3. Add S3 URL within the S3 validation error log line in order to better specify what went wrong
4. Add to gitignore: pytest.log

### Testing
1. Integration tests (ongoing)

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -43,6 +43,7 @@ from framework.tests_configuration.config_renderer import read_config_file
 from framework.tests_configuration.config_utils import get_all_regions
 from jinja2 import Environment, FileSystemLoader
 from network_template_builder import Gateways, NetworkTemplateBuilder, SubnetConfig, VPCConfig
+from permissions_template_builder import PermissionsTemplateBuilder
 from retrying import retry
 from utils import (
     InstanceTypesData,
@@ -52,6 +53,7 @@ from utils import (
     dict_has_nested_key,
     generate_stack_name,
     get_architecture_supported_by_instance_type,
+    get_identity,
     get_instance_info,
     get_network_interfaces_count,
     get_vpc_snakecase_value,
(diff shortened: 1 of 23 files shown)


tilne (dismissed):

tests/integration-tests/conftest.py
```diff
@@ -1023,3 +1025,75 @@ def _build_image(image_id, region, cluster_config):
...
+        logging.info(f"Identity restored in region {region}: {get_identity(region)}")
+
+    logging.info(f"Configuring AWS credentials in region {region} for scheduler: {scheduler}")
+
+    role_arn = get_role_arn(scheduler)
+    assume_role(region, role_arn)
+
+    yield
+
+    reset_role()
```
demartinofra, line 1061: I'm afraid this would break opt-in region tests cause it would leave no cred to be used by the tear down of other fixtures such as clusters_factory. 

This PR could address the problem but I did not have time to properly test it https://github.com/aws/aws-parallelcluster/pull/2777

tests/integration-tests/resources/customer_roles/awsbatch/user_awsbatch_policy_data.json
```diff
@@ -0,0 +1,80 @@
...
+        "s3:DeleteBucket",
+        "s3:DeleteObject",
+        "s3:DeleteObjectVersion"
+      ],
+      "Effect": "Allow",
+      "Resource": [
+        { "Fn::Sub": "arn:${AWS::Partition}:s3:::parallelcluster-*" }
+      ],
+      "Sid": "S3Delete"
+    }
```
demartinofra, line 77: you already have s3* above

gmarciani: Thanks. Removing in the next rev

tests/integration-tests/conftest.py
gmarciani: I think your approach should work at least in the scenario where we have only one level of set creds. 
If we want to support multiple levels, we should convert your backup dictionary to a backup stack of creds.

gmarciani: Working on that in the next rev

PR #2806, opened 2021-06-03
3.0: Implement DeleteImage API

ermanno:
### Notes
This patch implements the DeleteImage API. We create an `ImageBuilder` class with the given image id and retrieve the associated image or stack, then we delete the image and, in case we only had the stack, we retrieve it with another `ImageBuilder` object (the stack is initialized only once, and if we tried to retrieve it after calling the delete operation we might be unable to describe it as it might have already transitioned to the `DELETE_COMPLETE` state).

One thing to note here is that the `ImageBuilder.delete()` call might fail raising an `ImageBuilderActionError`, which is surfaced to the caller of the DeleteImage API as a 500 with the message: "Unexpected fatal exception. Please look at the application logs for details on the encountered failure."

### Tests
1. Unit tests
2. Delete of an available image:
```
curl --location --request DELETE 'http://<ENDPOINT>/v3/images/custom/test7?region=us-east-1'
{
  "image": {
    "imageBuildStatus": "DELETE_IN_PROGRESS",
    "imageId": "test7",
    "region": "us-east-1",
    "version": "3.0.0"
  }
}
```
3. Delete of a stack:
```
curl --location --request DELETE 'http://<ENDPOINT>/v3/images/custom/test1?region=us-east-1'
{
  "image": {
    "cloudformationStackArn": <CFN ARN>,
    "cloudformationStackStatus": "DELETE_IN_PROGRESS",
    "imageBuildStatus": "DELETE_IN_PROGRESS",
    "imageId": "test1",
    "region": "us-east-1",
    "version": "3.0.0"
  }
}
```

### Notes for the reviewer
I would like to solicit feedback on the following points:
1. In the [pcluster_api.py implementation](https://github.com/aws/aws-parallelcluster/blob/wip/pcluster3/cli/src/pcluster/api/pcluster_api.py#L327-L350) we get the image/stack **after** calling the `delete` operation, I am not sure this is correct. Among other things, the stack deletion is asynchronous, and so the stack might or might not have already reached the `DELETE_COMPLETE` status, in which case it won't be visible in the describe, thus leading to a `NonExistingStackError`.
2. As mentioned above, the `ImageBuilder.delete()` wraps, among other exceptions, `AWSClientError` in `ImageBuilderActionError` exceptions. I think that this is not what we want to happen, for example, in cases where we have a `REQUEST_LIMIT_EXCEEDED` in one of the EC2/CloudFormation calls, which I believe we would like to surface as a `LimitExceededException`. I see that in https://github.com/aws/aws-parallelcluster/pull/2776 for the `CreateCluster` we have wrapped these exceptions in a `InternalServiceException`, with a message that at least surfaces the message of the underlying exception, however I still have some questions regrading the error code (`InternalServiceException` are 500s). **UPDATE: new approach explained [here](https://github.com/aws/aws-parallelcluster/pull/2806#issuecomment-854806810)**

### DIFF
--- a/cli/src/pcluster/api/controllers/image_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/image_operations_controller.py
@@ -12,7 +12,6 @@ import os as os_lib
 from datetime import datetime
 
 from pcluster.api.controllers.common import configure_aws_region
-from pcluster.api.converters import cloud_formation_status_to_image_status
 from pcluster.api.errors import (
     BadRequestException,
     InternalServiceException,
@@ -136,17 +135,11 @@ def delete_image(image_id, region=None, client_token=None, force=None):
 
     imagebuilder.delete(force=force)
 
-    if stack:
-        cloud_formation_status_after_deletion = _get_cloud_formation_status_after_deletion(stack.name)
-        image_build_status_after_deletion = cloud_formation_status_to_image_status(
-            cloud_formation_status_after_deletion
-        )
-
     return DeleteImageResponseContent(
         image=ImageInfoSummary(
             image_id=image_id,
-            image_build_status=image_build_status_after_deletion if stack else ImageBuildStatus.DELETE_COMPLETE,
-            cloudformation_stack_status=cloud_formation_status_after_deletion if stack else None,
+            image_build_status=ImageBuildStatus.DELETE_IN_PROGRESS,
+            cloudformation_stack_status=CloudFormationStatus.DELETE_IN_PROGRESS if stack else None,
             cloudformation_stack_arn=stack.id if stack else None,
             region=os_lib.environ.get("AWS_DEFAULT_REGION"),
             version=stack.version if stack else image.version,
@@ -167,13 +160,6 @@ def _get_underlying_image_or_stack(imagebuilder):
                 "Unable to find an image or stack for ParallelCluster image id: {}".format(imagebuilder.image_id)
             )
     return image, stack
-
-
-def _get_cloud_formation_status_after_deletion(stack_name):
-    try:
-        return ImageBuilder(image_id=stack_name).stack.status
-    except NonExistingStackError:
-        return CloudFormationStatus.DELETE_COMPLETE
 
 
 @configure_aws_region()
--- a/cli/src/pcluster/api/controllers/image_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/image_operations_controller.py
@@ -34,13 +34,15 @@ from pcluster.api.models.delete_image_response_content import DeleteImageRespons
 from pcluster.api.models.image_build_status import ImageBuildStatus
 from pcluster.models.imagebuilder import (
     BadRequestImageBuilderActionError,
+    BadRequestImageError,
     ImageBuilder,
     LimitExceededImageBuilderActionError,
     LimitExceededImageError,
     NonExistingImageError,
 )
 
-from ...models.imagebuilder_resources import LimitExceededStackError, NonExistingStackError
+from ...aws.common import BadRequestError, LimitExceededError
+from ...models.imagebuilder_resources import BadRequestStackError, LimitExceededStackError, NonExistingStackError
 
 
 def convert_imagebuilder_errors():
@@ -51,9 +53,19 @@ def convert_imagebuilder_errors():
                 return func(*args, **kwargs)
             except ParallelClusterApiException as e:
                 error = e
-            except (LimitExceededImageError, LimitExceededStackError, LimitExceededImageBuilderActionError) as e:
+            except (
+                LimitExceededError,
+                LimitExceededImageError,
+                LimitExceededStackError,
+                LimitExceededImageBuilderActionError,
+            ) as e:
                 error = LimitExceededException(str(e))
-            except BadRequestImageBuilderActionError as e:
+            except (
+                BadRequestError,
+                BadRequestImageError,
+                BadRequestStackError,
+                BadRequestImageBuilderActionError,
+            ) as e:
                 error = BadRequestException(str(e))
             except Exception as e:
                 error = InternalServiceException(str(e))
--- a/cli/src/pcluster/api/controllers/image_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/image_operations_controller.py
@@ -7,10 +7,12 @@
 # limitations under the License.
 
 # pylint: disable=W0613
-
+import os as os_lib
 from datetime import datetime
 
 from pcluster.api.controllers.common import configure_aws_region
+from pcluster.api.converters import cloud_formation_status_to_image_status
+from pcluster.api.errors import BadRequestException, NotFoundException
 from pcluster.api.models import (
     BuildImageRequestContent,
     BuildImageResponseContent,
@@ -24,6 +26,8 @@ from pcluster.api.models import (
 )
 from pcluster.api.models.delete_image_response_content import DeleteImageResponseContent
 from pcluster.api.models.image_build_status import ImageBuildStatus
+from pcluster.models.imagebuilder import ImageBuilder, NonExistingImageError
+from pcluster.models.imagebuilder_resources import NonExistingStackError
 
 
 @configure_aws_region(is_query_string_arg=False)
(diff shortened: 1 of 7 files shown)


ermanno:
The new revision has a new way to handle errors, namely it uses subclasses of ImageBuilder errors to propagate particular error classes like LimitExceeded, etc. The three approaches we thought of were:
1. Directly propagate the AWSClientError
2. Add an error code field to the ImageBulider exceptions
3. Created subclasses of the ImageBuilder exceptions

We went with the third option (cleaner because it uses the type system instead of an optional string, and the error message is a concept associated more with AWS errors, which are not the only source of this "distinction" between error types... for example, we could have a BadRequest exception because the customer tries to deletee an image that is shared or is used by some instances)

lukeseawalker (commented):
1. I should check with @yuleiwan, but it could be possible that the get of image/stack is returning a cached version. If it's not that case, I think that your point makes sense and those resources should be retrieved before calling the delete.

EDIT
I confirm that we are getting the image/stack before calling the actual deregistration/delete of the image/stack, e.g. 
in `AWSApi.instance().ec2.deregister_image(self.image.id)` , `self.image retrieves the image before calling `deregister_image`, so when building `ImageBuilderImageInfo` a cached version is used. Same for stack. Nevertheless we can improve the code.

2. I think that the approach of creating subclasses of the ImageBuilder exceptions is reasonable and extensible in future

ermanno:
> 1. I should check with @yuleiwan, but it could be possible that the get of image/stack is returning a cached version. If it's not that case, I think that your point makes sense and those resources should be retrieved before calling the delete.
> 
> EDIT
> I confirm that we are getting the image/stack before calling the actual deregistration/delete of the image/stack, e.g.
> in `AWSApi.instance().ec2.deregister_image(self.image.id)` , `self.image` retrieves the image before calling `deregister_image`, so when building `ImageBuilderImageInfo` a cached version is used. Same for stack. Nevertheless we can improve the code.

I think that the patch handles this. In the stack case I have a `_get_cloud_formation_status_after_deletion` function that creates an `ImageBuilder` object and tries to get the stack **after** the deletion, to get the up to date CloudFormation status. For the case where only the `image` was present, I simply set the state to `DELETE_COMPLETE`. I **also** get the image/stack before the deletion only to get the general information (`image.version`, `stack.version`, etc.) **before** deleting the image/stack.

Above I had linked to the [pcluster_api.py](https://github.com/aws/aws-parallelcluster/blob/wip/pcluster3/cli/src/pcluster/api/pcluster_api.py#L327-L350) implementation, that initializes an `ImageBuilder` object **after** the deletion and tries to get the image and then the stack, eventually throwing an error if it does not find any... as I have written above, I think that this is not what we want to do, as in the case the deletion is fast, we would throw an error even in case of a successful deletion.

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -78,20 +83,62 @@ def __init__(self, message: str, validation_failures: list = None):
...
 
+class LimitExceededImageBuilderActionError(ImageBuilderActionError):
+    """Represent an error during the execution of an action on the imagebuilder due to limit exceeded."""
+
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
+class BadRequestImageBuilderActionError(ImageBuilderActionError):
+    """Represent an error during the execution of an action due to limit exceeded."""
```
yuleiwan, line 94: incorrect comment?

cli/src/pcluster/models/imagebuilder.py
```diff
@@ -78,20 +83,62 @@ def __init__(self, message: str, validation_failures: list = None):
...
 
 
+class LimitExceededImageBuilderActionError(ImageBuilderActionError):
+    """Represent an error during the execution of an action on the imagebuilder due to limit exceeded."""
+
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
+class BadRequestImageBuilderActionError(ImageBuilderActionError):
```
yuleiwan, line 93: Is BadRequestImageBuilderActionError mapping to BadRequestException and is ImageBuilderActionErrormapping to InternalServiceException? What are different use cases for BadRequestException and InternalServiceException?

yuleiwan (commented):
well done!

lukeseawalker:
> we set the old image or stack state in the response (i.e. the one they had before the deletion)
correct, while the status of the PCLuster image is set to `DELETE_IN_PROGRESS`. Let's sync on this if doesn't sound reasonable

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -87,18 +127,55 @@ def delete_image(image_id, region=None, client_token=None, force=None):
...
         )
     )
 
 
+def _get_underlying_image_or_stack(imagebuilder):
+    image = None
+    stack = None
+    try:
+        image = imagebuilder.image
+    except NonExistingImageError:
```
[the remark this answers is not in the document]
ermanno: No, I think it's correct to handle them as `InternalServiceExceptions` If they are not due to a problem in the request. I will add support for BadRequest errors (for example, with @demartinofra  we were saying that an `ParameterValidationError` when doing an AWS call should be a `BadRequestException`)

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -87,18 +127,55 @@ def delete_image(image_id, region=None, client_token=None, force=None):
 
     :rtype: DeleteImageResponseContent
     """
+    if client_token:
+        raise BadRequestException("clientToken is currently not supported for this operation")
+
+    force = force or False
```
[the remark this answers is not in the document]
ermanno: I see that in the rest of the code we followed this same pattern, there is a CheckStyle rule for it.

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -87,18 +127,55 @@ def delete_image(image_id, region=None, client_token=None, force=None):
...
 
+def _get_underlying_image_or_stack(imagebuilder):
+    image = None
+    stack = None
+    try:
+        image = imagebuilder.image
+    except NonExistingImageError:
+        try:
+            stack = imagebuilder.stack
+        except NonExistingStackError:
```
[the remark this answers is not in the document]
ermanno: See above.

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -87,18 +127,55 @@ def delete_image(image_id, region=None, client_token=None, force=None):
...
+            raise NotFoundException(
+                "Unable to find an image or stack for ParallelCluster image id: {}".format(imagebuilder.image_id)
+            )
+    return image, stack
+
+
+def _get_cloud_formation_status_after_deletion(stack_name):
+    try:
+        return ImageBuilder(image_id=stack_name).stack.status
+    except NonExistingStackError:
```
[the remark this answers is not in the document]
ermanno: Same reasoning as above.

cli/src/pcluster/models/imagebuilder.py
ermanno: Fixed phrasing in new review.

cli/src/pcluster/models/imagebuilder.py
ermanno: The `BadRequestException` represents an exception due to some problem in the request (for example, the customer attempts to delete an image that is shared without using the `force` option), it's a 4xx. whereas an `InternalServiceException` is a generic Server side exception (a 500). In the case of `AWSClientError` returned by AWS calls, I will separate these cases, as with @demartinofra we determined that, for example, `ParameterValidationError`s  should be surfaced as `BadRequestException`s

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -24,6 +32,48 @@
...
+from pcluster.models.imagebuilder import (
+    BadRequestImageBuilderActionError,
+    BadRequestImageError,
+    ImageBuilder,
+    LimitExceededImageBuilderActionError,
+    LimitExceededImageError,
+    NonExistingImageError,
+)
+
+from ...aws.common import BadRequestError, LimitExceededError
```
lukeseawalker, line 44: why not `pcluster.aws.common`?

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -24,6 +32,48 @@
...
+from ...models.imagebuilder_resources import BadRequestStackError, LimitExceededStackError, NonExistingStackError
```
lukeseawalker, line 45: same here

lukeseawalker (approved):

PR #2807, opened 2021-06-03
3.0: Keep logs on default and fix integration tests

hanwen-pcluste:
See commit descriptions for detail.

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -12,6 +12,7 @@
 # This module contains all the classes representing the Resources objects.
 # These objects are obtained from the configuration file through a conversion based on the Schema classes.
 #
+import json
 import logging
 import time
 from copy import deepcopy
@@ -310,6 +311,7 @@ class Cluster:
 
         except ValidationError as e:
             # syntactic failure
+            ClusterSchema.process_validation_message(e)
             validation_failures = [ValidationResult(str(e), FailureLevel.ERROR)]
             raise ConfigValidationError("Configuration is invalid", validation_failures=validation_failures)
 
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -297,6 +297,11 @@ class Cluster:
             LOGGER.info("Validating cluster configuration...")
             # syntactic validation
             cluster_config_dict = config_dict or self.source_config
+            if "DevSettings" in cluster_config_dict:
+                instance_types_data = cluster_config_dict["DevSettings"].get("InstanceTypesData")
+                if instance_types_data:
+                    # Set additional instance types data in AWSApi. Schema load will use the information.
+                    AWSApi.instance().ec2.additional_instance_types_data = json.loads(instance_types_data)
             config = ClusterSchema().load(cluster_config_dict)
 
             # semantic validation
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -18,6 +18,7 @@ import boto3
 import yaml
 from retrying import retry
 from utils import (
+    dict_add_nested_key,
     retrieve_cfn_outputs,
     retrieve_cfn_parameters,
     retrieve_cfn_resources,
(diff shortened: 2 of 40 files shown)


cli/src/pcluster/models/cluster.py
```diff
@@ -310,6 +316,14 @@ def _validate_and_parse_config(self, suppress_validators, validation_failure_lev
 
         except ValidationError as e:
             # syntactic failure
+            # Process error message. Because there are some @pre_load(s) changing the structure of the schema,
+            # We need to modify the error message to be aligned with the structure before @pre_load(s)
+            error_massage = e.args[0]
```
enrico-usai, line 321: massage? :D

cli/src/pcluster/models/cluster.py
```diff
@@ -310,6 +316,14 @@ def _validate_and_parse_config(self, suppress_validators, validation_failure_lev
...
+            if "Scheduling" in error_massage:
```
enrico-usai, line 322: 1. let's add another method called `process_error_message`
2. We have to do the same conversion for the pre_load of the storage type.

Anyway instead of doing this explicit conversion here in the model we should keep this conversion logic in the schema layer. 
We should double check what we have in the validation error object. If we have the the entire data information (i.e. schema metadata) I'm thinking to something where the conversion logic is saved as metadata in the schema layer and here we just need to call a generic converter if defined.

cli/src/pcluster/models/cluster.py
```diff
@@ -310,6 +316,14 @@ def _validate_and_parse_config(self, suppress_validators, validation_failure_lev
...
+                scheduling = error_massage["Scheduling"]
+                scheduler_settings = scheduling.get("Slurm") or scheduling.get("Awsbatch")
```
enrico-usai, line 324: temporarily let's define an enum for the available schedulers, in this way if we add another scheduler this code shouldn'ìt change.

tests/integration-tests/clusters_factory.py
```diff
@@ -63,20 +64,39 @@ def update(self, force=True):
             logging.error(error)
             raise Exception(error)
         logging.info("Cluster {0} updated successfully".format(self.name))
+        # Only update config file attribute if update is successful
+        self.config_file = config_file
+        with open(self.config_file) as conf_file:
+            self.config = yaml.load(conf_file, Loader=yaml.SafeLoader)
```
enrico-usai, line 70: the alternative to `yaml.load(... Loader=yaml.SafeLoader)` should be simply `yaml.safe_load`

tests/integration-tests/tests/update/test_update.py
```diff
@@ -444,76 +439,3 @@ def get_batch_spot_bid_percentage(stack_name, region):
...
-        cluster,
-        scheduler,
-        region,
-        "c5.2xlarge",
-        disable_hyperthreading=disable_hyperthreading,
-        cfn_scheduler_slots=cfn_scheduler_slots,
-    )
-
-
-def _check_compute_node_slots(
```
enrico-usai, line 491: Why are we removing these checks?

tests/integration-tests/clusters_factory.py
```diff
@@ -259,29 +279,32 @@ def create_cluster(self, cluster, extra_args=None, raise_on_error=True):
...
         time.sleep(30)
 
-    @retry(stop_max_attempt_number=5, wait_fixed=5000, retry_on_exception=retry_if_subprocess_error)
-    def destroy_cluster(self, name, keep_logs=False):
+    def destroy_cluster(self, name, delete_logs):
         """Destroy a created cluster."""
         logging.info("Destroying cluster {0}".format(name))
         if name in self.__created_clusters:
-            keep_logs = keep_logs or (self._keep_logs_on_failure and not self.__created_clusters[name].create_complete)
+            delete_logs = delete_logs and (self._keep_logs_on_failure and self.__created_clusters[name].create_complete)
```
enrico-usai, line 286: Let's talk about this logic

tests/integration-tests/tests/update/test_update.py
hanwen-pcluste: This checks were used for sge and torque. Now this code are not used

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Haha, sorry.

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Done by moving it to ClusterSchema.
I double checked the validation error object. There is no metadata inside the object.

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Done by reusing constant SUPPORTED_SCHEDULERS instead of creating a Enum

tests/integration-tests/clusters_factory.py
hanwen-pcluste: There are many appearances, I uniformed them all

enrico-usai (approved):

tests/integration-tests/conftest.py
```diff
@@ -837,7 +842,7 @@ def _create_vpc_stack(request, template, region, cfn_stacks_factory):
     return stack
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="class")
```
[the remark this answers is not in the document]
hanwen-pcluste: To make sure the teardown order is correct. For example, the code here should ensure s3_bucket_factory is teared down after cluster_factory. This requires the scope of s3_bucket_factory to be greater or equal to the cluster_factory. If the two scopes are equal, s3_bucket_factory should appear in front of cluster_factory.
This is to avoid failures during the final update to delete the log

PR #2814, opened 2021-06-04
3.0: Implement BuildImage API

ermanno:
### Notes
This patch implements the BuildImage API. It uses the operations defined in the `ImageBuilder` model class, in particular we have extracted the validation logic from the `create` method and in the case of a dry-run request we validate that the provided image id is valid and does not match any existing image or stack, then we we validate the configuration.

Additional notes for the reviewer:
1. Following what we have done in the `CreateCluster` patch (https://github.com/aws/aws-parallelcluster/pull/2776) the `BuildImageBadRequestException` taken an array of validation errors (which can be empty). Since the parameter afaik is optional, I would only add it to the response if it is non empty (at the moment it is added anyway to conform with what we did for the `CreateCluster`)
2. I have introduced some changes to the specification (`ConflictException`, `DryRunOperationException`) and changed the `ValidationError` to also include the validator type, as well as copying some common methods (parse of the configuration, conversion of the validation errors) in the common code/converters sections of the controller module
3. I notice that the `BuildImageBadRequestException` does not prepend `"Bad image:"` to the exception message as the regular `BadRequestException` does

### Tests
1. Unit tests
2. Dry run:
```
curl --location --request POST 'http://<ENDPOINT>/v3/images/custom?region=us-east-1&dryrun=true' \
> --header 'Content-Type: application/json' \
> --data-raw '{
>   "imageConfiguration": <BASE64 ENCODED CONFIG>,
>   "id": "testcreatestack",
>   "region": "us-east-1"
> }'
{"message": "Request would have succeeded, but DryRun flag is set."}
```
3. Create request:
```
curl --location --request POST 'http://<ENDPOINT>/v3/images/custom?region=us-east-1' \
> --header 'Content-Type: application/json' \
> --data-raw '{
>   "imageConfiguration": <BASE64 ENCODED CONFIG>,
>   "id": "testcreatestack",
>   "region": "us-east-1"
> }'
{
  "image": {
    "cloudformationStackArn": <CLOUDFORMATION STACK ARN>,
    "imageBuildStatus": "BUILD_IN_PROGRESS",
    "imageId": "testcreatestack",
    "region": "us-east-1",
    "version": "3.0.0"
  }
}
```


cli/src/pcluster/api/controllers/common.py
```diff
@@ -61,3 +64,23 @@ def check_cluster_version(cluster):
...
+
+def parse_config(base64_encoded_config: str) -> Dict:
+    try:
+        config = base64.b64decode(base64_encoded_config).decode("UTF-8")
+    except Exception as e:
+        LOGGER.error("Failed when decoding cluster configuration: %s", e)
+        raise BadRequestException("invalid configuration. Please make sure the string is base64 encoded.")
+
+    if not config:
+        raise BadRequestException("configuration is required and cannot be empty")
```
lukeseawalker, line 77: what about logging an error here as well?

ermanno: Agreed

lukeseawalker (approved):

PR #2815, opened 2021-06-07
3.0: Implement DescribeOfficialImages

ermanno:
### Notes
This patch implements the `DescribeOfficialImages` API. We search for images owned by "amazon" with the correct name prefix (`aws-parallelcluster-...`), optionally taking version, os and architecture parameters which are used to narrow down the search by constructing an appropriate prefix pattern. Note that in the `AmiInfo` structures in the response we need to parse the image name in order to retrieve the correct information about the operating system, as tags are not visible to accounts different from the one that owns the image (these are official images, not images that have been built by the customer), and the `PlatformDetails `section of the EC2 `describe_images` response is not specific enough (i.e. it generically identifies the images as a Linux image, for example, but won't distinguish between CentOS, Ubuntu, etc.)

### Tests
1. Unit tests
2. Test without parameters:
```
 curl --location --request GET 'http://<ENDPOINT>/v3/images/official?region=us-east-1'
{
  "items": [
    {
      "amiId": "ami-0012b0e32aeeb20bc",
      "architecture": "arm64",
      "name": "aws-parallelcluster-2.10.4-centos8-hvm-arm64-202105142101",
      "os": "centos8"
    },
    {
      "amiId": "ami-00144db1528d8942d",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.3.2a1-ubuntu-1404-lts-hvm-201904261542",
      "os": "linux"
    },
    {
      "amiId": "ami-006da8413e239334a",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.4.1-ubuntu-1404-lts-hvm-201907260654",
      "os": "linux"
    },
...
```
3. With architecture filter:
```
curl --location --request GET 'http://<ENDPOINT>/v3/images/official?region=us-east-1&architecture=x86_64' 
{
  "items": [
    {
      "amiId": "ami-006e29fa34278d9f1",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.9.1-centos6-hvm-x86_64-202009142226",
      "os": "linux"
    },
    {
      "amiId": "ami-009428e2bfe97a05c",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.8.0-amzn-hvm-x86_64-202007222100",
      "os": "linux"
    },
    {
      "amiId": "ami-0098ea5504ad7af97",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.8.0-centos6-hvm-x86_64-202007222100",
      "os": "linux"
    },
...
```
4. With architecture and os filter:
```
curl --location --request GET 'http://<ENDPOINT>/v3/images/official?region=us-east-1&architecture=x86_64&os=linux'
{
  "items": [
    {
      "amiId": "ami-006e29fa34278d9f1",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.9.1-centos6-hvm-x86_64-202009142226",
      "os": "linux"
    },
    {
      "amiId": "ami-009428e2bfe97a05c",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.8.0-amzn-hvm-x86_64-202007222100",
      "os": "linux"
    },
    {
      "amiId": "ami-0098ea5504ad7af97",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.8.0-centos6-hvm-x86_64-202007222100",
      "os": "linux"
    },
...
```
5. With version:
```
curl --location --request GET 'http://<ENDPOINT>/v3/images/official?region=us-east-1&version=2.0.0'
{
  "items": [
    {
      "amiId": "ami-017b013ef58a362fb",
      "architecture": "x86_64",
      "name": "aws-parallelcluster-2.0.0-centos7-hvm-201811120751",
      "os": "centos7"
    },
...
```

### Notes for the reviewer
I have decided to return '' as a suffix if the customer passes an os that is not present in our os-to-suffix map. This way we return an empty list, which is coherent with what happens in case we specify an architecture or version that does not exist (i.e. we build a name prefix that does not match against any official image name, thus return an empty list). See [related comment](https://github.com/aws/aws-parallelcluster/pull/2815#discussion_r647205194).

### DIFF
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -8,12 +8,19 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 from typing import List
 
 from pcluster import utils
 from pcluster.aws.aws_resources import ImageInfo, InstanceTypeInfo
 from pcluster.aws.common import AWSClientError, AWSExceptionHandler, Boto3Client, Cache, ImageNotFoundError
-from pcluster.constants import PCLUSTER_IMAGE_BUILD_STATUS_TAG, PCLUSTER_IMAGE_ID_TAG, SUPPORTED_ARCHITECTURES
+from pcluster.constants import (
+    IMAGE_NAME_PART_TO_OS_MAP,
+    OS_TO_IMAGE_NAME_PART_MAP,
+    PCLUSTER_IMAGE_BUILD_STATUS_TAG,
+    PCLUSTER_IMAGE_ID_TAG,
+    SUPPORTED_ARCHITECTURES,
+)
 
 
 class Ec2Client(Boto3Client):
@@ -221,6 +228,19 @@ class Ec2Client(Boto3Client):
             raise AWSClientError(function_name="describe_images", message="Cannot find official ParallelCluster AMI")
         return max(images, key=lambda image: image["CreationDate"]).get("ImageId")
 
+    def get_official_images(self, version=None, os=None, architecture=None):
+        """Get the list of official images, optionally filtered by os and architecture."""
+        try:
+            owners = ["amazon"]
+            version = version or "*"
+            os = OS_TO_IMAGE_NAME_PART_MAP.get(os, "") if os else "*"
+            architecture = architecture or "*"
+            name = f"aws-parallelcluster-{version}-{os}-{architecture}*"
+            filters = [{"Name": "name", "Values": [name]}]
+            return self.describe_images(ami_ids=[], owners=owners, filters=filters)
+        except ImageNotFoundError:
+            return []
+
     @AWSExceptionHandler.handle_client_exception
     @Cache.cached
     def get_eip_allocation_id(self, eip):
@@ -228,18 +248,19 @@ class Ec2Client(Boto3Client):
         return self._client.describe_addresses(PublicIps=[eip])["Addresses"][0]["AllocationId"]
 
     @staticmethod
+    def extract_os_from_official_image_name(name):
+        """Return the os from the os part in an official image name."""
+        name_pattern = "aws-parallelcluster-[^-]+-(?P<OsPart>.+-hvm)-.*"
+        matches = re.match(name_pattern, name)
+        os_part = matches.groupdict().get("OsPart") if matches else None
+        return IMAGE_NAME_PART_TO_OS_MAP.get(os_part, "linux")
+
+    @staticmethod
     def _get_official_image_name_prefix(os, architecture):
         """Return the prefix of the current official image, for the provided os-architecture combination."""
-        suffixes = {
-            "alinux2": "amzn2-hvm",
-            "centos7": "centos7-hvm",
-            "centos8": "centos8-hvm",
-            "ubuntu1804": "ubuntu-1804-lts-hvm",
-            "ubuntu2004": "ubuntu-2004-lts-hvm",
-        }
-        return "aws-parallelcluster-{version}-{suffix}-{arch}".format(
-            version=utils.get_installed_version(), suffix=suffixes[os], arch=architecture
-        )
+        version = utils.get_installed_version()
+        os_part = OS_TO_IMAGE_NAME_PART_MAP[os]
+        return f"aws-parallelcluster-{version}-{os_part}-{architecture}"
 
     @AWSExceptionHandler.handle_client_exception
     def terminate_instances(self, instance_ids):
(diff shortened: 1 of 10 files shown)


cli/src/pcluster/aws/ec2.py
```diff
@@ -221,24 +222,62 @@ def get_official_image_id(self, os, architecture, filters=None):
...
+    def _get_os_suffix(os):
+        """Return the os suffix to use in an image name prefix."""
         suffixes = {
             "alinux2": "amzn2-hvm",
             "centos7": "centos7-hvm",
             "centos8": "centos8-hvm",
             "ubuntu1804": "ubuntu-1804-lts-hvm",
             "ubuntu2004": "ubuntu-2004-lts-hvm",
         }
+        return suffixes.get(os, "*")
```
ermanno, line 254: Not sure this is what we want. We might consider throwing an exception here, which in the `get_official_images` we might convert into a `BadRequestException` (a subclass of `AWSClientError`?)

ermanno: I have decided to return `''` as a suffix if it is not present in the map. This way we return an empty list, which is coherent with what happens in case we specify an architecture or version that does not exist (i.e. we build a name prefix that does not match against any official image name, thus return an empty list).

cli/src/pcluster/aws/ec2.py
```diff
@@ -221,25 +228,39 @@ def get_official_image_id(self, os, architecture, filters=None):
...
+            return []
+
     @AWSExceptionHandler.handle_client_exception
     @Cache.cached
     def get_eip_allocation_id(self, eip):
         """Retrieve the allocation id of an Elastic IP."""
         return self._client.describe_addresses(PublicIps=[eip])["Addresses"][0]["AllocationId"]
 
+    @staticmethod
+    def extract_os_from_official_image_name(name):
```
demartinofra, line 251: not sure I'd leave this in the Ec2Client class, what do you think?

cli/tests/pcluster/api/controllers/test_image_operations_controller.py
```diff
@@ -74,20 +75,142 @@ def test_describe_image(self, client):
...
             "Accept": "application/json",
         }
-        response = client.open("/v3/images/official", method="GET", headers=headers, query_string=query_string)
-        assert_that(response.status_code).is_equal_to(200)
+        return client.open("/v3/images/official", method="GET", headers=headers, query_string=query_string)
+
+    @pytest.mark.parametrize(
+        "version, os, arch, mocked_response, expected_response",
+        [
+            pytest.param(
```
demartinofra, line 114: nice I was not aware of pytest.param that lets you specify the id together with the params. much better!

cli/tests/pcluster/api/controllers/test_image_operations_controller.py
```diff
@@ -74,20 +75,142 @@ def test_describe_image(self, client):
...
+                "us-east-",
+                {"message": "Bad Request: invalid or unsupported region 'us-east-'"},
+                id="test with malformed region",
+            ),
+            pytest.param(None, {"message": "Bad Request: region needs to be set"}, id="test without region"),
+        ],
+    )
+    def test_malformed_request(self, client, region, expected_response):
```
demartinofra, line 191: what about modifying the spec to add at least a regex validation for architecture, os and version? I'm also ok if we want to have an enumeration so that we throw automatically bad request on invalid values

demartinofra (commented):
LGTM. only a couple of minor comments

cli/src/pcluster/aws/ec2.py
ermanno: I don't really like it either from a "semantic" point of view, as this is not an EC2 call, however my reasoning is that I wanted to keep it close to where the "pattern" of the official ami name was defined, so that if we decided to change the naming (i.e. from `aws-parallelcluster-...` to something else) we would have to make the change is a single place.

cli/tests/pcluster/api/controllers/test_image_operations_controller.py
ermanno: In `constants.py` we have two lists; `SUPPORTED_OSES` and `SUPPORTED_ARCHITECTURES`, which list the possible values of os/architecture. Yesterday I discussed with @lukeseawalker whether we would like to throw a bad request if the customer tried to pass an invalid value (I am leaving the version out of this discussion, as we were saying that we wanted to remove it as an input), and his observation was that in the case of an `ec2-describe-images`, if the filter you set does not match anything you get back an empty list (this is the current behavior of the `DescribeOfficialImages`, if you pass it an arch/os for which there is no match you get back an empty list).

ermanno: For context, the Boto3 `ec2.describe_images` accepts as values for the `"name"` filter any string, even with "strange" Unicode characters, despite them [not being allowed in the AMI name](https://stackoverflow.com/questions/55017935/what-is-maximum-length-of-an-aws-ami-name).

ermanno:
### Updates
As agreed with @demartinofra the latest revision removes the version from the available optional filters, and adds it to the response. The API will search only images for the current ParallelCluster version (using `utlis.get_installed_version()`).

demartinofra (approved):

PR #2816, opened 2021-06-07
Add validator to prevent using data_compression_type with fsx_backup_id

rexcsn:
* DataCompressionType is not supported when creating filesystem from backup. The designed behavior is to use the existing compression setting from the backup. Add validator for this case
* DataCompressionType option cannot be specified in CFN when creating from backup, added logic to use `AWS:NoValue` if using backup

### DIFF
--- a/cloudformation/fsx-substack.cfn.json
+++ b/cloudformation/fsx-substack.cfn.json
@@ -605,10 +605,18 @@
             ]
           },
           "DataCompressionType": {
-            "Fn::Select": [
-              "19",
+            "Fn::If": [
+              "UseBackupId",
               {
-                "Ref": "FSXOptions"
+                "Ref": "AWS::NoValue"
+              },
+              {
+                "Fn::Select": [
+                  "19",
+                  {
+                    "Ref": "FSXOptions"
+                  }
+                ]
               }
             ]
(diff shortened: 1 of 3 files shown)


chenwany (approved):
Good job!

cloudformation/fsx-substack.cfn.json
```diff
@@ -605,10 +605,18 @@
...
+              "UseBackupId",
               {
-                "Ref": "FSXOptions"
+                "Ref": "AWS::NoValue"
+              },
+              {
+                "Fn::Select": [
+                  "19",
+                  {
+                    "Ref": "FSXOptions"
```
chenwany, line 617: What about if we don't specify `DataCompressionType`? Are we going to use "NONE" as DataCompressionType?

rexcsn: As we discussed offline, the FSx default of DataCompressionType is NONE, so our logic is inline with that. We cannot use `AWS:NoValue` as the default because of an issue with update, and we did the change in this commit: https://github.com/aws/aws-parallelcluster/commit/6c04a0e41ba25c3bc5d0df99c97f5d97713ac4e1

PR #2818, opened 2021-06-08
3.0: Add configuration: HeadNode.Imds.Secured

gmarciani:
### Changes
1. Add configuration: HeadNode.Imds.Secured (default True)
2. Run resume/suspend scripts as cluster admin user. 
3. Minor: CHANGELOG format fixes as per pre-commit enforcement.

#### Notes
This PR depends on https://github.com/aws/aws-parallelcluster-cookbook/pull/989

### Tests
1. Cluster creation with HeadNode.Imds.Secured omitted (default True)
2. Cluster creation with HeadNode.Imds.Secured = True
3. Cluster creation with HeadNode.Imds.Secured = False
4. Kitchen tests (pending)
5. Integration tests (pending)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -732,6 +733,17 @@ def make_resource(self, data, **kwargs):
         return Iam(**data)
 
 
+class ImdsSchema(BaseSchema):
+    """Represent the schema of IMDS for HeadNode."""
+
+    secured = fields.Bool(metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
```
tilne, line 739: couldn't update for this be supported?

tilne (commented):
Is there a way we can add a check for the case where the head node's IMDS is configured to be secured and Batch is used as the scheduler, and if so print a warning (or even fail config validation)?

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: > Is there a way we can add a check for the case where the head node's IMDS is configured to be secured and Batch is used as the scheduler, and if so print a warning (or even fail config validation)?

Sure, we are going to fail the validation when HeadNode.Imds.Secured is True and Scheduling.Scheduler is awsbatch.
Will add this in the next iteration

gmarciani: > couldn't update for this be supported?

Think it's possible. Working on it.

tilne: > Just to be precise: the update of HeadNode.Imds.Secured implies the update of Cfn resource HeadNodeLaunchTemplate.Metadata.AWS::CloudFormation::Init.cfnHupConfig

would that require the replacement of the instance? it would think not, but i could be wrong.

gmarciani: No, it doesn't.

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -43,3 +50,47 @@ def test_awsbatch_cluster_builder(mocker):
...
+    )
+    expected_head_node_dna_json = load_json_dict(test_datadir / expected_head_node_dna_json_file_name)
+
+    diff = DeepDiff(
+        expected_head_node_dna_json,
+        generated_head_node_dna_json,
+        ignore_order=True,
+    )
+
+    assert not diff
```
tilne, line 88: * can [assertpy](https://github.com/assertpy/assertpy) be used instead of the builtin assertion? it's used throughout the rest of the tests, and helps keep a constant experience for the stack traces resulting in test failures.
* rather than adding a new dependency on `deepdiff`, can `assertpy`'s [dict comparisons](https://github.com/assertpy/assertpy#dicts) be used?

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -43,3 +50,47 @@ def test_awsbatch_cluster_builder(mocker):
...
+
+@pytest.mark.parametrize(
+    "config_file_name, expected_head_node_dna_json_file_name",
+    [
+        ("slurm-imds-secured-true.yaml", "slurm-imds-secured-true.head-node.dna.json"),
+        ("slurm-imds-secured-false.yaml", "slurm-imds-secured-false.head-node.dna.json"),
+        ("awsbatch-imds-secured-false.yaml", "awsbatch-imds-secured-false.head-node.dna.json"),
+    ],
+)
+@freeze_time("2021-01-01T01:01:01")
```
tilne, line 63: can you add a comment explaining why time is frozen at this particular value?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -190,6 +192,42 @@ def test_update_slurm(region, pcluster_config_reader, s3_bucket_factory, cluster
...
+@pytest.mark.dimensions("us-west-1", "c5.xlarge", "*", "slurm")
+@pytest.mark.usefixtures("instance", "scheduler")
+@pytest.mark.parametrize(
+    "imds_secured, imds_secured_updated, users_allow_list",
+    [
+        (True, False, {"root": True, "pcluster": True, "slurm": True}),
+        (False, True, {"root": True, "pcluster": True, "slurm": False}),
+    ],
+)
+def test_update_imds_secured(
```
tilne, line 204: * rather than adding a dedicated test for the update of this one parameter, would it be possible to instead 
* if adding a test is necessary, is it important to test the update in both directions (i.e., secured to not secured and vice versa)?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -190,6 +192,42 @@ def test_update_slurm(region, pcluster_config_reader, s3_bucket_factory, cluster
...
+):
+    """Test IMDS access with different configurations"""
+    custom_ami = retrieve_latest_ami(region, os, ami_type="pcluster", architecture=architecture)
+    cluster_config = pcluster_config_reader(custom_ami=custom_ami, imds_secured=imds_secured)
+    cluster = clusters_factory(cluster_config, raise_on_error=False)
+
+    assert_head_node_is_running(region, cluster)
+
+    cluster_config_updated = pcluster_config_reader(
+        config_file="pcluster.config.update.yaml", custom_ami=custom_ami, imds_secured=imds_secured_updated
```
tilne, line 223: the templates that are used to generate the initial and updated configuration files look to be the same. can the update config be dropped?

tests/integration-tests/tests/common/assertions.py
```diff
@@ -125,3 +126,30 @@ def assert_num_instances_constant(cluster_name, region, desired, timeout=5):
...
+    username = get_username_for_os(cluster.os)
+    remote_command_executor = RemoteCommandExecutor(cluster, username=username)
+
+    for user, allowed in users_allow_list.items():
+        logging.info(f"Asserting access to AWS caller identity is {'allowed' if allowed else 'denied'} for user {user}")
+        command = f"sudo -u {user} aws sts get-caller-identity"
+        result = remote_command_executor.run_remote_command(command, raise_on_error=False)
+        logging.info(f"user={user} and result.failed={result.failed}")
+        logging.info(f"user={user} and result.stdout={result.stdout}")
+        assert result.failed is not allowed
```
tilne, line 155: same thing here as in the unit tests: opt for the use of [assertpy](https://github.com/assertpy/assertpy) where possible.

tests/integration-tests/configs/pcluster3.yaml
```diff
@@ -98,4 +98,10 @@ test-suites:
       dimensions:
         - regions: ["eu-west-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["centos7"]
\ No newline at end of file
+          oss: ["centos7"]
+    test_update.py::test_update_imds_secured:
+      dimensions:
+        - regions: [ "us-east-1" ]
```
tilne, line 104: does it make sense to add a china and govcloud region as well?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1067,6 +1068,13 @@ def _add_head_node(self):
                     "instance_types_data_s3_key": f"{self.bucket.artifact_directory}/configs/instance-types-data.json",
                     "custom_node_package": self.config.custom_node_package or "",
                     "custom_awsbatchcli_package": self.config.custom_aws_batch_cli_package or "",
+                    "head_node_imds_secured": "true"
+                    if self.config.head_node.imds.secured
+                    or (
+                        self.config.head_node.imds.secured is None
+                        and self.config.scheduling.scheduler in SCHEDULERS_SUPPORTING_IMDS_SECURED
```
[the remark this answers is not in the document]
gmarciani: We could in order to make the code cleaner, but I think the best place for this logic would rather be a function decorated with @property within the config object. I did not do it because `BaseClusterConfig` does not have information about the scheduler so we would have been forced to replicate it in every `[Scheduler]Config` and this is bad. I suppose this part will be covered by C&R. In the meantime I put the logic into a dedicated utility method just to make the code cleaner.

cli/tests/pcluster/templates/test_cluster_stack.py
gmarciani: There is no reason for that: it simply matches the value put into the yaml to check.

gmarciani: There is no reason for that: it simply matches the value put into the yaml to check.

tests/integration-tests/tests/update/test_update.py
gmarciani: 1. Truncated sentence in your comment. Do you mean to add it to an existent test?
2. I think it is necessary to both test from secured, to unsecured and vice versa because different branch in the cookbook are executed in the two cases.

tests/integration-tests/tests/update/test_update.py
gmarciani: I tried, but at the end I discovered that you must use two distinct files due to the current implementation of `pcluster_config_reader`. 
I think `pcluster_config_reader` should be changed accordingly, but since it is out of scope for this PR, I'd stay with this standard approach.

tests/integration-tests/configs/pcluster3.yaml
gmarciani: I think it would be useful only if images in those regions are peculiar in a way that they change the normal interactions with iptables. Otherwise, it would be redundant.
Do you have such suspect? If you don't I'd remain with one standard region.

cli/tests/pcluster/templates/test_cluster_stack.py
gmarciani: Sure, good catch!

tests/integration-tests/tests/common/assertions.py
gmarciani: Sure!

tests/integration-tests/tests/update/test_update.py
tilne: ugh, my bad. meant to say: 'rather than adding a dedicated test for the update of this one parameter, would it be possible to instead add it to existing update tests?'

cli/tests/pcluster/templates/test_cluster_stack.py
tilne: i was talking about the general freezing of time rather than the specific time at which we froze it, because if i were unfamiliar with what's going on with this test i think it would be difficult to understand why time needed to be frozen. either way it's a minor thing.

tilne (dismissed):

cli/tests/pcluster/templates/test_cluster_stack.py
tilne: > i was talking about the general freezing of time rather than the specific time

Ha, sorry I'm now looking at my first comment and I said the exact opposite.

gmarciani: Ok, I'll add a short comment to clarify it.

tests/integration-tests/tests/update/test_update.py
gmarciani: I got your point and I think you're right in terms of frugality.
Nevertheless, even if it is possible, I don't think it would be a good approach to pack everything within a unique update test because we would end up having one giant test for everything.

tilne (approved):

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ per cluster.
 - Add multiple queues and compute resources support for pcluster configure when the scheduler is Slurm.
 - Add prompt for availability zone in pcluster configure automated subnets creation.
 - Remove support for CentOS8.
+- Add configuration HeadNode.Imds.Secured to enable/disable restricted access to IMDS.
```
enrico-usai, line 60: Is it part of your task to add the default values in the output of the pcluster configure command?

tilne: no. that will be handled via a separate task because it's still pending a decision.

tilne (dismissed):

tilne (approved):

PR #2832, opened 2021-06-11
3.0: Prevent DryRun message to be printed as an error.

hanwen-pcluste:
When using DryRun for run_instaces call, it will throw a Client Exception even if the DryRun succeed. Therefore, we need to ignore the exception if the error code is DryRunOperation


cli/src/pcluster/aws/ec2.py
```diff
@@ -361,4 +363,8 @@ def describe_volume(self, volume_id):
     @AWSExceptionHandler.handle_client_exception
     def run_instances(self, **kwargs):
         """Describe network interfaces."""
-        return self._client.run_instances(**kwargs)
+        try:
+            self._client.run_instances(**kwargs)
+        except ClientError as e:
+            if e.response.get("Error").get("Code") != "DryRunOperation":
```
gmarciani, line 369: It is ok to not propagate the dry run exception to the console, but what about keeping track of it with a warning within the log file?

hanwen-pcluste: Our original behavior ignored DryRun exception right away. So I would say it is fine to be consistent with our original behavior

cli/tests/pcluster/aws/test_ec2.py
```diff
@@ -210,3 +210,29 @@ def test_get_ebs_snapshot_info(boto3_stubber, snapshot_id, error_message):
...
             Ec2Client().get_ebs_snapshot_info(snapshot_id)
             assert_that(clienterror.value.code).is_not_equal_to(0)
+
+
+@pytest.mark.parametrize(
+    "error_code, raise_exception",
+    [("DryRunOperation", False), ("UnsupportedOperation", True)],
+)
+def test_run_instances_dryrun(boto3_stubber, error_code, raise_exception):
+    """Verify that if run_instance doesn't generation exception if the error code is DryRunOperation."""
```
gmarciani, line 220: Typo in the comment. Should be "Verify that run_instance doesn't generate exception if the error code is DryRunOperation."

hanwen-pcluste: Done

gmarciani (approved):

PR #2850, opened 2021-06-16
3.0: ImageBuilder initialized with config tests, rather than dict

ermanno:
### Notes
Before, the `ImageBuilder` object was initialized with a dictionary which the caller populated from a parsed YAML configuration file. The dictionary was then written to S3. However in the process of deserializing and serializing the configuration, the content of the uploaded file did not match the one in the original file (for example, when parsing the YAML file into a dictionary comments were removed).  This patch makes the `ImageBuilder` constructor accept the YAML content as a text string. The file uploaded to S3 then has the same content as the one passed as input, and the file is converted into a dictionary only prior to loading the `ImageBuilderSchema` object with the `marshmallow` Python library.

### Tests
Unit tests


hanwen-pcluste (approved):
LGTM!

PR #2856, opened 2021-06-21
Add protected mode and integration test

chenwany:
Reverts commit ab11042
Add integration test

### DIFF
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -519,6 +519,18 @@ class SlurmCommands(SchedulerCommands):
         result = self.submit_command(**submit_command_args)
         return self.assert_job_submitted(result.stdout)
 
+    def get_partition_state(self, partition):
+        """Get the state of the partition."""
+        return self._remote_command_executor.run_remote_command(
+            f'/opt/slurm/bin/scontrol show partition={partition} | grep -oP "State=\\K(\\S+)"'
+        ).stdout
+
+    @retry(wait_fixed=seconds(20), stop_max_delay=minutes(5))
+    def wait_job_running(self, job_id):
+        """Wait till job starts running."""
+        result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id))
+        assert_that(result.stdout).contains("JobState=RUNNING")
+
 
 class TorqueCommands(SchedulerCommands):
     """Implement commands for torque scheduler."""
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -244,6 +244,27 @@ def test_error_handling(scheduler, region, instance, pcluster_config_reader, clu
     )
 
 
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+@pytest.mark.slurm_protected_mode
+def test_slurm_protected_mode(
+    scheduler, region, pcluster_config_reader, clusters_factory, test_datadir, s3_bucket_factory
+):
+    """Test that slurm protected mode logic can handle bootstrap failure nodes."""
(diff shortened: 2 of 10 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -233,6 +233,34 @@ def test_error_handling(scheduler, region, instance, pcluster_config_reader, clu
...
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "preinstall.sh"), "scripts/preinstall.sh")
+    cluster_config = pcluster_config_reader(bucket=bucket_name)
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = get_scheduler_commands(scheduler, remote_command_executor)
+    # Disable protected_mode by setting protected_failure_count to -1
+    _set_protected_failure_count(remote_command_executor, -1)
+    _inject_bootstrap_failures(cluster, bucket_name, pcluster_config_reader)
```
rexcsn, line 252: Very minor: Would it make sense to move these functions into `_test_disable_protected_mode`? This might make the code more readable. Also I consider these functions as the "pre-req" for `_test_disable_protected_mode`, and if we move these into `_test_disable_protected_mode` we can move testing functions around without having to worry about the order of executions.

Similar comments apply to `_enable_protected_mode` and `_set_protected_failure_count` calls below

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -979,3 +1007,150 @@ def _get_num_gpus_on_instance(instance_type_info):
...
+    cluster.stop()
+    _wait_for_computefleet_changed(cluster, "STOPPED")
+    cluster.update()
+    cluster.start()
+    _wait_for_computefleet_changed(cluster, "RUNNING")
+
+
+def _inject_bootstrap_failures(cluster, bucket_name, pcluster_config_reader):
+    """Update cluster to include pre-install script, which introduce bootstrap error."""
+    updated_config_file = pcluster_config_reader(config_file="pcluster.config.update.ini", bucket=bucket_name)
```
rexcsn, line 1032: Very minor: since you have `pcluster.config.recover.ini`, it might be worth to call this config used to break the cluster `pcluster.config.broken.ini`?

tests/integration-tests/tests/schedulers/test_slurm.py
chenwany: Good idea! Done

tests/integration-tests/tests/schedulers/test_slurm.py
chenwany: Renamed

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -519,6 +519,18 @@ def submit_command_and_assert_job_accepted(self, submit_command_args):
...
         return self.assert_job_submitted(result.stdout)
 
+    def get_partition_state(self, partition):
+        """Get the state of the partition."""
+        return self._remote_command_executor.run_remote_command(
+            f'/opt/slurm/bin/scontrol show partition={partition} | grep -oP "State=\\K(\\S+)"'
+        ).stdout
+
+    @retry(wait_fixed=seconds(20), stop_max_delay=minutes(5))
+    def wait_job_running(self, job_id):
```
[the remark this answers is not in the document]
rexcsn: [`wait_job_complete` uses scontrol for slurm](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/tests/common/schedulers_common.py#L293). We can define `_wait_job_state` only for `SlurmCommands` and use it only for `SlurmCommands`

rexcsn (approved):

PR #2860, opened 2021-06-22
3.0: Enable metrics and throttling for ParallelCluster API

demartinofra:


api/infrastructure/deploy-api.sh
```diff
@@ -0,0 +1,73 @@
...
+
+echo "Publishing OpenAPI specs to S3"
+aws s3 cp "${SCRIPT_DIR}/../spec/openapi/ParallelCluster.openapi.yaml" "${S3_UPLOAD_URI}"
+
+echo "Deploying API template"
+aws cloudformation create-stack \
+    --stack-name "ParallelClusterApi" \
+    --template-body file://${SCRIPT_DIR}/parallelcluster-api.yaml \
+    --parameters ParameterKey=ApiDefinitionS3Uri,ParameterValue="${S3_UPLOAD_URI}" ParameterKey=PublicEcrImageUri,ParameterValue="${ECR_ENDPOINT}/${ECR_REPO}:latest" \
+    --capabilities CAPABILITY_AUTO_EXPAND CAPABILITY_IAM
```
ermanno, line 72: I think that here `CAPABILITY_IAM` should be sufficient. `CAPABILITY_AUTO_EXPAND` afaik is used to auto-expand macros, which should not be needed in our case.

ermanno (dismissed):

api/infrastructure/deploy-api.sh
demartinofra: replaced create-stack with deploy

ermanno (approved):

PR #2866, opened 2021-06-23
3.0: HeadNode.Imds.Secured: default value and improvement to validation error messages

gmarciani:
### Changes
1. Set default value for HeadNode.Imds.Secured in pcluster configure
2. Remove conditional default for configuration HeadNode.Imds.Secured
3. Add configuration guidelines to HeadNode IMDS validation error messages

### Tests
1. Manual tests

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -597,7 +597,7 @@ class Imds(Resource):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.secured = Resource.init_param(secured)
+        self.secured = Resource.init_param(secured, default=True)
 
 
 class ClusterIam(Resource):
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -69,7 +69,6 @@ from pcluster.templates.cdk_builder_utils import (
     get_custom_tags,
     get_default_instance_tags,
     get_default_volume_tags,
-    get_property_head_node_imds_secured,
     get_retain_log_on_delete,
     get_shared_storage_ids_by_type,
     get_shared_storage_options_by_type,
@@ -1068,11 +1067,7 @@ class ClusterCdkStack(Stack):
                     "instance_types_data_s3_key": f"{self.bucket.artifact_directory}/configs/instance-types-data.json",
                     "custom_node_package": self.config.custom_node_package or "",
                     "custom_awsbatchcli_package": self.config.custom_aws_batch_cli_package or "",
-                    "head_node_imds_secured": "true"
-                    if get_property_head_node_imds_secured(
-                        self.config.head_node.imds.secured, self.config.scheduling.scheduler
-                    )
-                    else "false",
+                    "head_node_imds_secured": str(self.config.head_node.imds.secured).lower(),
                 },
                 "run_list": f"recipe[aws-parallelcluster::{self.config.scheduling.scheduler}_config]",
             },
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -797,6 +797,11 @@ class HeadNodeImdsValidator(Validator):
                 f"Cannot validate IMDS configuration with scheduler {scheduler}.",
                 FailureLevel.ERROR,
             )
+        elif imds_secured is None:
+            self._add_failure(
+                f"Cannot validate IMDS configuration with Imds.Secured {imds_secured}.",
+                FailureLevel.ERROR,
+            )
         elif imds_secured and scheduler not in SCHEDULERS_SUPPORTING_IMDS_SECURED:
             self._add_failure(
                 f"IMDS secured cannot be enabled in Head Node when using scheduler {scheduler}. "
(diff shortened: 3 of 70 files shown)


hanwen-pcluste (dismissed):
LGTM!

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -794,13 +795,24 @@ class HeadNodeImdsValidator(Validator):
...
         if scheduler is None:
             self._add_failure(
-                f"Cannot validate IMDS configuration with scheduler {scheduler}.",
+                f"Cannot validate IMDS configuration with scheduler {scheduler}."
+                f"Valid values are: {SUPPORTED_SCHEDULERS}.",
+                FailureLevel.ERROR,
+            )
+        elif imds_secured is None:
+            self._add_failure(
+                f"Cannot validate IMDS configuration with Imds.Secured {imds_secured}."
```
tilne, line 804: minor: this is going to result in an error message claiming that the value of this parameter was set to `None`, when more likely it means there's a bug in our handling of the config file. i can't think of a better error message at the moment, but i could see that as being confusing.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1068,11 +1067,7 @@ def _add_head_node(self):
                     "instance_types_data_s3_key": f"{self.bucket.artifact_directory}/configs/instance-types-data.json",
                     "custom_node_package": self.config.custom_node_package or "",
                     "custom_awsbatchcli_package": self.config.custom_aws_batch_cli_package or "",
-                    "head_node_imds_secured": "true"
-                    if get_property_head_node_imds_secured(
-                        self.config.head_node.imds.secured, self.config.scheduling.scheduler
-                    )
-                    else "false",
+                    "head_node_imds_secured": "true" if self.config.head_node.imds.secured else "false",
```
tilne, line 1070: minor: would it make sense to instead use `str(self.config.head_node.imds.secured).lower()` here?

The reason being that if the parameter object is `None` then this defaults to not securing IMDS on the head node. I don't think that can happen as currently written, so this is a bit overly paranoid.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -597,7 +597,7 @@ def __init__(
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.secured = Resource.init_param(secured)
+        self.secured = Resource.init_param(secured, default=True)
```
tilne, line 600: i'm pretty sure this is the change that nullifies my paranoid comment regarding the setting of the IMDS secured cookbook attribute?

tilne (dismissed):

cli/src/pcluster/validators/cluster_validators.py
gmarciani: I agree with you about the confusion between customer misconfig and bug. 
In any case, I would prefer to highlight a bug with an error message rather than assuming it does not happen and forward wrong values. 
In this way, if the bug is there, a customer will see the message repeatedly and at the end will submit a clear support request.

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I agree with you

cli/src/pcluster/config/cluster_config.py
gmarciani: Yes, this ensures that Imds.Secured cannot be None. 
I think the only place where it make sense to enforce that it must not be None is inside the validator.
So, I think we can assume it is not None when setting the dns.json

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I agree with you: in this place we can assume that imds.secured is not None.
I agree because, otherwise, we are silently introducing another defaulting behavior, which is not desired.

cli/src/pcluster/validators/cluster_validators.py
enrico-usai: The basic idea is that Validators should provide "semantic" error messages, without mentioning configuration parameters.

In this case we can print something like: "Cannot validate IMDS configuration if scheduler is not set" or "if IMDS Secured is not set".

gmarciani: So, shouldn't we provide a hint to the customer about valid values?

gmarciani: Talked with Enrico offline. I finally agree on that.

enrico-usai (approved):

tilne (dismissed):

tests/integration-tests/tests/arm_pl/test_arm_pl/test_arm_pl/pcluster.config.yaml
```diff
@@ -6,6 +6,8 @@ HeadNode:
     SubnetId: {{ public_subnet_id }}
   Ssh:
     KeyName: {{ key_name }}
+  Imds:
+    Secured: {{ imds_secured }}
```
tilne, line 10: why did this need to be added? we can't rely on the default for all but the Batch tests?

tilne (approved):

tests/integration-tests/tests/arm_pl/test_arm_pl/test_arm_pl/pcluster.config.yaml
gmarciani: Sure, but I think that as long as we have scheduler as a parameterized config in config file, the same should be done for imds secured because we do not know apriori anything about the scheduler. An alternative approach may be to put a jinja if statement within the config file, but I think this would be conceptually incorrect since config values should be injected by fixtures. 

What do you think?

PR #2868, opened 2021-06-23
3.0: Implement UpdateCluster API

ermanno:
### Notes
This patch implements the UpdateCluster API by calling the methods exposed by the Cluster business object. Things to note:

1. We throw a BadRequest execption in case of an empty change set, unless the caller specified force=True. Note that this change set does not take into account changes to parameters with print_succeded = False.  As far as we can tell at the moment this can happen only for SUCCEDED check result with policies IGNORED or READ_ONLY_RESOURCE_BUCKET. In the future we plan to make these be UNSUPPORTED.
2. The validation of the patch happens after the validation of the configuration. We do not return suppressed validation configuration errors in the BadRequest caused by a failed during patch validation, as we do not want to confuse the customer (i.e. the BadRequest was not caused by the configuration validation erros they suppressed). We do return the suppressed configuration validation errors in the response in the case of a successful request.

### Tests
1. Unit tests
2. Update cluster:
```
{
    "changeSet": [
        {
            "currentValue": "t2.micro",
            "parameter": "HeadNode.InstanceType",
            "requestedValue": "t2.small"
        }
    ],
    "cluster": {
        "cloudformationStackArn": "arn:aws:cloudformation:eu-west-1:395964158899:stack/test-cluster-1/f27a4e20-d2b4-11eb-957d-0a3927392745",
        "cloudformationStackStatus": "UPDATE_IN_PROGRESS",
        "clusterName": "test-cluster-1",
        "clusterStatus": "UPDATE_IN_PROGRESS",
        "region": "eu-west-1",
        "version": "3.0.0"
    },
    "validationMessages": [
        {
            "level": "WARNING",
            "message": "The custom AMI may not have been created by pcluster. You can ignore this warning if the AMI is shared or copied from another pcluster AMI. If the AMI is indeed not created by pcluster, cluster creation will fail. If the cluster creation fails, please gotohttps://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting.html#troubleshooting-stack-creation-failures for troubleshooting.",
            "type": "CustomAmiTagValidator"
        }
    ]
}
```

### DIFF
--- a/cli/src/pcluster/api/controllers/cluster_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/cluster_operations_controller.py
@@ -29,8 +29,10 @@ from pcluster.api.errors import (
     DryrunOperationException,
     InternalServiceException,
     NotFoundException,
+    UpdateClusterBadRequestException,
 )
 from pcluster.api.models import (
+    Change,
     CloudFormationStatus,
     ClusterConfigurationStructure,
     ClusterInfoSummary,
@@ -43,15 +45,23 @@ from pcluster.api.models import (
     InstanceState,
     ListClustersResponseContent,
     Tag,
+    UpdateClusterBadRequestExceptionResponseContent,
     UpdateClusterRequestContent,
     UpdateClusterResponseContent,
+    UpdateError,
     ValidationLevel,
 )
-from pcluster.api.models.cluster_status import ClusterStatus
 from pcluster.aws.aws_api import AWSApi
 from pcluster.aws.common import StackNotFoundError
 from pcluster.cli_commands.compute_fleet_status_manager import ComputeFleetStatus
-from pcluster.models.cluster import Cluster, ClusterActionError, ConfigValidationError
+from pcluster.config.update_policy import UpdatePolicy
+from pcluster.models.cluster import (
+    Cluster,
+    ClusterActionError,
+    ClusterUpdateError,
+    ConfigValidationError,
+    NotFoundClusterActionError,
+)
 from pcluster.models.cluster_resources import ClusterStack
 from pcluster.utils import get_installed_version
 from pcluster.validators.common import FailureLevel
@@ -270,6 +280,8 @@ def list_clusters(region=None, next_token=None, cluster_status=None):
 
 
 @configure_aws_region()
+@convert_errors()
+@http_success_status_code(202)
 def update_cluster(
     update_cluster_request_content: Dict,
     cluster_name,
@@ -305,15 +317,76 @@ def update_cluster(
 
     :rtype: UpdateClusterResponseContent
     """
+    # Set defaults
+    validation_failure_level = validation_failure_level or ValidationLevel.ERROR
+    dryrun = dryrun or False
+    force_update = force_update or False
     update_cluster_request_content = UpdateClusterRequestContent.from_dict(update_cluster_request_content)
-    return UpdateClusterResponseContent(
-        cluster=ClusterInfoSummary(
-            cluster_name="nameeee",
-            cloudformation_stack_status=CloudFormationStatus.CREATE_COMPLETE,
-            cloudformation_stack_arn="arn",
-            region="region",
-            version="3.0.0",
-            cluster_status=ClusterStatus.CREATE_COMPLETE,
+
+    # Validate inputs
+    if client_token:
+        raise BadRequestException("clientToken is currently not supported for this operation")
+    cluster_config = read_config(update_cluster_request_content.cluster_configuration)
+
+    try:
+        cluster = Cluster(cluster_name)
+        if not check_cluster_version(cluster, exact_match=True):
+            raise BadRequestException(
+                f"the update can be performed only with the same ParallelCluster version ({cluster.stack.version}) "
+                "used to create the cluster."
+            )
+
+        if dryrun:
+            cluster.validate_update_request(
+                target_source_config=cluster_config,
+                force=force_update,
+                validator_suppressors=get_validator_suppressors(suppress_validators),
+                validation_failure_level=FailureLevel[validation_failure_level],
+            )
+            raise DryrunOperationException()
+
+        changes, ignored_validation_failures = cluster.update(
+            target_source_config=cluster_config,
+            validator_suppressors=get_validator_suppressors(suppress_validators),
+            validation_failure_level=FailureLevel[validation_failure_level],
+            force=force_update,
+        )
+
+        change_set, _ = _analyze_changes(changes)
+        return UpdateClusterResponseContent(
+            cluster=ClusterInfoSummary(
+                cluster_name=cluster_name,
+                cloudformation_stack_status=CloudFormationStatus.UPDATE_IN_PROGRESS,
+                cloudformation_stack_arn=cluster.stack.id,
+                region=os.environ.get("AWS_DEFAULT_REGION"),
+                version=cluster.stack.version,
+                cluster_status=cloud_formation_status_to_cluster_status(CloudFormationStatus.UPDATE_IN_PROGRESS),
+            ),
+            validation_messages=validation_results_to_config_validation_errors(ignored_validation_failures),
+            change_set=change_set,
+        )
+    except ConfigValidationError as e:
+        raise _handle_config_validation_error(e)
+    except ClusterUpdateError as e:
+        raise _handle_cluster_update_error(e)
+    except (NotFoundClusterActionError, StackNotFoundError):
+        raise NotFoundException(
+            f"cluster '{cluster_name}' does not exist or belongs to an incompatible ParallelCluster major version."
+        )
+
+
+def _handle_cluster_update_error(e):
+    """Create an UpdateClusterBadRequestExceptionResponseContent in case of failure during patch validation.
+
+    Note that patch validation is carried out once we have successfully validated the configuration. For this reason, we
+    want to avoid adding the suppressed configuration validation errors (which we attach to the response in case of a
+    successful update) as we do not want to confuse the customer by showing them errors they suppressed, which did not
+    cause the BadRequest exception.
+    """
+    change_set, errors = _analyze_changes(e.update_changes)
+    return UpdateClusterBadRequestException(
+        UpdateClusterBadRequestExceptionResponseContent(
+            message=str(e), change_set=change_set, update_validation_errors=errors
         )
     )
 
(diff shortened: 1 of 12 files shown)


cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -305,16 +315,81 @@ def update_cluster(
...
+    except ClusterUpdateError as e:
+        raise _handle_cluster_update_error(e)
+    except (NotExistsClusterActionError, StackNotFoundError):
+        raise NotFoundException(
+            f"cluster '{cluster_name}' does not exist or belongs to an incompatible ParallelCluster major version."
+        )
+
+
+def _get_changeset(changes):
+    if changes is None or len(changes) <= 1:
```
demartinofra, line 376: nit: `if not changes` should produce the same behavior

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -305,16 +315,81 @@ def update_cluster(
...
+        return None
+
+    key_indexes = {key: index for index, key in enumerate(changes[0])}
+    return [
+        Change(
+            parameter=row[key_indexes["parameter"]],
```
demartinofra, line 382: let's show the full path for the parameter

cli/src/pcluster/models/cluster.py
```diff
@@ -672,6 +683,35 @@ def stop(self):
         except Exception as e:
             raise _cluster_error_mapper(e, f"Failed when stopping compute fleet with error: {str(e)}")
 
+    def validate_update_request(self, target_source_config, validator_suppressors, validation_failure_level, force):
```
demartinofra, line 686: let's keep adding typing info if possible to the new code

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -305,16 +315,81 @@ def update_cluster(
...
+            requested_value=row[key_indexes["new value"]],
+            current_value=row[key_indexes["old value"]],
+        )
+        for row in changes[1:]
+    ]
+
+
+def _handle_cluster_update_error(e):
+    return UpdateClusterBadRequestException(
+        UpdateClusterBadRequestExceptionResponseContent(message=str(e), change_set=_get_changeset(e.update_changes))
```
demartinofra, line 392: I think we are missing `updateValidationErrors` and `configurationValidationErrors` in this response

cli/src/pcluster/models/cluster.py
```diff
@@ -109,17 +111,26 @@ def __init__(self, message: str):
...
 
 class ConflictClusterActionError(ClusterActionError, Conflict):
-    """Represent an error due to another cluster with the same name already existing."""
+    """Represent an error due to a conflict, such as a stack already existing, or being in the wrong state."""
 
     def __init__(self, message: str):
         super().__init__(message)
 
 
+class NotExistsClusterActionError(ClusterActionError):
```
demartinofra, line 120: nit: NotFound... ?

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
```diff
@@ -1033,29 +1034,411 @@ def test_aws_api_errors(self, client, mocker):
 
 
 class TestUpdateCluster:
```
demartinofra, line 1036: Shall we add a test for force update?

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
```diff
@@ -1033,29 +1034,411 @@ def test_aws_api_errors(self, client, mocker):
...
+                    "clusterConfiguration": BASE64_ENCODED_CONFIG,
+                },
+                ["type:type1", "type:type2"],
+                ValidationLevel.WARNING,
+                False,
+            ),
+        ],
+        ids=["required", "all"],
+    )
+    def test_successful_update_request(
```
demartinofra, line 1099: Shall we test update with no config changes? what happens in this case?

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
```diff
@@ -1033,29 +1034,411 @@ def test_aws_api_errors(self, client, mocker):
...
+            ),
+            pytest.param(
+                {"clusterConfiguration": "aW52YWxpZA=="},
+                "us-east-1",
+                "clusterName",
+                None,
+                None,
+                None,
+                None,
+                {"message": "Bad Request: Cluster update failed.\nconfiguration must be a valid YAML document"},
```
demartinofra, line 1307: configuration not capitalized

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -114,7 +114,7 @@ def test_iam_policies(region, scheduler, pcluster_config_reader, clusters_factor
 
 def _test_s3_access(remote_command_executor, region):
     logging.info("Testing S3 Access")
-    result = remote_command_executor.run_remote_command(f"AWS_DEFAULT_REGION={region} sudo aws s3 ls").stdout
+    result = remote_command_executor.run_remote_command(f"AWS_DEFAULT_REGION={region} aws s3 ls").stdout
```
demartinofra, line 117: how come these changes to integ tests?

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: As agreed, we will compose it following the [YAML path specification](https://github.com/wwkimball/yamlpath/wiki/Segments-of-a-YAML-Path).

tests/integration-tests/tests/iam/test_iam.py
ermanno: They were in your patch, I have put them in a separate commit.

cli/src/pcluster/models/cluster.py
ermanno: Fixed

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: No, `if not changes` would be true if the `changes`  list contains one element (i.e. if it contained only the list of headers)

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
ermanno: The successful test does check that the `Cluster.update` is called with the `forceUpdates` that arrives to the controller. The behavior with the `force` is tested in the tests for the `Cluster` class, but we did not change that functionality in this patch.

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
ermanno: In this case the validation succeeds. The current PCluster 3 behavior is that the update is triggered and then the CLI notifies that there were no updates (see [code](https://github.com/aws/aws-parallelcluster/blob/wip/pcluster3/cli/src/pcluster/cli_commands/update.py#L138))

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
ermanno: Changed old `Cluster` exception code to `f"Cluster update failed.\n{str(e).upper()}"`

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: Added `updateValidationErrors` for issues with the patch. Note that when we run the check on the patch, the configuration has already been successfully validated. We do return the suppressed validation errors with the response in the case of a successful call, however since for patch validation failures we would throw a BadRequest. I was wondering whether adding the suppressed config validation errors to the exception thrown by the patch validation would be confusing (as the suppressed errors were not the cause of the BadRequest failure).

cli/src/pcluster/api/controllers/cluster_operations_controller.py
demartinofra: ah sorry I missed the <=

demartinofra: can you add a comment explaining why?

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
ermanno: As agreed offline, we will throw a BadRequestException in case of an empty changeset.

tests/integration-tests/tests/iam/test_iam.py
demartinofra: hum I might have added these changes by mistake when rebasing.

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -325,3 +398,47 @@ def _handle_config_validation_error(e: ConfigValidationError) -> CreateClusterBa
...
+def _create_message(failure_reason, action_needed):
+    message = None
+    if failure_reason:
+        message = failure_reason
+    if action_needed:
+        message = f"{message}. {action_needed}" if message else action_needed
+    return message or "Error during update"
+
+
+def _get_yaml_path(path, parameter):
```
demartinofra, line 434: can we add a unit test for this?

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -325,3 +398,47 @@ def _handle_config_validation_error(e: ConfigValidationError) -> CreateClusterBa
             configuration_validation_errors=config_validation_messages, message="Invalid cluster configuration"
         )
     )
+
+
+def _analyze_changes(changes):
```
demartinofra, line 403: shall we add a unit test?

cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py
```diff
@@ -1033,29 +1035,447 @@ def test_aws_api_errors(self, client, mocker):
...
+            ),
+            pytest.param(
+                {"clusterConfiguration": "config"},
+                "us-east-1",
+                "clusterName",
+                None,
+                None,
+                None,
+                None,
+                {"message": "Bad Request: invalid configuration. " "Please make sure the string is base64 encoded."},
```
demartinofra, line 1299: nit: join the strings

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: This is exercised through the test for the `update` public method, as the method below. Do we usually test private methods?

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: See above.

ermanno: Note: added `test_handling_of_change_sets` to check top level behavior against various changesets.

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ermanno: Note: added `test_handling_of_change_sets` to check top level behavior against various changesets.

demartinofra (approved):

PR #2871, opened 2021-06-24
[api/client] Add support for sigv4 signing of generated client code w…

charlesg3:

*Issue #, if available:*
2928

*Description of changes:*
- Add a shim for applying sigv4 headers to all client requests.
- Add gradle task to patch shim code on generated client code.


api/example.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
demartinofra, line 1: I'm wondering if we can group all these client specific artifacts in a single dir. A possibility could be:
```
- api/client
--- example.py
--- patch-client.sh
--- src/  # here we put the generated code
--- patch/
```

demartinofra (dismissed):

demartinofra (approved):

PR #2877, opened 2021-06-25
3.0: Remove schema manipulation for Scheduling and SharedStorage sections to fix regression on update checks

enrico-usai:
# Remove schema manipulation for Scheduling section to fix regression on update checks

We needed a way to differentiate Slurm queues schema from Awsbatch queues schema
and we were using an internal manipulation with pre_load to create internal fields
to be used to validate the different schemas.

It was causing problem during the check of the changes during an update
because we were cycling for all the schema variables and these internally created
fields were not present in the schema.
This bug was not discovered by config_patch tests because we were using the old
schema with `Scheduling > Slurm` structure rather than the current schema on
which "slurm" is not a subfolder but just an attribute of `Scheduling > Scheduler:Slurm`

Now instead of doing pre_load, post_dump actions to create internal fields
to differentiate the scheduling schemas are accepting both SlurmQueues and AwsBatchQueues
with different schemas and just differentiate the initialization according to the scheduler.

This is aligned with Slurm/AwsBatchSettings or Efs/Ebs/FsxLustreSettings,
this simplify validation because we don't need to post-process failures from marshmallow
to restore the right section name used by the customer.

## Tests
* Fixed unit tests for config_patch by using schema load and being sure we're using an accepted schema
* Added unit test for a change in the EbsSettings fields
* Updated configuration files in unit tests expected outputs and integration tests
* Added multiple unit tests for SchedulingSchema and SharedStorageSchema
* Updated integration tests configuration files

# Remove schema manipulation for SharedStorageSchema sections to fix regression on update checks

We needed a way to differentiate different storage schemas
and we were using an internal manipulation with pre_load to create internal fields
to be used to validate the different schemas.

It was causing problem during the check of the changes in the update
because we were cycling for all the schema variables and these internally created
fields were not present in the schema.

This bug was not discovered by config_patch tests because we were using the old
schema with `SharedStorage > Ebs` structure rather than the current schema on
which "Ebs" is not a subfolder but just an attribute of `SharedStorage > StorageType:Ebs`

In this case we don't really need pre_load, post_dump actions to create internal fields
to differentiate the storage schemas. We can simply accept both Settings fields
with different schemas and just differentiate the initialization according to the storage type.

Added unit tests for the new validators.

### DIFF
--- a/cli/tests/pcluster/schemas/test_cluster_schema.py
+++ b/cli/tests/pcluster/schemas/test_cluster_schema.py
@@ -17,7 +17,7 @@ import yaml
 from assertpy import assert_that
 from marshmallow.validate import ValidationError
 
-from pcluster.schemas.cluster_schema import ClusterSchema, ImageSchema, SlurmSchema
+from pcluster.schemas.cluster_schema import ClusterSchema, ImageSchema, SlurmSchema, SharedStorageSchema
 from pcluster.utils import load_yaml_dict
 from tests.pcluster.aws.dummy_aws_api import mock_aws_api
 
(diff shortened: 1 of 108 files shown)


hanwen-pcluste (approved):
LGTM!

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -50,7 +50,7 @@ def test_cluster_schema_slurm(mocker, test_datadir, config_file_name):
     _check_cluster_schema(test_datadir, config_file_name)
 
 
-@pytest.mark.parametrize("config_file_name", ["awsbatch.simple.yaml", "awsbatch.full.yaml"])
+@pytest.mark.parametrize("config_file_name", ["awsbatch.simple.yaml"])
```
hanwen-pcluste, line 53: Why deleting awsbatch.full.yaml?

enrico-usai: Fixed

hanwen-pcluste (approved):
LGTM

PR #2878, opened 2021-06-25
3.0: Add value of update key when updating a list element

ermanno:
### Notes
When computing a change, if we are updating a list element we want to embed the value of the element's update key in the path. For example, when modifying the EnableWriteAccess of a bucket in an S3 access policy, currently we would have:

**path: [..., S3Access, BucketName], param: EnableWriteAccess**

while with this patch we wolud have:

**path: [..., S3Access, BucketName[bucket2]], param: EnableWriteAccess**

Note that the BucketName in this case is the update key, which means that if we change the BucketName itself, we would continue to have something like the following instead:

**path: [..., S3Access], param: BucketName, old: bucket2, new: None**
**path: [..., S3Access], param: BucketName, old: None, new bucket3**

### Tests
Fixed and run unit tests.

### DIFF
--- a/cli/src/pcluster/config/config_patch.py
+++ b/cli/src/pcluster/config/config_patch.py
@@ -150,7 +150,7 @@ class ConfigPatch:
 
         If update_key is not set we're considering Name as identifier.
         """
-        update_key = field_obj.metadata.get("update_key", "Name")
+        update_key = field_obj.metadata.get("update_key")
 
         # Compare items in the list by searching the right item to compare through update_key value
         # First, compare all sections from target vs base config and mark visited base sections.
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/config/config_patch.py
```diff
@@ -150,7 +150,7 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
 
         If update_key is not set we're considering Name as identifier.
         """
-        update_key = field_obj.metadata.get("update_key", "Name")
+        update_key = field_obj.metadata.get("update_key")
```
enrico-usai, line 153: We should add a unit tests to verify all the arrays defined in the schema have an update_key defined, to be sure this won't fail at runtime.

ermanno: Agreed... I think that these should be added to the `tests/pcluster/schemas` tests

ermanno: A note... I tried doing a `inspect.getmembers(pcluster.schemas, inspect.ismodule)` which should have returned the list of all submodules, however I checked that it only returned the first one.

enrico-usai (dismissed):

enrico-usai (approved):
Great job!

PR #2886, opened 2021-06-29
3.0 [testing/api-client] Api tests for client <-> api interaction.

charlesg3:

*Issue #, if available:*
PCLUSTER-2987

*Description of changes:*
Added an api-client fixture and a test that exercises it.

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -102,6 +102,10 @@ def pytest_addoption(parser):
     parser.addoption(
         "--api-definition-s3-uri", help="URI of the Docker image for the Lambda of the ParallelCluster API"
     )
+    parser.addoption(
+        "--api-infrastructure-s3-uri", help="URI of the CloudFormation template for the ParallelCluster API"
+    )
+    parser.addoption("--api-uri", help="URI of an existing ParallelCluster API")
     parser.addoption("--instance-types-data-file", help="JSON file with additional instance types data")
     parser.addoption(
         "--credential", help="STS credential endpoint, in the format <region>,<endpoint>,<ARN>,<externalId>.", nargs="+"
@@ -311,6 +315,66 @@ def clusters_factory(request, region):
     yield _cluster_factory
     if not request.config.getoption("no_delete"):
         factory.destroy_all_clusters(test_passed=request.node.rep_call.passed)
+
+
+@pytest.fixture(scope="session")
+def api_server_factory(
+    cfn_stacks_factory, request, public_ecr_image_uri, api_definition_s3_uri, api_infrastructure_s3_uri
+):
+    """Creates a factory for deploying API servers on-demand to each region."""
+    api_servers = {}
+
+    def _api_server_factory(server_region):
+        api_stack_name = generate_stack_name("integ-tests-api", request.config.getoption("stackname_suffix"))
+
+        if not api_infrastructure_s3_uri:
+            stack_template_path = os.path.join("..", "..", "api", "infrastructure", "parallelcluster-api.yaml")
+            with open(stack_template_path) as stack_template_file:
+                stack_template_data = stack_template_file.read()
+
+        params = [
+            {"ParameterKey": "PublicEcrImageUri", "ParameterValue": public_ecr_image_uri},
+            {"ParameterKey": "ApiDefinitionS3Uri", "ParameterValue": api_definition_s3_uri},
+        ]
+
+        if server_region not in api_servers:
+            logging.info(f"Creating API Server stack: {api_stack_name} in {server_region}")
+            try:
+                set_credentials(server_region, request.config.getoption("credential"))
+                stack = CfnStack(
+                    name=api_stack_name,
+                    region=server_region,
+                    parameters=params,
+                    capabilities=["CAPABILITY_IAM", "CAPABILITY_AUTO_EXPAND"],
+                    template=api_infrastructure_s3_uri or stack_template_data,
+                )
+                cfn_stacks_factory.create_stack(stack)
+            finally:
+                unset_credentials()
+            api_servers[server_region] = stack
+        else:
+            logging.info(f"Found cached API Server stack: {api_stack_name} in {server_region}")
+
+        return api_servers[server_region]
+
+    yield _api_server_factory
+
+
+@pytest.fixture(scope="class")
+def api_client(region, api_server_factory, api_uri):
+    """Define a fixture for an API client that interacts with the pcluster api."""
+    from pcluster_client import ApiClient, Configuration
+
+    if api_uri:
+        host = api_uri
+    else:
+        stack = api_server_factory(region)
+        host = stack.cfn_outputs["ParallelClusterApiInvokeUrl"]
+
+    api_configuration = Configuration(host=host)
+
+    with ApiClient(api_configuration) as api_client_instance:
+        yield api_client_instance
 
 
 def _write_cluster_config_to_outdir(request, cluster_config):
@@ -855,14 +919,24 @@ def vpc_stack(vpc_stacks, region):
     return vpc_stacks[region]
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="session")
 def public_ecr_image_uri(request):
     return request.config.getoption("public_ecr_image_uri")
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="session")
+def api_uri(request):
+    return request.config.getoption("api_uri")
+
+
+@pytest.fixture(scope="session")
 def api_definition_s3_uri(request):
     return request.config.getoption("api_definition_s3_uri")
+
+
+@pytest.fixture(scope="session")
+def api_infrastructure_s3_uri(request):
+    return request.config.getoption("api_infrastructure_s3_uri")
 
 
 # If stack creation fails it'll retry once more. This is done to mitigate failures due to resources
--- a/tests/integration-tests/test_runner.py
+++ b/tests/integration-tests/test_runner.py
@@ -65,8 +65,10 @@ TEST_DEFAULTS = {
     "pre_install": None,
     "post_install": None,
     "vpc_stack": None,
+    "api_uri": None,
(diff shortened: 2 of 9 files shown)


tests/integration-tests/conftest.py
```diff
@@ -312,6 +313,60 @@ def _cluster_factory(cluster_config, extra_args=None, raise_on_error=True):
...
 
 
+@pytest.fixture(scope='class')
+def api_server_factory(cfn_stacks_factory, request, public_ecr_image_uri, api_definition_s3_uri):
+    """Creates a factory for deploying API servers on-demand to each region."""
+    api_servers = {}
+
+    def _api_server_factory(server_region):
+        api_stack_name = generate_stack_name("integ-tests-api", request.config.getoption("stackname_suffix"))
+        stack_template_path = os.path.join("..", "..", "api", "infrastructure", "parallelcluster-api.yaml")
```
demartinofra, line 323: can be done separately but we want to use the template passed through the --api-infrastructure-s3-uri test-runner arg so that we decouple the artifacts under test from the checked-out version of integ-tests framework. This is particularly useful when testing released artifacts.

tests/integration-tests/test_runner.py
```diff
@@ -305,6 +306,11 @@ def _init_argparser():
         help="S3 URI of the ParallelCluster API spec",
         default=TEST_DEFAULTS.get("public_ecr_image_uri"),
     )
+    debug_group.add_argument(
```
demartinofra, line 309: can we have an api_group and we move there all api related options?

tests/integration-tests/conftest.py
```diff
@@ -312,6 +313,60 @@ def _cluster_factory(cluster_config, extra_args=None, raise_on_error=True):
         factory.destroy_all_clusters(test_passed=request.node.rep_call.passed)
 
 
+@pytest.fixture(scope='class')
+def api_server_factory(cfn_stacks_factory, request, public_ecr_image_uri, api_definition_s3_uri):
```
demartinofra, line 317: are we sure the api_servers are reused across tests? isn't the class scoped fixture recreated and so the reference to the existing api_servers lost?

tests/integration-tests/conftest.py
```diff
@@ -312,6 +313,60 @@ def _cluster_factory(cluster_config, extra_args=None, raise_on_error=True):
...
+        return api_servers[server_region]
+
+    yield _api_server_factory
+
+
+@pytest.fixture(scope="class")
+def api_client(region, api_server_factory, api_uri):
+    """Define a fixture for an API client that interacts with the
+    pcluster api."""
+    from pcluster_client.api import cluster_operations_api
```
demartinofra, line 355: will the renaming be part of a separate PR?

tests/integration-tests/tests/api/test_api_client/test_api_client/pcluster.config.yaml
```diff
@@ -0,0 +1,21 @@
...
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  SlurmSettings:
+    ScaledownIdletime: {{ scaledown_idletime }}
```
demartinofra, line 12: feel free to remove this and keep the config minimal

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -0,0 +1,57 @@
...
+
+from tests.common.utils import generate_random_string
+
+# Client Interface
+from pcluster_client.model.create_cluster_request_content import CreateClusterRequestContent
+
+@pytest.mark.regions(["us-east-2"])
+@pytest.mark.instances(["c5.xlarge"])
+@pytest.mark.schedulers(["slurm"])
+@pytest.mark.oss(["alinux2"])
```
demartinofra, line 26: feel free to remove these markers that are not required with the new test config approach. we will have to remove them from every test function sooner or later

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _test_describe_cluster(client, cluster, region):
+    response = client.describe_cluster(cluster.name, region=region)
+    assert_that(response['cluster_name']).is_equal_to(cluster.name)
+
+
+def _test_create(api_client, cluster_config, region):
+    cluster_config_data = base64.b64encode(cluster_config.encode('utf-8')).decode('utf-8')
+    body = CreateClusterRequestContent(f"integ-tests-{generate_random_string()}",
```
demartinofra, line 54: you could use the generate_stack_name from utils so that also the suffix is added to the name. This is used to clean-up stacks that are not properly deleted for some reason.
we could also transform generate_stack_name into a fixture so that the suffix is automatically injected from config options

tests/integration-tests/conftest.py
```diff
@@ -312,6 +313,60 @@ def _cluster_factory(cluster_config, extra_args=None, raise_on_error=True):
...
+    if api_uri:
+        host = api_uri
+    else:
+        stack = api_server_factory(region)
+        host = stack.cfn_outputs["ParallelClusterApiInvokeUrl"]
+
+    api_configuration = Configuration(host=host)
+
+    with ApiClient(api_configuration) as api_client_instance:
+        yield cluster_operations_api.ClusterOperationsApi(api_client_instance)
```
demartinofra, line 367: do you plan to return api_client_instance at some point to test endpoints that are not part of the ClusterOperationsApi?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Client Interface
+from pcluster_client.model.create_cluster_request_content import CreateClusterRequestContent
+
+@pytest.mark.regions(["us-east-2"])
+@pytest.mark.instances(["c5.xlarge"])
+@pytest.mark.schedulers(["slurm"])
+@pytest.mark.oss(["alinux2"])
+@pytest.mark.usefixtures("region", "os", "instance")
+def test_api_client(scheduler, region, pcluster_config_reader, clusters_factory, api_client):
```
demartinofra, line 28: when adding tests for the custom image functionalities I'd put them in a separate test function so that image and cluster tests run in parallel.

tests/integration-tests/tests/api/test_api_client/test_api_client/pcluster.config.yaml
```diff
@@ -0,0 +1,21 @@
+Image:
```
demartinofra, line 1: as a reminder for a separate PR: it might be good to make this cluster config use as many features as possible so that we can better test the IAM policy attached to the API.

tests/integration-tests/conftest.py
charlesg3: This parameter didn't exist in my branch, so I added it and made it an override (so that things still work easily locally).

tests/integration-tests/test_runner.py
charlesg3: :+1: done.

tests/integration-tests/conftest.py
charlesg3: I was waiting to hear if it's worth it to try to do a "namespace" package (which would involve moving files), but if not the change to the package name is here:

https://github.com/aws/aws-parallelcluster/pull/2899

tests/integration-tests/conftest.py
charlesg3: Yes --  I sent DM separately, but figured this could be an exercise of getting things end-to-end and expanded to provide the various clients (especially as the first round is iterated on). It's possible to:
- add the other clients here as well
- just yield the `api_client_instance` and then create the `cluster_operations_api` from within the test.

charlesg3: Upon further thought, the second option feels better, at least for now so I pushed that change.

tests/integration-tests/tests/api/test_api_client.py
charlesg3: Agreed -- I renamed this function to `test_cluster_operations` so that is more obvious.

tests/integration-tests/conftest.py
charlesg3: Converted to session based fixtures and tested by adding two tests to simply list clusters, the second test used a cached instance of the server (did not re-create the stack) then the stack was deleted at the end of the tests.

demartinofra:
You can run the following to fix formatting and check code-linters:
```
cd cli
tox -e autoformat
tox -e code-linters
```

tests/integration-tests/conftest.py
```diff
@@ -313,6 +317,62 @@ def _cluster_factory(cluster_config, extra_args=None, raise_on_error=True):
...
+        return api_servers[server_region]
+
+    yield _api_server_factory
+
+
+@pytest.fixture(scope="class")
+def api_client(region, api_server_factory, api_uri):
+    """Define a fixture for an API client that interacts with the
+    pcluster api."""
+    from pcluster_client import Configuration, ApiClient
```
demartinofra, line 362: how is this import working?are you manually modifying your pythonpath to add the client dir?
I'm wondering if it makes sense to add the relative path to the client source in the requirements.txt defined under tests/integration-tests. The downside could be that `pip install -r requirements` will only work from within the tests/integration-tests dir.

charlesg3: I did a `pip install -e .` on the generated code...

_something_ needs to happen but I wasn't sure if it's jenkins as a separate external step or embedded in req's (or something else)?

ermanno:
I rebased the branch and added a commet with the checkstyle fixes.

demartinofra (approved):

ermanno:
As agreed with @demartinofra I have squashed together the various commits, and removed the changes to `tests/integration-tests/configs/*` as the dev Jenkins is not yet ready to run these tests.

demartinofra (approved):

PR #2897, opened 2021-06-30
3.0: Add integration tests covering the integration between pcluster multi-user with opsworks and active directory

gmarciani:
### Changes
1. Add integration tests covering the integration between pcluster multi-user with opsworks and active directory
1. Add to gitignore: pytest.log

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -1047,6 +1047,13 @@ def opsworks_stack_factory(region):
     yield create_stack
 
     for stack_id in stacks:
+        logging.info(f"Retrieving registered instances from opsworks stack: {stack_id}")
+        result = opsworks.describe_instances(StackId=stack_id)
+        instances = [instance["Ec2InstanceId"] for instance in result["Instances"]]
+        for instance in instances:
+            logging.info(f"Deregistering instance #{instance} from opsworks stack: {stack_id}")
+            opsworks.deregister_instance(InstanceId=instance)
+            logging.info(f"Deregistered instance #{instance} from opsworks stack: {stack_id}")
         logging.info(f"Deleting opsworks stack: {stack_id}")
         opsworks.delete_stack(StackId=stack_id)
 
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -1049,7 +1049,7 @@ def opsworks_stack_factory(region):
     for stack_id in stacks:
         logging.info(f"Retrieving registered instances from opsworks stack: {stack_id}")
         result = opsworks.describe_instances(StackId=stack_id)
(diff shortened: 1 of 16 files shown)


tests/integration-tests/conftest.py
```diff
@@ -815,7 +816,7 @@ def create_role(trusted_service, policies=()):
         time.sleep(60)
         logging.info(f"Iam role is ready: {iam_role_name}")
         roles.append({"role_name": iam_role_name, "policies": policies})
-        return iam_role_name
+        return iam_role_name, iam_role_arn
```
tilne, line 819: the uses of `role_factory` will need to be changed to handle the new return type. The only uses I see outside of the multi-user tests are those in test_iam.py:
* https://github.com/aws/aws-parallelcluster/blob/b542d594f3aa9c8980dc1c92ff78df33f7d3bcd0/tests/integration-tests/tests/iam/test_iam.py#L37
* https://github.com/aws/aws-parallelcluster/blob/b542d594f3aa9c8980dc1c92ff78df33f7d3bcd0/tests/integration-tests/tests/iam/test_iam.py#L38
* https://github.com/aws/aws-parallelcluster/blob/b542d594f3aa9c8980dc1c92ff78df33f7d3bcd0/tests/integration-tests/tests/iam/test_iam.py#L60

tests/integration-tests/conftest.py
```diff
@@ -828,6 +829,333 @@ def create_role(trusted_service, policies=()):
...
+
+    yield create_instance_profile
+
+    for instance_profile_name in instance_profiles:
+        logging.info(f"Deleting iam instance profile: {instance_profile_name}")
+        iam.delete_instance_profile(InstanceProfileName=instance_profile_name)
+
+
+@pytest.fixture(scope="class")
+def user_factory(region):
```
tilne, line 866: would it make sense to move these fixtures into the multi-user test?

tests/integration-tests/tests/multi-user/test_multi_user.py
```diff
@@ -0,0 +1,321 @@
...
+    # The opsworks-agent process is run as aws user, so it requires the access to IMDS
+    configure_imds_access(cluster, {"aws": True})
+
+    # Create new user credentials
+    key = RSA.generate(2048)
+    private_key = key.export_key()
+    private_key_path = f"{test_datadir}/private.key"
+    with (open(private_key_path, "wb")) as file:
+        file.write(private_key)
+    public_key = key.publickey().export_key().decode("utf-8")
```
tilne, line 189: it would be nice to wrap this in a function so that it can be used by all the multi-user tests that need to generate a keypair. In the OpenLDAP test case I used subprocess to call ssh-keyegn, but avoiding the external dependency seems cleaner for sure.

PR #2874, opened 2021-06-25
Separate Instance Policies Based on Node Role

tilne:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2922, opened 2021-07-08
3.0: Use shutdown instead of TerminateInstances for computes, Don't include dynamo permissions in Batch head node instance profile

tilne:
This commit changes the user data script run by compute nodes such that
the `shutdown` command is used instead of calling terminate-instances.
This is necessary because PR #2874 removed the permissions to call
TerminateInstances from the default instance profile attached to the
compute nodes.


cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -111,7 +110,7 @@ function error_exit
   # TODO: add possibility to disable this behavior
   # wait logs flush before signaling the failure
   sleep 10
-  aws --region ${!region} ec2 terminate-instances --instance-ids ${!instance_id}
+  shutdown
```
demartinofra, line 113: shall we execute `shutdown -h now` otherwise in case of logged in users it will wait for a default timeout?

tilne: indeed this was necessary

demartinofra (approved):

demartinofra (dismissed):

hanwen-pcluste (approved):

PR #2926, opened 2021-07-09
Reduce and move p4d.24xlarge integ tests

tilne:
This commit does the following two things:
* Reduces the number of p4d.24xlarge tests that are run as part of the
  nightly integration tests against the latest released version.
* Moves the remaining p4d.24xlarge tests to the eu-west-1 region (where
  tests are now only run in the euw1-az2 AZ).

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -512,7 +512,8 @@ AVAILABILITY_ZONE_OVERRIDES = {
     # NAT Gateway not available in sae1-az2
     "sa-east-1": ["sae1-az1", "sae1-az3"],
     # m6g.xlarge instances not available in euw1-az3
-    "eu-west-1": ["euw1-az1", "euw1-az2"],
+    # p4d.24xlarge instances only available in euw1-az2
+    "eu-west-1": ["euw1-az2"],
     # io2 EBS volumes not available in cac1-az4
     "ca-central-1": ["cac1-az1", "cac1-az2"],
     # instance can only be launch in placement group in eun1-az2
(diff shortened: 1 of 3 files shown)


tilne:
CI is failing due to linter errors related to exceptions not having names of the form `*Error`. I'm not planning on fixing those as part of this PR.

tests/integration-tests/conftest.py
```diff
@@ -512,6 +512,7 @@ def cfn_stacks_factory(request):
     # NAT Gateway not available in sae1-az2
     "sa-east-1": ["sae1-az1", "sae1-az3"],
     # m6g.xlarge instances not available in euw1-az3
+    # p4d.24xlarge instances only available in euw1-az2
     "eu-west-1": ["euw1-az1", "euw1-az2"],
```
hanwen-pcluste, line 516: If p4d only exists in euw1-az2, should we remove euw1-az1?

hanwen-pcluste (commented):
Another trick you could consider is to write those sections in released.yaml. THe test config with the same section name will overwrite the config in common.yaml. Therefore, we can keep the test for delvelop while reduce the tests for release

tests/integration-tests/conftest.py
tilne: lol yes definitely, dumb mistake on my part.

tilne:
> Another trick you could consider is to write those sections in released.yaml. The test config with the same section name will overwrite the config in common.yaml. Therefore, we can keep the test for delvelop while reduce the tests for release

doing this now. thank you.

hanwen-pcluste (dismissed):
LGTM

PR #2938, opened 2021-07-14
3.0: Do not convert generic Exceptions at the API controller level.

ermanno:
### Notes
We want to avoid to show generic exceptions' messages in the error returned to the customer. This patch makes the decorator in the controllers simply raise the generic exceptions and we then rely on the conversion in the flask app that produces a error with a generic message, telling the customer to look at the logs for more details.

### Tests
Unit tests.


ermanno:
Closing this PR, after discussion with @demartinofra we chose to leave it as it is for the moment in order to avoid potentially masking useful message from the `Cluster` and `ImageBuilder` business objects.

PR #2946, opened 2021-07-15
Default to DEFAULT_MAX_COUNT in UpdatePolicy.MAX_COUNT condition checker

ermanno:
### Notes
In the `UpdatePolicy.MAX_COUNT` condition checker we want to check that either the fleet is stopped, or the new max count is bigger or equal to the old one (because shrinking queues requires stopping the compute fleet). If the old configuration and/or the new one do not specify a max count, we default to `DEFAULT_MAX_COUNT` in the condition checker.

### Tests
Unit tests.

### DIFF
--- a/cli/src/pcluster/config/update_policy.py
+++ b/cli/src/pcluster/config/update_policy.py
@@ -10,6 +10,8 @@
 # limitations under the License.
 from enum import Enum
 
+from pcluster.constants import DEFAULT_MAX_COUNT
+
 
 class UpdatePolicy:
     """Describes the policy that rules the update of a configuration parameter."""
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/config/update_policy.py
```diff
@@ -149,7 +151,7 @@ def __eq__(self, other):
     fail_reason=lambda change, patch: "Shrinking a queue requires the compute fleet to be stopped first",
     action_needed=UpdatePolicy.ACTIONS_NEEDED["pcluster_stop"],
     condition_checker=lambda change, patch: not patch.cluster.has_running_capacity()
-    or change.new_value >= change.old_value,
+    or (change.new_value or DEFAULT_MAX_COUNT) >= (change.old_value or DEFAULT_MAX_COUNT),
```
enrico-usai, line 154: The  `or` approach is not the good one when working with numbers, if new_value is 0 the code will use DEFAULT_MAX_COUNT, the same is true for old_value.

```
>>> print(0 or "default_value")
default_value
```

enrico-usai: let's add unit tests with new_value or old_value set to 0

ermanno: You are right... no issue, we can write `change.new_value if change.new_value is not None else DEFAULT_MAX_COUNT`

enrico-usai (approved):

PR #2950, opened 2021-07-15
3.0: Add integration test for cluster operation APIs

ermanno:
### Notes
For the API test we want to cover just a minimal subset of feature while we will rely on CLI tests for full coverage (being the CLI a wrapper on top of the API). For the cluster APIs we have 2 test cases, one for Slurm and one for AWS Batch.

The tests' workflow starts by creating a cluster, waiting until it has been created and then performing a list and describe, verifying that it can be found. Then we do a dry-run update, verifying that an eventual update would have succeeded, and lastly we delete the cluster, verifying that it does not appear anymore in the listing. For the moment we do not have cluster instance or compute fleet API calls, when these APIs are ready we will add them in the tests.

### Tests
Successfully run cluster API integration tests.


tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -29,16 +31,32 @@ def _cloudformation_wait(region, stack_name, status):
     waiter.wait(StackName=stack_name)
 
 
-@pytest.mark.usefixtures("region", "os", "instance")
-def test_cluster_operations(request, scheduler, region, pcluster_config_reader, clusters_factory, api_client):
-    cluster_config_path = pcluster_config_reader(scaledown_idletime=3)
+@pytest.mark.usefixtures("os", "instance")
+@pytest.mark.schedulers(["slurm"])
+def test_cluster_slurm(request, region, pcluster_config_reader, api_client):
```
demartinofra, line 36: can you also inject the scheduler fixture and assert that scheduler==slurm? this because we rely on those fixture to publish test metrics and we want to know this test is related to slurm schedueler

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -29,16 +31,32 @@ def _cloudformation_wait(region, stack_name, status):
     waiter.wait(StackName=stack_name)
 
 
-@pytest.mark.usefixtures("region", "os", "instance")
-def test_cluster_operations(request, scheduler, region, pcluster_config_reader, clusters_factory, api_client):
-    cluster_config_path = pcluster_config_reader(scaledown_idletime=3)
+@pytest.mark.usefixtures("os", "instance")
+@pytest.mark.schedulers(["slurm"])
```
demartinofra, line 35: you can remove this

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -47,6 +65,12 @@ def test_cluster_operations(request, scheduler, region, pcluster_config_reader,
 
     _test_list_clusters(client, cluster, region)
     _test_describe_cluster(client, cluster, region)
```
demartinofra, line 67: can we run a describe cluster also right after the create and assert the status?

tests/integration-tests/tests/api/test_api_client.py
ermanno: I have talked about this with @enrico-usai , are you sure that the fixture is not injected when using this annotation? The reasoning here was that we wanted to "force" the usage of `slurm` here since this is a Slurm specific test case, and we did not want a situation like someone specifying a different scheduler in the test config.

tests/integration-tests/tests/api/test_api_client.py
ermanno: See above.

tests/integration-tests/tests/api/test_api_client.py
ermanno: Synced up with @demartinofra off-line, reporting here the notes. Configs have precedence over this annotation. I will do as suggested (i.e. take the config value for the schedulers and add an assertion in the test).

ermanno:
I see that we have compute fleet and cluster instances API available, I will add those four calls to the test cluster workflow in a separate PR after merging this patch.

demartinofra (approved):

PR #2952, opened 2021-07-15
Use new commands for integration tests

hanwen-pcluste:

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -9,6 +9,7 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import json
 import logging
 import re
 import subprocess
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1062,7 +1062,7 @@ def _test_protected_mode(scheduler_commands, remote_command_executor, cluster):
         ],
     )
     # Assert bootstrap failure queues are inactive and compute fleet status is PROTECTED
-    assert_that(cluster.status()).contains("ComputeFleetStatus: PROTECTED")
+    assert_that(cluster.describe_cluster()).contains("ComputeFleetStatus: PROTECTED")
```
charlesg3, line 1065: Looks like this and the below line can be changed into a `check_status` call.

hanwen-pcluste: Done

tests/integration-tests/clusters_factory.py
```diff
@@ -54,13 +55,22 @@ def update(self, config_file, force=True):
...
-            command.append("--force")
-        command.append(self.name)
+            command.extend(["--force-update", "true"])
         result = run_command(command)
-        if "Status: {0} - UPDATE_COMPLETE".format(self.cfn_name) not in result.stdout:
-            error = "Cluster update failed for {0} with output: {1}".format(self.name, result.stdout)
+        logging.info("update-cluster --wait response: %s", result.stdout)
+        response = json.loads(result.stdout)
+        if response.get("cloudFormationStatus") != "UPDATE_COMPLETE":
+            error = f"Cluster update failed for {self.name}"
```
charlesg3, line 73: Should we include the json output in the error?

Otherwise I would combine this line with the following

charlesg3 (approved):
Small nit in output (see comments), otherwise LGTM.

tests/integration-tests/clusters_factory.py
hanwen-pcluste: The output is already printed on Line 70: ` logging.info("update-cluster --wait response: %s", result.stdout)` I wanted to print the output before `json.loads` in case `json.loads` fails.

The `error` string is used both on Line 74 and 75, so it would ok to keep the it. That said, I am not sure about the necessity of putting the `error` in Exception on Line 75.

charlesg3: Ah okay --  yeah that all checks out. Seems reasonable to keep it as-is. :+1:

charlesg3 (approved):
Nice work!

hanwen-pcluste:
Closed in favor of this PRhttps://github.com/aws/aws-parallelcluster/pull/2980

PR #2958, opened 2021-07-16
Add tag Name to shared storage

gmarciani:
### Changes
1. Add tag Name to shared storage, using the value specified in the shared storage name config. Since the name is used as a tag value, it must comply with the standard AWS tag regular expression.

This change will be reflected in integration tests during a dedicated activity on integration testing covering shared storage.

### Tests
1. Manual tests

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -30,6 +30,7 @@ from aws_cdk.core import (
     CfnParameter,
     CfnResourceSignal,
     CfnStack,
+    CfnTag,
     Construct,
     CustomResource,
     Fn,
@@ -654,6 +655,7 @@ class ClusterCdkStack(Stack):
                 storage_type=shared_fsx.fsx_storage_type,
                 subnet_ids=self.config.compute_subnet_ids,
                 security_group_ids=self._get_compute_security_groups(),
+                tags=[CfnTag(key="Name", value=shared_fsx.name)],
             )
             fsx_id = fsx_resource.ref
             # Get MountName for new filesystem
@@ -709,6 +711,7 @@ class ClusterCdkStack(Stack):
                 provisioned_throughput_in_mibps=shared_efs.provisioned_throughput,
                 throughput_mode=shared_efs.throughput_mode,
             )
+            efs_resource.tags.set_tag(key="Name", value=shared_efs.name)
             efs_id = efs_resource.ref
 
         checked_availability_zones = []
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -639,6 +640,7 @@ def _add_fsx_storage(self, id: str, shared_fsx: SharedFsx):
                 storage_type=shared_fsx.fsx_storage_type,
                 subnet_ids=self.config.compute_subnet_ids,
                 security_group_ids=self._get_compute_security_groups(),
+                tags=[CfnTag(key="Name", value=shared_fsx.name)],
```
tilne, line 643: do we want to add a prefix for this like we do for other resources that we create as part of a cluster workflow?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -561,6 +563,27 @@ def _validate(self, efs_id, head_node_avail_zone: str):
...
+class SharedStorageNameValidator(Validator):
+    """
+    Shared storage name validator.
+
+    Validate if the provided name for the shared storage complies with the acceptable pattern.
+    Since the storage name is used as a tag, the provided name must comply with the tag pattern.
+    """
+
+    def _validate(self, name: str):
+        if not re.match(PCLUSTER_SHARED_STORAGE_NAME_REGEX % PCLUSTER_SHARED_STORAGE_NAME_MAX_LENGTH, name):
```
tilne, line 575: The regex is only used here. Can you avoid defining a constant for it? It makes it difficult to know what the resulting regex is here.

gmarciani: The regex is the standard regex for a tag value. 
I think it makes sense to have it defined as a constant, as it has been done for other regex because it can be useful somewhere else.
I will rename the variable with a name that gives the idea that it is meant for tagging values.

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I prefer having the name tag without prefix because:
1. the customer would otherwise have two tags for the same concept: `Name` and `parallelcluster:Name`;
2. it makes sense to prefix non standard tags, but `Name` is a standard tag;

gmarciani: Talked offline with @demartinofra about this. He also agrees on having 'Name' without prefix also because it would be displayed by default on the console.

tilne (approved):

tilne (approved):

PR #2960, opened 2021-07-16
Move ClusterNameValidator into config validation logic

enrico-usai:
This permits to use validator_suppressors logic to skip ClusterNameValidator and permits to have the cluster_name available in the config class to be used for other validators (e.g. dns domain + cluster name total length).

`cluster_name` is now a mandatory parameter for the ClusterSchema initialization.

## Test

Before:
```
($ pcluster create -c cluster-config.yaml test-clusteré£ --suppress-validators
Beginning cluster creation for cluster: test-clusteré£
...
Validating cluster configuration...
ERROR: Cluster creation failed. Configuration is invalid.
Validation failures:
ERROR: Error: The cluster name can contain only alphanumeric characters (case-sensitive) and hyphens. It must start with an alphabetic character and can't be longer than 60 characters.
```

After:
```
$ pcluster create -c cluster-config.yaml test-clusteré£ --suppress-validators
Beginning cluster creation for cluster: test-clusteré£
...
Validating cluster configuration...
Validation succeeded.
...
```

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -43,6 +43,7 @@ from pcluster.validators.awsbatch_validators import (
 )
 from pcluster.validators.cluster_validators import (
     ArchitectureOsValidator,
+    ClusterNameValidator,
     ComputeResourceLaunchTemplateValidator,
     ComputeResourceSizeValidator,
     CustomAmiTagValidator,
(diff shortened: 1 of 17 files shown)


demartinofra (approved):

chenwany (approved):
Great job on moving ClusterNameValidator is moving into the config!

PR #2968, opened 2021-07-19
3.0: Add integration tests for custom image APIs

ermanno:
### Notes
This patch adds integration tests for the custom image APIs.

The workflow is as follows:
* `BuildImage`
* `DescribeImage` after waiting for the CFN stack to be deleted
* `ListImages` of available images
* `DeleteImage` and check that describe does not find it

The patch also adds a separate test for the `DescribeOfficialImages`.

### Tests
Successfully run the integration test.

### DIFF
--- a/tests/integration-tests/tests/api/test_api_client.py
+++ b/tests/integration-tests/tests/api/test_api_client.py
@@ -12,18 +12,27 @@
 
 
 import base64
+import logging
 
 import boto3
 import pytest
 from assertpy import assert_that, soft_assertions
 from botocore.config import Config
 from pcluster_client import ApiException
-from pcluster_client.api import cluster_operations_api
+from pcluster_client.api import cluster_operations_api, image_operations_api
+from pcluster_client.exceptions import NotFoundException
+from pcluster_client.model.build_image_request_content import BuildImageRequestContent
 from pcluster_client.model.cloud_formation_status import CloudFormationStatus
 from pcluster_client.model.cluster_status import ClusterStatus
 from pcluster_client.model.create_cluster_request_content import CreateClusterRequestContent
+from pcluster_client.model.image_build_status import ImageBuildStatus
+from pcluster_client.model.image_status_filtering_option import ImageStatusFilteringOption
 from pcluster_client.model.update_cluster_request_content import UpdateClusterRequestContent
 from utils import generate_stack_name
+
+from tests.common.utils import retrieve_latest_ami
+
+LOGGER = logging.getLogger(__name__)
 
 
 def _cloudformation_wait(region, stack_name, status):
@@ -136,3 +145,79 @@ def _test_delete_cluster(client, cluster, region):
     response = client.list_clusters(region=region)
     cluster_names = [c["cluster_name"] for c in response["items"]]
     assert_that(cluster_names).does_not_contain(cluster_name)
+
+
+def test_official_images(region, api_client):
+    client = image_operations_api.ImageOperationsApi(api_client)
+    response = client.describe_official_images(region=region)
+    assert_that(response.items).is_not_empty()
+
+
+@pytest.mark.usefixtures("instance")
+def test_custom_image(request, region, os, pcluster_config_reader, api_client):
+    base_ami = retrieve_latest_ami(region, os)
+
+    config_file = pcluster_config_reader(config_file="image.config.yaml", parent_image=base_ami)
+    with open(config_file) as config_file:
+        config = config_file.read()
+
+    image_id = generate_stack_name("integ-tests-build-image", request.config.getoption("stackname_suffix"))
+    client = image_operations_api.ImageOperationsApi(api_client)
+
+    _test_build_image(client, image_id, region, config)
+
+    _test_describe_image(client, image_id, region, "BUILD_IN_PROGRESS")
+    _test_list_images(client, image_id, region, "PENDING")
+
+    # CFN stack is deleted as soon as image is available
+    _cloudformation_wait(region, image_id, "stack_delete_complete")
+
+    _test_describe_image(client, image_id, region, "BUILD_COMPLETE")
+    _test_list_images(client, image_id, region, "AVAILABLE")
+
+    _delete_image(client, image_id, region)
+
+
+def _test_build_image(client, image_id, region, config):
+    image_config_data = base64.b64encode(config.encode("utf-8")).decode("utf-8")
+    body = BuildImageRequestContent(image_config_data, image_id, region=region)
+    response = client.build_image(body)
+    LOGGER.info("Build image response: %s", response)
+    assert_that(response.image.image_id).is_equal_to(image_id)
+
+
+def _test_describe_image(client, image_id, region, status):
+    response = client.describe_image(image_id, region=region)
+    LOGGER.info("Describe image response: %s", response)
+    assert_that(response.image_id).is_equal_to(image_id)
+    assert_that(response.image_build_status).is_equal_to(ImageBuildStatus(status))
+
+
+def _test_list_images(client, image_id, region, status):
+    response = client.list_images(image_status=ImageStatusFilteringOption(status), region=region)
+    target_image = _get_image(response.items, image_id)
+
+    while "next_token" in response and not target_image:
+        response = client.list_images(
+            image_status=ImageStatusFilteringOption(status), region=region, next_token=response.next_token
+        )
+        target_image = _get_image(response.items, image_id)
+
+    LOGGER.info("Target image in ListImages response is: %s", target_image)
+
+    assert_that(target_image).is_not_none()
+
+
+def _get_image(images, image_id):
+    for image in images:
+        if image.image_id == image_id:
+            return image
+    return None
+
+
+def _delete_image(client, image_id, region):
+    client.delete_image(image_id, region=region)
+
+    error_message = f"No image or stack associated to parallelcluster image id {image_id}."
+    with pytest.raises(NotFoundException, match=error_message):
+        client.describe_image(image_id, region=region)
--- a/tests/integration-tests/tests/api/test_api_client/test_custom_image/image.config.yaml
+++ b/tests/integration-tests/tests/api/test_api_client/test_custom_image/image.config.yaml
@@ -1,0 +1,4 @@
+Build:
+  InstanceType: {{ instance }}
+  ParentImage: {{ parent_image }}
+  UpdateOsAndReboot: true
--- a/cli/src/pcluster/api/controllers/image_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/image_operations_controller.py
@@ -39,6 +39,7 @@ from pcluster.api.models import (
     ImageInfoSummary,
     ImageStatusFilteringOption,
     ListImagesResponseContent,
+    Tag,
     ValidationLevel,
 )
 from pcluster.api.models.delete_image_response_content import DeleteImageResponseContent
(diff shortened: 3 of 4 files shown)


tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
...
+
+@pytest.mark.usefixtures("instance")
+def test_custom_image(request, region, os, pcluster_config_reader, api_client):
+    base_ami = retrieve_latest_ami(region, os)
+
+    config_file = pcluster_config_reader(config_file="image.config.yaml", parent_image=base_ami)
+    with open(config_file) as config_file:
+        config = config_file.read()
+
+    stack_name = generate_stack_name("integ-tests", request.config.getoption("stackname_suffix"))
```
demartinofra, line 155: we could follow the same approach as in test_createami.py and have a prefix "integ-tests-build-image-"

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
...
+        LOGGER.info("Build image response: %s", resp)
+        image = resp["image"]
+        image_id = image["image_id"]
+        assert_that(image_id).is_equal_to(stack_name)
+
+        # CFN stack is deleted as soon as image is available
+        _cloudformation_wait(region, stack_name, "stack_delete_complete")
+
+        _test_describe_image(client, image, region)
+        _test_list_images(client, image, region)
```
demartinofra, line 169: similar to cluster can we do a describe and a list also when the creation is in progress?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
...
+    response = client.describe_image(image_id, region=region)
+    LOGGER.info("Describe image response: %s", response)
+    assert_that(response.image_id).is_equal_to(image_id)
+
+
+def _test_list_images(client, image, region):
+    image_id = image["image_id"]
+
+    # Despite being a paginated call, available images are returned all in the first page
+    response = client.list_images(image_status="AVAILABLE", region=region)
```
demartinofra, line 190: is there a constant/enum the client packages expose for image status?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
...
+    LOGGER.info("List images response: %s", response)
```
demartinofra, line 191: I know this is useful for debugging but especially in the main dev account this list could become quite long. Shall we log only the entry if found?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
...
+    image_ids = [item["image_id"] for item in response["items"]]
```
demartinofra, line 192: let's use sets rather than list when we need to check that an item is present. same elsewhere

tests/integration-tests/tests/api/test_api_client/test_custom_image/image.config.yaml
```diff
@@ -0,0 +1,10 @@
+Build:
+  InstanceType: {{ instance }}
+  ParentImage: {{ parent_image }}
+  UpdateOsAndReboot: true
+
+DevSettings:
+  Cookbook:
+    ChefCookbook: https://github.com/aws/aws-parallelcluster-cookbook/tarball/refs/heads/develop
+  NodePackage: https://github.com/aws/aws-parallelcluster-node/tarball/refs/heads/develop
+  AwsBatchCliPackage: https://github.com/aws/aws-parallelcluster/tarball/refs/heads/develop
```
demartinofra, line 10: I believe you can remove these since they are injected by the framework when given as param to the test-runner

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +142,62 @@ def _test_delete_cluster(client, cluster, region):
     response = client.list_clusters(region=region)
     cluster_names = [c["cluster_name"] for c in response["items"]]
     assert_that(cluster_names).does_not_contain(cluster_name)
+
+
+@pytest.mark.usefixtures("instance")
+def test_custom_image(request, region, os, pcluster_config_reader, api_client):
```
demartinofra, line 148: can you also add a separate test for official images and we have it fail so that we can realize if the feature is broken after the initial release?

tests/integration-tests/tests/api/test_api_client.py
ermanno: Afaik, there are only constructors that take a string as an input, and validate it against a set of possible values. If you see in the tests below the status is passed to the constructors for the cluster and cfn statuses.

tests/integration-tests/tests/api/test_api_client.py
ermanno: I will re-write this call to unwrap the result, since if we want to check the list of pending images right after a build_image, we cannot rely on getting the target image in the first page of results.

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +143,88 @@ def _test_delete_cluster(client, cluster, region):
...
+        LOGGER.info("Build image response: %s", resp)
+        image = resp["image"]
+        image_id = image["image_id"]
+        assert_that(image_id).is_equal_to(stack_name)
+
+        _test_describe_image(client, image, region)
+        _test_list_images(client, image, region, "PENDING")
+
+        # CFN stack is deleted as soon as image is available
+        _cloudformation_wait(region, stack_name, "stack_delete_complete")
```
demartinofra, line 176: isn't this going to timeout?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +143,88 @@ def _test_delete_cluster(client, cluster, region):
...
+    base_ami = retrieve_latest_ami(region, os)
+
+    config_file = pcluster_config_reader(config_file="image.config.yaml", parent_image=base_ami)
+    with open(config_file) as config_file:
+        config = config_file.read()
+
+    stack_name = generate_stack_name("integ-tests-build-image", request.config.getoption("stackname_suffix"))
+    client = image_operations_api.ImageOperationsApi(api_client)
+
+    with soft_assertions():
```
demartinofra, line 165: does it make sense to keep all of the tests under the soft_assertions?wondering if in most of the cases it makes sense to move ahead with the test if some of the assertions fail

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +143,88 @@ def _test_delete_cluster(client, cluster, region):
...
+
+    response = client.list_images(image_status=ImageStatusFilteringOption(status), region=region)
+    target_image = _get_image(response["items"], image_id)
+    next_token = response.get("next_token", None)
+
+    while next_token and not target_image:
+        response = client.list_clusters(
+            image_status=ImageStatusFilteringOption(status), region=region, next_token=next_token
+        )
+        target_image = _get_cluster(response["items"], image_id)
```
demartinofra, line 208: list_clusters and get_clusters?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +143,88 @@ def _test_delete_cluster(client, cluster, region):
...
+            return image
+    return None
+
+
+def _delete_image(client, image, region):
+    image_id = image["image_id"]
+    client.delete_image(image_id, region=region)
+
+    # Deletion of available image is synchronous, so we do not have to wait as in the cluster case
+    response = client.list_images(image_status="AVAILABLE", region=region)
```
demartinofra, line 228: what about a describe_image which is faster and expect the not found exception?

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -136,3 +143,88 @@ def _test_delete_cluster(client, cluster, region):
...
+    client = image_operations_api.ImageOperationsApi(api_client)
+
+    with soft_assertions():
+        resp = _test_build_image(client, config, region, stack_name)
+        LOGGER.info("Build image response: %s", resp)
+        image = resp["image"]
+        image_id = image["image_id"]
+        assert_that(image_id).is_equal_to(stack_name)
+
+        _test_describe_image(client, image, region)
```
demartinofra, line 172: shall we assert status of the described image?

tests/integration-tests/tests/api/test_api_client.py
ermanno: Not sure about what you mean. [Waiters](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/clients.html#waiters) will poll a resource until it reaches the desired state or an error occurs.

tests/integration-tests/tests/api/test_api_client.py
ermanno: Okay, you are saying that it would make more sense to abort the test as soon as we have an error since the next steps would depend on the state of previous calls... I guess it makes sense, I will remove it also from the cluster APIs tests.

tests/integration-tests/tests/api/test_api_client.py
ermanno: Sure

tests/integration-tests/tests/api/test_api_client.py
ermanno: Sure

tests/integration-tests/tests/api/test_api_client.py
ermanno: Fixed

tests/integration-tests/tests/api/test_api_client.py
ermanno: Actually, now that I look at other integration test, I see that we don't use this context manager. I will remove it from this patch, and in the subsequent PR for the missing cluster instance and compute fleet APIs I will remove it from the cluster tests.

tests/integration-tests/tests/api/test_api_client.py
demartinofra: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Waiter.StackDeleteComplete

Polls CloudFormation.Client.describe_stacks() every 30 seconds until a successful state is reached. An error is returned after 120 failed checks.

according to that it timeouts after 1hour. it can be increased

tests/integration-tests/tests/api/test_api_client.py
```diff
@@ -12,19 +12,28 @@
...
+from pcluster_client.api import cluster_operations_api, image_operations_api
+from pcluster_client.exceptions import NotFoundException
+from pcluster_client.model.build_image_request_content import BuildImageRequestContent
 from pcluster_client.model.cloud_formation_status import CloudFormationStatus
 from pcluster_client.model.cluster_status import ClusterStatus
 from pcluster_client.model.create_cluster_request_content import CreateClusterRequestContent
+from pcluster_client.model.image_build_status import ImageBuildStatus
+from pcluster_client.model.image_status_filtering_option import ImageStatusFilteringOption
 from pcluster_client.model.update_cluster_request_content import UpdateClusterRequestContent
 from utils import generate_stack_name
```
demartinofra, line 31: shall we change this to import from tests.common.utils?

demartinofra (approved):

tests/integration-tests/tests/api/test_api_client.py
ermanno: I see that in all other tests (security groups, createami, etc.) we use this import. I tried changing it to `from tests.common.utils import generate_stack_name`, but with this last import in is not found.

PR #2972, opened 2021-07-19
Add parameter DeleteOnTermination in head node root volume and DeletionPolicy in Shared EBS

hanwen-pcluste:
See commit descriptions for details


tests/integration-tests/tests/storage/test_deletion_policy.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+from assertpy import assert_that
+from botocore.exceptions import ClientError
+from utils import get_compute_nodes_instance_ids, get_root_volume_id
+
+from tests.tags.test_tag_propagation import get_head_node_instance_id
+
+
+@pytest.mark.usefixtures("instance", "scheduler")
+def test_retain_on_deletion(pcluster_config_reader, clusters_factory, region, os):
```
demartinofra, line 24: can we extract resource cleanup in a fixture in order to make it more robust and attempt a cleanup also when assertions fail? An idea could be to have the following fixture

```
@pytest.fixture
def cleanup_retained_resources_for_cluster():
    cluster = None
    def _register_cluster(c):
        cluster = c

   yield _register_cluster

   # do the cleanup by searching for resources and deleting them (if one of them fails to be deleted let's not stop deletion of the others)
```

tests/integration-tests/tests/storage/test_deletion_policy.py
```diff
@@ -0,0 +1,72 @@
...
+        compute_root_volumes.append(get_root_volume_id(compute_node, region, os))
+    logging.info("Compute root volume %s", compute_root_volumes)
+
+    logging.info("Checking no snapshot with the tag of stack id is created after stack deletion")
+    ec2_client = boto3.client("ec2")
+    snapshots = ec2_client.describe_snapshots(
+        Filters=[
+            {"Name": "tag:aws:cloudformation:stack-id", "Values": [stack_arn]},
+        ]
+    )["Snapshots"]
```
demartinofra, line 46: can you extract portions of this test in private functions, especially the duplicated ones like this? so that you also improve code readability

tests/integration-tests/tests/storage/test_deletion_policy.py
hanwen-pcluste: Done by doing the clean up in cluster_factory

tests/integration-tests/tests/storage/test_deletion_policy.py
hanwen-pcluste: Done

tests/integration-tests/tests/storage/test_deletion_policy.py
```diff
@@ -0,0 +1,72 @@
...
+    head_node_root_volume = get_root_volume_id(head_node_instance_id, region, os)
+    compute_node_instance_ids = get_compute_nodes_instance_ids(cluster.name, region)
+    logging.info("Checking at least one compute node is running")
+    assert_that(len(compute_node_instance_ids)).is_greater_than_or_equal_to(1)
+    compute_root_volumes = []
+    for compute_node in compute_node_instance_ids:
+        compute_root_volumes.append(get_root_volume_id(compute_node, region, os))
+    logging.info("Compute root volume %s", compute_root_volumes)
+
+    logging.info("Checking no snapshot with the tag of stack id is created after stack deletion")
```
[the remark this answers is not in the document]
hanwen-pcluste: Done

tests/integration-tests/clusters_factory.py
```diff
@@ -88,6 +102,7 @@ def delete(self, delete_logs=False):
         """Delete this cluster."""
         if self.has_been_deleted:
             return
+        self.cfn_stack_arn  # Cache cfn stack arn before stack deletion.
```
demartinofra, line 105: nit: I'd rather initialize self.__cfn_stack_arn in create() function

tests/integration-tests/clusters_factory.py
```diff
@@ -113,6 +128,7 @@ def delete(self, delete_logs=False):
                 error = "Cluster deletion failed for {0} with output: {1}".format(self.name, result.stdout)
                 logging.error(error)
                 raise Exception(error)
+            self._delete_resource_by_stack_id_tag()
```
demartinofra, line 131: shall we move this before the if so that we try this deletion also in case of a delete failred?

tests/integration-tests/clusters_factory.py
```diff
@@ -285,12 +301,41 @@ def cfn_resources(self):
...
         self.__cfn_outputs = None
         self.__cfn_resources = None
 
+    def _delete_resource_by_stack_id_tag(self):
+        self._delete_snapshots()
+        self._delete_volumes()
+
+    @suppress_and_log_exception
+    def _delete_snapshots(self):
+        ec2_client = boto3.client("ec2")
```
demartinofra, line 327: do we need to set the region? same in the other function

tests/integration-tests/clusters_factory.py
```diff
@@ -28,6 +30,17 @@
 )
 
 
+def suppress_and_log_exception(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            logging.error(e)
```
demartinofra, line 39: let's improve this log message to something like: `"Failed when running function {func.__name__}. Ignoring failure. Error: {str(e)}"`

tests/integration-tests/tests/storage/test_deletion_policy.py
```diff
@@ -0,0 +1,68 @@
...
+    head_node_root_volume = get_root_volume_id(head_node_instance_id, region, os)
+    compute_node_instance_ids = get_compute_nodes_instance_ids(cluster.name, region)
+    logging.info("Checking at least one compute node is running")
+    assert_that(len(compute_node_instance_ids)).is_greater_than_or_equal_to(1)
+    compute_root_volumes = []
+    for compute_node in compute_node_instance_ids:
+        compute_root_volumes.append(get_root_volume_id(compute_node, region, os))
+    logging.info("Compute root volume %s", compute_root_volumes)
+
+    ec2_client = boto3.client("ec2")
```
demartinofra, line 40: region_name?

tests/integration-tests/clusters_factory.py
hanwen-pcluste: Done

tests/integration-tests/clusters_factory.py
hanwen-pcluste: Done.
My concern was it would be possible to delete root volume attached to an instance if the stack deletion failed. I tried to call boto3 `delete_volume` to delete an attached root volume and it just threw an exception without deleting the volume. So this new change is safe.

tests/integration-tests/clusters_factory.py
hanwen-pcluste: IMO, it is optional. But I added them

tests/integration-tests/tests/storage/test_deletion_policy.py
hanwen-pcluste: Added

hanwen-pcluste: In individual test, it is very safe to not add the region_name, because it is set in the environment in through a fixture

tests/integration-tests/clusters_factory.py
demartinofra: I believe without region it breaks if running tests in sequential mode when multiple regions tests are executed in the same process, or are we properly exporting the AWS_DEFAULT_REGION var before each test?

tests/integration-tests/tests/storage/test_deletion_policy.py
demartinofra: you are right I had missed we have that fixture

demartinofra (approved):

demartinofra (approved):

PR #2902, opened 2021-07-01
3.0 Improve build image test and create image factory

yuleiwan:
* Refactor build image integration test. 
** image factory to manage image creation, deletion, describe and log retrieve
** use describe-images to check image tags instead of using awscli describe images
** retrieve and check build log in test_build_image_wrong_pcluster_version

Signed-off-by: Yulei Wang <yuleiwan@amazon.com>

**Please See** [Git Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions)

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3003, opened 2021-07-22
3.0: Clean up test_api_client resources in case of test failures

ermanno:
### Notes
This patch introduces two fixtures to handle the clean up of clusters of images in case the API tests fail half way.

Note that we have another patch changing the way we do cleanup for the `build_image` operation (https://github.com/aws/aws-parallelcluster/pull/2902). After merging this patch we will be able to re-write the `build_image` for the API tests in a similar fashion (to something more like what we have in this patch for the `create_cluster`).

### Tests
Run integration tests with erroneous configurations, verifying that the stacks and images were removed.

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -62,6 +62,13 @@ class Cluster:
         attrs = ", ".join(["{key}={value}".format(key=key, value=repr(value)) for key, value in self.__dict__.items()])
         return "{class_name}({attrs})".format(class_name=self.__class__.__name__, attrs=attrs)
 
+    def mark_as_created(self):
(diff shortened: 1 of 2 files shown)


tests/integration-tests/clusters_factory.py
```diff
@@ -375,7 +382,7 @@ def create_cluster(self, cluster, extra_args=None, raise_on_error=True):
                 raise Exception(error)
         else:
             logging.info("Cluster {0} created successfully".format(name))
-            cluster.create_complete = True
+            cluster.mark_an_created()
```
demartinofra, line 385: typo

demartinofra (approved):

PR #3008, opened 2021-07-22
[API] Align timestamp formats.

charlesg3:

*Issue #, if available:*
PCLUSTER-3680

*Description of changes:*
Make all api responses return timestamp in ISO 8601 format with milliseconds.

### DIFF
--- a/cli/src/pcluster/utils.py
+++ b/cli/src/pcluster/utils.py
@@ -8,6 +8,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+import datetime
 import itertools
 import json
 import logging
@@ -19,12 +20,12 @@ import sys
 import time
 import urllib.request
 import zipfile
-from datetime import datetime
 from io import BytesIO
 from shlex import quote
 from typing import NoReturn
 from urllib.parse import urlparse
 
+import dateutil.parser
 import pkg_resources
 import yaml
 from dateutil import tz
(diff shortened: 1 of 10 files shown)


cli/src/pcluster/utils.py
```diff
@@ -298,7 +318,8 @@ def timestamp_to_isoformat(timestamp, timezone=None):
     if not timezone:
         timezone = tz.tzlocal()
     # Forcing microsecond to 0 to avoid having them displayed.
-    return datetime.fromtimestamp(timestamp / 1000, tz=timezone).isoformat(timespec="seconds")
+    time_ = datetime.datetime.fromtimestamp(timestamp / 1000, tz=timezone)
+    return time_.isoformat(timespec="seconds")
```
hanwen-pcluste, line 322: Could you explain this change?

cli/tests/pcluster/api/controllers/test_cluster_instances_controller.py
```diff
@@ -10,6 +10,7 @@
 
 from pcluster.api.models import NodeType
 from pcluster.aws.common import StackNotFoundError
+from pcluster.utils import to_iso_time
```
hanwen-pcluste, line 13: minor: should we try to avoid using `pcluster.utils` in testing?

cli/src/pcluster/utils.py
charlesg3: I needed access to the `datetime.date` member of the `datetime` module. Previously we were using `from datetime import datetime` which I wanted to convert to `import datetime`... so when I made this change we ended up with the `datetime.datetime.fromtimestamp(...)` call, which made the line pretty long so I decided to break it up from a chain of calls so the line length would be shorter.

cli/tests/pcluster/api/controllers/test_cluster_instances_controller.py
charlesg3: I'm not sure here -- I've generally used converters in the like in testing code to ensure that the logic internally conforms to the conversion operation.

charlesg3: Has this come up in the past? Another option is to do the conversions so we have exact expected output.

cli/src/pcluster/utils.py
```diff
@@ -123,6 +124,25 @@ def camelcase(snake_case_word):
     return "".join(word.title() for word in parts)
 
 
+def to_iso_time(time_in):
```
hanwen-pcluste, line 127: How does this compare with `def timestamp_to_isoformat`?

cli/tests/pcluster/api/controllers/test_cluster_instances_controller.py
hanwen-pcluste: I vaguely remember someone mentioned it. But in this case, I believe it is fine

hanwen-pcluste (approved):
LGTM!

PR #3027, opened 2021-07-26
integ-tests: test cluster CLI commands

hanwen-pcluste:
* Test `create-cluster` (`--dryrun`, `--suppress-validators`, `--validation-failure-level`, error response, successful response).

* Test `update-cluster` (`--dryrun`, `--suppress-validators`, `--validation-failure-level`, error response). The successful case of `update-cluster` is tested in `test_update_slurm` and `test_update_awsbatch`

* Add wait (default is True) parameter on tests. Therefore we can test the response with `wait==False`. (Note that when `wait==True`, the response is always from describe-cluster, so we need to set it to `False` if we want to test the response for create/update/delete

* Split `_test_pcluster_instances_and_status` to `_test_describe_instances` and `_test_pcluster_compute_fleet` to make testing function names to be consistent with the commands under testing

* Other minor renaming

* Fix retrieval of `cfn_stack_arn` in `class Cluster`: Previous code could not retrieve the arn if stack creation failed. The `cfn_stack_arn` is used to clean up resources according to stack-id tag

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -56,7 +56,7 @@ class Cluster:
         self.__cfn_parameters = None
         self.__cfn_outputs = None
         self.__cfn_resources = None
-        self.cfn_stack_arn = None
+        self.__cfn_stack_arn = None
 
     def __repr__(self):
         attrs = ", ".join(["{key}={value}".format(key=key, value=repr(value)) for key, value in self.__dict__.items()])
@@ -69,10 +69,14 @@ class Cluster:
         """
         self.create_complete = True
 
-    def update(self, config_file, force=True):
+    def update(self, config_file, force=True, extra_args=None, raise_on_error=True, wait=True, log_error=True):
         """
         Update a cluster with an already updated config.
         :param force: if to use --force flag when update
+        :param extra_args: list of strings; extra args to pass to `pcluster update-cluster`
+        :param raise_on_error: raise exception if cluster creation fails
+        :param wait: wait for update completion
+        :param log_error: log error when error occurs. This can be set to False when error is expected
         """
         # update the cluster
         logging.info("Updating cluster {0} with config {1}".format(self.name, config_file))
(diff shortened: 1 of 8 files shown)


tests/integration-tests/clusters_factory.py
```diff
@@ -187,25 +188,38 @@ def describe_cluster(self):
...
+            response = json.loads(result.stdout)
+            logging.info("Describe cluster %s compute fleet successfully", self.name)
+            return response
+        except subprocess.CalledProcessError as e:
+            logging.error(
+                "Failed when getting cluster compute fleet with error:\n%s\nand output:\n%s", e.stderr, e.stdout
+            )
+            raise
+
+    def instances(self, node_type=None, queue_name=None):
```
charlesg3, line 205: I think we might want to update this function name to more directly represent the underlying call that it makes to pcluster (e.g. "describe_cluster_instances")... though a bit verbose I think it's better to be consistent / explicit.

charlesg3: Looking further down, I'm actually curious if it's better to have it just return the instances (e.g. `return response["instances"]`) or if it should be called `get_cluster_incance_ids` to reflect that it is just returning the ids?

tests/integration-tests/tests/cli_commands/test_cli_commands.py
```diff
@@ -19,43 +20,240 @@
...
+            "type": "KeyPairValidator",
+            "message": ".*you do not specify a key pair.*",
+        }
+        invalid_cluster_configuration_message = "Invalid cluster configuration"
+        expected_creation_response = {
+            "configurationValidationErrors": [custom_ami_tag_warning, key_pair_warning, name_error],
+            "message": invalid_cluster_configuration_message,
+        }
+        # Set log_error=False if error is expected. Therefore, the test log won't be flooded with errors
+        cluster = clusters_factory(
```
charlesg3, line 88: It looks like there's a decent amount of code below here that is structurally the same... I'm curious if we can turn this into a more data-based approach where we have a list of args and a list of expected responses and then iterate through each of them performing the same:
```
    logging.info(f"Testing {test_description}")
    cluster = clusters_factory(cluster_config_with_warning, extra_args, ...)
    _check_response_with_configuration_validation(cluster.creation_response, expected_creation_response)
```

This would put the inputs next to the outputs and only have the actual operations in a single place.

hanwen-pcluste: Done

tests/integration-tests/clusters_factory.py
hanwen-pcluste: Done by renaming it to get_cluster_instance_ids.

charlesg3 (approved):

PR #3029, opened 2021-07-27
Fix master and compute node volume size regex

yuleiwan:
* Support size range from 35 to 9999 GiB

*Issue #, if available:*

*Description of changes:*


cli/src/pcluster/config/mappings.py
```diff
@@ -160,7 +160,7 @@
     "efs_fs_id": r"^fs-[0-9a-z]{8}$|^fs-[0-9a-z]{17}|NONE$",
     "file_path": r"^\/?[^\/.\\][^\/\\]*(\/[^\/.\\][^\/]*)*$",
     "fsx_fs_id": r"^fs-[0-9a-z]{17}|NONE$",
-    "greater_than_35": r"^([1-9]+[0-9]{2}|[4-9][0-9]|3[5-9])$",
+    "greater_than_35": r"^([1-9][0-9]{3}|[1-9][0-9]{2}|[4-9][0-9]|3[5-9])$",
```
enrico-usai, line 163: We can avoid the upper-bound limit, the regex should be:
`r"^([1-9][0-9]{2,}|[4-9][0-9]|3[5-9])$"`

yuleiwan: updated, but I am not sure we should remove upper-bound limit, because the maximum volume size is 16TiB

enrico-usai (approved):

PR #3030, opened 2021-07-27
Scope PassRole permissions based on specified instance roles

tilne:
### Scope PassRole permissions based on specified instance roles

This commit modifies the PassRole permissions that are attached to the
default head node instance profile. Previously it was assumed that the
head node would need only enable PassRole permissions for IAM roles with
a name like "/parallelcluster/". This default permission is still
included when a custom instance role is not specified.

### Restrict PassRole permissions to cluster resources

This commit modifies the PassRole permissions granted by default to the
instance profile created for the head node such that it can only perform
the action for resources specific to a cluster. It also modifies the
names given to some of the resources in order to enable this restriction.

### Return results of IAM API calls

Previously the methods calling IAM APIs were not returning the results.
This was because the callers were only interested in whether the methods
raised an exception or not.

### Include PassRole permissions for custom compute node instance profiles

This commit modifies the PassRole permissions attached to the default
head node instance profile such the action can now be called on custom
instance profiles specified for compute queues. Previously this logic
only accounted for the custom instance roles specified for compute
queues.


cli/src/pcluster/templates/slurm_builder.py
```diff
@@ -677,6 +675,30 @@ def _add_compute_resource_launch_template(
...
+        default_pass_role_resource = self._format_arn(
+            service="iam",
+            region="",
+            resource=f"role{self._cluster_scoped_iam_path()}*",
+        )
+
+        # If there are any queues where a custom instance role was specified,
+        # enable the head node to pass permissions to those roles.
+        custom_queue_role_arns = [
+            arn for queue in self.config.scheduling.queues for arn in queue.iam.instance_role_arns
```
hanwen-pcluste, line 689: What if the same custom role is specified in multiple queues? Is it ok to have redundant resources in the policy?

hanwen-pcluste (commented):
Great work!

cli/src/pcluster/templates/slurm_builder.py
tilne: I tested with a config containing duplicated instance role values in two different compute queues. Everything worked as expected, but I agree that this isn't the best way to handle this. The method implementing the `iam.instance_role_arns` param could remove duplicates first, but I think it also makes more sense to have `_generate_head_node_pass_role_resources` remove the duplicates, because there's a default value that may need to be added there.

That consideration also points out that the logic to figure out whether or not to include the default value currently uses the length of the value returned by `iam.instance_role_arns`. Since a given queue can have multiple instance role arns pointed to by a custom instance profile, that doesn't work. I'll change that as well so that it looks for any queues that don't specify any instance role or instance profile.

hanwen-pcluste (approved):

cli/src/pcluster/templates/slurm_builder.py
hanwen-pcluste: Thank you for testing it out. Just out of curiosity, did IAM carry those duplicated resources or merge them together?

tilne: The duplicates are not present in the template generated for the cluster stack. They are being removed sometime between when the resource policy construct is initialized and when the cdk app is synthesized. My guess is that happens when the construct is initialized but I don't know for sure.

demartinofra (approved):

PR #3035, opened 2021-07-28
3.0: Handle differences between AWS Batch and Slurm in API tests

ermanno:
### Notes
When Slurm clusters are created, the instances are ready as soon as the stack creation is completed. With AWS Batch this is not true, so we add a step to wait until the desired number of instances are present in the cluster. Furthermore, the queue name for AWS Batch clusters is not populated (not that we can specify only one queue), and so this patch handles also that corner case when creating the compute_node_map.

### Tests
Run API integration tests.

### DIFF
--- a/tests/integration-tests/tests/api/test_api.py
+++ b/tests/integration-tests/tests/api/test_api.py
@@ -16,6 +16,7 @@ import logging
 import boto3
 import pytest
 from assertpy import assert_that
+from benchmarks.common.util import get_instance_vcpus
 from botocore.config import Config
 from clusters_factory import Cluster, ClustersFactory
 from pcluster_client import ApiException
@@ -39,9 +40,11 @@ from pcluster_client.model.update_cluster_request_content import UpdateClusterRe
 from pcluster_client.model.update_compute_fleet_request_content import UpdateComputeFleetRequestContent
 from utils import generate_stack_name
 
+from tests.common.assertions import wait_for_num_instances_in_cluster
 from tests.common.utils import retrieve_latest_ami
 
 LOGGER = logging.getLogger(__name__)
+NUM_OF_COMPUTE_INSTANCES = 2
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/api/test_api.py
```diff
@@ -99,10 +101,18 @@ def _cloudformation_wait(region, stack_name, status):
...
+
 def _ec2_wait_terminated(region, instances):
+    _ec2_wait(region, instances, "terminated")
+
+
+def _ec2_wait(region, instances, status):
     config = Config(region_name=region)
     ec2 = boto3.client("ec2", config=config)
-    waiter = ec2.get_waiter("instance_terminated")
+    waiter = ec2.get_waiter(f"instance_{status}")
```
lukeseawalker, line 115: minor: since this is a generic `get_waiter` and not just a getter for `instance_*` waiters, I would keep this generic, and pass the full waiter name from the outside

tests/integration-tests/tests/api/test_api.py
```diff
@@ -220,6 +237,10 @@ def _test_stop_compute_fleet(region, cluster_compute_fleet_client, cluster_insta
     compute_node_map = _test_describe_cluster_compute_nodes(region, cluster_instances_client, cluster_name)
     instances_to_terminate = _get_instances_to_terminate(compute_node_map)
 
+    # EC2 instances in "running" or "stopping" state cannot be terminated, so we will wait for the instances we
```
lukeseawalker, line 240: comment is confusing, you are stating that instance in "running" cannot be terminated and then you wait instance to be running to terminate them. Can you please clarify?

tests/integration-tests/tests/api/test_api.py
```diff
@@ -187,15 +201,18 @@ def _test_describe_cluster_compute_nodes(region, client, cluster_name, all_termi
 
 def _add_compute_nodes(instances, compute_node_map):
     for instance in instances:
-        if instance.queue_name not in compute_node_map:
-            compute_node_map[instance.queue_name] = set()
-        compute_node_map[instance.queue_name].add(instance.instance_id)
+        # The AWS Batch queue name is not populated
+        queue_name = instance.get("queue_name", "aws_queue")
```
lukeseawalker, line 205: what is "aws_queue"? did you mean "awsbatch_queue"?

tests/integration-tests/tests/api/test_api.py
ermanno: Agreed.

tests/integration-tests/tests/api/test_api.py
ermanno: Typo, I meant to write "pending".

tests/integration-tests/tests/api/test_api/test_cluster_slurm/pcluster.config.update.yaml
```diff
@@ -1,75 +1,30 @@
 Image:
   Os: {{ os }}
-Tags:
```
hanwen-pcluste, line 3: Why would the test be simplified?

hanwen-pcluste (commented):
Great work!

tests/integration-tests/tests/api/test_api/test_cluster_slurm/pcluster.config.update.yaml
ermanno: The Slurm test was unnecessarily complex for what we intended to test (I originally copied it from the configuration used for the update config tests), I reduced it to a more "basic" case that is sufficient for our purposes.

demartinofra (approved):

PR #3054, opened 2021-07-29
3.0: Create infrastructure to update API PC image in local repository with ImageBuilder pipeline

ermanno:
### Notes
This patch adds to the SAM template to provision the API infrastructure the resources needed to support updating a given version of the ParallelCluster Docker image to the latest version using an ImageBuilder pipeline. For upgrades to a new version of the API we simply re-run the SAM template. 

This patch crease an ImageBuilder pipeline, an SNS topic to use to be notified when the updated image is available, and a Lambda to make the ParallelClusetr Lambda point to the new Docker image. We also modify the custom resource responsible for cleanup by making it also remove all ECR Docker images of the version of the API that we are deleting or upgrading, and also all the ImageBuilder images created by the ImageBuilder pipeline (the first one, created by the CFN template, is delete automatically when updating or deleting the stack).

#### Minor changes
I did the following additional changes to the custom resource used for deletion (in separate commits):
1. Create a physical resource id in order to avoid having to always recreate the custom resource (by default the log stream name is used, which frequently trigger replacement on updates, given the reality of log rotation).
1. Use the `reason` parameter to communicate the reason for the CFN status (see [module source code](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html#w2ab1c27c23c16b9c15) for usage and [custom resource response object](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html) for the definition of the response fields).

#### Notes for the reviewer
I have added the PC version to the IB pipeline name so as to trigger a replacement when we update the stack with a new template. Technically if we change only the url of the PC parent image it could be updated in place (indeed from [the documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-imagebuilder-imagepipeline.html) changing the name is the only thing that can trigger a replacement, all other changes are made in place), but logically I think it makes more sense to replace the resource (i.e. since we delete the old PC imagebuilder and ecr images) so as to have clear statistics (i.e. last time the pipeline was run, etc.). If you don't agree it's a simple matter of removing the PC version from the pipeline name.

### Tests
Manual tests (create stack + update with pipeline, upgrade to new version + update with pipeline, delete stack).


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -969,47 +1064,70 @@ Resources:
...
-              build_number = parts[-1]
-              return '{}-{}'.format(version, build_number)
+          imagebuilder = boto3.client('imagebuilder')
+
+          def get_image_ids(repository_name, version):
+              image_ids = []
+              paginator = ecr.get_paginator('list_images')
+              response_iterator = paginator.paginate(repositoryName=repository_name, filter={'tagStatus': 'TAGGED'})
+              for response in response_iterator:
+                  image_ids.extend([{'imageTag': image_id['imageTag']} for image_id in response['imageIds'] if image_id['imageTag'].startswith(version)])
```
demartinofra, line 1078: to make this more robust we need to probably check for startswith(version + "-") otherwise this check fails with versions such as 3.0.1 that are included in 3.0.10

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -1050,15 +1168,22 @@ Resources:
               - Effect: Allow
                 Action:
                   - ecr:BatchDeleteImage
+                  - ecr:ListImages
                 Resource: !GetAtt PrivateEcrRepository.Arn
+              - Effect: Allow
+                Action:
+                  - imagebuilder:ListImagePipelineImages
+                  - imagebuilder:DeleteImage
```
demartinofra, line 1176: can this be scoped down to not allow deletion of any image?

demartinofra (commented):
LGTM! dropped a couple of minor comments

charlesg3:
Comments addressed, moved to this PR: https://github.com/aws/aws-parallelcluster/pull/3107

demartinofra (approved):
Will merge this and address comments in separate PR

PR #3055, opened 2021-07-29
Drop support for configure AssignPublicIp for head node

chenwany:
Drop support for configure AssignPublicIp for head node because we will need to create an ENI for head node. ENI and AssignPublicIp can not be configured together.


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -528,6 +528,19 @@ class HeadNodeNetworkingSchema(BaseNetworkingSchema):
...
+        """Validate that elastic_ip and assign_public_ip do not co-exist."""
+        if kwargs.get("partial"):
+            # If the schema is to be loaded partially, do not check existence constrain.
+            return
+        elastic_ip = data.get("elastic_ip")
+        assign_public_ip = data.get("assign_public_ip")
+        if assign_public_ip is not None and elastic_ip is not None:
+            raise ValidationError("ElasticIp and AssignPublicIp can not be configured together.")
+        if assign_public_ip is False:
+            raise ValidationError("AssignPublicIp can not be False for headnode.")
```
enrico-usai, line 542: This is wrong. It can be False for example when the entire cluster is in a private subnet accessible through a bastion or Direct Connect.

lukeseawalker (changes requested):
ENI must not be dropped, because it's in charge to keep the private IP of the head node and will allow us to enable the feature to replace the head node

enrico-usai (dismissed):

enrico-usai:
Unit tests are failing.

enrico-usai (approved):

PR #3064, opened 2021-08-01
[API/log] Add unit testing for log controllers.

charlesg3:

*Issue #, if available:*

*Description of changes:*
 -  Align logs vs log_streams in controllers / model
 - Add api testing for new apis (list images / streams)
 - Align documentation on returned errors
 - Ensure errors in model are specific enough to propagate suitable errors through the api
   - Add tests to image_operations_controller and update imagebuilder model to ensure that we are returning convertible errors when images are shared or in-user
 - Track down and optimize imagebuilder tests that were going to AWS API and thus running slowly locally (25s faster running locally)
 - Fix tests warning about deprecated yaml loading without specifying a loader (i.e. just use yaml.safe_load(...) )
- Update handling of timestamps so that we use datetime objects with utc timezone internally and convert to/from string only at the periphery.


cli/src/pcluster/api/pcluster_api.py
```diff
@@ -30,7 +30,7 @@
 from pcluster.models.compute_fleet_status_manager import ComputeFleetStatus
```
demartinofra, line 30: Not part of this PR: we should probably get rid of this file since most if not all functionalities have been ported in the controllers classes.

cli/src/pcluster/models/cluster.py
```diff
@@ -993,24 +990,28 @@ def get_log_events(
...
                 limit=limit,
                 start_from_head=start_from_head,
                 next_token=next_token,
             )
+
             return LogStream(self.stack_name, log_stream_name, log_events_response)
         except AWSClientError as e:
-            raise _cluster_error_mapper(e, f"Unexpected error when retrieving log events: {e}")
+            if e.message.startswith("The specified log group"):
+                LOGGER.debug("Log Group %s doesn't exist.", self.stack.log_group_name)
```
demartinofra, line 1009: API/CLI logging level defaults to info. Would it make sense to write such log messages with info level? or we want to exclude them from the logs by default?

demartinofra (approved):
Flawless :)

PR #3065, opened 2021-08-02
Add validator to avoid different subnets in the queues

chenwany:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1087,16 +1087,29 @@ class SchedulingSchema(BaseSchema):
             ("awsbatch", "aws_batch_settings", "aws_batch_queues"),
             ("slurm", "slurm_settings", "slurm_queues"),
         ]:
-            # Verify the settings section is associated to the right storage type
+            # Verify the settings section is associated to the right scheduler type
             configured_scheduler = data.get("scheduler")
-            if data.get(settings, None) and scheduler != configured_scheduler:
+            if settings in data and scheduler != configured_scheduler:
                 raise ValidationError(
                     f"Scheduling > *Settings section is not appropriate to the Scheduler: {configured_scheduler}."
                 )
-            if data.get(queues, None) and scheduler != configured_scheduler:
+            if queues in data and scheduler != configured_scheduler:
                 raise ValidationError(
                     f"Scheduling > *Queues section is not appropriate to the Scheduler: {configured_scheduler}."
                 )
+
+    @validates_schema
+    def same_subnet_in_different_queues(self, data, **kwargs):
+        """Validate subnet_ids configured in different queues are the same."""
+        for queues in ["slurm_queues", "aws_batch_queues"]:
+            if queues in data:
+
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1098,6 +1098,19 @@ def right_scheduler_schema(self, data, **kwargs):
...
                 )
 
+    @validates_schema
+    def same_subnet_in_different_queues(self, data, **kwargs):
+        """Validate subnet_ids configured in different queues are the same."""
+        for queues in ["slurm_queues", "aws_batch_queues"]:
+            if data.get(queues, None):
+                queues_list = data.get(queues)
+                subnet_ids_list = []
+                for queue in queues_list:
```
charlesg3, line 1108: could use `for queue in data[queues]`: here and avoid the extra variable created in line 1106

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1098,6 +1098,19 @@ def right_scheduler_schema(self, data, **kwargs):
                     f"Scheduling > *Queues section is not appropriate to the Scheduler: {configured_scheduler}."
                 )
 
+    @validates_schema
+    def same_subnet_in_different_queues(self, data, **kwargs):
+        """Validate subnet_ids configured in different queues are the same."""
+        for queues in ["slurm_queues", "aws_batch_queues"]:
+            if data.get(queues, None):
```
charlesg3, line 1105: Feels more pythonic to use `if queue in data:` here

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1098,6 +1098,19 @@ def right_scheduler_schema(self, data, **kwargs):
...
+                queues_list = data.get(queues)
+                subnet_ids_list = []
```
charlesg3, line 1107: if you make `subnet_ids` a `set` you can check `if len(subnet_ids) > 1` in the condition below (`if not all ...`) which would be shorter / read cleaner.

charlesg3: Would also be possible to do this all with a set-comprehension...

```.py
    def queue_has_subnet_ids(queue)
        return queue.networking and queue.networking.subnet_ids
    subnet_ids = {q.networking.subnet_ids for q in data.get(queues) if queue_has_subnet_ids(q)}
    if len(subnet_ids) > 1:
        raise ValidationError("SubnetIds configured in different queues should be the same.")
```

cli/src/pcluster/schemas/cluster_schema.py
chenwany: Good idea! Changed

cli/src/pcluster/schemas/cluster_schema.py
chenwany: I was using `data.get(queues, None)` to be aligned with the other validation function in our code, now change other places to use `if queues in data` to looks more clear

charlesg3 (approved):

PR #3080, opened 2021-08-05
Update suggested action given by update-cluster to stop compute fleet

tilne:

### DIFF
--- a/cli/src/pcluster/config/update_policy.py
+++ b/cli/src/pcluster/config/update_policy.py
@@ -101,11 +101,11 @@ class UpdatePolicy:
 UpdatePolicy.FAIL_REASONS = {
     "ebs_volume_resize": "Updating the file system after a resize operation requires commands specific to your "
     "operating system.",
-    "shared_storage_change": "Shared Storage cannot be added or removed during a 'pcluster update' operation",
+    "shared_storage_change": "Shared Storage cannot be added or removed during a 'pcluster update-cluster' operation",
     "cookbook_update": lambda change, patch: (
         "Updating cookbook related parameter is not supported because it only "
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/config/update_policy.py
```diff
@@ -116,7 +116,8 @@ def __eq__(self, other):
         "modify-ebs-volume",
     ),
     "pcluster_stop": lambda change, patch: (
-        f"Stop the cluster with the following command:\npcluster stop {patch.cluster_name}"
+        "Stop the cluster with the following command:\npcluster update-compute-fleet --cluster-name  "
```
enrico-usai, line 119: I'd say stop the compute fleet. 

Then is STOP_REQUESTED valid for both Slurm and Batch? If it's not the case we can just say stop the compute fleet, without mentioning the command.

tilne: > is STOP_REQUESTED valid for both Slurm and Batch?

I assume so, based on the fact that `UpdatePolicy.AWSBATCH_CE_MAX_RESIZE` has uses the same reason.

enrico-usai: Unfortunately it is not true: https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/api/controllers/cluster_compute_fleet_controller.py#L66

tilne: so does it need to be DISABLED instead of STOP_REQUESTED for batch?

tilne: > without mentioning the command

i think not including  the command to run defeats the purpose of providing suggested actions

enrico-usai (approved):

enrico-usai (approved):

PR #3083, opened 2021-08-05
Improvements to integ-tests when running with high parallelism

demartinofra:
* Increase the number of retries and timeout when boto3 accesses the IMDS service to retrieve credentials. This fixes the "Unable to locate credentials" problem happening when running with a high parallelism on an EC2 instance.
* Add the process id to log entries to facilitate extracting logs for a single test process when running in parallel mode.
* Bumping the default boto3 retry to 10 for all boto3 service calls

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -163,6 +163,21 @@ def pytest_configure(config):
     _setup_custom_logger(config.getoption("tests_log_file"))
 
 
+def pytest_sessionstart(session):
+    # The number of seconds before a connection to the instance metadata service should time out.
+    # When attempting to retrieve credentials on an Amazon EC2 instance that is configured with an IAM role,
+    # a connection to the instance metadata service will time out after 1 second by default. If you know you're
+    # running on an EC2 instance with an IAM role configured, you can increase this value if needed.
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest.py
```diff
@@ -163,15 +163,32 @@ def pytest_configure(config):
     _setup_custom_logger(config.getoption("tests_log_file"))
 
 
+def pytest_sessionstart(session):
+    # The number of seconds before a connection to the instance metadata service should time out.
+    # When attempting to retrieve credentials on an Amazon EC2 instance that is configured with an IAM role,
+    # a connection to the instance metadata service will time out after 1 second by default. If you know you're
+    # running on an EC2 instance with an IAM role configured, you can increase this value if needed.
+    os.environ["AWS_METADATA_SERVICE_TIMEOUT"] = "5"
```
charlesg3, line 171: out of curiosity -- what is the default here?

charlesg3 (approved):

tests/integration-tests/conftest.py
demartinofra: 1 second. I left a note in the comment above:
>a connection to the instance metadata service will time out after 1 second by default.

charlesg3: Ah -- missed that while skimming. :+1:

PR #3090, opened 2021-08-06
[cli] Handle failure when waiting for cluster status transition

demartinofra:
Previously the unhandled error was resulting in:
```
{
  "message": "Unexpected fatal exception. Please look at the application logs for details on the encountered failure."
}
```

### DIFF
--- a/cli/src/pcluster/cli/middleware.py
+++ b/cli/src/pcluster/cli/middleware.py
@@ -19,12 +19,17 @@ function that is provided as the first argument and passing the **kwargs
 provided.
 """
 
+import logging
+
 import argparse
 import boto3
 import jmespath
+from botocore.exceptions import WaiterError
 
 import pcluster.cli.model
-from pcluster.cli.exceptions import ParameterException
+from pcluster.cli.exceptions import APIOperationException, ParameterException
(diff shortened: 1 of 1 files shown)


enrico-usai (approved):

cli/src/pcluster/cli/middleware.py
```diff
@@ -73,7 +78,11 @@ def update_cluster(func, _body, kwargs):
     if wait and not kwargs.get("dryrun"):
         cloud_formation = boto3.client("cloudformation")
         waiter = cloud_formation.get_waiter("stack_update_complete")
-        waiter.wait(StackName=kwargs["cluster_name"])
+        try:
+            waiter.wait(StackName=kwargs["cluster_name"])
+        except WaiterError as e:
+            LOGGER.error("Failed when waiting for cluster update with error: %s", e)
+            raise APIOperationException(_cluster_status(_body["clusterName"]))
```
charlesg3, line 85: it looks like this might need to be `_cluster_status(kwargs["cluster_name"])` because the cluster name is provided as a query parameter (which is also what is affecting the naming convention difference here)

PR #3092, opened 2021-08-06
2.11: Upgrade OSU benchmark to 5.7.1 and add other collective and pt2pt benchmarks

enrico-usai:
Upgrade OSU benchmark to 5.7.1 and add other collective and pt2pt benchmarks

OSU pt2pt benchmarks cannot be executed with more than 2 nodes.
We're running it in 2 instances with 1 proc per instance, defined by map-by parameter.

OSU collective benchmarks can be executed with any number of instances,
4 instances are enough to see performance differences.

For all the tests I collected benchmarks results for both Pcluster 2.11.0 and 2.11.1
and using as expected value one close to the 2.11.1 results.

Check avg latency for all packet sizes, with a tolerance of 10-20%.
Accept a max number of 3-4 failures on a total of 19-24 tests.
This helps to avoid failures for minor spikes.

Extended timeout for slurm jobs because OSU benchmarks are time expensive
(e.g. alltoall takes about 10 minutes).
Run a subset of benchmarks in efa-enabled-by-default and all of them in efa-enabled.

Done two separate commits to add benchmark results using pcluster 2.11.0 and pcluster 2.11.1.

In this patch I'm also updating `config.guess` and `config.sub` files to support ARM architecture. Arm tests were failing with:
```
checking build system type... ./config.guess: unable to guess system type
E
E           This script, last modified 2009-11-20, has failed to recognize
E           the operating system you are using. It is advised that you
E           download the most up to date version of the config scripts from
E
E             http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD
E           and
E             http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD
```

In the micro-benchmarks-5.6.3 archive the timestamp was 2013.


tests/integration-tests/tests/efa/test_efa/test_hit_efa/osu_pt2pt_submit_openmpi.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+BENCHMARK_NAME={{ benchmark_name }}
+OSU_BENCHMARK_VERSION={{ osu_benchmark_version }}
```
tilne, line 5: why not use an arg instead of templating these?

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -288,7 +288,7 @@ def __init__(self, remote_command_executor):
             for value in ["EndTime=Unknown", "JobState=RUNNING", "JobState=COMPLETING", "JobState=CONFIGURING"]
         ),
         wait_fixed=seconds(3),
-        stop_max_delay=minutes(7),
+        stop_max_delay=minutes(14),
```
tilne, line 291: a comment for how you got 14 minutes would be nice (e.g., twice the average that you saw when testing)

tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -180,47 +224,91 @@ def _test_osu_benchmarks_multiple_bandwidth(
...
     scheduler_commands.wait_job_completed(job_id)
     scheduler_commands.assert_job_succeeded(job_id)
 
-    output = remote_command_executor.run_remote_command("cat /shared/osu.out").stdout
+    output = remote_command_executor.run_remote_command(f"cat /shared/{benchmark_name}.out").stdout
     return output
 
 
+def _check_osu_benchmarks_results(test_datadir, mpi_version, benchmark_name, output):
+    packet_sizes = [4, 64, 256, 8192, 65536, 524288, 1048576]
```
tilne, line 274: it would be nice to have a comment explaining where these packet sizes come from

tests/integration-tests/tests/common/schedulers_common.py
enrico-usai: Changed to 10 minutes. the longest test seems to be alltoall that takes about 9 minutes.

tests/integration-tests/tests/efa/test_efa.py
enrico-usai: Changed approach, by comparing all the packet sizes and accepting a max numbers of 3 failures on 24 tests.

tests/integration-tests/tests/efa/test_efa/test_hit_efa/osu_pt2pt_submit_openmpi.sh
enrico-usai: Because this script is not executed by the framework but passed to Slurm to be submitted as is.

tilne (approved):

PR #3101, opened 2021-08-09
Fix custom cookbook S3 url

chenwany:
Fix custom cookbook s3 url not work 
The presign url not work previously because of:
lacking instance policy to presign s3 url 
aws presign ... --region should be bucket region instead of instance region

Add policy to presign url and get bucket region
Retrieve bucket region instead of using instance region
Remove not used parts when presign s3 url in the code to fix error: Warning: Illegal date format for -z, --time-cond (and not a file name).

Manually tests:
Before adding policy:
```
cat: /etc/chef/aws-parallelcluster-cookbook.tgz.date: No such file or directory
Warning: Illegal date format for -z, --time-cond (and not a file name). 
< HTTP/1.1 403 Forbidden
```
After adding IAM policy:
```
< HTTP/1.1 200 OK
```
Add a cookbook block in the manually test custom cookbook:
execute 'test custom cookbook' do
  command "echo 'This is use to test custom cookbook'"
end
I am able to see the execution result in chef-client.log:
```
execute] This is use to test custom cookbook
[2021-08-09T21:21:01+00:00] INFO: execute[test custom cookbook] ran successfully
    - execute echo 'This is use to test custom cookbook'
    - ```

### CODE REVIEW

File: cli/src/pcluster/resources/head_node/user_data.sh [hunk 1]
```diff
@@ -70,8 +70,6 @@ if [ "${!custom_cookbook}" != "NONE" ]; then
   else
     cookbook_url=${!custom_cookbook}
   fi
-else
```
yuleiwan, line 73: since we don't have runtime baking, we can remove this line in compute node user data too.

cli/src/pcluster/resources/head_node/user_data.sh
```diff
@@ -87,7 +85,7 @@ else
   error_exit "This AMI was not baked by ParallelCluster. Please use pcluster createami command to create an AMI by providing your AMI as parent image."
 fi
 if [ "${!custom_cookbook}" != "NONE" ]; then
-  curl --retry 3 -v -L -o /etc/chef/aws-parallelcluster-cookbook.tgz -z "$(cat /etc/chef/aws-parallelcluster-cookbook.tgz.date)" ${!cookbook_url}
+  curl --retry 3 -v -L -o /etc/chef/aws-parallelcluster-cookbook.tgz ${!cookbook_url}
```
yuleiwan, line 88: the presign url should be created with --region option which is same as cluster region, otherwise, we can't get the https 200

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -578,6 +583,32 @@ def _add_pcluster_policies_to_role(self, role_ref: str, name: str):
             roles=[role_ref],
         )
 
+    def _add_custom_cookbook_policies_to_role(self, role_ref: str, name: str):
```
yuleiwan, line 586: why we can't add multiple resources to parallelcluster policy rather than create a new policy?
```
if self._condition_custom_cookbook_with_s3_url():
            resources=[
                self.format_arn(
                    service="s3",
                    resource="{0}-aws-parallelcluster/*".format(self.region),
                    region="",
                    account="",
                ),
             self.format_arn(
                    service="s3",
                    resource="{0}/*".format(bucket_name),
                    region="",
                    account="",
                )
            ],
  else:
            resources=[
              self.format_arn(
                  service="s3",
                  resource="{0}-aws-parallelcluster/*".format(self.region),
                  region="",
                  account="",
              ),
              ],
  
```

cli/src/pcluster/resources/head_node/user_data.sh
chenwany: Done

cli/src/pcluster/resources/head_node/user_data.sh
chenwany: After our discussion, add bucket_region and additional policy to retrieve the bucket region

cli/src/pcluster/templates/cluster_stack.py
chenwany: I think we also create s3 policy for S3Access if there's S3Access. So I create new policy for customcook if there's s3 url to be aligned. Besides, if we add more policies in the parallelcluster policy in the futher, the if.. else will become longer and longer. So I think a new policy is more maintainable

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -578,6 +583,38 @@ def _add_pcluster_policies_to_role(self, role_ref: str, name: str):
...
+        bucket_name = re.match(r"s3://(.*?)/(.*)", self.config.dev_settings.cookbook.chef_cookbook).group(1)
```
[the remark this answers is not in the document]
chenwany: Done. After adding object key, the policy scope down
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::bucketforlogcwy/aws-parallelcluster-cookbook-3.0.0.tgz",
            "Effect": "Allow"
        },
        {
            "Action": "s3:GetBucketLocation",
            "Resource": "arn:aws:s3:::bucketforlogcwy",
            "Effect": "Allow"
        }
    ]
}
```

yuleiwan (approved):

PR #3103, opened 2021-08-09
Change EFA and Placement Group Defaults

tilne:
* Disable EFA by default, rather than a conditional default based on whether the instance type supports it.
* Require the use of a placement group when EFA is enabled.
* Warn user when instance type supports EFA but it is not enabled.

### DIFF
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -286,13 +286,18 @@ class EfaValidator(Validator):
 
     def _validate(self, instance_type, efa_enabled, gdr_support):
 
-        if efa_enabled:
-            if not AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported():
-                self._add_failure(
-                    f"Instance type '{instance_type}' does not support EFA.",
-                    FailureLevel.WARNING,
-                )
-        elif gdr_support:
+        instance_type_supports_efa = AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported()
+        if efa_enabled and not instance_type_supports_efa:
+            self._add_failure(
+                f"Instance type '{instance_type}' does not support EFA.",
+                FailureLevel.ERROR,
+            )
+        if instance_type_supports_efa and not efa_enabled:
+            self._add_failure(
+                f"Instance type '{instance_type}' supports EFA, but it is not enabled.",
+                FailureLevel.WARNING,
+            )
+        if gdr_support and not efa_enabled:
             self._add_failure(
                 "The EFA GDR Support can be used only if EFA is enabled.",
                 FailureLevel.ERROR,
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -302,8 +302,15 @@ class EfaValidator(Validator):
 class EfaPlacementGroupValidator(Validator):
     """Validate placement group if EFA is enabled."""
 
-    def _validate(self, efa_enabled, placement_group_id, placement_group_enabled):
-        if efa_enabled and not placement_group_id and not placement_group_enabled:
+    def _validate(self, efa_enabled, placement_group_enabled, placement_group_config_implicit):
+        if efa_enabled and placement_group_config_implicit:
+            self._add_failure(
+                "The placement group for EFA-enabled compute resources must be explicit. "
+                "You may see better performance using a placement group, but if you don't wish to use one please add "
+                "'Enabled: false' to the compute resource's configuration section.",
+                FailureLevel.ERROR,
+            )
+        elif efa_enabled and not placement_group_enabled:
             self._add_failure(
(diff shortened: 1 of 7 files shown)


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -286,13 +286,18 @@ class EfaValidator(Validator):
...
-                self._add_failure(
-                    f"Instance type '{instance_type}' does not support EFA.",
-                    FailureLevel.WARNING,
-                )
-        elif gdr_support:
+        instance_type_supports_efa = AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported()
+        if efa_enabled and not instance_type_supports_efa:
+            self._add_failure(
+                f"Instance type '{instance_type}' does not support EFA.",
+                FailureLevel.WARNING,
```
hanwen-pcluste, line 293: Should it be an `ERROR`?

hanwen-pcluste (commented):
Integration test need to be rolled back https://github.com/aws/aws-parallelcluster/pull/2689/files

cli/src/pcluster/validators/cluster_validators.py
tilne: yes it should

tilne: fixed

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -302,10 +307,10 @@ def _validate(self, instance_type, efa_enabled, gdr_support):
 class EfaPlacementGroupValidator(Validator):
     """Validate placement group if EFA is enabled."""
 
-    def _validate(self, efa_enabled, placement_group_id, placement_group_enabled):
-        if efa_enabled and not placement_group_id and not placement_group_enabled:
+    def _validate(self, efa_enabled, placement_group_enabled):
+        if efa_enabled and not placement_group_enabled:
             self._add_failure(
-                "You may see better performance using a placement group for the queue.", FailureLevel.WARNING
+                "Using a placement group is required for EFA-enabled compute resources.", FailureLevel.ERROR
```
demartinofra, line 313: here we don't want to require the usage of a placement group, but we want to require the PlacementGroup setting to be explicitly included in the config. Also we need to extend PlacementGroup accepted values in order to include not only dynamic and id but also a keyword to explicitly disable it (disabled?).

The message here should point out the fact that you need to explicitly disable placement groups with PlacementGroup: disabled if needed.

enrico-usai:
We also need to update all the tests configuration that are considering EFA is enabled by default (test_efa, test_nccl, ...)

cli/src/pcluster/validators/cluster_validators.py
tilne: PlacementGroup is not a single parameter; it's a section:

```yaml
PlacementGroup:
  Enabled: Boolean  # true
  Id: String
```

I see what you're saying about requiring an explicit value as opposed to requiring the use of a placement group, and how that would in turn affect the right error message to display to users. Instead of extending the valid values for the PlacementGroup section's Enabled field, why not do this by instead ensuring that PlacementGroup's Enabled field is not implicit when EFA is enabled? This would let us to continue defining the Enabled field as a Bool rather than a Str field where we need to convert to a bool.

demartinofra: yes totally. I had completely forgotten that now we have the Enabled boolean field.

tilne:
> We also need to update all the tests configuration that are considering EFA is enabled by default (test_efa, test_nccl, ...)

@enrico-usai @hanwen-pcluste is there a reason to keep the `efa-enabled-by-default` queue in the EFA integ test introduced by the [PR that Hanwen linked to](https://github.com/aws/aws-parallelcluster/pull/2689)? i've kept it for now because i wasn't sure if it was useful for the test to have another queue for whatever reason.

tests/integration-tests/tests/nccl/test_nccl/test_nccl/pcluster.config.yaml
```diff
@@ -20,7 +20,6 @@ Scheduling:
           MinCount: 2
           Efa:
             Enabled: true
-            GdrSupport: true
```
demartinofra, line 23: why removing GdrSupport?

hanwen-pcluste:
> @enrico-usai @hanwen-pcluste is there a reason to keep the efa-enabled-by-default queue in the EFA integ test introduced by the PR that Hanwen linked to? i've kept it for now because i wasn't sure if it was useful for the test to have another queue for whatever reason.

You can remove everything from that PR. `efa-enabled-by-default` was used to tested the efa enabled by default without specifying the `EFA` section. Now it is useless

tilne:
> You can remove everything from that PR. efa-enabled-by-default was used to tested the efa enabled by default without specifying the EFA section. Now it is useless

I saw that @enrico-usai only runs some of the tests added in #3095 on efa-enabled-by-default.

PR #3104, opened 2021-08-09
Allow same compute resource name in different queues

chenwany:

*Issue #, if available:*

*Description of changes:*


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1051,17 +1051,19 @@ def _test_disable_protected_mode(remote_command_executor, cluster, bucket_name,
...
         submit_command_args={"command": "sleep 3000", "nodes": 1, "partition": "half-broken", "constraint": "c5.xlarge"}
     )
     # Wait for the job to run
     scheduler_commands.wait_job_running(cancel_job_id)
+
+    # Re-enable protected mode
+    _enable_protected_mode(remote_command_executor)
+    # Decrease protected failure count for quicker enter protected mode.
+    _set_protected_failure_count(remote_command_executor, 2)
+
```
demartinofra, line 1066: out of curiosity why you moved this?

Also can we modify an existing integ test config to use same compute resource name in different queues?

chenwany: The reason is that here we are testing if there's job running, we will not disable the queue. In the test we have static compute fleets that are problematic. Before moving the order of the codes, it is possible that static nodes failed before the job submission/job begin to run, the queue will be put in protected mode and we may not able to test it. After changing the order, the job will begin to rum before bootstrap failure happen, so we are able to test `queue will not be disable if jobs are running`

Added

demartinofra (approved):

PR #3107, opened 2021-08-10
[API/infrastructure] Add support for updating the API

charlesg3:

*Issue #, if available:*

*Description of changes:*
- Add support for updating the API via imagebuilder


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -1050,15 +1168,23 @@ Resources:
...
                 Action:
                   - ecr:BatchDeleteImage
+                  - ecr:ListImages
                 Resource: !GetAtt PrivateEcrRepository.Arn
+              - Effect: Allow
+                Action:
+                  - imagebuilder:ListImagePipelineImages
+                  - imagebuilder:DeleteImage
+                Resource:
+                  - !Sub 'arn:${AWS::Partition}:imagebuilder:${Region}:${AWS::AccountId}:image/parallelclusterimage-*'
```
demartinofra, line 1178: I believe there are a couple of issues here:
1. ListImagePipelineImages needs to have a resource of type [imagePipeline](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2imagebuilder.html#amazonec2imagebuilder-imagePipeline)
2. The resource used for DeleteImage I don't think matches the arn of the images the pipeline generates. The prefix needs to match the images generated when the image builder pipeline runs. From a quick test they currently look something like `arn:aws:imagebuilder:eu-west-1:653949452088:image/importpublicecrimage-3-0-0-adb6e170-f5ed-11eb-95c3-069f44f40d5b/3.0.0/2`. Either we modify the prefix in the Name property of EcrImageRecipe or we change the resource here

See docs: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2imagebuilder.html

PR #3124, opened 2021-08-12
3.0: HeadNode Replacement

yuleiwan:

*Issue #, if available:*

*Description of changes:*


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -501,6 +501,37 @@ Resources:
...
+            Sid: AutoScalingList
+          - Action:
+              - autoscaling:CreateAutoScalingGroup
+              - autoscaling:DeleteAutoScalingGroup
+              - autoscaling:UpdateAutoScalingGroup
+              - autoscaling:SuspendProcesses
+              - autoscaling:TerminateInstanceInAutoScalingGroup
+              - autoscaling:ResumeProcesses
+              - autoscaling:SetDesiredCapacity
+            Resource: '*'
```
demartinofra, line 526: Can we restrict this resource and the above one to only work with ASGs that are created as part of ParallelCluster stacks? a ParallelCluster prefix in the ASG name can be a way to filter them

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -501,6 +501,37 @@ Resources:
...
+              - autoscaling:DescribeAutoScalingInstances
+              - autoscaling:DescribeScheduledActions
+            Resource: '*'
+            Effect: Allow
+            Condition: !If
+              - IsMultiRegion
+              - !Ref AWS::NoValue
+              - StringEquals:
+                  aws:RequestedRegion:
+                    - !Ref Region
```
demartinofra, line 516: if we add an arn in the Resource we can get rid of this and use the region or * in the arn

cli/requirements.txt
```diff
@@ -19,6 +19,8 @@ aws-cdk.aws-logs~=1.116
 aws-cdk.aws-route53~=1.116
 aws-cdk.aws-ssm~=1.116
 aws-cdk.aws-sqs~=1.116
+aws-cdk.aws-autoscaling~=1.116
+aws-cdk.aws-cloudformation~=1.116
```
demartinofra, line 23: do we need to update deps in setup.py too?

tests/integration-tests/conftest.py
```diff
@@ -1192,3 +1192,34 @@ def _pcluster_ami_without_standard_naming(version):
...
+
+    This AMI is used to test head node and compute node AMI update
+    """
+    copy_ami_id = None
+    client = boto3.client("ec2", region_name=region)
+
+    def _copy_image(image_id, test_name):
+        nonlocal copy_ami_id
+        copy_ami_id = client.copy_image(
+            Name=f"copyed_image_{test_name}", SourceImageId=image_id, SourceRegion=region
```
demartinofra, line 1210: in case the image is not cleaned-up for a previous run or in case tests run in parallel with the same name (for any reason) this would fail.

one option is to add a random string in the name. is there a way to leverage the AMI cleanup logic in Jenkins in case the clean-up is not executed bu tests?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1025,16 +1039,16 @@ def _add_head_node(self):
...
-                    "deployConfigFiles",
                     "cfnHupConfig",
                     "chefPrepEnv",
                     "shellRunPreInstall",
                     "chefConfig",
                     "shellRunPostInstall",
                     "chefFinalize",
                 ],
-                "update": ["deployConfigFiles", "chefUpdate"],
+                "update": ["deployConfigFiles", "chefUpdate", "sendSignal"],
```
demartinofra, line 1051: what if chefUpdate fails?in that case we need to signal a failure.

cli/src/pcluster/resources/head_node/user_data.sh
```diff
@@ -61,6 +61,12 @@ function vendor_cookbook
   export HOME="${!HOME_BAK}"
 }
 [ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh
+
+# deploy config files
+export PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin
```
tilne, line 66: minor: why set an absolute path rather than appending on to the existing path?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -100,23 +100,23 @@ class HeadNodeRootVolumeSchema(BaseSchema):
         validate=get_field_validator("volume_type"),
         metadata={
             "update_policy": UpdatePolicy(
-                UpdatePolicy.UNSUPPORTED, action_needed=UpdatePolicy.ACTIONS_NEEDED["ebs_volume_update"]
```
tilne, line 103: based on offline conversations from earlier today i think we still want to keep all of these as unsupported

cli/src/pcluster/resources/head_node/user_data.sh
yuleiwan: This is from legacy code, I think because we don't know what does existing path include and the path order in existing path may affect our execution, it stops at the first one it finds.

cli/src/pcluster/schemas/cluster_schema.py
yuleiwan: correct, done

cli/src/pcluster/templates/cluster_stack.py
yuleiwan: good catch, added failure siganl, but in worst case, it will be timeout in 30 minutes.

cli/requirements.txt
yuleiwan: done

tests/integration-tests/conftest.py
yuleiwan: random string is added in test_name. For AMI cleanup pipeline, we only clean up ami with name matching to *aws-parallelcluster* regex.

api/infrastructure/parallelcluster-api.yaml
yuleiwan: we don't have parallelcluster prefix in ASG name now, for example, 
`arn:aws:autoscaling:eu-west-3:109993544819:autoScalingGroup:ae217271-ec3f-495f-b259-91db02ea9bec:autoScalingGroupName/integ-tests-f9c8ssa4wvv8udci-p3-HeadNode-ZH9RRWSB8X1P`
it is composed of {cluster_name}-HeadNode-{random_string}, shall I filter it with `arn:${AWS::Partition}:autoscaling:${Region}:${AWS::AccountId}:autoScalingGroup:*/*-HeadNode-*` or rename the ASG name with parallelcluster prefix?

tests/integration-tests/conftest.py
tilne: > we only clean up ami with name matching to aws-parallelcluster regex

is there a reason we can't use that name format here?

api/infrastructure/parallelcluster-api.yaml
tilne: my personal preference is to rename the ASG with a parallelcluster prefix, since that would seem to better ensure that the policy can only be used for pcluster-specific resources.

yuleiwan: done

api/infrastructure/parallelcluster-api.yaml
yuleiwan: done

tests/integration-tests/conftest.py
yuleiwan: add tag for copied ami to leverage cleanup_ami pipeline

tilne (approved):

cli/src/pcluster/resources/head_node/user_data.sh
```diff
@@ -61,6 +61,12 @@ function vendor_cookbook
...
+cfn-init -s ${AWS::StackName} -v -c deployFiles -r HeadNodeLaunchTemplate --region ${AWS::Region}
```
lukeseawalker, line 67: what if this fails? Are we able to signal an error?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1229,21 +1268,14 @@ def _add_outputs(self):
 
         CfnOutput(
             self,
-            "HeadNodeInstanceID",
-            description="ID of the head node instance",
-            value=self.head_node_instance.ref,
+            "HeadNodeAutoScalingGroupID",
```
lukeseawalker, line 1271: Where was the "HeadNodeInstanceID" used? Are we sure that can be replaced by "HeadNodeAutoScalingGroupID"?

cli/src/pcluster/resources/head_node/user_data.sh
lukeseawalker: later downstream, a `cd /tmp` is done before calling `cfn-init`. Is this a requirement that needs to be done also here?

api/infrastructure/parallelcluster-api.yaml
demartinofra: minor: to shorten the size of the policy can we use resource here as well with something like: `!Sub arn:${AWS::Partition}:autoscaling:${Region}:${AWS::AccountId}:*`>

cli/src/pcluster/templates/cluster_stack.py
yuleiwan: we can't get instance id from ASG since we use launch template, I don't find this cfn output is used in other place.

cli/src/pcluster/resources/head_node/user_data.sh
yuleiwan: Move `/tmp` at the begining, we can't send the signal in deployFiles because we can't get wait condition ref, but if this command fails, creation will fail in 30 mins due to timeout of wait condition.

api/infrastructure/parallelcluster-api.yaml
yuleiwan: done

lukeseawalker (approved):

tilne (approved):

PR #3139, opened 2021-08-17
Prevent installation of Intel packages for Batch clusters

tilne:

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1207,6 +1207,21 @@ class ClusterSchema(BaseSchema):
         """Validate tags."""
         validate_no_reserved_tag(tags)
 
+    @validates_schema
+    def no_intel_select_solutions_for_batch(self, data, **kwargs):
+        """Ensure IntelSelectSolutions section is not included when AWS Batch is the scheduler."""
+        scheduling = data.get("scheduling")
+        additional_packages = data.get("additional_packages")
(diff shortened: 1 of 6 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1207,6 +1207,21 @@ def validate_tags(self, tags):
...
         validate_no_reserved_tag(tags)
 
+    @validates_schema
+    def no_intel_select_solutions_for_batch(self, data, **kwargs):
+        """Ensure IntelSelectSolutions section is not included when AWS Batch is the scheduler."""
+        scheduling = data.get("scheduling")
+        additional_packages = data.get("additional_packages")
+        if (
+            scheduling
+            and scheduling.scheduler == "awsbatch"
```
enrico-usai, line 1217: I cannot remember if this validation is executed after the Scheduling and AdditionalPackagest schema validation. I'm asking this to be sure scheduler exists.

You can simply try removing the scheduler keyword from the Scheduling section.

Then I'd add an unit test for it.

tilne: without the first check of `scheduling` (to make sure it's not `None`) the existing unit tests fail because that parameter isn't set. i think that's what you're asking?

tilne: unit tests added

enrico-usai (approved):

PR #3167, opened 2021-08-20
2.11: Avoid conflict between os module and fixture names

enrico-usai:
The check was failing with:
```
if instance == "p4d.24xlarge" and "centos" not in os:
argument of type 'module' is not iterable
```


mauri-melato (dismissed):
lgtm

mauri-melato (approved):

PR #3171, opened 2021-08-23
Fix remaining update integ tests

tilne:
In f02f0902 some integ tests were updated to account for the fact that
cluster updates done via the Cluster class were no longer forced by
default. This commit does more of the same. It also updates other tests
that weren't failing but were using a now-ineffective kwarg to attempt
to force an update.


hanwen-pcluste (approved):
LGTM

PR #3183, opened 2021-08-25
2.11: Remove placement group from EFA test when using p4d

enrico-usai:
We have to avoid failures for OSU benchmark tests because it is tuned for Placement Group usage.


mauri-melato (dismissed):

enrico-usai:
Closed in favour of: https://github.com/aws/aws-parallelcluster/pull/3180/files

PR #3228, opened 2021-09-01
Support Region parameter in cluster and image configuration file

hanwen-pcluste:

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -907,12 +907,19 @@ class BaseClusterConfig(Resource):
         additional_packages: AdditionalPackages = None,
         tags: List[Tag] = None,
         iam: ClusterIam = None,
+        config_region: str = None,
         custom_s3_bucket: str = None,
         additional_resources: str = None,
         dev_settings: ClusterDevSettings = None,
     ):
         super().__init__()
         self.__region = None
+        # config_region represents the region parameter in the configuration file
+        # and is only used by configure_aws_region_from_config in controllers.
+        # Since the region is already set by configure_aws_region_from_config to the environment variable,
+        # the self.config_region is never used. It has to be here to make sure cluster_config stores all information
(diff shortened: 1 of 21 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -967,6 +968,7 @@ def __init__(
         self.additional_packages = additional_packages
         self.tags = tags
         self.iam = iam
+        self.config_region = config_region
```
enrico-usai, line 971: I think we can avoid this extra attribute and just set the `self.__region` in row 962

cli/src/pcluster/config/cluster_config.py
```diff
@@ -953,6 +953,7 @@ def __init__(
         additional_packages: AdditionalPackages = None,
         tags: List[Tag] = None,
         iam: ClusterIam = None,
+        config_region: str = None,
```
enrico-usai, line 956: I'd call it just `region`

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: I removed this line since the region of config has been parsed and used in controller layer. And `config_region` is left as an unused argument.

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Calling it `region` breaks unit tests because the output of `cluster_schema.dump(cluster)` in `test_cluster_schema` will contains region information getting from the `property` `region` in `cluster_config`.
We can talk if this reply is confusing

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Even removing this line fails the unit tests in a tricky way. I have to restore the code before your review. We can talk more

enrico-usai (approved):

PR #3245, opened 2021-09-03
Check bucket existence before using it in exception messages

enrico-usai:
In the modified methods we were accessing the bucket.name when raising the exception
but if the failure causing the exception is due to the bucket absence the raising will fail too.

With this patch we're adding a check for bucket existence out of the global try/catch
using the bucket name in the message.

We can access to the bucket name only in the S3Bucket class but not in the
cluster and imagebuilder module because we cannot be sure the bucket exists.


cli/src/pcluster/models/cluster.py
```diff
@@ -230,6 +231,12 @@ def _get_cluster_config(self):
                 e, f"Unable to load configuration from bucket '{self.bucket.name}/{self.s3_artifacts_dir}'.\n{e}"
             )
 
+    def _check_bucket_existence(self):
+        try:
+            return self.bucket
```
yuleiwan, line 236: bucket property is cached, and bucket exists check is executed at the first call of this property in the init_s3_bucket, is it necessary to check cached property in all methods?

yuleiwan (approved):

demartinofra (approved):

PR #3283, opened 2021-09-20
3.0 Remove Proxy and PlacementGroup from Awsbatch schema and config

yuleiwan:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -395,13 +395,10 @@ class Proxy(Resource):
 class _BaseNetworking(Resource):
     """Represent the networking configuration shared by head node and compute node."""
 
-    def __init__(
-        self, security_groups: List[str] = None, additional_security_groups: List[str] = None, proxy: Proxy = None
-    ):
+    def __init__(self, security_groups: List[str] = None, additional_security_groups: List[str] = None):
         super().__init__()
         self.security_groups = Resource.init_param(security_groups)
         self.additional_security_groups = Resource.init_param(additional_security_groups)
-        self.proxy = proxy
 
     def _register_validators(self):
         self._register_validator(SecurityGroupsValidator, security_group_ids=self.security_groups)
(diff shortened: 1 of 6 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -438,16 +436,29 @@ def _register_validators(self):
...
+class SlurmQueueNetworking(_QueueNetworking):
+    """Represent the networking configuration for the slurm Queue."""
+
+    def __init__(self, subnet_ids: List[str], placement_group: PlacementGroup = None, proxy: Proxy = None, **kwargs):
+        super().__init__(subnet_ids, **kwargs)
         self.placement_group = placement_group
+        self.proxy = proxy
+
+
+class AwsBatchQueueNetworking(_QueueNetworking):
```
chenwany, line 457: Why do we need AwsBatchQueueNetworking? Is it the same as _QueueNetworking?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -883,10 +894,9 @@ def __init__(self, local_storage: LocalStorage = None, **kwargs):
 class BaseQueue(Resource):
     """Represent the generic Queue resource."""
 
-    def __init__(self, name: str, networking: QueueNetworking, capacity_type: str = None):
+    def __init__(self, name: str, capacity_type: str = None):
```
chenwany, line 897: def get_common_user_data_env(node: Union[HeadNode, BaseQueue], config: BaseClusterConfig) -> dict:
in cdk_builder_utils.py also need to be changed accordingly.

cli/src/pcluster/config/cluster_config.py
yuleiwan: Although QueueNetworking has same properties as AwsBatchQueueNetworking, AwsBatchQueueNetworking and SlurmQueueNetworking Inherit from QueueNetworking, and we should not expose QueueNetworking to avoid confusion.

cli/src/pcluster/config/cluster_config.py
yuleiwan: good catch, updated

cli/src/pcluster/config/cluster_config.py
chenwany: Is there some thing we can do to prevent to directly create instance from  _QueueNetworking class or use it? For example, define it as abstract class?

chenwany (approved):
Great job!

cli/src/pcluster/config/cluster_config.py
yuleiwan: I don't think we can set it as abstract class, otherwise _register_validators can't be used in parent class BaseNetworking. Underscore prefix helps us to distinguish it with other exposed classes.

PR #3290, opened 2021-09-22
[WIP]Add conversion for SharedStorage, Region, Os, Image, AdditionalP…

chenwany:
…ackage and Tags

*Issue #, if available:*

*Description of changes:*


cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        required=True,
+    )
+    convert_parser.set_defaults(func=convert)
+
+    return parser.parse_args(argv)
+
+
+def convert(args=None):
+    try:
+        pcluster3_config_converter = Pcluster3ConfigConverter(args.config_file, args.cluster_template, args.output_file)
```
charlesg3, line 574: `pcluster3_config_convert` feels verbose here -- could use `converter` for a local variable (since there are no other types of converters possible)

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+                tags = json.loads(tags)
+            except Exception as e:
+                self.error(f"Tags should be in the format of dictionary, {e}")
+            self.pcluster3_configuration["Tags"] = []
+            for key, value in tags.items():
+                self.pcluster3_configuration[section].append({"Key": key, "Value": value})
+
+    def convert_shared_storage(self, section):
+        shared_storage = []
+        function_section_mappings = {
```
charlesg3, line 141: Feels more intuitive to map the string to the function here, even though they are just looped over as pairs below

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        ]:
+            self.convert_single_field(*item)
+        for item in additional_items:
+            pc2_label, pc3_label = item[0], item[1]
+            conversion_type = item[2] if len(item) == 3 else None
+            self.convert_single_field(section_label, pc2_label, settings, pc3_label, conversion_type)
+        return storage_section, settings, section_label
+
+    def convert_ebs_settings(self, section):
+        ebs_settings = self.config_parser.get(self.cluster_section_name, "ebs_settings", fallback=None)
```
charlesg3, line 173: This line is repeated quite a few times through this file -- perhaps it's better to make a helper function so that you can simply call `self.config_get("ebs_settings")` and not repeat the same parameters each time (at least where the parameters are the same)

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+            self.convert_single_field(section_label, "retention_days", cloudwatch, "RetentionInDays", "getint")
+
+        dashboards = {}
+        dasnboard_cloudwatch = {}
+        dashboard_settings = self.config_parser.get(self.cluster_section_name, "dashboard_settings", fallback=None)
+        if dashboard_settings:
+            section_label = f"dashboard {dashboard_settings.strip()}"
+            self.convert_single_field(section_label, "enable", dasnboard_cloudwatch, "Enabled", "getboolean")
+
+        if cloudwatch:
```
charlesg3, line 301: It seems like this particular logic is repeated many times in the conversion. I would suggest adding a simple helper like:
```
def add_if(self, section, section_name, value):
   if value:
       section[section_name] = value
```

Then the following can become:
```
    self.add_if(monitoring, "Logs", logs)
    self.add_if(dashboards, "CloudWatch", dashboard_cloudwatch)
    self.add_if(monitoring, "Dashboards", dashboards)
    self.add_if(self.pcluster3_configuration, section_name, monitoring)
```

Which seems easier to read because there's less repeated code and less if then, if then, if then...

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        self.convert_single_field(self.cluster_section_name, "post_install", on_node_configured, "Script")
+        pre_install_args_list = []
+        post_install_args_list = []
+        pre_install_args = self.config_parser.get(self.cluster_section_name, "pre_install_args", fallback=None)
+        if pre_install_args:
+            pre_install_args_list.append(pre_install_args)
+        post_install_args = self.config_parser.get(self.cluster_section_name, "post_install_args", fallback=None)
+        if post_install_args:
+            post_install_args_list.append(post_install_args)
+        if pre_install_args_list:
```
charlesg3, line 475: more `add_if` opportunity here

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        s3_read_resource = self.config_parser.get(self.cluster_section_name, "s3_read_resource", fallback=None)
+        s3_read_write_resource = self.config_parser.get(
+            self.cluster_section_name, "s3_read_write_resource", fallback=None
+        )
+        if s3_read_resource:
+            bucket_name, key = get_bucket_name_and_key(s3_read_resource)
+            s3_access.append({"BucketName": bucket_name, "KeyName": key})
+        if s3_read_write_resource:
+            s3_access.append({"EnableWriteAccess": True, "BucketName": bucket_name, "KeyName": key})
+        if s3_access:
```
charlesg3, line 501: more `add_if` here

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        if vpc_settings:
+            self.vpc_section_label = f"vpc {vpc_settings.strip()}"
+
+    def convert_local_storage(self, headnode):
+        local_storage = {}
+        root_volume = {}
+        self.convert_single_field(self.cluster_section_name, "master_root_volume_size", root_volume, "Size", "getint")
+        ephemeral_volume = {}
+        self.convert_single_field(self.cluster_section_name, "ephemeral_dir", ephemeral_volume, "EphemeralVolume")
+        if root_volume:
```
charlesg3, line 443: more `add_if` opportunity here

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+            )
+            if vpc_security_group_id:
+                security_groups.append(vpc_security_group_id)
+
+            additional_sg = self.config_parser.get(self.vpc_section_label, "additional_sg", fallback=None)
+            if additional_sg:
+                additional_security_groups.append(additional_sg)
+        proxy = {}
+        self.convert_single_field(self.cluster_section_name, "proxy_server", proxy, "HttpProxyAddress")
+        if proxy:
```
charlesg3, line 404: more `add_if` opportunity here

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        self.warning_messages = []
+        self.error_messages = []
+        self.__sections = OrderedDict({})
+        self.model = dict()
+        self.pcluster3_configuration = dict()
+        self.cluster_section_name = None
+        self.vpc_section_label = None
+        self.config_parser = configparser.ConfigParser(inline_comment_prefixes=("#", ";"))
+
+    def init_config_parser(self):
```
charlesg3, line 68: Is there any reason that we can't just call init in the constructor?

charlesg3 (commented):
- I think we can call the file `configuration_converter` to be less verbose.
- I would love to see a testing system where we have lots of small test file snippets and we're just running the full converter on that file (perhaps turning off partial validation) rather than the unit testing that is expanding here... It's much easier to convince ourselves that it's working if we're doing a pure end-to-end test on many many different inputs rather than ensuring that all of the testing code is correct -- testing it purely from the user's point of view with verified outputs.

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+from collections import OrderedDict
+
+import argparse
+import configparser
+import yaml
+
+from pcluster.utils import get_partition
+
+
+class MyDumper(yaml.Dumper):
```
charlesg3, line 24: Let's call this "ConfigDumper" to be more specific

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        return super(MyDumper, self).increase_indent(flow, False)
+
+
+def _role_name_to_arn(role_name):
+    return "arn:{0}:iam::aws:role/{1}".format(get_partition(), role_name)
+
+
+def get_bucket_name_and_key(resource_arn):
+    match = re.match(r"arn:(.*?):s3:::([^\/\*]*)((\/)?(.*))?", resource_arn)
+    bucket_name = match.groups()[1]
```
charlesg3, line 35: You could combine these into:
```
    bucket_name, key_name = match.groups()[1], match.groups()[4]
```
Since they're pulling out of the same list.

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+    key_name = match.groups()[4]
+    return bucket_name, key_name
+
+
+# arn:aws:s3:::my_corporate_bucket/Development/*
```
charlesg3, line 40: Do these comments belong in the above function?

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+                if ebs_settings:
+                    ebs_section[section] = ebs_settings
+                if ebs_section:
+                    raid_list.append(ebs_section)
+            return raid_list
+        return None
+
+    def convert_efs_settings(self, section):
+        efs_settings = self.config_parser.get(self.cluster_section_name, "efs_settings", fallback=None)
+        if efs_settings:
```
charlesg3, line 227: You can return early here.

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+                if efs_dict:
+                    efs_section[section] = efs_dict
+                if efs_section:
+                    efs_list.append(efs_section)
+            return efs_list
+        return None
+
+    def convert_fsx_settings(self, section):
+        fsx_settings = self.config_parser.get(self.cluster_section_name, "fsx_settings", fallback=None)
+        if fsx_settings:
```
charlesg3, line 248: You can return early here.

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+    def convert_scheduling(self, section):
+        scheduling = {}
+        self.convert_single_field(self.cluster_section_name, "scheduler", scheduling, "Scheduler")
+        scheduler = self.config_parser.get(self.cluster_section_name, "scheduler", fallback=None)
+        if scheduler == "slurm":
+            self.convert_slurm_queue()
+        elif scheduler == "awsbatch":
+            pass
+
+        if scheduling:
```
charlesg3, line 519: `add_if` opportunity here

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,593 @@
...
+        self.convert_single_field(
+            self.cluster_section_name, "enable_intel_hpc_platform", intel_software, "IntelHpcPlatform", "getboolean"
+        )
+        if intel_software:
+            additional_packages["IntelSoftware"] = intel_software
+        if additional_packages:
+            self.pcluster3_configuration[section] = additional_packages
+
+    def convert_tags(self, section):
+        tags = self.config_parser.get(self.cluster_section_name, "tags", fallback=None)
```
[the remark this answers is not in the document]
chenwany: Good idea! Changed

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
chenwany: The reason we write it as {function: "String"} instead of {"String": function} is that "EbsSettings" map both self.convert_ebs_settings and self.convert_raid_settings

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
chenwany: Added a helper function: 
def cluster_config_get(self, param):
        return self.config_parser.get(self.cluster_section_name, param, fallback=None)

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
chenwany: Good idea! Changed

cli/src/pcluster/cli/commands/pcluster3_configuration_converter.py
chenwany: Changed

cli/tests/pcluster/cli/test_pcluster3_configuration_converter.py
```diff
@@ -0,0 +1,230 @@
...
+
+from pcluster.cli.commands.pcluster3_configuration_converter import Pcluster3ConfigConverter
+
+
+@pytest.mark.parametrize("expected_input, expected_output", [("slurm_full.ini", "slurm_full.yaml")])
+def test_pcluster3_configuration_converter(test_datadir, tmpdir, expected_input, expected_output):
+    pcluster3_config_converter = Pcluster3ConfigConverter(
+        test_datadir / expected_input, "default", tmpdir / "output_yaml"
+    )
+    pcluster3_config_converter.init_config_parser()
```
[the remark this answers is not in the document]
chenwany: Changed

cli/src/pcluster/cli/commands/configuration_converter.py
```diff
@@ -0,0 +1,608 @@
...
+from pcluster.utils import get_partition
+
+
+class ConfigDumper(yaml.Dumper):
+    def increase_indent(self, flow=False, indentless=False):
+        return super(ConfigDumper, self).increase_indent(flow, False)
+
+
+def _role_name_to_arn(role_name):
+    return "arn:{0}:iam::aws:role/{1}".format(get_partition(), role_name)
```
demartinofra, line 30: is this correct?or aws needs to be replaced by the account id?

chenwany: You are right, aws needs to be replaced by the account id. Fixed

cli/setup.py
```diff
@@ -77,6 +77,7 @@ def readme():
     entry_points={
         "console_scripts": [
             "pcluster = pcluster.cli.entrypoint:main",
+            "pcluster3-config-converter = pcluster.cli.commands.configuration_converter:main",
```
demartinofra, line 80: if possible I'd move `pcluster3-config-converter` outside of the pcluster dir to something like cli/src/pcluster3_config_converter and avoid any dependency with the modules defined in pcluster package if possible

chenwany: Moved it outside

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ CHANGELOG
 - The region parameter is now retrieved from the provider chain, thus supporting the use of profiles and defaults specified
 in the `~/.aws/config` file.
 - Add support for `UseEc2Hostnames` in the cluster configuration file. When set to `true`, use EC2 default hostnames (e.g. ip-1-2-3-4) for compute nodes.
+- Add configuration converter to convert pcluster2 config to pcluster3 condfig.
```
charlesg3, line 11: `condfig` typo

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+    print("Note: {0}".format(message))
+
+
+def _get_partition(region):
+    """Get partition from the given region."""
+    return next(("aws-" + partition for partition in ["us-gov", "cn"] if region.startswith(partition)), "aws")
+
+
+def _add_if(section, section_name, value):
+    """Add value to the dictionary only if the value exist."""
```
charlesg3, line 60: "exists." in the comment.

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+            except OSError as e:
+                if e.errno != errno.EEXIST:
+                    raise
+
+            # Fix permissions
+            with open(self.output_file, "a", encoding="utf-8"):
+                os.chmod(self.output_file, stat.S_IRUSR | stat.S_IWUSR)
+
+        # Write configuration to disk
+        if self.output_file:
```
charlesg3, line 403: Curious what this guard is / why it's here? If we don't have an output file it seems like there's some other error that should be caught elsewhere / earlier (?)

Otherwise, it seems like it's worth returning earlier in this function when it isn't specified (e.g. basically as a pre-condition)

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+                    default_flow_style=False,
+                )
+
+    def validate_cluster_section_name(self):
+        """
+        Validate cluster section name.
+
+        If cluster_template is specified in the command, use specifeid cluster_template to find the cluster section.
+        If cluster_label is not specified, use cluster template specified in global section.
+        IF no cluster_label sepcified in the cluster section, verify if default cluster section exist.
```
charlesg3, line 420: `IF` should be `If`

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+            self.cluster_section_name = "cluster default"
+        else:
+            _error("Can not find a valid cluster section.")
+        if self.cluster_section_name not in self.config_parser.sections():
+            _error("The specified cluster section is not in the configuration.")
+
+    def validate_required_fields(self):
+        """Validate required fields for conversion."""
+        self.comments += (
+            "# Volume encrypted is default to True in ParallelCluster3 while it is default to False in "
```
charlesg3, line 437: `is default to` can be `defaults to`

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+            "ParallelCluster2.\n"
+        )
+        _note(
+            "Volume encrypted is default to True in ParallelCluster3 while it is default to False in "
+            "ParallelCluster2."
+        )
+        if self.cluster_config_get("scheduler") == "slurm":
+            self.comments += (
+                "# Imds/Secured is default to True in ParallelCluster3 for Slurm while it is default to "
```
charlesg3, line 446: I think it might be more common to have nested structures represented with dot-notation, so here `Imds.Secured` rather than `Imds/Secured` (esp considering something like CSS selectors or JmesPath)

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+        if placement == "DYNAMIC":
+            placement_group["Enabled"] = True
+        elif placement:
+            placement_group["Id"] = placement
+        _add_if(networking, "PlacementGroup", placement_group)
+
+        proxy = dict()
+        self.convert_single_field(self.cluster_section_name, "proxy_server", proxy, "HttpProxyAddress")
+        _add_if(networking, "Proxy", proxy)
+
```
charlesg3, line 510: extra newline?

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+        self.convert_single_field(self.vpc_section_label, "ssh_from", ssh, "AllowedIps")
+        _add_if(headnode, "Ssh", ssh)
+
+    def validate_vpc_settings(self):
+        """Validate vpc settings exists in pcluster2 config."""
+        vpc_settings = self.cluster_config_get("vpc_settings")
+        if vpc_settings:
+            self.vpc_section_label = f"vpc {vpc_settings.strip()}"
+        else:
+            _error("Missing vpc_settings in the configuration file")
```
charlesg3, line 570: Missing period `.` at end of sentence.

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+                "The provided scheduler is deprecated in ParallelCluster3, please check [here]"
+                "(https://docs.aws.amazon.com/parallelcluster/latest/ug/schedulers-v3.html) for supported schedulers."
+            )
+        elif scheduler not in ["slurm", "awsbatch"]:
+            _error(f"Wrong value for scheduler: {scheduler}.")
+
+    def validate_single_field(self, section, field, message):
+        """Validate whether a filed in pcluster2 is supported in pcluster3."""
+        value = self.config_parser.get(section, field, fallback=None)
+        if value:
```
charlesg3, line 792: Seems like you could return early here.

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+            if message == "ignored":
+                self.comments += f"# {field} = {value} is ignored\n"
+                _warn(f"Parameter {field} in {value} section is deprecated. Ignoring it during conversion.")
+            if message == "need_remove":
```
charlesg3, line 796: `elif` and same for below `if` statements.

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
```diff
@@ -0,0 +1,1000 @@
...
+            "Indicates the 'cluster' section of the configuration file to convert. "
+            "If not specified the script will look for the cluster_template parameter in the [global] section "
+            "or will search for '[cluster default]'."
+        ),
+        required=False,
+    )
+    convert_parser.add_argument(
+        "-o",
+        "--output-file",
+        help="Configuration file to be written as output. By default the output will be written to the stdout.",
```
charlesg3, line 964: `written to stdout` (no `the` required here)

chenwany: Changed

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
chenwany: Changed

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
chenwany: Deleted

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
chenwany: Changed

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
chenwany: changed

cli/src/pcluster3_config_converter/pcluster3_config_converter.py
chenwany: Fixed. I miss to change this part and forgot about adding it to stdout.
Now fixed the function to add if there's no self.out_outfile, write the result to stdout
```
    def write_configuration_file(self):
        """Write converted yaml to the output file."""
        if self.output_file:
            if os.path.isfile(self.output_file):
                _error(f"File {self.output_file} already exists, please select another output file.")
            else:
                try:
                    config_folder = os.path.dirname(self.output_file) or "."
                    os.makedirs(config_folder)
                except OSError as e:
                    if e.errno != errno.EEXIST:
                        raise
            # Fix permissions
            with open(self.output_file, "a", encoding="utf-8"):
                os.chmod(self.output_file, stat.S_IRUSR | stat.S_IWUSR)

            # Write configuration to disk
            with open(self.output_file, "w", encoding="utf-8") as config_file:
                config_file.write(self.comments)
                yaml.dump(
                    self.pcluster3_configuration,
                    config_file,
                    sort_keys=True,
                    Dumper=ConfigDumper,
                    default_flow_style=False,
                )
            print(f"Configuration file written to {self.output_file}")
        else:
            sys.stdout.write(self.comments)
            yaml.dump(
                self.pcluster3_configuration,
                sys.stdout,
                sort_keys=True,
                Dumper=ConfigDumper,
                default_flow_style=False,
            )
            sys.stdout.flush()
```

charlesg3 (dismissed):

PR #3304, opened 2021-09-30
Pin to connexion transitive dependencies

tilne:
This change is being made to fix an issue related to the release of a new version of the jsonschema package, which is a transitive dependency of the API code via connexion. The list of dependencies was taken from [here](https://github.com/zalando/connexion/blob/main/setup.py).

### DIFF
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -51,6 +51,13 @@ REQUIRES = [
     # This does not impact any functionality. Also next connexion release will lift this limitation
     # and we will be able to upgrade.
     "connexion==2.7.0",
+    # Pin to connexion transitive dependencies.
+    # These can be removed after the next connexion release (assuming they pin these dependencies as expected).
+    "clickclick>=1.2,<21",
+    "jsonschema>=2.5.1,<4",
+    "requests>=2.9.1,<3",
+    "inflection>=0.3.1,<0.6",
+    "openapi-spec-validator>=0.2.4,<0.4",
     "flask~=2.0",
     "jmespath~=0.10",
 ]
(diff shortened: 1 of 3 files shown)


cli/setup.py
```diff
@@ -51,6 +51,12 @@ def readme():
     # This does not impact any functionality. Also next connexion release will lift this limitation
     # and we will be able to upgrade.
     "connexion==2.7.0",
+    # pin to connexion transitive dependencies
+    "clickclick>=1.2,<21",
+    "jsonschema>=2.5.1,<4",
+    "requests>=2.9.1,<3",
+    "inflection>=0.3.1,<0.6",
+    "openapi-spec-validator>=0.2.4,<0.4",
```
hanwen-pcluste, line 59: Why giving a range instead of `~=`?
Does dependabot work with the range?

tilne: I left them exactly as they were defined in the connexion source linked to in the PR description. Theoretically we could use `~=` to avoid using a range, but the goal was to use the same constraints as connexion.

cli/setup.py
```diff
@@ -51,6 +51,12 @@ def readme():
     # This does not impact any functionality. Also next connexion release will lift this limitation
     # and we will be able to upgrade.
     "connexion==2.7.0",
+    # pin to connexion transitive dependencies
```
demartinofra, line 54: could you add a note that this can be removed once we upgrade to a connexion version that includes such constraints?

tilne: Added.

hanwen-pcluste (approved):

PR #3374, opened 2021-10-21
Include tags in RunInstances dry run

tilne:
This change is motivated by users whose organizational security policies
require certain tags to be present when they're attempting to launch
instances. Without this change, the dry run performed as part of
validation would fail, and the CLI would incorrectly tell the customer
that there's an issue with their configuration file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+2.11.3
+-----
+
+**CHANGES**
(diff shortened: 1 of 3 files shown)


yuleiwan (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
```
yuleiwan, line 1: minor: not related to this PR, could you fix dependency github action before merge? For example in develop, https://github.com/aws/aws-parallelcluster/pull/3284

PR #3417, opened 2021-10-29
Integ-tests: fix pcluster AMI retrival for released tests

hanwen-pcluste:

### DIFF
--- a/tests/integration-tests/tests/common/utils.py
+++ b/tests/integration-tests/tests/common/utils.py
@@ -62,7 +62,7 @@ AMI_TYPE_DICT = {
 }
 
 
-def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64", additional_filters=None):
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/common/utils.py
```diff
@@ -62,7 +62,7 @@
 }
 
 
-def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64", additional_filters=None):
+def retrieve_latest_ami(region, os, ami_type="official", architecture="x86_64", additional_filters=None, request=None):
```
yuleiwan, line 65: If ami_type is "pcluster", shall we set request as required parameter? Since there are other tests using retrieve_latest_ami when ami_type is "pcluster"

hanwen-pcluste: How do you make it as required conditioned on the value of the other parameter?

yuleiwan (approved):

PR #3489, opened 2021-11-15
Add support to specify SchedulerDefinition as an HTTPS or S3 url

yuleiwan:
* Add support to specify SchedulerDefinition as an HTTPS or S3 url
* Pass instance types info as input to Jinja rendering

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -537,6 +537,7 @@ class Cluster:
             rendered_template = template.render(
                 cluster_configuration=ClusterSchema(cluster_name=self.name).dump(deepcopy(self.config)),
                 cluster_name=self.name,
+                instance_types_info=self.config.get_instance_types_data(),
             )
         except Exception as e:
             raise BadRequestClusterActionError(
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -16,9 +16,12 @@
 
 import copy
 import re
+from urllib.request import urlopen
 
-from marshmallow import ValidationError, fields, post_load, pre_dump, validate, validates, validates_schema
+import yaml
+from marshmallow import ValidationError, fields, post_load, pre_dump, pre_load, validate, validates, validates_schema
 
+from pcluster.aws.aws_api import AWSApi
 from pcluster.config.cluster_config import (
     AdditionalPackages,
     AmiSearchFilters,
@@ -103,6 +106,7 @@ from pcluster.constants import (
     SCHEDULER_PLUGIN_MAX_NUMBER_OF_USERS,
     SUPPORTED_OSES,
 )
+from pcluster.models.s3_bucket import parse_bucket_url
 from pcluster.schemas.common_schema import (
     AdditionalIamPolicySchema,
     BaseDevSettingsSchema,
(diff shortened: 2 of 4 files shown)


cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -575,3 +580,76 @@ def test_scheduler_plugin_scheduler_definition_schema(plugin_interface_version,
...
+        (
+            {"PluginInterfaceVersion": "1.0", "Events": {"HeadInit": {"ExecuteCommand": {"Command": "env"}}}},
+            True,
+            None,
+            None,
+            None,
+        ),
+    ],
+)
+def test_scheduler_plugin_settings_schema(
```
demartinofra, line 643: can you also test what happens when the text you try to decode is not a valid YAML? is the error clear enough or we need to more granularly handle such case?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1436,6 +1440,32 @@ def make_resource(self, data, **kwargs):
...
+                    result = AWSApi.instance().s3.get_object(
+                        bucket_name=bucket_parsing_result["bucket_name"], key=bucket_parsing_result["object_key"]
+                    )
+                    data["SchedulerDefinition"] = yaml.safe_load(result["Body"].read().decode("utf-8"))
+                elif original_scheduler_definition.startswith("https"):
+                    with urlopen(original_scheduler_definition) as f:
+                        data["SchedulerDefinition"] = yaml.safe_load(f.read().decode("utf-8"))
+                else:
+                    raise ValidationError(
+                        f"SchedulerDefinition {data['SchedulerDefinition']} is invalid, "
```
demartinofra: nit: what about "The provided value for SchedulerDefinition ('{url}') is invalid. You can specify this as an S3 URL, HTTPS URL or as an inline YAML object."

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1436,6 +1440,32 @@ def make_resource(self, data, **kwargs):
...
         return SchedulerPluginSettings(**data)
 
+    @pre_load
+    def fetch_scheduler_definition(self, data, **kwargs):
+        """Fetch scheduler definition if it is s3 or https url."""
+        original_scheduler_definition = data["SchedulerDefinition"]
+        if type(original_scheduler_definition) is str:
+            try:
+                if original_scheduler_definition.startswith("s3"):
+                    bucket_parsing_result = parse_bucket_url(original_scheduler_definition)
```
demartinofra, line 1450: what if the string starts with s3 but then the format is wrong, is the error message clear enough? same for https.

cli/src/pcluster/models/cluster.py
```diff
@@ -535,7 +535,9 @@ def _render_and_upload_scheduler_plugin_template(self):
             )
             template = environment.from_string(file_content)
             rendered_template = template.render(
-                cluster_configuration=parse_config(self.source_config_text), cluster_name=self.name
+                cluster_configuration=parse_config(self.source_config_text),
```
demartinofra: Can we extend `test_render_and_upload_scheduler_plugin_template` to test that jinja rendering is properly performed by trying to use both the hash function and also the passed parameters?

cli/tests/pcluster/schemas/test_cluster_schema.py
yuleiwan: if it is not valid YAML, yaml.safe_load will raise an error and it'll be caught in Exception and throw ValidationError

cli/src/pcluster/schemas/cluster_schema.py
yuleiwan: if it is not valid format of s3,  it'll raise an error (Invalid s3 url: {url}) and be caught in Exception, throw ValidationError with wrapped error message. For https, if it is invalid url, it'll raise urlopen error and be caught in Exception as well.

cli/src/pcluster/models/cluster.py
yuleiwan: Added return value for get_instance_types_data method in test. Shall we add some jinja template examples here?

cli/src/pcluster/schemas/cluster_schema.py
yuleiwan: done

demartinofra (approved):

PR #3544, opened 2021-11-29
[ad_integration] Use LDAPS for Simple AD integration test

tilne:
* Explicitly set file permissions being tested
*  Use NLB to implement LDAPS for SimpleAD
* Create integ test directory in private subnets
* Use better resource names in AD stack
* Manage existence of ~/.ssh directories when testing automatic keypair generation
* Reset SSH connections after cluster update
* Run AD integration integ test on ubuntu20

### DIFF
--- a/tests/integration-tests/tests/ad_integration/cluster_user.py
+++ b/tests/integration-tests/tests/ad_integration/cluster_user.py
@@ -24,14 +24,14 @@ class ClusterUser:
         # TODO: randomly generate this. It's hardcoded here because it's also hard-coded in the script
         #       that creates users as part of the directory stack.
         self.password = "ApplesBananasCherries!"
-        self._validate_automatic_homedir_creation(self.cluster.head_node_ip)
-        self._configure_public_ssh_keys()
         self._personalized_remote_command_executor = RemoteCommandExecutor(
             self.cluster, username=self.alias, alternate_ssh_key=self.ssh_private_key_path
         )
         self._personalized_scheduler_commands = get_scheduler_commands(
             scheduler, self._personalized_remote_command_executor
         )
+        self.validate_password_auth_and_automatic_homedir_creation()
+        self._configure_public_ssh_keys()
 
     def _generate_ssh_keypair(self):
         """Create an RSA SSH keypair for the user."""
@@ -50,7 +50,7 @@ class ClusterUser:
         ]
         run_command(cmd)
 
-    def _copy_public_ssh_key_to_authorized_keys(self):
+    def copy_public_ssh_key_to_authorized_keys(self):
         """Copy user's public SSH key to authorized keys file on cluster's head node."""
         user_home_dir = f"/home/{self.alias}"
         user_ssh_dir = f"{user_home_dir}/.ssh"
@@ -69,7 +69,7 @@ class ClusterUser:
 
     def _configure_public_ssh_keys(self):
         self._generate_ssh_keypair()
-        self._copy_public_ssh_key_to_authorized_keys()
+        self.copy_public_ssh_key_to_authorized_keys()
 
     def submit_script(self, script, **submit_command_kwargs):
         """Wrapper around SchedulerCommand's submit_script method."""
@@ -97,11 +97,11 @@ class ClusterUser:
         logging.info("Removing home directory for user %s (%s)", self.alias, user_home_dir)
         self._default_user_remote_command_executor.run_remote_command(f"sudo rm -rf {user_home_dir}")
 
-    def _validate_automatic_homedir_creation(self, head_node_ip, port=22):
-        """Ensure password can be used to login to cluster."""
+    def validate_password_auth_and_automatic_homedir_creation(self, port=22):
+        """Ensure password can be used to login to cluster and that user's home directory is created."""
         ssh = SSHClient()
         ssh.set_missing_host_key_policy(AutoAddPolicy())
-        ssh.connect(head_node_ip, port, self.alias, self.password, allow_agent=False, look_for_keys=False)
+        ssh.connect(self.cluster.head_node_ip, port, self.alias, self.password, allow_agent=False, look_for_keys=False)
 
         homedir = f"/home/{self.alias}"
         command = f"[ -d {homedir} ] || echo failure"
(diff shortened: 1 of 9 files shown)


tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -46,17 +47,20 @@ def get_infra_stack_outputs(stack_name):
     }
 
 
-def get_ad_config_param_vals(stack_name, bucket_name, password_secret_arn):
+def get_ad_config_param_vals(directory_stack_name, nlb_stack_name, bucket_name, password_secret_arn):
     """Return a dict used to set values for config file parameters."""
-    infra_stack_outputs = get_infra_stack_outputs(stack_name)
+    stack_outputs = {
+        stack_type: get_infra_stack_outputs(stack_name)
+        for stack_type, stack_name in [("directory", directory_stack_name), ("nlb", nlb_stack_name)]
```
hanwen-pcluste: minor: why using a loop instead of laying out the dict directly? Or maybe just use two variables `directory_stack_outputs` and `nlb_stack_outputs`?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -155,7 +159,9 @@ def _store_secret(secret):
 
 @pytest.fixture(scope="module")
 def directory_factory(request, cfn_stacks_factory):
-    created_directory_stacks = {}
+    # TODO: do we need to worry about regions?
```
hanwen-pcluste: Why wouldn't we care about the region?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -155,7 +159,9 @@ def _store_secret(secret):
...
+    # TODO: protect modification of this dict via file locking
+    created_directory_stacks = defaultdict(dict)
```
hanwen-pcluste, line 275: Why the change is necessary

tests/integration-tests/tests/ad_integration/cluster_user.py
```diff
@@ -22,10 +24,10 @@ def __init__(self, user_num, test_datadir, cluster, scheduler, default_user_remo
         # TODO: randomly generate this. It's hardcoded here because it's also hard-coded in the script
         #       that creates users as part of the directory stack.
         self.password = "ApplesBananasCherries!"
-        self._validate_automatic_homedir_creation(cluster.head_node_ip)
+        self._validate_automatic_homedir_creation(self.cluster.head_node_ip)
```
hanwen-pcluste: Why the change is necessary? Same question for line 30

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -164,78 +170,140 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
...
+            if request.config.getoption("no_delete"):
+                logging.info(
+                    "Not deleting %s stack named %s in region %s because --no-delete option was specified",
+                    stack_type,
+                    stack_name,
+                    region,
+                )
+            else:
+                logging.info("Deleting %s stack named %s in region %s", stack_type, stack_name, region)
+                boto3.client("cloudformation", region_name=region).delete_stack(StackName=stack_name)
```
hanwen-pcluste: Sorry, this is a bug from me. It should've been
```
cfn_stacks_factory.delete_stack(stack_name, region)
```

tilne: fixed.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
tilne: the main reason i made the change was because it's now a nested dict with region and then stack type keys. This prevents having to check whether a dict has already been initialized in the region with something like:

```python
if region not in dict:
  dict[region] = dict()
```

it wasn't strictly necessary.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
tilne: even within the same region clusters won't these tests probably be running in their own processes, and thus not be able to share this data structure?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
tilne: i don't have a great reason. i think the idea was to use one line instead of two, but i agree looking at it now that it would probably be simpler to read if i used two variables instead. changing it now.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hanwen-pcluste: Yes. If the tests are in different processes, they couldn't share the AD.
The idea was to reduce the creation time of AD and share it within each process.
Sharing it across processes is not easy (e.g. we need to handle racing condition)

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hanwen-pcluste: Great. I wasn't aware of the `defaultdict`

tests/integration-tests/tests/ad_integration/test_ad_integration.py
tilne: fixed.

hanwen-pcluste (dismissed):

hanwen-pcluste (dismissed):

tests/integration-tests/tests/ad_integration/test_ad_integration.py
tilne: yeah agreed. some sort of lock would be needed. the reason i leave the TODO asking if we need to handle regions is because i think we can assume that tests running within the same process will be in the same region. but it doesn't hurt anything to handle it, and it makes it safer. i'll remove the region TODO.

tilne: done

hanwen-pcluste (approved):

PR #3546, opened 2021-11-30
2.11: Use c6gn.large in place of c6gn.16xlarge for EFA tests

enrico-usai:
The test is useful to verify EFA kernel module is installed correctly on ARM instances with alinux2 and ubuntu1804 and that openmpi is installed correctly but it's not needed to use the biggest c6gn instance type.


enrico-usai:
Closed because EFA can be used on c6gn.16xlarge only, see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html#efa-instance-types

PR #3567, opened 2021-12-04
[develop] Avoid using DHCP options set in test_ad_integration

tilne:
The only portion of the test that requires the use of the DNS provided
by the directory is the configuration of the admin node. Clusters
function fine with default VPC DNS configurations.


hanwen-pcluste (approved):
Awesome!

PR #3571, opened 2021-12-06
Use validator for max compute resources and max queue for scheduler plugin

chenwany:
Use validator for max compute resources and max queue for scheduler plugin instead of validate through schema, so the validator can be suppressed when more queues and compute resources are configured

Manually test with using a config with more than 10 queues and 5 compute resources.
Manually test with changing MaxCount of QueueConstraints and ComputeResourceConstraints in the config

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -39,7 +39,7 @@ from pcluster.constants import (
     SCHEDULER_PLUGIN_QUEUE_CONSTRAINTS_MAX_SUBNETS_COUNT,
     SUPPORTED_OSES,
 )
-from pcluster.utils import get_partition, get_resource_name_from_resource_arn, replace_url_parameters
+from pcluster.utils import get_attr, get_partition, get_resource_name_from_resource_arn, replace_url_parameters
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1851,6 +1851,25 @@ def _register_validators(self):
         self._register_validator(
             DuplicateNameValidator, name_list=[queue.name for queue in self.queues], resource_name="Queue"
         )
+        self._register_validator(
+            MaxCountValidator,
+            resources_length=len(self.queues),
+            max_length=self.settings.scheduler_definition.requirements.queue_constraints.max_count
+            if self.settings.scheduler_definition.requirements
+            and self.settings.scheduler_definition.requirements.queue_constraints
+            else MAX_NUMBER_OF_QUEUES,
```
demartinofra: nit: you could use the utility get_attr from utils.py: `get_attr(self.settings.scheduler_definition, "equirements.queue_constraints.max_count", default=MAX_NUMBER_OF_QUEUES)`. Same below

chenwany: Good idea! Changed

demartinofra (approved):

PR #3575, opened 2021-12-07
Add validation for Requirements config

chenwany:
### Description of changes
* Add SchedulerPluginOsArchitectureValidator to validate Requirements.SupportedDistros in scheduler plugin
* Add SchedulerPluginRegionValidator to validate Requirements.SupportedRegions
* Add SupportedVersionsValidator to validate Requirements.SupportedParallelClusterVersions
* Remove MaxSubnetsCount and MaxInstanceTypesCount in the config
* Add validation for Requirements.SupportsClusterUpdate 

### Tests
* Add unit tests for the validators
* Manually tests with create a cluster with a config with dry-run to check the validation message

### Checklist


cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -336,6 +336,15 @@ def update_cluster(
...
             )
+        if (
+            cluster.config.scheduling.scheduler == "plugin"
+            and cluster.config.scheduling.scheduler_settings.scheduler_definition.requirements
+            and cluster.config.scheduling.scheduler_settings.scheduler_definition.requirements.supports_cluster_update
+            is False
+        ):
+            raise BadRequestException(
+                "Cluster update is not supported with SchedulerPlugin SupportsClusterUpdate set to False."
+            )
```
demartinofra: Can we bake this logic into an update policy so that we block the update only when the config under the `Scheduling` section changes? What I mean is that we want to fail the update only when SupportsClusterUpdate=False and the config is changed under the Scheduling section.


Also I'd slightly rephrase the error message: "The scheduler plugin used for this cluster does not support updating the scheduling configuration."

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +23,67 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+
+    def _validate(self, region, supported_regions):
+        if region not in supported_regions:
+            self._add_failure(
+                "The specified region {0} is not supported by Scheduler Plugin: {1}.".format(region, supported_regions),
+                FailureLevel.ERROR,
+            )
+
+
+class SupportedVersionsValidator(Validator):
```
demartinofra, line 59: Can we reuse the logic in `CliRequirementsMatcher` from awsbatch-cli package?We could assume that if no operator is prepended to the version number we assume `==`. In the general case we support any combination of operator+version, e.g. `>=3.0.0, <3.2.0, !=3.1.0`.

Shall we also add a schema level validator based on a regex to validate this string?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -592,6 +592,12 @@ class QueueNetworkingSchema(BaseNetworkingSchema):
 class SlurmQueueNetworkingSchema(QueueNetworkingSchema):
```
demartinofra: let's leave the changes to subnet_ids out of this PR for now

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +23,67 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
                 "but these privileges were not granted because GrantSudoPrivileges was not set or set to false.",
                 FailureLevel.ERROR,
             )
+
+
+class SchedulerPluginOsArchitectureValidator(Validator):
+    """Verify that head node architecture and os combination is supported by Scheduler Plugin."""
+
+    def _validate(self, os, architecture: str, supported_x86, supported_arm64):
```
demartinofra, line 35: are supported_x86 and supported_arm64 initialized to empty lists when not specified in SchedulerPlugin?

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +23,67 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+        oss_and_supported_architectures_mapping = {"x86_64": supported_x86, "arm64": supported_arm64}
+        supported_oss = oss_and_supported_architectures_mapping.get(architecture)
+        if os not in supported_oss:
+            self._add_failure(
+                "The Scheduler Plugin supports the OSs {0} for architecture {1}, none of which are "
```
demartinofra: Scheduler Plugin -> scheduler plugin, same everywhere else

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +23,67 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+                FailureLevel.ERROR,
+            )
+
+
+class SchedulerPluginRegionValidator(Validator):
+    """
+    Validate instance type and architecture combination.
+
+    Verify that head node and compute instance types imply compatible architectures.
+    """
```
demartinofra: to be updated

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +23,67 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+
+    def _validate(self, region, supported_regions):
+        if region not in supported_regions:
+            self._add_failure(
+                "The specified region {0} is not supported by Scheduler Plugin: {1}.".format(region, supported_regions),
```
demartinofra: ...by the scheduler plugin. Supported regions are: {}

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Changed to reuse the logic in CliRequirementsMatcher. Meet a question during implementation: do we want to have "or" or "and" for these version requirements? When the input is "3.0.0, 3.0.1", I think we want to take it as “3.0.0” or “3.0.1", if the input of SupportedParallelClusterVersions is “>=3.0.0, <=3.2.0, !=3.0.1”, it is likely that we want to take it as version >=3.0.0 and version<=3.2.0 and version!=3.0.1,"

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: The default of supported_x86 and supported_arm64 initialized to SUPPORTED_OSES = ["alinux2", "centos7", "ubuntu1804", "ubuntu2004"]
Change the code to
```
        self._register_validator(
            SchedulerPluginOsArchitectureValidator,
            os=self.image.os,
            architecture=self.head_node.architecture,
            supported_x86=get_attr(scheduler_definition, "requirements.supported_distros.x86", default=SUPPORTED_OSES),
            supported_arm64=get_attr(
                scheduler_definition, "requirements.supported_distros.arm64", default=SUPPORTED_OSES
            ),
        )
```
If Requirements is not specified, it will validate if it is in SUPPORTED_OSES

cli/src/pcluster/api/controllers/cluster_operations_controller.py
chenwany: Working on it

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Changed

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Done

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Changed

cli/src/pcluster/schemas/cluster_schema.py
chenwany: Ok, changed back

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: After our discussion, change it to either support a range or string. Supported format: ">=3.0.0, <=3.1.2", "3.0.0, 3.0.1b"...

cli/src/pcluster/api/controllers/cluster_operations_controller.py
chenwany: After our discussion, add it in validate_update_request

cli/src/pcluster/models/cluster.py
```diff
@@ -1081,3 +1083,21 @@ def get_log_events(
     def _stack_events_stream_name(self):
         """Return the name of the stack events log stream."""
         return STACK_EVENTS_LOG_STREAM_NAME_FORMAT.format(self.stack_name)
+
+    def _validate_scheduling_update(self, changes, target_config):
+        """Update of Scheduling is not supported when SupportsClusterUpdate of the scheduler plugin is set to false."""
+        if (
+            self.config.scheduling.scheduler == "plugin"
+            and self.config.scheduling.settings.scheduler_definition.requirements
+            and self.config.scheduling.settings.scheduler_definition.requirements.supports_cluster_update is False
```
demartinofra: you can use get_attr from utils

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1235,7 +1233,9 @@ class SchedulerPluginRequirementsSchema(BaseSchema):
     )
     requires_sudo_privileges = fields.Bool(metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
     supports_cluster_update = fields.Bool(metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    supported_parallel_cluster_versions = fields.Str(metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
+    supported_parallel_cluster_versions = fields.Str(
+        metadata={"update_policy": UpdatePolicy.UNSUPPORTED}, validate=validate.Regexp(r"[a-z0-9\>\=\<\.]")
```
demartinofra: Are you sure the regex is correct? isn't it just matching any single char among `a-z0-9\>\=\<\.`?

cli/src/pcluster/models/cluster.py
```diff
@@ -1081,3 +1083,21 @@ def get_log_events(
...
+        ) and target_config.source_config.get("Scheduling") != self.config.source_config.get("Scheduling"):
+            scheduling_changes = []
+            for change in changes[1:]:
+                if change[0][0] == "Scheduling":
```
demartinofra: Can you add a comment on why we iterate from the second element and on what is change[0][0]?

cli/src/pcluster/models/cluster.py
```diff
@@ -1081,3 +1083,21 @@ def get_log_events(
...
+                    scheduling_changes.append(change)
+            if len(scheduling_changes) >= 1:
```
demartinofra, line 1136: is it possible that `target_config.source_config.get("Scheduling") != self.config.source_config.get("Scheduling")` but there are no changes?

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +29,82 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+                    if not comparison_operators[req.operator](
+                        packaging.version.parse(installed_version),
+                        packaging.version.parse(req.version),
+                    ):
+                        self._add_failure(
+                            "The installed version {0} is not supported by the scheduler plugin ({1}).".format(
+                                installed_version, supported_versions_string
+                            ),
+                            FailureLevel.ERROR,
+                        )
```
demartinofra: nit: can we move this into a separate private function?

cli/tests/pcluster/validators/test_scheduler_plugin_validators.py
```diff
@@ -36,6 +41,123 @@
...
+            None,
+        ),
+        (
+            "3.0.0b",
+            "3.0.1,     3.0.0b",
+            None,
+        ),
+        (
+            "3.1.0",
+            ">=2.9.0, <=3.1.1",
```
demartinofra, line 136: what happens with multiple ranges?would it work?

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +29,82 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+            supported_versions_list = [version.strip() for version in supported_versions_string.split(",")]
+            if any(x in supported_versions_string for x in [">", "<"]):  # Input string is a range
+                requirements = []
+                for version in supported_versions_list:
+                    try:
+                        match = re.search(r"^([<>=!]+)([\d.]+)", version)
+                        requirements.append(CliRequirement(operator=match.group(1), version=match.group(2)))
+                    except Exception as e:
+                        self._add_failure(
+                            "Unable to parse SupportedParallelClusterVersions {0} in the scheduler plugin: {1}.".format(
```
demartinofra: let's give a hint on the supported values/format in the error message.

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -8,9 +8,16 @@
...
 # limitations under the License.
+import operator
+import re
+from collections import namedtuple
+
+from pkg_resources import packaging
 
 from pcluster.validators.common import FailureLevel, Validator
 
+CliRequirement = namedtuple("Requirement", "operator version")
```
demartinofra: we could move this to SupportedVersionsValidator class

cli/tests/pcluster/models/test_cluster.py
```diff
@@ -657,6 +663,113 @@ def test_render_and_upload_scheduler_plugin_template(self, mocker, cluster, temp
...
+            config=plugin_old_configuration,
+        )
+
+        mocker.patch("pcluster.aws.cfn.CfnClient.stack_exists", return_value=True)
+
+        if match:
+            with pytest.raises(ClusterUpdateError, match=match):
+                cluster.validate_update_request(
+                    target_source_config=plugin_new_configuration, validator_suppressors={AllValidatorsSuppressor()}
+                )
```
demartinofra: shall we also assert the changes?

demartinofra (commented):
LGTM! only few minor comments

cli/src/pcluster/models/cluster.py
chenwany: Good idea, changed

cli/src/pcluster/models/cluster.py
chenwany: Yes, it is possible, for example: {queues: [q1, q2]} is not equal to {queues: [q2, q1]},  but it doesn’t matter here because we also check changes in scheduling to check if there’s scheduling update. Add a comment in the code to explain that

cli/src/pcluster/schemas/cluster_schema.py
chenwany: With (r"[a-z0-9\>\=\<\.]"), it is able to filter out "!" and match ">=3.0.0, <=3.1.0" with multiple matches. Changed to `r"^[a-z0-9\>\=\<\.\s]*$"` so it match the entire string instead of each char

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Moved inside validation function

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Done

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Yes, moved

cli/tests/pcluster/models/test_cluster.py
chenwany: Added the assertion for scheduling_changes

cli/tests/pcluster/validators/test_scheduler_plugin_validators.py
chenwany: It works with multiple range with "and" operator

cli/src/pcluster/models/cluster.py
chenwany: Yes, added a comment to explain that, also added example format of changes in the code

cli/src/pcluster/schemas/cluster_schema.py
demartinofra: this regex is always true because you are asking to match 0 or more occurrences of the chars in []. What about something like this: `^((>|<|>=|<=)?[0-9]+\.[0-9]+\.[0-9]+,\s*)*(>|<|>=|<=)?[0-9]+\.[0-9]+\.[0-9]+$`? You can use tools like this to double check what it does: https://regex101.com/r/M0Fk1W/1

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +27,89 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+        }
+        requirements = []
+        for version in supported_versions_list:
+            try:
+                match = re.search(r"^([<>=!]+)([\d.]+)", version)
+                requirements.append(CliRequirement(operator=match.group(1), version=match.group(2)))
+            except Exception as e:
+                self._add_failure(
+                    "Unable to parse SupportedParallelClusterVersions {0} in the scheduler plugin: {1}. "
+                    "The input of SupportedParallelClusterVersions can either contains multiple version"
```
demartinofra: typos: contain multiple versions

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -22,3 +27,89 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
+                    ),
+                    FailureLevel.ERROR,
+                )
+        for req in requirements:
+            if not comparison_operators[req.operator](
+                packaging.version.parse(installed_version),
+                packaging.version.parse(req.version),
+            ):
+                self._add_failure(
+                    "The installed version {0} is not supported by the scheduler plugin ({1}).".format(
```
demartinofra: nit: The installed version {0} is not supported by the scheduler plugin. Supported versions are: {1}.

demartinofra (commented):
Good to go once we fix the regex.

cli/src/pcluster/schemas/cluster_schema.py
chenwany: Thank you! I see, with `^((>|<|>=|<=)?[0-9]+\.[0-9]+\.[0-9]+,\s*)*(>|<|>=|<=)?[0-9]+\.[0-9]+\.[0-9]+$`, we also make it not possible to filter out format like `<>3.0.0` to be invalid. I was only think about filter out string contains "!"
Changed

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Fixed

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Changed

demartinofra (approved):

Issue #3590, opened 2021-12-10
Parallel Cluster launch fails with EFA enabled


gallanik-zs:
- AWS ParallelCluster version [e.g. 2.9.0]: 3.0.2
 - Full cluster configuration without any credentials or personal data: 
[hpc-with-efa.yaml.txt](https://github.com/aws/aws-parallelcluster/files/7691618/hpc-with-efa.yaml.txt)
 - Cluster name: hpc-with-efa

**Bug description and how to reproduce:**
I am trying to create a Parallel cluster with EFA enabled, and have created a security group as instructed here [AWS Documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html). But while launching, it fails with the following error -

```
{
      "level": "ERROR",
      "type": "EfaSecurityGroupValidator",
      "message": "An EFA requires a security group that allows all inbound and outbound traffic to and from the security group itself. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security"
}
```

The security group I am using is configured correctly as shown here - 
![sg-outbound](https://user-images.githubusercontent.com/95228016/145565944-89e614b3-6330-4a73-b757-35acb1dfca4b.jpg)
![sg-inbound](https://user-images.githubusercontent.com/95228016/145565951-8682835f-9a5e-4559-a021-979af1d41765.jpg)

I was able to launch a standalone EC2 instance with EFA enabled and the same security group without any issues.

hanwen-pcluste:
Hello Gallanik,

Your security groups look good to me. You can use `--suppress-validators type:EfaSecurityGroupValidator` in the [`create-cluster` command](https://docs.aws.amazon.com/parallelcluster/latest/ug/pcluster.create-cluster-v3.html) to suppress the error and continue the cluster creation.

Moreover, I couldn't reproduce your issue. To help us understand the issue better, could you run the following python code and show us the printed content?
```
import boto3
sg = boto3.client("ec2").describe_security_groups(GroupIds=["sg-09f28e9fbfd632c3b"]).get("SecurityGroups")[0]
print(sg)
```

Thank you,
Hanwen

gallanik-zs:
Here is the pretty output (with masked sensitive values):

```
{
  "Description": "Allow access to compute nodes",
  "GroupName": "ComputeSecurityGroup-XHCTUIZWJFML",
  "IpPermissions": [
    {
      "IpProtocol": "-1",
[...]

PR #3602, opened 2021-12-14
Check all user id group pairs inside a security group rule

hanwen-pcluste:
### Description of changes
The assumption of "UserIdGroupPairs is always of length 1" is wrong. This led to false validator failures.
https://github.com/aws/aws-parallelcluster/issues/3590

### Checklist


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ CHANGELOG
 - Fix exporting of cluster logs when there is no prefix specified, previously exported to a `None` prefix.
 - Fix rollback not being performed in case of cluster update failure.  
 - Fix RootVolume schema for the HeadNode.
+- Fix EfaSecurityGroupValidator. Previously, it may produce false failures when custom security groups were provided and EFA was enabled.
```
enrico-usai, line 28: What about something like: `Fix EfaSecurityGroupValidator to correctly validate custom Security Groups with more than 1 UserIdGroupPairs when EFA is enabled.`

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -242,6 +242,31 @@ def test_efa_placement_group_validator(
             [{"IpProtocol": "-1", "UserIdGroupPairs": [{"UserId": "123456789012", "GroupId": "sg-12345678"}]}],
             None,
         ),
+        # right sg. Test when UserIdGroupPairs contains more entries
```
enrico-usai, line 245: Do we have an integration test with custom Security Group? If so I'd extend it to have more than 1 group pair.

CHANGELOG.md
hanwen-pcluste: I tried to use more user-friendly wordings in the change log, and more technical wordings in the commit description. But I am ok to change the message here

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: We do. But adding another group pair does not guarantee we could discover the issue from the integration tests. If the boto3 call returns the group pairs with the correct order, the test would not discover the issue. Moreover, it feels like a unit test to validate our validators is enough

enrico-usai (approved):

PR #3610, opened 2021-12-17
Add support for custom tags in plugin definition

chenwany:
### Description of changes
* Add support for: scheduler plugin tags:
```
Scheduling:
  Scheduler: plugin
  SchedulerSettings:
    SchedulerDefinition:
      Tags:
        - Key: key1
          Value: value1
```
* Add the tags to all resources if tags are specified

### Tests
* Add integration test to verify tags are in all resources

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1033,9 +1033,11 @@ class BaseClusterConfig(Resource):
         )
         self._register_storage_validators()
         self._register_validator(
-            HeadNodeLaunchTemplateValidator, head_node=self.head_node, ami_id=self.head_node_ami, tags=self.tags
+            HeadNodeLaunchTemplateValidator,
+            head_node=self.head_node,
+            ami_id=self.head_node_ami,
+            tags=self.get_cluster_tags(),
         )
-
         if self.head_node.dcv:
             self._register_validator(
                 DcvValidator,
@@ -1246,6 +1248,10 @@ class BaseClusterConfig(Resource):
                 self.image.os, self.head_node.architecture, ami_filters
             )
         return self._official_ami
+
+    def get_cluster_tags(self):
+        """Return tags configured in the cluster configuration."""
+        return self.tags
 
 
 class AwsBatchComputeResource(BaseComputeResource):
@@ -1817,6 +1823,7 @@ class SchedulerPluginDefinition(Resource):
         plugin_resources: SchedulerPluginPluginResources = None,
         monitoring: SchedulerPluginMonitoring = None,
         system_users: [SchedulerPluginUser] = None,
+        tags: List[Tag] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -1828,6 +1835,7 @@ class SchedulerPluginDefinition(Resource):
         self.events = events
         self.monitoring = monitoring
         self.system_users = system_users
+        self.tags = tags
 
 
 class SchedulerPluginSettings(Resource):
@@ -1910,6 +1918,12 @@ class SchedulerPluginClusterConfig(BaseClusterConfig):
                 result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
         return result
 
+    def get_cluster_tags(self):
+        """Return tags configured in the root of the cluster config and under scheduler definition."""
+        return (self.tags if self.tags else []) + get_attr(
+            self.scheduling, "settings.scheduler_definition.tags", default=[]
+        )
+
     def _register_validators(self):
         super()._register_validators()
         self._register_validator(SchedulerOsValidator, scheduler=self.scheduling.scheduler, os=self.image.os)
@@ -1936,7 +1950,10 @@ class SchedulerPluginClusterConfig(BaseClusterConfig):
 
         for queue in self.scheduling.queues:
             self._register_validator(
-                ComputeResourceLaunchTemplateValidator, queue=queue, ami_id=self.image_dict[queue.name], tags=self.tags
+                ComputeResourceLaunchTemplateValidator,
+                queue=queue,
+                ami_id=self.image_dict[queue.name],
+                tags=self.get_cluster_tags(),
             )
             queue_image = self.image_dict[queue.name]
             if queue_image not in checked_images and queue.queue_ami:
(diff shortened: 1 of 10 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1030,9 +1030,6 @@ def _register_validators(self):
             SubnetsValidator, subnet_ids=self.compute_subnet_ids + [self.head_node.networking.subnet_id]
         )
         self._register_storage_validators()
-        self._register_validator(
-            HeadNodeLaunchTemplateValidator, head_node=self.head_node, ami_id=self.head_node_ami, tags=self.tags
-        )
```
demartinofra: why don't we keep the validator only in the base class and in `SchedulerPluginDefinition` we set self.tags as the union of the two list of tags? Also do we validate that the tags are unique?

tests/integration-tests/utils.py
```diff
@@ -475,3 +475,8 @@ def instance_stream_name(instance, stream_name):
     """Return a stream name given an instance."""
     ip_str = instance["privateIpAddress"].replace(".", "-")
     return "ip-{}.{}.{}".format(ip_str, instance["instanceId"], stream_name)
+
+
+def get_head_node_instance_id(cluster):
```
demartinofra: would it make sense to have this as a property of the Cluster class (defined in clusters_factory.py) as it's done for head_node_ip?

tests/integration-tests/tests/scheduler_plugin/test_scheduler_plugin/test_scheduler_plugin_integration/pcluster.config.yaml
```diff
@@ -51,6 +51,9 @@ Scheduling:
           EnableImds: false
         - Name: {{ user2 }}
           EnableImds: true
+      Tags:
```
demartinofra, line 67: shall we also define tags in the root of the cluster config to make sure the 2 lists are properly merged?

cli/src/pcluster/config/cluster_config.py
chenwany: Because in the base class we don't have self.scheduling:
scheduler_tags = self.scheduling.settings.scheduler_definition.tags

Yes, tags are unique are validate through ComputeResourceLaunchTemplateValidator:
"message": "Unable to validate configuration parameters for instance type t2.micro. Please double check your cluster configuration. Duplicate tag key 'key1' specified."
For example, if we have tags in scheduler plugin:
```
      Tags:
        - Key: key1
          Value: value1
```
and Tags in cluster:
```
Tags:
  - Key: key1
    Value: value1
  - Key: key2
    Value: value1
```
It will raise the validation error

tests/integration-tests/tests/scheduler_plugin/test_scheduler_plugin/test_scheduler_plugin_integration/pcluster.config.yaml
chenwany: Yes, added

tests/integration-tests/utils.py
chenwany: 👍 Cool! Good idea, that simplify the code a lot that we don't need to import this function everywhere. Changed

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1931,10 +1933,12 @@ def _register_validators(self):
...
 
         checked_images = []
-
         for queue in self.scheduling.queues:
             self._register_validator(
-                ComputeResourceLaunchTemplateValidator, queue=queue, ami_id=self.image_dict[queue.name], tags=self.tags
+                ComputeResourceLaunchTemplateValidator,
+                queue=queue,
+                ami_id=self.image_dict[queue.name],
+                tags=(self.tags if self.tags else []),
```
demartinofra: nit: isn't self tags already defaulting to `[]`?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1817,6 +1822,7 @@ def __init__(
         plugin_resources: SchedulerPluginPluginResources = None,
         monitoring: SchedulerPluginMonitoring = None,
         system_users: [SchedulerPluginUser] = None,
+        tags: List[Tag] = None,
```
demartinofra, line 1826: is this really needed? probably we can remove this and the assignment in `__init__` and rely on the base class.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -174,6 +174,20 @@ def get_custom_tags(config: BaseClusterConfig, raw_dict: bool = False):
...
+            custom_scheduler_plugin_tags = (
+                {tag.key: tag.value for tag in scheduler_plugin_tags} if scheduler_plugin_tags else {}
+            )
+            custom_tags.update(custom_scheduler_plugin_tags)
+        else:
+            custom_scheduler_plugin_tags = (
+                [CfnTag(key=tag.key, value=tag.value) for tag in scheduler_plugin_tags] if scheduler_plugin_tags else []
+            )
+            custom_tags += custom_scheduler_plugin_tags
+
```
demartinofra: can we get rid of this custom logic and use `config.get_cluster_tags()` instead of config.tags in the existing code above?

tests/integration-tests/tests/scheduler_plugin/test_scheduler_plugin/test_scheduler_plugin_integration/pcluster.config.yaml
demartinofra: I don't see `Tags:` in the root of cluster config

cli/src/pcluster/config/cluster_config.py
chenwany: self.tags defaults to None with: tags: List[Tag] = None

cli/src/pcluster/templates/cdk_builder_utils.py
chenwany: Good idea! Changed

tests/integration-tests/tests/scheduler_plugin/test_scheduler_plugin/test_scheduler_plugin_integration/pcluster.config.yaml
chenwany: Forgot to commit, added

demartinofra (approved):

PR #3615, opened 2021-12-20
[AD integration]  improve ad integration tests to test LDAPS with certificate verification

hanwen-pcluste:

LDAPS connection has been tested, but without certificate verification. This commit create a self-signed certificate and instruct sssd agent to trust the certificate.

The reason for changing the `store_secret_in_secret_manager` from troposphere CloudFormation to boto3 calls is that CloudFormation only stores SecretString, but not SecretBytes

This commit also reduce the instance size used for the test.

### DIFF
--- a/tests/integration-tests/configs/ad_integration.yaml
+++ b/tests/integration-tests/configs/ad_integration.yaml
@@ -5,6 +5,6 @@ test-suites:
     test_ad_integration.py::test_ad_integration:
       dimensions:
         - regions: ["us-east-1"]
-          instances: ["c5n.18xlarge"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["alinux2", "ubuntu2004"]
           schedulers: ["slurm"]
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -2,7 +2,7 @@ ad_integration:
   test_ad_integration.py::test_ad_integration:
     dimensions:
       - regions: ["us-east-1"]
-        instances: ["c5n.18xlarge"]
+        instances: {{ common.INSTANCES_DEFAULT_X86 }}
         oss: {{ common.OSS_COMMERCIAL_X86 }}
         schedulers: ["slurm"]
 arm_pl:
--- a/tests/integration-tests/tests/ad_integration/test_ad_integration.py
+++ b/tests/integration-tests/tests/ad_integration/test_ad_integration.py
@@ -17,6 +17,7 @@ import os as os_lib
 import random
 import re
 import string
+import time
 import zipfile
 from collections import defaultdict
 
@@ -30,8 +31,6 @@ from OpenSSL.crypto import FILETYPE_PEM, TYPE_RSA, X509, dump_certificate, dump_
 from remote_command_executor import RemoteCommandExecutor
 from retrying import retry
 from time_utils import seconds
-from troposphere import Template
-from troposphere.secretsmanager import Secret
 from utils import generate_stack_name
 
 from tests.ad_integration.cluster_user import ClusterUser
@@ -51,20 +50,27 @@ def get_infra_stack_outputs(stack_name):
     }
 
 
-def get_ad_config_param_vals(directory_stack_name, nlb_stack_name, password_secret_arn):
+def get_infra_stack_parameters(stack_name):
+    cfn = boto3.client("cloudformation")
+    return {
+        entry.get("ParameterKey"): entry.get("ParameterValue")
+        for entry in cfn.describe_stacks(StackName=stack_name)["Stacks"][0]["Parameters"]
+    }
+
+
+def get_ad_config_param_vals(directory_stack_outputs, nlb_stack_parameters, password_secret_arn, ldap_tls_ca_cert):
     """Return a dict used to set values for config file parameters."""
-    directory_stack_outputs = get_infra_stack_outputs(directory_stack_name)
-    nlb_stack_outputs = get_infra_stack_outputs(nlb_stack_name)
     ldap_search_base = ",".join(
         [f"dc={domain_component}" for domain_component in directory_stack_outputs.get("DomainName").split(".")]
     )
     read_only_username = directory_stack_outputs.get("ReadOnlyUserName")
     return {
-        "ldaps_uri": nlb_stack_outputs.get("LDAPSURL"),
+        "ldaps_uri": nlb_stack_parameters.get("DomainName"),
         "ldap_search_base": ldap_search_base,
         # TODO: is the CN=Users portion of this a valid assumption?
         "ldap_default_bind_dn": f"CN={read_only_username},CN=Users,{ldap_search_base}",
         "password_secret_arn": password_secret_arn,
+        "ldap_tls_ca_cert": ldap_tls_ca_cert,
     }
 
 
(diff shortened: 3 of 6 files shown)


tests/integration-tests/configs/ad_integration.yaml
```diff
@@ -5,6 +5,6 @@ test-suites:
     test_ad_integration.py::test_ad_integration:
       dimensions:
         - regions: ["us-east-1"]
-          instances: ["c5n.18xlarge"]
+          instances: ["c5.xlarge"]
```
gmarciani: Why c5.xlarge (8GB memory, 0.17$/h)? Isn't a c5.large fairly enough (4GB memory, 0.085$/h)?

tests/integration-tests/configs/common/common.yaml
```diff
@@ -2,7 +2,7 @@ ad_integration:
   test_ad_integration.py::test_ad_integration:
     dimensions:
       - regions: ["us-east-1"]
-        instances: ["c5n.18xlarge"]
+        instances: ["c5.xlarge"]
```
gmarciani: Same as above

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -51,20 +50,27 @@ def get_infra_stack_outputs(stack_name):
...
     """Return a dict used to set values for config file parameters."""
-    directory_stack_outputs = get_infra_stack_outputs(directory_stack_name)
-    nlb_stack_outputs = get_infra_stack_outputs(nlb_stack_name)
     ldap_search_base = ",".join(
         [f"dc={domain_component}" for domain_component in directory_stack_outputs.get("DomainName").split(".")]
     )
     read_only_username = directory_stack_outputs.get("ReadOnlyUserName")
     return {
-        "ldaps_uri": nlb_stack_outputs.get("LDAPSURL"),
+        "ldaps_uri": nlb_stack_parameters.get("DomainName"),
```
gmarciani, line 68: Afaik, ldaps_uri should contain an LDAP or LDAPS Uri, not a domain name.
Why are we using a domain name here?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -315,10 +348,19 @@ def _directory_factory(
             nlb_stack_name = created_directory_stacks.get(region, {}).get("nlb")
             logging.info("Using NLB stack named %s created by another test", nlb_stack_name)
         else:
-            certificate_arn = _generate_certificate()
+            common_name = "ldap.simplead.multiuser.pcluster"
```
gmarciani: The common name should not be hard wired to simpleAD because we should use the same approach also for MsAD.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: Looking at your code below, I understood your approach and I think it's ok

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/NLB_SimpleAD.yaml
```diff
@@ -23,6 +23,12 @@ Parameters:
     Description: IP Address of secondary Simple AD instance
     AllowedPattern: "(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})"
     Type: String
+  CertificateSecretArn:
+    Description: The Arn of the secret storing the certifiacate
```
gmarciani: Typo: certificate

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update.yaml
```diff
@@ -42,7 +42,8 @@ DirectoryService:
   DomainAddr: {{ ldaps_uri }}
   PasswordSecretArn: {{ password_secret_arn }}
   DomainReadOnlyUser: {{ ldap_default_bind_dn }}
-  LdapTlsReqCert: never
+  LdapTlsCaCert: {{ ldap_tls_ca_cert }}
+  # LdapTlsReqCert: never
```
gmarciani, line 46: I think the certificate verification should be parameterizable because we publicly state that we support SimpleAD and MsAD over LDAP, LDAPS and LDAPS without certificate verification.

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.yaml
```diff
@@ -42,7 +42,8 @@ DirectoryService:
   DomainAddr: {{ ldaps_uri }}
   PasswordSecretArn: {{ password_secret_arn }}
   DomainReadOnlyUser: {{ ldap_default_bind_dn }}
-  LdapTlsReqCert: never
+  LdapTlsCaCert: {{ ldap_tls_ca_cert }}
+  # LdapTlsReqCert: never
```
gmarciani, line 46: Same as above

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update.yaml
hanwen-pcluste: Are you suggesting we should run three tests to create three clusters to test all three scenarios?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hanwen-pcluste: What would be the alternative to hard-coding?

gmarciani: for instance:

`common_name = f"{directory_type}.multiuser.pcluster"`

gmarciani: I think the `ldap` prefix is not strictly required to characterize the endpoint because both SimpleAd and MsAD are always LDAP

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update.yaml
gmarciani: I think in the long run (**not in this PR**), we must test:
1. SimpleAD over LDAP
2. SimpleAD over LDAPS without certificate verification
3. SimpleAD over LDAPS with certificate verification
4. MsAD over LDAP
5. MsAD over LDAPS without certificate verification
6. MsAD over LDAPS with certificate verification

The above ones are the scenarios we publicly support so we should test them.
What I'm suggesting here is not to implement all the above scenarios, but to make the code that you touch ready for it

tests/integration-tests/configs/ad_integration.yaml
hanwen-pcluste: c5.xlarge is the common instance type we are using in integration tests.
However, the hard-coding was not ideal. I changed it to {{ common.INSTANCES_DEFAULT_X86 }}

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hanwen-pcluste: Done

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/NLB_SimpleAD.yaml
hanwen-pcluste: Thx

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update.yaml
hanwen-pcluste: I totally agree with you. Here we are testing the case when `LdapTlsReqCert` is not specified. Can we leave the further improvement to the future.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -315,10 +348,19 @@ def _directory_factory(
             nlb_stack_name = created_directory_stacks.get(region, {}).get("nlb")
             logging.info("Using NLB stack named %s created by another test", nlb_stack_name)
         else:
-            certificate_arn = _generate_certificate()
+            common_name = f"{directory_type}.multiuser.pcluster"
```
gmarciani: Nitpick: enforce common name to be lowercase. directory_type string could be uppercase/camelcase/whatever, but we want the common name to be lowercase.

gmarciani: Common name matching during cert verification should be case insensitivity. that's why this is a minor nitpick. Nevertheless, it looks like there could be situations where the verification is case sensitive and it ciuld cause some headaches :-)

If your tests succeed, let's consider this a nitpick, otherwise, please change.

hanwen-pcluste: Done

gmarciani (approved):

PR #3624, opened 2021-12-23
3.1 Fix DomainAddrValidator when checking TLS disablement in SSD additional config

gmarciani:
### Description of changes
1. Fix DomainAddrValidator when checking TLS disablement in SSD additional config. The bug causes boolean values to be  wrongly evaluated. Without this fix, customer cannot create a cluster integrated with LDAP because validator raises an exception
1. Added unit tests to cover patch

### Tests
1. Unit tests
1. Manual test

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -37,7 +37,8 @@ class DomainAddrValidator(Validator):
         elif domain_addr_scheme == "ldap":
             warning_message = "The use of the ldaps protocol is strongly encouraged for security reasons."
             tls_disabled = (
-                additional_sssd_configs.get("ldap_auth_disable_tls_never_use_in_production", "false").lower() == "true"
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -37,7 +37,8 @@ def _validate(self, domain_addr, additional_sssd_configs):
         elif domain_addr_scheme == "ldap":
             warning_message = "The use of the ldaps protocol is strongly encouraged for security reasons."
             tls_disabled = (
-                additional_sssd_configs.get("ldap_auth_disable_tls_never_use_in_production", "false").lower() == "true"
+                str(additional_sssd_configs.get("ldap_auth_disable_tls_never_use_in_production", "false")).lower()
```
enrico-usai, line 40: We should include a unit test to cover the bug.

gmarciani: Added in next rev

enrico-usai (approved):

PR #3653, opened 2022-01-12
3.x. Modify Intel MPI check to support 2021 version

enrico-usai:
### Description of changes
* With version 2021 the package is no longer installed from rpm package but with standalone installer from Intel.

### Tests
* Tested with build-ami process

### References
* Related to: https://github.com/aws/aws-parallelcluster-cookbook/pull/1313

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
@@ -235,9 +235,11 @@ phases:
                   rpm -qa | grep libfabric && rpm -qa | grep efa-
                   [[ $? -ne 0 ]] && echo "Check efa rpm failed" && exit 1
 
-                  echo "Checking intel mpi installed..."
-                  rpm -qa | grep intel-mpi
-                  [[ $? -ne 0 ]] && echo "Check intel-mpi rpm failed" && exit 1
+                  echo "Checking Intel MPI 20xx installed and module available..."
+                  unset MODULEPATH
+                  source /etc/profile.d/modules.sh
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
```diff
@@ -235,9 +235,8 @@ phases:
                   rpm -qa | grep libfabric && rpm -qa | grep efa-
                   [[ $? -ne 0 ]] && echo "Check efa rpm failed" && exit 1
 
-                  echo "Checking intel mpi installed..."
-                  rpm -qa | grep intel-mpi
-                  [[ $? -ne 0 ]] && echo "Check intel-mpi rpm failed" && exit 1
+                  echo "Checking Intel MPI installed..."
+                  [[ ! -d "/opt/intel/mpi" ]] && echo "Check Intel MPI failed" && exit 1
```
gmarciani: Is it possible to verify the installation in a more robust way rather than a directory check?
For instance calling the command version or similar.

gmarciani (changes requested):

cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
enrico-usai: Done, I'm not checking the specific 2021 version to avoid the need to modify cli package when upgrading the package from the cookbook.

We already have a check for the version in kitchen tests.

gmarciani (dismissed):

gmarciani (approved):

PR #3666, opened 2022-01-15
Add security group ingress if custom security groups are used

hanwen-pcluste:

### Description of changes
If custom security groups are provided in the cluster configuration to some queues and/or the head node, but not to all of the cluster nodes, the pcluster created security groups will allow inbound connections from the custom security groups.

Before this change, the mixture of managed and custom security groups led to cluster nodes not accessible to each other. Specifically, nodes in custom security groups cannot access nodes in managed security groups.

With this change, we ensure managed security groups allows access from nodes in custom security groups. However, if the custom security groups are not properly configured, there will still be connection issues among cluster nodes.

Mixing managed security groups and custom security groups is not common. So a warning is added.

### Tests
`test_security_group` has been passed.
configs/pcluster3.yaml is under testing
A unit test is added for the warning validator. `test_security_group` integration test is expanded to test the scenario when a managed security group and a custom security group are mixed.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ CHANGELOG
 - Install NVIDIA drivers and CUDA library for ARM.
 - Add `parallelcluster:compute-resource-name` tag to LaunchTemplates used by compute nodes.
 - Explicitly set cloud-init datasource to be EC2. This save boot time for Ubuntu and CentOS platforms.
+- Improve Security Groups created within the cluster to allow inbound connections from custom security groups when `SecurityGroups` parameters are specified for some head node and/or queues.`
 
 **CHANGES**
 - Upgrade Slurm to version 21.08.5.
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -74,6 +74,7 @@ from pcluster.validators.cluster_validators import (
     IntelHpcArchitectureValidator,
     IntelHpcOsValidator,
     MaxCountValidator,
+    MixedSecurityGroupOverwriteValidator,
     NameValidator,
     NumberOfStorageValidator,
     OverlappingMountDirValidator,
@@ -2015,6 +2016,11 @@ class SlurmClusterConfig(BaseClusterConfig):
         self._register_validator(SchedulerOsValidator, scheduler=self.scheduling.scheduler, os=self.image.os)
         self._register_validator(
             HeadNodeImdsValidator, imds_secured=self.head_node.imds.secured, scheduler=self.scheduling.scheduler
+        )
+        self._register_validator(
+            MixedSecurityGroupOverwriteValidator,
+            head_node_security_groups=self.head_node.networking.security_groups,
+            queues=self.scheduling.queues,
         )
         if self.scheduling.settings and self.scheduling.settings.dns and self.scheduling.settings.dns.hosted_zone_id:
             self._register_validator(
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -417,39 +417,77 @@ class ClusterCdkStack(Stack):
     def _add_security_groups(self):
         """Associate security group to Head node and queues."""
         # Head Node Security Group
-        head_security_group = None
-        if not self.config.head_node.networking.security_groups:
-            head_security_group = self._add_head_security_group()
+        managed_head_security_group = None
+        custom_head_security_groups = self.config.head_node.networking.security_groups or []
+        if not custom_head_security_groups:
+            managed_head_security_group = self._add_head_security_group()
+            head_node_security_groups = [managed_head_security_group.ref]
+        else:
+            head_node_security_groups = custom_head_security_groups
 
         # Compute Security Groups
         managed_compute_security_group = None
-        if any(not queue.networking.security_groups for queue in self.config.scheduling.queues):
+        custom_compute_security_groups = set()
+        managed_compute_security_group_required = False
+        for queue in self.config.scheduling.queues:
+            queue_security_groups = queue.networking.security_groups
+            if queue_security_groups:
+                for security_group in queue_security_groups:
+                    custom_compute_security_groups.add(security_group)
+            else:
+                managed_compute_security_group_required = True
+        compute_security_groups = list(custom_compute_security_groups)
+        if managed_compute_security_group_required:
             managed_compute_security_group = self._add_compute_security_group()
+            compute_security_groups.append(managed_compute_security_group.ref)
 
-        if head_security_group and managed_compute_security_group:
-            # Access to head node from compute nodes
-            ec2.CfnSecurityGroupIngress(
-                self,
-                "HeadNodeSecurityGroupComputeIngress",
-                ip_protocol="-1",
-                from_port=0,
-                to_port=65535,
-                source_security_group_id=managed_compute_security_group.ref,
-                group_id=head_security_group.ref,
-            )
+        self._add_inbounds_to_managed_security_groups(
+            compute_security_groups,
+            custom_compute_security_groups,
+            head_node_security_groups,
+            managed_compute_security_group,
+            managed_head_security_group,
+        )
 
+        return managed_head_security_group, managed_compute_security_group
+
+    def _add_inbounds_to_managed_security_groups(
+        self,
+        compute_security_groups,
+        custom_compute_security_groups,
+        head_node_security_groups,
+        managed_compute_security_group,
+        managed_head_security_group,
+    ):
+        if managed_head_security_group:
+            for security_group in compute_security_groups:
+                # Access to head node from compute nodes
+                self._allow_all_ingress(
+                    "HeadNodeSecurityGroupComputeIngress", security_group, managed_head_security_group.ref
+                )
+        if managed_compute_security_group:
             # Access to compute nodes from head node
-            ec2.CfnSecurityGroupIngress(
-                self,
-                "ComputeSecurityGroupHeadNodeIngress",
-                ip_protocol="-1",
-                from_port=0,
-                to_port=65535,
-                source_security_group_id=head_security_group.ref,
-                group_id=managed_compute_security_group.ref,
-            )
+            for security_group in head_node_security_groups:
+                self._allow_all_ingress(
+                    "ComputeSecurityGroupHeadNodeIngress", security_group, managed_compute_security_group.ref
+                )
+            for security_group in custom_compute_security_groups:
+                self._allow_all_ingress(
+                    "ComputeSecurityGroupCustomComputeSecurityGroupIngress",
+                    security_group,
+                    managed_compute_security_group.ref,
+                )
 
-        return head_security_group, managed_compute_security_group
+    def _allow_all_ingress(self, description, source_security_group_id, group_id):
+        ec2.CfnSecurityGroupIngress(
+            self,
+            description,
+            ip_protocol="-1",
+            from_port=0,
+            to_port=65535,
+            source_security_group_id=source_security_group_id,
+            group_id=group_id,
+        )
 
     def _add_compute_security_group(self):
         compute_security_group = ec2.CfnSecurityGroup(
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -10,6 +10,7 @@
 # limitations under the License.
 import re
 from abc import ABC
+from enum import Enum
 
 from pcluster.aws.aws_api import AWSApi
 from pcluster.aws.common import AWSClientError
@@ -28,6 +29,7 @@ from pcluster.constants import (
 from pcluster.utils import get_installed_version, get_supported_os_for_architecture, get_supported_os_for_scheduler
 from pcluster.validators.common import FailureLevel, Validator
 
+# pylint: disable=C0302
 NAME_MAX_LENGTH = 25
 SHARED_STORAGE_NAME_MAX_LENGTH = 30
 NAME_REGEX = r"^[a-z][a-z0-9\-]*$"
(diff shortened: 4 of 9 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1574,6 +1575,21 @@ def _register_validators(self):
             resource_name="SlurmQueues",
         )
 
+    def get_security_group_overwrite_status(self):
```
enrico-usai: See code scanning message. It means the code structure is not returning a value for each case. It is due to multiple if blocks and missing return at the end.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1574,6 +1575,21 @@ def _register_validators(self):
...
+            if queue.networking.security_groups:
+                custom = True
+            else:
+                managed = True
+        if custom and managed:
+            return "mixed"
+        if custom:
+            return "custom"
+        if managed:
+            return "managed"
```
enrico-usai: Instead of returning string like these you should define an enum and use it here for returning the value and in the caller to check the result. This helps to have a more organized and extendible code and to avoid issues if there are typos in the strings.

cli/src/pcluster/config/cluster_config.py
enrico-usai: at the end you should have something like `return x if y else z`

enrico-usai: the name of the function should include the "queues" word

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -416,39 +416,77 @@ def _add_head_eni(self):
...
+            managed_head_security_group = self._add_head_security_group()
+            head_node_security_groups = [managed_head_security_group.ref]
+        else:
+            head_node_security_groups = custom_head_security_groups
 
         # Compute Security Groups
         managed_compute_security_group = None
-        if any(not queue.networking.security_groups for queue in self.config.scheduling.queues):
+        custom_compute_security_groups = set()
+        need_to_create_managed_compute_security_group = False
```
enrico-usai: `need_to_create_managed_compute_security_group` --> `managed_compute_security_group_required`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1574,6 +1575,21 @@ def _register_validators(self):
...
+        custom = False
+        for queue in self.queues:
+            if queue.networking.security_groups:
+                custom = True
+            else:
+                managed = True
+        if custom and managed:
+            return "mixed"
+        if custom:
+            return "custom"
```
enrico-usai: This function should be covered by unit test to be sure our logic is correct (you can pass a fake dictionary for queue networking security groups items).

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -697,6 +697,24 @@ def _validate(self, name_list, resource_name):
...
 
+class MixedSecurityGroupOverwriteValidator(Validator):
+    """Warn if some nodes are using custom security groups while others are using managed security groups."""
+
+    def _validate(self, head_node_security_groups, compute_security_group_overwrite_status):
+        if (head_node_security_groups and compute_security_group_overwrite_status != "custom") or (
+            head_node_security_groups is None and compute_security_group_overwrite_status != "managed"
+        ):
+            self._add_failure(
+                "Specifying security groups for head node and queues stops ParallelCluster from "
```
enrico-usai: I'd write just:
```
"Please make sure that all cluster nodes are reachable to each other, "
"or consider using additional security groups rather than replacing ParallelCluster security groups."
```

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -697,6 +697,24 @@ def _validate(self, name_list, resource_name):
             )
 
 
+class MixedSecurityGroupOverwriteValidator(Validator):
+    """Warn if some nodes are using custom security groups while others are using managed security groups."""
+
+    def _validate(self, head_node_security_groups, compute_security_group_overwrite_status):
+        if (head_node_security_groups and compute_security_group_overwrite_status != "custom") or (
```
enrico-usai: let's write unit test for all the cases because there are many conditions.

tests/integration-tests/remote_command_executor.py
```diff
@@ -27,12 +27,21 @@ def __init__(self, result):
...
+        """
+        Initiate SSH connection
+
+        By default, commands are executed on head node. If `compute_node_ip` is specified, execute  commands on compute.
+        """
         if not username:
             username = get_username_for_os(cluster.os)
+        if compute_node_ip:
+            node_ip = compute_node_ip
+            bastion = f"{username}@{cluster.head_node_ip}"
```
enrico-usai, line 41: not clear what we're changing here, can you add more info in the commit message?

tests/integration-tests/remote_command_executor.py
```diff
@@ -27,12 +27,21 @@ def __init__(self, result):
 class RemoteCommandExecutor:
     """Execute remote commands on the cluster head node."""
 
-    def __init__(self, cluster, username=None, bastion=None, alternate_ssh_key=None):
-        head_node_ip = cluster.head_node_ip
+    def __init__(self, cluster, compute_node_ip=None, username=None, bastion=None, alternate_ssh_key=None):
+        """
+        Initiate SSH connection
+
+        By default, commands are executed on head node. If `compute_node_ip` is specified, execute  commands on compute.
```
enrico-usai: minor: extra space

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
...
+            vpc_security_group_id=custom_security_group_id,
+        )
+        cluster.update(str(updated_config_file), force_update="true")
+        _check_connections_between_head_node_and_compute_nodes(cluster)
+
+
+def _check_connections_between_head_node_and_compute_nodes(cluster):
+    username = get_username_for_os(cluster.os)
+    head_node_public_ip = cluster.describe_cluster_instances(node_type="HeadNode")[0].get("publicIpAddress")
+    head_node_private_ip = cluster.describe_cluster_instances(node_type="HeadNode")[0].get("privateIpAddress")
```
enrico-usai: you can avoid two calls to describe_cluster instances.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
...
+    head_node_username_ip = f"{username}@{head_node_public_ip}"
+    compute_nodes = cluster.describe_cluster_instances(node_type="Compute")
```
enrico-usai: let's add a new line between head node variable initialization and compute node code blocks.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
...
+    head_node_remote_command_executor = RemoteCommandExecutor(cluster)
+    for compute_node in compute_nodes:
+        compute_node_ip = compute_node.get("privateIpAddress")
+        head_node_remote_command_executor.run_remote_command(f"ping {compute_node_ip} -c 5")
```
enrico-usai: You can define a variable for the command and reuse it everywhere.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
         assert_that(mount_target_security_groups[0]).is_equal_to(custom_security_group_id)
         assert_that(mount_target_security_groups).is_length(1)
 
+    if scheduler == "slurm":
+        _check_connections_between_head_node_and_compute_nodes(cluster)
+        updated_config_file = pcluster_config_reader(
+            config_file="pcluster.config.update.yaml",
+            vpc_security_group_id=custom_security_group_id,
+        )
+        cluster.update(str(updated_config_file), force_update="true")
```
enrico-usai, line 99: let'add a comment here describing we're modifying SG for both head node and compute and why this update is useful

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
...
+            "SecurityGroupSelfIngressResource",
             IpProtocol="-1",
             FromPort=0,
             ToPort=65535,
-            SourceSecurityGroupId=Ref(security_group),
+            CidrIp=vpc_cidr,
+            GroupId=Ref(security_group),
+        )
+    )
+    template.add_resource(
```
enrico-usai, line 155: let's add a comment describing why we need this

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -85,27 +86,67 @@ def test_overwrite_sg(region, custom_security_group, pcluster_config_reader, clu
...
+    compute_nodes = cluster.describe_cluster_instances(node_type="Compute")
+    head_node_remote_command_executor = RemoteCommandExecutor(cluster)
+    for compute_node in compute_nodes:
+        compute_node_ip = compute_node.get("privateIpAddress")
+        head_node_remote_command_executor.run_remote_command(f"ping {compute_node_ip} -c 5")
+        compute_remote_command_executor = RemoteCommandExecutor(
+            cluster, compute_node_ip=compute_node_ip, bastion=head_node_username_ip
+        )
+        compute_remote_command_executor.run_remote_command(f"ping {head_node_private_ip} -c 5")
+        for other_compute_node in compute_nodes:
```
enrico-usai, line 120: here you're also testing connection with the node itself

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Done

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Yes. For the simplicity of the code, we can keep it

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Done

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Done

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: The code seems straightforward enough. Can we keep it?

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Done

cli/src/pcluster/validators/cluster_validators.py

cli/src/pcluster/templates/cluster_stack.py

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Done.
Thank you for the explanation. The code scanning was clear to me. I just didn't feel the necessity to address it

cli/src/pcluster/config/cluster_config.py

tests/integration-tests/remote_command_executor.py
hanwen-pcluste: Done

tests/integration-tests/remote_command_executor.py
hanwen-pcluste: Done

cli/src/pcluster/validators/cluster_validators.py

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ CHANGELOG
 - Upgrade Slurm to version 21.08.5.
 - Disable EC2 ImageBuilder enhanced image metadata when building ParallelCluster custom images.
 - Disable packages update at instance launch time on Amazon Linux 2.
+- If security groups are specified for some head node and computes but not all, the security groups created by ParallelCluster allows inbound connections from the custom security groups.
```
enrico-usai: I'd move this to enhancements. I'd write something like:

Improve Security Groups created within the cluster allowing inbound connections from custom security groups when `SecurityGroups` parameter is specified for Head Node or Queue, but not both.`

In general we should start the sentence with an action (Add/Improve/Disable, etc).

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1574,6 +1576,22 @@ def _register_validators(self):
             resource_name="SlurmQueues",
         )
 
+    def get_queues_security_group_overwrite_status(self):
```
enrico-usai: This function is validator specific, we should move it in the validator class.

When registering the validator we can pass: `self.head_node.networking.security_groups` and 
 `self.scheduling.queues` then it's the validator that should know the logic to identify the status analyzing all the queues.

By moving the method you don't need to import the new Enum here.

enrico-usai: Then after moving will be easier to create unit tests for `get_queues_security_group_overwrite_status`private method of the validator (that is not yet covered).

See for example unit test for `_is_instance_type_format`. private method of `AwsBatchInstancesArchitectureCompatibilityValidator`

tests/integration-tests/tests/networking/test_security_groups.py
enrico-usai: The point is if we are going to replace the `ping` command or add a parameter to ping command (e.g. timeout, interval, verbosity) we need to remember to change it in 3 places.

tests/integration-tests/tests/networking/test_security_groups.py
enrico-usai: From the command (and from the code) I can understand we're allowing inbound SSH connection but it's not clear "why" we need to open SSH from everywhere. I think it's because the test instance needs to connect to it to execute commands in the node, but we should clarify this in the comment.

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: We have been passing simple data types to validators, which helps to decrease the complexity of unit tests.
I agree that moving the method to the validator do reduce the complexity of source code. But it increases complexity of the unit tests, because we are passing the `queues` list of object to the validator

CHANGELOG.md
hanwen-pcluste: Done

tests/integration-tests/tests/networking/test_security_groups.py
hanwen-pcluste: Done

cli/src/pcluster/config/cluster_config.py
enrico-usai: By looking at our validators I can see some of them are doing a cycle for all the compute resources of a given queue. This is a similar case, here the responsibility to check the status of this parameter should be of the validator, the overwrite status is useless in the config module and we also have an Enum that is very specific to the validator so I'd move this code.

In general let's focus on source code complexity rather than unit tests complexity (even if hard to write unit tests means bad smell in the code structure).

Anyway let's talk about this before proceeding.

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py

enrico-usai (approved):

PR #3721, opened 2022-01-29
Fix FSx namespace for FSx metrics displayed in Cloudwatch dashboard

chenwany:
Fix FSx metrics not displayed in Cloudwatch Dashboard. The namespace should be AWS/FSx.
Add unit test to test sharedstorage namespace


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 CHANGELOG
 =========
+3.x.x
+------
+**BUG FIXES**
+Fix FSx metrics not displayed in Cloudwatch dashboard
```
mauri-melato: If we merge it soon, this is going to be part of 3.1.0

cli/tests/pcluster/templates/test_cw_dashboard_builder.py
```diff
@@ -75,9 +75,10 @@ def _verify_ec2_metrics_conditions(cluster_config, output_yaml):
     # Check each section title
     for storage_type, storages in storage_resource.items():
         if len(storages) > 0:
-            assert_that(output_yaml).contains(storage_type_title_dict[storage_type])
+            assert_that(output_yaml).contains(storage_type_title_dict[storage_type].get("title"))
```
enrico-usai: nit: one way to avoid code replication is:
```python
for field in ["title", "namespace"]:
      assert_that(output_yaml).contains(storage_type_title_dict[storage_type].get(field))
```

enrico-usai (commented):
All the fields in the PR description template should be populated. Can you please update it?

cli/tests/pcluster/templates/test_cw_dashboard_builder.py
chenwany: The reason I didn't use the for loop is that it only use for the "if" case but not in the "else" case.
Because in the "else" case, EBS and Raid both have nane space "AWS/EBS".
Change to:
```
    for storage_type, storages in storage_resource.items():
        if len(storages) > 0:
            for field in ["title", "namespace"]:
                assert_that(output_yaml).contains(storage_type_title_dict[storage_type].get(field))
        else:
            assert_that(output_yaml).does_not_contain(storage_type_title_dict[storage_type].get("title"))
```

enrico-usai (approved):

Issue #3711, opened 2022-01-26
(2.10.0 - 2.11.4) Tags in number interpreted as integer instead of string possible cause value error in Compute resource launch template


chenwany:
**Bug description**

When creating cluster with tags set in the cluster configuration, if the tag’s key or value are in number,  they are interpreted as integer instead of string, possibly causing unexpected value conversion when tagging compute resources.
For example, cluster section with value tags = {"key" : "00000122"} is interpreted as “Key: key, Value: 82” for compute resources launched from launch template, because "00000122" is considered an octal and its string representation is converted to decimal.

This issue affects versions >=2.10.0, <=2.11.4 when creating cluster with number as tags.

**Causes of the problem**
ParallelCluster create compute fleet launch templates and pass the tags specified in the cluster configuration through cfn yaml template. The problem is that when yaml passing number without quote, it interprets it as integer instead of string. 
```
- Key: {{ tag.Key }}  
  Value: {{ tag.Value }}
```
Quotes is needed if the passing value are numbers in case we want to interpret them as string.
```
- Key: "{{ tag.Key }}"  
  Value: "{{ tag.Value }}"
```

**Workaround**
You can find the mitigation of the problem [here](https://github.com/aws/aws-parallelcluster/wiki/(2.10.0---2.11.4)-Tags-in-number-interpreted-as-integer-instead-of-string-possible-cause-value-error-in-Compute-resource-launch-template).

enrico-usai:
[2.11.5](https://github.com/aws/aws-parallelcluster/releases/tag/v2.11.5) version is available on PyPI.

PR #3732, opened 2022-02-01
Fix Tags in number interpret as integer

chenwany:
This PR is created to fix this issue: https://github.com/aws/aws-parallelcluster/issues/3711
When creating cluster with tags set in the cluster configuration, if the tag’s key or value are in number, they are interpreted as integer instead of string, possibly causing unexpected value conversion when tagging compute resources.
For example, cluster section with value tags = {"key" : "00000122"} is interpreted as “Key: key, Value: 82” for compute resources launched from launch template, because "00000122" is considered an octal and its string representation is converted to decimal.

### DIFF
--- a/cloudformation/compute-fleet-hit-substack.cfn.yaml
+++ b/cloudformation/compute-fleet-hit-substack.cfn.yaml
@@ -201,8 +201,8 @@ Resources:
               - Key: aws-parallelcluster-node-type
                 Value: Compute
     {%- for tag in tags %}
-              - Key: {{ tag.Key }}
-                Value: {{ tag.Value }}
+              - Key: '{{ tag.Key }}'
+                Value: '{{ tag.Value }}'
     {%- endfor %}
           - ResourceType: volume
             Tags:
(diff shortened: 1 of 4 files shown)


enrico-usai:
Can you please populate the PR description? Thanks!

cloudformation/compute-fleet-hit-substack.cfn.yaml
```diff
@@ -201,8 +201,8 @@ Resources:
               - Key: aws-parallelcluster-node-type
                 Value: Compute
     {%- for tag in tags %}
-              - Key: {{ tag.Key }}
-                Value: {{ tag.Value }}
+              - Key: "{{ tag.Key }}"
+                Value: "{{ tag.Value }}"
```
demartinofra: do we have any validation on the chars allowed as part of a tag Key or Value? what happens if the tag key contains a double quote?

chenwany: We don't have validation on tags key and value.
Cloudformation has regrex to check the keyy and value, 
If the tag key or value contain quote, cluster creation will fail with: 
```
  - AWS::IAM::Role RootRole 2 validation errors detected: Value ''00000122'' at 'tags.4.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]*; Value '"InternalOrderNumber"' at 'tags.4.member.key' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]+ (Service: AmazonIdentityManagement; Status Code: 400; Error Code: ValidationError; Request ID: f3dde7b7-ded8-47b4-a5c0-f8480d6e7c0d; Proxy: null)
```

demartinofra (dismissed):

hanwen-pcluste (approved):

PR #3736, opened 2022-02-01
Do not grant system users sudo privileges in case GrantSudoPrivileges is set to false

chenwany:
Add a validator so that we fail in case `GrantSudoPrivileges: false` and SystemUsers in plugin definition have SudoerConfiguration

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -118,6 +118,7 @@ from pcluster.validators.s3_validators import (
     UrlValidator,
 )
 from pcluster.validators.scheduler_plugin_validators import (
+    GrantSudoPrivilegesValidator,
     SchedulerPluginOsArchitectureValidator,
     SchedulerPluginRegionValidator,
     SudoPrivilegesValidator,
@@ -1877,6 +1878,12 @@ class SchedulerPluginSettings(Resource):
             requires_sudo_privileges=self.scheduler_definition.requirements.requires_sudo_privileges
             if self.scheduler_definition.requirements
             else None,
+        )
+
+        self._register_validator(
+            GrantSudoPrivilegesValidator,
+            grant_sudo_privileges=self.grant_sudo_privileges,
+            system_users=get_attr(self.scheduler_definition, "system_users"),
         )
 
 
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1878,6 +1879,18 @@ def _register_validators(self):
             if self.scheduler_definition.requirements
             else None,
         )
+        system_users = get_attr(self.scheduler_definition, "system_users")
+        if system_users:
+            sudoer_configuration = [
+                get_attr(user, "sudoer_configuration")
+                for user in system_users
+                if get_attr(user, "sudoer_configuration")
+            ]
```
demartinofra: Shall we move this entire logic in the validator and rather than passing sudoer_configuration to it we pass system_uses?

cli/src/pcluster/validators/scheduler_plugin_validators.py
```diff
@@ -29,6 +29,18 @@ def _validate(self, grant_sudo_privileges: bool, requires_sudo_privileges: bool)
...
 
 
+class GrantSudoPrivilegesValidator(Validator):
+    """Verify SystemUsers does not have SudoerConfiguration set when GrantSudoPrivileges is true."""
+
+    def _validate(self, grant_sudo_privileges: bool, sudoer_configuration):
+        if not grant_sudo_privileges and sudoer_configuration:
+            self._add_failure(
+                "SudoerConfiguration of SystemUsers can not be set because GrantSudoPrivileges is not set or set to "
+                "false.",
```
demartinofra: reword to: `The used scheduler plugin requires the creation of SystemUsers with sudo access. To grant such rights please set GrantSudoPrivileges to true.`

cli/src/pcluster/config/cluster_config.py
chenwany: Passing the logic into the validator will make it more complicate with unit test. In this case we need to create Systemusers config object in unit test

chenwany: Changed

cli/src/pcluster/validators/scheduler_plugin_validators.py
chenwany: Changed

demartinofra (approved):

PR #3762, opened 2022-02-08
Add BucketOwner and Checksum for SchedulerDefinition

chenwany:
### Description of changes
* Add support for `SchedulerDefinitionS3BucketOwner` and `SchedulerDefinitionChecksum` when `SchedulerDefinition` is url.
* Add unit tests
* `SchedulerDefinitionChecksum` and `SchedulerDefinitionS3BucketOwner` is not allowed when SchedulerDefinition is not url
* `SchedulerDefinitionS3BucketOwner` is not allowed when SchedulerDefinition is not s3 url


scheduler_plugins/plugin_template/examples/cluster_configuration.yaml
```diff
@@ -10,6 +10,8 @@ Scheduling:
   Scheduler: plugin
   SchedulerSettings:
     SchedulerDefinition: <PLUGIN_DEFINITION>
+    SchedulerDefinitionS3BucketOwner: "012345678910"
+    SchedulerDefinitionChecksum: "0b33ca1576c6adbced3205ee2a1c270e424294e29576e8e92c9379a0fbc2924b"
```
demartinofra: we should not hardcode this here. Check how this is done in plugin_definition for the other artifacts. The upload_artifacts.sh script takes care of setting those values.

Also we should do the same for the Slurm plugin

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1474,6 +1476,72 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+                )
+
+    def _validate_s3_bucket_onwer(self, s3_bucket_owner, e):
+        if s3_bucket_owner and isinstance(e, AWSClientError) and e.error_code == "AccessDenied":
+            message = f" This can be due to bucket owner not matching the expected one '{s3_bucket_owner}'"
+        else:
+            message = ""
+        return message
+
+    def _validate_schedluer_defination_url(
```
demartinofra: why we renamed this if it's still fetching? I'd rather keep the previous name

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1474,6 +1476,72 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+        metadata={"update_policy": UpdatePolicy.UNSUPPORTED}, validate=validate.Regexp(r"^[A-Fa-f0-9]{64}$")
+    )
+
+    def _verify_checksum(self, file_content, original_definition, expected_checksum):
+        if expected_checksum:
+            actual_checksum = hashlib.sha256(file_content.encode()).hexdigest()
+            if actual_checksum != expected_checksum:
+                raise ValidationError(
+                    f"Error when validating SchedulerDefinition '{original_definition}': "
+                    f"checksum: {actual_checksum} does not match expected one: {expected_checksum}"
```
demartinofra: nit: checksum ({actual_checksum}) does not match expected one ({expected_checksum})

cli/src/pcluster/schemas/cluster_schema.py
demartinofra: Can we extract from this function the logic to fetch from s3 in a separate function with its own error handling and the logic to fetch from url in a separate function with its own error handling?

chenwany: I didn't  rename the function `fetch_scheduler_definition`, it is still in the code. The function `fetch_scheduler_definition` is too long after adding the check for bucket owner and checksum, so I break it into different functions. For S3 checksum and https, we firstly download the artifact, then we verify the checksum for it, `_validate_schedluer_definition_url` is similar with what we do in cluster.py. In this function, we also handle a case when it is not S3 or https

scheduler_plugins/plugin_template/examples/cluster_configuration.yaml
chenwany: I was thinking they are just some examples so I hardcode some values there. Yes, added for Slurm plugin

cli/src/pcluster/schemas/cluster_schema.py
chenwany: Changed

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1474,6 +1476,100 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+            data["SchedulerDefinition"] = yaml.safe_load(scheduler_definition)
+        except YAMLError as e:
+            raise ValidationError(
+                f"The retrieved SchedulerDefinition ({original_scheduler_definition}) is not a valid YAML."
+            ) from e
+        except Exception as e:
+            error_message = (
+                f"Error while downloading scheduler definition from {original_scheduler_definition}: {str(e)}"
+            )
+            raise ValidationError(error_message) from e
```
demartinofra: is this still needed?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1474,6 +1476,100 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+                key=bucket_parsing_result["object_key"],
+                expected_bucket_owner=s3_bucket_owner,
+            )
+            scheduler_definition = result["Body"].read().decode("utf-8")
+            return scheduler_definition
+        except AWSClientError as e:
+            if s3_bucket_owner and e.error_code == "AccessDenied":
+                error_message = (
+                    f"Error while downloading scheduler definition from {original_scheduler_definition}: "
+                    f"{str(e)} This can be due to bucket owner not matching the expected "
```
demartinofra: do we need a `.` before `This`?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1474,6 +1476,100 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+                    f"one '{s3_bucket_owner}'"
+                )
+            else:
+                error_message = (
+                    f"Error while downloading scheduler definition from {original_scheduler_definition}: {str(e)}"
+                )
```
demartinofra: Can be simplified to:
```
error_message = (
                    f"Error while downloading scheduler definition from {original_scheduler_definition}: {str(e)}."
                )
if s3_bucket_owner and e.error_code == "AccessDenied":
  error_message = f"{error_message}. This can be due to bucket owner no...."
```

scheduler_plugins/plugin_template/examples/cluster_configuration.yaml
```diff
@@ -10,6 +10,8 @@ Scheduling:
   Scheduler: plugin
   SchedulerSettings:
     SchedulerDefinition: <PLUGIN_DEFINITION>
+    SchedulerDefinitionS3BucketOwner: <ACCOUNT_ID>
+    SchedulerDefinitionChecksum: <CHECKSUM>
```
demartinofra: let's sync on how to have this auto-replaced

demartinofra (commented):
only a few minor comments then good to go

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1477,6 +1479,93 @@ class SchedulerPluginSettingsSchema(BaseSchema):
...
+            scheduler_definition = result["Body"].read().decode("utf-8")
+            return scheduler_definition
+        except AWSClientError as e:
+            error_message = (
+                f"Error while downloading scheduler definition from {original_scheduler_definition}: {str(e)}"
+            )
+            if s3_bucket_owner and e.error_code == "AccessDenied":
+                error_message = (
+                    f"{error_message}. This can be due to bucket owner not matching the expected "
+                    f"one '{s3_bucket_owner}"
```
demartinofra: missing `'`

chenwany: Good catch! Fixed, also add a test case to catch it

demartinofra (approved):

PR #3870, opened 2022-03-09
Support using custom template with S3 Block Public Access

hanwen-pcluste:
1. Relax URL checks in CLI and integration tests. When a S3 URL (e.g. https://...s3...) is provided, we use boto3 call to validate the URL. A simple https request can only validate the URL if the object is public while a boto3 call can validate the object as long as the CLI role has access to the object.
2. When uploading templates to S3 bucket, set the ACL to private. Therefore, the upload does not require a public bucket.

### Tests
I ran the following test
```
  configure:
    test_pcluster_configure.py::test_pcluster_configure:
      dimensions:
        - regions: [ "ap-northeast-2" ]
          oss: ["alinux2"]
          schedulers: [ "slurm" ]
```

### Checklist

### DIFF
--- a/cli/src/pcluster/config/validators.py
+++ b/cli/src/pcluster/config/validators.py
@@ -714,7 +714,12 @@ def url_validator(param_key, param_value, pcluster_config):
 
     else:
         try:
-            urllib.request.urlopen(param_value)  # nosec nosemgrep
+            match = re.match(r"https://(.*?)\.s3\.(.*?)amazonaws\.com(.*?)/(.*)", param_value)
+            if match and len(match.groups()) == 4:  # If S3 URL is provided.
+                bucket_name, object_name = match.group(1), match.group(4)
+                boto3.client("s3").head_object(Bucket=bucket_name, Key=object_name)
+            else:
+                urllib.request.urlopen(param_value)  # nosec nosemgrep
         except urllib.error.HTTPError as e:
             warnings.append("{0} {1} {2}".format(param_value, e.code, e.reason))
         except urllib.error.URLError as e:
--- a/tests/integration-tests/test_runner.py
+++ b/tests/integration-tests/test_runner.py
@@ -334,7 +334,7 @@ def _is_url(value):
     scheme = urllib.request.urlparse(value).scheme
     if scheme in ["https", "s3", "file"]:
         try:
-            if scheme == "s3":
+            if scheme == "s3":  # If S3 URI is provided
                 match = re.match(r"s3://(.*?)/(.*)", value)
                 if not match or len(match.groups()) < 2:
                     raise argparse.ArgumentTypeError(f"'{value}' is not a valid S3url")
@@ -342,7 +342,12 @@ def _is_url(value):
                     bucket_name, object_name = match.group(1), match.group(2)
                     boto3.client("s3").head_object(Bucket=bucket_name, Key=object_name)
             else:
-                urllib.request.urlopen(value)
+                match = re.match(r"https://(.*?)\.s3\.(.*?)amazonaws\.com(.*?)/(.*)", value)
+                if match and len(match.groups()) == 4:  # If S3 URL is provided.
+                    bucket_name, object_name = match.group(1), match.group(4)
+                    boto3.client("s3").head_object(Bucket=bucket_name, Key=object_name)
+                else:
+                    urllib.request.urlopen(value)
             return value
         except Exception as e:
             raise argparse.ArgumentTypeError(f"'{value}' is not a valid url:{e}")
--- a/util/upload-cfn-templates.py
+++ b/util/upload-cfn-templates.py
@@ -38,10 +38,13 @@ def get_template_extension(templates_dir, template_name):
     return ".cfn." + extension
 
 
-def put_object_to_s3(s3_client, bucket, key, region, data, template_name):
+def put_object_to_s3(s3_client, bucket, key, region, data, template_name, private):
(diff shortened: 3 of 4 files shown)


util/uploadTemplate.sh
```diff
@@ -108,8 +108,8 @@ main() {
     sed -i "s#.*aws-parallelcluster.*/templates/master-server-substack-\${version}.cfn.yaml.*#\"https://${_s3_folder_url}/master-server-substack.cfn.yaml\",#" ${_temp_dir}/aws-parallelcluster.cfn.json
 
     # upload templates
-    aws ${_profile} --region "${_region}" s3 cp --acl public-read ${_temp_dir}/aws-parallelcluster.cfn.json s3://${_bucket}/${_templates_folder}/ || _error_exit 'Failed to push cloudformation template to S3'
-    aws ${_profile} --region "${_region}" s3 cp --acl public-read --recursive --exclude "*" --include "*substack.cfn.json" --include "*substack.cfn.yaml" ${_srcdir}/cloudformation/ s3://${_bucket}/${_templates_folder}/ || _error_exit 'Failed to push substack cfn templates to S3'
+    aws ${_profile} --region "${_region}" s3 cp ${_temp_dir}/aws-parallelcluster.cfn.json s3://${_bucket}/${_templates_folder}/ || _error_exit 'Failed to push cloudformation template to S3'
+    aws ${_profile} --region "${_region}" s3 cp --recursive --exclude "*" --include "*substack.cfn.json" --include "*substack.cfn.yaml" ${_srcdir}/cloudformation/ s3://${_bucket}/${_templates_folder}/ || _error_exit 'Failed to push substack cfn templates to S3'
```
hanwen-pcluste, line 112: This script is only used when we manually upload the templates. It is not used in anywhere else, so remove the --acl is safe here

tests/integration-tests/test_runner.py
```diff
@@ -334,15 +334,20 @@ def _is_url(value):
...
+            if scheme == "s3":  # If S3 URI is provided
                 match = re.match(r"s3://(.*?)/(.*)", value)
                 if not match or len(match.groups()) < 2:
                     raise argparse.ArgumentTypeError(f"'{value}' is not a valid S3url")
                 else:
                     bucket_name, object_name = match.group(1), match.group(2)
                     boto3.client("s3").head_object(Bucket=bucket_name, Key=object_name)
             else:
-                urllib.request.urlopen(value)
+                match = re.match(r"https://(.*?)\.s3\.(.*?)amazonaws\.com/(.*)", value)
```
enrico-usai: This doesn't work with china partition, because in that case the suffix is: `.amazonaws.com.cn`.

util/upload-cfn-templates.py
```diff
@@ -38,10 +38,13 @@ def get_template_extension(templates_dir, template_name):
...
 
-def put_object_to_s3(s3_client, bucket, key, region, data, template_name):
+def put_object_to_s3(s3_client, bucket, key, region, data, template_name, private):
     try:
         object = s3_client.Object(bucket, key)
-        response = object.put(Body=data, ACL="public-read")
+        if private:
+            response = object.put(Body=data)
+        else:
+            response = object.put(Body=data, ACL="public-read")
```
enrico-usai: This can be simplified with:
```python
acl = None if private else "public-read"
response = object.put(Body=data, ACL=acl)
```

enrico-usai: or:
```python
acl = "private" if private else "public-read"
response = object.put(Body=data, ACL=acl)
```

util/upload-cfn-templates.py
```diff
@@ -54,7 +57,10 @@ def put_object_to_s3(s3_client, bucket, key, region, data, template_name):
             s3_client.BucketVersioning(bucket).enable()
             print("Created %s bucket. Bucket versioning is enabled, " "please enable bucket logging manually." % bucket)
             b = s3_client.Bucket(bucket)
-            res = b.put_object(Body=data, ACL="public-read", Key=key)
+            if private:
```
enrico-usai: same here. An alternative is:
```python
put_object_args = {"Body": body, "Key": key}
if not private:
    put_object_args["ACL"] = "public-read"
res = b.put_object(**put_object_args)
```

cli/src/pcluster/config/validators.py
```diff
@@ -714,7 +714,12 @@ def url_validator(param_key, param_value, pcluster_config):
 
     else:
         try:
-            urllib.request.urlopen(param_value)  # nosec nosemgrep
+            match = re.match(r"https://(.*?)\.s3\.(.*?)amazonaws\.com/(.*)", param_value)
```
enrico-usai: not supported by china partition: amazonaws.com.cn

tests/integration-tests/test_runner.py
hanwen-pcluste: Done

util/upload-cfn-templates.py
hanwen-pcluste: ACL can not be `None`. So I went for your last alternative

util/upload-cfn-templates.py
hanwen-pcluste: Done

cli/src/pcluster/config/validators.py
hanwen-pcluste: Done

enrico-usai (approved):

PR #3871, opened 2022-03-10
Integration Tests: cover consistent creation of home directory and SSH key at SSH login and switch user.

gmarciani:
**HIGHLIGHT: Please, do not merge before [this cookbook PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/1385)** because the PR depends on that one.

### Description of changes
1. Make AD integration tests to cover the consistent creation of home directory and SSH key at SSH login and switch user.

#### Notes
1. In this PR I left commented out the case with `sudo -u` and `sudo -u -i`. I've verified manually on the same cluster used by the tests that they actually succeed. I will further investigate on this before merging the PR.

### Tests
1. Tests succeeded when launched locally making the cluster configuration used the cookbook with the required changes.
2. Kitchen tests
3. Integration tests

### DIFF
--- a/tests/integration-tests/tests/ad_integration/test_ad_integration.py
+++ b/tests/integration-tests/tests/ad_integration/test_ad_integration.py
@@ -26,6 +26,7 @@ from assertpy import assert_that
 from cfn_stacks_factory import CfnStack
 from OpenSSL import crypto
 from OpenSSL.crypto import FILETYPE_PEM, TYPE_RSA, X509, dump_certificate, dump_privatekey
+from paramiko import RSAKey
 from remote_command_executor import RemoteCommandExecutor
 from retrying import retry
 from time_utils import seconds
@@ -467,7 +468,7 @@ def _check_files_permissions(users):
         _check_failed_result_for_permission_denied(result)
         previous_user = users[index - 1]
         for path in [
-            f"/home/{user.alias}/my_file",
+            f"{user.home_dir}/my_file",
             f"/shared/{user.alias}_file",
             f"/ebs/{user.alias}_file",
             f"/efs/{user.alias}_file",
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -486,28 +487,134 @@ def _check_failed_result_for_permission_denied(result):
...
+    user.ssh_connect()
+    _check_home_directory(user, remote_command_executor)
+    _check_ssh_key(user, generate_ssh_keys_for_user, remote_command_executor, scheduler_commands)
+    logging.info(
+        "Verified SSH key generation for user %s on SSH login (expected generation: %s)",
+        user.alias,
+        generate_ssh_keys_for_user,
+    )
+
+    ssh_key_path = f"/home/{user.alias}/.ssh/id_rsa"
```
francesco-giordano: Is it correct to use this path hardcoded? It is because in the tests we configure this value explicitly?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -486,28 +487,134 @@ def _check_failed_result_for_permission_denied(result):
...
+    result = remote_command_executor.run_remote_command(check_ownership)
+    assert_that(result.failed).is_false()
+    assert_that(result.stdout.strip()).is_equal_to(user.alias)
+    logging.info("Verified home directory for user %s", user.alias)
+
+
+def _check_ssh_key(user, ssh_generation_enabled, remote_command_executor, scheduler_commands):
+    logging.info("Checking SSH key for user %s (expected to exist: %s)", user.alias, ssh_generation_enabled)
+
+    ssh_key_path = f"/home/{user.alias}/.ssh/id_rsa"
```
francesco-giordano: Same as above

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -486,28 +487,134 @@ def _check_failed_result_for_permission_denied(result):
...
+
+    # Check existence
+    check_existence = f"sudo ls {ssh_key_path}"
+    result = remote_command_executor.run_remote_command(check_existence, raise_on_error=ssh_generation_enabled)
+    assert_that(result.failed).is_equal_to(not ssh_generation_enabled)
```
francesco-giordano, line 574: More comment could better explain this.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -604,7 +710,7 @@ def test_ad_integration(
     scheduler_commands = scheduler_commands_factory(remote_command_executor)
     assert_that(NUM_USERS_TO_TEST).is_less_than_or_equal_to(NUM_USERS_TO_CREATE)
     users = []
-    for user_num in range(1, NUM_USERS_TO_TEST + 1):
+    for user_num in range(1, NUM_USERS_TO_TEST):
```
francesco-giordano: Why do the range start from 1?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -486,28 +487,134 @@ def _check_failed_result_for_permission_denied(result):
...
+        result = remote_command_executor.run_remote_command(check_permissions)
+        assert_that(result.failed).is_false()
+        assert_that(result.stdout.strip()).is_equal_to(f"{user.alias} 600")
+        logging.info(
+            "Verified SSH key permissions for user %s (expected to exist: %s)",
+            user.alias,
+            ssh_generation_enabled,
+        )
+
+        # Check SSH log in with SSH key to head node and to a static compute nodes
```
francesco-giordano: log in -> login

francesco-giordano:
I would suggest to add more comment in the main function `test_ad_integration` to better explain the overall test.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: In our integration tests we assume users to have the standard home directory, so it is correct to have it hardcoded. Btw, we could improve the code by including it as a user class attribute. Will do in the next iter

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: Same as above

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: Good catch! It is a change I've introduced during debugging, will remove it. Thanks a lot!

gmarciani: I think it was like that just to have user names starting with 1 rather than 0.
I will check if it needed for some other reasons. Otherwise, I will make it start from 0.

gmarciani: Seems there is no need to start the numeration that way. Will change it in the next rev. Please, consider that such change will imply the change on the ldap access filter in the update cluster config from Pcluster3 to Pcluster2.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
francesco-giordano: Perfect

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: Added

francesco-giordano (dismissed):
LGTM

francesco-giordano (dismissed):

francesco-giordano (approved):

PR #3874, opened 2022-03-11
Support for environments without a default VPC

maclema:
Allows passing a VPC and subnet ID to be used for the image build infrastructure to allow for launching in an environment where no default VPC exists.

Of note, the subnet passed in as a parameter should have internet access (i.e private subnet with NAT gateway routes, or a public subnet). When a public subnet is used, it must have MapPublicIpOnLaunch enabled as image builder won't automatically map a public IP address and the image build will timeout.

### Tests
Manually launched parallel cluster (via parallel cluster manager) stacks using a stack referencing stack set templates with these updates which exist in my personal AWS account S3 bucket.


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -964,6 +975,13 @@ Resources:
       Roles:
         - !Ref ImageBuilderInstanceRole
 
+  InfrastructureConfigurationSecurityGroup:
+    Condition: NonDefaultVpc
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      VpcId: !Ref ImageBuilderVpcId
+      GroupDescription: Parallel cluster image builder security group
```
enrico-usai: Parallel cluster --> ParallelCluster :)

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -74,6 +74,14 @@ Parameters:
       - true
       - false
 
+  ImageBuilderVpcId:
+    Description: Optional. Provide a specific vpc id to use for building the container images. Use this if you don't have a default vpc available.
+    Type: String
+
+  ImageBuilderSubnetId:
+    Description: Optional. Provide a specific subnet id to use for building the container images. Use this if you don't have a default vpc available.
+    Type: String
```
demartinofra: Could you make the `Default: ''` explicit for these 2 new params?

demartinofra (commented):
Thanks for adding this support.

Could you please add a line to the changelog, something like:

```
**ENHANCEMENTS**
- Add support to deploy API infrastructure in environments without a default VPC.
```

maclema:
Added defaults and changelog entry.

enrico-usai (approved):
Thanks @maclema for the improvement!

PR #3879, opened 2022-03-14
Add validator for property DirectoryService.DomainName

gmarciani:
**WARNING: This PR depends on [PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/1391)**

### Description of changes
Add validator for property `DirectoryService.DomainName` that accepts only a value in FQDN or LDAP DN format. The validator fails the validation with an error, otherwise.

Added unit tests to cover the validation.

Examples:
```sh
# FQDN
DirectoryService:
  DomainName: corp.mgiacomo.com

# LDAP DN
DirectoryService:
  DomainName: DC=corp,DC=mgiacomo,DC=com
```

### Tests
1. Unit testing
2. Build&Test in Jenkins

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -9,7 +9,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import re
 from urllib.parse import urlparse
 
 from pcluster.validators.common import FailureLevel, Validator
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -48,6 +48,25 @@ def _validate(self, domain_addr, additional_sssd_configs):
             self._add_failure(warning_message, FailureLevel.WARNING)
 
 
+class DomainNameValidator(Validator):
+    """Domain name validator."""
+
+    FQDN_PATTERN = "^([a-zA-Z0-9_-]+)(\\.[a-zA-Z0-9_-]+)*$"
```
francesco-giordano, line 54: This accept ip addresses, is it intended to do so?

francesco-giordano (commented):
I do not know if these would fit but regarding regex these can be 2 links worth checking:

https://regexr.com/3l4au
https://regexr.com/3g5j0

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: IP addresses should not be accepted.
Will iterate on the above regexp

gmarciani: The regexp you suggested regarding LDAP DN does not fit our case because:
1. it is a full LDAP DN, whereas we need here only the domain part
2. it accepts strings like `DC=corp,DC=mgiac/mo,DC=com`, which cannot be accepted

gmarciani: I will take the FQDN regexp you suggested as it fits our case

gmarciani: The proposed regexp excluded the possibility of underscores in DNS name, which have been excluded by recent RFC, but I think we should be as much flexible as possible and accepts also old format.

gmarciani: The regexp to match FQDN and LDAP DN in a strict way are so complex that I consider them hardly maintainable.
I suggest to use a more open, but maintainable regexp as it would have a better ROI.

francesco-giordano (dismissed):

francesco-giordano (approved):

PR #3905, opened 2022-03-22
Improve retrieve of custom template with S3 URI

hanwen-pcluste:
The GetObject call of S3 can retrieve object across regions within a partition. In addition, there is one limitation in commercial partition, where the call does not work to get an object in optin region from non optin region or vice versa

### DIFF
--- a/cli/src/pcluster/utils.py
+++ b/cli/src/pcluster/utils.py
@@ -1076,7 +1076,25 @@ def read_remote_file(url):
         if urlparse(url).scheme == "s3":
             match = re.match(r"s3://(.*?)/(.*)", url)
             bucket, key = match.group(1), match.group(2)
-            file_contents = boto3.resource("s3").Object(bucket, key).get()["Body"].read().decode("utf-8")
+            try:
+                if get_partition() == "aws":
+                    s3_resource = boto3.resource("s3", region_name="us-east-1")
+                else:
+                    s3_resource = boto3.resource("s3")
+                file_contents = s3_resource.Object(bucket, key).get()["Body"].read().decode("utf-8")
+            except ClientError as e:
+                error = e.response["Error"]
+                if error["Code"] == "IllegalLocationConstraintException":
+                    # When S3 GetObject call is made in a different Region from where the bucket exists
+                    # (e.g. bucket in an optin region while this code is in us-east-1),
+                    # make the call again by using the region from the error message, that is in the form:
+                    # "The af-south-1 location constraint is incompatible
+                    # for the Region specific endpoint this request was sent to.".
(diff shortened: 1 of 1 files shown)


enrico-usai:
Could you please add a more complete description of the change in the commit message and in PR description?
I mean why we are changing this and what problem we're solving.

cli/src/pcluster/utils.py
```diff
@@ -1076,7 +1076,28 @@ def read_remote_file(url):
         if urlparse(url).scheme == "s3":
             match = re.match(r"s3://(.*?)/(.*)", url)
             bucket, key = match.group(1), match.group(2)
-            file_contents = boto3.resource("s3").Object(bucket, key).get()["Body"].read().decode("utf-8")
+            try:
+                file_contents = (
+                    boto3.resource("s3", region_name="us-east-1")
```
enrico-usai, line 1081: We should move this call in the `pcluster/aws/s3_resource.py` module, in the `S3Resource` class.

cli/src/pcluster/utils.py
```diff
@@ -1076,7 +1076,21 @@ def read_remote_file(url):
...
+            try:
+                if get_partition() == "aws":
+                    s3_resource = boto3.resource("s3", region_name="us-east-1")
+                else:
+                    s3_resource = boto3.resource("s3")
+                file_contents = s3_resource.Object(bucket, key).get()["Body"].read().decode("utf-8")
+            except ClientError as e:
+                error = e.response["Error"]
+                if error["Code"] == "IllegalLocationConstraintException":
+                    # Try to make the call again if the object is in opt in region.
```
enrico-usai: I'd improve this comment by saying:

> We were trying to access the bucket from a different Region than where the bucket exists (e.g.  bucket in optin region from us-east-1). 
Make the call again by using the region from the error message, that is in the form: "The <region> location constraint is incompatible for the Region specific endpoint this request was sent to.".

cli/src/pcluster/utils.py
```diff
@@ -1076,7 +1076,21 @@ def read_remote_file(url):
...
+                    match = re.match(r"The (.*) location constraint", error["Message"])
```
enrico-usai, line 1093: note: the matching word might be `unspecified` or `EU`  

See: 
* https://docs.aws.amazon.com/code-samples/latest/catalog/cpp-s3-create_bucket.cpp.html
* https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html#API_GetBucketLocation_ResponseSyntax

enrico-usai (approved):

PR #3961, opened 2022-04-12
Add validation of DirectoryService/AdditionalSssdConfigs: emitting a failure when the customer is overriding protected properties

gmarciani:
### Description of changes
1. Added a validator for `DirectoryService/AdditionalSssdConfigs` that fails when additional properties overrides protected properties with unacceptable values. In particular the validation must fail in the following cases:
  1. `id_provider != ldap`
  2. `access_provider != ldap` when a `DirectoryService/LdapAccessFilter` is specified 

Notice that we decided to not validate the property `ldap_schema` in order to provide customer with highest flexibility in case of on-prem customizations.

### Tests
1. Manually tested cluster creation and update triggering some critical configurations leveraging AdditionalSssdConfigs (e.g. switching from LdapAccessFilter to a simple access filter)
4. Integ test (SimpleAD/MsAD with LDAPS) locally launched
5. [ONGOING] Build&Test

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -12,6 +12,7 @@
 import re
 from urllib.parse import urlparse
 
+from pcluster.constants import DIRECTORY_SERVICE_RESERVED_SETTINGS
 from pcluster.validators.common import FailureLevel, Validator
 
 
@@ -78,3 +79,32 @@ class LdapTlsReqCertValidator(Validator):
                 f"For security reasons it's recommended to use {' or '.join(values_requiring_cert_validation)}",
                 FailureLevel.WARNING,
             )
+
+
+class AdditionalSssdConfigsValidator(Validator):
+    """AdditionalSssdConfigs validator."""
+
+    def _validate(self, additional_sssd_configs, ldap_access_filter):
+        """Validate that AdditionalSssdConfigs does not introduce unacceptable values."""
+        for config_key, accepted_value in DIRECTORY_SERVICE_RESERVED_SETTINGS.items():
+            if config_key in additional_sssd_configs:
+                actual_value = additional_sssd_configs[config_key]
+                if actual_value != accepted_value:
+                    self._add_failure(
+                        f"Cannot override the SSSD property '{config_key}' "
+                        f"with value '{actual_value}'. "
+                        f"Allowed value is: '{accepted_value}'. "
+                        "Please refer to ParallelCluster official documentation for more information.",
+                        FailureLevel.ERROR,
+                    )
+
+        if "access_provider" in additional_sssd_configs:
+            actual_access_provider = additional_sssd_configs["access_provider"]
+            if ldap_access_filter is not None and actual_access_provider != "ldap":
+                self._add_failure(
+                    "Cannot override the SSSD property 'access_provider' "
+                    f"with value '{actual_access_provider}' when LdapAccessFilter is specified. "
+                    "Allowed value is: 'ldap'. "
+                    "Please refer to ParallelCluster official documentation for more information.",
+                    FailureLevel.ERROR,
+                )
--- a/cli/tests/pcluster/validators/test_directory_service_validators.py
+++ b/cli/tests/pcluster/validators/test_directory_service_validators.py
@@ -11,6 +11,7 @@
 import pytest
 
 from pcluster.validators.directory_service_validators import (
+    AdditionalSssdConfigsValidator,
     DomainAddrValidator,
     DomainNameValidator,
     LdapTlsReqCertValidator,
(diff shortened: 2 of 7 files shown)


cli/tests/pcluster/validators/test_directory_service_validators.py
```diff
@@ -108,3 +109,55 @@ def test_domain_addr_protocol(domain_addr, additional_sssd_configs, expected_mes
...
+            "Allowed value is: 'ldap'.",
+        ),
+    ],
+)
+def test_additional_sssd_configs(additional_sssd_configs, ldap_access_filter, expected_message):
+    actual_failures = AdditionalSssdConfigsValidator().execute(
+        additional_sssd_configs=additional_sssd_configs, ldap_access_filter=ldap_access_filter
+    )
+    for f in actual_failures:
+        print(f"xxxx {f.message}")
```
gmarciani: to be removed

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -78,3 +78,39 @@ def _validate(self, ldap_tls_reqcert):
...
+class AdditionalSssdConfigsValidator(Validator):
+    """AdditionalSssdConfigs validator."""
+
+    def _validate(self, additional_sssd_configs, ldap_access_filter):
+        """Validate that AdditionalSssdConfigs does not introduce unacceptable values."""
+        if "ldap_schema" in additional_sssd_configs:
+            value = additional_sssd_configs["ldap_schema"]
+            if value != "AD":
+                self._add_failure(
+                    f"Cannot override the SSSD property 'ldap_schema' "
```
enrico-usai: f string is not needed if there are no variables in this part of the string

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -78,3 +78,39 @@ def _validate(self, ldap_tls_reqcert):
...
             )
+
+
+class AdditionalSssdConfigsValidator(Validator):
+    """AdditionalSssdConfigs validator."""
+
+    def _validate(self, additional_sssd_configs, ldap_access_filter):
+        """Validate that AdditionalSssdConfigs does not introduce unacceptable values."""
+        if "ldap_schema" in additional_sssd_configs:
+            value = additional_sssd_configs["ldap_schema"]
```
enrico-usai: minor: I'd use a more explicative term other than "value". 
Even because we're using the same `value`word in the next case and it can hide issues if we miss to set the right "value" to this variable.

I'd use `ldap_schema` or `actual_ldap_schema`.

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -78,3 +78,39 @@ def _validate(self, ldap_tls_reqcert):
...
+            if value != "AD":
+                self._add_failure(
+                    f"Cannot override the SSSD property 'ldap_schema' "
+                    f"with value '{value}'. "
+                    f"Allowed value is: 'AD'.",
+                    FailureLevel.ERROR,
+                )
+
+        if "id_provider" in additional_sssd_configs:
+            value = additional_sssd_configs["id_provider"]
```
enrico-usai: same here for the value naming.

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -78,3 +78,39 @@ def _validate(self, ldap_tls_reqcert):
...
+            if value != "ldap":
+                self._add_failure(
```
enrico-usai: the first 2 checks are executing the same code and have the same error in the same form.
What about doing something like:
```python
for config_key, accepted_value in ("ldap_schema", "AD"), ("id_provider", "ldap"):
...
```

enrico-usai: Or we can also define a RESERVED_AD_SETTINGS dict containing config_key and accepted value in constants.py module, similar to what you added in the cookbook.

enrico-usai:
Should we validate if they are inserting `ldap_access_filter` but passing a `access_provider` other than `ldap`?

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Fixed in the next rev

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Fixed in the next rev

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Fixed in the next rev

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: I agree with the 2nd approach

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -78,3 +79,31 @@ def _validate(self, ldap_tls_reqcert):
...
             )
+
+
+class AdditionalSssdConfigsValidator(Validator):
+    """AdditionalSssdConfigs validator."""
+
+    def _validate(self, additional_sssd_configs, ldap_access_filter):
+        """Validate that AdditionalSssdConfigs does not introduce unacceptable values."""
+        for config_key, accepted_value in DIRECTORY_SERVICE_RESERVED_SETTINGS.items():
+            print(f"xxxxx {config_key} = {accepted_value}")
```
enrico-usai: to be removed

enrico-usai (dismissed):

enrico-usai (dismissed):

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update2.yaml
```diff
@@ -0,0 +1,59 @@
...
+    debug_level: "0x1ff"
+    {% if directory_protocol == "ldap" %}
+    ldap_auth_disable_tls_never_use_in_production: True
+    {% endif %}
+    access_provider: simple
+    simple_deny_users: PclusterUser0
+DevSettings:
+  AmiSearchFilters:
+    Owner: "447714826191"
+  Cookbook:
```
gmarciani: To remove

enrico-usai (approved):

enrico-usai (approved):

PR #3973, opened 2022-04-15
[ad integration] Fix integration tests: using the correct variable for head node instance type and setting max count equal to the first update.

gmarciani:
### Description of changes
Fix integration tests: using the correct variable for head node instance type and setting max count equal to the first update.
The correct variable for head node instance type is `instance`: the bug was introduced by a recent cherry-pick from develop, where the variable has different name.
The correct value for max compute count is `3000`, which may seem pretty high, but it must be aligned to the first update.

### Tests
1. Launched test locally


hanwen-pcluste (approved):
LGTM

PR #3981, opened 2022-04-20
Add validator to prevent using reserved directories as mount dirs

hanwen-pcluste:
Add more detailed check considering the difference between shared storage and local storage:

1. Shared storage directories can not overlap/duplicate with each other.
2. Shared storage directories can not overlap/duplicate with ephemeral storage directories.
3. Ephemeral storage directories can overlap/duplicate with each other, because they are local to compute nodes.

Note this PR always reserves`/scratch` for ephemeral storage, forbidding shared storage from using it. Although only some of the instance types have ephemeral storage. This universal ban will facilitate instance update functionality in the future.

Signed-off-by: Hanwen <hanwenli@amazon.com>

### Checklist
- [ ] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [ ] Check all commits' messages are clear, describing what and why vs how.
- [ ] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [ ] Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3999, opened 2022-04-27
Fix bug in shared mount directory list

hanwen-pcluste:
The list should only contain directories from shared storage. There is another property `local_mount_dir_set` to collect the set of local mount directories.

### References
This PR fixes a bug in https://github.com/aws/aws-parallelcluster/pull/3981

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1206,10 +1206,6 @@ class BaseClusterConfig(Resource):
         if self.shared_storage:
             for storage in self.shared_storage:
                 mount_dir_list.append(storage.mount_dir)
-
-        if self.head_node.local_storage.ephemeral_volume:
-            mount_dir_list.append(self.head_node.local_storage.ephemeral_volume.mount_dir)
-
         return mount_dir_list
 
     @property
--- a/cli/tests/pcluster/validators/test_all_validators.py
+++ b/cli/tests/pcluster/validators/test_all_validators.py
@@ -247,8 +247,12 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
         [call(kms_key_id="1234abcd-12ab-34cd-56ef-1234567890ab", encrypted=True)]
     )
     fsx_architecture_os_validator.assert_has_calls([call(architecture="x86_64", os="alinux2")])
-    duplicate_mount_dir_validator.assert_has_calls(
-        [call(mount_dir_list=["/my/mount/point1", "/my/mount/point2", "/my/mount/point3", "/scratch"])]
+    # Scratch mount directories are retrieved from a set. So the order of them is not guaranteed.
+    # The first item in call_args is regular args, the second item is keyword args.
+    mount_dir_list = duplicate_mount_dir_validator.call_args[1]["mount_dir_list"]
+    mount_dir_list.sort()
+    assert_that(mount_dir_list).is_equal_to(
+        ["/my/mount/point1", "/my/mount/point2", "/my/mount/point3", "/scratch", "/scratch_head"]
     )
     number_of_storage_validator.assert_has_calls(
         [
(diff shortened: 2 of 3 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1206,10 +1206,6 @@ def shared_mount_dir_list(self):
         if self.shared_storage:
             for storage in self.shared_storage:
                 mount_dir_list.append(storage.mount_dir)
-
-        if self.head_node.local_storage.ephemeral_volume:
-            mount_dir_list.append(self.head_node.local_storage.ephemeral_volume.mount_dir)
-
         return mount_dir_list
 
     @property
```
hanwen-pcluste, line 1211: The property below contains the mount dir of ephemeral volume of head node. So the old code count the mount dir of head node ephemeral volume twice, generating the duplicate mount dir error

chenwany:
Shall we add a test case in test_all_validators to catch such issue?

hanwen-pcluste (commented):
Good idea. I added a check there

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -247,9 +247,11 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
         [call(kms_key_id="1234abcd-12ab-34cd-56ef-1234567890ab", encrypted=True)]
     )
     fsx_architecture_os_validator.assert_has_calls([call(architecture="x86_64", os="alinux2")])
-    duplicate_mount_dir_validator.assert_has_calls(
-        [call(mount_dir_list=["/my/mount/point1", "/my/mount/point2", "/my/mount/point3", "/scratch"])]
-    )
+    mount_dir_list = duplicate_mount_dir_validator.call_args.kwargs["mount_dir_list"]
+    expected_mount_dir_list = ["/my/mount/point1", "/my/mount/point2", "/my/mount/point3", "/scratch", "/scratch_head"]
+    assert_that(mount_dir_list).is_length(len(expected_mount_dir_list))
```
hanwen-pcluste: Since the ephemeral mount dirs are from a set. The order of set is not consistent. So we need to check the items in the list one by one

chenwany (dismissed):
LGTM!

cli/src/pcluster/config/cluster_config.py
enrico-usai: This is a good explanation to add in the PR description. This explains what is the error seen by the user without your patch.

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -247,9 +247,36 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
         [call(kms_key_id="1234abcd-12ab-34cd-56ef-1234567890ab", encrypted=True)]
     )
     fsx_architecture_os_validator.assert_has_calls([call(architecture="x86_64", os="alinux2")])
-    duplicate_mount_dir_validator.assert_has_calls(
-        [call(mount_dir_list=["/my/mount/point1", "/my/mount/point2", "/my/mount/point3", "/scratch"])]
-    )
+    try:
+        # Scratch mount directories are retrieved from a set. So the order of them is not guaranteed.
+        # Using try block to check one of the call was made
+        duplicate_mount_dir_validator.assert_has_calls(
```
enrico-usai: You could instead retrieve the args used to call the validator, retrieve the sorted mount_dir_list and compare with an ordered list.

Something like:
```python
mount_dir_call_args = duplicate_mount_dir_validator.call_args_list.kwargs.get("mount_dir_list").sort()
assert_that(mount_dir_call_args).is_equals_to(your-ordered-list)
```

Ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args

hanwen-pcluste: Retrieving call args has some incompatibility issue across on python 3.7

The kwargs was added in python 3.8

enrico-usai: I think you can retrieve `call_args` that is a tuple and then retrieve the N value of the tuple, in this case the only value there. kwargs should be just an extension we can skip

hanwen-pcluste: Done

enrico-usai (approved):

chenwany (approved):

PR #4003, opened 2022-04-28
Add support for LDAPS to the 1-click AD integration template

gmarciani:
### Description of changes
Add support for LDAPS to the 1-click AD integration template.

#### Notes
With this change I've also added the policy `DomainCertificateSecretReadPolicy`. Such policy is not required for the integration to work with LDAPS, but it can be very handy for our customers following the tutorial because they can simply attach that policy to the head node to make it able to retrieve the certificate from Secrets Manager within a bootstrap script.

### Tests
1. Launched the 1-click template on my personal account;
2. Tested the Ad integration, both using LDAP and LDAPS (both with/without cert verification);
3. Deleted the stack and verified that all related resources have been deleted (the stack contains custom resources so this must be checked);

### DIFF
--- a/cloudformation/ad/ad-integration.yaml
+++ b/cloudformation/ad/ad-integration.yaml
@@ -192,8 +192,7 @@ Resources:
   PrepLambda:
     Type: AWS::Lambda::Function
     Properties:
-      Description: Lambda function to prepare stack
-      FunctionName: !Sub ${AWS::StackName}-Prep
+      Description: !Sub "${AWS::StackName}: custom resource handler to prepare the stack."
       Handler: index.handler
       MemorySize: 128
       Role: !GetAtt PrepRole.Arn
@@ -317,7 +316,16 @@ Resources:
                   - arn:${AWS::Partition}:ds:${AWS::Region}:${AWS::AccountId}:directory/${DirectoryId}
                   # - { DirectoryId: !If [CreateAD, !Ref Directory, !Ref Ad ] }
                   - { DirectoryId: !Ref Directory }
-          PolicyName: !Sub ${AWS::StackName}resetuserpassword
+          PolicyName: ResetUserPassword
+        - PolicyDocument:
+            Statement:
+              - Action:
+                  - secretsmanager:PutSecretValue
+                Effect: Allow
+                Resource:
+                  - !Ref DomainCertificateSecret
+                  - !Ref DomainPrivateKeySecret
+          PolicyName: PutDomainCertificateSecrets
   JoinProfile:
     Type: AWS::IAM::InstanceProfile
     Properties:
@@ -347,6 +355,7 @@ Resources:
               krb5-workstation: []
               openldap-clients: []
               policycoreutils-python: []
+              openssl: []
     Properties:
       IamInstanceProfile:
         Ref: JoinProfile
@@ -356,6 +365,9 @@ Resources:
       SecurityGroupIds:
         - Ref: AdDomainAdminNodeSecurityGroup
       SubnetId: !If [CreateVPC, !Ref PclusterVpcSubnet1, !Ref PrivateSubnetOne ]
+      Tags:
+        - Key: "Name"
+          Value: !Sub [ "AdDomainAdminNode-${StackIdSuffix}", {StackIdSuffix: !Select [1, !Split ['/', !Ref 'AWS::StackId']]}]
       UserData:
         Fn::Base64:
           !Sub
@@ -365,6 +377,8 @@ Resources:
               yum update -y aws-cfn-bootstrap
               /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource AdDomainAdminNode --configsets setup --region ${AWS::Region}
               echo "Domain Name: " ${DirectoryDomain}
+              echo "Domain Certificate Secret: " ${DomainCertificateSecretArn}
+              echo "Domain Private Key Secret: " ${DomainPrivateKeySecretArn}
               cat << EOF > /etc/resolv.conf
               search           ${DirectoryDomain}
               nameserver       ${DnsIp1}
@@ -380,13 +394,31 @@ Resources:
               sleep 0.5
               echo "Registering User..."
               echo $ADMIN_PW | adcli create-user -x -U Admin --domain=${DirectoryDomain} --display-name=${UserName} ${UserName}
+
+              echo "Creating domain certificate..."
+              PRIVATE_KEY="${DirectoryDomain}.key"
+              CERTIFICATE="${DirectoryDomain}.crt"
+              printf ".\n.\n.\n.\n.\n${DirectoryDomain}\n.\n" | openssl req -x509 -sha256 -nodes -newkey rsa:2048 -keyout $PRIVATE_KEY -days 365 -out $CERTIFICATE
+
+              echo "Storing domain private key to Secrets Manager..."
+              aws secretsmanager put-secret-value --secret-id ${DomainPrivateKeySecretArn} --secret-string file://$PRIVATE_KEY --region ${AWS::Region}
+
+              echo "Storing domain certificate to Secrets Manager..."
+              aws secretsmanager put-secret-value --secret-id ${DomainCertificateSecretArn} --secret-string file://$CERTIFICATE --region ${AWS::Region}
+
+              echo "Deleting private key and certificate from local file system..."
+              rm -rf $PRIVATE_KEY $CERTIFICATE
+
               /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource AdDomainAdminNode --region ${AWS::Region}
 
             - { DirectoryDomain: !GetAtt Prep.DomainName,
                 AdminPassword: !Ref AdminPassword,
                 UserName: !Ref UserName,
                 DnsIp1: !GetAtt Prep.DnsIpAddress1,
-                DnsIp2: !GetAtt Prep.DnsIpAddress2}
+                DnsIp2: !GetAtt Prep.DnsIpAddress2,
+                DomainCertificateSecretArn: !Ref DomainCertificateSecret,
+                DomainPrivateKeySecretArn: !Ref DomainPrivateKeySecret,
+            }
 
   PostRole:
     Type: AWS::IAM::Role
@@ -410,7 +442,6 @@ Resources:
                 - logs:CreateLogStream
                 - logs:PutLogEvents
                 Effect: Allow
-                # Resource: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${AWS::StackName}-Prep:*
                 Resource: '*'
         - PolicyName: ResetPassword
           PolicyDocument:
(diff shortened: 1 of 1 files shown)


cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,237 @@ Resources:
...
+                  logger.info('Importing certificate into ACM...')
+                  certificate_arn = acm.import_certificate(Certificate=domain_certificate, PrivateKey=domain_private_key)["CertificateArn"]
+                  logger.info('Storing ACM certificate ARN into Secrets Manager...')
+                  sm.put_secret_value(SecretId=acm_certificate_secret_arn, SecretString=certificate_arn)
+                  response_data['DomainCertificateArn'] = certificate_arn
+                elif event['RequestType'] == 'Delete':
+                  physical_resource_id = event['PhysicalResourceId']
+                  certificate_arn = sm.get_secret_value(SecretId=acm_certificate_secret_arn)["SecretString"]
+                  logger.info('Deleting ACM certificate {}...'.format(certificate_arn))
+                  acm.delete_certificate(CertificateArn=certificate_arn)
```
gmarciani: This call may fail due to ACM certificate still being in use. We must implement a retry

cloudformation/ad/ad-integration.yaml
```diff
@@ -318,6 +318,15 @@ Resources:
...
           PolicyName: !Sub ${AWS::StackName}resetuserpassword
+        - PolicyDocument:
+            Statement:
+              - Action:
+                  - secretsmanager:PutSecretValue
+                Effect: Allow
+                Resource:
+                  - !Ref DomainCertificateSecret
+                  - !Ref DomainPrivateKeySecret
+          PolicyName: !Sub ${AWS::StackName}-PutDomainCertificateSecrets
```
demartinofra: the limit on StackName is 128 chars which matches the limit on policy name, this means that if a user picks a stack name that can't fit "-PutDomainCertificateSecrets" within the 128 chars the stack creation fails. Low probability but worth mentioning.

cloudformation/ad/ad-integration.yaml
```diff
@@ -380,13 +395,28 @@ Resources:
...
               echo "Registering User..."
               echo $ADMIN_PW | adcli create-user -x -U Admin --domain=${DirectoryDomain} --display-name=${UserName} ${UserName}
+
+              echo "Creating domain certificate..."
+              PRIVATE_KEY="${DirectoryDomain}.key"
+              CERTIFICATE="${DirectoryDomain}.crt"
+              printf ".\n.\n.\n.\n.\n${DirectoryDomain}\n.\n" | openssl req -x509 -sha256 -nodes -newkey rsa:2048 -keyout $PRIVATE_KEY -days 90 -out $CERTIFICATE
+
+              echo "Storing domain private key to Secrets Manager..."
+              aws secretsmanager put-secret-value --secret-id ${DomainPrivateKeySecretArn} --secret-string "$(cat $PRIVATE_KEY)" --region ${AWS::Region}
```
demartinofra: is there the possibility to pass a file reference and avoid printing the file with cat? otherwise if tracing is enabled or in other cases we risk to dump the secrets in plain text

cloudformation/ad/ad-integration.yaml
```diff
@@ -380,13 +395,28 @@ Resources:
               sleep 0.5
               echo "Registering User..."
               echo $ADMIN_PW | adcli create-user -x -U Admin --domain=${DirectoryDomain} --display-name=${UserName} ${UserName}
+
+              echo "Creating domain certificate..."
+              PRIVATE_KEY="${DirectoryDomain}.key"
+              CERTIFICATE="${DirectoryDomain}.crt"
+              printf ".\n.\n.\n.\n.\n${DirectoryDomain}\n.\n" | openssl req -x509 -sha256 -nodes -newkey rsa:2048 -keyout $PRIVATE_KEY -days 90 -out $CERTIFICATE
```
demartinofra: out of curiosity why 90 days?

cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,256 @@ Resources:
...
+        - PolicyName: LogOutput
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Action:
+                  - logs:CreateLogGroup
+                  - logs:CreateLogStream
+                  - logs:PutLogEvents
+                Effect: Allow
+                # Resource: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${AWS::StackName}-Prep:*
```
demartinofra: did you intend to keep this comment?

cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,256 @@ Resources:
...
+                Resource: '*'
+        - PolicyName: ManageDomainCertificate
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - acm:ImportCertificate
+                  - acm:DeleteCertificate
+                Resource: !Sub arn:${AWS::Partition}:acm:${AWS::Region}:${AWS::AccountId}:certificate/*
```
demartinofra, line 648: This is quite permissive

cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,256 @@ Resources:
...
+          sm = boto3.client("secretsmanager")
+
+          def create_physical_resource_id():
+              alnum = string.ascii_uppercase + string.ascii_lowercase + string.digits
+              return ''.join(random.choice(alnum) for _ in range(16))
+
+          def handler(event, context):
+              redact_keys = {}
+              print('Context: {}'.format(context))
+              print({k:event[k] if k not in redact_keys else "[REDACTED]" for k in event.keys()})
```
demartinofra: isn't redact_keys empty?

cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,256 @@ Resources:
...
+                  logger.info('Deleting ACM certificate {}...'.format(certificate_arn))
+                  max_attempts = 10
+                  sleep_time = 60
+                  for attempt in range(0, max_attempts):
+                    try:
+                      acm.delete_certificate(CertificateArn=certificate_arn)
+                      break
+                    except acm.exceptions.ResourceInUseException as e:
+                      logger.info('(Attempt {}/{}) Cannot delete ACM certificate because it is in use. Retrying in {} seconds...'.format(attempt+1, max_attempts, sleep_time))
+                      time.sleep(sleep_time)
```
demartinofra: nit: can be moved to its own function

cloudformation/ad/ad-integration.yaml
```diff
@@ -507,22 +537,256 @@ Resources:
...
+                else:
+                    physical_resource_id = event['PhysicalResourceId']
```
demartinofra: nit: you could initialize physical_resource_id to event.get("PhysicalResourceId", create_physical_resource_id())

cloudformation/ad/ad-integration.yaml
demartinofra: where is this key generated? is it generated with the right permissions? do we need to keep the key locally or we can delete it once we upload it to secrets manager?

also if not needed on this instance have we considered generating this directly in Lambda?

gmarciani: > out of curiosity why 90 days?

There is no specific reason for that. We use it as default duration in our integration tests and I aligned to it. 
Currently, the industry standard is 1 year duration (https://knowledge.digicert.com/generalinformation/2-year_Certificate_Availability.html?), but I thought 3 months were enough for a tutorial. Do you think it should be increased?

charlesg3: > industry standard is 1 year duration

I think we should go with 1 year here too

gmarciani: > where is this key generated? is it generated with the right permissions?

Both the certificate and the private key are generated locally to the instance, but they are not meant to be used there. In fact, they are only meant to be uploaded to secrets manager. So, permissions is not an issue and both files can be deleted once uploaded to Secrets Manager.

> do we need to keep the key locally or we can delete it once we upload it to secrets manager?

We can delete both certificate and private key from the instance. Good point, will fix in the next rev.

gmarciani: > also if not needed on this instance have we considered generating this directly in Lambda?

Generating certificate and private key within Lambda was actually the first approach that I considered. At the end I discarded it because it would have required to include external dependencies in the Lambda (pyOpenSSL). So, doing that within a Lambda would have introduced more complexity to the template (Lambda code on S3, Lambda Layers, ...)

cloudformation/ad/ad-integration.yaml
gmarciani: No, will remove in the next rev. Thanks

cloudformation/ad/ad-integration.yaml
gmarciani: You're right. the solution to overcome this would introduce a bit of additional complexity, but we can scope it down.

### Alternative Solution
Have a Lambda that is responsible only for the certificate import *with tags* (https://docs.aws.amazon.com/acm/latest/APIReference/API_DeleteCertificate.html). 
Have a separate cleanup Lambda that will be responsible for deletion only.
In this case the cleanup Lambda could have a scoped down DeleterCertificate permission because at that point the Certificate ARN is known.

What do you think?

cloudformation/ad/ad-integration.yaml
gmarciani: Yes, will fix in the next rev

cloudformation/ad/ad-integration.yaml
gmarciani: right, will do in the next rev

cloudformation/ad/ad-integration.yaml
gmarciani: Right, thanks

cloudformation/ad/ad-integration.yaml
gmarciani: Thanks for the heads up. I did not known it was possible, Testing it.

cloudformation/ad/ad-integration.yaml
gmarciani: Thanks for the heads up on this. Since they are inline policies I think we can also avoid to include the StackName into the policy name

gmarciani: Same argument applies to Lambda function names (64 chars limit).
In order to be robust we should omit Lambda Function names at all in order to make CFN generate a random one.

cloudformation/ad/ad-integration.yaml
demartinofra: >Generating certificate and private key within Lambda was actually the first approach that I considered. At the end I discarded it because it would have required to include external dependencies in the Lambda (pyOpenSSL). So, doing that within a Lambda would have introduced more complexity to the template (Lambda code on S3, Lambda Layers, ...)

makes sense, for some reason I was assuming that openssl is available in the lambda default runtime and you could just use it.

cloudformation/ad/ad-integration.yaml
gmarciani: Tested that it is possible to specify a file as parameter. Will do that way in the next rev

cloudformation/ad/ad-integration.yaml
gmarciani: I thought about alternatives ways, but did not find any. Proceeding with the cleanup lambda approch in order to scope down both certificate import and certificate deletion.

demartinofra (dismissed):

demartinofra (approved):

PR #4005, opened 2022-04-28
Add validator for DirectoryService/PasswordSecretArn

gmarciani:
### Description of changes
Add validator for DirectoryService/PasswordSecretArn. The validation logic works as follows:

If the user policy has the permission to verify that the secret exists and the secret actually exists, the validation succeedsa and the creation/update proceeds.

If the user policy have the permission to verify that the secret exists and the secret does not actually exist, the following error is returned and the creation/update is blocked:

```
{
    {
      "level": "ERROR",
      "type": "PasswordSecretArnValidator",
      "message": "Secrets Manager can't find the specified secret."
    }
  ],
}
```

If the user policy does not have the permission to verify that the secret exists, the following warning is returned and the creation/update proceeds:

```
"validationMessages": [
    {
      "level": "WARNING",
      "type": "PasswordSecretArnValidator",
      "message": "Cannot validate PasswordSecretArn due to lack of permissions. Please refer to ParallelCluster official documentation for more information."
    }
  ]
```

### Tests
1. Manual tests:
    1. User policy having `secretmanager:DescribeSecret` and the secret exists, the validation is ok;
    2. User policy having `secretmanager:DescribeSecret` and the secret does not exist, the validation fails with error;
    3. User policy missing `secretmanager:DescribeSecret` and the secret exists, the warning is emitted about missing validation and the creation will succeed.
    4. User policy missing `secretmanager:DescribeSecret` and the secret does not exist, the warning is emitted about missing validation and the creation will fail.
2. Unit tests;

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -12,6 +12,8 @@
 import re
 from urllib.parse import urlparse
 
+from pcluster.aws.aws_api import AWSApi
+from pcluster.aws.common import AWSClientError
 from pcluster.constants import DIRECTORY_SERVICE_RESERVED_SETTINGS
 from pcluster.validators.common import FailureLevel, Validator
 
@@ -66,6 +68,31 @@ class DomainNameValidator(Validator):
                 "Supported formats are FQDN (corp.example.com) or LDAP Distinguished Name (DC=corp,DC=example,DC=com).",
                 FailureLevel.ERROR,
             )
+
+
+class PasswordSecretArnValidator(Validator):
+    """PasswordSecretArn validator."""
+
+    def _validate(self, password_secret_arn):
+        """Validate that PasswordSecretArn contains an ARN of a readable secret in AWS Secrets Manager."""
+        try:
+            # We only require the secret to exist; we do not validate its content.
+            AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
+        except AWSClientError as e:
+            if e.error_code == "ResourceNotFoundExceptionSecrets":
+                self._add_failure(f"The secret {password_secret_arn} does not exist.", FailureLevel.ERROR)
+            elif e.error_code == "AccessDeniedException":
+                self._add_failure(
+                    f"Cannot validate secret {password_secret_arn} due to lack of permissions. "
+                    "Please refer to ParallelCluster official documentation for more information.",
+                    FailureLevel.WARNING,
+                )
+            else:
+                self._add_failure(
+                    f"Cannot validate secret {password_secret_arn}. "
+                    "Please refer to ParallelCluster official documentation for more information.",
+                    FailureLevel.WARNING,
+                )
 
 
 class LdapTlsReqCertValidator(Validator):
--- a/cli/tests/pcluster/validators/test_directory_service_validators.py
+++ b/cli/tests/pcluster/validators/test_directory_service_validators.py
@@ -10,13 +10,16 @@
 # limitations under the License.
 import pytest
 
+from pcluster.aws.common import AWSClientError
+from pcluster.validators.common import FailureLevel
 from pcluster.validators.directory_service_validators import (
     AdditionalSssdConfigsValidator,
     DomainAddrValidator,
     DomainNameValidator,
     LdapTlsReqCertValidator,
+    PasswordSecretArnValidator,
 )
-from tests.pcluster.validators.utils import assert_failure_messages
+from tests.pcluster.validators.utils import assert_failure_level, assert_failure_messages
 
 DOMAIN_NAME_ERROR_MESSAGE = (
     "Unsupported domain address format. "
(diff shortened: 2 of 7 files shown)


cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -68,6 +70,31 @@ def _validate(self, domain_name):
...
+    """PasswordSecretArn validator."""
+
+    def _validate(self, password_secret_arn):
+        """Validate that PasswordSecretArn contains an ARN of a readable secret in AWS Secrets Manager."""
+        try:
+            # We only require the secret to exist; we do not validate its content.
+            AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
+        except AWSClientError as e:
+            if e.error_code == "ResourceNotFoundExceptionSecrets":
+                self._add_failure("The secrets provided in PasswordSecretArn does not exist.", FailureLevel.ERROR)
```
enrico-usai: the secret

enrico-usai: In general the validators should provide "semantic" validation and they shouldn't mention parameters names in the error messages. In this way we are decoupling the semantic validation from the schema.

The error message should be: `The Secret Arn {secret_arn} does not exist.`.

We could also rename the `PasswordSecretArnValidator` to `SecretValidator` in this way we could use the same validator to validate different secrets in future and we can extend it to verify multiple things and not just the arn existence in future.

For the same reason we can rename the param from `password_secret_arn` to `secret_arn`.

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -68,6 +70,31 @@ def _validate(self, domain_name):
...
+            elif e.error_code == "AccessDeniedException":
+                self._add_failure(
+                    "Cannot validate PasswordSecretArn due to lack of permissions. "
```
enrico-usai: Same here:

`Cannot validate Secret Arn {secret_arn} due to lack of permissions`

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -68,6 +70,31 @@ def _validate(self, domain_name):
...
+                    "Please refer to ParallelCluster official documentation for more information.",
```
enrico-usai, line 87: it would be good to put a link in the error.

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -68,6 +70,31 @@ def _validate(self, domain_name):
...
+                    FailureLevel.WARNING,
+                )
+            else:
+                self._add_failure(
+                    "Cannot validate PasswordSecretArn. "
```
enrico-usai: Cannot validate Secret Arn {secret_arn}. Please refer to ParallelCluster official documentation for more information.

cli/src/pcluster/validators/directory_service_validators.py
enrico-usai: let's talk about this if needed.

cli/tests/pcluster/validators/test_directory_service_validators.py
```diff
@@ -111,6 +113,49 @@ def test_ldap_tls_reqcert_validator(ldap_tls_reqcert, expected_message):
     assert_failure_messages(actual_failures, expected_message)
 
 
+@pytest.mark.parametrize(
+    "password_secret_arn, error_from_secrets_manager, expected_message, expected_error",
```
enrico-usai: just secret_arn to be more generic

cli/tests/pcluster/validators/utils.py
```diff
@@ -28,3 +30,15 @@ def assert_failure_messages(actual_failures, expected_messages):
     else:
         print(actual_failures)
         assert_that(actual_failures).is_empty()
+
+
+def assert_failure_level(actual_failures, expected_error):
```
enrico-usai: I don't like so much using a "boolean" variable to check True/False/None values. 

I'd like to use:
* True --> we're expecting an error
* False/None --> we're not expecting an error.

But according to the function name you're checking for the "failure level" so you're trying to identify if there are errors or warnings.

In this case you could use `expected_failure_level` and have:
* FailureLevel.ERROR --> check if error exist
* FailureLevel.WARNING --> check if warning exist (you can reuse the same code, just search for the other level)
* None --> assert there are no error nor warnings, actual failures is empty

Let's talk if needed.

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: The sentence "Please refer...." has been approved as is (without links) by the PM.
I thin k it makes sense to not couple error message with docs links.

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Will fix in next rev, thanks

gmarciani: Will replace the config. name with the actual secret.

But, I'm not sure that generalizing `PasswordSecretArnValidator` into `SecretValidator` is the right choice at this point because:
1. In general I think that generalizations should follow the [Rule of Three Strikes](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)). We can reduce it to 2 strikers, but here we have only one :-)
2. Currently, the name of the validator helps users to identify the offending property. In the long term we should consider to include in the error message the list of offending properties in order to decouple validator's name from the offending properties. Nevertheless, for the time being it is not like that.

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Will fix in the next rev, thanks

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: Will fix in the next rev, thanks.

cli/tests/pcluster/validators/test_directory_service_validators.py
gmarciani: This fix is conditioned to the generalization of `PasswordSecretArnValidator` to `SecretValidator`, which at this point does not really convince me.

cli/tests/pcluster/validators/utils.py
gmarciani: I agree, will fix in the next rev

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -68,6 +70,33 @@ def _validate(self, domain_name):
...
+    """PasswordSecretArn validator."""
+
+    def _validate(self, password_secret_arn):
+        """Validate that PasswordSecretArn contains an ARN of a readable secret in AWS Secrets Manager."""
+        try:
+            # We only require the secret to exist; we do not validate its content.
+            AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
+        except AWSClientError as e:
+            if e.error_code == "ResourceNotFoundExceptionSecrets":
+                self._add_failure("The secret {0} does not exist.".format(password_secret_arn), FailureLevel.ERROR)
```
enrico-usai: Why not using f-strings?
https://www.scivision.dev/python-f-string-speed/

cli/tests/pcluster/validators/test_directory_service_validators.py
```diff
@@ -111,6 +114,49 @@ def test_ldap_tls_reqcert_validator(ldap_tls_reqcert, expected_message):
...
+        (
+            "arn:PARTITION:secretsmanager:REGION:ACCOUNT:secret:NOT_ACCESSIBLE_SECRET",
+            "ResourceNotFoundExceptionSecrets",
+            "The secret arn:PARTITION:secretsmanager:REGION:ACCOUNT:secret:NOT_ACCESSIBLE_SECRET does not exist.",
+            FailureLevel.ERROR,
+        ),
+        (
+            "arn:PARTITION:secretsmanager:REGION:ACCOUNT:secret:ANY_SECRET",
+            "AccessDeniedException",
+            "Cannot validate secret arn:PARTITION:secretsmanager:REGION:ACCOUNT:secret:ANY_SECRET "
```
enrico-usai, line 129: Up to you: if you want to simplify the parameters you can use `.*` in the `expected_message`.

cli/tests/pcluster/validators/test_directory_service_validators.py
enrico-usai: I think this could be modified in any case, anyway not strictly needed for now.

cli/tests/pcluster/validators/test_directory_service_validators.py
gmarciani: Thanks for the hint. Actually, I prefer the exact match

enrico-usai (approved):

PR #4048, opened 2022-05-19
Support mounting existing FSx for Ontap and OpenZFS

hanwen-pcluste:
1. Add section of FsxOntap and FsxOpenZfs settings in configuration file:
```
SharedStorage:
  - MountDir: /my/mount/point1
    Name: name1
    StorageType: FsxOpenZfs
    FsxOpenZfsSettings:
      VolumeId: fsvol-12345678901234567
  - MountDir: /my/mount/point2
    Name: name2
    StorageType: FsxOntap
    FsxOntapSettings:
      VolumeId: fsvol-12345678901234567
```
2. Use class inheritance in cluster_config.py for FSx for Lustre, Optap, OpenZFS
3. Pass two more attributes to cookbook:
`fsx_volume_junction_paths`: It is a comma separated list. There are strings between the comma if it is FSx for Ontap and OpenZFS. There is empty if it is FSx for Lustre.
`fsx_fs_types`: It is a comma separated list. It tells the cookbook what mount method to use according to the file system (Lustre, Ontap, OpenZFS)
4. Improve FSx security group validator to check different ports for different file system
5. Improve integration test to test the new file system. Move fsx_factory to conftest, so it can be reused by other tests

Test:
Manual tests are done with test_ad_integration and test_cluster_in_no_internet_subnet

### Tests
Improved test_multiple_fsx tests to check Ontap and OpenZFS,
Manual tests are done with test_ad_integration and test_cluster_in_no_internet_subnet

### References
* This PR works together with https://github.com/aws/aws-parallelcluster-cookbook/pull/1446

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -11,6 +11,8 @@
 import logging
 
 from pcluster.constants import (
+    LUSTRE,
+    OPENZFS,
     PCLUSTER_IMAGE_BUILD_LOG_TAG,
     PCLUSTER_IMAGE_CONFIG_TAG,
     PCLUSTER_IMAGE_ID_TAG,
@@ -248,14 +250,25 @@ class FsxFileSystemInfo:
         self.file_system_data = file_system_data
 
     @property
+    def file_system_type(self):
+        """Return the type of FSx file system (LUSTRE, WINDOWS, ONTAP, or OPENZFS). WINDOWS is not supported."""
+        return self.file_system_data.get("FileSystemType")
+
+    @property
     def mount_name(self):
-        """Return MountName of the filesystem."""
-        return self.file_system_data.get("LustreConfiguration").get("MountName")
+        """Return MountName of the FSx Lustre file system."""
+        return (
+            self.file_system_data.get("LustreConfiguration").get("MountName") if self.file_system_type == LUSTRE else ""
+        )
 
     @property
     def dns_name(self):
-        """Return DNSName of the filesystem."""
-        return self.file_system_data.get("DNSName")
+        """
+        Return DNSName of the file system.
+
+        Lustre, OpenZFS have DNS name on file systems. Ontap has DNS name on storage virtual machines.
+        """
+        return self.file_system_data.get("DNSName") if self.file_system_type in [LUSTRE, OPENZFS] else ""
 
     @property
     def file_system_id(self):
--- a/cli/src/pcluster/aws/fsx.py
+++ b/cli/src/pcluster/aws/fsx.py
@@ -18,6 +18,8 @@ class FSxClient(Boto3Client):
     def __init__(self):
         super().__init__("fsx")
         self.cache = {}
+        self.svm_cache = {}
+        self.volume_cache = {}
 
     @AWSExceptionHandler.handle_client_exception
     def get_file_systems_info(self, fsx_fs_ids):
@@ -41,6 +43,44 @@ class FSxClient(Boto3Client):
                 file_system_info = FsxFileSystemInfo(file_system)
                 self.cache[file_system_info.file_system_id] = file_system_info
                 result.append(file_system_info)
+        return result
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_storage_virtual_machines(self, storage_virtual_machine_ids):
+        """Describe storage virtual machines."""
+        result = []
+        missed_storage_virtual_machine_ids = []
+        for storage_virtual_machine_id in storage_virtual_machine_ids:
+            cached_data = self.svm_cache.get(storage_virtual_machine_id)
+            if cached_data:
+                result.append(cached_data)
+            else:
+                missed_storage_virtual_machine_ids.append(storage_virtual_machine_id)
+        if missed_storage_virtual_machine_ids:
+            response = self._client.describe_storage_virtual_machines(
+                StorageVirtualMachineIds=missed_storage_virtual_machine_ids
+            )["StorageVirtualMachines"]
+            for storage_virtual_machine in response:
+                self.svm_cache[storage_virtual_machine.get("StorageVirtualMachineId")] = storage_virtual_machine
+                result.append(storage_virtual_machine)
+        return result
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_volumes(self, volume_ids):
+        """Describe FSx volumes."""
+        result = []
+        missed_volume_ids = []
+        for volume_id in volume_ids:
+            cached_data = self.volume_cache.get(volume_id)
+            if cached_data:
+                result.append(cached_data)
+            else:
+                missed_volume_ids.append(volume_id)
+        if missed_volume_ids:
+            response = self._client.describe_volumes(VolumeIds=missed_volume_ids)["Volumes"]
+            for volume in response:
+                self.volume_cache[volume.get("VolumeId")] = volume
+                result.append(volume)
         return result
 
     @AWSExceptionHandler.handle_client_exception
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -34,12 +34,15 @@ from pcluster.constants import (
     EBS_VOLUME_SIZE_DEFAULT,
     EBS_VOLUME_TYPE_DEFAULT,
     EBS_VOLUME_TYPE_IOPS_DEFAULT,
+    LUSTRE,
     MAX_EBS_COUNT,
     MAX_EXISTING_STORAGE_COUNT,
     MAX_NEW_STORAGE_COUNT,
     MAX_NUMBER_OF_COMPUTE_RESOURCES,
     MAX_NUMBER_OF_QUEUES,
     NODE_BOOTSTRAP_TIMEOUT,
+    ONTAP,
+    OPENZFS,
     SCHEDULER_PLUGIN_INTERFACE_VERSION,
     SCHEDULER_PLUGIN_INTERFACE_VERSION_LOW_RANGE,
     SUPPORTED_OSES,
@@ -294,13 +297,37 @@ class SharedEfs(Resource):
             self._register_validator(KmsKeyIdEncryptedValidator, kms_key_id=self.kms_key_id, encrypted=self.encrypted)
 
 
-class SharedFsx(Resource):
+class BaseSharedFsx(Resource):
+    """Represent the shared FSX resource."""
+
+    def __init__(self, mount_dir: str, name: str):
+        super().__init__()
+        self.mount_dir = Resource.init_param(mount_dir)
+        self.name = Resource.init_param(name)
+        self.shared_storage_type = SharedStorageType.FSX
+        self.__file_system_data = None
+
+    def _register_validators(self):
+        self._register_validator(SharedStorageNameValidator, name=self.name)
+
+    @property
+    def file_system_data(self):
+        """Return filesystem information if using existing FSx."""
+        if not self.__file_system_data and self.file_system_id:
+            self.__file_system_data = AWSApi.instance().fsx.get_file_systems_info([self.file_system_id])[0]
+        return self.__file_system_data
+
+    @property
+    def existing_dns_name(self):
+        """Return DNSName if using existing FSx filesystem."""
+        return self.file_system_data.dns_name if self.file_system_id else ""
+
+
+class SharedFsxLustre(BaseSharedFsx):
     """Represent the shared FSX resource."""
 
     def __init__(
         self,
-        mount_dir: str,
-        name: str,
         storage_capacity: int = None,
         deployment_type: str = None,
         data_compression_type: str = None,
@@ -318,11 +345,9 @@ class SharedFsx(Resource):
         auto_import_policy: str = None,
         drive_cache_type: str = None,
         fsx_storage_type: str = None,
+        **kwargs,
     ):
-        super().__init__()
-        self.mount_dir = Resource.init_param(mount_dir)
-        self.name = Resource.init_param(name)
-        self.shared_storage_type = SharedStorageType.FSX
+        super().__init__(**kwargs)
         self.storage_capacity = Resource.init_param(storage_capacity)
         self.fsx_storage_type = Resource.init_param(fsx_storage_type)
         self.deployment_type = Resource.init_param(deployment_type, default="SCRATCH_2")
@@ -340,11 +365,10 @@ class SharedFsx(Resource):
         self.file_system_id = Resource.init_param(file_system_id)
         self.auto_import_policy = Resource.init_param(auto_import_policy)
         self.drive_cache_type = Resource.init_param(drive_cache_type)
-        self.fsx_storage_type = Resource.init_param(fsx_storage_type)
-        self.__file_system_data = None
+        self.file_system_type = LUSTRE
 
     def _register_validators(self):
-        self._register_validator(SharedStorageNameValidator, name=self.name)
+        super()._register_validators()
         self._register_validator(
             FsxS3Validator,
             import_path=self.import_path,
@@ -399,21 +423,61 @@ class SharedFsx(Resource):
             )
 
     @property
-    def file_system_data(self):
-        """Return filesystem information if using existing FSx."""
-        if not self.__file_system_data and self.file_system_id:
-            self.__file_system_data = AWSApi.instance().fsx.get_file_systems_info([self.file_system_id])[0]
-        return self.__file_system_data
-
-    @property
     def existing_mount_name(self):
         """Return MountName if using existing FSx filesystem."""
         return self.file_system_data.mount_name if self.file_system_id else ""
 
+
+class ExistingFsxOpenZfs(BaseSharedFsx):
+    """Represent the shared FSX for OpenZFS resource."""
+
+    def __init__(self, volume_id: str, **kwargs):
+        super().__init__(**kwargs)
+        self.volume_id = volume_id
+        self.file_system_type = OPENZFS
+
+    @property
+    def file_system_id(self):
+        """Return the file system id behind the volume."""
+        return AWSApi.instance().fsx.describe_volumes([self.volume_id])[0]["FileSystemId"]
+
+    @property
+    def volume_path(self):
+        """Return the volume path."""
+        return AWSApi.instance().fsx.describe_volumes([self.volume_id])[0]["OpenZFSConfiguration"]["VolumePath"]
+
+
+class ExistingFsxOntap(BaseSharedFsx):
+    """Represent the shared FSX for Ontap resource."""
+
+    def __init__(self, volume_id: str, **kwargs):
+        super().__init__(**kwargs)
+        self.volume_id = volume_id
+        self.file_system_type = ONTAP
+
+    @property
+    def file_system_id(self):
+        """Return the file system id behind the volume."""
+        return AWSApi.instance().fsx.describe_volumes([self.volume_id])[0]["FileSystemId"]
+
+    @property
+    def storage_virtual_machine_id(self):
+        """Return the storage virtual machine behind the volume."""
+        return AWSApi.instance().fsx.describe_volumes([self.volume_id])[0]["OntapConfiguration"][
+            "StorageVirtualMachineId"
+        ]
+
+    @property
+    def junction_path(self):
+        """Return the junction path."""
+        return AWSApi.instance().fsx.describe_volumes([self.volume_id])[0]["OntapConfiguration"]["JunctionPath"]
+
     @property
     def existing_dns_name(self):
-        """Return DNSName if using existing FSx filesystem."""
-        return self.file_system_data.dns_name if self.file_system_id else ""
+        """Return DNSName of the SVM of existing FSx filesystem."""
+        return AWSApi.instance().fsx.describe_storage_virtual_machines([self.storage_virtual_machine_id])[0][
+            "Endpoints"
+        ]["Nfs"]["DNSName"]
 
 
 # ---------------------- Networking ---------------------- #
@@ -1106,10 +1170,13 @@ class BaseClusterConfig(Resource):
             self._register_validator(S3BucketRegionValidator, bucket=self.custom_s3_bucket, region=self.region)
 
     def _register_storage_validators(self):
-        ebs_count = 0
-        new_storage_count = defaultdict(int)
-        existing_storage_count = defaultdict(int)
         if self.shared_storage:
+            ebs_count = 0
+            new_storage_count = defaultdict(int)
+            existing_storage_count = defaultdict(int)
+
+            self._cache_describe_volume()
+
             self._register_validator(
                 DuplicateNameValidator,
                 name_list=[storage.name for storage in self.shared_storage],
@@ -1120,11 +1187,12 @@ class BaseClusterConfig(Resource):
                 name_list=self.existing_fs_id_list,
                 resource_name="Shared Storage IDs",
             )
+
             existing_fsx = []
             for storage in self.shared_storage:
                 self._register_validator(SharedStorageNameValidator, name=storage.name)
                 self._register_validator(SharedStorageMountDirValidator, mount_dir=storage.mount_dir)
-                if isinstance(storage, SharedFsx):
+                if isinstance(storage, BaseSharedFsx):
                     if storage.file_system_id:
                         existing_storage_count["fsx"] += 1
                         existing_fsx.append(storage.file_system_id)
@@ -1133,6 +1201,8 @@ class BaseClusterConfig(Resource):
                     self._register_validator(
                         FsxArchitectureOsValidator, architecture=self.head_node.architecture, os=self.image.os
                     )
+                if isinstance(storage, (ExistingFsxOpenZfs, ExistingFsxOntap)):
+                    existing_fsx.append(storage.file_system_id)
                 if isinstance(storage, SharedEbs):
                     if storage.raid:
                         new_storage_count["raid"] += 1
@@ -1156,26 +1226,11 @@ class BaseClusterConfig(Resource):
                 are_all_security_groups_customized=self.are_all_security_groups_customized,
             )
 
-            for storage_type in ["efs", "fsx", "raid"]:
-                self._register_validator(
-                    NumberOfStorageValidator,
-                    storage_type=f"new {storage_type.upper()}",
-                    max_number=MAX_NEW_STORAGE_COUNT.get(storage_type),
-                    storage_count=new_storage_count[storage_type],
-                )
-                self._register_validator(
-                    NumberOfStorageValidator,
-                    storage_type=f"existing {storage_type.upper()}",
-                    max_number=MAX_EXISTING_STORAGE_COUNT.get(storage_type),
-                    storage_count=existing_storage_count[storage_type],
-                )
-            self._register_validator(
-                NumberOfStorageValidator,
-                storage_type="EBS",
-                max_number=MAX_EBS_COUNT,
-                storage_count=ebs_count,
-            )
+            self._validate_max_storage_count(ebs_count, existing_storage_count, new_storage_count)
 
+        self._validate_mount_dirs()
+
+    def _validate_mount_dirs(self):
         self._register_validator(
             DuplicateMountDirValidator,
             mount_dir_list=self.shared_mount_dir_list + list(self.local_mount_dir_set),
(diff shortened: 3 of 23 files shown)


cli/src/pcluster/aws/aws_resources.py
```diff
@@ -11,6 +11,8 @@
 import logging
 
 from pcluster.constants import (
+    FSX_LUSTRE,
+    FSX_OPENZFS,
```
gmarciani: since `FSX_ONTAP` is missing, I assume there will be some differences in the way we interact with Ontap. May you please clarify?

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -247,15 +249,24 @@ class FsxFileSystemInfo:
...
+        return self.file_system_data.get("FileSystemType")
+
     @property
     def mount_name(self):
         """Return MountName of the filesystem."""
-        return self.file_system_data.get("LustreConfiguration").get("MountName")
+        return (
+            self.file_system_data.get("LustreConfiguration").get("MountName")
+            if self.file_system_type == FSX_LUSTRE
+            else ""
```
gmarciani: Why `MountName` is empty for FsxOpenZsf and FsxOntap? In the config file we are able to set the mount name for every type of FS so this looks like unexpected. May you please clarify both in this conversation and with a comment on code?

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -247,15 +249,24 @@ class FsxFileSystemInfo:
...
+        )
 
     @property
     def dns_name(self):
         """Return DNSName of the filesystem."""
-        return self.file_system_data.get("DNSName")
+        return self.file_system_data.get("DNSName") if self.file_system_type in [FSX_LUSTRE, FSX_OPENZFS] else ""
```
gmarciani: Same as above: I'm not getting why the `DnsName` is returned only for FsxLustre and FsxOpenZfs and it's not for FsxOntap. May you please clarify both in this conversation and with a comment on code?

cli/src/pcluster/aws/fsx.py
```diff
@@ -18,6 +18,8 @@ class FSxClient(Boto3Client):
     def __init__(self):
         super().__init__("fsx")
         self.cache = {}
+        self.svm_cache = {}
```
gmarciani, line 21: I suppose that *svm* stands for storage virtual machines, but I needed to go through the rest of the code to understand it? I suggest to avoid acronyms whenever it is possible.

cli/src/pcluster/aws/fsx.py
```diff
@@ -43,6 +45,44 @@ def get_file_systems_info(self, fsx_fs_ids):
...
 
+    @AWSExceptionHandler.handle_client_exception
+    def describe_storage_virtual_machines(self, storage_virtual_machine_ids):
+        """Describe storage virtual machines."""
+        result = []
+        missed_storage_virtual_machine_ids = []
+        for storage_virtual_machine_id in storage_virtual_machine_ids:
+            cached_data = self.svm_cache.get(storage_virtual_machine_id)
+            if cached_data:
+                result.append(cached_data)
```
gmarciani, line 56: We are caching without cache expiration strategy. 
I suppose that if we are caching we are expecting this command to be executed a lot of times.
Are we sure that the cached data is not invalidated over time?

cli/src/pcluster/aws/fsx.py
```diff
@@ -43,6 +45,44 @@ def get_file_systems_info(self, fsx_fs_ids):
...
+            else:
+                missed_storage_virtual_machine_ids.append(storage_virtual_machine_id)
+        if missed_storage_virtual_machine_ids:
+            response = self._client.describe_storage_virtual_machines(
```
gmarciani, line 60: Here we are introducing a new AWS API call. Let's be sure that we are aligning the user policies accordingly in our public docs.
If it is not yet included, we must document it and review it with the security team.

cli/src/pcluster/aws/fsx.py
```diff
@@ -43,6 +45,44 @@ def get_file_systems_info(self, fsx_fs_ids):
...
+        result = []
+        missed_volume_ids = []
+        for volume_id in volume_ids:
+            cached_data = self.volume_cache.get(volume_id)
+            if cached_data:
+                result.append(cached_data)
+            else:
+                missed_volume_ids.append(volume_id)
+        if missed_volume_ids:
+            response = self._client.describe_volumes(VolumeIds=missed_volume_ids)["Volumes"]
```
gmarciani, line 80: Same as above: let's be sure that this call is allowed by the user policy in our public docs.
If it is not yet included, we must document it and review it with the security team.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -294,13 +297,37 @@ def _register_validators(self):
...
+        self.shared_storage_type = SharedStorageType.FSX
+        self.__file_system_data = None
+
+    def _register_validators(self):
+        self._register_validator(SharedStorageNameValidator, name=self.name)
+
+    @property
+    def file_system_data(self):
+        """Return filesystem information if using existing FSx."""
+        if not self.__file_system_data and self.file_system_id:
```
gmarciani, line 316: Here we are assuming that there cannot be the case of missing both file system data and file system id. Are we sure that our validators already prevent this?

gmarciani, line 316: I don't see the class attribute `file_system_id` defined anywhere in the class constructor. Where are we expecting it to be defined? If we are expecting this class attribute to be defined in a subclass, I think it's not a good practice: a class should reference its own data. In this specific case, I think we should move `file_system_id` in the base class, right?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -318,11 +345,9 @@ def __init__(
...
         fsx_storage_type: str = None,
+        **kwargs,
     ):
-        super().__init__()
-        self.mount_dir = Resource.init_param(mount_dir)
-        self.name = Resource.init_param(name)
-        self.shared_storage_type = SharedStorageType.FSX
+        super().__init__(**kwargs)
         self.storage_capacity = Resource.init_param(storage_capacity)
         self.fsx_storage_type = Resource.init_param(fsx_storage_type)
```
gmarciani, line 352: Isn't `fsx_storage_type` implicit in `file_system_type`? What about reducing knowledge duplication?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -398,22 +421,55 @@ def _register_validators(self):
...
-            self.__file_system_data = AWSApi.instance().fsx.get_file_systems_info([self.file_system_id])[0]
-        return self.__file_system_data
-
     @property
     def existing_mount_name(self):
         """Return MountName if using existing FSx filesystem."""
         return self.file_system_data.mount_name if self.file_system_id else ""
 
+
+class ExistingFsxOpenZfs(BaseSharedFsx):
```
gmarciani, line 431: According to PR description, it seems that both Ontap and OpenZFS have junction paths. Why this concept is missing in`ExistingFsxOpenZfs ` and it is implement as a property in the below `ExistingFsxOntap`?

gmarciani: According to the code, it seems that the two file systems share the concept of volumes, but not the concept of junction paths, so why are we using the naming *junction path instead of *volumes* is cookbook attributes?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1111,7 +1167,7 @@ def _register_validators(self):
             self._register_validator(S3BucketValidator, bucket=self.custom_s3_bucket)
             self._register_validator(S3BucketRegionValidator, bucket=self.custom_s3_bucket, region=self.region)
 
-    def _register_storage_validators(self):
+    def _register_storage_validators(self):  # noqa: C901
```
gmarciani: Why are we ignoring this flake rule?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1139,6 +1204,10 @@ def _register_storage_validators(self):
                     self._register_validator(
                         FsxArchitectureOsValidator, architecture=self.head_node.architecture, os=self.image.os
                     )
+                if isinstance(storage, ExistingFsxOpenZfs):
+                    existing_fsx.append(storage.file_system_id)
+                if isinstance(storage, ExistingFsxOntap):
+                    existing_fsx.append(storage.file_system_id)
```
gmarciani: Code duplication: these two lines are equivalent to the above ones.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1480,7 +1549,7 @@ def _register_validators(self):
 
         if self.shared_storage:
             for storage in self.shared_storage:
-                if isinstance(storage, SharedFsx):
+                if isinstance(storage, SharedFsxLustre):
                     self._register_validator(AwsBatchFsxValidator)
```
gmarciani, line 1562: Are we sure that the batch validator must be applied only when FsxLustre is used?

cli/src/pcluster/constants.py
```diff
@@ -52,6 +52,16 @@
 FSX_SSD_THROUGHPUT = {"PERSISTENT_1": [50, 100, 200], "PERSISTENT_2": [125, 250, 500, 1000]}
 FSX_HDD_THROUGHPUT = [12, 40]
 
+FSX_LUSTRE = "LUSTRE"
+FSX_OPENZFS = "OPENZFS"
+FSX_ONTAP = "ONTAP"
+
+FSX_PORTS = {
```
gmarciani, line 66: May you please add a comment with links to the official documentation that declares these ports, if any?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -429,6 +434,22 @@ def validate_backup_id_unsupported_parameters(self, data, **kwargs):
...
+    volume_id = fields.Str(
+        validate=validate.Regexp(r"^fsvol-[0-9a-f]{17}$"), metadata={"update_policy": UpdatePolicy.UNSUPPORTED}
+    )
+
+
+class FsxOntapSettingsSchema(BaseSchema):
+    """Represent the FSX Ontap schema."""
+
+    volume_id = fields.Str(
+        validate=validate.Regexp(r"^fsvol-[0-9a-f]{17}$"), metadata={"update_policy": UpdatePolicy.UNSUPPORTED}
```
gmarciani: 1. We are duplicating the same volume regexp for OpenZFS and Ontap. What about putting it as a constant in constants.py file close to where we define file system related constants?
2. May you please link the official doc where the regexp is documented?
3. Here we are including in the schema only the parameter `volume_id`, but in other parts of this PR we are assuming that the settings may have other parameters, e.g. StorageVirtualMachineId, VolumeJunctionPath. So, why are they excluded from the schema?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -472,16 +503,24 @@ def make_resource(self, data, **kwargs):
...
             return SharedEfs(**shared_volume_attributes)
-        elif storage_type == "FsxLustre":
-            return SharedFsx(**shared_volume_attributes)
         elif storage_type == "Ebs":
             return SharedEbs(**shared_volume_attributes)
+        elif storage_type == "FsxLustre":
+            return SharedFsxLustre(**shared_volume_attributes)
+        elif storage_type == "FsxOpenZfs":
+            return ExistingFsxOpenZfs(**shared_volume_attributes)
+        elif storage_type == "FsxOntap":
```
gmarciani: The following strings are duplicated many times: "FsxLustre", "FsxOpenZfs", "FsxOntap".
What about defining them as constants instead or re-writing them many times?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -492,14 +531,25 @@ def restore_child(self, data, **kwargs):
         if adapted_data.shared_storage_type == "efs":
             storage_type = "efs"
         elif adapted_data.shared_storage_type == "fsx":
-            storage_type = "fsx_lustre"
+            if adapted_data.file_system_type == FSX_LUSTRE:
+                storage_type = "fsx_lustre"
```
gmarciani: I'm not getting why in the lines from the previous diff block `storage_type` is expected to have values "FsxLustre", "FsxOpenZfs", "FsxOntap"; whereas in this diff block it is expected to have "fsx_lustre", "fsx_open_zfs" and "fsx_ontap". I think we should consider to align these naming conventions to reduce the risk of errors.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -492,14 +531,25 @@ def restore_child(self, data, **kwargs):
...
+            elif adapted_data.file_system_type == FSX_OPENZFS:
+                storage_type = "fsx_open_zfs"
+            elif adapted_data.file_system_type == FSX_ONTAP:
+                storage_type = "fsx_ontap"
         else:  # "raid", "ebs"
             storage_type = "ebs"
         setattr(adapted_data, f"{storage_type}_settings", copy.copy(adapted_data))
```
gmarciani, line 542: the missing final else within the first if makes `storage_type` potentially uninitialized

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -604,17 +608,24 @@ def _add_shared_storage(self, storage):
             storage_list.extend(self._add_raid_volume(cfn_resource_id, storage))
         self.shared_storage_mount_dirs[storage.shared_storage_type].append(storage.mount_dir)
 
-    def _add_fsx_storage(self, id: str, shared_fsx: SharedFsx):
+    def _add_fsx_storage(self, id: str, shared_fsx: BaseSharedFsx):
         """Add specific Cfn Resources to map the FSX storage."""
         fsx_id = shared_fsx.file_system_id
+        mount_name = ""
+        dns_name = ""
+        volume_junction_path = ""
```
gmarciani, line 616: Overall in the PRs we mention volumes, junctions paths, volume path and volume junction paths. We need to be more clear about their different roles and relations.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -410,6 +411,9 @@ class ExistingFsxNetworkingValidator(Validator):
     FSx networking validator.
 
     Validate file system mount point according to the head node subnet.
+    File system ids can be specified together with storage virtual machine ids. The code will get the file system ids
+    from storage virtual machine ids and combine the two file system ids lists to validate.
+    The reason to have this structure is to make boto3 calls as few as possible.
```
gmarciani, line 414: What is the upper bound of the number of calls we are expecting and how much are we reducing them?
Are we sure that aiming to this reduction does not make the code more error prone or, even worst, make the validation less clear from the customer perspective?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -521,8 +521,9 @@ def test_queue_name_validator(name, expected_message):
             [{"IpProtocol": "-1", "UserIdGroupPairs": [{"UserId": "123456789012", "GroupId": "sg-12345678"}]}],
             False,
             ["eni-09b9460295ddd4e5f"],
+            "The current security group settings on file system .* does not satisfy mounting requirement. "
             "The file system must be associated to a security group that "
-            "allows inbound and outbound TCP traffic through port 988.",
+            r"allows inbound and outbound TCP traffic through ports \[988\].",
```
gmarciani, line 548: Is the `r` prefix really necessary?
Furthermore, here we are assuming ports 988, so basically we are covering only the case of FsxLustre. What about the other ones?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -260,11 +271,46 @@ def test_fsx_lustre_backup(region, pcluster_config_reader, clusters_factory, sch
     _test_delete_manual_backup(manual_backup, region)
 
 
+def create_fsx_ontap(fsx_factory, num):
+    return fsx_factory(
+        ports=[111, 635, 2049, 4046],
+        ip_protocols=["tcp", "udp"],
```
gmarciani, line 277: Aren't we assuming to use TCP? Why are we including also UDP?

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: FSx names the path differently, e.g. `junction path` in Ontap vs `volume path` in OpenZFS. But the concept is the same. Therefore, the property names are different in the code. However, when passing the path to cookbook, I unified them into `fsx_volume_junction_paths`

If we pass volumes instead of paths to cookbook, an additional AWS API is required in Cookbook. AWS API call should be done on the CLI to reduce API throttling when scaling up, and to better support cluster in isolated subnet.

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Probably `fsx_storage_type` should be renamed as `fsx_lustre_storage_type`?

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: Mount name is specific to FSx Lustre, and is retrieved from AWS API

In config file, we can specify mount dir for all the file systems.

cli/src/pcluster/aws/fsx.py
hanwen-pcluste: SVM stands for storage virtual machine

cli/src/pcluster/aws/fsx.py
hanwen-pcluste: Within a single pcluster command execution, we can assume the API response is the same. Across multiple pcluster commands. the cache is always cleared

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: `file_system_data` is only used when existing FSx is provided, when `file_system_id` always exists.
If a new FSx Lustre is being created, the code does not use `file_system_data`

hanwen-pcluste: The retrieval of `file_system_id` is different depending on the file system type (Lustre, Ontap, OpenZFS)

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: It is function too complex warning. An alternative is to split the function into different function. What do you think?

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Good catch. I fix this along with other bugs from the renaming

cli/src/pcluster/constants.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: 1. We only duplicated once here. Can we keep them here?
2. Other regex in this file is not commented. The document is here https://docs.aws.amazon.com/fsx/latest/APIReference/API_DescribeVolumes.html#FSx-DescribeVolumes-request-VolumeIds. Should we add it to the doc? Are we assuming the link is valid permenantly?
3. Other information is implied by volume id

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: This is some magic with the schema. It is not easy to align them

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: The upper bound is 50 file systems. We are reducing them to constant number of calls regardless of the number of the file system

From the customer perspective, nothing is changed

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: `r` is needed when writing regex with escapes.
Yes , here we only cover Lustre

tests/integration-tests/tests/storage/test_fsx_lustre.py
hanwen-pcluste: Ontap and OpenZFS requires both

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: Because the condition statements below only affects some of the file system types

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: Done

cli/src/pcluster/aws/aws_resources.py
gmarciani: Sure, the question is: why Ontap differs w.r.t. the others.

cli/src/pcluster/aws/aws_resources.py
gmarciani: Thanks for adding a comment.
You say "Ontap has DNS name on storage virtual machines.".
So, we are not retrieving DNS name for Ontap here because in terms of clean coding is not correct to retrieve it from here, or because we do not need to retrieve it at all?

cli/src/pcluster/aws/fsx.py
gmarciani: Got it. So, within a single pcluster command how many times are we expecting to execute such API calls to justify a caching strategy?

cli/src/pcluster/config/cluster_config.py
gmarciani: So, file_system_id is an "abstract" attribute whose value is set within file system specific sub-classes.
Such abstract behaviour should be implemented in a more robust way, e.g. defined as an `@abstractmethod @property` within the super-class. See https://towardsdatascience.com/how-to-use-abstract-classes-in-python-d4d2ddc02e90

cli/src/pcluster/config/cluster_config.py
gmarciani: Checked the [docs](https://docs.aws.amazon.com/parallelcluster/latest/ug/SharedStorage-v3.html#yaml-SharedStorage-FsxLustreSettings-StorageType). there is no knowledge duplication here:`fsx_storage_type` (HDD,SSD) cannot be retrieved from `file_system_type` (LUSTRE).

Let's keep it as it is.

cli/src/pcluster/config/cluster_config.py
gmarciani: For the cookbook: let's keep concepts separated and self-contained. Generally speaking, variables should be associated to a single precise concept. We should avoid to have a variable like `fsx_volume_junction_paths` if it can contain volume ids OR paths. We should have a variable for junction paths and a variable for volume ids

cli/src/pcluster/config/cluster_config.py
gmarciani: I think it would be better to split as long as it improves readability and maintainability, that's generally the case.

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: 1. I think it's better to not duplicating the information at all
2. Thanks for sharing the link. I think adding the link in a comment would add some value because even if invalidated sooner or later it will provide information about what we considered the source of thruth for that regexp, i.e. the FSX DescribeVolumes API.
3. That does not match with what I understood from the design discussions. Let's have a follow up on this, please.

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: If something appears as magic, then we should improve something I suppose :-) 
Let's be sure to track this in a refactoring task.

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: Still not seeing the fix. By the way, what about using a dictionary instead of a nested if?

cli/tests/pcluster/validators/test_cluster_validators.py
gmarciani: Shouldn't we extend  it to cover also the other file systems?

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: The DNS name is needed. But for clean coding, we need to get the DNS name form storage virtual machine

cli/src/pcluster/aws/fsx.py
hanwen-pcluste: The maximum file system number is 50, So we are comparing 1 call vs 50 calls

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Volume id is not passed to cookbook. All the values in fsx_volume_junction_paths are paths

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: On point 3: checked the final decision in design and we need only volume id in the schema. That's fine.

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: For existing FSx for Lustre, `file_system_id` is a field from cluster configuration.
For existing FSx for Ontap and OpenZFS, `file_system_id` is property, retrieved using volume id.

So it not easy to do anything in the superclass

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: Done

cli/src/pcluster/schemas/cluster_schema.py
hanwen-pcluste: Done

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: Done

cli/src/pcluster/aws/aws_resources.py
gmarciani: We had a offline follow up on this. That's fine.

cli/src/pcluster/aws/fsx.py
gmarciani: Ok, so please correct me if I'm wrong.

We are caching SVM data because we can mount different volumes having the same SVM, so with caching we are limiting the number of times we retrieve SVM data.

Correct?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -294,13 +297,37 @@ def _register_validators(self):
...
+
+    def __init__(self, mount_dir: str, name: str):
+        super().__init__()
+        self.mount_dir = Resource.init_param(mount_dir)
+        self.name = Resource.init_param(name)
+        self.shared_storage_type = SharedStorageType.FSX
+        self.__file_system_data = None
+
+    def _register_validators(self):
+        self._register_validator(SharedStorageNameValidator, name=self.name)
```
gmarciani, line 311: Is it possible to test the inclusion of SharedStorageNameValidator (see codecov comment below)?

cli/src/pcluster/config/cluster_config.py
gmarciani: I'm not sure about that. Btw, let's consider this as as non blocking comment.

cli/src/pcluster/config/cluster_config.py
gmarciani: Got it, thanks

cli/src/pcluster/config/cluster_config.py
gmarciani: Did you fix it?

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: Thanks

cli/tests/pcluster/validators/test_cluster_validators.py
gmarciani: Thanks

tests/integration-tests/conftest.py
```diff
@@ -27,6 +27,7 @@
 
 import boto3
 import pytest
+import utils
```
gmarciani: What about the Code Scanning comment below?

cli/src/pcluster/aws/aws_resources.py
gmarciani: Questions:
1. What is the expected content of `file_system_data.get("DNSName")` for ONTAP?
1. What about adding the DNSName also for ONTAP within `file_system_data` before passing it to the constructor of `FsxFileSystemInfo`?

hanwen-pcluste: 1. There is no such key in the response. So the result is None
2. the DNSName of Ontap is on the storage virtual machine. Why adding it to file system data?

cli/src/pcluster/aws/fsx.py
hanwen-pcluste: Yes.
It is not only for the same SVM, but also for different SVMs.
There could've been a boto3 call to describe a list of SVMs, or many single calls to describe each SVM.
Here, we are using the first approach to reduce the number of calls to once

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Done

tests/integration-tests/conftest.py
hanwen-pcluste: Done

cli/src/pcluster/config/cluster_config.py
gmarciani: Not sure about that, but let's consider this as not blocking for the moment.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -604,17 +608,24 @@ def _add_shared_storage(self, storage):
...
         if fsx_id:
-            # Initialize DNSName and MountName for existing filesystem, if any
-            self.shared_storage_attributes[shared_fsx.shared_storage_type]["MountNames"].append(
-                shared_fsx.existing_mount_name
-            )
-            self.shared_storage_attributes[shared_fsx.shared_storage_type]["DNSNames"].append(
-                shared_fsx.existing_dns_name
-            )
+            if isinstance(shared_fsx, ExistingFsxOntap):
+                dns_name = AWSApi.instance().fsx.describe_storage_virtual_machines(
```
gmarciani: Not a blocker, just an improvement: what about setting the `dns_name` value within `shared_fsx.existing_dns_name`, like in the other cases (probably with different Boto calls)?

gmarciani (approved):

Issue #3576, opened 2021-12-07
Multiple AZ/subnets support


zbarr:
There's an issue open for this but it is specifically for awsbatch whereas I need this for Slurm. Our own proprietary "Slurm with AWS" implementation had this so we've gone backwards a bit in that respect by moving to ParallelCluster. Hoping you guys are able to get to this soon! Documentation indicates it's coming.

hanwen-pcluste:
Hello zbarr,

Thank you for the information. Multiple AZ support is a popular feature request. We are discussing this and will let you know any progress.

Cheers,
Hanwen

devanshkv:
Any updates or timelines for the multiple AZ support?

enrico-usai:
We don't have yet a timeline to share for this feature enhancement.

An high level alternative is to use Slurm Federation to manage multiple Slurm clusters. See Workshop: [Using AWS ParallelCluster for Research](https://catalog.us-east-1.prod.workshops.aws/workshops/d431e0b1-9f08-4d82-822e-ea56962b2a0b/en-US/summary).

gns-build-server:
I successfully did multi-AZ / multi-subnet with ParallelCluster 2.11.3 running SGE.  I am not sure if the issues people are seeing are SLURM-specific, or if there's something else I'm doing right here that AWS doesn't necessarily know about.  @enrico-usai , can you weigh in?

We are using:

scheduler = sge
base_os = ubuntu1804
cluster_type = spot

with 
master_subnet_id and compute_subnet_id set to the same subnet.  Create the cluster, and then edit the autoscaling group manually.  Under the "Network" section, change the one AZ/subnet listed to a full set of subnets you've created, one in each AZ.

As you can see in the screenshots, simply changing the ASG causes the autoscaler to spot request in multiple AZs, and running jobs span across the AZs:

![image](https://user-images.githubusercontent.com/4273527/160922296-6ff19359-b648-4e7c-b5a7-4717c9c6604e.png)
![image](https://user-images.githubusercontent.com/4273527/160922230-61e366db-bb78-4c8d-8269-0a4d86ad01f1.png)
![image](https://user-images.githubusercontent.com/4273527/160922368-a04c3e45-6e07-4015-b53c-daa9047d4481.png)
![image](https://user-images.githubusercontent.com/4273527/160922387-f8ec0a1d-6af0-4e36-b0b9-66fdedee9d89.png)
[...]

PR #4050, opened 2022-05-19
Multi-AZ

sean-smith:
Enable using a separate Availability Zone per-partition. Warn about cross AZ charges if multiple-AZ's are specified.

This enables users to use multiple availability zones. This is part of https://github.com/aws/aws-parallelcluster/issues/3576

### DIFF
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -47,6 +47,15 @@ class SubnetsValidator(Validator):
                         FailureLevel.ERROR,
                     )
 
+            # warn if subnets are not in the same AZ
+            azs = {subnet.get("AvailabilityZone") for subnet in subnets}
+            if len(azs) > 1:
+                self._add_failure(
+                    f"Subnets in multiple Availibility Zones {azs} may result in data transfer charges. "
+                    + "Please review the EC2 docs: https://aws.amazon.com/ec2/pricing/on-demand/#Data_Transfer",
+                    FailureLevel.WARNING,
+                )
+
             # Check for DNS support in the VPC
             if not AWSApi.instance().ec2.is_enable_dns_support(vpc_id):
                 self._add_failure(f"DNS Support is not enabled in the VPC {vpc_id}.", FailureLevel.ERROR)
--- a/cli/tests/pcluster/schemas/test_cluster_schema.py
+++ b/cli/tests/pcluster/schemas/test_cluster_schema.py
@@ -273,23 +273,6 @@ def dummy_slurm_compute_resource(name, instance_type):
             },
             None,
         ),
-        (
-            {
-                "Scheduler": "slurm",
-                "SlurmQueues": [
-                    dummy_slurm_queue(),
-                    {
-                        "Name": "queue2",
-                        "Networking": {"SubnetIds": ["subnet-00000000"]},
-                        "ComputeResources": [
-                            {"Name": "compute_resource3", "InstanceType": "c5.2xlarge", "MaxCount": 5},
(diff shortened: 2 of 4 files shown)


cli/src/pcluster/validators/networking_validators.py
```diff
@@ -47,6 +47,15 @@ def _validate(self, subnet_ids: List[str]):
...
                     )
 
+            # warn if subnets are not in the same AZ
+            azs = list(dict.fromkeys([subnet.get("AvailabilityZone") for subnet in subnets]))
+            if len(azs) > 1:
+                self._add_failure(
+                    f"Subnets in multiple Availibility Zones {azs} may result in data transfer charges. "
+                    + "Please review the EC2 docs: https://aws.amazon.com/ec2/pricing/on-demand/#Data_Transfer",
+                    FailureLevel.WARNING,
+                )
```
demartinofra, line 57: I'm ok with moving this check from the schema level into a suppressable validator however we can't mark this as a warning until the shared storage functionalities are also made compatible with multi az. I suggest to keep this as an error for now and re-word the error message.

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -47,6 +47,15 @@ def _validate(self, subnet_ids: List[str]):
                         FailureLevel.ERROR,
                     )
 
+            # warn if subnets are not in the same AZ
+            azs = list(dict.fromkeys([subnet.get("AvailabilityZone") for subnet in subnets]))
```
demartinofra: this can be simply replaced by a set rather than doing list -> dict -> list conversion.

Also differently from the schema level validator this is also including the head node subnet which becomes a breaking change if, as suggested, we transform the failure below into an error.

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -273,23 +273,6 @@ def dummy_slurm_compute_resource(name, instance_type):
...
-                        "Networking": {"SubnetIds": ["subnet-00000000"]},
-                        "ComputeResources": [
-                            {"Name": "compute_resource3", "InstanceType": "c5.2xlarge", "MaxCount": 5},
-                            {"Name": "compute_resource4", "InstanceType": "c4.2xlarge"},
-                        ],
-                    },
-                ],
-            },
-            "The SubnetIds used for all of the queues should be the same",
-        ),
```
demartinofra, line 292: the code change needs to be ideally covered by a new test

enrico-usai:
Thanks Sean for the changes.
Given that this commit is just a partial movement in the direction of the Multi-AZ support I'd modify the changelog, the PR title and the commit message accordingly. E.g. we can say: `Add possibility to suppress subnet ids validator by moving validator from schema to config`.

Then I'd remove the `Solves #3576` message from the description because the feature is still open. Note: there are some [GitHub keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) that automatically closes the related issues when the PR is merged, be sure to not use them in this case.

cli/src/pcluster/validators/networking_validators.py
sean-smith: What's the issue with shared storage?

cli/src/pcluster/validators/networking_validators.py
sean-smith: Gotcha so I transformed this into a set however I'd like to keep it as a warning so users don't have to set an extra flag to get it working

lukeseawalker:
superseded by https://github.com/aws/aws-parallelcluster/pull/4541

PR #4074, opened 2022-06-01
Fix test_fsx_lustre_backup

hanwen-pcluste:
With this commit https://github.com/aws/aws-parallelcluster/commit/271aba696ae9e7154582a859624928c74d5d8167, we added the default lustre deployment type `SCRATCH_2` and Lustre server version`2.12.

However, if backup id is specified, those defaults shouldn't be specified because the Lustre file system configuration will be created according to the backup configuration.

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -350,7 +350,9 @@ class SharedFsxLustre(BaseSharedFsx):
         super().__init__(**kwargs)
         self.storage_capacity = Resource.init_param(storage_capacity)
         self.fsx_storage_type = Resource.init_param(fsx_storage_type)
-        self.deployment_type = Resource.init_param(deployment_type, default="SCRATCH_2")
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -366,6 +366,7 @@ def __init__(
         self.auto_import_policy = Resource.init_param(auto_import_policy)
         self.drive_cache_type = Resource.init_param(drive_cache_type)
         self.file_system_type = LUSTRE
+        self.file_system_type_version = "2.12" if backup_id is None else None
```
chenwany: What happen if file_system_id is provided?

hanwen-pcluste: Good catch.
Although the self.file_system_type_version is ignored when `file_system_id` is provided, I improve the code to not provide any defaults if  `file_system_id` is provided

chenwany (approved):

PR #4100, opened 2022-06-10
Not require PlacementGroup/Enabled to be set to true when using PlacementGroup/Id with EFA

chenwany:
### Description of changes
* Do not require PlacementGroup/Enabled to be set to true when passing an existing PlacementGroup/Id. Fixed EfaPlacementGroupValidator to not required PlacementGroup/Enabled to be set to true when using PlacementGroup/Id with EFA enabled

### Tests
* unit test
* manual test
  * Ok:
Enabled=true
Id=1234
  * Ok:
Id=1234
  * Error:
Enabled=false
Id=1234
  * Ok:
Enabled=false
#Id=1234
  * Suppress validator, id is not used
Enabled=false
Id=1234
### References

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -1324,7 +1324,9 @@ class ComputeFleetConstruct(Construct):
 
             queue_placement_group = None
             if queue.networking.placement_group:
-                if queue.networking.placement_group.id:
+                if queue.networking.placement_group.id and (
+                    queue.networking.placement_group.enabled or queue.networking.placement_group.is_implied("enabled")
+                ):  # Do not use `PlacementGroup/Id` when `PlacementGroup/Enabled` is explicitly set to `false`
                     queue_placement_group = queue.networking.placement_group.id
                 elif queue.networking.placement_group.enabled:
                     queue_placement_group = managed_placement_groups[queue.name].ref
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -288,7 +288,13 @@ class EfaValidator(Validator):
 class EfaPlacementGroupValidator(Validator):
     """Validate placement group if EFA is enabled."""
 
-    def _validate(self, efa_enabled, placement_group_enabled, placement_group_config_implicit):
+    def _validate(self, efa_enabled, placement_group):
+        placement_group_enabled = placement_group and (
+            placement_group.enabled or (placement_group.id and placement_group.is_implied("enabled"))
+        )
+        placement_group_config_implicit = placement_group is None or (
+            placement_group.is_implied("enabled") and placement_group.id is None
+        )
         if efa_enabled and placement_group_config_implicit:
             self._add_failure(
                 "The placement group for EFA-enabled compute resources must be explicit. "
--- a/cli/src/pcluster/validators/ec2_validators.py
+++ b/cli/src/pcluster/validators/ec2_validators.py
@@ -102,12 +102,20 @@ class KeyPairValidator(Validator):
 class PlacementGroupIdValidator(Validator):  # TODO: add tests
     """Placement group id validator."""
 
-    def _validate(self, placement_group_id: str):
-        if placement_group_id:
-            try:
-                AWSApi.instance().ec2.describe_placement_group(placement_group_id)
-            except AWSClientError as e:
-                self._add_failure(str(e), FailureLevel.ERROR)
+    def _validate(self, placement_group):
+        if placement_group.id:
+            if not placement_group.is_implied("enabled") and not placement_group.enabled:
+                self._add_failure(
+                    "PlacementGroup Id can not be set when setting 'Enabled: false' in queue's "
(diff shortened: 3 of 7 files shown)


cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -195,26 +195,25 @@ def test_efa_validator(mocker, boto3_stubber, instance_type, efa_enabled, gdr_su
...
-        (True, False, False, "may see better performance using a placement group"),
-        (True, False, True, "placement group for EFA-enabled compute resources must be explicit"),
-        (True, True, True, "placement group for EFA-enabled compute resources must be explicit"),
-        (True, True, False, None),
+        (True, None, "placement group for EFA-enabled compute resources must be explicit"),
+        (True, PlacementGroup(), "placement group for EFA-enabled compute resources must be explicit"),
+        (True, PlacementGroup(enabled=True), None),
+        (True, PlacementGroup(id="test"), None),
+        (True, PlacementGroup(enabled=True, id="test"), None),
+        (True, PlacementGroup(enabled=False), "may see better performance using a placement group for the queue"),
```
EddyMM, line 212: Could we test a scenario where the PlacementGroup is specified but we explicitly disable it?
```
(True, PlacementGroup(enabled=False, Id="test"), "may see better performance using a placement group for the queue")
```

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -288,7 +288,11 @@ def _validate(self, instance_type, efa_enabled, gdr_support):
 class EfaPlacementGroupValidator(Validator):
     """Validate placement group if EFA is enabled."""
 
-    def _validate(self, efa_enabled, placement_group_enabled, placement_group_config_implicit):
+    def _validate(self, efa_enabled, placement_group):
+        placement_group_enabled = placement_group and (placement_group.enabled or placement_group.id)
```
EddyMM: This will evaluate to `true` if the PlacementGroup/enabled is set to `false` but the name/id is specified, right?

EddyMM (commented):
- Suggested a test scenario in which the PlacementGroup/enabled is set to `false` but the PlacementGroup/Id is specified

cli/tests/pcluster/validators/test_cluster_validators.py
chenwany: The current behavior is that, when enabled=False and Id="test", it is still using the placement group in the id for the queue

chenwany: After our discussion, changing the current behavior and add a test

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -102,12 +102,20 @@ def _validate(self, key_name: str):
...
-            try:
-                AWSApi.instance().ec2.describe_placement_group(placement_group_id)
-            except AWSClientError as e:
-                self._add_failure(str(e), FailureLevel.ERROR)
+    def _validate(self, placement_group):
+        if placement_group.id:
+            if not placement_group.is_implied("enabled") and not placement_group.enabled:
+                self._add_failure(
+                    "PlacementGroup Id can not be set when setting 'Enabled: false' in compute "
+                    "resource's configuration section. Please remove PlacementGroup Id if you don't want"
```
EddyMM: NIT: The placement group is in the Queue's [Networking Section](https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html#Scheduling-v3-SlurmQueues-Networking).

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1324,7 +1324,9 @@ def _add_launch_templates(self, managed_placement_groups, instance_profiles):
 
             queue_placement_group = None
             if queue.networking.placement_group:
-                if queue.networking.placement_group.id:
+                if queue.networking.placement_group.id and (
```
EddyMM, line 1327: Could you kindly add a comment on why we're checking the `enabled` property of the PlacementGroup here?
If a customer defines the PlacementGroup Name and it's been validated, we can go ahead and use it, right?

EddyMM (commented):
The cluster_stack change needs some more clarification on why we're checking for the PlacementGroup/enabled property.

cli/src/pcluster/templates/cluster_stack.py
chenwany: In the cluster_stack.py, I add the check there because in case customers skip the Validators, and set `PlacementGroup/Id` and `PlacementGroup/Enabled` is explicitly set to `false`, the placementgroup id will not be used

cli/src/pcluster/validators/ec2_validators.py
chenwany: Done

cli/src/pcluster/templates/cluster_stack.py
EddyMM: I see, that makes sense, Thanks!

EddyMM (approved):

hanwen-pcluste (approved):

PR #4115, opened 2022-06-15
Fixed bug that prevented generating log widgets

HWong-sudo:
### Description of changes
* Added conditional statement to function "_is_logs_section_empty". Original code was preventing a log widget with no pre-existing conditions from being formed.   

### Tests
* Passed unit tests on PCluster2 


HWong-sudo:
Discovered error when trying to create log widget with no specified condition. Solved by editing function "_is_logs_section_empty" . Before edits if you wanted to create a log with no conditions, not possible to reach the desired empty_section = false in specified function. Now possible to create log with no condition while still checking valid for conditions  on sections where condition exists

chenwany:
Good catch! Please also update the changelog to include the the fix.

HWong-sudo:
Edited comment: 

Fixed bug to allow for cloud watch log without log stream conditions

chenwany (dismissed):
LGTM!

enrico-usai (changes requested):
@HWong-sudo thanks for creating this patch!
Before merging it could you please update the PR description and the commit message to match the following rules: 
* [Limit the subject line to 50 characters](https://cbea.ms/git-commit/#limit-50)
* [Capitalize the subject line](https://cbea.ms/git-commit/#capitalize)
* [Use the imperative mood in the subject line](https://cbea.ms/git-commit/#imperative)
* [Use the body to explain what and why vs. how](https://cbea.ms/git-commit/#why-not-how)

For example the title might be something like: `Permit creation of CW Dashboard Widgets without conditions for the Log Stream`

Then in general when creating a PR we should populate the template sections: 
* Description of changes -  what you're changing and why you're doing these changes
* Tests - Describe the automated and/or manual tests executed to validate the patch.
* References (if available)
* Checklist (the flags must be checked after verifying they have been addressed)

@chenwany I think in this case we can avoid to add a line in the changelog because it's an internal change required for future extensions, we're not fixing an user facing bug.

enrico-usai (approved):

PR #4116, opened 2022-06-17
Add threshold check to benchmark fixture

francesco-giordano:
### Description of changes
* Add to the benchmark fixture a flag to check the OSU benchmark result with the result of 32 node cluster of c5n.18xlarge. The method employed sets the threshold to the average plus five times the standard deviation

### Tests
* Manual Tests in test_disable_multithreading

### DIFF
--- a/tests/integration-tests/tests/trainium/test_trainium/test_trainium/neuron-installation.sh
+++ b/tests/integration-tests/tests/trainium/test_trainium/test_trainium/neuron-installation.sh
@@ -44,7 +44,10 @@ EOF
 
   # Install packages from S3 --> FIXME they should be installed from configured repository
   DEBS=$(aws secretsmanager get-secret-value --secret-id arn:aws:secretsmanager:us-east-1:447714826191:secret:TrainiumPreviewRepository --region us-east-1 --query 'SecretString' --output text | jq -r '.debs')
-  sudo dpkg -i ${DEBS}
+  for DEB in $DEBS
+  do
+    sudo dpkg -i $DEB
+  done
 
   # Install Python venv and activate Python virtual environment to install Neuron pip packages.
   local OS_VERSION="$(grep "^VERSION_ID=" /etc/os-release | cut -d"=" -f 2 | xargs)"
@@ -77,7 +80,10 @@ EOF
 
   # Install packages from S3 --> FIXME they should be installed from configured repository
   RPMS=$(aws secretsmanager get-secret-value --secret-id arn:aws:secretsmanager:us-east-1:447714826191:secret:TrainiumPreviewRepository --region us-east-1 --query 'SecretString' --output text | jq -r '.rpms')
-  sudo rpm -i ${RPMS}
+  for RPM in $RPMS
+  do
+    sudo rpm -i $RPM
+  done
 
   python3 -m venv /home/ec2-user/aws_neuron_venv_pytorch
 }


tests/integration-tests/conftest.py
```diff
@@ -1394,12 +1397,23 @@ def _run_benchmarks(remote_command_executor, scheduler_commands, **kwargs):
...
                     )
+                    # Load reference results to confront benchmark and Assert failure
+                    if check_result:
+                        reference_results = results_csv_to_dict()
+                        check_errors = []
+
                     for metric_data in metric_data_list:
+                        # Check threshold
+                        if check_result:
+                            check_errors = check_threshold(reference_results, metric_data)
```
enrico-usai: minor: check_threshold --> check_tresholds ?

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
     return job_id, output
 
 
+def results_csv_to_dict() -> dict:
+    filename = OSU_COMMON_DATADIR / "reference_results.csv"
```
enrico-usai: let's define a constant for this file, close to the `OSU_COMMON_DATADIR` one.

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+    dict_result = {}
+    with open(filename, newline="", encoding="utf-8-sig") as results:
+        reader = csv.DictReader(results, delimiter=";")
+        for row in reader:
+            mpi_version = row["mpi_version"]
+            test_name = row["test_name"]
+            packet_size = row["packet_size"]
+            if not dict_result.get(mpi_version):
```
enrico-usai: nit: I?d add a space after variable initialization, to have a better separation between variable initialization and core calculation

enrico-usai: you can replace the IFs with:
```python
dict_result.setdefault(mpi_version, {})
dict_result[mpi_version].setdefault(test_name, {})
dict_result[mpi_version][test_name].setdefault(packet_size, {})
```
`setdefault` will set the value for that key only if it doesn't exist.

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+            dict_result[mpi_version][test_name][packet_size]["average"] = float(
+                row["average"].replace(".", "").replace(",", ".")
+            )
+            dict_result[mpi_version][test_name][packet_size]["std"] = float(
+                row["std"].replace(".", "").replace(",", ".")
+            )
+    return dict_result
+
+
+def check_threshold(reference_results, metric_data, variability_factor=5) -> list:
```
enrico-usai: let's add a docstring to explain what variability_factor is.

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+            test_name = row["test_name"]
+            packet_size = row["packet_size"]
+            if not dict_result.get(mpi_version):
+                dict_result[mpi_version] = {}
+            if not dict_result[mpi_version].get(test_name):
+                dict_result[mpi_version][test_name] = {}
+            if not dict_result[mpi_version][test_name].get(packet_size):
+                dict_result[mpi_version][test_name][packet_size] = {}
+
+            dict_result[mpi_version][test_name][packet_size]["average"] = float(
```
enrico-usai: minor: I'd create an internal function to perform this conversion:

```python
def _value_to_float(value):
  return float(value.replace(".", "").replace(",", "."))
```

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+            dict_result[mpi_version][test_name][packet_size]["std"] = float(
+                row["std"].replace(".", "").replace(",", ".")
+            )
+    return dict_result
+
+
+def check_threshold(reference_results, metric_data, variability_factor=5) -> list:
+    check_errors = []
+    for metric in metric_data:
+        if metric["MetricName"] == "Latency":
```
enrico-usai: let's add a comment before the for/ifs to clarify what we're doing in the next lines.

enrico-usai: minor: I'd use `dimension` in place of `d` to be self explanatory.

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+            for d in metric["Dimensions"]:
```
enrico-usai: I'd define an internal variable for values we're checking more than 1 time. E.g.
```python
for dimension_name, value in metric["Dimensions"].items():
   name = dimension["Name"]
   value = str(dimension["Value"])

    if name == "PacketSize":
        packet_size = value
    elif name == ...
        test_name = value
...
```

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+                if d["Name"] == "PacketSize":
+                    packet_size = str(d["Value"])
+                elif d["Name"] == "OsuBenchmarkName":
+                    test_name = d["Value"]
+                elif d["Name"] == "MpiVariant":
+                    mpi_variant = d["Value"]
+        try:
+            threshold = reference_results[mpi_variant][test_name][packet_size]["average"] + (
```
enrico-usai: I'd add a comment here too, describing what we're doing

tests/integration-tests/tests/common/osu_common.py
```diff
@@ -104,6 +105,60 @@ def run_individual_osu_benchmark(
...
+                variability_factor * reference_results[mpi_variant][test_name][packet_size]["std"]
+            )
+        except KeyError:
```
enrico-usai: if you want to avoid KeyError you can use `d.get("key")` rather than `d["key"]`.

tests/integration-tests/conftest.py
```diff
@@ -1394,12 +1397,23 @@ def _run_benchmarks(remote_command_executor, scheduler_commands, **kwargs):
...
+
                         cloudwatch_client.put_metric_data(
                             Namespace=metric_namespace,
                             MetricData=metric_data,
                         )
         logging.info("Finished benchmarks for %s", function_name)
+        if check_result:
```
enrico-usai: let's remove this if and always return check_errors even if it's an empty list and let's add in the docstring of the function what we're returning. In this way the caller can have a consistent workflow in both the cases.

tests/integration-tests/conftest.py
```diff
@@ -1354,7 +1354,7 @@ def to_pascal_case(snake_case_word):
 
 
 @pytest.fixture()
-def run_benchmarks(request, mpi_variants, test_datadir, instance, os, region, benchmarks):
+def run_benchmarks(request, mpi_variants, test_datadir, instance, os, region, benchmarks):  # noqa: C901
```
enrico-usai: The C901 is an alert to suggest us to try to refactor the code into a subfunctions, because probably we have too many if/else branches or loops

tests/integration-tests/conftest.py
enrico-usai: I'd move this logic in the `if key == "check_result":` to avoid an extra by rewriting the code with:

```python
if key == "check_result" and value:
     reference_results = results_csv_to_dict()
```

and moving check_errors initialization outside of all the loops, in this way you'll have a check_errors empty list to return in any case and make the function consistent (always returning the same info).

tests/integration-tests/conftest.py
```diff
@@ -1394,12 +1397,23 @@ def _run_benchmarks(remote_command_executor, scheduler_commands, **kwargs):
                         test_datadir,
                         dimensions,
                     )
+                    # Load reference results to confront benchmark and Assert failure
+                    if check_result:
+                        reference_results = results_csv_to_dict()
+                        check_errors = []
```
enrico-usai: I'd rename `check_errors`  to `failing_benchmarks`. 

`check` seems an action we want to perform on the errors.

tests/integration-tests/conftest.py
```diff
@@ -1394,12 +1397,23 @@ def _run_benchmarks(remote_command_executor, scheduler_commands, **kwargs):
                         test_datadir,
                         dimensions,
                     )
+                    # Load reference results to confront benchmark and Assert failure
+                    if check_result:
+                        reference_results = results_csv_to_dict()
```
hanwen-pcluste: Minor: the idea was to keep `run_benchmarks` fixture agnostic to the benchmark type. So an alternative is to move the logic here to `osu_common.py`

tests/integration-tests/conftest.py
francesco-giordano: Done

tests/integration-tests/tests/common/osu_common.py
francesco-giordano: Done

tests/integration-tests/conftest.py
hanwen-pcluste: After offline discussion, we decided to keep it as it is and improve later.

PR #4118, opened 2022-06-17
Prompt user to enable EFA for supported instance-types when using the pcluster configure wizard

EddyMM:
### Description of changes
* Users who create a cluster configuration using `pcluster configure` have minimal awareness of EFA being able to optimise performance for supported instance types
* These changes aim to address this by taking the following actions when a user specifies an EFA-supported compute instance type in `pcluster configure`:
   * Explain the benefits of enabling EFA
   * Prompt the user to enable EFA (or choose not to)
   * If the user enables EFA, they will be prompted to either provide an existing Placement Group or let parallelCluster create one for them

### Tests
* Tests have been added for the EFA prompt workflow to confirm the appropriate configuration is generated
  * If EFA is enabled, and existing placement group is provided the appropriate output and configuration file is generated
  * If EFA is enabled, and non-existent placement group is provided, the user is informed that the Placement Group is unknown and requested to try again
  * If EFA is enabled, and no placement group is provided (ParallelCluster will create one by default) the appropriate output and configuration file is generated
  * If EFA is NOT enabled, the user is not prompted to provide a Placement Group
  * If EFA is NOT enabled, the appropriate output and configuration file is generated

### References
* [ParallelCluster EFA](https://docs.aws.amazon.com/parallelcluster/latest/ug/efa.html)
* [AWS EFA](https://aws.amazon.com/hpc/efa/)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ x.x.x
 - Show `requested_value` and `current_value` values in the change set when adding or removing a section.
 - Add new configuration parameter `Scheduling/SlurmSettings/EnableMemoryBasedScheduling` to configure memory-based
   scheduling in Slurm.
+- Prompt user to enable EFA for supported instance types when using `pcluster configure` wizard.
 
 **CHANGES**
 - Remove support for Python 3.6.
--- a/cli/src/pcluster/cli/commands/configure/easyconfig.py
+++ b/cli/src/pcluster/cli/commands/configure/easyconfig.py
@@ -31,6 +31,8 @@ from pcluster.cli.commands.configure.utils import (
     get_regions,
     get_resource_tag,
     handle_client_exception,
+    instance_type_supports_efa,
+    placement_group_exists,
     prompt,
     prompt_iterable,
 )
@@ -196,7 +198,10 @@ def configure(args):  # noqa: C901
                 )
             )
         compute_resources = []
+        efa_enabled_in_queue = False
         for compute_resource_index in range(number_of_compute_resources):
+            efa_enabled_in_compute_resource = False
+            efa_supported_by_instance_type = False
             if scheduler != "awsbatch":
                 while True:
                     compute_instance_type = prompt(
@@ -213,6 +218,12 @@ def configure(args):  # noqa: C901
                         "resources in the same queue. Please insert a different instance type."
                     )
                 compute_resource_name = re.sub(r"[^A-Za-z0-9]", "", compute_instance_type)
+
+                efa_supported_by_instance_type = instance_type_supports_efa(compute_instance_type)
+                if efa_supported_by_instance_type:
+                    efa_enabled_in_compute_resource = _prompt_for_efa()
+                    if efa_enabled_in_compute_resource:
+                        efa_enabled_in_queue = True
             min_cluster_size = DEFAULT_MIN_COUNT
             max_cluster_size = int(
                 prompt(
@@ -232,25 +243,44 @@ def configure(args):  # noqa: C901
                     }
                 )
             else:
-                compute_resources.append(
-                    {
-                        "Name": compute_resource_name,
-                        "InstanceType": compute_instance_type,
-                        "MinCount": min_cluster_size,
-                        "MaxCount": max_cluster_size,
-                    }
-                )
+                compute_resource = {
+                    "Name": compute_resource_name,
+                    "InstanceType": compute_instance_type,
+                    "MinCount": min_cluster_size,
+                    "MaxCount": max_cluster_size,
+                }
+                if efa_supported_by_instance_type:
+                    compute_resource["Efa"] = {"Enabled": efa_enabled_in_compute_resource}
+
+                compute_resources.append(compute_resource)
                 compute_instance_types.append(compute_instance_type)
 
             queue_names.append(queue_name)
             cluster_size += max_cluster_size  # Fixme: is it the right calculation for awsbatch?
-        queues.append({"Name": queue_name, "ComputeResources": compute_resources})
+
+        queue = {
+            "Name": queue_name,
+            "ComputeResources": compute_resources,
+        }
+        if efa_enabled_in_queue:
+            placement_group = {"Enabled": True}
+            placement_group_name = _prompt_for_placement_group()
+            if placement_group_name:
+                placement_group["Id"] = placement_group_name
+
+            networking = queue.get("Networking", {})
+            networking["PlacementGroup"] = placement_group
+            queue["Networking"] = networking
+
+        queues.append(queue)
 
     vpc_parameters = _create_vpc_parameters(scheduler, head_node_instance_type, compute_instance_types, cluster_size)
 
     # Here is the end of prompt. Code below assembles config and write to file
     for queue in queues:
-        queue["Networking"] = {"SubnetIds": [vpc_parameters["compute_subnet_id"]]}
+        networking = queue.get("Networking", {})
+        networking["SubnetIds"] = [vpc_parameters["compute_subnet_id"]]
+        queue["Networking"] = networking
 
     head_node_config = {
         "InstanceType": head_node_instance_type,
@@ -401,9 +431,27 @@ def _prompt_for_subnet(all_subnets, qualified_subnets, message, default_subnet=N
     return prompt_iterable(message, qualified_subnets, default_value=default_subnet)
 
 
+def _prompt_for_efa():
+    print(
+        "To get results faster with the instance selected at no additional charge, enable the "
+        "Elastic Fabric Adapter (https://docs.aws.amazon.com/parallelcluster/latest/ug/efa.html)"
+    )
+    enable_efa = prompt(
+        "Enable EFA on compute instances (EFA) (y/n)", validator=lambda x: x in ("y", "n"), default_value="y"
+    )
+    return enable_efa == "y"
+
+
+def _prompt_for_placement_group():
+    print(
+        "Enabling EFA requires compute instances to be placed within a Placement Group, Specify an existing "
+        "Placement Group name or leave blank for ParallelCluster to create one"
+    )
+
+    return prompt("Placement Group name", validator=lambda x: x == "" or placement_group_exists(x), default_value="")
+
+
 # Availability zone utilities
-
-
 def _get_common_supported_az_for_multi_instance_types(instance_types):
     supported_az = AWSApi.instance().ec2.get_supported_az_for_instance_types(instance_types)
     common_az = None
--- a/cli/src/pcluster/cli/commands/configure/utils.py
+++ b/cli/src/pcluster/cli/commands/configure/utils.py
@@ -17,6 +17,8 @@ import boto3
 from botocore.exceptions import BotoCoreError, ClientError
 from tabulate import tabulate
 
+from pcluster.aws.aws_api import AWSApi
+from pcluster.aws.common import AWSClientError
 from pcluster.constants import SUPPORTED_REGIONS
 
 LOGGER = logging.getLogger(__name__)
@@ -27,7 +29,7 @@ def handle_client_exception(func):
     def wrapper(*args, **kwargs):  # pylint: disable=inconsistent-return-statements
         try:
             return func(*args, **kwargs)
-        except (BotoCoreError, ClientError) as e:
+        except (BotoCoreError, ClientError, AWSClientError) as e:
             print(f"Failed with error: {e}")
             if isinstance(e, ClientError) and "credentials" in str(e):
                 print("To set the credentials, run 'aws configure' or set them as environment variables")
@@ -201,6 +203,28 @@ def get_regions():
     regions = [region.get("RegionName") for region in regions if region.get("RegionName") in SUPPORTED_REGIONS]
     regions.sort()
     return regions
+
+
+@handle_client_exception
+def instance_type_supports_efa(instance_type: str) -> bool:
+    """Return whether of not a specified instance type supports EFA.
+
+    It uses the instance type information returned from AWS API.
+    :param instance_type:
+    :return: True if instance type support EFA, False otherwise
+    """
+    return AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported()
+
+
+@handle_client_exception
+def placement_group_exists(placement_group_name: str) -> bool:
+    """Return whether of not a specified placement group exists based on the placement group name.
+
+    It does so by querying the AWS EC2 API for the placement group by name.
+    :param placement_group_name: Name of the placement group
+    :return: Placement Group if it exists, None if placement group does not exist or unknown in a region
+    """
+    return AWSApi.instance().ec2.describe_placement_group(placement_group_name)
 
 
 def get_resource_tag(resource, tag_name):
--- a/cli/tests/pcluster/cli/configure/test_pcluster_configure.py
+++ b/cli/tests/pcluster/cli/configure/test_pcluster_configure.py
@@ -301,6 +301,7 @@ def _mock_aws_api_required_calls(mocker):
         "g3.8xlarge",
         "m6g.xlarge",
         "p4d.24xlarge",
+        "c5n.18xlarge",
     ]
     mock_aws_api(mocker, mock_instance_type_info=False)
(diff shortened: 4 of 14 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ x.x.x
 - Show `requested_value` and `current_value` values in the change set when adding or removing a section.
 - Add new configuration parameter `Scheduling/SlurmSettings/EnableMemoryBasedScheduling` to configure memory-based
   scheduling in Slurm.
+- Prompt user to enable EFA for supported instance types when using `pcluster configure` wizard
```
enrico-usai: minor: missing final period.

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
-                    }
-                )
+                compute_resource = {
+                    "Name": compute_resource_name,
+                    "InstanceType": compute_instance_type,
+                    "MinCount": min_cluster_size,
+                    "MaxCount": max_cluster_size,
+                }
+                if efa_supported_by_instance_type:
+                    if efa_enabled_in_compute_resource:
```
enrico-usai: minor: you can rewrite the entire if/else block to:
```python
compute_resource["Efa"] = {"Enabled": efa_enabled_in_compute_resource}
```

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
                 compute_instance_types.append(compute_instance_type)
 
             queue_names.append(queue_name)
             cluster_size += max_cluster_size  # Fixme: is it the right calculation for awsbatch?
-        queues.append({"Name": queue_name, "ComputeResources": compute_resources})
+
+        placement_group = {}
+        if efa_enabled_in_queue:
+            placement_group_name = _prompt_for_placement_group()
+            if placement_group_name:
```
enrico-usai: minor: to avoid code duplication you can write:
```python
placement_group = {"Enabled": True}
if placement_group_name:
    placement_group["Id"] = placement_group_name

```

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
+        if placement_group:
+            queue.update({"Networking": {"PlacementGroup": placement_group}})
+        queues.append(queue)
 
     vpc_parameters = _create_vpc_parameters(scheduler, head_node_instance_type, compute_instance_types, cluster_size)
 
     # Here is the end of prompt. Code below assembles config and write to file
     for queue in queues:
-        queue["Networking"] = {"SubnetIds": [vpc_parameters["compute_subnet_id"]]}
+        if "Networking" in queue:
```
enrico-usai: an alternative to avoid if/else is:
```python
networking = queue.get("Networking", {})
networking["SubnetIds"] = [vpc_parameters["compute_subnet_id"]]
queue["Networking"] = networking
```

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
+            if placement_group_name:
+                placement_group = {"Enabled": True, "Id": placement_group_name}
+            else:
+                placement_group = {"Enabled": True}
+        queue = {
+            "Name": queue_name,
+            "ComputeResources": compute_resources,
+        }
+        if placement_group:
+            queue.update({"Networking": {"PlacementGroup": placement_group}})
```
enrico-usai: If networking contains other fields they will be removed, we should use the same logic used below

cli/src/pcluster/cli/commands/configure/utils.py
```diff
@@ -203,6 +205,32 @@ def get_regions():
...
+    return AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported()
+
+
+@handle_client_exception
+def placement_group_exists(placement_group_name: str) -> bool:
+    """Return whether of not a specified placement group exists based on the placement group name.
+
+    It does so by querying the AWS EC2 API for the placement group by name.
+    :param placement_group_name: Name of the placement group
+    :return: True if placement group exists, False if placement group does not exist or unknown ina  region
```
enrico-usai: minor: typo at the end

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
+            placement_group_name = _prompt_for_placement_group()
+            if placement_group_name:
+                placement_group = {"Enabled": True, "Id": placement_group_name}
+            else:
+                placement_group = {"Enabled": True}
+        queue = {
+            "Name": queue_name,
+            "ComputeResources": compute_resources,
+        }
+        if placement_group:
```
enrico-usai: you could move the initialization of the queue dict before the `if efa_enabled_in_queue` block, then move this `if placement_group` into the `if efa_enabled_in_queue` block.

cli/src/pcluster/cli/commands/configure/utils.py
```diff
@@ -203,6 +205,32 @@ def get_regions():
     return regions
 
 
+@handle_client_exception
```
demartinofra, line 208: `AWSApi.instance().ec2.get_instance_type_info` is already decorated with `@AWSExceptionHandler.handle_client_exception` which translates botocore exceptions in internal exceptions. I'm wondering if the handle_client_exception here works as you expect it to.

cli/src/pcluster/cli/commands/configure/utils.py
```diff
@@ -203,6 +205,32 @@ def get_regions():
...
+    """Return whether of not a specified instance type supports EFA.
+
+    It uses the instance type information returned from AWS API.
+    :param instance_type:
+    :return: True if instance type support EFA, False otherwise
+    """
+    return AWSApi.instance().ec2.get_instance_type_info(instance_type).is_efa_supported()
+
+
+@handle_client_exception
```
demartinofra, line 219: same here

cli/src/pcluster/cli/commands/configure/utils.py
```diff
@@ -203,6 +205,32 @@ def get_regions():
...
+    """Return whether of not a specified placement group exists based on the placement group name.
+
+    It does so by querying the AWS EC2 API for the placement group by name.
+    :param placement_group_name: Name of the placement group
+    :return: True if placement group exists, False if placement group does not exist or unknown ina  region
+    """
+    try:
+        return AWSApi.instance().ec2.describe_placement_group(placement_group_name)
+    except AWSClientError as e:
+        print(f"Failed with error: {e}")
```
enrico-usai: I don't think we want to print the error here. Should be in charge of @handle_client_exception

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
+                    "Name": compute_resource_name,
+                    "InstanceType": compute_instance_type,
+                    "MinCount": min_cluster_size,
+                    "MaxCount": max_cluster_size,
+                }
+                if efa_supported_by_instance_type:
+                    if efa_enabled_in_compute_resource:
+                        compute_resource["Efa"] = {"Enabled": True}
+                    else:
+                        compute_resource["Efa"] = {"Enabled": False}
```
demartinofra: nit: it's a matter of personal taste but I find more readable a compact form like:

```
compute_resource = {
                    "Name": compute_resource_name,
                    "InstanceType": compute_instance_type,
                    "MinCount": min_cluster_size,
                    "MaxCount": max_cluster_size,
                    "Efa": { "Enabled": efa_supported_by_instance_type and efa_enabled_in_compute_resource }
                }
```

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -232,25 +243,47 @@ def configure(args):  # noqa: C901
...
+
+                compute_resources.append(compute_resource)
                 compute_instance_types.append(compute_instance_type)
 
             queue_names.append(queue_name)
             cluster_size += max_cluster_size  # Fixme: is it the right calculation for awsbatch?
-        queues.append({"Name": queue_name, "ComputeResources": compute_resources})
+
+        placement_group = {}
```
demartinofra: Dropping it here but it's more of a general comment: this configure function is getting extremely long and hard to parse. Not asking to refactor the entire code as part of this PR cause the effort might be too big but if you could come up with some partial improvements especially around the areas you are modifying it would be beneficial for code quality

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -401,9 +434,27 @@ def _prompt_for_subnet(all_subnets, qualified_subnets, message, default_subnet=N
...
+        "Enable EFA on compute instances (EFA) (y/n)", validator=lambda x: x in ("y", "n"), default_value="y"
+    )
+    return enable_efa == "y"
 
 
+def _prompt_for_placement_group():
+    print(
+        "Enabling EFA requires compute instances to be placed within a Placement Group, Specify an existing "
+        "Placement Group name or leave blank for ParallelCluster to create one"
+    )
```
demartinofra, line 449: What if I don't want a PG? shall we allow Y/N or PG name selection?

cli/tests/pcluster/cli/configure/test_pcluster_configure.py
```diff
@@ -689,14 +692,88 @@ def general_wrapper_for_prompt_testing(
...
+    input_composer.add_first_flow(op_sys, head_node_instance)
+    input_composer.add_compute_instance(compute_instance, max_size)
     input_composer.add_no_automation_no_empty_vpc(vpc_id, head_node_id, compute_id)
     input_composer.mock_input(mocker)
 
     _run_configuration(mocker, temp_path_for_config)
     return True
 
 
+def test_enabled_efa_default_placement_group(mocker, temp_path_for_config, capsys, test_datadir):
```
demartinofra: nit: if the newly added tests share most of the test code you could consider replacing them with a single parametrized test_efa test case where pg setting is a parameter

cli/tests/pcluster/cli/configure/test_pcluster_configure/test_enabled_efa_default_placement_group/output.txt
```diff
@@ -0,0 +1,57 @@
+INFO: Configuration file {{ CONFIG_FILE }} will be written.
```
demartinofra, line 1: Don't we capture and validate the questions/prompts as part of the unit testing? is there a technical limitation?

EddyMM: We only capture the output, errors and configs at the moment.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: - With the wizard workflow, we went for requiring a placement group with EFA.
- The placement group can always be set to `false` in the config file but for the initial configuration we're trying to get the most efficient setup

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: I agree. At the moment the `configure` function is not properly modularised. I tried to move the prompting logic to separate functions but a refactor may be needed later for other prompts and logic.

cli/src/pcluster/cli/commands/configure/utils.py
EddyMM: `@AWSExceptionHandler.handle_client_exception` will ultimately raise an error.
`@handle_client_exception` from `utils` will only catch the exception and print out the Client Error, which is the behaviour we need here.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: Yes, I'll keep the outer `if` though, to avoid defining the "Efa" property for instance-types that don't have support for it anyway.

```
if efa_supported_by_instance_type:
    compute_resource["Efa"] = {"Enabled": efa_enabled_in_compute_resource}
```

CHANGELOG.md
EddyMM: Done.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: Done.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: Done.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: Done.

cli/src/pcluster/cli/commands/configure/easyconfig.py
EddyMM: Done.

cli/src/pcluster/cli/commands/configure/utils.py
demartinofra: But are you sure the correct exception is catched by handle_client_exception from utils? Cause I see a mismatch between what @AWSExceptionHandler.handle_client_exception raises and what the other one handles

cli/tests/pcluster/cli/configure/test_pcluster_configure/test_enabled_efa_default_placement_group/output.txt
demartinofra: is there an easy way to also capture the prompted question? I think that would be extremely valuable to test with the added value of having the entire configure workflow dumped in a file

EddyMM: I can take a look at the mechanism used to capture the input to see if it's possible.

cli/src/pcluster/cli/commands/configure/utils.py
EddyMM: `handle_client_exception` from utils catches the AWSClientError exception raised from `@AWSExceptionHandler.handle_client_exception`. Only difference is that the utils version will handle it by printing the exception (instead of raising it the next level in the hierarchy & potentially ultimately exiting).

cli/tests/pcluster/cli/configure/test_pcluster_configure.py
```diff
@@ -689,14 +692,64 @@ def general_wrapper_for_prompt_testing(
...
     input_composer.mock_input(mocker)
 
     _run_configuration(mocker, temp_path_for_config)
     return True
 
 
+@pytest.mark.parametrize(
+    "test_dir_name, supports_efa, efa_enabled, placement_group_names, placement_group_exists",
+    [
+        ("test_enabled_efa_default_placement_group", True, "y", [""], lambda x: False),
```
enrico-usai, line 707: what about a test with supports_efa = False?

EddyMM: Most of the other tests in `test_pcluster_configure.py` cover such a scenario, but I can explicitly include it here to be sure.

EddyMM: Added a scenario where `supports_efa = False`

cli/src/pcluster/cli/commands/configure/utils.py
demartinofra: Oh ok sorry for some reason I had missed your change where you added AWSClientError to the except cause

enrico-usai (dismissed):

enrico-usai (approved):

PR #4134, opened 2022-06-22
Set the default root volume size according to AMIs being used

hanwen-pcluste:
1. Set the default root volume sizes of all nodes to the size of snapshot in the AMI used for each node respectively. The old behavior was to have a constant 35GB default
2. Users can overwrite the defaults by the `Size` parameter under `RootVolume`
3. Add a validator to make sure the `Size` specified by the user is equal or greater than the size of the snapshot in the AMI. The old validator to make sure the `Size` is greater than 35GB is removed.
4. Refactor the code by creating a common class between Slurm and Scheduler Plugin config to reduce code replication.

For example:
```
Region: us-east-1
Image:
  Os: alinux2
HeadNode:
  ...
Scheduling:
  Scheduler: slurm
  SlurmQueues:
  - Name: queue1
    ...
  - Name: queue2
    Image:
      CustomAmi: ami-00f25fb4d0adb70a5
    ...
```
where `ami-00f25fb4d0adb70a5` is a deep learning AMI. The head node and compute nodes in queue1 will have 35GB of root volume, the size of official ParallelCluster AMI. The compute nodes in queue2 will have 167GB of root volume, the size of deep learning AMI

### Description of changes

### Tests
* Added a unit test to test validator
* I didn't add any integration tests. Existing integration tests are running
* Manually tested the default volume size behaves as expected

### Checklist


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1078,6 +1078,20 @@ def _test_compute_resource(
...
 
+class RootVolumeSizeValidator(Validator):
+    """Verify the root volume size is equal or greater to the size of the snapshot of the AMI."""
+
+    def _validate(self, root_volume_size, ami_id):
+        if root_volume_size:
+            ami_volume_size = AWSApi.instance().ec2.describe_image(ami_id).volume_size
+            if root_volume_size < ami_volume_size:
+                self._add_failure(
+                    f"Root volume size {root_volume_size} GB must be equal or greater than the volume size of "
```
enrico-usai: should be GiB not GB

cli/src/pcluster/validators/ebs_validators.py
```diff
@@ -45,18 +45,19 @@ class EbsVolumeTypeSizeValidator(Validator):
...
-                self._add_failure(
-                    f"The size of {volume_type} volumes can not exceed {max_size} GiB.",
-                    FailureLevel.ERROR,
-                )
-            elif volume_size < min_size:
-                self._add_failure(
-                    f"The size of {volume_type} volumes must be at least {min_size} GiB.",
-                    FailureLevel.ERROR,
-                )
+        if volume_size is not None:
```
enrico-usai: you can compact the two ifs.

```python
if volume_size is not None and volume_type in ..:
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -218,9 +219,9 @@ def __init__(self, mount_dir: str = None):
 class LocalStorage(Resource):
     """Represent the entire node storage configuration."""
 
-    def __init__(self, root_volume: Ebs = None, ephemeral_volume: EphemeralVolume = None, **kwargs):
+    def __init__(self, root_volume: RootVolume = None, ephemeral_volume: EphemeralVolume = None, **kwargs):
```
gmarciani, line 222: Why not using `RootVolume(implied=True)` as the default value for `root_volume`, instead of `None`?
Asking because, according to line 224 it seems that this is the desired behaviour.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -248,6 +250,7 @@ def __init__(
         **kwargs,
     ):
         super().__init__(**kwargs)
+        self.size = Resource.init_param(size, default=EBS_VOLUME_SIZE_DEFAULT)
```
gmarciani, line 253: Here we re using 35Gb as the root volume default size.
According to the PR description, it seems that we want to use the AMI snapshot size as default size.
Maybe I'm missing something, but the risk here is: if the customer does not specify the root volume size, it will fall back to 35Gb and the validators we are introducing may fail, whereas the customer expects it to succeed when leaving default values.
May you please explain?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1173,6 +1176,15 @@ def _register_validators(self):
         if self.custom_s3_bucket:
             self._register_validator(S3BucketValidator, bucket=self.custom_s3_bucket)
             self._register_validator(S3BucketRegionValidator, bucket=self.custom_s3_bucket, region=self.region)
+        self._register_validator(SchedulerOsValidator, scheduler=self.scheduling.scheduler, os=self.image.os)
+        self._register_validator(
+            HeadNodeImdsValidator, imds_secured=self.head_node.imds.secured, scheduler=self.scheduling.scheduler
+        )
```
gmarciani, line 1182: How are these validators related to the goal of this PR?
I was expecting to see only changes related to root volume, not to OS or IMDs.
May you please explain?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2192,7 +2200,48 @@ def _register_validators(self):
             )
 
 
-class SchedulerPluginClusterConfig(BaseClusterConfig):
+class CommonSlurmAndSchedulerPluginClusterConfig(BaseClusterConfig):
```
gmarciani: I got your point here and that's fine. Nevertheless I'd like to share with you some points.

1. The goal of this PR is about root volume default size, so it's always better to not mix up general/broader refactoring with feature core code changes. Better to split core changes and broader refactoring in distinct PRs to speed up iterations on the code.
2. It's not a good practice to name a class with a name that is defined by the name of the classes that will use it. If we end up doing it, sometimes it may be the signal of something wrong in the class/interface design. The name of a class should be defined by the class's responsibilities, not by the external classes that will make use of them. For instance: if a class is used by X and Y and we name it `CommonUsedByXAndY`, what if a new class Z will then need to make us of it too? We should end up renaming the class as `CommonUsedByXAndYAndZ` which is not good. In this specific case, it looks to me that the responsibility of this class is to define validators for scheduling queues, so the name could be `BaseSchedulingQueuesConfig`.

cli/src/pcluster/validators/ebs_validators.py
```diff
@@ -45,18 +45,19 @@ class EbsVolumeTypeSizeValidator(Validator):
...
+            if volume_type in EBS_VOLUME_TYPE_TO_VOLUME_SIZE_BOUNDS:
+                min_size, max_size = EBS_VOLUME_TYPE_TO_VOLUME_SIZE_BOUNDS.get(volume_type)
+                if volume_size > max_size:
+                    self._add_failure(
+                        f"The size of {volume_type} volumes can not exceed {max_size} GiB.",
+                        FailureLevel.ERROR,
+                    )
+                elif volume_size < min_size:
+                    self._add_failure(
+                        f"The size of {volume_type} volumes must be at least {min_size} GiB.",
```
gmarciani, line 58: Type: `GiB` => `GB`

cli/src/pcluster/config/cluster_config.py
enrico-usai: I think this name is not a good choice because we're adding the information about the classes that are inheriting it.

We can call it `CommonSchedulerClusterConfig` or `OnPremSchedulerClusterConfig` (if you want to highlight this is not for Batch) or something else.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1137,3 +1138,25 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
     )
     expected_message = "make sure.*cluster nodes are reachable" if expect_warning else None
     assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
+    "root_volume_size, expected_message, ami_size",
```
gmarciani: Minor: I think is better to put parameters in the following order: inputs followe by the expected outcome.
So, in this case: `root_volume_size, ami_size, expected_message`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1173,6 +1176,15 @@ def _register_validators(self):
         if self.custom_s3_bucket:
             self._register_validator(S3BucketValidator, bucket=self.custom_s3_bucket)
             self._register_validator(S3BucketRegionValidator, bucket=self.custom_s3_bucket, region=self.region)
+        self._register_validator(SchedulerOsValidator, scheduler=self.scheduling.scheduler, os=self.image.os)
+        self._register_validator(
+            HeadNodeImdsValidator, imds_secured=self.head_node.imds.secured, scheduler=self.scheduling.scheduler
```
enrico-usai, line 1181: When doing this kind of refactoring it's better to split the refactoring change from the new feature in two different commits.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1137,3 +1138,25 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
...
+    [
+        (65, None, 50),
+        (25, None, 8),
+        (
+            25,
+            "Root volume size 25 GB must be equal or greater than .* 50 GB.",
```
gmarciani: Better to put `the volume size of the AMI .* 50GB.` because the portion to wildcard is only the AMI ID.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1137,3 +1138,25 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
     )
     expected_message = "make sure.*cluster nodes are reachable" if expect_warning else None
     assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
+    "root_volume_size, expected_message, ami_size",
+    [
+        (65, None, 50),
+        (25, None, 8),
```
gmarciani: Why should we test `root_volume_size=25,ami_size=8` if we already have `root_volume_size=65,ami_size=50`?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -191,8 +191,9 @@ def _register_validators(self):
 class RootVolume(Ebs):
     """Represent the root volume configuration."""
 
-    def __init__(self, delete_on_termination: bool = None, **kwargs):
+    def __init__(self, size: int = None, delete_on_termination: bool = None, **kwargs):
         super().__init__(**kwargs)
+        self.size = Resource.init_param(size)
```
enrico-usai, line 196: Why did you remove the default size from EBS and so from the RootVolume?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1137,3 +1138,25 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
...
+    ],
+)
+def test_root_volume_size_validator(mocker, root_volume_size, expected_message, ami_size):
+    mock_aws_api(mocker)
+    mocker.patch(
+        "pcluster.aws.ec2.Ec2Client.describe_image",
+        return_value=ImageInfo({"BlockDeviceMappings": [{"Ebs": {"VolumeSize": ami_size}}]}),
+    )
+    actual_failures = RootVolumeSizeValidator().execute(root_volume_size, "ami-123456789a8a37250")
+    assert_failure_messages(actual_failures, expected_message)
```
gmarciani, line 1161: We are testing the failure messages, but we are not testing the failure type (ERROR vs WARNING).
I know that it the majority of our tests we do not cover it, but I think it's good to.
Let's consider this as an improvement out of scope for this PR, but let's be sure to track t.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1555,10 +1567,6 @@ def __init__(self, cluster_name: str, scheduling: AwsBatchScheduling, **kwargs):
     def _register_validators(self):
         super()._register_validators()
         self._register_validator(AwsBatchRegionValidator, region=self.region)
```
enrico-usai, line 1569: This kind of change of moving `SchedulerOsValidator` and `HeadNodeImdsValidator` from AwsBatchClusterConfig to BaseClusterConfig should stay in a separate commit.

cli/src/pcluster/validators/cluster_validators.py
enrico-usai, line 1089: Should be GiB not GB

cli/tests/pcluster/validators/test_cluster_validators.py
enrico-usai: nit: I think it's more readable to have as first parameters our "variables" so root_volume_size and ami_size and as third parameter the value we're checking (expected_message).

cli/src/pcluster/validators/ebs_validators.py
enrico-usai: You change the logic of this EbsVolumeTypeSizeValidator, can you add a test case for volume_size set to None?

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Because the instance root volume will be created with the size of AMI automatically. It is a EC2 default. We don't need to specify the root volume size in our code

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: [Not valid anymore] Good question. I didn't think about it. Our code has this style. Can we keep it to be aligned within our code?

After a deeper look, we have to use the current approach, because whatever default set in the parameter will be overwritten by the `None` passed from cluster_schema

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: This class is for the shared EBS, not the root volume. We have to give shared EBS a default size. Otherwise it will be a breaking change

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: It is a refactor. When I added RootVolumeSizeValidator, I noticed I had to add it in two places. So I did the refactor

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Thanks for the suggestion.

I split the PR in to two commits

cli/src/pcluster/validators/ebs_validators.py
hanwen-pcluste: Done

hanwen-pcluste: Done

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: Done

hanwen-pcluste: Done

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: I removed one of them

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: Using regex makes the test less likely to break if we make some minor change to the error message

cli/src/pcluster/validators/cluster_validators.py

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: I split the PR in to two commits

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: I split the PR in to two commits

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: I split the PR in to two commits

gmarciani: Thanks

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1078,6 +1078,20 @@ def _test_compute_resource(
...
+class RootVolumeSizeValidator(Validator):
+    """Verify the root volume size is equal or greater to the size of the snapshot of the AMI."""
+
+    def _validate(self, root_volume_size, ami_id):
+        if root_volume_size:
+            ami_volume_size = AWSApi.instance().ec2.describe_image(ami_id).volume_size
+            if root_volume_size < ami_volume_size:
+                self._add_failure(
+                    f"Root volume size {root_volume_size} GiB must be equal or greater than the volume size of "
+                    f"the AMI {ami_id}: {ami_volume_size} GiB. ",
```
enrico-usai: minor: trailing whitespace

gmarciani (dismissed):

enrico-usai (approved):

PR #4142, opened 2022-06-23
Fix bug where EbsVolumeIopsValidator compared NoneType with Int

hanwen-pcluste:
This bug came from https://github.com/aws/aws-parallelcluster/commit/fbb995d4221b56b62cb8d7edd5ffa91ee643f002, where the default of root volume size was removed to respect the size of AMIs.

Therefore, EbsVolumeIopsValidator could receive None as the `volume_size`, causing the bad comparison.

This commit moves two relevant validators from base class `Ebs` to other places: `ShareEbs` (to validate shared EBS), `BaseClusterConfig` (to validate head node root volume), `CommonSchedulerClusterConfig` (to validate compute nodes root volume).

This commit also improves RootVolumeSizeValidator to use `ami_volume_size` instead of `ami_id` because `ami_volume_size` is shared with multiple validators.

### References
* Issue introduced with: https://github.com/aws/aws-parallelcluster/pull/4134

### Checklist


cli/tests/pcluster/validators/test_ebs_validators.py
```diff
@@ -94,30 +94,24 @@ def test_ebs_volume_iops_validators(volume_type, volume_size, volume_iops, expec
         ("standard", 15, None),
         ("standard", 0, "The size of standard volumes must be at least 1 GiB"),
         ("standard", 1025, "The size of standard volumes can not exceed 1024 GiB"),
-        ("standard", None, None),
```
hanwen-pcluste, line 97: These were added in the previous PR. Now, the input parameter of the validator cannot be None, therefore I removed these tests

chenwany (approved):

PR #4153, opened 2022-06-30
[PCluster Configure] Add integration tests for enabling EFA through the pcluster configure wizard

EddyMM:
### Description of changes
* The `pcluster configure` command now includes prompts to enable EFA for supported instance-types
* These tests check to confirm if the wizard is able to generate the expected configuration file as well as successfully create a cluster with the same configuration

### Tests
* User is prompted to enable EFA for supported instance-type
* User is prompted to either provide an existing PlacementGroup or let PCluster create one for them
* A configuration file with expected properties is generated
* The configuration file can be used to successfully create a cluster

Result | Test | Duration | 
-- | -- | -- |
Passed | configure/test_pcluster_configure.py::test_efa_and_placement_group[us-east-1-c5n.9xlarge-alinux2-slurm-n-efa_config2-none] | 464.64 |  
Passed | configure/test_pcluster_configure.py::test_efa_and_placement_group[us-east-1-c5n.9xlarge-alinux2-slurm-y-efa_config0-default] | 492.55 |  
Passed | configure/test_pcluster_configure.py::test_efa_and_placement_group[us-east-1-c5n.9xlarge-alinux2-slurm-y-efa_config1-custom] | 466.77

### References
* [Prompt user to enable EFA for supported instance-types when using the pcluster configure wizard](https://github.com/aws/aws-parallelcluster/pull/4118)
* [ParallelCluster EFA](https://docs.aws.amazon.com/parallelcluster/latest/ug/efa.html)

### Checklist

### DIFF
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -110,6 +110,19 @@ configure:
         - regions: ["eu-north-1"] # must be regions that do not have t2.micro
           oss: ["centos7"]
           schedulers: ["slurm"]
+  test_pcluster_configure.py::test_efa_and_placement_group:
+    dimensions:
+      - regions: ["us-east-1"]
+        instances: {{ common.INSTANCES_EFA_SUPPORTED_X86 }}
+        oss: ["alinux2"]
+        schedulers: ["slurm"]
+  test_pcluster_configure.py::test_efa_unsupported:
+    dimensions:
+      - regions: ["us-east-1"]
+        instances: {{ common.INSTANCES_EFA_UNSUPPORTED_X86 }}
+        oss: ["alinux2"]
+        schedulers: ["slurm"]
+
 create:
   test_create.py::test_create_wrong_os:
     dimensions:
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -51,6 +51,7 @@ from jinja2 import Environment, FileSystemLoader
 from network_template_builder import Gateways, NetworkTemplateBuilder, SubnetConfig, VPCConfig
 from retrying import retry
 from troposphere import Ref, Template, ec2
+from troposphere.ec2 import PlacementGroup
 from troposphere.fsx import FileSystem, StorageVirtualMachine, Volume, VolumeOntapConfiguration
 from utils import (
     InstanceTypesData,
@@ -1244,6 +1245,25 @@ def network_interfaces_count(request, instance, region):
         network_interfaces_count = get_network_interfaces_count(instance, region)
         request.config.cache.set(f"{instance}/network_interfaces_count", network_interfaces_count)
     return network_interfaces_count
+
+
+@pytest.fixture(scope="class")
+def placement_group_stack(cfn_stacks_factory, request, region):
+    """Placement group stack contains a placement group."""
+    placement_group_template = Template()
+    placement_group_template.set_version()
+    placement_group_template.set_description("Placement group stack created for testing existing placement group")
+    placement_group_template.add_resource(PlacementGroup("PlacementGroup", Strategy="cluster"))
+    stack = CfnStack(
+        name=generate_stack_name("integ-tests-placement-group", request.config.getoption("stackname_suffix")),
+        region=region,
+        template=placement_group_template.to_json(),
+    )
+    cfn_stacks_factory.create_stack(stack)
+
+    yield stack
+
+    cfn_stacks_factory.delete_stack(stack.name, region)
 
 
 @pytest.fixture()
--- a/tests/integration-tests/tests/configure/test_pcluster_configure.py
+++ b/tests/integration-tests/tests/configure/test_pcluster_configure.py
@@ -19,6 +19,34 @@ import yaml
 from assertpy import assert_that
 from conftest import inject_additional_config_settings
 
+PROMPTS = {
+    "region": lambda region: {"prompt": r"AWS Region ID \[.*\]: ", "response": region},
+    "key_pair": lambda key_name: {"prompt": r"EC2 Key Pair Name \[.*\]: ", "response": key_name},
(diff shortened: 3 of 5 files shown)


tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -19,23 +19,48 @@
...
+        "response": f"{n}",
+        "skip_for_batch": True,
+    },
+    "compute_instance_type": lambda resource, queue_name, instance: {
+        "prompt": rf"Compute instance type for compute resource {resource} in {queue_name} \[t.\.micro\]: ",
+        "response": instance,
+        "skip_for_batch": True,
+    },
+    "enable_efa": lambda response: {
+        "prompt": r"Enable EFA on compute instances \(EFA\) \(y/n\) \[y\]:",
```
gmarciani: Why do we need the ending `(EFA)` in the prompt? It sounds redundant.

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -96,27 +117,149 @@ def test_region_without_t2micro(vpc_stack, pcluster_config_reader, key_name, reg
...
-        vpc_stack.cfn_outputs["VpcId"],
+        region,
         vpc_stack.cfn_outputs["PublicSubnetId"],
         vpc_stack.cfn_outputs["PrivateSubnetId"],
+        config_path,
     )
+
+
+@pytest.mark.parametrize(
+    "efa_response, efa_config, placement_group_response_type",
```
gmarciani, line 124: What do you mean with `*_response`?

tests/integration-tests/conftest.py
```diff
@@ -1246,6 +1247,25 @@ def network_interfaces_count(request, instance, region):
     return network_interfaces_count
 
 
+@pytest.fixture(scope="class")
+def placement_group_stack(cfn_stacks_factory, request, region):
```
gmarciani, line 1251: What is the benefit of moving this function from `tests/integration-tests/tests/networking/test_placement_group.py`?

tests/integration-tests/configs/common/common.yaml
```diff
@@ -110,6 +110,13 @@ configure:
         - regions: ["eu-north-1"] # must be regions that do not have t2.micro
           oss: ["centos7"]
           schedulers: ["slurm"]
+  test_pcluster_configure.py::test_efa_and_placement_group:
+    dimensions:
+      - regions: ["us-east-1"]
+        instances: {{ common.INSTANCES_EFA_SUPPORT_X86 }}
```
gmarciani: Shouldn't we test both the case of instances supporting EFA and instances not supporting it?

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -96,27 +117,149 @@ def test_region_without_t2micro(vpc_stack, pcluster_config_reader, key_name, reg
     In other words, t3.micro is retrieved when the region does not contain t2.micro
     """
     config_path = test_datadir / "config.yaml"
-    stages = orchestrate_pcluster_configure_stages(
-        region,
+    region_without_t2micro_prompts = (
```
gmarciani, line 104: Wy do we need to test region with absence of t2.micro?

tests/integration-tests/tests/configure/test_pcluster_configure.py
```diff
@@ -19,23 +19,48 @@
 from assertpy import assert_that
 from conftest import inject_additional_config_settings
 
+prompts = {
```
gmarciani: Since this is a constant, what about naming with uppercase?

tests/integration-tests/tests/configure/test_pcluster_configure.py
EddyMM: This will be re-worded. For now I'll update it to use a regex that matches both the current and updated wording.

tests/integration-tests/conftest.py
EddyMM: To make the stack available for other tests.
In this case the `pcluster configure` test needs to use a placement group that already exists.

tests/integration-tests/configs/common/common.yaml
EddyMM: All the other pcluster configure tests use instance types that don't support EFA so I figured we could save some time and resources by skipping this. I could add the test in order to be thorough though.

tests/integration-tests/tests/configure/test_pcluster_configure.py
EddyMM: This is from a test that was already part of the suite. I only refactored it.
The test is meant to confirm that the default instance-type is retrieved dynamically. E.g. For regions that don't have `t2.micro`, the default instance-type is `t3.micro`.

tests/integration-tests/tests/configure/test_pcluster_configure.py
EddyMM: The response to the EFA prompt. Using something like `efa_enabled` would be confusing since one would expect a boolean. In this case however we're anticipating a "y" or "n". Hence the choice to use `efa_response`

gmarciani: Got it: you mean the end.user response to the configure prompt

tests/integration-tests/configs/common/common.yaml
gmarciani: Having other tests use non-EFA instances is enough as long as they verify that the user is not prompted with EFA related questions.

tests/integration-tests/tests/configure/test_pcluster_configure.py
gmarciani: Got it thanks.

gmarciani (approved):

PR #4156, opened 2022-07-01
Add integration tests for scontrol reboot functionality

jdeamicis:
### Description of changes
* Add integration tests for scontrol reboot functionality

### Tests
* Integration tests for scontrol reboot via test_runner

### Checklist

### DIFF
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -465,6 +465,12 @@ schedulers:
         instances: {{ common.INSTANCES_DEFAULT_X86 }}
         oss: ["alinux2"]
         schedulers: ["slurm"]
+  test_slurm.py::test_scontrol_reboot:
+    dimensions:
+      - regions: ["us-east-1"]
+        instances: {{ common.INSTANCES_DEFAULT_X86 }}
+        oss: ["ubuntu2004"]
+        schedulers: ["slurm"]
 spot:
   test_spot.py::test_spot_default:
     dimensions:
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -467,7 +467,7 @@ schedulers:
         schedulers: ["slurm"]
   test_slurm.py::test_scontrol_reboot:
     dimensions:
-      - regions: ["us-east-1"]
+      - regions: ["us-west-2"]
         instances: {{ common.INSTANCES_DEFAULT_X86 }}
         oss: ["ubuntu2004"]
         schedulers: ["slurm"]
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -377,6 +377,52 @@ def test_slurm_memory_based_scheduling(
     )
 
 
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+@pytest.mark.slurm_scontrol_reboot
+def test_scontrol_reboot(
+        pcluster_config_reader,
+        clusters_factory,
+        test_datadir,
+        scheduler_commands_factory,
+):
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = scheduler_commands_factory(remote_command_executor)
+
+    # Test that an idle static node can be rebooted
+    _test_scontrol_reboot_idle_static_node(
+        remote_command_executor,
+        slurm_commands,
+        "queue1-st-t2medium-1",
+        asap=False,
+    )
+
+    # Test that an idle static node can be rebooted ASAP
+    _test_scontrol_reboot_idle_static_node(
+        remote_command_executor,
+        slurm_commands,
+        "queue1-st-t2medium-1",
+        asap=True,
+    )
+
+    # Test that an allocated static node can be rebooted
+    _test_scontrol_reboot_alloc_static_node(
+        remote_command_executor,
+        slurm_commands,
+        "queue1-st-t2medium-1",
+        asap=False,
+    )
+
+    # Test that an allocated static node can be rebooted ASAP
+    _test_scontrol_reboot_alloc_static_node(
+        remote_command_executor,
+        slurm_commands,
+        "queue1-st-t2medium-1",
+        asap=True,
+    )
+
+
 def _assert_cluster_initial_conditions(
     scheduler_commands, expected_num_dummy, expected_num_instance_node, expected_num_static
 ):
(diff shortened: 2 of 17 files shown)


tests/integration-tests/configs/common/common.yaml
```diff
@@ -465,6 +465,12 @@ schedulers:
         instances: {{ common.INSTANCES_DEFAULT_X86 }}
         oss: ["alinux2"]
         schedulers: ["slurm"]
+  test_slurm.py::test_scontrol_reboot:
+    dimensions:
+      - regions: ["us-east-1"]
```
gmarciani: Is it necessary to execute this test in us-east-1? If possible it would be better to distribute tests also in other regions.

tests/integration-tests/configs/common/common.yaml
```diff
@@ -465,6 +465,12 @@ schedulers:
...
+        instances: {{ common.INSTANCES_DEFAULT_X86 }}
+        oss: ["ubuntu2004"]
```
gmarciani, line 472: Is there any specific reason why you choose ubuntu2004?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -377,6 +377,70 @@ def test_slurm_memory_based_scheduling(
...
+    test_datadir,
+    scheduler_commands_factory,
+):
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = scheduler_commands_factory(remote_command_executor)
+
+    # Clear clustermgtd logs before starting the tests
+    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/clustermgtd")
```
gmarciani, line 394: why is this necessary?

tests/integration-tests/tests/schedulers/test_slurm/test_scontrol_reboot/pcluster.config.yaml
```diff
@@ -0,0 +1,24 @@
...
+Scheduling:
+  Scheduler: {{ scheduler }}
+  {{ scheduler_prefix }}Queues:
+    - Name: queue1
+      Networking:
+        SubnetIds:
+          - {{ private_subnet_id }}
+      ComputeResources:
+        - Name: t2medium
+          InstanceType: t2.medium
```
gmarciani: Why not using a t2.micro (4x cheaper) or a t2.small (2x cheaper) for this type of test?

tests/integration-tests/tests/schedulers/test_slurm/test_scontrol_reboot/pcluster.config.yaml
```diff
@@ -0,0 +1,24 @@
...
+          MinCount: 2
+          MaxCount: 4
+SharedStorage:
```
gmarciani: Why do we need a shared storage for this type of test?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1557,3 +1621,156 @@ def _test_memory_based_scheduling_enabled_true(
...
+
+    # Reboot nodes according to this logic:
+    # - 1 static idle node
+    # - 1 static idle node asap
+    # - 1 dynamic idle node
+    # - 1 dynamic idle node asap
+    static_nodes, dynamic_nodes = get_partition_nodes(nodes_in_queue)
+    assert_that(len(static_nodes)).is_equal_to(2)
+    assert_that(len(dynamic_nodes)).is_equal_to(2)
+    slurm_commands.reboot_compute_node(static_nodes[0], False)
```
gmarciani: In order to improve readability, may you please specify named parameter `asap=[True/False]`?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1557,3 +1621,156 @@ def _test_memory_based_scheduling_enabled_true(
...
+    slurm_commands.reboot_compute_node(static_nodes[1], True)
+    slurm_commands.reboot_compute_node(dynamic_nodes[0], False)
+    slurm_commands.reboot_compute_node(dynamic_nodes[1], True)
+
+    # Check that nodes enter a reboot state
+    time.sleep(2)
```
gmarciani: 1. Is this wait time fair enough or we risk to check the state too early?
2. Isn't the wait time to check transition already implicit in `wait_for_compute_nodes_states()`
1. I see you always wait 2 seconds before checking state transitions. What about having such wait time defined in a variable so that we have a single place to control the wait time?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1557,3 +1621,156 @@ def _test_memory_based_scheduling_enabled_true(
...
+    assert_compute_node_states(slurm_commands, nodes_in_queue, "reboot^")
+
+    # Wait that nodes come back after a while, without having triggered clustermgtd
```
gmarciani: It's not clear to me the relation between this comment and the actual check in line 1658. 
Why do we consider the absence of the log line `Found the following unhealthy static nodes` as a signal that clustermgtd has not been triggered?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1557,3 +1621,156 @@ def _test_memory_based_scheduling_enabled_true(
...
+    wait_for_compute_nodes_states(slurm_commands, nodes_in_queue, expected_states=["idle"])
+    assert_no_msg_in_logs(
+        remote_command_executor,
+        ["/var/log/parallelcluster/clustermgtd"],
+        ["Found the following unhealthy static nodes"],
+    )
+
+
+def _test_scontrol_reboot_alloc_nodes(
```
gmarciani: The logic in this function seems very similar to the logic in `_test_scontrol_reboot_idle_nodes()`.
Do you think it could make sense to generalize i a single function?

tests/integration-tests/configs/common/common.yaml
jdeamicis: No. I just wanted to put something different from `alinux2`. Or we can stick to `alinux2` if we prefer.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: Because as of now static nodes are always marked as unhealthy at the beginning of the life of the cluster, during the first iteration of clustermgtd. So we have to remove those messages if we want the test to be meaningful.

Alternatively we could build a function that checks the logs only from a certain point in time - which is out of scope for the time being, and also probably quite complicated to generalize for all possible log files we might want to analyze.

tests/integration-tests/tests/schedulers/test_slurm/test_scontrol_reboot/pcluster.config.yaml
jdeamicis: Addressing in the next commit.

tests/integration-tests/configs/common/common.yaml
jdeamicis: Discussed separately: we are going to set `us-west-2`. Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: Agreed. Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm/test_scontrol_reboot/pcluster.config.yaml
jdeamicis: We really don't. I copied a fairly basic configuration from another slurm test and left it there. Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: 1. No, for such clusters 2 seconds should always be enough. The transition should be instantaneous, but I'm putting some buffer for safety.
2. We are doing an `assert_compute_node_states` before the `wait_for_compute_nodes_states`.
3. Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: Because if it had been triggered (i.e. if the node had been marked as unhealthy), you would see it in the logs with `Found the following unhealthy static nodes (x1) : ['nodename'(..."`, which comes from [here](https://github.com/aws/aws-parallelcluster-node/blob/9b9e1145fe212fb7b079821453e969b3adb5ae23/src/slurm_plugin/clustermgtd.py#L734).

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: I was afraid generalizing it would have made the test workflow less readable, but indeed it came out fine. Addressed in the next commit.

tests/integration-tests/configs/common/common.yaml
gmarciani: That's fine in both cases, just wanted to understand the reasoning behind the choice :-)

tests/integration-tests/tests/schedulers/test_slurm.py
gmarciani: Agree, let's stick with the current implementation.

tests/integration-tests/tests/schedulers/test_slurm/test_scontrol_reboot/pcluster.config.yaml
gmarciani: Ok, so let's remove it if it's not necessary

gmarciani (approved):

PR #4158, opened 2022-07-01
Make SharedStorageMountDirValidator accept /scratch and improve error message

hanwen-pcluste:
### Description of changes
This is an amendament of https://github.com/aws/aws-parallelcluster/pull/3981.
See commit descriptions for details

### Tests
The unit test has been improved to test the logic. 
`test_ebs_multiple` is improved to test `/scratch` used by a shared EBS on instances without instance store

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -205,6 +205,10 @@ class InstanceTypeInfo:
 
         return vcpus
 
+    def instance_storage_supported(self) -> bool:
+        """Indicate whether instance storage is supported."""
+        return self.instance_type_data.get("InstanceStorageSupported")
+
     def supported_architecture(self):
         """Return the list of supported architectures."""
         supported_architectures = self.instance_type_data.get("ProcessorInfo").get("SupportedArchitectures")
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1266,13 +1266,13 @@ class BaseClusterConfig(Resource):
     def _validate_mount_dirs(self):
         self._register_validator(
             DuplicateMountDirValidator,
-            shared_mount_dir_list=self.shared_mount_dir_list,
-            local_mount_dir_list=list(self.local_mount_dir_set),
+            shared_storage_name_mount_dir_tuple_list=self.shared_storage_name_mount_dir_tuple_list,
+            local_mount_dir_instance_types_dict=self.local_mount_dir_instance_types_dict,
         )
         self._register_validator(
             OverlappingMountDirValidator,
-            shared_mount_dir_list=self.shared_mount_dir_list,
-            local_mount_dir_list=list(self.local_mount_dir_set),
+            shared_mount_dir_list=[mount_dir for mount_dir, _ in self.shared_storage_name_mount_dir_tuple_list],
+            local_mount_dir_list=list(self.local_mount_dir_instance_types_dict.keys()),
         )
 
     def _validate_max_storage_count(self, ebs_count, existing_storage_count, new_storage_count):
@@ -1321,37 +1321,37 @@ class BaseClusterConfig(Resource):
         return get_partition()
 
     @property
-    def shared_mount_dir_list(self):
-        """Retrieve the list of shared mount dirs."""
+    def shared_storage_name_mount_dir_tuple_list(self):
+        """Retrieve the list of shared storage names and mount dirs."""
         mount_dir_list = []
         if self.shared_storage:
             for storage in self.shared_storage:
-                mount_dir_list.append(storage.mount_dir)
+                mount_dir_list.append((storage.name, storage.mount_dir))
         return mount_dir_list
 
     @property
-    def local_mount_dir_set(self):
-        """Retrieve the list of local mount dirs of head compute nodes ephemeral volume."""
-        mount_dir_set = (
-            {
+    def local_mount_dir_instance_types_dict(self):
+        """Retrieve a dictionary of local mount dirs and corresponding instance types."""
+        mount_dir_instance_types_dict = defaultdict(set)
+        if self.head_node.instance_type_info.instance_storage_supported():
+            mount_dir_instance_types_dict[
                 self.head_node.local_storage.ephemeral_volume.mount_dir
                 if self.head_node.local_storage.ephemeral_volume
                 else DEFAULT_EPHEMERAL_DIR
-            }
-            if self.head_node.instance_type_info.instance_storage_supported()
-            else set()
-        )
+            ].add(self.head_node.instance_type)
+
         scheduling = self.scheduling
         if isinstance(scheduling, (SchedulerPluginScheduling, SlurmScheduling)):
             for queue in scheduling.queues:
-                if queue.has_instance_storage():
-                    mount_dir_set.add(
+                instance_types_with_instance_storage = queue.instance_types_with_instance_storage
+                if instance_types_with_instance_storage:
+                    mount_dir_instance_types_dict[
                         queue.compute_settings.local_storage.ephemeral_volume.mount_dir
                         if queue.compute_settings.local_storage.ephemeral_volume
                         else DEFAULT_EPHEMERAL_DIR
-                    )
+                    ].update(instance_types_with_instance_storage)
 
-        return mount_dir_set
+        return mount_dir_instance_types_dict
 
     @property
     def existing_storage_id_list(self):
@@ -1794,12 +1794,14 @@ class SlurmQueue(_CommonQueue):
         """Return the list of instance types associated to the Queue."""
         return [compute_resource.instance_type for compute_resource in self.compute_resources]
 
-    def has_instance_storage(self):
-        """Indicate if any of the compute resource has instance storage."""
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
         for compute_resource in self.compute_resources:
             if compute_resource.instance_type_info.instance_storage_supported():
-                return True
-        return False
+                result.add(compute_resource.instance_type)
+        return result
 
 
 class Dns(Resource):
@@ -1918,12 +1920,14 @@ class SchedulerPluginQueue(_CommonQueue):
         """Return the list of instance types associated to the Queue."""
         return [compute_resource.instance_type for compute_resource in self.compute_resources]
 
-    def has_instance_storage(self):
-        """Indicate if any of the compute resource has instance storage."""
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
         for compute_resource in self.compute_resources:
             if compute_resource.instance_type_info.instance_storage_supported():
-                return True
-        return False
+                result.add(compute_resource.instance_type)
+        return result
 
 
 class SchedulerPluginSupportedDistros(Resource):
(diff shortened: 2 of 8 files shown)


cli/src/pcluster/aws/aws_resources.py
```diff
@@ -205,6 +205,10 @@ def vcpus_count(self) -> int:
 
         return vcpus
 
+    def instance_storage_supported(self) -> bool:
+        """Indicate whether instance storage is supported."""
+        return self.instance_type_data.get("InstanceStorageSupported")
```
gmarciani, line 210: Is the field `InstanceStorageSupported ` always present in `instance_type_data`?
If we are not sure about it, we should consider to return false by default.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1331,19 +1332,24 @@ def shared_mount_dir_list(self):
...
-            if self.head_node.local_storage.ephemeral_volume
-            else DEFAULT_EPHEMERAL_DIR
-        }
+        mount_dir_set = (
+            {
+                self.head_node.local_storage.ephemeral_volume.mount_dir
+                if self.head_node.local_storage.ephemeral_volume
+                else DEFAULT_EPHEMERAL_DIR
+            }
+            if self.head_node.instance_type_info.instance_storage_supported()
```
gmarciani: Correct me if I'm wrong.
This change makes me think that we are fixing a bug that we had regardless of the main goal of this PR.
In particular, we were assuming to always have a local mount dir set even if the instance was not supporting instance storage. Is this right?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1905,6 +1918,13 @@ def instance_type_list(self):
         """Return the list of instance types associated to the Queue."""
         return [compute_resource.instance_type for compute_resource in self.compute_resources]
 
+    def has_instance_storage(self):
```
gmarciani: We are duplicating the same `has_instance_storage` function in both `SchedulerPluginQueue` and `SlurmQueue`. So, what about adding the function in `_CommonQueue`, instead?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -589,6 +588,13 @@ def _validate(self, mount_dir_list):
                 ),
                 FailureLevel.ERROR,
             )
+        for local_mount_dir in local_mount_dir_list:
+            if local_mount_dir in shared_mount_dir_list:
+                self._add_failure(
+                    f"Ephemeral drive mount directory `{local_mount_dir}` is overwritten by shared storage "
+                    f"mount directory. Please consider changing either directory to avoid the overwrite.",
+                    FailureLevel.WARNING,
```
gmarciani: **Product decision alert**
I think that this kind of failure should be an error rather than a warning. Overwriting a local mount point should block the cluster creation in the same way a shared mount point overwrites another shared mount point.

Pending product decision here.

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -224,14 +224,14 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
         any_order=True,
     )
     key_pair_validator.assert_has_calls([call(key_name="ec2-key-name")])
-    instance_type_validator.assert_has_calls([call(instance_type="t2.micro")])
+    instance_type_validator.assert_has_calls([call(instance_type="c5d.xlarge")])
```
gmarciani, line 227: It's not clear to me why you are removing the t2.micro case.
I was expecting to add tests/assertions not to replace existing ones?
May you please clarify?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -827,32 +827,43 @@ def test_fsx_architecture_os_validator(architecture, os, expected_message):
...
+            [],
             "Mount directory dir1 cannot be specified for multiple file systems",
         ),
         (
             ["dir1", "dir2", "dir3", "dir2", "dir1"],
+            [],
             "Mount directories dir2, dir1 cannot be specified for multiple file systems",
         ),
+        (
+            ["dir1", "dir2", "/scratch"],
```
gmarciani: Since `/scratch` is a notable mount point, I think it's worth it to have a test case verifying that when shared_mount_dir_list contains `/shared` and local mount dir does not, the validator succeeds

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: It comes from the doc of boto3 https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_instance_types. So I believe this is strong enough to assume the field always exist

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: You can call it a bug.

cli/tests/pcluster/validators/test_all_validators.py
hanwen-pcluste: This is to use a instance type that has an instance store

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: I included the test case in integration test.

gmarciani: I think we should add it in unit tests because:
1. validator logic is expected to be covered by unit tests 
2. this is a notable test case for the validator
3. if something can be tested both with unit tests and integ test, better to have in unit test at forst because they are faster and cheaper

cli/tests/pcluster/validators/test_all_validators.py
gmarciani: Got it: so we are replacing t2.micro with c5d.xlarge because the paths triggered by the latter is a superset of the paths triggered by the former

cli/src/pcluster/config/cluster_config.py
gmarciani: This duplication is actually a signal of a defect in the aforementioned class design.
By the way, this would required a refactoring that is out of scope for this PR.
So, let's proceed with the proposed implementation, but let's be sure to track the need for improvement.

cli/src/pcluster/validators/cluster_validators.py
gmarciani: Product decision is to be more conservative and keep the warning level.

cli/tests/pcluster/validators/test_cluster_validators.py
hanwen-pcluste: Done

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1320,32 +1321,37 @@ def partition(self):
...
     @property
-    def local_mount_dir_set(self):
-        """Retrieve the list of local mount dirs of head compute nodes ephemeral volume."""
-        mount_dir_set = {
-            self.head_node.local_storage.ephemeral_volume.mount_dir
-            if self.head_node.local_storage.ephemeral_volume
-            else DEFAULT_EPHEMERAL_DIR
-        }
+    def local_mount_dir_instance_types_dict(self):
+        """Retrieve a dictionary of local mount dirs and corresponding instance types."""
```
gmarciani, line 1334: Why do we need a mapping between local mount dir and instance types?
Isn't it enough to have a list of local mount dir?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1320,32 +1321,37 @@ def partition(self):
...
+        mount_dir_instance_types_dict = defaultdict(set)
+        if self.head_node.instance_type_info.instance_storage_supported():
+            mount_dir_instance_types_dict[
+                self.head_node.local_storage.ephemeral_volume.mount_dir
+                if self.head_node.local_storage.ephemeral_volume
+                else DEFAULT_EPHEMERAL_DIR
```
gmarciani, line 1340: Minor: this logic is implemented the same both for head node and compute nodes. We should consider to put it in common and reduce code duplication.
May you please keep track of this refactoring in our backlog?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -581,16 +581,28 @@ class DuplicateMountDirValidator(Validator):
...
+                    "Shared storage mount directories should be unique. "
+                    "Please change the mount directory configuration of the shared storage.",
+                    FailureLevel.ERROR,
+                )
+        for local_mount_dir, instance_types in local_mount_dir_instance_types_dict.items():
+            shared_storage_names = mount_dir_to_names.get(local_mount_dir)
+            if shared_storage_names:
+                self._add_failure(
+                    f"The mount directory `{local_mount_dir}` used for shared storage {shared_storage_names} "
+                    f"clashes with the one used for ephemeral store volumes of the instances {list(instance_types)}. "
```
gmarciani: We refer to "ephemeral store volume", but AWS docs typically refer to this storage as "instance store volume".
Furthermore, we refer to that one time as "ephemeral store volume" and another time as "ephemeral volume", let's use unique naming.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -581,16 +581,28 @@ class DuplicateMountDirValidator(Validator):
...
-        duplicated_mount_dirs = _find_duplicate_params(mount_dir_list)
-        if duplicated_mount_dirs:
-            self._add_failure(
-                "Mount {0} {1} cannot be specified for multiple file systems".format(
-                    "directories" if len(duplicated_mount_dirs) > 1 else "directory",
-                    ", ".join(mount_dir for mount_dir in duplicated_mount_dirs),
-                ),
-                FailureLevel.ERROR,
-            )
+    def _validate(self, shared_storage_name_mount_dir_tuple_list, local_mount_dir_instance_types_dict):
```
gmarciani, line 584: In this validation we do a two steps verification:
1. first verify that there is no clash among shared storages
2. then, verify that there is no clash between shared storages and local storages.
Did you do it this way because the goal was to provide distinct messages for shared v/s shared and shared v/s local?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1905,6 +1920,15 @@ def instance_type_list(self):
         """Return the list of instance types associated to the Queue."""
         return [compute_resource.instance_type for compute_resource in self.compute_resources]
 
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
+        for compute_resource in self.compute_resources:
+            if compute_resource.instance_type_info.instance_storage_supported():
```
gmarciani, line 1928: Does this introduce the risk of having a lot of calls to EC2 to retrieve such information?
In case it does, we should cache the info about instance storage support.

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: Because we want the error message to include instance type information

cli/src/pcluster/config/cluster_config.py
hanwen-pcluste: There is cache for instance_type_info. So no duplicate EC2 is required

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: Yes, because of the different messages

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: Done. Thanks

cli/src/pcluster/config/cluster_config.py
gmarciani: Got it, that's fine

cli/src/pcluster/config/cluster_config.py
gmarciani: Great, thanks

gmarciani (approved):

PR #4168, opened 2022-07-04
Add functional integration tests for memory-based scheduling

jdeamicis:
### Description of changes
* Add functional integration tests for memory-based scheduling

### Tests
* Integration tests for usage of Slurm memory flags with memory-based scheduling disabled vs enabled.

### Checklist

### DIFF
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -404,9 +404,13 @@ class SlurmCommands(SchedulerCommands):
         result = self._remote_command_executor.run_remote_command(check_core_cmd)
         return re.search(r"(\d+)", result.stdout).group(1)
 
-    def get_job_info(self, job_id):
+    def get_job_info(self, job_id, field=None):
         """Return job details from slurm"""
-        return self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id)).stdout
+        result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id)).stdout
+        if field is not None:
+            match = re.search(rf"({field})=(\S*)", result)
+            return match.group(2)
+        return result
 
     def cancel_job(self, job_id):
         """Cancel a job"""
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -405,7 +405,7 @@ class SlurmCommands(SchedulerCommands):
         return re.search(r"(\d+)", result.stdout).group(1)
 
     def get_job_info(self, job_id, field=None):
-        """Return job details from slurm"""
+        """Return job details from slurm. If field is provided, only the field is returned"""
         result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id)).stdout
         if field is not None:
             match = re.search(rf"({field})=(\S*)", result)
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -1657,7 +1657,7 @@ def _test_memory_based_scheduling_enabled_false(
 
     # Compile C program to test memory allocations
     remote_command_executor.run_remote_command("gcc memory_allocation_chars.c")
-    assert_that(remote_command_executor.run_remote_command("ls ./a.out").stdout).contains("a.out")
+    remote_command_executor.run_remote_command("ls ./a.out")
 
     # Check that I can use the `--mem` flag to filter compute nodes
     # Try to allocate on nodes with not enough memory
@@ -1712,6 +1712,7 @@ def _test_memory_based_scheduling_enabled_false(
     slurm_commands.wait_job_completed(job_id_2)
 
     # Check that without memory constraining, jobs might contend memory on the compute node
+    # (memory constraining makes sense only if memory is set as consumable resource)
     job_id_1 = slurm_commands.submit_command_and_assert_job_accepted(
         submit_command_args={
             "nodes": 1,
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -359,8 +359,6 @@ def test_slurm_memory_based_scheduling(
 
     # test Slurm without memory-based scheduling feature
     _test_memory_based_scheduling_enabled_false(
-        pcluster_config_reader,
-        cluster,
         remote_command_executor,
         slurm_commands,
         test_datadir,
@@ -371,8 +369,6 @@ def test_slurm_memory_based_scheduling(
 
     # test Slurm with memory-based scheduling feature
     _test_memory_based_scheduling_enabled_true(
-        pcluster_config_reader,
-        cluster,
         remote_command_executor,
         slurm_commands,
         test_datadir,
@@ -1589,8 +1585,6 @@ def _test_update_with_queue_params(pcluster_config_reader, cluster, remote_comma
 
 
 def _test_memory_based_scheduling_enabled_false(
-    pcluster_config_reader,
-    cluster,
     remote_command_executor,
     slurm_commands,
     test_datadir,
@@ -1673,7 +1667,7 @@ def _test_memory_based_scheduling_enabled_false(
     slurm_commands.wait_job_completed(job_id_1)
     slurm_commands.wait_job_completed(job_id_2)
 
-    # Check that without memory-constraining, jobs might contend memory on the compute node
+    # Check that without memory constraining, jobs might contend memory on the compute node
     job_id_1 = slurm_commands.submit_command_and_assert_job_accepted(
         submit_command_args={
             "nodes": 1,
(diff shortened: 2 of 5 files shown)


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -404,9 +404,13 @@ def get_node_cores(self, partition=None):
         result = self._remote_command_executor.run_remote_command(check_core_cmd)
         return re.search(r"(\d+)", result.stdout).group(1)
 
-    def get_job_info(self, job_id):
+    def get_job_info(self, job_id, field=None):
         """Return job details from slurm"""
```
gmarciani: Minor: include in the function comment that if field is specified, only the specified filed will be returned.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
...
+    # Upload files for memory allocation tests
+    remote_command_executor._copy_additional_files(
+        [
+            str(test_datadir / "memory_allocation_chars.c"),
+        ],
+    )
+
+    # Compile C program to test memory allocations
+    remote_command_executor.run_remote_command("gcc memory_allocation_chars.c")
+    assert_that(remote_command_executor.run_remote_command("ls ./a.out").stdout).contains("a.out")
```
gmarciani: If `a.out` is not there, the `ls` command would fail, so the assertion on `contains` is redundant.
You could simply run the `ls` command to verify that it is there

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
...
+
+    # Check that I can use the `--mem` flag to filter compute nodes
+    # Try to allocate on nodes with not enough memory
+    result = slurm_commands.submit_command(
+        nodes=1,
+        command="sleep 1",
+        constraint="ondemand1-i1",
+        other_options="--mem=4000 --test-only",
```
gmarciani, line 1668: Why should we include `--test-only`?
Is it like a dryrun?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
...
+    # Here two jobs submitted with `--mem=2000` can fit on a node with less than 4000 MiB memory
+    # because without memory as consumable resource, Slurm doesn't track the memory usage of
+    # each job.
+    assert_that(slurm_commands.get_job_info(job_id_1, field="NodeList")).is_equal_to(
+        slurm_commands.get_job_info(job_id_1, field="NodeList")
+    )
+    slurm_commands.wait_job_completed(job_id_1)
+    slurm_commands.wait_job_completed(job_id_2)
+
+    # Check that without memory constraining, jobs might contend memory on the compute node
```
gmarciani, line 1714: the command below has `--mem`, so why are you saying "without memory constraining"?
Correct me if I'm wrong, but I'd rather say "without memory as a consumable resource"

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
...
+            "nodes": 1,
+            "slots": 1,
+            "command": "srun ./a.out 2000000000",
+            "other_options": "--mem=2500 -w queue1-st-ondemand1-i1-1",
+            "raise_on_error": False,
+        }
+    )
+    slurm_commands.wait_job_completed(job_id_1)
+    slurm_commands.wait_job_completed(job_id_2)
+    # In this scenario the second job will have stolen memory from the first job, causing
```
gmarciani, line 1737: Why do we expect the second job to steal memory instead of having the first job acquiring it and the second one failing?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
...
+    job_id_1 = slurm_commands.submit_command_and_assert_job_accepted(
+        submit_command_args={
+            "nodes": 1,
+            "slots": 1,
+            "command": "sleep 30",
+            "other_options": "--mem=2000 -w queue1-st-ondemand1-i1-1",
+            "raise_on_error": False,
+        }
+    )
+    time.sleep(jiff)
```
gmarciani, line 1784: Here and in other part of the code: I don't get the jiff between job submissions. I see the goal of jiff when making assertion on nodes states, but not between job submissions.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1587,13 +1585,14 @@ def _test_update_with_queue_params(pcluster_config_reader, cluster, remote_comma
 
 
 def _test_memory_based_scheduling_enabled_false(
```
gmarciani, line 1631: I know it may be out of scope for this specific PR, but what about putting the tests related to memory based scheduling in a dedicated script instead of enlarging `test_slurm.py`

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1605,24 +1604,182 @@ def _test_memory_based_scheduling_enabled_false(
     assert_that(slurm_commands.get_node_attribute("queue1-st-ondemand1-i1-1", "Memory")).is_equal_to("3891")
     assert_that(slurm_commands.get_node_attribute("queue1-dy-ondemand1-i3-1", "Memory")).is_equal_to("31129")
 
-    # TODO: Add functional tests for memory-based scheduling
+    # Upload files for memory allocation tests
+    remote_command_executor._copy_additional_files(
+        [
+            str(test_datadir / "memory_allocation_chars.c"),
```
gmarciani, line 1654: Cool that you have implemented a C program to allocate a given amount of memory.
Just out of my curiosity: why not simply have a bash command allocating an array?

tests/integration-tests/tests/common/schedulers_common.py
jdeamicis: Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: You are right. Addressing in the next commit.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: Yes. This way you can test the submission command without running the actual job. Here we are interested only in checking that the `--mem` flag works as intended, so it's faster to use `--test-only`.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: This is subtle. The thing that actually prevents memory contention is the memory constraining (i.e. `ConstrainRAMSpace=yes`). Though, memory constraining makes sense only if memory is set as a consumable resource. 

So what you say is right, but my comment is more precise. I will make it clearer in the comment.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: That's what I've always seen happening. I would leave it as it is, and if someday we see these particular assertions failing, we will learn something new.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: I want to make sure that the first job has been submitted and processed by the scheduler, before submitting a new job. It's just a precaution: we can remove it if you prefer.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: For the way integration tests for schedulers are implemented, this is the only way that makes sense.

We can consider splitting the schedulers integration tests in a separate PR.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: I guess I'm more familiar with memory allocation in C rather than bash :)

tests/integration-tests/tests/schedulers/test_slurm.py
gmarciani: Thanks for making it more specific

tests/integration-tests/tests/schedulers/test_slurm.py
gmarciani: I think we should remove them

tests/integration-tests/tests/schedulers/test_slurm.py
gmarciani: May you please track this activity in our backlog?

tests/integration-tests/tests/schedulers/test_slurm.py
gmarciani: Fair enough :-)

gmarciani (approved):

tests/integration-tests/tests/schedulers/test_slurm.py
demartinofra: alternatively we could use the `--dependency after` flag: https://slurm.schedmd.com/sbatch.html#OPT_dependency

PR #4185, opened 2022-07-11
[Testing] Add integration tests to verify EFA using fabtests

gmarciani:
### Description of changes
Add integration tests to verify EFA using [fabtests](https://github.com/ofiwg/libfabric/tree/main/fabtests), which is the official  integration testing suite for libfabric.

The minimum subset of valuable tests considered in this change has been suggested by EFA engineer.

### Tests
1. Tests launched locally succeeded with instances c5n.9xlarge (cannot use p4d.24xlarge)
2. Tests will be verified with p4d.24xlarge on dev pipeline

### DIFF
--- a/tests/integration-tests/tests/efa/test_fabric.py
+++ b/tests/integration-tests/tests/efa/test_fabric.py
@@ -1,0 +1,110 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import logging
+
+import boto3
+import xmltodict
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+
+from tests.common.assertions import assert_no_errors_in_logs
+from tests.common.utils import run_system_analyzer
+from tests.performance_tests.common import read_remote_file, wait_process_completion
+
+FABTESTS_BASIC_TESTS = ["rdm_tagged_bw", "rdm_tagged_pingpong"]
+
+FABTESTS_GDRCOPY_TESTS = ["runt"]
+
+
+def test_fabric(
+    os,
+    region,
+    scheduler,
+    instance,
+    pcluster_config_reader,
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+    request,
+    s3_bucket_factory,
+):
+    """
+    Test tha EFA is working according to Fabtests, that is the official libfabric test suite.
+    See https://github.com/ofiwg/libfabric/tree/main/fabtests
+    """
+
+    # Bootstrap script to use p4d targeted ODCR
+    bucket_name = ""
+    if instance == "p4d.24xlarge":
+        bucket_name = s3_bucket_factory()
+        bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+        bucket.upload_file(str(test_datadir / "run_instance_override.sh"), "run_instance_override.sh")
+
+    cluster_config = pcluster_config_reader(
+        bucket_name=bucket_name,
+    )
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+
+    run_system_analyzer(cluster, scheduler_commands_factory, request, partition="q1")
+
+    fabtests_report = _execute_fabtests(remote_command_executor, test_datadir, instance)
+
+    num_tests = int(fabtests_report.get("testsuites", {}).get("testsuite", {}).get("@tests", None))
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/efa/test_fabric.py
```diff
@@ -0,0 +1,105 @@
...
+    s3_bucket_factory,
+):
+    """
+    Test tha EFA is working according to Fabtests, that is the official libfabric test suite.
+    See https://github.com/ofiwg/libfabric/tree/main/fabtests
+    """
+
+    # Bootstrap script to use p4d targeted ODCR
+    bucket_name = ""
+    if instance == "p4d.24xlarge":
```
enrico-usai, line 47: in future we could have an auto-use fixture to check instance type and perform this upload when using p4d

tests/integration-tests/tests/efa/test_fabric.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+
+    run_system_analyzer(cluster, scheduler_commands_factory, request, partition="q1")
+
+    fabtests_report = _execute_fabtests(remote_command_executor, test_datadir, instance)
+
+    num_tests = int(fabtests_report["testsuites"]["testsuite"]["@tests"])
+    num_failures = int(fabtests_report["testsuites"]["testsuite"]["@failures"])
```
enrico-usai: minor: I'd use `get` with a default value, otherwise let's catch KeyError.

tests/integration-tests/tests/efa/test_fabric.py
```diff
@@ -0,0 +1,105 @@
...
+        bucket_name=bucket_name,
+    )
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+
+    run_system_analyzer(cluster, scheduler_commands_factory, request, partition="q1")
+
+    fabtests_report = _execute_fabtests(remote_command_executor, test_datadir, instance)
+
+    num_tests = int(fabtests_report["testsuites"]["testsuite"]["@tests"])
```
enrico-usai: minor: it would be useful to have an example of the structure of the report we're parsing. In this way if in future the output changes we know what to change here.

tests/integration-tests/tests/efa/test_fabric/test_fabric/run-fabtests.sh
```diff
@@ -0,0 +1,54 @@
...
+mkdir -p $(dirname $LOG_FILE)
+mkdir -p $(dirname $REPORT_FILE)
+
+python3 $FABTESTS_RUNNER \
+  -b efa \
+  -p "$FABTESTS_BIN_DIR" \
+  --expression "$TEST_EXPRESSION" \
+  --timeout $FABTESTS_TIMEOUT \
+  --junit-xml $REPORT_FILE \
+  $TEST_ENVIRONMENT_OPTIONS $COMPUTE_IP_1 $COMPUTE_IP_2 > $LOG_FILE 2>&1 &
```
enrico-usai: TEST_ENVIRONMENT_OPTIONS is not defined. See final "S"

enrico-usai: We have a ShellCheck GitHub action that should discover this issues, could you check why it's not finding them? maybe we are not parsing all the folders.

tests/integration-tests/tests/efa/test_fabric.py
gmarciani: You're right. Tracked in our backlog.

tests/integration-tests/tests/efa/test_fabric.py
gmarciani: I'll go with a catch because there is no meaningful default to use

gmarciani: Will got with get+assertion of the presence of the required fields. Thanks

tests/integration-tests/tests/efa/test_fabric.py
gmarciani: I generally agree with you but in this specific case the report schema is so big that it may not be worth it.

tests/integration-tests/tests/efa/test_fabric/test_fabric/run-fabtests.sh
gmarciani: Good catch! And I did not discover it with my manual tests because when not running on p4d the absence of that variable is not impactful. Fix in the next rev.

gmarciani: The file is included and considered for ShellCheck inspection.
Nevertheless our ShellCheck action did not discover it because:
1. ShellCheck does not consider an undefined variable as an error, but as a warning (verified locally that it would have been able to detect it as a warning)
2. We execute ShellCheck with the error level filter `-S error`

enrico-usai: ok I'd say we should change it to warning, but this is another task.

enrico-usai (approved):

tests/integration-tests/tests/efa/test_fabric/test_fabric/run-fabtests.sh
gmarciani: I agree, tracked in our backlog

PR #4191, opened 2022-07-12
Improve ExistingFsxNetworkingValidator Validator

hanwen-pcluste:
1. Only display one error message if multiple ports of a single file system are absent
2. Improve the validator to check UDP protocol when necessary. This was not implementated because we had no time
3. Refactor the code avoid function too complex code linters error

### Tests
Tests in config/pcluster3.yaml have been passed
Manually tested the validator works as expected when the ports are not completed

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1221,27 +1221,28 @@ class BaseClusterConfig(Resource):
                 resource_name="Shared Storage IDs",
             )
 
-            existing_fsx = []
+            existing_fsx = set()
             for storage in self.shared_storage:
                 self._register_validator(SharedStorageNameValidator, name=storage.name)
                 self._register_validator(SharedStorageMountDirValidator, mount_dir=storage.mount_dir)
-                if isinstance(storage, BaseSharedFsx):
+                if isinstance(storage, SharedFsxLustre):
                     if storage.file_system_id:
                         existing_storage_count["fsx"] += 1
-                        existing_fsx.append(storage.file_system_id)
+                        existing_fsx.add(storage.file_system_id)
                     else:
                         new_storage_count["fsx"] += 1
                     self._register_validator(
                         FsxArchitectureOsValidator, architecture=self.head_node.architecture, os=self.image.os
                     )
-                if isinstance(storage, (ExistingFsxOpenZfs, ExistingFsxOntap)):
-                    existing_fsx.append(storage.file_system_id)
-                if isinstance(storage, SharedEbs):
+                elif isinstance(storage, (ExistingFsxOpenZfs, ExistingFsxOntap)):
+                    existing_storage_count["fsx"] += 1
+                    existing_fsx.add(storage.file_system_id)
+                elif isinstance(storage, SharedEbs):
                     if storage.raid:
                         new_storage_count["raid"] += 1
                     else:
                         ebs_count += 1
-                if isinstance(storage, SharedEfs):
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -438,9 +440,16 @@ def _check_sg_rules_for_port(rule, port_to_check):
     # if ip_protocol is -1, all ports are allowed
     if ip_protocol == "-1":
         return True
-    # tcp == protocol 6,
-    # if the ip_protocol is tcp, from_port and to_port must >= 0 and <= 65535
-    if (ip_protocol in ["tcp", "6"]) and (from_port <= port_to_check <= to_port):
+    # Add protocol number in addition to the protocol name
+    if protocol == "tcp":
+        expected_protocol = [protocol, "6"]
```
gmarciani, line 445: Why do we need these magic numbers: 6 and 17? Who is in control of these numbers? Could these numbers be dynamically retrieved to avoid tech debts?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -505,23 +514,40 @@ def _check_file_systems(self, are_all_security_groups_customized, file_systems,
...
+        missing_ports = []
+        for port in ports:
+            fs_access = False
+            for network_interface in network_interfaces:
+                # Get list of security group IDs
+                sg_ids = [sg.get("GroupId") for sg in network_interface.get("Groups")]
+                if _check_in_out_access(
+                    sg_ids,
+                    port=port,
+                    is_cidr_optional=are_all_security_groups_customized,
```
gmarciani, line 542: Why `is_cidr_optional=are_all_security_groups_customized`? I don't get why the CIDR is optional when all SGs are customized.

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: According to the [boto3 doc](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_security_groups), Either the string or the number is included in the boto3 result.

To make it dynamically, we need to include a Python dependencies https://pypi.org/project/sockets/. I believe we don't want to include the dependencies just for these numbers?

gmarciani: For sure we don't want a new external deps. LEt's go with the proposed solution. thanks

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: If all security groups are customized, CIDR specification is not required. Because custom security groups can allow traffics by specifying security group id of themselves or others.
On the contrary, if cluster nodes use ParallelCluster managed security group, the security groups of the file system has to have CIDR specification to enable traffics from the file system to the nodes

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1254,7 +1255,7 @@ def _register_storage_validators(self):
                         new_storage_count["efs"] += 1
             self._register_validator(
                 ExistingFsxNetworkingValidator,
```
gmarciani, line 1257: Minor: Please keep track in our backlog that we must cover these kind of changes in unit tests like `test_validators_are_called_with_correct_argument` to be sure that expected validators are called with the correct arguments

gmarciani (approved):

PR #4193, opened 2022-07-13
[Validation] Removed validation of `DisableSimultaneousMultithreading` and adapted disabling logic accordingly.

gmarciani:
### Description of changes
1. Removed validation of `DisableSimultaneousMultithreading` in order to allow the disablement of multi-threading on all instances supporting it or to skip it on instances not supporting it.
2. Adapted internal logic to select disablement via CpuOptions or manually

Proved via scripting that single-threading can be set via `CpuOptions` for all instances having 1 within `VcpuInfo.ValidThreadsPerCore`, even if they have `VcpuInfo.DefaultThreadsPerCore=1` and regardless the supported architecture is `x86_64` or `arm64`.

1. DefaultThreadsPerCore > 1 AND ValidThreadsPerCore contains 1 => multi-threading must be disabled via CpuOptions
1. DefaultThreadsPerCore > 1 AND ValidThreadsPerCore not reported => multi-threading must be disabled manually
1. DefaultThreadsPerCore = 1 AND ValidThreadsPerCore contains 1 => nothing to do
1. DefaultThreadsPerCore = 1 AND ValidThreadsPerCore not reported => nothing to do

### Tests
1. Unit tests succeeded
3. Verified that the criteria to select the instances supporting the single threading via CpuOptions is correct by setting it for every instance types in DUB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ CHANGELOG
 **BUG FIXES**
 - Fix default for disable validate and test components when building custom AMI. The default was to disable those components, but it wasn't effective.
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
+- Dropped validation that would prevent ARM instance type to be used when `DisableSimultaneousMultithreading` was set to true.
 
 3.1.4
 ------
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -224,10 +224,9 @@ class InstanceTypeInfo:
         return self.instance_type_data.get("InstanceType")
 
     def is_cpu_options_supported_in_lt(self):
-        """Check whether hyperthreading can be disabled via CPU options."""
-        # If default threads per core is 1, HT doesn't need to be disabled
+        """Check whether CPUOptions can be set in launch template."""
         # When CpuOptions is not supported, valid threads per core can't be found in vcpu info
-        return self.default_threads_per_core() > 1 and 1 in self.valid_threads_per_core()
+        return len(self.valid_threads_per_core()) != 0
 
     def is_ebs_optimized(self):
         """Check whether the instance has optimized EBS support."""
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -68,7 +68,6 @@ from pcluster.validators.cluster_validators import (
     ComputeResourceSizeValidator,
     CustomAmiTagValidator,
     DcvValidator,
-    DisableSimultaneousMultithreadingArchitectureValidator,
     DuplicateMountDirValidator,
     DuplicateNameValidator,
     EfaOsArchitectureValidator,
@@ -975,11 +974,6 @@ class HeadNode(Resource):
 
     def _register_validators(self):
         self._register_validator(InstanceTypeValidator, instance_type=self.instance_type)
-        self._register_validator(
-            DisableSimultaneousMultithreadingArchitectureValidator,
-            disable_simultaneous_multithreading=self.disable_simultaneous_multithreading,
-            architecture=self.architecture,
-        )
 
     @property
     def architecture(self) -> str:
@@ -1000,7 +994,7 @@ class HeadNode(Resource):
     @property
     def pass_cpu_options_in_launch_template(self) -> bool:
         """Check whether CPU Options must be passed in launch template for head node."""
-        return self.disable_simultaneous_multithreading and self.instance_type_info.is_cpu_options_supported_in_lt()
+        return self.disable_simultaneous_multithreading_via_cpu_options
 
     @property
     def is_ebs_optimized(self) -> bool:
@@ -1022,12 +1016,20 @@ class HeadNode(Resource):
     @property
     def disable_simultaneous_multithreading_via_cpu_options(self) -> bool:
         """Return true if simultaneous multithreading must be disabled through cpu options."""
-        return self.disable_simultaneous_multithreading and self.instance_type_info.is_cpu_options_supported_in_lt()
+        return (
+            self.disable_simultaneous_multithreading
+            and self.instance_type_info.default_threads_per_core() > 1
+            and self.instance_type_info.is_cpu_options_supported_in_lt()
+        )
 
     @property
     def disable_simultaneous_multithreading_manually(self) -> bool:
         """Return true if simultaneous multithreading must be disabled with a cookbook script."""
-        return self.disable_simultaneous_multithreading and not self.instance_type_info.is_cpu_options_supported_in_lt()
+        return (
+            self.disable_simultaneous_multithreading
+            and self.instance_type_info.default_threads_per_core() > 1
+            and not self.instance_type_info.is_cpu_options_supported_in_lt()
+        )
 
     @property
     def instance_role(self):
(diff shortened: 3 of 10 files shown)


tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -54,7 +54,7 @@ def test_efa(
         multithreading_disabled = True
     else:
         head_node_instance = "c6g.16xlarge"
-        multithreading_disabled = False
+        multithreading_disabled = True
```
lukeseawalker: what about removing this variable if there is only one case? Than you can also change the config to remove the placeholder

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -57,7 +55,14 @@ def max_network_interface_count(self):
...
 
     def default_threads_per_core(self):
-        return self._default_threads_per_core
+        # There are more instance types, but for the simplicity of the mock,
+        # we consider only t2 as having one thread per core.
+        return 1 if self._instance_type.startswith("t2") else 2
+
+    def valid_threads_per_core(self):
+        # There are more instance types, but for the simplicity of the mock,
+        # we consider only t2 as having one thread per core.
```
lukeseawalker: "we consider only t2 as having one thread per core."
probably you meant to consider only t2 has not reporting valid thread per core

cli/tests/pcluster/validators/test_all_validators/test_validators_are_called_with_correct_argument/slurm.yaml
```diff
@@ -30,8 +30,10 @@ Scheduling:
         - Name: compute_resource1
           InstanceType: c5.4xlarge
           MaxCount: 5
+          DisableSimultaneousMultithreading: true
```
lukeseawalker, line 33: is it worth adding the case "instance not supporting multi-thread disablement" + DisableSimultaneousMultithreading = true?, i.e.
```
InstanceType: t2.large
DisableSimultaneousMultithreading: true
```

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ CHANGELOG
 **BUG FIXES**
 - Fix default for disable validate and test components when building custom AMI. The default was to disable those components, but it wasn't effective.
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
+- Fix validation for configuration parameter `Scheduling/SlurmQueues/ComputeResources/DisableSimultaneousMultithreading`
```
lukeseawalker: isn't the DisableSimultaneousMultithreading fixed also for the HeadNode?

tests/integration-tests/tests/efa/test_efa.py
gmarciani: Fixed in the next rev.

cli/tests/pcluster/aws/dummy_aws_api.py
gmarciani: Yes, copy-paste error. Rephrasing in the next rev.

cli/tests/pcluster/validators/test_all_validators/test_validators_are_called_with_correct_argument/slurm.yaml
gmarciani: Good point, thanks. Added in the next rev.

CHANGELOG.md
gmarciani: Good point, thanks! It is. I would put a general 

"Fix validation for configuration parameter `DisableSimultaneousMultithreading` to allow the disablement of multi-threading on all instances supporting it." then

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -224,10 +224,13 @@ def instance_type(self):
...
 
     def is_cpu_options_supported_in_lt(self):
-        """Check whether hyperthreading can be disabled via CPU options."""
-        # If default threads per core is 1, HT doesn't need to be disabled
+        """Check whether CPUOptions can be set in launch template."""
         # When CpuOptions is not supported, valid threads per core can't be found in vcpu info
-        return self.default_threads_per_core() > 1 and 1 in self.valid_threads_per_core()
+        return len(self.valid_threads_per_core()) != 0
+
+    def is_multi_threading_disablement_supported(self):
```
lukeseawalker: I think that the logic on the method "disable_simultaneous_multithreading_manually" should be revisited to account this change. Can you please check and in case add also a test to cover this case? I'm thinking about for example a metal instance with DefaultThreadsPerCore = 2 but that doesn't support cpu_options.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -224,10 +224,13 @@ def instance_type(self):
         return self.instance_type_data.get("InstanceType")
 
     def is_cpu_options_supported_in_lt(self):
```
lukeseawalker, line 226: since you have changed the method behavior, please also check all the callers of this method, to understand if they need to be changed as well.

gmarciani: Thanks, done in the next rev.

cli/src/pcluster/aws/aws_resources.py
gmarciani: We decided to remove the validator because we have the following cases:
1. DefaultThreadsPerCore > 1 AND ValidThreadsPerCore contains 1 => multi-threading must be disabled via CpuOptions
1. DefaultThreadsPerCore > 1 AND ValidThreadsPerCore not reported => multi-threading must be disabled manually
1. DefaultThreadsPerCore = 1 AND ValidThreadsPerCore contains 1 => nothing to do
1. DefaultThreadsPerCore = 1 AND ValidThreadsPerCore not reported => nothing to do

So there is no case where the validators should fail.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ CHANGELOG
 **BUG FIXES**
 - Fix default for disable validate and test components when building custom AMI. The default was to disable those components, but it wasn't effective.
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
+- Removed validation of `DisableSimultaneousMultithreading` in order to allow the disablement of multi-threading on all instances supporting it
+  or to skip it on instances not supporting it.
```
lukeseawalker: " skip it on instances not supporting it." we are disabling multithreading also in the case of instance not supporting it, through the manual disable flag.
I would rewrite the bullet with something "Dropped validation that would prevent ARM instance type to be used when DisableSimultaneousMultithreading was set to true."

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1022,12 +1016,20 @@ def instance_type_info(self) -> InstanceTypeInfo:
     @property
     def disable_simultaneous_multithreading_via_cpu_options(self) -> bool:
         """Return true if simultaneous multithreading must be disabled through cpu options."""
-        return self.disable_simultaneous_multithreading and self.instance_type_info.is_cpu_options_supported_in_lt()
+        return (
+            self.disable_simultaneous_multithreading
+            and self.instance_type_info.default_threads_per_core() > 1
```
lukeseawalker, line 1021: What about removing "and self.instance_type_info.default_threads_per_core() > 1"?
This will semplify the logic and will use cpu options even in the case default_threads_per_core = 1

CHANGELOG.md
gmarciani: Totally agree with you. Your proposal is a lot more customer-centric, thanks

cli/src/pcluster/config/cluster_config.py
gmarciani: If this is not a strong concern for you, I think we should keep this conditions as it is because it guarantees that we don't do any unnecessary setup action, whether it is manual or via template.

Furthermore, since sooner or later we are moving towards a scenario where we will perform this specific disablement only manually, it would be even better to skip manual steps when not required.

What do you think?

lukeseawalker: ACK

lukeseawalker (approved):

PR #4199, opened 2022-07-15
[FsxS3Access] Add missing permissions needed to import/export from S3 when using Fsx for Lustre via API

EddyMM:
### Description of changes
- Creation of a cluster with FsxLuster storage via ParallelCluster API fails due to the user role missing some permissions needed to import/export from S3.
- This commit creates an inline-policy with the needed permissions and associates it with the user role.
- The inline-policy is only created if the API stack is created with certain parameters (EnableFSxS3Access and FsxS3Buckets).

### Tests
* Manually tested by first reproducing the error through API Gateway
* Applied the changes in this commit to a copy of the current CFN Template
* Recreated the API and verified that the `FSxS3AccessPolicy` is created and associated with the `ParallelClusterUserRole`
* Made an API request using API Gateway to create a cluster
* Got a 202 OK response status with these results

```
{
  "cluster": {
    "cloudformationStackArn": "<STACK_ARN>",
    "clusterName": "<CLUSTER_NAME>",
    "clusterStatus": "CREATE_IN_PROGRESS",
    "region": "eu-west-1",
    "version": "3.1.4"
  }
}
```

### References
* [CloudFormation Conditions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-condition.html)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ CHANGELOG
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
 - Dropped validation that would prevent ARM instance type to be used when `DisableSimultaneousMultithreading` was set to true.
 - Add missing policies for EcrImageDeletionLambda and ImageBuilderInstance roles that were causing failure when upgrading ParallelCluster API from one version to another.
+- Add missing permissions needed to import/export from S3 when using FSx for Lustre via ParallelCluster API.
 
 3.1.4
 ------
--- a/api/infrastructure/parallelcluster-api.yaml
+++ b/api/infrastructure/parallelcluster-api.yaml
@@ -59,6 +59,27 @@ Parameters:
       - true
       - false
 
+  EnableFSxS3Access:
+    Description: |
+      When set to true the ParallelCluster API can access, write to the S3 buckets specified in the Filed FsxS3Bucket, it is needed to import/export from/to S3 when creating an FSx filesystem.
+      NOTE - setting this to true grants the Lambda function S3 Get*, List* and PutObject privileges on the buckets specified in FsxS3Buckets.
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
+
+  FsxS3Buckets:
+    Description: |
+      Comma separated list of S3 bucket ARNs, to allow the lambda function to import/export from/to S3 when creating an FSx filesystem.
+      NOTE - The setting is used only when EnableFSxS3Access is set to true. (example arn:aws:s3:::<S3_BUCKET_1>,arn:aws:s3:::<S3_BUCKET_2>)
+    Type: String
+    Default: ''
+    AllowedPattern: ^((arn:[a-z\-]*:s3:[a-z0-9\-]*:([0-9]{12})*:[^,\s\/\*]+)?(,arn:[a-z\-]*:s3:[a-z0-9\-]*:([0-9]{12})*:[^,\s\/\*]+)*)$|^\*$
+    ConstraintDescription: |
+      The list of S3 buckets is incorrectly formatted. The list should have the format: arn:<PARTITION>:s3:<REGION>:<ACCOUNT_ID>:<BUCKET_NAME>[,arn:<PARTITION>:s3:<REGION>:<ACCOUNT_ID>:<BUCKET_NAME>,...]
+      Example: arn:aws:s3:::test-bucket-1,arn:aws:s3:::test-bucket-2,arn:aws:s3:::test-bucket-3
+
   PermissionsBoundaryPolicy:
     Description: |
       ARN of a IAM policy to use as PermissionsBoundary for all IAM resources created by ParallelCluster API.
@@ -155,6 +176,10 @@ Conditions:
     Fn::And:
       - !Not [!Equals [!Ref ImageBuilderVpcId, ""]]
       - !Not [!Equals [!Ref ImageBuilderSubnetId, ""]]
+  EnableFSxS3AccessCondition: !And
+    - !Equals [!Ref EnableFSxS3Access, true]
+    - !Condition CreateIamResources
+  UseAllBucketsForFSxS3: !Equals [!Ref FsxS3Buckets, "*"]
 
 Resources:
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ CHANGELOG
 - Fix default for disable validate and test components when building custom AMI. The default was to disable those components, but it wasn't effective.
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
 - Dropped validation that would prevent ARM instance type to be used when `DisableSimultaneousMultithreading` was set to true.
+- Fix S3BucketUriValidator error that occurs when creating a cluster with FsxStorage using ParallelCluster API
```
gmarciani: What about "Add missing permissions needed to import/export from S3 when using Fsx for Lustre via ParallelCluster API."?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -59,6 +59,23 @@ Parameters:
       - true
       - false
 
+  EnableFSxS3Access:
+    Description: |
+      When set to true the ParallelCluster API can access, write to the S3 buckets specified in the Filed FsxS3Bucket, it is needed to import/export from/to S3 when creating an FSx filesystem.
+      WARNING - setting this to true grants to the Lambda function S3 Get*, List* and PutObject privileges on the buckets specified in FsxS3Buckets.
```
gmarciani: Let's replace "WARNING - setting ..." with "NOTE - Setting ..."

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -59,6 +59,23 @@ Parameters:
...
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
+
+  FsxS3Buckets:
+    Description: |
+      Comma separated arn of the S3 buckets, to allow the lambda function to import/export from/to S3 when creating an FSx filesystem.
```
gmarciani: Comma separated list of S3 bucket ARNs

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -59,6 +59,23 @@ Parameters:
...
+      WARNING - The setting is used only when EnableFSxS3Access is set to true. (example arn:aws:s3:::<S3_BUCKET_1>,arn:aws:s3:::<S3_BUCKET_2>)
```
gmarciani: 1. Let's replace WARNING with NOTE.
!. The example should rather be before the NOTE

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -59,6 +59,23 @@ Parameters:
...
+    Type: String
+    AllowedPattern: ^($|arn:.*:s3:[a-z0-9\-]*:([0-9]{12})*:.+)
```
gmarciani: Where did you get this regex?
Be careful that it matches strings such as "arn:aws:s3:::S3_BUCKET_1,   arn:aws:s3:::S3_BUCKET_2", (with blank spaces), that are not split as expected within the policy statement

tests/iam_policies/user-role.cfn.yaml
```diff
@@ -20,12 +20,31 @@ Parameters:
     AllowedValues:
       - true
       - false
+  EnableFSxS3Access:
```
gmarciani, line 23: Let's align this template to the changes in parallelcluster-api.yaml

api/infrastructure/parallelcluster-api.yaml
EddyMM: It's a custom regex. I'll test it against such variations of S3 arns.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ CHANGELOG
 - Fix default for disable validate and test components when building custom AMI. The default was to disable those components, but it wasn't effective.
 - Handle corner case in the scaling logic when instance is just launched and the describe instances API doesn't report yet all the EC2 info.
 - Dropped validation that would prevent ARM instance type to be used when `DisableSimultaneousMultithreading` was set to true.
+- Add missing permissions needed to import/export from S3 when using Fsx for Lustre via ParallelCluster API.
```
gmarciani: Minor: Fsx should be FSx.

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -459,6 +483,31 @@ Resources:
...
+              - iam:AttachRolePolicy
+              - iam:PutRolePolicy
+            Resource: !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:role/aws-service-role/s3.data-source.lustre.fsx.amazonaws.com/*
+            Effect: Allow
+            Sid: FSxS3PoliciesAttach
+          - Action:
+              - s3:Get*
+              - s3:List*
+              - s3:PutObject
+            Resource: !Split [",", !Ref FsxS3Buckets]
```
gmarciani: The `s3:PutObject` requires the resource to be BUCKET_NAME/*, doesn't it?

EddyMM: Yes, I'll run some additional tests to confirm. This can be specified in the FsxS3Buckets parameter but perhaps we may want to implicitly add the `*` wildcard.

EddyMM: Good catch, i've updated the parameter pattern to make sure the user provides a list with bucket ARNs that have `/*`

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -459,6 +483,41 @@ Resources:
...
+              - ","
+              - !Join # "abc,xyz,qrs,abc/*,xyz/*,qrs/*"
+                - ""
+                - - !Ref FsxS3Buckets
+                  - ","
+                  - !Join # "abc/*,xyz/*,qrs/*"
+                    - "/*,"
+                    - !Split # ["abc", "xyz" , "qrs/*]"
+                      - ","
+                      - !Join ["", [!Ref FsxS3Buckets, "/*"]] ### "abc,xyz,qrs/*"
```
demartinofra: This looks good but I'm wondering if we can leverage !Sub to make it more readable. For example something like:
```
!Sub
- "${FsxS3Buckets},${FsxS3BucketsObjects}"
- FsxS3BucketsObjects: !Join ["/*,", !Split [",", !Sub "${FsxS3Buckets}/*"]]
```

syntax to be checked :)

demartinofra (commented):
LGTM, only left a minor comment on a potential improvement

api/infrastructure/parallelcluster-api.yaml
EddyMM: I'll take a look into it. `!Sub` would definitely help simplify it.

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -459,6 +484,37 @@ Resources:
...
+              - s3:Get*
+              - s3:List*
+              - s3:PutObject
+            Resource: !Split
+              - ","
+              - !If [
+                UseAllBucketsForFSxS3,
+                "*",
+                !Sub ["${FsxS3Buckets},${FsxS3BucketsObjects}", FsxS3BucketsObjects: !Join ["/*,", !Split [",", !Sub "${FsxS3Buckets}/*"]]]
+                ]
```
demartinofra: nit: for the !If since you are not inlining you could use `-` instead of `[` which is more yaml friendly.

demartinofra (approved):

PR #4213, opened 2022-07-22
Add test for all validators for scheduler plugin

chenwany:
### Description of changes
* Add test for all validators for scheduler plugin

### Tests
* unit tests

### References

### Checklist


cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -295,3 +309,296 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
     ebs_volume_size_snapshot_validator.assert_called()
     shared_ebs_volume_id_validator.assert_called()
     fsx_persistent_options_validator.assert_called()
+
+
+def test_scheduler_plugin_all_validators_are_called(test_datadir, mocker):
+    """Verify that all validators are called during validation."""
```
jdeamicis, line 306: Would it get too ugly to generalize the already existing `test_all_validators_are_called()` and `test_validators_are_called_with_correct_argument()` adding a new `scheduler` parameters in the signature, which can then be controlled via a `pytest.mark.parametrize` annotation? This way it would be easier to see any difference in the validation between "normal scheduler" and scheduler plugin. Or is it really better to split the two sets of validators?

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -295,3 +309,296 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
...
+    # Assert validators are called
+    for m in mockers:
+        if m["name"] in [
+            "TagKeyValidator",
+            "ClusterNameValidator",
+            "InstanceProfileValidator",
+            "RoleValidator",
+            "MixedSecurityGroupOverwriteValidator",
+            "HostedZoneValidator",
+            "InstanceTypeMemoryInfoValidator",
```
jdeamicis, line 349: Why are we skipping these 3 additional validators with respect to `test_all_validators_are_called()`?

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
```diff
@@ -0,0 +1,263 @@
...
+        Logs:
+          Files:
+            - FilePath: /var/log/slurmctld.log
+              TimestampFormat: "%Y-%m-%dT%H:%M:%S.%f"
+              NodeType: ALL
+              LogStreamName: slurmctld.log
+            - FilePath: /var/log/slurmd.log
+              TimestampFormat: "%Y-%m-%dT%H:%M:%S.%f"
+              NodeType: HEAD
+              LogStreamName: slurmctld.log
```
jdeamicis, line 107: Minor: maybe we want to generalize the names of these logs, since the rest of your SchedulerPlugin parameters are generic (for example, the name `my_scheduler`).

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -24,12 +25,25 @@
...
+        kms_validators,
+        iam_validators,
+        networking_validators,
+        s3_validators,
+    ]
+    if additional_modules:
+        modules += additional_modules
     for module in [
         cluster_validators,
         ebs_validators,
```
jdeamicis: Shouldn't we be looping over `modules` now instead of using an explicit list here? Or am I missing something?

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_validators_are_called_with_correct_argument/scheduler_plugin.yaml
```diff
@@ -0,0 +1,180 @@
...
+        Logs:
+          Files:
+            - FilePath: /var/log/slurmctld.log
+              TimestampFormat: "%Y-%m-%dT%H:%M:%S.%f"
+              NodeType: ALL
+              LogStreamName: slurmctld.log
+            - FilePath: /var/log/slurmd.log
+              TimestampFormat: "%Y-%m-%dT%H:%M:%S.%f"
+              NodeType: HEAD
+              LogStreamName: slurmctld.log
```
jdeamicis, line 79: Same as above (generalize log names).

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
```diff
@@ -0,0 +1,263 @@
...
+  ClusterTemplate: https://tests/aws-parallelcluster-template-3.0.tgz
+  Cookbook:
+    ChefCookbook: https://tests/aws-parallelcluster-cookbook-3.0.tgz
+    ExtraChefAttributes: |
+      {"cluster": {"scheduler_slots": "cores"}}
+  AwsBatchCliPackage: s3://test/aws-parallelcluster-batch-3.0.tgz
+  NodePackage: s3://test/aws-parallelcluster-node-3.0.tgz
+  Timeouts:
+    HeadNodeBootstrapTimeout: 1201  # Default 1800 (seconds)
+    ComputeNodeBootstrapTimeout: 1001  # Default 1800 (seconds)
```
jdeamicis: Watch out, missing end-of-line character here.

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_validators_are_called_with_correct_argument/scheduler_plugin.yaml
```diff
@@ -0,0 +1,180 @@
...
+    Name: name2
+    StorageType: Efs
+    EfsSettings:
+      Encrypted: True
+      KmsKeyId: 1234abcd-12ab-34cd-56ef-1234567890ab
+  - MountDir: /my/mount/point3
+    Name: name3
+    StorageType: FsxLustre
+    FsxLustreSettings:
+      StorageCapacity: 3600
```
jdeamicis: Watch out, missing end-of-line character here.

jdeamicis (commented):
Good job, I just wonder if we should generalize the `test_all_validators...` functions to cover both normal scheduler and scheduler plugin.

cli/tests/pcluster/validators/test_all_validators.py
chenwany: Good catch, replaced with modules

cli/tests/pcluster/validators/test_all_validators.py
chenwany: I think it is better to split the two sets of validators. The reason is that: The cluster configurations are different, for example, you will see 
`    instance_architecture_compatibility_validator.assert_has_calls(
        [
            call(instance_type="t2.large", architecture="x86_64"),
            call(instance_type="c4.2xlarge", architecture="x86_64"),
            call(instance_type="c5.4xlarge", architecture="x86_64"),
            call(instance_type="c5d.xlarge", architecture="x86_64"),
            call(instance_type="t2.large", architecture="x86_64"),
        ]
    )`
in slurm.yaml and scheduler_plugin.yaml Although they are using same validators `instance_architecture_compatibility_validator`, they have different numbers of calls and also different parameter values. If we use pytest.mark.parametrize, there are lots of validators in the test, we will need to mark all these validators as parameters. Each time we add a new validator, we need to add a parameter in `pytest.mark.parametrize`, it is not easy to maintain.
What is more, there are some validators differences for slurm scheduler and plugin scheduler. Then we will add if... else...to check which is the scheduler if we put it in one test. IMO it is easier to add validators and test cases when in the future we add more validators for slurm and scheduler plugin.

cli/tests/pcluster/validators/test_all_validators.py
chenwany: Because these validators are only call for scheduler plugin. For slurm schedulers, they are not called. `test_all_validators_are_called` are for slurm scheduler. Renamed the function

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
chenwany: These log names are just some example to used for the validators. If people use slurm plugin, they can configure the logstream name from the paths.

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_validators_are_called_with_correct_argument/scheduler_plugin.yaml
chenwany: Thanks! added

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -295,3 +300,296 @@ def test_validators_are_called_with_correct_argument(test_datadir, mocker):
...
+    mocker.patch(
+        "pcluster.config.cluster_config.HeadNodeNetworking.availability_zone",
+        new_callable=PropertyMock(return_value="us-east-1a"),
+    )
+    mocker.patch(
+        "pcluster.config.cluster_config.BaseClusterConfig.head_node_ami",
+        new_callable=PropertyMock(return_value="ami-12345678"),
+    )
+    mocker.patch(
+        "pcluster.config.cluster_config.SchedulerPluginClusterConfig.get_instance_types_data",
```
hanwen-pcluste, line 327: Can we put these in to a function and reuse it in the slurm test?

hanwen-pcluste (approved):

cli/tests/pcluster/validators/test_all_validators.py
chenwany: Here some mocker patch are different from slurm and plugin 
`SchedulerPluginComputeResource. architecture ` and `SchedulerPluginClusterConfig. get_instance_types_data`, they can not be reused in slurm test.
Only the fields from the base config can be reused

PR #4247, opened 2022-08-04
[3.3][DFSM] Move head node tags from launch template to instance definition.

gmarciani:
### Description of changes
Moved head node tags from head node launch template to head node instance definition. This  change is required by Dynamic File systems Mount (DFSM) in order to not trigger head node replacement on file systems changes.

Added unit tests that verifies tags in head node launch template and instance definition.

### Tests
1. Unit tests
2. Created a cluster and verified that the head node has the expected tags. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
 **CHANGES**
 - Remove support for Python 3.6 in aws-parallelcluster-batch-cli.
 - Upgrade Python and NodeJS versions in API infrastructure, API Docker container and cluster Lambda resources.
+- Move head node tags from launch template to instance definition to avoid head node replacement on tags updates.
 
 3.2.0
 ------
--- a/cli/tests/pcluster/templates/test_cluster_stack.py
+++ b/cli/tests/pcluster/templates/test_cluster_stack.py
@@ -240,3 +240,127 @@ def _get_cfn_init_file_content(template, resource, file):
     content_separator = content_join[0]
     content_elements = content_join[1]
     return content_separator.join(str(elem) for elem in content_elements)
+
+
+@freeze_time("2021-01-01T01:01:01")
+@pytest.mark.parametrize(
+    "config_file_name, expected_instance_tags, expected_volume_tags",
+    [
+        (
+            "slurm.full.yaml",
+            {},
+            {
+                "parallelcluster:cluster-name": "clustername",
+                "parallelcluster:node-type": "HeadNode",
+                "String": "String",
+                "two": "two22",
+            },
+        ),
+        (
+            "awsbatch.full.yaml",
+            {},
+            {
+                "parallelcluster:cluster-name": "clustername",
+                "parallelcluster:node-type": "HeadNode",
+                "String": "String",
+                "two": "two22",
+            },
+        ),
+        (
+            "scheduler_plugin.full.yaml",
+            {},
+            {
+                "parallelcluster:cluster-name": "clustername",
+                "parallelcluster:node-type": "HeadNode",
+                "String": "String",
+                "two": "two22",
+            },
+        ),
+    ],
+)
+def test_head_node_tags_from_launch_template(mocker, config_file_name, expected_instance_tags, expected_volume_tags):
+    mock_aws_api(mocker)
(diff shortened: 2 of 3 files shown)


chenwany (commented):
Why don't we move tags outside launch template and use tags to the instance instead? 
```
 head_node_instance = ec2.CfnInstance(
        self,
        "HeadNode",
        launch_template=ec2.CfnInstance.LaunchTemplateSpecificationProperty(
            launch_template_id=head_node_launch_template.ref,
            version=head_node_launch_template.attr_latest_version_number,
        ),
        tags=<move to here>,
    )
```

gmarciani:
> Why don't we move tags outside launch template and use tags to the instance instead?
> 
> ```
>  head_node_instance = ec2.CfnInstance(
>         self,
>         "HeadNode",
>         launch_template=ec2.CfnInstance.LaunchTemplateSpecificationProperty(
>             launch_template_id=head_node_launch_template.ref,
>             version=head_node_launch_template.attr_latest_version_number,
>         ),
>         tags=<move to here>,
>     )
> ```

Totally agree with you.
I did that way in the first proposal of this PR because at that time that was the approach we agreed on.
Since we changed our mind on this, I've revised the approach by moving tags from head node launch template to instance definition.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
 **CHANGES**
 - Remove support for Python 3.6 in aws-parallelcluster-batch-cli.
 - Upgrade Python and NodeJS versions in API infrastructure, API Docker container and cluster Lambda resources.
+- Move head node tags from launch template to instance definition to avoid head node replacement on tags updates.
```
enrico-usai, line 10: The first sentence might be: "Add support for Tags update."

With this change you're adding support for Tags update, so you can change the Tags Update Policy too (probably from `UNSUPPORTED` to `COMPUTE_FLEET_STOP`). If you don't want to do it right now (because this requires extending current integ tests for update functionality) let's add a todo in the     `tags = fields.Nested(...)` definition.

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -240,3 +240,127 @@ def _get_cfn_init_file_content(template, resource, file):
...
+            {
+                "parallelcluster:cluster-name": "clustername",
+                "parallelcluster:node-type": "HeadNode",
+                "String": "String",
+                "two": "two22",
+            },
+        ),
+    ],
+)
+def test_head_node_tags_from_launch_template(mocker, config_file_name, expected_instance_tags, expected_volume_tags):
```
enrico-usai, line 281: Great job adding this test?

Can you double check if all the tags listed here for the head node are present? https://docs.aws.amazon.com/parallelcluster/latest/ug/resources-tags-v3.html 

We can add a link to this documentation in the test too.

CHANGELOG.md
gmarciani: I prefer to keep the two things separated; so, I'll go with a TODO in this PR.
The goal of this PR is to unblock tags updates, not to introduce the complete support for such updates.
By the way, I think we should go with `QUEUE_UPDATE_STRATEGY` rather than `COMPUTE_FLEET_STOP `

cli/tests/pcluster/templates/test_cluster_stack.py
gmarciani: Among the tags listed in [ParallelCluster Docs](https://docs.aws.amazon.com/parallelcluster/latest/ug/resources-tags-v3.html), only those prefixed with `parallelcluster:` are set via instance definition.

I noticed that in our docs we are missing the following tags and must be aligned:
1. `parallelcluster:attributes`
2. `parallelcluster:filesystem`
3. `parallelcluster:networking`

Furthermore, the docs list the tag `parallelcluster:version` that is actually correct, but currently some extra refactoring is required to make it checkable within this test. Adding a TODO for that.

enrico-usai: We decided to not publish them in our official doc because they are internal tags and we're not sure we want to preserve them in future.

enrico-usai (approved):

PR #4250, opened 2022-08-08
[CLI] Add configuration parameters needed to specify multiple instance types for a compute resource

EddyMM:
### Description of changes
* As part of supporting multiple instance types in a compute resource, these configuration parameters are being added
```
Scheduling:
  Scheduler: slurm
  SlurmQueues:
  - Name: ...
    AllocationStrategy: lowest-price|capacity-optimized
    ComputeResources:
    - Name: ...
      InstanceTypeList:
        - InstanceType: ...
        - InstanceType: ...
```

- A Compute Resource may define a list of possible instance-types that can be launched based on an allocation strategy
- A queue may have compute resources that define a single instance type or a list of instance-types
- The AllocationStrategy is defined at Queue Level and will apply to all the compute fleets launched under that queue.

### Tests
* A queue can list all instance types it contains (including the compute resources with multiple instance types)
* A queue can support compute Resources containing a either Single Instance Type or Multiple Instance Types 
* For launch templates generated for a compute fleet with multiple instance types:
    * The smallest MaxNetworkCards value will be used to determine the number of network interfaces to use
    * The InstanceType and CpuInfo fields are excluded

### References
* [AllocationStrategy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ CHANGELOG
 - Remove support for Python 3.6 in aws-parallelcluster-batch-cli.
 - Upgrade Python and NodeJS versions in API infrastructure, API Docker container and cluster Lambda resources.
 - Move head node tags from Launch Template to instance definition to avoid head node replacement on tags updates.
-- Disable Multithreading through script executed by cloud-init and not through CpuOptions set into Launch Template.   
+- Disable Multithreading through script executed by cloud-init and not through CpuOptions set into Launch Template.
+- Add support for multiple instance types in the same Compute Resource.
 
 **BUG FIXES**
 - Fix validation of parameter `SharedStorage/EfsSettings`: now validation fails when `FileSystemId` is specified 
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -220,12 +220,13 @@ class InstanceTypeInfo:
         return self.instance_type_data.get("InstanceType")
 
     def is_ebs_optimized(self):
-        """Check whether the instance has optimized EBS support."""
-        ebs_optimized = False
-        ebs_info = self.instance_type_data.get("EbsInfo")
-        if ebs_info:
-            ebs_optimized = ebs_info.get("EbsOptimizedSupport") != "unsupported"
-        return ebs_optimized
+        """
+        Check whether the instance has optimized EBS support.
+
+        Defaults to `unsupported` if EbsInfo is not available for the instance type.
+        """
+        support_level = self.instance_type_data.get("EbsInfo", {}).get("EbsOptimizedSupport", "unsupported")
+        return support_level != "unsupported"
 
     def supported_usage_classes(self):
         """Return the list supported usage classes."""
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -23,7 +23,6 @@ from pcluster.constants import (
     OS_TO_IMAGE_NAME_PART_MAP,
     PCLUSTER_IMAGE_BUILD_STATUS_TAG,
     PCLUSTER_IMAGE_ID_TAG,
-    SUPPORTED_ARCHITECTURES,
 )
 from pcluster.utils import get_partition
 
@@ -244,10 +243,7 @@ class Ec2Client(Boto3Client):
     def get_supported_architectures(self, instance_type):
         """Return a list of architectures supported for the given instance type."""
         instance_info = self.get_instance_type_info(instance_type)
-        supported_architectures = instance_info.supported_architecture()
-
-        # Some instance types support multiple architectures (x86_64 and i386). Filter unsupported ones.
-        return list(set(supported_architectures) & set(SUPPORTED_ARCHITECTURES))
+        return instance_info.supported_architecture()
 
     @AWSExceptionHandler.handle_client_exception
     @Cache.cached
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -13,6 +13,7 @@
 # These objects are obtained from the configuration file through a conversion based on the Schema classes.
 #
 import logging
+from abc import abstractmethod
 from collections import defaultdict
 from enum import Enum
 from typing import Dict, List, Union
@@ -1585,12 +1586,11 @@ class AwsBatchClusterConfig(BaseClusterConfig):
         return pkg_resources.resource_filename(__name__, "../resources/batch")
 
 
-class SlurmComputeResource(BaseComputeResource):
+class _BaseSlurmComputeResource(BaseComputeResource):
     """Represent the Slurm Compute Resource."""
 
     def __init__(
         self,
-        instance_type: str = None,
         max_count: int = None,
         min_count: int = None,
         spot_price: float = None,
@@ -1600,16 +1600,121 @@ class SlurmComputeResource(BaseComputeResource):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.instance_type = Resource.init_param(instance_type)
         self.max_count = Resource.init_param(max_count, default=DEFAULT_MAX_COUNT)
         self.min_count = Resource.init_param(min_count, default=DEFAULT_MIN_COUNT)
         self.spot_price = Resource.init_param(spot_price)
         self.disable_simultaneous_multithreading = Resource.init_param(
             disable_simultaneous_multithreading, default=False
         )
-        self.__instance_type_info = None
         self.efa = efa or Efa(enabled=False, implied=True)
         self.schedulable_memory = Resource.init_param(schedulable_memory)
+        self._instance_types_with_instance_storage = []
+        self._instance_type_info_map = {}
+
+    @staticmethod
+    def fetch_instance_type_info(instance_type) -> InstanceTypeInfo:
+        """Return instance type information."""
+        return AWSApi.instance().ec2.get_instance_type_info(instance_type)
+
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the Compute Resource that have local instance storage."""
+        if not self._instance_types_with_instance_storage:
+            self._instance_types_with_instance_storage = [
+                instance_type
+                for instance_type in self.instance_types
+                if self.instance_type_info_map[instance_type].instance_storage_supported()
+            ]
+        return self._instance_types_with_instance_storage
+
+    @property
+    def instance_type_info_map(self) -> Dict[str, InstanceTypeInfo]:
+        """List of Instance Type information for each instance type in the compute resource.
+
+        :returns: [ "InstanceType1": {... InstanceTypeInfo ...}, "InstanceType1": {... InstanceTypeInfo ...} ]
+        """
+        if not self._instance_type_info_map:
+            self._instance_type_info_map = {
+                instance_type: self.fetch_instance_type_info(instance_type) for instance_type in self.instance_types
+            }
+        return self._instance_type_info_map
+
+    @property
+    @abstractmethod
+    def disable_simultaneous_multithreading_manually(self) -> bool:
+        pass
+
+    @property
+    @abstractmethod
+    def max_network_interface_count(self) -> int:
+        pass
+
+    @property
+    def is_ebs_optimized(self) -> bool:
+        return all(
+            self.instance_type_info_map[instance_type].is_ebs_optimized() for instance_type in self.instance_types
+        )
+
+    @property
+    @abstractmethod
+    def instance_types(self) -> List[str]:
+        pass
+
+
+class FlexibleInstanceType(Resource):
+    """Represent an instance type listed in the InstanceTypeList of a ComputeResources."""
+
+    def __init__(self, instance_type: str, **kwargs):
+        super().__init__(**kwargs)
+        self.instance_type = Resource.init_param(instance_type)
+
+
+class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
+    """Represents a Slurm Compute Resource with Multiple Instance Types."""
+
+    def __init__(self, instance_type_list: List[FlexibleInstanceType], **kwargs):
+        super().__init__(**kwargs)
+        self.instance_type_list = Resource.init_param(instance_type_list)
+
+    @property
+    def instance_types(self) -> List[str]:
+        """Return list of instance type names in this compute resource."""
+        return [flexible_instance_type.instance_type for flexible_instance_type in self.instance_type_list]
+
+    @property
+    def disable_simultaneous_multithreading_manually(self) -> bool:
+        """Return true if simultaneous multithreading must be disabled with a cookbook script."""
+        return self.disable_simultaneous_multithreading
+
+    @property
+    def max_network_interface_count(self) -> int:
+        """Return max number of NICs for the compute resource.
+
+        In this case the Compute Resource may have multiple instance types, hence the instance-type with
+        the least MaxNetworkInterfaceCards value will be considered.
+        """
+        least_max_nics = self.instance_type_info_map[self.instance_types[0]].max_network_interface_count()
+        if len(self.instance_types) > 1:
+            for instance_type in self.instance_types[1:]:
+                instance_type_info = self.instance_type_info_map[instance_type]
+                max_nics = instance_type_info.max_network_interface_count()
+                if max_nics < least_max_nics:
+                    least_max_nics = max_nics
+        return least_max_nics
+
+
+class SlurmComputeResource(_BaseSlurmComputeResource):
+    """Represents a Slurm Compute Resource with a Single Instance Type."""
+
+    def __init__(self, instance_type, **kwargs):
+        super().__init__(**kwargs)
+        self.instance_type = Resource.init_param(instance_type)
+        self.__instance_type_info = None
+
+    @property
+    def instance_types(self) -> List[str]:
+        """List of instance types under this compute resource."""
+        return [self.instance_type]
 
     @property
     def instance_type_info(self) -> InstanceTypeInfo:
@@ -1697,17 +1802,36 @@ class _CommonQueue(BaseQueue):
 
 
 class SlurmQueue(_CommonQueue):
-    """Represent the Slurm Queue resource."""
+    """Represents a Slurm Queue that has Compute Resources with both Single and Multiple Instance Types."""
 
     def __init__(
         self,
-        compute_resources: List[SlurmComputeResource],
+        compute_resources: List[_BaseSlurmComputeResource],
         networking: SlurmQueueNetworking,
+        allocation_strategy: str = "lowest-price",
         **kwargs,
     ):
         super().__init__(**kwargs)
         self.compute_resources = compute_resources
         self.networking = networking
+        if any(isinstance(compute_resource, SlurmFlexibleComputeResource) for compute_resource in compute_resources):
+            self.allocation_strategy = allocation_strategy
+
+    @property
+    def instance_type_list(self):
+        """Return the list of instance types associated to the Queue."""
+        instance_types = set()
+        for compute_resource in self.compute_resources:
+            instance_types.update(compute_resource.instance_types)
+        return list(instance_types)
+
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
+        for compute_resource in self.compute_resources:
+            result.update(compute_resource.instance_types_with_instance_storage)
+        return result
 
     def _register_validators(self):
         super()._register_validators()
@@ -1724,9 +1848,6 @@ class SlurmQueue(_CommonQueue):
         )
         for compute_resource in self.compute_resources:
             self._register_validator(
-                CapacityTypeValidator, capacity_type=self.capacity_type, instance_type=compute_resource.instance_type
-            )
-            self._register_validator(
                 EfaSecurityGroupValidator,
                 efa_enabled=compute_resource.efa.enabled,
                 security_groups=self.networking.security_groups,
@@ -1737,20 +1858,12 @@ class SlurmQueue(_CommonQueue):
                 efa_enabled=compute_resource.efa.enabled,
                 placement_group=self.networking.placement_group,
             )
-
-    @property
-    def instance_type_list(self):
-        """Return the list of instance types associated to the Queue."""
-        return [compute_resource.instance_type for compute_resource in self.compute_resources]
-
-    @property
-    def instance_types_with_instance_storage(self):
-        """Return a set of instance types in the queue that have instance store."""
-        result = set()
-        for compute_resource in self.compute_resources:
-            if compute_resource.instance_type_info.instance_storage_supported():
-                result.add(compute_resource.instance_type)
-        return result
+            for instance_type in compute_resource.instance_types:
+                self._register_validator(
+                    CapacityTypeValidator,
+                    capacity_type=self.capacity_type,
+                    instance_type=instance_type,
+                )
 
 
 class Dns(Resource):
@@ -2226,14 +2339,15 @@ class CommonSchedulerClusterConfig(BaseClusterConfig):
                 checked_images.append(queue_image)
                 self._register_validator(AmiOsCompatibleValidator, os=self.image.os, image_id=queue_image)
             for compute_resource in queue.compute_resources:
-                self._register_validator(
-                    InstanceTypeBaseAMICompatibleValidator,
-                    instance_type=compute_resource.instance_type,
-                    image=queue_image,
-                )
+                for instance_type in compute_resource.instance_types:
+                    self._register_validator(
+                        InstanceTypeBaseAMICompatibleValidator,
+                        instance_type=instance_type,
+                        image=queue_image,
+                    )
                 self._register_validator(
                     InstanceArchitectureCompatibilityValidator,
-                    instance_type=compute_resource.instance_type,
+                    instance_type_info_list=list(compute_resource.instance_type_info_map.values()),
                     architecture=self.head_node.architecture,
                 )
                 self._register_validator(
@@ -2315,8 +2429,9 @@ class SlurmClusterConfig(CommonSchedulerClusterConfig):
         result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
         for queue in self.scheduling.queues:
             for compute_resource in queue.compute_resources:
-                instance_type_info = compute_resource.instance_type_info
-                result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
+                for instance_type in compute_resource.instance_types:
+                    instance_type_info = compute_resource.instance_type_info_map[instance_type]
+                    result[instance_type] = instance_type_info.instance_type_data
         return result
 
     def _register_validators(self):
@@ -2338,11 +2453,12 @@ class SlurmClusterConfig(CommonSchedulerClusterConfig):
         for queue in self.scheduling.queues:
             for compute_resource in queue.compute_resources:
                 if self.scheduling.settings.enable_memory_based_scheduling:
-                    self._register_validator(
-                        InstanceTypeMemoryInfoValidator,
-                        instance_type=compute_resource.instance_type,
-                        instance_type_data=instance_types_data[compute_resource.instance_type],
-                    )
+                    for instance_type in compute_resource.instance_types:
+                        self._register_validator(
+                            InstanceTypeMemoryInfoValidator,
+                            instance_type=instance_type,
+                            instance_type_data=instance_types_data[compute_resource.instance_type],
+                        )
 
     @property
     def image_dict(self):
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -51,6 +51,7 @@ from pcluster.config.cluster_config import (
     EphemeralVolume,
     ExistingFsxOntap,
     ExistingFsxOpenZfs,
+    FlexibleInstanceType,
     HeadNode,
     HeadNodeImage,
     HeadNodeNetworking,
@@ -96,6 +97,7 @@ from pcluster.config.cluster_config import (
     SharedFsxLustre,
     SlurmClusterConfig,
     SlurmComputeResource,
+    SlurmFlexibleComputeResource,
     SlurmQueue,
     SlurmQueueNetworking,
     SlurmScheduling,
@@ -1051,6 +1053,17 @@ class QueueCustomActionsSchema(BaseSchema):
         return CustomActions(**data)
 
 
+class InstanceTypeSchema(BaseSchema):
+    """Schema of a compute resource that supports a pool of instance types."""
+
+    instance_type = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return FlexibleInstanceType(**data)
+
+
 class HeadNodeSchema(BaseSchema):
     """Represent the schema of the HeadNode."""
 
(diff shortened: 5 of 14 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1614,12 +1614,11 @@ def scheduler_resources(self):
         return pkg_resources.resource_filename(__name__, "../resources/batch")
 
 
-class SlurmComputeResource(BaseComputeResource):
+class BaseSlurmComputeResource(BaseComputeResource):
```
enrico-usai: minor: we can add a starting underscore to identify it's a class used within the module only.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,17 +1628,85 @@ def __init__(
...
+
+class FlexibleInstanceType(Resource):
+    """Represent an instance type listed under InstanceTypes under ComputeResources."""
+
+    def __init__(self, instance_type: str, **kwargs):
+        super().__init__(**kwargs)
+        self.instance_type = Resource.init_param(instance_type)
+
+
+def get_instance_type_info(instance_type) -> InstanceTypeInfo:
```
enrico-usai: I'd avoid defining a function that is one line and with the same name of the called function, we can call the `get_instance_type_info` directly.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,17 +1628,85 @@ def __init__(
...
+        super().__init__(**kwargs)
+        self.instance_types = Resource.init_param(instance_types)
+
+    @property
+    def instance_type_names(self) -> List[str]:
+        """Return list of instance type names in this compute resource."""
+        return [flexible_instance_type.instance_type for flexible_instance_type in self.instance_types]
+
+    @property
+    def disable_simultaneous_multithreading_manually(self) -> bool:
```
enrico-usai, line 1685: Here we're overriding methods from the parent class.

What about removing these methods from `BaseSlurmComputeResource` and have them defined just in the `SlurmFlexibleComputeResource` and  `SlurmComputeResource`.

Then we could mark `BaseSlurmComputeResource`  as abstract class and set the methods as `abstractmethods` to be sure they are defined by children classes.

cli/src/pcluster/config/cluster_config.py
enrico-usai: or we can define this function into the `SlurmFlexibleComputeResource` as static method given that it's used just by that class.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,17 +1628,85 @@ def __init__(
...
+                instance_type_info = get_instance_type_info(instance_type)
+                max_nics = instance_type_info.max_network_interface_count()
+                if max_nics < least_max_nics:
+                    least_max_nics = max_nics
+        return least_max_nics
+
+    @property
+    def is_ebs_optimized(self) -> bool:
+        """Return True if all the instances have optimized EBS support."""
+        for instance_type in self.instance_type_names:
```
enrico-usai: nit: If I'm not wrong this can be rewritten with `any`

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1173,6 +1211,9 @@ class SlurmQueueSchema(_CommonQueueSchema):
     @post_load
     def make_resource(self, data, **kwargs):
         """Generate resource."""
+        allocation_strategy = data.get("allocation_strategy")
```
enrico-usai: ComputeResources with InstanceTypes and InstanceType can coexist in the same Queue. It means we cannot have SlurmFlexibleQueue, but add all the logic to manage both of them to the SlurmQueue.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1161,6 +1195,10 @@ class _CommonQueueSchema(BaseQueueSchema):
 class SlurmQueueSchema(_CommonQueueSchema):
     """Represent the schema of a Slurm Queue."""
 
+    allocation_strategy = fields.Str(
+        validate=validate.OneOf(["lowest-price", "capacity-optimized"]),
+        metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP},
```
enrico-usai: Did we evaluate `QUEUE_UPDATE_STRATEGY`?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1796,6 +1860,39 @@ def _register_validators(self):
                 placement_group=self.networking.placement_group,
             )
 
+
+class SlurmFlexibleQueue(BaseSlurmQueue):
```
enrico-usai: We cannot have a separate Queue class for flexible instance types. Both types of compute resources should coexist in the same Queue.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2373,8 +2485,13 @@ def get_instance_types_data(self):
         result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
         for queue in self.scheduling.queues:
             for compute_resource in queue.compute_resources:
-                instance_type_info = compute_resource.instance_type_info
-                result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
+                if isinstance(compute_resource, SlurmComputeResource):
+                    instance_type_info = compute_resource.instance_type_info
+                    result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
+                if isinstance(compute_resource, SlurmFlexibleComputeResource):
+                    for instance_type in compute_resource.instance_type_names:
```
enrico-usai: can we generalize the compute_resource interface to return a list of instance_type_names for both the cases?

enrico-usai: I'd add a property called:  `instance_type_list` in the base compute resource (like the one we have in the Queue), that returns a list with an item in the basic case or a list of instance types in the flexible case.

enrico-usai: I'm for calling the function "instance_type_list" and not "instance_type_names", to avoid introducing another naming. We're already using "instance_type_list" at queue level so we can have the same at compute resource level.

cli/src/pcluster/config/cluster_config.py
EddyMM: I think [all](https://docs.python.org/3/library/functions.html#all) is what would work here.
I went for this approach to avoid trying to build an iterable/list in order to use `all`. 
But there might be a possibility to use a generator (or generator comprehension). I'll give it a try to see the results.

EddyMM: Eventually decided to avoid using `all` here.
Using `all` works but it's not easily readable.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1082,7 +1095,12 @@ class _ComputeResourceSchema(BaseSchema):
 class SlurmComputeResourceSchema(_ComputeResourceSchema):
     """Represent the schema of the Slurm ComputeResource."""
 
-    instance_type = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    instance_type = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    instance_types = fields.Nested(
+        FlexibleInstanceTypeSchema,
+        many=True,
+        metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP, "update_key": "InstanceType"},
```
lukeseawalker: UpdatePolicy.COMPUTE_FLEET_STOP_ON_REMOVE

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1082,7 +1095,12 @@ class _ComputeResourceSchema(BaseSchema):
 class SlurmComputeResourceSchema(_ComputeResourceSchema):
     """Represent the schema of the Slurm ComputeResource."""
 
-    instance_type = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    instance_type = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    instance_types = fields.Nested(
+        FlexibleInstanceTypeSchema,
```
lukeseawalker: What about going for "InstanceTypeSchema"? I think it's simpler

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,17 +1629,127 @@ def __init__(
...
+        return False
+
+    @property
+    def max_network_interface_count(self) -> int:
+        """Return max number of NICs for the compute resource.
+
+        In this case the Compute Resource may have multiple instance types, hence the instance-type with
+        the least MaxNetworkInterfaceCards value will be considered.
+        """
+        least_max_nics = SlurmFlexibleComputeResource.get_instance_type_info(
```
lukeseawalker: if this method works for a instance type list with just 1 element, I will define it in the Base class so that it can be used for both SlurmFlexibleComputeResource and SlurmComputeResource

cli/src/pcluster/schemas/cluster_schema.py
EddyMM: I initially used `InstanceTypeSchema` but figured it might be confused with the properties in an InstanceTypeInfo object.
But since this schema is used in the context of a FlexibleComputeResource, I can simply use `InstanceTypeSchema`.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,17 +1629,127 @@ def __init__(
...
+        if len(self.instance_types) > 1:
+            for instance_type in self.instance_types[1:]:
+                instance_type_info = SlurmFlexibleComputeResource.get_instance_type_info(instance_type)
+                max_nics = instance_type_info.max_network_interface_count()
+                if max_nics < least_max_nics:
+                    least_max_nics = max_nics
+        return least_max_nics
+
+    @property
+    def is_ebs_optimized(self) -> bool:
```
lukeseawalker: same here, this can be defined in the Base class

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2284,16 +2403,17 @@ def _register_validators(self):
...
-                )
+                for instance_type in compute_resource.instance_types:
+                    self._register_validator(
+                        InstanceTypeBaseAMICompatibleValidator,
+                        instance_type=instance_type,
+                        image=queue_image,
+                    )
+                    self._register_validator(
+                        InstanceArchitectureCompatibilityValidator,
+                        instance_type=instance_type,
```
lukeseawalker: let's think about passing directly a list of InstanceTypeInfo and we move the architecture method in there

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1755,17 +1865,37 @@ def queue_ami(self):
...
+        for compute_resource in self.compute_resources:
+            instance_types.update(compute_resource.instance_types)
+        return list(instance_types)
+
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
+        for instance_type in self.instance_type_list:
+            if SlurmFlexibleComputeResource.get_instance_type_info(instance_type).instance_storage_supported():
```
lukeseawalker: I think we shouldn't use the SlurmFlexibleComputeResource here.
What about having a method to get the list of instance type info, something like instance_type_info_list?
This method will be exposed by the compute_resource, by both the SlurmFlexibleComputeResource and SlurmComputeResource, but in this way, the caller (instance_types_with_instance_storage) doesn't have to know about them.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2373,8 +2493,9 @@ def get_instance_types_data(self):
         result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
         for queue in self.scheduling.queues:
             for compute_resource in queue.compute_resources:
-                instance_type_info = compute_resource.instance_type_info
-                result[instance_type_info.instance_type()] = instance_type_info.instance_type_data
+                for instance_type in compute_resource.instance_types:
+                    instance_type_info = SlurmFlexibleComputeResource.get_instance_type_info(instance_type)
```
lukeseawalker: same consideration here, compute_resource should have a method to retrieve the instance_type_info_list

cli/tests/pcluster/config/test_cluster_config.py
```diff
@@ -60,6 +62,29 @@ def base_slurm_cluster_config(self):
...
         )
 
+    def test_instance_type_list_in_slurm_queue(self):
+        queue = SlurmQueue(
+            name="queue0",
+            networking=SlurmQueueNetworking(subnet_ids=["subnet"]),
+            compute_resources=[
+                SlurmComputeResource(name="compute_resource_1", instance_type="c5n.4xlarge"),
+                SlurmComputeResource(name="compute_resource_2", instance_type="t2.micro"),
+                SlurmFlexibleComputeResource(
```
lukeseawalker, line 72: add a SlurmFlexibleComputeResource with just one element

cli/src/pcluster/config/cluster_config.py
EddyMM: I agree, the `Queue` Class shouldn't rely on the `get_instance_type_info` method from `SlurmFlexibleComputeResource`.
I'll push the responsibility to obtain the list of instance_types that have instance store down to the compute resources as suggested.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ CHANGELOG
 - Remove support for Python 3.6 in aws-parallelcluster-batch-cli.
 - Upgrade Python and NodeJS versions in API infrastructure, API Docker container and cluster Lambda resources.
 - Move head node tags from launch template to instance definition to avoid head node replacement on tags updates.
+- Support using multiple instance types in a compute resource based on an allocation strategy
```
enrico-usai: I'd write: Add support for multiple instance types in the same Compute Resource.
Anyway missing final period.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -236,6 +236,19 @@ def is_ebs_optimized(self):
...
+
+        See API Reference:
+        https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html#:~:text=ebs%2Dinfo.ebs
+        %2Doptimized%2Dsupport.
+        Defaults to `unsupported` if EbsInfo is not available for the instance type.
+        """
+        ebs_info = self.instance_type_data.get("EbsInfo")
+        if ebs_info:
+            return ebs_info.get("EbsOptimizedSupport")
+        return "unsupported"
```
enrico-usai: an alternative one-liner way is:
```python
return self.instance_type_data.get("EbsInfo", {}).get("EbsOptimizedSupport", "unsupported")
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,16 +1629,140 @@ def __init__(
...
+        pass
+
+    @property
+    @abstractmethod
+    def max_network_interface_count(self) -> int:
+        pass
+
+    @property
+    def is_ebs_optimized(self) -> bool:
+        if "unsupported" in [
```
enrico-usai: Why did you define `ebs_optimized_support()`? We already had `is_ebs_optimized` in the instance_type_info class that is doing the check for `unsupported` for you.

enrico-usai: Then you can have an one-liner check:
```python
return all([self.instance_type_info_map[instance_type].is_ebs_optimized() for instance_type in self.instance_types])

File: cli/src/pcluster/config/cluster_config.py [hunk 20]
```diff
@@ -1629,16 +1629,140 @@ def __init__(
...
+
+    @property
+    def max_network_interface_count(self) -> int:
+        """Return max number of NICs for the compute resource.
+
+        In this case the Compute Resource may have multiple instance types, hence the instance-type with
+        the least MaxNetworkInterfaceCards value will be considered.
+        """
+        least_max_nics = self.instance_type_info_map[self.instance_types[0]].max_network_interface_count()
+        if len(self.instance_types) > 1:
```
enrico-usai, line 1697: let's add a unit test for this property calculation

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1755,17 +1874,36 @@ def queue_ami(self):
...
+        instance_types = set()
+        for compute_resource in self.compute_resources:
+            instance_types.update(compute_resource.instance_types)
+        return list(instance_types)
+
+    @property
+    def instance_types_with_instance_storage(self):
+        """Return a set of instance types in the queue that have instance store."""
+        result = set()
+        for compute_resource in self.compute_resources:
```
enrico-usai, line 1832: you can use set comprehension:
```python
return {compute_resource.instance_types_with_instance_storage for compute_resource in self.compute_resources}
```

enrico-usai: same in the instance_type_list property.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1092,9 +1110,24 @@ class SlurmComputeResourceSchema(_ComputeResourceSchema):
...
+    def no_coexist_instance_type_flexibility(self, data, **kwargs):
+        """Validate that 'instance-type' and 'instance-types' do not co-exist."""
+        if self.fields_coexist(
+            data,
+            ["instance_type", "instance_type_list"],
+            one_required=True,
+            **kwargs,
+        ):
+            raise ValidationError(
+                "A Compute Resource needs to specify either a single InstanceType or a list of InstanceTypes."
```
enrico-usai: a ComputeResource ... or InstanceTypesList

cli/src/pcluster/aws/aws_resources.py
enrico-usai: Anyway do we really need this or can we use the existing `is_ebs_optimized` ?

If we need both let's call `ebs_optimized_support` from `is_ebs_opmitized`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,16 +1629,140 @@ def __init__(
...
         self.min_count = Resource.init_param(min_count, default=DEFAULT_MIN_COUNT)
         self.spot_price = Resource.init_param(spot_price)
         self.disable_simultaneous_multithreading = Resource.init_param(
             disable_simultaneous_multithreading, default=False
         )
-        self.__instance_type_info = None
         self.efa = efa or Efa(enabled=False, implied=True)
         self.schedulable_memory = Resource.init_param(schedulable_memory)
+        self.__instance_types_with_instance_storage = []
+        self.__instance_type_info_map = {}
```
enrico-usai: why are you using two underscore? I'd stay consistent with the rest of the code using a single underscore for the private attributes.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,16 +1629,140 @@ def __init__(
...
+
+    @staticmethod
+    def fetch_instance_type_info(instance_type) -> InstanceTypeInfo:
+        """Return instance type information."""
+        return AWSApi.instance().ec2.get_instance_type_info(instance_type)
+
+    @property
+    def instance_types_with_instance_storage(self):
+        if not self.__instance_types_with_instance_storage:
+            for instance_type in self.instance_types:
```
enrico-usai: nit: you can use list comprehension

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -176,6 +177,100 @@ def test_scheduler_plugin_substack(mocker, config_file_name, expected_scheduler_
...
+    # mock bucket initialization parameters
+    mock_bucket(mocker)
+
+    input_yaml, cluster = load_cluster_model_from_yaml(config_file_name, test_datadir)
+    generated_template = CDKTemplateBuilder().build_cluster_template(
+        cluster_config=cluster, bucket=dummy_cluster_bucket(), stack_name="clustername"
+    )
+
+    instance_type = get_launch_template_data_property("InstanceType")
+    assert bool(instance_type) == has_instance_type
```
enrico-usai: you can use `assert_that` from pytest for all the check you're doing.

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -65,6 +65,20 @@ def _load_and_validate(config_path):
...
+def _assert_instance_architecture(expected_instance_architecture_validator_input, validator):
+    for call_index, validator_call in enumerate(validator.call_args_list):
+        args, kwargs = validator_call
+        instance_type_list = [
+            instance_type_info.instance_type() for instance_type_info in kwargs.get("instance_type_info_list")
+        ]
+        architecture = kwargs.get("architecture")
+        expected_instance_type_list = expected_instance_architecture_validator_input[call_index].get("instance_types")
+        expected_architecture = expected_instance_architecture_validator_input[call_index].get("architecture")
+        assert len(instance_type_list) == len(expected_instance_type_list)
```
enrico-usai: you can use:
```python
assert_that(instance_type_list).is_length(len(expected_instance_type_list))
```

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -65,6 +65,20 @@ def _load_and_validate(config_path):
...
+        assert len(set(instance_type_list) - set(expected_instance_type_list)) == 0
+        assert architecture == expected_architecture
```
enrico-usai: assert_that(...).is_equal_to(...)

cli/src/pcluster/aws/ec2.py
```diff
@@ -244,9 +244,13 @@ def get_instance_type_info(self, instance_type):
...
-        supported_architectures = instance_info.supported_architecture()
+        return self.extract_supported_architectures(instance_info)
 
-        # Some instance types support multiple architectures (x86_64 and i386). Filter unsupported ones.
+    @staticmethod
+    def extract_supported_architectures(instance_type_info):
+        """Return a list that filters out unsupported architectures that an instance type may have."""
+        supported_architectures = instance_type_info.supported_architecture()
+        # Some instance types support multiple architectures (x86_64 and i386). Filter out unsupported ones.
         return list(set(supported_architectures) & set(SUPPORTED_ARCHITECTURES))
```
lukeseawalker: isn't this filtering out already done in InstanceTypeInfo.supported_architecture()?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1629,16 +1629,140 @@ def __init__(
...
+        """Return list of instance type names in this compute resource."""
+        return [flexible_instance_type.instance_type for flexible_instance_type in self.instance_type_list]
+
+    @property
+    def disable_simultaneous_multithreading_manually(self) -> bool:
+        """Return true if simultaneous multithreading must be disabled with a cookbook script."""
+        return self.disable_simultaneous_multithreading
+
+    @property
+    def disable_simultaneous_multithreading_via_cpu_options(self) -> bool:
```
lukeseawalker: this and following property have been removed as part of https://github.com/aws/aws-parallelcluster/pull/4260

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1051,6 +1053,17 @@ def make_resource(self, data, **kwargs):
         return CustomActions(**data)
 
 
+class InstanceTypeSchema(BaseSchema):
+    """Schema of a compute resource that supports a pool of instance types."""
+
+    instance_type = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+
```
lukeseawalker, line 1060: merge conflict

cli/src/pcluster/schemas/cluster_schema.py
lukeseawalker: docstring not updated

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1092,9 +1110,24 @@ class SlurmComputeResourceSchema(_ComputeResourceSchema):
...
+            raise ValidationError(
+                "A Compute Resource needs to specify either a single InstanceType or a list of InstanceTypes."
+            )
+
     @post_load
     def make_resource(self, data, **kwargs):
         """Generate resource."""
+        if data.get("instance_type_list"):
+            return SlurmFlexibleComputeResource(**data)
         return SlurmComputeResource(**data)
```
lukeseawalker, line 1129: What about "A Compute Resource needs to specify either InstanceType or InstanceTypeList."

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1392,25 +1394,33 @@ def _add_compute_resource_launch_template(
                 ),
             )
 
+        conditional_template_properties = {}
+
+        if compute_resource.is_ebs_optimized:
+            conditional_template_properties.update({"ebs_optimized": True})
+        if compute_resource.pass_cpu_options_in_launch_template:
```
lukeseawalker: cpu_options dropped in https://github.com/aws/aws-parallelcluster/pull/4260

cli/src/pcluster/config/cluster_config.py
EddyMM: I agree, we can reuse the `is_ebs_optimized` method from InstanceTypeInfo to find out EBS-Support at ComputeResource level.
I used `ebs_optimized_support()` originally to separate the logic used to determine EBS-Support followed at ComputeResource level vs InstanceType level.

cli/src/pcluster/config/cluster_config.py
EddyMM: This is tested in the test_cluster_stack: [#L235-L235](https://github.com/EddyMM/aws-parallelcluster/blob/wip/multiple-instance-types/cli/tests/pcluster/templates/test_cluster_stack.py#L235-L235)
It selects the least number of NICs in the list.

cli/src/pcluster/config/cluster_config.py
EddyMM: `compute_resource.instance_types_with_instance_storage` and `compute_resource.instance_types` return a **list** of instance_types so set comprehension won't work since lists are unhashable.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -221,11 +221,17 @@ def instance_type(self):
 
     def is_ebs_optimized(self):
         """Check whether the instance has optimized EBS support."""
-        ebs_optimized = False
-        ebs_info = self.instance_type_data.get("EbsInfo")
-        if ebs_info:
-            ebs_optimized = ebs_info.get("EbsOptimizedSupport") != "unsupported"
-        return ebs_optimized
+        return self.ebs_optimized_support() != "unsupported"
```
lukeseawalker: cannot we leave just `def is_ebs_optimized`?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1092,9 +1110,22 @@ class SlurmComputeResourceSchema(_ComputeResourceSchema):
     disable_simultaneous_multithreading = fields.Bool(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
     schedulable_memory = fields.Int(metadata={"update_policy": UpdatePolicy.QUEUE_UPDATE_STRATEGY})
 
+    @validates_schema
+    def no_coexist_instance_type_flexibility(self, data, **kwargs):
+        """Validate that 'instance-type' and 'instance-types' do not co-exist."""
```
lukeseawalker: fix doc string, e.g.
```
"""Validate that 'instance_type' and 'instance_type_list' do not co-exist."""
```

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -234,17 +237,22 @@ class InstanceArchitectureCompatibilityValidator(Validator):
...
-        if head_node_architecture not in compute_architectures:
-            self._add_failure(
-                "The specified compute instance type ({0}) supports the architectures {1}, none of which are "
-                "compatible with the architecture supported by the head node instance type ({2}).".format(
-                    instance_type, compute_architectures, head_node_architecture
-                ),
-                FailureLevel.ERROR,
-            )
+
+        compute_architectures = set()
```
lukeseawalker: the `set` is not needed because you need to compare head_node to each_compute resource and not head_node to the incremental set of compute_resources

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -316,6 +318,104 @@ def test_scheduling_schema(mocker, config_dict, failure_message):
...
+                    {
+                        "Name": "compute_resource1",
+                        "InstanceTypeList": [{"InstanceType": "c5.2xlarge"}, {"InstanceType": "c4.2xlarge"}],
+                    },
+                    {"Name": "compute_resource2", "InstanceType": "c4.2xlarge"},
+                ],
+            },
+            "",
+        ),
+        # Failing to specify either InstanceType or InstanceTypes should return a validation error
```
lukeseawalker: add test case where both (InstanceType or InstanceTypeList) are specified (for the same compute resource), this should return a validation error

EDIT, I saw it is covered in test_slurm_flexible_compute_resources. Is it possible to merge the two tests?

lukeseawalker: also fix comment InstanceTypes -> InstanceType or InstanceTypeList

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -176,6 +177,100 @@ def test_scheduler_plugin_substack(mocker, config_file_name, expected_scheduler_
...
+        ("cluster-using-flexible-instance-types.yaml", False, 2, False, False, None),
+        ("cluster-using-single-instance-type.yaml", True, 3, True, True, True),
+    ],
+)
+def test_compute_launch_template_properties(
+    mocker,
+    config_file_name,
+    has_instance_type,
+    no_of_network_interfaces,
+    includes_cpu_options,
```
lukeseawalker: includes_cpu_options not used

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1600,16 +1600,131 @@ def __init__(
...
         self.disable_simultaneous_multithreading = Resource.init_param(
             disable_simultaneous_multithreading, default=False
         )
-        self.__instance_type_info = None
         self.efa = efa or Efa(enabled=False, implied=True)
         self.schedulable_memory = Resource.init_param(schedulable_memory)
+        self._instance_types_with_instance_storage = []
+        self._instance_type_info_map = {}
+
+    @staticmethod
```
lukeseawalker, line 1614: can this be a property?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1600,16 +1600,131 @@ def __init__(
...
+        )
+
+    @property
+    @abstractmethod
+    def instance_types(self) -> List[str]:
+        pass
+
+
+class FlexibleInstanceType(Resource):
+    """Represent an instance type listed under InstanceTypes under ComputeResources."""
```
lukeseawalker: InstanceTypes -> InstanceTypeList

cli/src/pcluster/config/cluster_config.py
lukeseawalker: you can remove the `disable_simultaneous_multithreading_via_cpu_options` method

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1600,16 +1600,131 @@ def __init__(
...
+        """Return true if simultaneous multithreading must be disabled with a cookbook script."""
+        return self.disable_simultaneous_multithreading
+
+    @property
+    def disable_simultaneous_multithreading_via_cpu_options(self) -> bool:
+        """CPU Options not supported for Compute Resources with Flexible Instance Types."""
+        return False
+
+    @property
+    def pass_cpu_options_in_launch_template(self) -> bool:
```
lukeseawalker: you can remove this method, isn't used anywhere

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2226,14 +2349,15 @@ def _register_validators(self):
...
+                for instance_type in compute_resource.instance_types:
+                    self._register_validator(
+                        InstanceTypeBaseAMICompatibleValidator,
+                        instance_type=instance_type,
+                        image=queue_image,
+                    )
                 self._register_validator(
                     InstanceArchitectureCompatibilityValidator,
-                    instance_type=compute_resource.instance_type,
+                    instance_type_info_list=list(compute_resource.instance_type_info_map.values()),
```
lukeseawalker, line 2350: is `InstanceArchitectureCompatibilityValidator` accepting a instance_type_info_list ?

lukeseawalker: Let's keep passing `instance_type` for now (this require to add a for loop to iterate of the compute resource instance type

lukeseawalker: nevermind, I saw that InstanceArchitectureCompatibilityValidator was modified to reflect the changes

EddyMM: Yes.

lukeseawalker (approved):

Q-Hilbtec:
This is great, thanks @lukeseawalker! I noted that it is not currently compatible with Memory Based Scheduling - would it be possible to loosen the constraint and allow users to specify multiple instance types - assuming all of them share the same RAM amount?

PR #4253, opened 2022-08-08
Fix cluster update when changing the order of SharedStorage 

chenwany:
### Description of changes
* Before the fix, update the cluster with changing the orders of multiple shared storages will fail with shared storages replacement. This is because the logical ID of the shared storages are determined by the order in which they appear.
With the fix, changing ShareStorage orders during update will work. The logical ID of the shared storages will be decided by the shared storage name.

### Tests
* Manually test with create and update cluster with switching share storage order.

### References

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ CHANGELOG
 **BUG FIXES**
 - Fix validation of parameter `SharedStorage/EfsSettings`: now validation fails when `FileSystemId` is specified 
   along with other `SharedStorage/EfsSettings` parameters, whereas it was previously ignoring them.
+- Fix cluster update when changing the order of SharedStorage together with other changes in the configuration.
 
 3.2.0
 ------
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ CHANGELOG
 **BUG FIXES**
 - Fix validation of parameter `SharedStorage/EfsSettings`: now validation fails when `FileSystemId` is specified 
   along with other `SharedStorage/EfsSettings` parameters, whereas it was previously ignoring them.
+- Fix cluster update when switching the order of SharedStorage configuration.
```
demartinofra: if I'm not wrong we don't support yet an update of this section. In that case this is not really a bug but will be part of adding support for shared FS updates.

enrico-usai: It's a bug because it's triggering the update of the shared storages when there are no changes on them except for the order. 
This issue is triggered only when there are other changes in the config.

So probably we should add something like:
`...together with other changes in the configuration.`

chenwany: We support update of some of the parameters in SharedStorage section, for example, IOPS, Throughput, DeletionPolicy. Changelog is updated to add `together with other changes in the configuration.`

enrico-usai (approved):

PR #4255, opened 2022-08-10
[DFSM] Add SHARED_STORAGE_UPDATE_POLICY

chenwany:
### Description of changes
* Add new update policy `SHARED_STORAGE_UPDATE_POLICY` for SharedStorages section, it applies to adding and removing share storages and changing MountDir attribute.
  * It requires compute fleet stop when removing shared storages or changing `MountDir`
  * Update is UNSUPPORT for awsbatch scheduler
  * Computefleet stop is required for plugin scheduler
  * For adding new shared storages, you can either stop compute fleet or apply queue update strategy.

### Tests
* Unit tests
* Manual tests
  * Not stop compute fleet and add EBS, EFS, FSX without queue strategy
  ```
  "message": "All compute nodes must be stopped or QueueUpdateStrategy must be set. Stop the compute fleet with the 
  pcluster update-compute-fleet command, or set QueueUpdateStrategy in the configuration used for the 'update-cluster' 
  operation",
  "currentValue": "-"
  ```

  * Stop and add EBS, EFS, FSX without queue strategy -- OK
  * Remove EBS, EFS or FSX with queue strategy -- All compute nodes must be stopped. Stop the compute fleet with the pcluster update-compute-fleet command
  * Not stop and add EBS, EFS, FSX without queue strategy -- All compute nodes must be stopped or QueueUpdateStrategy must be set. Stop the compute fleet with the pcluster update-compute-fleet command, or set QueueUpdateStrategy in the configuration used for the 'update-cluster' operation"
  * Remove EFS wit queue strategy without stop -- "All compute nodes must be stopped. Stop the compute fleet with the pcluster update-compute-fleet command."
*  For awsbatch, changing the parameters is not support
*  For plugin scheduler, it requires compute fleet stop to add, remove sections or changing MountDir
### Checklist


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ CHANGELOG
 
 3.x.x
 -----
+**ENHANCEMENTS**
+- Add support for updating SharedStorage configuration.
```
gmarciani: I suggest to highlight the config name as we did in other lines, i.e. `SharedStorage`

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +125,39 @@ def is_slurm_queues_change(change):
     return any(path.startswith("SlurmQueues[") for path in change.path)
 
 
+def is_slurm_scheduler(patch):
+    return patch.cluster.stack.scheduler == "slurm"
```
gmarciani: Here and in other places in our codebase, we redefine the string "slurm".
I think we should define it once in a constant and reference the constant instead.

What about adding such a constant in `cli/src/pcluster/constants.py`?

We can use such constant here and do the larger refactoring later on in a dedicated activity.

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +125,39 @@ def is_slurm_queues_change(change):
...
+
+
+def is_awsbatch_scheduler(_, patch):
+    return patch.cluster.stack.scheduler == "awsbatch"
```
gmarciani: Here and in other places in our codebase, we redefine the string "awsbatch".
I think we should define it once in a constant and reference the constant instead.

What about adding such a constant in cli/src/pcluster/constants.py?

We can use such constant here and do the larger refactoring later on in a dedicated activity.

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +125,39 @@ def is_slurm_queues_change(change):
...
+        return True
+    return False
+
+
+def fail_reason_shared_storage_update_policy(change, patch):
+    if is_awsbatch_scheduler(change, patch):
+        return f"Update actions are not currently supported for the '{change.key}' parameter"
+    reason = "All compute nodes must be stopped"
+    # QueueUpdateStrategy can override UpdatePolicy of parameters under SlurmQueues
+    if not is_stop_required_for_shared_storage(change) and is_slurm_scheduler(patch):
```
gmarciani, line 155: With these conditions we would have the following case. If the scheduler is Slurm and a stop is not required, we would fail with message "All compute nodes must be stopped or QueueUpdateStrategy must be set". This may confuse the end-user. What do you think?

cli/src/pcluster/config/update_policy.py
```diff
@@ -149,11 +182,47 @@ def condition_checker_queue_update_strategy(change, patch):
...
+    if is_awsbatch_scheduler(change, patch):
+        return (
+            f"Restore '{change.key}' value to '{change.old_value}'"
+            if change.old_value is not None
+            else "{0} the parameter '{1}'".format("Restore" if change.is_list else "Remove", change.key)
+            + ". If you need this change, please consider creating a new cluster instead of updating the existing one."
+        )
+    actions = "Stop the compute fleet with the pcluster update-compute-fleet command"
+    if not is_stop_required_for_shared_storage(change) and is_slurm_scheduler(patch):
+        actions += ", or set QueueUpdateStrategy in the configuration used for the 'update-cluster' operation"
```
gmarciani, line 195: Missing full stop at the end of the message.

cli/src/pcluster/config/update_policy.py
```diff
@@ -149,11 +182,47 @@ def condition_checker_queue_update_strategy(change, patch):
...
 
+def actions_needed_shared_storage_update(change, patch):
+    if is_awsbatch_scheduler(change, patch):
+        return (
+            f"Restore '{change.key}' value to '{change.old_value}'"
+            if change.old_value is not None
+            else "{0} the parameter '{1}'".format("Restore" if change.is_list else "Remove", change.key)
+            + ". If you need this change, please consider creating a new cluster instead of updating the existing one."
+        )
+    actions = "Stop the compute fleet with the pcluster update-compute-fleet command"
```
gmarciani, line 193: Missing period at the end of the message if the condition below does not hold.

cli/src/pcluster/config/update_policy.py
```diff
@@ -149,11 +182,47 @@ def condition_checker_queue_update_strategy(change, patch):
...
+        actions += ", or set QueueUpdateStrategy in the configuration used for the 'update-cluster' operation"
+
+    return actions
+
+
+def condition_checker_shared_storage_update_policy(change, patch):
+    """
+    Check different requirements for different schedulers.
+
+    Compute fleet stop is required for awsbatch and plugin scheduler.
```
gmarciani: When the scheduler is AWS Batch then the shared storage update should not be support, right?

cli/src/pcluster/config/update_policy.py
```diff
@@ -149,11 +182,47 @@ def condition_checker_queue_update_strategy(change, patch):
...
+    QueueUpdateStrategy can override UpdatePolicy of parameters under SlurmQueues for slurm scheduler.
+    """
+    if is_awsbatch_scheduler(change, patch):
```
gmarciani, line 208: We cover awsbatch and slurm here, but the function comment also mentions the scheduler plugin.
What about the scheduler plugin?

cli/src/pcluster/config/update_policy.py
```diff
@@ -149,11 +182,47 @@ def condition_checker_queue_update_strategy(change, patch):
...
+        return None
```
gmarciani: I supposed this function would have returned only True or False. What is the meaning of returning None?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -491,7 +491,7 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
             ),
             iam.PolicyStatement(
                 sid="Ec2TagsAndVolumes",
-                actions=["ec2:AttachVolume", "ec2:CreateTags"],
+                actions=["ec2:AttachVolume", "ec2:CreateTags", "ec2:DetachVolume"],
```
gmarciani, line 494: This kind of change should be covered by unit tests.
I known that we do not have an existing unit test to modify, but we should.
May you please keep track of the lack of unit testing the iam policies created by the template?

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -453,3 +453,357 @@ def test_compute_fleet_stop_fail_reason_and_actions_needed(
...
+    new_value,
+    update_strategy,
+    expected_result,
+    expected_fail_reason,
+    expected_actions_needed,
+    scheduler,
+):
+    cluster = dummy_cluster()
+    cluster_has_running_capacity_mock = mocker.patch.object(
+        cluster, "has_running_capacity", return_value=not is_fleet_stopped
```
gmarciani, line 773: Minor comment: why having a variable `is_fleet_stopped` to negate instead of having a variable `is_fleet_running` to check positively?

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -453,3 +453,357 @@ def test_compute_fleet_stop_fail_reason_and_actions_needed(
...
+            [],
+            None,
+            {"MountDir": "/ebs3", "Name": "ebs3", "StorageType": "Ebs", "EbsSettings": {"VolumeType": "gp3"}},
+            None,
+            False,
+            "All compute nodes must be stopped or QueueUpdateStrategy must be set",
+            "Stop the compute fleet with the pcluster update-compute-fleet command, or set QueueUpdateStrategy in the "
+            "configuration used for the 'update-cluster' operation",
+            "slurm",
+            id="running fleet and EBS changed with no update strategy set",
```
gmarciani: We say here that the test case is "EBS changed", but we are actually adding a new EBS section right? If this is the case, let's rather say "add a new EBS section".

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -453,3 +453,357 @@ def test_compute_fleet_stop_fail_reason_and_actions_needed(
...
+            [],
+            None,
+            {"MountDir": "/ebs3", "Name": "ebs3", "StorageType": "Ebs", "EbsSettings": {"VolumeType": "gp3"}},
+            None,
+            None,
+            "Update actions are not currently supported for the 'SharedStorage' parameter",
+            "Restore the parameter 'SharedStorage'. If you need this change, please consider creating a new cluster "
+            "instead of updating the existing one.",
+            "awsbatch",
+            id="running fleet and EBS changed with no update strategy set with batch scheduler",
```
gmarciani: Typically we refer to the AWS Batch scheduler with the name "awsbatch" rather than "batch".

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -453,3 +453,357 @@ def test_compute_fleet_stop_fail_reason_and_actions_needed(
...
+            "SharedStorage",
+            [],
+            None,
+            {"MountDir": "/ebs3", "Name": "ebs3", "StorageType": "Ebs", "EbsSettings": {"VolumeType": "gp3"}},
+            QueueUpdateStrategy.TERMINATE.value,
+            True,
+            None,
+            None,
+            "slurm",
+            id="running fleet and Ebs added with update strategy TERMINATE",
```
gmarciani: In `id` string here and below: Ebs => EBS

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -453,3 +453,357 @@ def test_compute_fleet_stop_fail_reason_and_actions_needed(
...
+            "SharedStorage",
+            [],
+            None,
+            {"MountDir": "/efs4", "Name": "efs4", "StorageType": "Efs"},
+            QueueUpdateStrategy.TERMINATE.value,
+            True,
+            None,
+            None,
+            "slurm",
+            id="running fleet and EFs added with update strategy DRAIN",
```
gmarciani: The update strategy is actually TERMINATE, not DRAIN.

gmarciani (changes requested):

cli/src/pcluster/config/update_policy.py
chenwany: added

cli/src/pcluster/config/update_policy.py
chenwany: added to constants.py

cli/src/pcluster/config/update_policy.py
chenwany: Could you explain why do you think this is confusing? This is aligned with the error message when Parameters require QueueUpdateStrategy.

cli/src/pcluster/config/update_policy.py
chenwany: This is aligned with the error message when Parameters require QueueUpdateStrategy.

cli/src/pcluster/config/update_policy.py
chenwany: Yup, thanks, update the comment

cli/src/pcluster/config/update_policy.py
chenwany: The function also cover scheduler plugin. line `result = not patch.cluster.has_running_capacity()`, cover scheduler plugin case

cli/src/pcluster/config/update_policy.py
chenwany: good catch, this should return False

cli/tests/pcluster/config/test_update_policy.py
chenwany: I think `is_fleet_stopped` is a name better explain itself. Refactoring of `is_fleet_stopped` is a function defined before and it is out of scope here.
Also, `is_fleet_stopped` is not opposite to `is_fleet_running`. If the compute fleet have 0 compute nodes, the cluster compute fleet status is running, but it is not Is_fleet_stop.

cli/tests/pcluster/config/test_update_policy.py
chenwany: Changed the id

cli/tests/pcluster/config/test_update_policy.py
chenwany: Changed the id

cli/tests/pcluster/config/test_update_policy.py
chenwany: Changed the id

cli/tests/pcluster/config/test_update_policy.py
chenwany: Changed the id

cli/src/pcluster/templates/cdk_builder_utils.py
chenwany: Created a SIM to track it

cli/src/pcluster/config/update_policy.py
gmarciani: If a stop is not required, but we receive a message saying "All compute nodes must be stopped", then this sounds to be confusing. 
Nevertheless, I agree that we must be consistent with the messaging returned by QueueUpdateStrategy, so let's verify with the beta testing if such messages create confusion or not in the end user.
Let's proceed as it is.

cli/src/pcluster/config/update_policy.py
gmarciani: In general, I think that messages should end with a period.
Nevertheless, as long as we are consistent with ourselves that's fine. Let's go with it.

cli/src/pcluster/config/update_policy.py
gmarciani: Ok, got it

cli/src/pcluster/config/update_policy.py
gmarciani: In general, I think that messages should end with a period.
Nevertheless, as long as we are consistent with ourselves that's fine. Let's go with it.

cli/src/pcluster/templates/cdk_builder_utils.py
gmarciani: Thank you so much.

cli/tests/pcluster/config/test_update_policy.py
gmarciani: Got it thanks!

gmarciani (dismissed):

gmarciani (commented):
Diffs look good to me. Nevertheless, some checks are failing so please verify and fix the root cause before merging.

gmarciani (approved):

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -469,7 +471,9 @@ class SharedStorageSchema(BaseSchema):
     """Represent the generic SharedStorage schema."""
 
     mount_dir = fields.Str(
-        required=True, validate=get_field_validator("file_path"), metadata={"update_policy": UpdatePolicy.UNSUPPORTED}
+        required=True,
+        validate=get_field_validator("file_path"),
+        metadata={"update_policy": UpdatePolicy.SHARED_STORAGE_UPDATE_POLICY},
```
hanwen-cluster, line 476: Why is this change needed? Isn't the only change needed on [line 1942](https://github.com/aws/aws-parallelcluster/pull/4255/files#diff-ea5b87cce964033aecc0072b33dd35596bf4f5a3d63ae4dd76fb7b318ba22318R1942)

PR #4258, opened 2022-08-10
Fix integration test of retrieve slurm nodes

chenwany:
Fix test fast capacity tests and queue parameters update tests for node states assertion failure

### Description of changes

### Tests

### References

### Checklist


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -382,13 +382,15 @@ def assert_job_state(self, job_id, expected_state):  # noqa: D102
...
+    def get_compute_nodes(self, filter_by_partition=None, all_nodes=None):  # noqa: D102
         command = "sinfo --Node --noheader --responding"
         if filter_by_partition:
             command += " --partition {}".format(filter_by_partition)
         # Print first and fourth columns to get nodename and state only (default partition contains *)
         # Filter out nodes that are not responding or in power saving states
-        command += " | awk '{print $1, $4}' | grep -v '[*#~%]' | awk '{print $1}'"
+        command += (
+            " | awk '{print $1, $4}' | grep -v '[*#~%]' | awk '{print $1}'" if not all_nodes else " | awk '{print $1}'"
+        )
```
jdeamicis: I believe we should also move the `--responding` flag to the if condition in line 392: otherwise the case with `all_nodes` will still miss the non-responding nodes.

Out of scope: we should rewrite this whole method (by default this implementation doesn't return all nodes in a cluster, only the responding ones, and I am afraid this is not consistent with the definition in the parent class).

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -471,7 +473,7 @@ def wait_job_running(self, job_id):
 
     def get_node_info(self, nodename):
         """Get node info."""
-        return self._remote_command_executor.run_remote_command("scontrol show nodes {0}".format(nodename))
+        return self._remote_command_executor.run_remote_command("scontrol show nodes {0}".format(nodename)).stdout
```
jdeamicis, line 476: Is this method even used in the code?

Also, we might implement it the same way we implemented `get_job_info()` (a bit above in the code):

```python
    def get_job_info(self, job_id, field=None):
        """Return job details from slurm. If field is provided, only the field is returned"""
        result = self._remote_command_executor.run_remote_command("scontrol show jobs -o {0}".format(job_id)).stdout
        if field is not None:
            match = re.search(rf"(\s{field})=(\S*)", result)
            return match.group(2)
        return 
```

jdeamicis (changes requested):
In general I'm fine with the development: I'm proposing some improvements we can address in this PR.

tests/integration-tests/tests/common/schedulers_common.py
chenwany: removed

tests/integration-tests/tests/common/schedulers_common.py
chenwany: the method is used in the code. We don't use field=None here so lets consider add it when it is needed in the future

jdeamicis: Are you sure it's being used? I only see references to a `get_nodes_info` (with the s)  method used in node, but nothing about `get_node_info`. Anyway, it's not too important.

tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -382,13 +382,15 @@ def assert_job_state(self, job_id, expected_state):  # noqa: D102
...
+    def get_compute_nodes(self, filter_by_partition=None, all_nodes=None):  # noqa: D102
+        command = "sinfo --Node --noheader"
         if filter_by_partition:
             command += " --partition {}".format(filter_by_partition)
         # Print first and fourth columns to get nodename and state only (default partition contains *)
         # Filter out nodes that are not responding or in power saving states
-        command += " | awk '{print $1, $4}' | grep -v '[*#~%]' | awk '{print $1}'"
+        command += (
+            " | awk '{print $1, $4}' | grep -v '[*#~%]' | awk '{print $1}'" if not all_nodes else " | awk '{print $1}'"
+        )
```
jdeamicis, line 393: I'm not sure leaving the `| grep -v '[*#~%]'` is equivalent to having `--responding` in the `command` string.

In my [previous comment](https://github.com/aws/aws-parallelcluster/pull/4258#discussion_r942723551) I meant *moving* the `--responding` option into the if condition like this:
```python
        command += (
            " --responding | awk '{print $1, $4}' | grep -v '[*#~%]' | awk '{print $1}'" if not all_nodes else " | awk '{print $1}'"
        )
```

jdeamicis (changes requested):
We should probably keep the `--responding` flag in case `all_nodes` is not used.

tests/integration-tests/tests/common/schedulers_common.py
chenwany: yes I am sure. get_node_info is used, see code change in my PR: tests/integration-tests/tests/common/hit_common.py

tests/integration-tests/tests/common/schedulers_common.py
jdeamicis: We just checked the sinfo state codes, and we saw that `*` means non-responding. So the `grep -v '[*#~%]'` is already filtering out non-responding nodes. So we can safely remove `--responding` from the initial `command` string.

jdeamicis (approved):

hanwen-pcluste (approved):

PR #4260, opened 2022-08-10
[develop] Change logic for disable multithreading

lukeseawalker:

### Description of changes
* The manual disable multithreading logic has been moved from a chef recipe to cloud-init, using bootcmd directive, so that the disabling is persistent against instance reboot, and it's done in the early stages of instance boot.

The script used to perform the disable multithreading has been taken from https://aws.amazon.com/blogs/compute/disabling-intel-hyper-threading-technology-on-amazon-linux/, with a minor change to take into account the fact that certain EC2 instance types will have different delimiters in the thread_siblings_list depending on CPU architecture and Operating System.

```
cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list
| tr '-' ','       # change to account that some kernels and CPU architectures use a hyphen (-) instead of a comma (,)
| cut -s -d, -f2-
| tr ',' '\n'
| sort -un
```

* Drop CpuOption for disable Multithreading

Disable Multithreading is always performed through script executed by cloud-init and not through CpuOptions set into Launch Template anymore.

This is done using bootcmd directive so that the disabling is done very early in the boot process and it's persistent against instance reboot.

### Tests
* integration tests already in place
* changes tested locally with the following conf
```
---
scheduler-plugins:
  slurm_plugin:
    scheduler-definition: "../../scheduler_plugins/slurm/utils/upload_artifacts.sh"
    scheduler-commands: "tests.common.schedulers_common.SlurmCommands"
    requires-sudo: true
test-suites:
  disable_hyperthreading:
    test_disable_hyperthreading.py::test_hit_disable_hyperthreading:
      dimensions:
        - regions: ["eu-west-1"]
          instances: ["m4.xlarge"]
          oss: ["alinux2", "centos7"]
          schedulers: ["slurm"]
        - regions: ["eu-west-1"]
          instances: ["c5.xlarge"]
          oss: ["ubuntu1804", "ubuntu2004"]
          schedulers: ["slurm"]
          benchmarks:
            - mpi_variants: ["openmpi", "intelmpi"]
              num_instances: [20] # Change the head node instance type if you'd test more than 30 instances
              slots_per_instance: 2
              partition: "ht-disabled"
              osu_benchmarks:
                collective: ["osu_allreduce", "osu_alltoall"]
```

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1497

### Checklist


CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ CHANGELOG
 **CHANGES**
 - Remove support for Python 3.6 in aws-parallelcluster-batch-cli.
 - Upgrade Python and NodeJS versions in API infrastructure, API Docker container and cluster Lambda resources.
-- Move head node tags from launch template to instance definition to avoid head node replacement on tags updates.
+- Move head node tags from Launch Template to instance definition to avoid head node replacement on tags updates.
+- Disable Multithreading is performed through script executed by cloud-init and not through CpuOptions set into Launch Template anymore.
```
enrico-usai: What about:
`Disable multithreading through ...`

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -39,6 +39,9 @@ fi
 Content-Type: text/cloud-config; charset=us-ascii
 MIME-Version: 1.0
 
+bootcmd:
+  - if [ "${DisableHyperThreadingManually}" = "true" ]; then for cpunum in $(cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list | tr '-' ',' | cut -s -d, -f2- | tr ',' '\n' | sort -un); do echo 0 > /sys/devices/system/cpu/cpu$cpunum/online; done; fi
```
enrico-usai: I'd add a comment here to explain what is the expected input and output of the cat | commands.

or add something like:
`First, it finds the total number of CPUs. Then, it loops over the second half of the CPUs. Finally, it disables each CPU in turn.`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1699,23 +1654,10 @@ def _instance_type_info(self) -> InstanceTypeInfo:
...
-
     @property
     def disable_simultaneous_multithreading_manually(self) -> bool:
         """Return true if simultaneous multithreading must be disabled with a cookbook script."""
-        return (
-            self.disable_simultaneous_multithreading
-            and self.instance_type_info.default_threads_per_core() > 1
-            and not self.instance_type_info.is_cpu_options_supported_in_lt()
-        )
+        return self.disable_simultaneous_multithreading and self.instance_type_info.default_threads_per_core() > 1
```
enrico-usai, line 1660: I just noticed we had a lot of common methods for `HeadNode`  and `BaseComputeResource`. 
We should have them in a common parent class (we can think to this separately)

cli/src/pcluster/resources/compute_node/user_data.sh
francesco-giordano: From the tests I have seen that you checked with Intel CPU. Do you check this with AMD?
It is not possible to get here with `DisableHyperThreadingManually=true` and ARM, am I recalling right?

CHANGELOG.md
lukeseawalker: changed

cli/src/pcluster/resources/compute_node/user_data.sh
lukeseawalker: comment added

cli/src/pcluster/config/cluster_config.py
lukeseawalker: I will leave this for a separate refactoring

cli/src/pcluster/resources/compute_node/user_data.sh
lukeseawalker: * AMD case it's covered as well, e.g. c5a.xlarge
```
# cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list
0,2
1,3
0,2
1,3
```

* ARM case, `disable_simultaneous_multithreading_manually` is false because this condition `self.instance_type_info.default_threads_per_core() > 1` is not verified, nevertheless code still works because
output of the cat will be empty

enrico-usai (approved):

francesco-giordano (approved):
LGTM

PR #4291, opened 2022-08-23
[integ-tests-2.11.7]Temporarily disabling build-image tests in us-gov region

demartinofra:
### Description of changes
* Temporarily disabling build-image tests in us-gov region due to DLAMI not available in GovCloud


chenwany (approved):

chenwany:
Do we want to remove it or just comment it out so we don't forget to re-enable it after it becomes available?

PR #4303, opened 2022-08-30
[API] Encrypt the EcrImageBuilder SNSTopic using the default AWS SNS KeyMasterKey

EddyMM:
### Description of changes
- The EcrImageBuilderSNSTopic is used to notify the UpdateParallelCluster Lambda function about docker image builder events
- The body of the SNS topic message needs to be encrypted using SSE (https://docs.aws.amazon.com/sns/latest/dg/sns-server-side-encryption.html)

### Tests
* Manually created the ParallelCluster API with the Encrypted SNS Topic

### References
* [sns-server-side-encryption](https://docs.aws.amazon.com/sns/latest/dg/sns-server-side-encryption.html#:~:text=AWS%20KMS%20for%20Amazon%20SNS%20is%20always%20alias/aws/sns)

### Checklist


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
 **ENHANCEMENTS**
 - Add new configuration parameter `DeletionPolicy` for EFs and FSx for Lustre shared storage 
   to support storage retention on deletion.
+- Encrypt the `EcrImageBuilderSNSTopic` used to communicate docker image build events using Server-Side Encryption.
```
demartinofra: nit: be explicit that this change is in the API infrastructure template. e.g. Enable server side encryption for the SNS topic created when deploying ParallelCluster API and used to communicate docker image build events.

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -1191,7 +1192,7 @@ Resources:
                 Action:
                   - logs:CreateLogStream
                   - logs:PutLogEvents
-                Resource: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/ParallelClusterApi-UpdateParallelClusterLambda-*
+                Resource: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/*
```
demartinofra: This unfortunately grants the Lambda permissions to write in any log group of any lambda function. One way to address this is to follow what we do in our cluster template, where we force the name of the lambda to be some prefix + a hash we get from the sack id. This way with a unique name we can scope down logs related permissions, given the fact that Lambda log groups names are based on the function name.

Also can you check if the same fix is required for the other Lambda functions present in this template?

EddyMM: > Also can you check if the same fix is required for the other Lambda functions present in this template?

I'll look into this.

demartinofra (approved):

PR #4304, opened 2022-08-30
[CLI] Include validation checks for Compute Resources with multiple instance types

EddyMM:
### Description of changes
- EFA Requirements:
    - Instance types should have the same EFA support status if EFA is enabled in the Compute Resource
    - If EFA is not enabled and one or more instance types support EFA, print a warning validation message
- Networking Requirements:
    - The lowest value for the MaximumNetworkInterfaceCards among the Instance Types is used. The user needs to be warned about this selection.
    - Compute Resources with multiple instance types with a PlacementGroup increases chances of an Insufficient Capacity Error. The user needs to be warned about this.
- CPU requirements:
    - Instance types should have the same number of CPUs or same number of Cores if Simultaneous Multithreading is disabled
- Accelerator requirements:
    - Instance types should have the same number of accelerators. The accelerators should also have the same manufacturer

### Tests
* Automated Tests to check validation messages for:
    * Instance Types should have the same number of CPUs
    * InstanceTypes should have the same number of cores if simultaneous multithreading is disabled
    * Instance Types should have the same number of Accelerators
    * Instance Types should have the same GPU manufacturer
    * Instance Types should have the same Accelerator Name (Inferentia)
    * If EFA is NOT enabled and one or more instance types supports EFA, a WARNING message should be printed
    * Instance Types should have the same EFA support status if EFA is enabled
    * Instance Types with varying Maximum NICs will have the smallest one used when setting the launch template
    * Using a placement group while having compute resources with multiple instance types increases the chances of getting an Insufficient Capacity Error

### References
* [describe-instance-types](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html)

### Checklist


cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -190,6 +191,62 @@ def test_schedulable_memory_validator(schedulable_memory, ec2memory, instance_ty
     assert_failure_messages(actual_failures, expected_message)
 
 
+# ---------------- Flexible Instance Types validators ---------------- #
+
+
+@pytest.mark.parametrize(
+    "queue_name, compute_resource_name, instance_types_info, disable_simultaneous_multithreading, efa_enabled, "
+    "placement_group_enabled, expected_message",
+    [
```
lukeseawalker: add a test case where there is no failure, for both disable_simultaneous_multithreading choices

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -183,6 +183,36 @@ def gpu_count(self):
...
 
+    def gpu_manufacturers(self):
+        """Return the list of GPU manufacturers supported by this instance type."""
+        gpu_info = self.instance_type_data.get("GpuInfo", None)
+
+        if gpu_info:
+            return list({gpu.get("Manufacturer") for gpu in gpu_info.get("Gpus", [])})
+        return []
+
+    def inference_accelerator_names(self):
```
lukeseawalker: `inference_accelerator_manufacturer` without the `s`. I do not expect an instance type have more than 1 inference accelerator of different manufacturers.

Thus, this should return a single value, not a list.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -183,6 +183,36 @@ def gpu_count(self):
...
+        """Return the list of Inference Accelerator Manufacturers supported by this instance type."""
+        inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", None)
+
+        if inference_accelerator_info:
+            return list({accelerator.get("Name") for accelerator in inference_accelerator_info.get("Accelerators", [])})
```
lukeseawalker: `accelerator.get("Name")` -> `accelerator.get("Manufacturer")`

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -183,6 +183,36 @@ def gpu_count(self):
 
         return gpu_count
 
+    def gpu_manufacturers(self):
```
lukeseawalker: `gpu_manufacturer` without the `s`. I do not expect an instance type have more than 1 gpu of different manufacturers.

Thus, this should return a single value, not a list.

Can we move the warning "ParallelCluster currently does not offer native support for '%s' GPUs.  etc..." in this method. I think it's more appropriate having the check when checking the manufacturer rather when counting the GPUs

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -183,6 +183,36 @@ def gpu_count(self):
...
+        inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", None)
+
+        accelerator_count = 0
+        if inference_accelerator_info:
+            for accelerator in inference_accelerator_info.get("Accelerators", []):
+                accelerator_name = accelerator.get("Name", "")
+                if accelerator_name.lower() == "inferentia":
+                    accelerator_count += accelerator.get("Count", 0)
+                else:
+                    LOGGER.warning("ParallelCluster currently does not support %s accelerators.", accelerator_name)
```
lukeseawalker: I would rather move this on the `inference_accelerator_manufacturer` method. Then we can align the messaging similar to the one for the GPU case, i.e. "ParallelCluster currently does not offer native support for ..."

This should be a FailureLevel.WARNING, otherwise it only appears in the log

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -120,6 +121,67 @@ def validate_cpu_requirements(
...
+            f"Inference Accelerators.",
+            FailureLevel.ERROR,
+        )
+
+    def validate_accelerator_manufactures(
+        self, compute_resource_name: str, instance_types_info: Dict[str, InstanceTypeInfo]
+    ):
+        """Instance Types should have the same manufacturer type."""
+        unique_gpu_manufacturers = set()
+        for instance_type_info in instance_types_info.values():
```
lukeseawalker: this will be simplified if we consider only one manufacturer per instance type

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -120,6 +121,67 @@ def validate_cpu_requirements(
                 FailureLevel.ERROR,
             )
 
+    def validate_accelerator_count(self, compute_resource_name: str, instance_types_info: Dict[str, InstanceTypeInfo]):
+        """Instance Types should have the same number of accelerators."""
+        self.validate_size(
+            {instance_type_info.gpu_count() for instance_type_info in instance_types_info.values()},
+            1,
+            f"Instance types listed under Compute Resource {compute_resource_name} must have the same number of GPUs.",
```
lukeseawalker: Can we print the different values we have found for the count?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -120,6 +121,67 @@ def validate_cpu_requirements(
...
+            FailureLevel.ERROR,
+        )
+
+        self.validate_size(
+            {instance_type_info.inference_accelerator_count() for instance_type_info in instance_types_info.values()},
+            1,
+            f"Instance types listed under Compute Resource {compute_resource_name} must have the same number of "
```
lukeseawalker: Can we print the different values we have found for the count?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -120,6 +121,67 @@ def validate_cpu_requirements(
...
+        """Instance Types should have the same manufacturer type."""
+        unique_gpu_manufacturers = set()
+        for instance_type_info in instance_types_info.values():
+            unique_gpu_manufacturers.update(instance_type_info.gpu_manufacturers())
+
+        self.validate_size(
+            unique_gpu_manufacturers,
+            1,
+            (
+                f"Instance types listed under Compute Resource {compute_resource_name} must have the same GPU "
```
lukeseawalker: Can we print the different values we have found for the manufacturers?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -120,6 +121,67 @@ def validate_cpu_requirements(
...
+
+        unique_accelerator_names = set()
+        for instance_type_info in instance_types_info.values():
+            unique_accelerator_names.update(instance_type_info.inference_accelerator_names())
+
+        self.validate_size(
+            unique_accelerator_names,
+            1,
+            (
+                f"Instance types listed under Compute Resource {compute_resource_name} must have the same inference "
```
lukeseawalker: Can we print the different values we have found for the manufacturers?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -225,6 +225,138 @@ def test_schedulable_memory_validator(schedulable_memory, ec2memory, instance_ty
             "Instance types listed under Compute Resource TestComputeResource must have the same number of CPU "
             "cores when Simultaneous Multithreading is disabled.",
         ),
+        # Instance Types should have the same number of GPUs
```
lukeseawalker: Provide all the tests where there is no failure, e.g. same count, same manufacturer are respected

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -73,6 +73,54 @@
...
+        disable_simultaneous_multithreading: bool,
+    ):
+        """Confirm CPU requirements for Flexible Instance Types.
+
+        Instance types should have the same number of CPUs or same number of Cores if Simultaneous Multithreading
+        is disabled.
+        """
+        if disable_simultaneous_multithreading:
+            self.validate_size(
+                {instance_type_info.cores_count() for instance_type_info in instance_types_info.values()},
```
lukeseawalker: code it's not optimal since the check can fail as soon as there is a mismatch, without having to traverse all the list. But deviation isn't much if the list isn't big. Let's follow up on this

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -253,7 +252,7 @@ def supported_architecture(self):
 
     def is_efa_supported(self):
         """Check whether EFA is supported."""
-        return self.instance_type_data.get("NetworkInfo").get("EfaSupported")
+        return self.instance_type_data.get("NetworkInfo", {}).get("EfaSupported")
```
lukeseawalker: what about returning a default for "EfaSupported"? i.e. `.get("EfaSupported", false)`

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -182,6 +183,63 @@ def validate_accelerator_requirements(
...
+            all_instance_types = set(instance_types_info.keys())
+            instance_types_without_efa_support = {
+                instance_type_name
+                for instance_type_name, instance_type_info in instance_types_info.items()
+                if not instance_type_info.is_efa_supported()
+            }
+
+            # If all the instance types have EFA support, `instance_types_without_efa_support` should be empty
+            # --> No failure expected
+            # If all the instance types DO NOT have EFA support, `instance_types_without_efa_support` should be the same
```
lukeseawalker: > If all the instance types DO NOT have EFA support, `instance_types_without_efa_support` should be the same as `all_instance_types` (Set difference of 0) --> No failure expected

I think this should be a failure because EFA is enabled on LT and NO instance will be launched, since none of them support EFA

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -182,6 +183,63 @@ def validate_accelerator_requirements(
...
+            # as `all_instance_types` (Set difference of 0) --> No failure expected
+            # If there is a mixed support for EFA, `instance_types_without_efa_support` & `all_instance_types`
+            # will have different instance types (Set difference greater than 0) --> Validation Failure expected and
+            # a message with the instance types that DO NOT support EFA is included
+            if instance_types_without_efa_support and len(all_instance_types - instance_types_without_efa_support) > 0:
+                self._add_failure(
+                    (
+                        "Instance types ({0}) in Compute Resource {1} do not support EFA.".format(
```
lukeseawalker: `types` -> `type(s)`

lukeseawalker: what about adding the reason of the failure here? Something like "and cannot be launched because you have enabled EFA [in CR configuration]"

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -357,6 +357,90 @@ def test_schedulable_memory_validator(schedulable_memory, ec2memory, instance_ty
             "Instance types listed under Compute Resource TestComputeResource must have the same inference "
             "accelerator manufacturer",
         ),
+        # Instance Types should have the same EFA support status if EFA is enabled
```
lukeseawalker: add the test cases when there is no failure (e.g EFA enabled and all instances support EFA, EFA disabled and all instances don't support EFA)

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -441,6 +441,43 @@ def test_schedulable_memory_validator(schedulable_memory, ec2memory, instance_ty
             "update the cluster's configuration to enable EFA ("
             "https://docs.aws.amazon.com/parallelcluster/latest/ug/efa-v3.html).",
         ),
+        # Instance Types with varying Maximum NICs will have the smallest one used when setting the launch template
```
lukeseawalker: add the test cases for when there is no WARNING

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -278,6 +281,18 @@ def validate_networking_requirements(
                 FailureLevel.WARNING,
             )
 
+    def validate_allocation_strategy(self, compute_resource_name: str, capacity_type: Enum, allocation_strategy: Enum):
+        """On-demand Capacity type only supports "lowest-price" allocation strategy."""
+        if (
+            capacity_type == cluster_config.CapacityType.ONDEMAND
+            and allocation_strategy == cluster_config.AllocationStrategy.CAPACITY_OPTIMIZED
```
lukeseawalker: change to `allocation_strategy` != `LOWEST_PRICE` so that if a new allocation strategy is added and not supported by ONDEMAND, the check will be still valid

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -278,6 +281,18 @@ def validate_networking_requirements(
...
+        ):
+            self._add_failure(
+                f"Compute Resource {compute_resource_name} is using an OnDemand CapacityType. OnDemand CapacityType "
+                "can only use 'lowest-price' allocation strategy.",
```
lukeseawalker: use the enum instead of the 'lowest-price' string, so that if value changes there is only one place where it has to be changed and message will be still valid

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -478,10 +500,26 @@ def test_schedulable_memory_validator(schedulable_memory, ec2memory, instance_ty
             "use of multiple instance types for Compute Resource: TestComputeResource ("
             "https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-groups.html#placement-groups-cluster).",
         ),
+        # OnDemand Capacity type only supports "lowest-price" allocation strategy
```
lukeseawalker: add test cases when there is no failure

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -73,6 +73,246 @@
...
+        capacity_type: Enum,
+        allocation_strategy: Enum,
+        compute_resource_name: str,
+        instance_types_info: Dict[str, InstanceTypeInfo],
+        disable_simultaneous_multithreading: bool,
+        efa_enabled: bool,
+        placement_group_enabled: bool,
+        memory_scheduling_enabled: bool,
+    ):
+        self.validate_cpu_requirements(compute_resource_name, instance_types_info, disable_simultaneous_multithreading)
```
lukeseawalker: I'm worried about putting all the checks in a single validator. This because if user wants to disable validator to bypass cpu check, then all the other checks will be bypassed as well. I'm rather on the opinion that the check needs to be disabled on demand, one by one. Let's follow up on this.

cli/src/pcluster/aws/aws_resources.py
EddyMM: > Thus, this should return a single value, not a list

👍, Just did a quick filter on EC2 console, all the instance types seem to return a single manufacturer.

> it's more appropriate having the check when checking the manufacturer

+1

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
 
+    def gpu_manufacturer(self) -> str:
+        """Return the list of GPU manufacturers supported by this instance type."""
+        gpu_info = self.instance_type_data.get("GpuInfo", None)
+
+        gpu_manufacturers = list({gpu.get("Manufacturer", "") for gpu in gpu_info.get("Gpus", [])}) if gpu_info else []
+
+        # Only one GPU manufacturer is associated with each Instance Type's GPU
+        manufacturer = gpu_manufacturers[0] if gpu_manufacturers else ""
+        if manufacturer.upper() != "NVIDIA":
```
enrico-usai, line 187: This warning message will be printed even if manufacturer is empty. Is this expected?

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
+                "ParallelCluster currently does not offer native support for '%s' GPUs. "
+                "Please make sure to use a custom AMI with the appropriate drivers in order to leverage "
+                "GPUs functionalities",
+                manufacturer,
+            )
+        return manufacturer
+
+    def inference_accelerator_manufacturer(self) -> str:
+        """Return the list of Inference Accelerator Manufacturers supported by this instance type."""
+        inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", None)
```
enrico-usai: to save an if/else you could do:
```python
inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", {})
...
inference_accelerator_manufacturers = (
            list(
                {
                    accelerator.get("Manufacturer", "")
                    for accelerator in inference_accelerator_info.get("Accelerators", [])
                }
            )
        )
```

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
+                if manufacturer.upper() == "NVIDIA":
+                    gpu_count += gpu.get("Count", 0)
 
         return gpu_count
 
+    def gpu_manufacturer(self) -> str:
+        """Return the list of GPU manufacturers supported by this instance type."""
+        gpu_info = self.instance_type_data.get("GpuInfo", None)
+
+        gpu_manufacturers = list({gpu.get("Manufacturer", "") for gpu in gpu_info.get("Gpus", [])}) if gpu_info else []
```
enrico-usai: to save an if/else you can do:
```python
gpu_info = self.instance_type_data.get("GpuInfo", {})

gpu_manufacturers = list({gpu.get("Manufacturer", "") for gpu in gpu_info.get("Gpus", [])})
```

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
+                    for accelerator in inference_accelerator_info.get("Accelerators", [])
+                }
+            )
+            if inference_accelerator_info
+            else []
+        )
+        # Only one accelerator manufacturer is associated with each Instance Type's accelerator
+
+        accelerator_manufacturer = inference_accelerator_manufacturers[0] if inference_accelerator_manufacturers else ""
+        if accelerator_manufacturer.upper() != "AWS":
```
enrico-usai, line 208: same here, we'll print a warning with an empty accelerator_manufacturer, is this expected?

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
+            LOGGER.warning(
+                "ParallelCluster currently does not offer native support for '%s' GPUs. "
+                "Please make sure to use a custom AMI with the appropriate drivers in order to leverage "
+                "GPUs functionalities",
+                manufacturer,
+            )
+        return manufacturer
+
+    def inference_accelerator_manufacturer(self) -> str:
+        """Return the list of Inference Accelerator Manufacturers supported by this instance type."""
```
enrico-usai: wrong comment: this is not returning a list.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
-                    )
+            for gpu in gpu_info.get("Gpus", []):
+                manufacturer = gpu.get("Manufacturer", "")
+                if manufacturer.upper() == "NVIDIA":
+                    gpu_count += gpu.get("Count", 0)
 
         return gpu_count
 
+    def gpu_manufacturer(self) -> str:
+        """Return the list of GPU manufacturers supported by this instance type."""
```
enrico-usai: wrong comment: this is not returning a list.

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -164,28 +164,85 @@ def __init__(self, instance_type_data):
...
+                "Please make sure to use a custom AMI with the appropriate drivers in order to leverage the "
+                "accelerators functionalities",
+                accelerator_manufacturer,
+            )
+
+        return inference_accelerator_manufacturers[0] if inference_accelerator_manufacturers else ""
+
+    def inference_accelerator_count(self):
+        """Return the total number of Inference Accelerators associated with this instance type."""
+        inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", None)
```
enrico-usai: same here, you can save the if by doing:
```python
inference_accelerator_info = self.instance_type_data.get("InferenceAcceleratorInfo", {})
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1806,21 +1812,32 @@ def queue_ami(self):
...
+        allocation_strategy: str = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
         self.compute_resources = compute_resources
         self.networking = networking
         if any(isinstance(compute_resource, SlurmFlexibleComputeResource) for compute_resource in compute_resources):
-            self.allocation_strategy = allocation_strategy
+            self.allocation_strategy = (
+                AllocationStrategy[allocation_strategy.replace("-", "_").upper()]
```
enrico-usai: you can use `to_camel_case` utility function in `utils.py`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2462,8 +2479,32 @@ def _register_validators(self):
...
+                        instance_types_info=compute_resource.instance_type_info_map,
+                        disable_simultaneous_multithreading=compute_resource.disable_simultaneous_multithreading,
+                        efa_enabled=compute_resource.efa and compute_resource.efa.enabled,
+                        placement_group_enabled=(
+                            queue.networking.placement_group and queue.networking.placement_group.enabled
+                        ),
+                        memory_scheduling_enabled=self.scheduling.settings.enable_memory_based_scheduling,
+                    )
+                    flexible_instance_types_validators = [
+                        InstanceTypesListCPUValidator,
```
enrico-usai: we should rename all of them to be: `InstanceTypeList`, rather than `InstanceTypesList` to be consistent with the configuration parameter name.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1151,3 +1151,258 @@ def _validate(self, scheduler):
...
+        **kwargs,
+    ):
+        """Check if CPU requirements are met.
+
+        Instance types should have the same number of CPUs or same number of Cores if Simultaneous Multithreading
+        is disabled.
+        """
+        if disable_simultaneous_multithreading:
+            self.validate_property_homogeneity(
+                instance_type_info_list=list(instance_types_info.values()),
```
enrico-usai: to avoid repetition and simplify the code you could:
1. accept `instance_types_info` in  `validate_property_homogeneity` and make the get of values and conversion to a list within the `validate_property_homogeneity` or alternatively you can define a method that does it in the `_FlexibleInstanceTypesValidatorMixin` and just call it every time you need to perform the conversion
1. pass just a piece of the message and add the initial part that is it in common  within the `validate_property_homogeneity` (I mean: `Instance types listed under Compute Resource`) 
1. avoid to pass `failure_level given` that it's always set to ERROR and just set it internally
1. (optional) instead of passing a callback function for the `property callback` you could simply pass the method/property name (i.e. `"vcpus_count"`, `"cores_count"`, ...).

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1151,3 +1151,258 @@ def _validate(self, scheduler):
...
+                property_callback=lambda instance_type_info: instance_type_info.vcpus_count(),
+                failure_message_callback=lambda heterogeneous_values: f"Instance types listed under Compute Resource "
+                f"{compute_resource_name} must have the same number of vCPUs ({heterogeneous_values}).",
+                failure_level=FailureLevel.ERROR,
+            )
+
+
+class InstanceTypesListAcceleratorsValidator(Validator, _FlexibleInstanceTypesValidatorMixin):
```
enrico-usai: InstanceTypeList (without the s), everywhere.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1151,3 +1151,258 @@ def _validate(self, scheduler):
...
+            # --> No failure expected
+            # If all the instance types DO NOT have EFA support, `instance_types_without_efa_support` should be the same
+            # as `all_instance_types` (Set difference of 0) --> No failure expected
+            # If there is a mixed support for EFA, `instance_types_without_efa_support` & `all_instance_types`
+            # will have different instance types (Set difference greater than 0) --> Validation Failure expected and
+            # a message with the instance types that DO NOT support EFA is included
+            if instance_types_without_efa_support and len(all_instance_types - instance_types_without_efa_support) > 0:
+                self._add_failure(
+                    (
+                        "Instance type(s) ({0}) in Compute Resource {1} do not support EFA and cannot be launched "
```
enrico-usai: What about: 
`"Instance type(s) ({0}) do not support EFA and cannot be launched when EFA is enabled in Compute Resource: {1}."`

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+                    }
+                ),
+                "g5.xlarge": InstanceTypeInfo(
+                    {
+                        "InstanceType": "g5.xlarge",
+                        "VCpuInfo": {"DefaultVCpus": 4, "DefaultCores": 2},
+                        "GpuInfo": {
+                            "Gpus": [
+                                {
+                                    "Name": "A10G",
```
enrico-usai: To simplify test code you could avoid to define items in dictionary that are not used by the validator. (maybe it's already so).

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+                                    "Count": 2,
+                                    "MemoryInfo": {"SizeInMiB": 24576},
+                                }
+                            ],
+                            "TotalGpuMemoryInMiB": 24576,
+                        },
+                    }
+                ),
+            },
+            "Instance types listed under Compute Resource TestComputeResource must have the same GPU manufacturer ({"
```
enrico-usai: Up to you: You could add `.*` in place of the common part of the message `Instance types listed under Compute Resource`

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
     mock_aws_api(mocker)
     actual_failures = RootVolumeSizeValidator().execute(root_volume_size, ami_size)
     assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
```
enrico-usai: I think it's worth to move your validator into a separate module. 
So move them from `cluster_validators` to `instance_type_list_validators` and do the same for the tests: from `test_cluster_validators` to `test_instance_type_list_validators`.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+    "compute_resource_name, instance_types_info, efa_enabled, expected_message",
+    [
+        # Instance Types should have the same EFA support status if EFA is enabled
+        (
+            "TestComputeResource",
+            {
+                "t2.micro": InstanceTypeInfo(
+                    {
+                        "VCpuInfo": {"DefaultVCpus": 2, "DefaultCores": 2},
```
enrico-usai: For example, do we really need vcpu info here to test this validator?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+                "c5n.9xlarge": InstanceTypeInfo(
+                    {
+                        "VCpuInfo": {"DefaultVCpus": 2, "DefaultCores": 2},
+                        "NetworkInfo": {"EfaSupported": False},
+                    }
+                ),
+                "c5n.18xlarge": InstanceTypeInfo(
+                    {
+                        "VCpuInfo": {"DefaultVCpus": 2, "DefaultCores": 2},
+                        "NetworkInfo": {"EfaSupported": False},
```
enrico-usai: I'd use t3.micro and t2.micro in this example, to be consistent with EfaSupported False value.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+            {
+                "t2.micro": InstanceTypeInfo(
+                    {
+                        "VCpuInfo": {"DefaultVCpus": 4, "DefaultCores": 2},
+                        "NetworkInfo": {"MaximumNetworkCards": 4},
+                    }
+                ),
+                "t3.micro": InstanceTypeInfo(
+                    {
+                        "VCpuInfo": {"DefaultVCpus": 4, "DefaultCores": 2},
```
enrico-usai: same here, do we really need vcpu info to test this validator?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1162,3 +1171,568 @@ def test_root_volume_size_validator(mocker, root_volume_size, ami_size, expected
...
+    actual_failures = InstanceTypesListAllocationStrategyValidator().execute(
+        compute_resource_name, capacity_type, allocation_strategy
+    )
+    assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
+    "compute_resource_name, instance_types_info, memory_scheduling_enabled, expected_message",
+    [
+        # Memory-based scheduling is supported for Compute Resource that use either 'InstanceType' or have a sinlge
```
enrico-usai: typo: single

cli/src/pcluster/aws/aws_resources.py
lukeseawalker: I would say yes, since there is an accelerator data but the manufacturer is missing

cli/src/pcluster/config/cluster_config.py
```diff
@@ -82,6 +82,7 @@
     HeadNodeLaunchTemplateValidator,
     HostedZoneValidator,
     InstanceArchitectureCompatibilityValidator,
+    InstanceTypesListCPUValidator,
```
lukeseawalker: InstanceTypesListCPUValidator -> InstanceTypeListCPUValidator

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1272,3 +1254,76 @@ def _validate(
...
+            all_instance_types = set(instance_types_info.keys())
+            instance_types_without_efa_support = {
+                instance_type_name
+                for instance_type_name, instance_type_info in instance_types_info.items()
+                if not instance_type_info.is_efa_supported()
+            }
+
+            # If all the instance types have EFA support, `instance_types_without_efa_support` should be empty
+            # --> No failure expected
+            # If all the instance types DO NOT have EFA support, `instance_types_without_efa_support` should be the same
```
lukeseawalker: > If all the instance types DO NOT have EFA support, instance_types_without_efa_support should be the same as all_instance_types (Set difference of 0) --> No failure expected

Please fix this: this should be a failure because EFA is enabled on LT and NO instance will be launched, since none of them support EFA

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1272,3 +1254,76 @@ def _validate(
...
+                    (
+                        "Instance type(s) ({0}) in Compute Resource {1} do not support EFA and cannot be launched "
+                        "when EFA is enabled in Compute Resource: {1}.".format(
+                            ",".join(sorted(instance_types_without_efa_support)),
+                            compute_resource_name,
+                        )
+                    ),
+                    FailureLevel.ERROR,
+                )
+            if len(all_instance_types - instance_types_without_efa_support) == 0:
```
lukeseawalker: I believe the two if can be simplified into just one `instance_types_without_efa_support > 0`.

cli/src/pcluster/aws/aws_resources.py
EddyMM: Good point. I think logging that only 'AWS' manufacturer is supported, then show the `accelerator_info`.
This will be enough to communicate why the accelerator wasn't considered.

cli/src/pcluster/config/cluster_config.py
EddyMM: Thanks, I think `to_snake_case` is the right utility function.

cli/tests/pcluster/validators/test_cluster_validators.py
EddyMM: The `vCPUInfo` can be removed for some of the validators.

cli/tests/pcluster/validators/test_cluster_validators.py
EddyMM: Thanks, I'll use `.*` for messages in the logs. For the validator messages I'd prefer matching the exact text since they're customer facing.

cli/src/pcluster/validators/cluster_validators.py
EddyMM: 1. 👍
2. I'm giving the caller of `validate_property_homogeneity` flexibility to use the mismatched values for now, since the wording hasn't been decided on yet.
3. 👍  I can set `ERROR` as the default argument for the `failure_level` parameter and avoid passing it explicitly.
4. I chose a callback to give the caller the flexibility to process the `instance_type_info` object

cli/src/pcluster/validators/cluster_validators.py
EddyMM: I agree, we can use the same messaging as well.

enrico-usai (approved):
Great job!

PR #4305, opened 2022-08-30
3.x Change CUDA install validation test to support both v11.4 and v11.x > 11.4 for ARM

davprat:
### Description of changes
This change will use the correct path to build and execute the deviceQuery utility during image validation on ARM architectures depending on which version of the CUDA toolkit is installed.

Starting with NVIDIA CUDA Toolkit v11.5, the samples where the deviceQuery source code is installed was moved out of the CUDA install package and into its own GitHub repository. The structure of the sample directy tree also changed thus requiring different paths for compiling and executing the deviceQuery utility.

### Tests
* Ran developer build with CUDA 11.7 installed, validation test passed.
* Ran developer build with CUDA 11.4 installed, validation test passed.

### References
* [NVIDIA CUDA 11.7.1 EULA](https://docs.nvidia.com/cuda/eula/index.html#nvidia-cuda-samples-preface)
* [NVIDIA CUDA Samples](https://github.com/NVIDIA/cuda-samples)

### Checklist


cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
```diff
@@ -167,6 +167,36 @@ phases:
...
 
+      - name: CudaSamplesSrcDir
+        action: ExecuteBash
+        inputs:
+          commands:
+            - |
+              set -v
+              cuda_ver="{{ validate.CudaVersion.outputs.stdout }}"
+              if [ ${cuda_ver} \> '11.4' ]; then
+                PATTERN=$(grep -F "default['cluster']['nvidia']['cuda_sample_version']" {{ CookbookDefaultFile }})
```
lukeseawalker, line 178: I couldn't find `default['cluster']['nvidia']['cuda_sample_version']`.
Is this related to another PR? Can you please reference it into the description?

lukeseawalker (commented):
> Starting with NVIDIA CUDA Toolkit v11.5, the samples where the deviceQuery source code is installed was moved out of the CUDA install package and into its own GitHub repository.

Does it mean that the samples are not in the CUDA bundle we use? If yes, when/where we are downloading this samples?

cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
```diff
@@ -167,6 +167,36 @@ phases:
...
               echo ${VERSION}
 
+      - name: CudaSamplesSrcDir
+        action: ExecuteBash
+        inputs:
+          commands:
+            - |
+              set -v
+              cuda_ver="{{ validate.CudaVersion.outputs.stdout }}"
+              if [ ${cuda_ver} \> '11.4' ]; then
```
lukeseawalker, line 177: Mind that this is a string comparison, and `cuda_ver=011.5` will fail the test. It's OK if we are confident about the versioning format

davprat: This will be retrieved from 'attributes/default.rb' in the cookbook recipes, so it is under our control. I can add some more processing to make sure it's normalized though if you like?

davprat: So the sample download will come from github and is done as part of my yet to be generated pull-request for the cookbook package. I'm just trying to set up the test here so that it will work with either the embedded samples in 11.4 or the external samples in 11.7 - or at least that is my goal. Let me know if there is a better way to do this.

cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
davprat: That change is coming in an as yet to be generated PR for parallelcluster-cookbook. So technically, with this PR, cuda_ver will be empty, so the logic here still kind of works but is definitely sloppy. I will fix this.

cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
lukeseawalker: If the value is under our control I'm good with this, tnx

lukeseawalker (approved):

PR #4330, opened 2022-09-09
3.3 Add the Networking section to SlurmQueues.ComputeResources including only PlacementGroup

dreambeyondorange:
This change allows for creating configs that include placement groups for specific compute resources.  It changes the way Queue level placement groups are configured as well.  This aligns better with AWS EC2 placement group guidance to reduce insufficient capacity errors.  Parallel cluster managed placement groups will no longer be created to include multiple compute resources.

If enabled at the queue level and not assigned a name, every compute resource will get a separate parallel cluster managed placement group.

If enabled and assigned a name at the queue level, there will be a single placement group for all compute resources in the queue.

If enabled at the compute resource level and not assigned a name, this config will override the queue level config and that resource will get its own managed placement group.

If enabled at the compute resource level and assigned a name, this config will override the queue level config and that resource will use the assigned placement group.

Signed-off-by: Ryan Anderson <ndry@amazon.com>

### Description of changes
* See above

### Tests
* Unit tests for logic in cluster config
* Unit tests updated for validators

### References
* https://quip-amazon.com/kaLSAnnbyR7I/Design-Add-Support-for-On-Demand-Capacity-Reservations-ODCRs-in-Cluster-Configuration-File

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -518,8 +518,8 @@ class Proxy(Resource):
 class _BaseNetworking(Resource):
     """Represent the networking configuration shared by head node and compute node."""
 
-    def __init__(self, security_groups: List[str] = None, additional_security_groups: List[str] = None):
-        super().__init__()
+    def __init__(self, security_groups: List[str] = None, additional_security_groups: List[str] = None, **kwargs):
+        super().__init__(**kwargs)
         self.security_groups = Resource.init_param(security_groups)
         self.additional_security_groups = Resource.init_param(additional_security_groups)
 
@@ -548,16 +548,34 @@ class HeadNodeNetworking(_BaseNetworking):
 
 
 class PlacementGroup(Resource):
-    """Represent the placement group for the Queue networking."""
+    """Represent the placement group for networking."""
 
-    def __init__(self, enabled: bool = None, name: str = None, id: str = None):
-        super().__init__()
-        self.enabled = Resource.init_param(enabled, default=False)
+    def __init__(self, enabled: bool = None, name: str = None, id: str = None, **kwargs):
+        super().__init__(**kwargs)
+        self.enabled = Resource.init_param(enabled)
         self.name = Resource.init_param(name)
         self.id = Resource.init_param(id)  # Duplicate of name
 
     def _register_validators(self):
         self._register_validator(PlacementGroupNamingValidator, placement_group=self)
+
+    @property
+    def is_enabled_and_unassigned(self) -> bool:
+        """Check if the PlacementGroup is enabled without a name or id."""
+        return not (self.id or self.name) and self.enabled
+
+    @property
+    def assignment(self) -> str:
+        """Check if the placement group has a name or id and get it, preferring the name if it exists."""
+        return self.name or self.id
+
+
+class SlurmComputeResourceNetworking(Resource):
+    """Represent the networking configuration for the compute resource."""
+
+    def __init__(self, placement_group: PlacementGroup = None, **kwargs):
+        super().__init__(**kwargs)
+        self.placement_group = placement_group or PlacementGroup(implied=True)
 
 
 class _QueueNetworking(_BaseNetworking):
@@ -574,7 +592,7 @@ class SlurmQueueNetworking(_QueueNetworking):
 
     def __init__(self, placement_group: PlacementGroup = None, proxy: Proxy = None, **kwargs):
         super().__init__(**kwargs)
-        self.placement_group = placement_group
+        self.placement_group = placement_group or PlacementGroup(implied=True)
         self.proxy = proxy
 
 
@@ -1623,6 +1641,7 @@ class _BaseSlurmComputeResource(BaseComputeResource):
         disable_simultaneous_multithreading: bool = None,
         schedulable_memory: int = None,
         capacity_reservation_target: CapacityReservationTarget = None,
+        networking: SlurmComputeResourceNetworking = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -1637,6 +1656,7 @@ class _BaseSlurmComputeResource(BaseComputeResource):
         self.capacity_reservation_target = capacity_reservation_target
         self._instance_types_with_instance_storage = []
         self._instance_type_info_map = {}
+        self.networking = networking or SlurmComputeResourceNetworking(implied=True)
 
     @staticmethod
     def fetch_instance_type_info(instance_type) -> InstanceTypeInfo:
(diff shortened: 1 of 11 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1364,47 +1364,38 @@ def _add_policies_to_cleanup_resources_lambda_role(self):
...
-                    queue_placement_group = managed_placement_groups[queue.name].ref
-
             queue_pre_install_action, queue_post_install_action = (None, None)
             if queue.custom_actions:
                 queue_pre_install_action = queue.custom_actions.on_node_start
                 queue_post_install_action = queue.custom_actions.on_node_configured
 
-            for compute_resource in queue.compute_resources:
-                launch_template = self._add_compute_resource_launch_template(
+            for resource in queue.compute_resources:
```
hanwen-pcluste, line 1390: Just curious: Why did you do the renaming here?

cli/src/pcluster/utils.py
```diff
@@ -333,3 +333,33 @@ def get_attr(obj, attributes, default=None):
...
+def is_fully_defined_placement_group(placement_group) -> bool:
+    return (placement_group.id or placement_group.name) and (
+        placement_group.enabled or placement_group.is_implied("enabled")
+    )
+
+
+def is_managed_placement_group(placement_group) -> bool:
+    return (placement_group.id or placement_group.name) and (
+        placement_group.enabled or placement_group.is_implied("enabled")
+    )
```
hanwen-pcluste: Can we move the two functions to two methods of PlacementGroup class in cluster_config.py?

cli/src/pcluster/utils.py
```diff
@@ -333,3 +333,33 @@ def get_attr(obj, attributes, default=None):
         if obj is None:
             return default
     return obj
+
+
+def get_placement_group_for_compute_resource(queue, managed_placement_groups, compute_resource):
```
hanwen-pcluste: Can we try to split this function in to some methods in `cluster_config.py` and `cluster_stack.py`

Below is a rough idea. The final implementation can be something different.

In `SlurmQueue` in `cluster_config.py`:
```
def is_fully_defined_placement_group(compute_resource_name) -> bool:
    ...

def get_fully_defined_placement_group(compute_resource_name) -> str:
    ...
```
Then, put other logic in `cluster_stack.py`

hanwen-pcluste (commented):
Great job! Thank you!

cli/src/pcluster/templates/cluster_stack.py
dreambeyondorange: I just didn't like the extra qualifier aha

cli/src/pcluster/utils.py
dreambeyondorange: Yup, that makes sense to move them.  However I realized that the managed group function also needs to check the queue level config... so may just get rid of that one

cli/src/pcluster/utils.py
dreambeyondorange: will do!

dreambeyondorange:
PR is ready for review @hanwen-pcluste

cli/src/pcluster/config/cluster_config.py
```diff
@@ -548,17 +548,40 @@ def availability_zone(self):
 
 
 class PlacementGroup(Resource):
-    """Represent the placement group for the Queue networking."""
+    """Represent the placement group for networking."""
 
-    def __init__(self, enabled: bool = None, name: str = None, id: str = None):
-        super().__init__()
-        self.enabled = Resource.init_param(enabled, default=False)
```
hanwen-pcluste, line 555: Let's double check with the team to make sure this change is safe

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1829,6 +1870,43 @@ def queue_ami(self):
...
+    def get_managed_placement_group_keys(self) -> List[str]:
+        managed_placement_group_keys = []
+        if self.networking.placement_group.is_enabled_and_unassigned:
+            for resource in self.compute_resources:
+                if resource.networking.placement_group.should_be_managed:
+                    managed_placement_group_keys.append(f"{self.name}-{resource.name}")
+        else:
+            for resource in self.compute_resources:
+                if resource.networking.placement_group.is_enabled_and_unassigned:
+                    managed_placement_group_keys.append(f"{self.name}-{resource.name}")
```
hanwen-pcluste: Can we simplify the code:
```
        for resource in self.compute_resources: 
            placement_group_setting = resource.networking.placement_group if not resource.placement_group.implied else self.networking.placement_group
            if placement_group_setting.is_enabled_and_unassigned:
                managed_placement_group_keys.append(f"{self.name}-{resource.name}")
```

hanwen-pcluste: If the simplification can be done, `should_be_managed` method can be removed too.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -315,24 +315,18 @@ def _validate(self, instance_type, efa_enabled, gdr_support):
...
-        placement_group_config_implicit = placement_group is None or (
-            placement_group.is_implied("enabled") and placement_group.id is None
-        )
-        if efa_enabled and placement_group_config_implicit:
+    def _validate(self, efa_enabled: bool, placement_group_key: str, placement_group_disabled: bool):
+        if efa_enabled and placement_group_disabled:
+            self._add_failure(
+                "You may see better performance using a placement group for the queue.", FailureLevel.WARNING
+            )
+        elif efa_enabled and placement_group_key is None:
```
hanwen-pcluste, line 323: A further improvement can be extracting `efa_enabled`
e.g.
```
if efa_enabled:
  if placement_group_disabled:
    ...
  elif placement_group_key is None:
    ...
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1829,6 +1870,43 @@ def queue_ami(self):
...
+        return managed_placement_group_keys
+
+    def get_placement_group_key_for_compute_resource(
```
hanwen-pcluste, line 1880: I am wondering whether we should split this method to
`get_placement_group_name_for_compute_resource` and `is_placement_group_managed_for_compute_resource`?
Because this method has some repetitions of `"{self.name}-{compute_resource.name}"`, which are not really used in the consumers of this method.

hanwen-pcluste (commented):
Great job (detailed description, thorough testing)! Thank you!

I left some comments!

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: the logic here is slightly more performant because it only checks if the queue level is enabled once instead of checking it for every compute resource, but I would be OK simplifying if that is more important.  Do you think that is the case here?

cli/src/pcluster/validators/cluster_validators.py
dreambeyondorange: That could be true, but it seems to complicate the code and add an extra layer of nesting, which I didn't like.

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: I wrote this function initially for the cluster_stack implementation, but I agree that splitting it might be cleaner for the other consumers to distinguish.

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: will simplify the logic here

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: will leave this for now

cli/src/pcluster/validators/cluster_validators.py
dreambeyondorange: will leave this for now

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: nothing from the team so far

hanwen-pcluste (approved):
LGTM

PR #4331, opened 2022-09-09
3.x Update integration test to use normalized directory name for CUDA rather than versioned directory name.

davprat:
### Description of changes
* Updated EFA integration test to refer to /usr/local/cuda linked directory rather than /usr/local/cuda-11.4
* This is so that the test will work with newer versions of CUDA.
* Additional change to upload_artifacts.sh shell script to support both BSD (MacOS) and Linux style `sed` options.
* `-i` requires an argument on Mac - changed so that integration tests could be run from developer Macs.

### Tests
* Ran EFA integration test from developer desktop on developer account.
* EFA integration test was modified to CUDA version agnostic.

### Checklist


scheduler_plugins/slurm/utils/upload_artifacts.sh
```diff
@@ -19,6 +19,11 @@ function usage {
     exit 1
 }
 
+BUILD_OS=$(uname)
+if [ "${BUILD_OS}" != "Linux" ]; then
+  SED_BAK_EXT="''"
```
gmarciani, line 24: Afaik the safest approach to reuse the same sed parameters both in Linux and Mac is to use `gsed` instead of `sed` on Mac.
Btw, no problem at all with this approach as far as it's working on our specific case. 
Thanks for proposing it: I did not know about it.

gmarciani (approved):

PR #4356, opened 2022-09-19
Add support for Slurm Accounting

jdeamicis:
### Description of changes
* Add basic cluster configuration parameters to enable Slurm Accounting.
* Add validators and unit tests for validators for Slurm Accounting.
* Propagate Slurm Accounting cluster parameters to the dna.json of the head node.
* Add GetSecretValue IAM permission in case Slurm database is enabled.

### Tests
* Added unit tests for the new feature and modified some existing ones (e.g. test_all_validators).
* Ran unit tests with the patched code.
* Manual creation of cluster with Slurm accounting.
* Manual update of cluster of cluster to enable/disable the functionality.
* Integration tests to be added at a later stage.

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1530

### Checklist
- [x] Make sure **to have added unit tests ~~or integration tests~~** to cover the new/modified code.


cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Check if netloc can be parsed
+        if not self._check_netloc(uri, uri_parse):
+            return
+
+        # Check if port is provided and acceptable
+        if not self._check_port(uri_parse):
+            return
+
+    def _check_trailing_slash(self, uri: str) -> bool:
```
EddyMM: The term **trailing** applies to the end of a string. I believe you meant `check_leading_slash`?

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,94 @@
...
+        if uri and uri[0] == "/":
```
EddyMM: [startswith](https://docs.python.org/3/library/stdtypes.html#str.startswith) can also work.

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,94 @@
...
+
+        uri_parse = urlparse(uri)
+        if not uri_parse.netloc:
+            # This happens if users provide an URI without explicit scheme followed by ://
+            # (for example 'test.example.com:3306' instead of 'mysql://test.example.com:3306`).
+            uri_parse = urlparse("//" + uri)
+
+        # Throw error if the URI contains a scheme
+        if not self._check_scheme(uri_parse):
+            return
```
EddyMM, line 39: For consistency in validator behaviour, should we avoid returning on the first encounter of an error?
I believe the current behaviour is to build a list of validation failures then print all of them.
Using this `return` approach will break that pattern.

cli/src/pcluster/validators/database_validators.py
jdeamicis: You are right. Fixing in the next commit.

cli/src/pcluster/validators/database_validators.py
jdeamicis: Fixing in the next commit.

cli/src/pcluster/validators/database_validators.py
jdeamicis: That's a very good question. When looking at the validators, most validation functions had a structure like `if ... elif ...`, meaning that they implicitly return only one type of validation error.

Some validators though show the pattern you are talking about (for example [`ExistingFsxNetworkingValidator`](https://github.com/aws/aws-parallelcluster/blob/874f95fac31731eae1b5156c5ca274c482d58cb4/cli/src/pcluster/validators/cluster_validators.py#L445)). But I believe there we are talking about looping through different entities to be validated (essentially one check per file system). Do you have any example about multiple validation errors for the same validated entity?

I'm not sure what the best user experience is. Should we tell the user all the criteria the string does not comply with, even if the string will most likely change to cope with the first validation error, potentially leading to a different outcome at the next iteration of the validation?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1329,6 +1330,26 @@ def make_resource(self, data, **kwargs):
...
+    uri = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    user_name = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    password_secret_arn = fields.Str(
+        required=True,
+        validate=validate.Regexp(r"^arn:.*:secret"),
+        metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP},
+    )
+    # We don't want to take ownership of the whole structure of the StorageParameters config
+    # parameter in slurmdbd.conf, so we define a generic dictionary rather than a subschema.
+    slurmdbd_storage_parameters = fields.Dict(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
```
demartinofra: Leaving a note so that we can discuss whether we want to leave this as is, extend to allow any slurmdbd param to ba passed or drop for now and expect users to modify slurmdbd.conf in this first iteration.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -122,6 +122,23 @@ def get_slurm_specific_dna_json_for_head_node(config: SlurmClusterConfig, schedu
     }
 
 
+def get_slurm_database_dna_json_for_head_node(config: BaseClusterConfig) -> dict:
```
demartinofra: why we need this to explicitly go into the dna.json and can't only rely on the config?

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,128 @@
...
+from urllib.parse import ParseResult, urlparse
+
+from pcluster.constants import SLURMDBD_STORAGE_PARAMETERS
+from pcluster.validators.common import FailureLevel, Validator
+
+
+class DatabaseUriValidator(Validator):
+    """Slurm database URI validator."""
+
+    def _validate(self, uri: str):
```
demartinofra, line 20: if you want to avoid all these "if not -> return" you could put in and all the function calls, the first returning False should stop the flow.

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,128 @@
...
+        default_mysql_port = 3306
+        try:
+            port = uri_parse.port
+        except ValueError as e:
+            self._add_failure("Invalid URI specified. " + str(e), FailureLevel.ERROR)
+            return False
+        if not port:
+            self._add_failure(
+                f"No port specified in the URI. Assuming the use of port {default_mysql_port}",
+                FailureLevel.WARNING,
```
demartinofra: can this be an INFO?

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,128 @@
...
+                "'SSL_CA' Slurmdbd StorageParameter not provided. This may prevent database "
+                "server identity verification from slurmdbd daemon.",
+                FailureLevel.WARNING,
+            )
+        for key, value in slurmdbd_storage_parameters.items():
+            if key not in SLURMDBD_STORAGE_PARAMETERS:
+                self._add_failure(
+                    f"'{key}' not available as Slurmdbd StorageParameter. The provided parameter will be ignored.",
+                    FailureLevel.INFO,
+                )
```
demartinofra: why not failing? a typo might go unnoticed otherwise

cli/src/pcluster/validators/database_validators.py
```diff
@@ -0,0 +1,128 @@
...
+            return False
+        if not port:
+            self._add_failure(
+                f"No port specified in the URI. Assuming the use of port {default_mysql_port}",
+                FailureLevel.WARNING,
+            )
+        return True
+
+
+class SlurmdbdStorageParametersValidator(Validator):
```
demartinofra: todo: review the validation/messaging flow

cli/src/pcluster/validators/database_validators.py
jdeamicis: Yes, I will change it.

cli/src/pcluster/validators/database_validators.py
jdeamicis: We need to discuss about this implementation. I did not create a schema for the `StorageParameter` parameter because I think we should not take ownership of that parameter (and in the longer run we should have a more generalized solution to allow users to customize the Slurm configuration).

Regarding the FailureLevel, I guess I can increase it to ERROR.

cli/src/pcluster/validators/database_validators.py
jdeamicis: Done.

cli/src/pcluster/schemas/cluster_schema.py
jdeamicis: Yes, we should discuss this.

cli/src/pcluster/templates/cdk_builder_utils.py
jdeamicis: At first I thought the dna.json was the right way to export parameters to make them available to Chef (because it's ubiquitous in the code). After your comment I dug a bit more in the cookbook and found [this](https://github.com/aws/aws-parallelcluster-cookbook/blob/da0302f6a491987dc35b078a3a9e3f0a89d61f4a/cookbooks/aws-parallelcluster-config/recipes/config.rb#L18-L22):

```chef
include_recipe 'aws-parallelcluster-config::base'

fetch_config 'Fetch and load cluster configs' unless node['cluster']['scheduler'] == 'awsbatch'

include_recipe 'aws-parallelcluster-slurm::config' if node['cluster']['scheduler'] == 'slurm'
```

which I guess is why all the operations performed in `base.rb` currently rely on parameters defined in either `defaults.rb` or exported to the `dna.json` (e.g. directory service).

Since the configuration of Slurm happens after the config fetching operation I can now remove what I had exported into the `dna.json` and use the imported structure. Is there a reason we fetch the config only after the `base.rb` recipe has been called?

jdeamicis: Actually, we do need at least a trigger (`['slurm']['database']['enabled']) for the Cloudwatch agent: https://github.com/aws/aws-parallelcluster-cookbook/pull/1532/ .

@demartinofra @davprat maybe we should just remove the `feature_conditions` for the export of `/var/log/slurmdbd.log`?

jdeamicis: Done.

cli/src/pcluster/validators/database_validators.py
jdeamicis: Done

cli/src/pcluster/validators/database_validators.py
jdeamicis: Done.

cli/src/pcluster/validators/database_validators.py
EddyMM: > most validation functions had a structure like if ... elif ..., meaning that they implicitly return only one type of validation error.

This seems to happen when comparing a specific property as opposed to a range of properties.

For the intended experience, it may be worth discussing it with the wider team but it's not a technical blocker.

cli/src/pcluster/templates/cdk_builder_utils.py
jdeamicis: Fdm says it's OK to remove the feature_conditions.

cli/src/pcluster/schemas/cluster_schema.py
jdeamicis: This is going to be removed from the beta. Resolving this conversation.

cli/src/pcluster/validators/database_validators.py
jdeamicis: This is going to be removed from the beta anyway.

cli/src/pcluster/validators/database_validators.py
jdeamicis: Removed from the beta.

cli/src/pcluster/validators/database_validators.py
jdeamicis: We will leave it like this for the time being. Waiting for feedback from beta testers.

cli/src/pcluster/templates/cdk_builder_utils.py
jdeamicis: Done.

davprat (dismissed):

davprat (approved):

Q-Hilbtec:
Dear @davprat, thanks for all those changes.. Would you guys consider also enabling `jobacct_gather` (e.g. see https://aws.amazon.com/blogs/compute/enabling-job-accounting-for-hpc-with-aws-parallelcluster-and-amazon-rds/) in order to get real-time job stats via `sstat` and whatnot? I would make a separate issue but not sure of the visibility. It seems quite trivial (few lines of slurm config) to enable it now that you're getting accounting / a database as a whole.

PR #4358, opened 2022-09-21
Allow `apt-get` to upgrade openssh-server in 3.x

stephenmsachs:
This fixes a hang in `pcluster build-image` when
`Build:UpdateOsPackages:Enabled:true` is set.

`apt-get upgrade` will prompt for user input during upgrade of `openssh-server` as `/etc/ssh/ssd_config` is locally modified and would be overwritten by package provided file. The proposed patch removes the prompt and chooses to keep the existing configuration file.

I can see this issue pop up on ubuntu-{1804,2004} in pcluster-[3.1.1-3.1.4]. I did not test older releases, but would expect it to occur there do.

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
@@ -119,7 +119,7 @@ phases:
                 while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done
                 flock $(apt-config shell StateDir Dir::State/d | sed -r "s/.*'(.*)\/?'$/\1/")/daily_lock systemctl disable --now apt-daily.timer apt-daily.service apt-daily-upgrade.timer apt-daily-upgrade.service
                 sed "/Update-Package-Lists/s/\"1\"/\"0\"/; /Unattended-Upgrade/s/\"1\"/\"0\"/;" /etc/apt/apt.conf.d/20auto-upgrades > "/etc/apt/apt.conf.d/51pcluster-unattended-upgrades"
-                DEBIAN_FRONTEND=noninteractive apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --with-new-pkgs --allow-change-held-packages upgrade && apt-get --purge autoremove -y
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -119,7 +119,7 @@ phases:
                 while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done
                 flock $(apt-config shell StateDir Dir::State/d | sed -r "s/.*'(.*)\/?'$/\1/")/daily_lock systemctl disable --now apt-daily.timer apt-daily.service apt-daily-upgrade.timer apt-daily-upgrade.service
                 sed "/Update-Package-Lists/s/\"1\"/\"0\"/; /Unattended-Upgrade/s/\"1\"/\"0\"/;" /etc/apt/apt.conf.d/20auto-upgrades > "/etc/apt/apt.conf.d/51pcluster-unattended-upgrades"
-                DEBIAN_FRONTEND=noninteractive apt-get -y update && apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --with-new-pkgs upgrade && apt-get --purge autoremove -y
+                DEBIAN_FRONTEND=noninteractive apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --with-new-pkgs --allow-change-held-packages upgrade && apt-get --purge autoremove -y
```
enrico-usai: Thanks Stephen for t he patch.

What about:
`export DEBIAN_FRONTEND=noninteractive && apt-get -y update && apt-get -y -o ...`

enrico-usai: Instead I'd remove the `--allow-change-held-packages`, according to official documentation this is potentially dangerous:
> --allow-change-held-packages
           Force yes; this is a dangerous option that will cause apt to continue without
           prompting if it is changing held packages. It should not be used except in very
           special situations. Using it can potentially destroy your system! Configuration Item:
           APT::Get::allow-change-held-packages. Introduced in APT 1.1.

stephenmsachs: You are correct. it works wihtout `--allow-change-held-packages`

enrico-usai (approved):

PR #4362, opened 2022-09-22
Allow `apt-get` to upgrade openssh-server in 3.x (#4358)

stephenmsachs:
Allow `apt-get` to upgrade openssh-server

This fixes a hang in `pcluster build-image` when `Build:UpdateOsPackages:Enabled:true` is set.

`apt-get upgrade` will prompt for user input during upgrade of `openssh-server` as `/etc/ssh/ssd_config` is locally modified and would be overwritten by package provided file. The proposed patch removes the prompt and chooses to keep the existing configuration file.

I can see this issue pop up on ubuntu-{1804,2004} in pcluster-[3.1.1-3.1.4]. I did not test older releases, but would expect it to occur there do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
 **BUG FIXES**
 - Avoid failing on DescribeCluster when cluster configuration is not available.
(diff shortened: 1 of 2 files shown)


enrico-usai (approved):

lukeseawalker:
Closing this since it didn't make to 3.2.1

PR #4371, opened 2022-09-23
[integ-tests] Add integ test for DFSM

chenwany:
### Description of changes
* Add integration test for DFSM

### Tests
* Integration test:
```
  update:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["c5.xlarge"]
          oss: ["alinux2"]
          schedulers: ["slurm"]
```

### References

### Checklist


tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
     _check_queue_ami(cluster, ec2, pcluster_ami_id, "queue1")
     _check_queue_ami(cluster, ec2, pcluster_copy_ami_id, "queue2")
     assert_compute_node_states(scheduler_commands, queue1_nodes, "idle")
+
+
+@pytest.mark.usefixtures("instance")
+def test_dynamic_file_systems_update(
```
gmarciani, line 767: The dynamic file system update is a feature that is strictly related to the storage. I think it would make sense to move this test in `tests/storage` than being in `tests/update`

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+    fsx_factory,
+    svm_factory,
+    open_zfs_volume_factory,
+):
+    """Test update shared storages."""
+    existing_ebs_mount_dir = "/existing_ebs_mount_dir"
+    existing_efs_mount_dir = "/existing_efs_mount_dir"
+    fsx_lustre_mount_dir = "/fsx_lustre_mount_dir"
+    fsx_ontap_mount_dir = "/fsx_ontap_mount_dir"
+    fsx_open_zfs_mount_dir = "/fsx_open_zfs_mount_dir"
```
gmarciani: These file systems are existing, not managed. What about having mountidrs `existing_fsx_lustre_mount_dir`,`existing_fsx_ontap_mount_dir`,`existing_fsx_open_zfs_mount_dir`?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+    new_ebs_mount_dir = "/new_ebs_mount_dir"
+    new_raid_mount_dir = "/raid_dir"
```
gmarciani: What about naming this `new_raid_dir`, like you did in the others?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
+    (
+        ebs_volume_id,
+        existing_efs_id,
+        existing_fsx_lustre_fs_id,
+        fsx_ontap_volume_id,
+        fsx_open_zfs_volume_id,
```
gmarciani: Since all of these resources are existing file systems, what about having all variable names starting with `existing_`?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+        new_ebs_mount_dir=new_ebs_mount_dir,
+        new_raid_mount_dir=new_raid_mount_dir,
+        new_lustre_mount_dir=new_lustre_mount_dir,
+        new_efs_mount_dir=new_efs_mount_dir,
+    )
+
+    cluster.update(str(update_cluster_config))
+
+    # check chef client log contains expected log
+    assert_errors_in_logs(
```
gmarciani: The name of this function is misleading: we are checking lines in log, that may not be error logs, right?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+    cluster = clusters_factory(init_config_file)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = scheduler_commands_factory(remote_command_executor)
+    cluster_nodes = scheduler_commands.get_compute_nodes()
+    queue1_nodes = [node for node in cluster_nodes if "queue1" in node]
+
+    # submit a job to queue1
+    queue1_job_id = scheduler_commands.submit_command_and_assert_job_accepted(
+        submit_command_args={
+            "command": "sleep 3000",
```
gmarciani: At first, such a long sleep time sounds unreasonable, but then looking at the rest of the test it makes sense. I suggest to describe in a short comment why you need such a long sleep time.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+        new_raid_mount_dir,
+        efs_mount_dirs,
+        fsx_mount_dirs,
+    )
+
+    # check newly mounted file systems are not visible on compute nodes that are running jobs
+    for node_name in queue1_nodes:
+        _test_directory_not_mounted(remote_command_executor, all_mount_dirs, node_type="compute", node_name=node_name)
+
+    scheduler_commands.cancel_job(queue1_job_id)
```
gmarciani, line 896: Shouldn't we wait some time between job cancellation and the verification that the expected shared storages are mounted on compute nodes?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+    new_efs_mount_dir = "/new_efs_mount_dir"
+    new_lustre_mount_dir = "/new_lustre_mount_dir"
+    ebs_mount_dirs = [new_ebs_mount_dir, existing_ebs_mount_dir]
+    efs_mount_dirs = [existing_efs_mount_dir, new_efs_mount_dir]
+    fsx_mount_dirs = [new_lustre_mount_dir, fsx_lustre_mount_dir, fsx_open_zfs_mount_dir, fsx_ontap_mount_dir]
+    all_mount_dirs = ebs_mount_dirs + [new_raid_mount_dir] + efs_mount_dirs + fsx_mount_dirs
+
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
```
gmarciani, line 802: My understanding is that we upload this file to a bucket attached to the FSx Lustre file system.
Why do we need that? Isn't enough to mount a FSx Lustre file system without an underlying bucket?

tests/integration-tests/tests/update/test_update.py
gmarciani, line 767: I think we should test the dynamic file system update in the following scenarios:
1. QueueUpdateStrategy=DRAIN
2. QueueUpdateStrategy=TERMINATE
3. QueueUpdateStrategy=COMPUTE_FLEET_STOP

If you agree with that, I think this test should be parameterized by that strategy.

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -0,0 +1,103 @@
...
+        Type: 0
+        NumberOfVolumes: 5
+  - MountDir: {{ new_ebs_mount_dir }}
+    Name: /ebs1
+    StorageType: Ebs
+    EbsSettings:
+      VolumeType: gp3
+      DeletionPolicy: Delete
+  - MountDir: {{ existing_ebs_mount_dir }}
+    Name: existing_eebs
```
gmarciani: typo: "eebs" => "ebs"

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -0,0 +1,103 @@
...
+    StorageType: Ebs
+    EbsSettings:
+      VolumeType: gp2
+      VolumeId: {{ volume_id }}
+  - MountDir: {{ new_efs_mount_dir }}
+    Name: efs1
+    StorageType: Efs
+    EfsSettings: # Customize EFS settings to have better coverage outside other EFS tests.
```
gmarciani: I don't get the goal of the comment in this line

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -0,0 +1,103 @@
...
+  - MountDir: {{ new_lustre_mount_dir }}
+    Name: fsx1
+    StorageType: FsxLustre
+    FsxLustreSettings:
+      StorageCapacity: 1200
+      ImportPath: s3://{{ bucket_name }}
+      ExportPath: s3://{{ bucket_name }}/export_dir
+      DeploymentType: PERSISTENT_1
+      PerUnitStorageThroughput: 200
+  - MountDir: {{ fsx_lustre_mount_dir }} # the index of mount dirs starts from the second item. Because the first is used for the new FSx.
```
gmarciani: I don't get the goal of the comment in this line

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -0,0 +1,103 @@
...
+            - InstanceType: c5.xlarge
+          MinCount: 1
+          MaxCount: 2
+      Networking:
+        SubnetIds:
+          - {{ private_subnet_id }}
+SharedStorage:
+  - MountDir: /raid_dir
+    StorageType: Ebs
+    Name: raid-type1
```
gmarciani: We have both managed and existing file systems. What about prefixing managed file system names with "managed-" as you did for "existing-"?

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -0,0 +1,103 @@
...
+  Os: {{ os }}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Iam:
+    S3Access:
+      - BucketName: {{ bucket_name }}
```
gmarciani, line 11: Why does the head node require read access to the bucket if we do not execute any explicit call to AWS S3 targeting that bucket? Same question for compute nodes below.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+        bucket_name,
+        scheduler_commands_factory,
+        ebs_mount_dirs,
+        new_raid_mount_dir,
+        efs_mount_dirs,
+        fsx_mount_dirs,
+    )
+
+    # check newly mounted file systems are not visible on compute nodes that are running jobs
+    for node_name in queue1_nodes:
```
gmarciani, line 892: Here we are verifying that queues with running jobs don't see the mounting while having running jobs.
I think we should also verify that queues with no running jobs they immediately see the mounting.
What do you think?

tests/integration-tests/tests/update/test_update.py
```diff
@@ -749,3 +761,295 @@ def _test_update_queue_strategy_with_running_job(
...
+        _test_directory_not_mounted(remote_command_executor, all_mount_dirs, node_type="compute", node_name=node_name)
+
+    scheduler_commands.cancel_job(queue1_job_id)
+    for mount_dir in all_mount_dirs:
+        verify_directory_correctly_shared(remote_command_executor, mount_dir, scheduler_commands, partition="queue1")
+        verify_directory_correctly_shared(remote_command_executor, mount_dir, scheduler_commands, partition="queue2")
```
gmarciani: Queue2 does not require the job in queue1 to be cancelled in order to see the mounting. What about moving this check before the job cancellation?

tests/integration-tests/tests/update/test_update.py
chenwany: Yes, it make sense to put it in test/storages. I think it also make sense put it under tests/update because it only related to update of FS and related to queue_update_stragegy.
Both queue_update_strategy and update functionality are under test_update

tests/integration-tests/tests/update/test_update.py
chenwany: sure, renamed

tests/integration-tests/tests/update/test_update.py
chenwany: Raid can only be new one and there's no existing raid. But we can rename it to be aligned

tests/integration-tests/tests/update/test_update.py
chenwany: added

tests/integration-tests/tests/update/test_update.py
chenwany: No, we don't need to.  Expect shared storages are mounted on compute nodes is tested by submitting a job. Right after cancel the job, the old nodes will be DRAIN, those nodes accept the new jobs are new nodes

tests/integration-tests/tests/update/test_update.py
chenwany: The bucket is used in the integration test to test fsx functionality after update cluster with fsx to test data repository task and export path

tests/integration-tests/tests/update/test_update.py
chenwany: This test takes long time to run, different queue update strategies are already tested in unit tests, it is also covered in test_queue_update_strategy integration test. We already test it for all the architectures and oss. I don't think we should test the combinations of each parameter here

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
chenwany: Fixed

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
chenwany: removed

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
chenwany: removed

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
chenwany: It is used in the test: `sudo aws s3 cp --region`. Test will fail without specifying the s3 bucket

tests/integration-tests/tests/update/test_update.py
chenwany: We already verify that right after cancel the job in queue1
Do you mean I should verify it before canceling the job in queue1? The job cancelation command is fast, and it is not related to queue2
```
    # check newly mounted file systems are not visible on compute nodes that are running jobs
    for node_name in queue1_nodes:
        _test_directory_not_mounted(remote_command_executor, all_mount_dirs, node_type="compute", node_name=node_name)

    scheduler_commands.cancel_job(queue1_job_id)
    for mount_dir in all_mount_dirs:
        verify_directory_correctly_shared(remote_command_executor, mount_dir, scheduler_commands, partition="queue1")
        verify_directory_correctly_shared(remote_command_executor, mount_dir, scheduler_commands, partition="queue2")
```

tests/integration-tests/tests/update/test_update.py
chenwany: Moved

tests/integration-tests/tests/update/test_update.py
chenwany: Just see the below comment, moved

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
chenwany: good idea, changed

tests/integration-tests/tests/update/test_update.py
chenwany: yes, this name is misleading, let's rename it

tests/integration-tests/tests/update/test_update.py
gmarciani: Not a blocking comment, but I still think it should be within the storage folder because the organization of the integ tests folders is mostly by feature/component rather than by workflow.

tests/integration-tests/tests/update/test_update.py
gmarciani: Yes, let's keep them aligned. Thank you

tests/integration-tests/tests/update/test_update.py
gmarciani: That's great, thanks for clarifying it. Let's proceed with your solution.

tests/integration-tests/tests/update/test_update.py
gmarciani: Thanks for the explanation. I agree with you that the more we test the better it is, but I'd like to understand this point a bit better. The goal of the test here is to verify that file systems are correctly mounted/unmounted in a way that is ruled by the QueueUpdateStrategy. Isn't enough to check the mount points instead of using S3?

tests/integration-tests/tests/update/test_update.py
gmarciani: Got it now, thanks

gmarciani (approved):
You did a great job, thank you!
Please consider the remaining comments as non blocking, just for me to better understand your position.
Before merging: launch the test locally to verify that it behaves as expected (did see such validation in PR description)

tests/integration-tests/tests/update/test_update.py
chenwany: Previously, we don't support update cluster to add filesystems. As we support it now, I just want to verify that when we update a cluster to add fsx, everything of fsx works as expected. I agree it is a nice to have, I don't have a strong opinion on keeping it or removing it. Both are fine to me.

PR #4390, opened 2022-09-30
Add a template to create a serverless database

davprat:
### Description of changes
* Add a 'tutorial' level CloudFormation template for creating a serverless database cluster.

### Tests
* Deployed into an existing VPC
* Created a cluster with accounting enabled
* Verified connection from head node with 'sacctmgr list users -Pn'
* Verified connection from head node by creating a job with sbatch and then executing 'sacct -j <job id>'
* Deployed stack without existing VPC

### Checklist

### DIFF
--- a/cloudformation/database/serverless-database.yaml
+++ b/cloudformation/database/serverless-database.yaml
@@ -1,0 +1,344 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  This template is provided as part of a tutorial on how to enable Slurm Accounting using Parallel Cluster.
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: "Database Cluster Configuration"
+        Parameters:
+          - ClusterAdmin
+          - AdminPasswordSecretName
+          - AdminPasswordSecretString
+          - MinCapacity
+          - MaxCapacity
+      - Label:
+          default: "Network Configuration"
+        Parameters:
+          - Vpc
+          - DatabaseClusterSubnetOne
+          - DatabaseClusterSubnetTwo
+    ParameterLabels:
+      Vpc:
+        default: "Which VPC should the database cluster be created in?"
+      ClusterAdmin:
+        default: "What should the database administrator user name be?"
+      AdminPasswordSecretName:
+        default: "What should the name of the secret containing the administrator password be?"
+      AdminPasswordSecretString:
+        default: "What should the administrator password be?"
+      MinCapacity:
+        default: "What should the minimum scaling capacity of the database cluster be?"
+      MaxCapacity:
+        default: "What should the maximum scaling capacity of the database cluster be?"
+      DatabaseClusterSubnetOne:
+        default: "What is the first subnet the database cluster should be deployed in?"
+      DatabaseClusterSubnetTwo:
+        default: "What is the second subnet the database cluster should be deployed in?"
+Parameters:
+  AdminPasswordSecretName:
+    Description: Name of then Secrets Manager secret for the database cluster administrator password.
+    Type: String
+    MinLength: 2
+    MaxLength: 1024
+  AdminPasswordSecretString:
+    Description: Password for the database cluster administrator (leave BLANK to import the password from Secrets Manager).
+    Type: String
+    NoEcho: true
+    # Only allow 'Medium' or better strength passwords according to https://dev.mysql.com/doc/refman/8.0/en/validate-password.html
+    AllowedPattern: (?=^.{8,}$)(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\w\s])^.*|^$
+    ConstraintDescription: Password must be BLANK or at least 8 characters with at least one uppercase letter, one lower case letter, one numeric digit, and one special (non-alphanumeric) character.
+    Default: ""
+  ClusterAdmin:
+    Description: Database cluster administrator user name.
+    Type: String
+    Default: clusteradmin
+    MinLength: 3
+    MaxLength: 64
+  Vpc:
+    Description: VPC ID to create the database cluster in (leave BLANK to create a new one).
+    Type: String
+    # Type: AWS::EC2::VPC::Id
+    Default: ""
+  DatabaseClusterSubnetOne:
+    Description: Subnet ID of the first subnet (leave BLANK to create).
+    Type: String
+    # Type: AWS::EC2::Subnet::Id
+    Default: ""
+  DatabaseClusterSubnetTwo:
+    Description: Subnet ID of the second subnet (leave BLANK to create). This subnet should be in different availability zone from the first subnet.
+    Type: String
+    # Type: AWS::EC2::Subnet::Id
+    Default: ""
+  MinCapacity:
+    Description: Minimum scaling capacity for the database cluster. Must be less than or equal to MaxCapacity.
+    Type: Number
+    Default: 1
+    AllowedValues:
+      - 1
+      - 2
+      - 4
+      - 8
+      - 16
+      - 32
+      - 64
+      - 128
+      - 256
+  MaxCapacity:
+    Description: Maximum scaling capacity for the database cluster. Must be greater than or equal to MinCapacity
+    Type: Number
+    Default: 4
+    AllowedValues:
+      - 1
+      - 2
+      - 4
+      - 8
+      - 16
+      - 32
+      - 64
+      - 128
+      - 256
+Transform: AWS::Serverless-2016-10-31
+Conditions:
+  CreateVPC: !Equals [!Ref Vpc, '']
+  CreateSubnets: !Or [!Condition CreateVPC, !Equals [!Ref DatabaseClusterSubnetOne, ''], !Equals [!Ref DatabaseClusterSubnetTwo, '']]
+  CreateAdminSecret: !Not [!Equals [!Ref AdminPasswordSecretString, '']]
+Rules:
+  EmptyVpcAssertions:
+    RuleCondition: !Equals
+      - !Ref Vpc
+      - ''
+    Assertions:
+      - Assert: !Equals
+          - !Ref DatabaseClusterSubnetOne
+          - ''
+        AssertDescription: DatabaseClusterSubnetOne must be empty if Vpc is not provided.
+      - Assert: !Equals
+          - !Ref DatabaseClusterSubnetTwo
+          - ''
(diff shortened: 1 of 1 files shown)


cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+          - MaxCapacity
+      -
+        Label:
+          default: "Network Configuration"
+        Parameters:
+          - Vpc
+          - PrivateSubnetOne
+          - PrivateSubnetTwo
+    ParameterLabels:
+      Vpc:
```
gmarciani, line 22: Why do we need a ParameterLabel for Vpc? Isn't enough to have the Description for it?
Why having ParameterLabels for Vpc and not for others, e.g. PrivateSubnetOne and PrivateSubnetTwo?

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+        default: "Which VPC should this be deployed to?"
+Parameters:
+  AdminPasswordSecret:
```
gmarciani: We are assuming that the user has already defined a Secret in SecretsManager.
I think it could be useful to request the Password as template parameters, let the template create the secret and expose the secret ARN as a template output.
what do you think?

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+      InstanceTenancy: default
+      Tags:
+        - Key: Name
+          Value: ServerlessDatabaseClusterVPC
+        - Key: 'parallel-cluster:accounting'
+          Value: vpc
+        - Key: 'parallel-cluster:accounting:scheduler'
+          Value: slurm
+        - Key: 'parallel-cluster:accounting:version'
+          Value: '1.0'
```
gmarciani: Here and below: I think we can avoid all tags, but Name.
I'm curious: why do you think they may be useful?

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+      VpcSecurityGroupIds:
+        - !GetAtt
+          - ServerlessDatabaseClusterSecurityGroup
+          - GroupId
+    UpdateReplacePolicy: Snapshot
+    DeletionPolicy: Snapshot
```
gmarciani: I understand the reasoning behind `DeletionPolicy: Snapshot` and `UpdateReplacePolicy: Snapshot`. Nevertheless, the template is meant to be used for tutorials and beta testing, so we do not want to introduce extra cost for users once the stack will be deleted.
I think all the resources should be deleted at stack deletion.

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+      Name: !Sub 'parallel-cluster:accounting:database:port:${AWS::StackName}'
+  parallelclusterdatabaseadminuser:
+    Value: !Ref ClusterAdmin
+    Export:
+      Name: !Sub 'parallel-cluster:accounting:database:admin-user:${AWS::StackName}'
+  parallelclusterdatabasesecretsname:
+    Value: !Ref AdminPasswordSecret
+    Export:
+      Name: !Sub 'parallel-cluster:accounting:database:secrets:name:${AWS::StackName}'
+  parallelclusterdatabasevpc:
```
gmarciani, line 335: I think it could be useful to also have the Subnets as stack outputs.

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+      DestinationSecurityGroupId: !GetAtt
+        - ServerlessDatabaseClusterSecurityGroup
+        - GroupId
+      FromPort: !GetAtt
+        - ServerlessDatabaseCluster
+        - Endpoint.Port
+      ToPort: !GetAtt
+        - ServerlessDatabaseCluster
+        - Endpoint.Port
+Outputs:
```
gmarciani, line 310: General comment on `Outputs`:
1. why should we declare `Exports: Name` instead of simply setting their name with the key?
I mean having this:

```
Outputs:
  ParameterName:
    Value: ParameterValue
```

instead of:

```
Outputs:
  OutputName:
    Value: OutputValue
    Exports:
      Name: AnotherOutputName
```

1. I think it's not necessary to prefix all outputs with `parallelcluster` because (i) the stack does not contain anything specific to pcluster and (ii) even if the stack would be really specific to pcluster, we can consider it implicit.

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,345 @@
...
+    Type: String
+    Default: clusteradmin
+    MinLength: 3
+    MaxLength: 64
+  Vpc:
+    Description: VPC ID to create the Database Cluster in (leave BLANK to create a new one).
+    Type: String
+    # Type: AWS::EC2::VPC::Id
+    Default: ""
+  PrivateSubnetOne:
```
gmarciani: Is it necessary for the subnets to be private?
I mean: from a security standpoint that's better because we do not want to expose a public ip for the database, but that is not mandatory.
What do you think?

cloudformation/database/serverless-database.yaml
davprat: Changed to `Delete`.

cloudformation/database/serverless-database.yaml
davprat: Is there a good reason to not make them private? My thought is that by default, they should be as locked down as possible unless there is a reason not to?

cloudformation/database/serverless-database.yaml
davprat: I was wondering about this. My thinking was that these subnets should be specific and private to the database and not something really exposed for use. For the outputs, I was trying to expose everything that was needed to configure Slurm accounting in the pcluster config file. What is the reason for exposing the subnets?

cloudformation/database/serverless-database.yaml
davprat: I think I am trying to make sure the resources created by this stack are easily discernible in their relevant consoles, but maybe this isn't a good idea?

cloudformation/database/serverless-database.yaml
davprat: I was just trying to make the form look a bit nicer. I can remove it though.

cloudformation/database/serverless-database.yaml
gmarciani: Isn't it necessary for the cluster and database to be deployed in the same subnets?
I assumed it was necessary. If it is, I suggest to expose them in Outputs.

cloudformation/database/serverless-database.yaml
gmarciani: Yes, this is the best practice, but it's not mandatory and the name of the parameter let me think that it will only work if using private subnets.
Example: a customer can also deploy the whole cluster in a public subnet, even if we suggest to deploy compute fleet in a private one.

I suggest to remove the prefix "Private" from subnets Parameters and add in the parameter description the suggestion of using private subnets

cloudformation/database/serverless-database.yaml
gmarciani: I tried the template and it actually looks nicer! Let's keep them and thanks for them

cloudformation/database/serverless-database.yaml
davprat: I have been deploying them into different subnets. I think it is only necessary (at least without doing some amount more work) that the cluster and database reside in the same VPC. Of course, I may be missing something here.

cloudformation/database/serverless-database.yaml
davprat: I think it's a good idea - I've added the option to provide a password to the template.

cloudformation/database/serverless-database.yaml
davprat: Ah OK, I think I understand what you are saying now. I've renamed the parameters.

cloudformation/database/serverless-database.yaml
gmarciani: I think you're right. In any case I think it's useful to have the subnets exposed as Outputs, especially because your template support the creation of brand new VPC and Subnets.

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,344 @@
...
+      ClusterAdmin:
+        default: "What should the database administrator user name be?"
+      AdminPasswordSecretName:
+        default: "What should the name of the secret containing the administrator password be?"
+      AdminPasswordSecretString:
+        default: "What should the administrator password be?"
+      MinCapacity:
+        default: "What should the minimum scaling capacity of the database cluster be?"
+      MaxCapacity:
+        default: "What should the maximum scaling capacity of the database cluster be?"
```
gmarciani, line 33: Minor, here and in other descriptions: I'd put the description as a description rather than a question, i.e.:

```
The maximum scaling capacity of the database cluster.
```

instead of 

```
What should the maximum scaling capacity of the database cluster be?
```

cloudformation/database/serverless-database.yaml
```diff
@@ -0,0 +1,344 @@
...
+    Properties:
+      CidrBlock: 10.0.0.0/16
+      EnableDnsHostnames: true
+      EnableDnsSupport: true
+      InstanceTenancy: default
+      Tags:
+        - Key: Name
+          Value: ServerlessDatabaseClusterVPC
+        - Key: 'parallel-cluster:accounting'
+          Value: vpc
```
gmarciani, line 136: Minor, here and elsewhere: I understand the reason why you're adding these tags and I agree. Nevertheless, I have the following observations:
1. Let's align key names with the ones used by ParallelCluster, i.e. `parallelcluster:something`, instead of `parallel-cluster:something`
2. We need a tag that declares a relation with the use case of the infrastructure. In this case the use case is Slurm Accounting, so I'd put a tag with name `parallelcluster:infrastructure` or `parallelcluster:usecase` and value  `Slurm Accounting`. We could replicate this in any other 1-click template , e.g. the one with the AD infrastructure would have `parallelcluster:usecase=Multi User`
3. I would avoid tags like `parallelcluster:accounting=resource` (e.g. `parallelclluster:accountig=vpc` becausde the type of resource is implicit in the resource itself and it does not need a tag for it.

gmarciani (approved):

cloudformation/database/serverless-database.yaml
davprat: Yes, this makes a lot more sense.

PR #4393, opened 2022-09-30
[3.3] Create an update policy for managed placement groups

dreambeyondorange:
### Description of changes
Managed placement group deletion requires the compute fleet to be stopped. The update policy for placement groups and all of the parameters within needs to check for deletion when updating the cluster.  Also, the networking section of the compute resource only contains placement group, so that section follows this policy as well.

Signed-off-by: Ryan Anderson <ndry@amazon.com>

### Tests
* Manual tests for update policy
* Unit tests added for update policy related to managed placement groups
* The relevant pcluster 3 integration tests are passing with these changes

### References
* https://quip-amazon.com/kaLSAnnbyR7I/Design-Add-Support-for-On-Demand-Capacity-Reservations-ODCRs-in-Cluster-Configuration-File

### Checklist

### DIFF
--- a/cli/src/pcluster/config/update_policy.py
+++ b/cli/src/pcluster/config/update_policy.py
@@ -8,6 +8,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 from enum import Enum
 
 from pcluster.config.cluster_config import QueueUpdateStrategy
@@ -112,6 +113,14 @@ def actions_needed_queue_update_strategy(change, _):
     return actions
 
 
+def actions_needed_managed_placement_group(change, patch):
+    if is_managed_placement_group_deletion(change, patch):
+        actions = "Stop the compute fleet with the pcluster update-compute-fleet command."
+    else:
+        actions = actions_needed_queue_update_strategy(change, patch)
+    return actions
+
+
 def condition_checker_compute_fleet_stop_on_remove(change, patch):
     result = not patch.cluster.has_running_capacity()
     # SlurmQueue or ComputeResource can be added but removal require compute fleet stop
@@ -123,6 +132,60 @@ def condition_checker_compute_fleet_stop_on_remove(change, patch):
 
 def is_slurm_queues_change(change):
     return any(path.startswith("SlurmQueues[") for path in change.path)
+
+
+def extract_type_and_name_from_path(path):
+    # Example path = 'SlurmQueues[slurm-q-name]'
+    # This function returns the type and name extracted like this: 'SlurmQueues', 'slurm-q-name'
+    obj_type = re.search(".+(?=\\[)", path).group(0)  # Get the sub string before the '['
+    obj_name = re.search("(?<=\\[).+(?=\\])", path).group(0)  # Get the sub string between '[' and ']'
+    return obj_type, obj_name
+
+
+def get_q_from_config(change, config):
+    # Example path=['Scheduling', 'SlurmQueues[q-pg-enabled]', 'ComputeResources[cr-pg-enabled]']
+    # This function would return the dictionary 'q-pg-enabled' from the config using the key from the change path
+    q_type, q_name = None, None
+    for path in change.path:
+        if re.search("Queues\\[", path):
+            q_type, q_name = extract_type_and_name_from_path(path)
+    if q_type and q_name:
+        for queue in config.get("Scheduling", {}).get(q_type, {}):
+            if queue.get("Name", None) == q_name:
+                return queue
+    return {}
+
+
+def get_cr_from_config(change, config):
+    # Example path=['Scheduling', 'SlurmQueues[q-pg-enabled]', 'ComputeResources[cr-pg-enabled]']
+    # This function would return the dictionary 'cr-pg-enabled' from the config using the key from the change path
+    cr_type, cr_name = None, None
+    for path in change.path:
+        if re.search("ComputeResources\\[", path):
+            cr_type, cr_name = extract_type_and_name_from_path(path)
+    if cr_type and cr_name:
+        queue = get_q_from_config(change, config)
+        for compute_resource in queue.get(cr_type, {}):
+            if compute_resource.get("Name", None) == cr_name:
+                return compute_resource
+    return {}
+
+
+def is_placement_group_managed_for_compute_resource(queue_networking, compute_resource_networking):
+    chosen_pg = compute_resource_networking.get("PlacementGroup") or queue_networking.get("PlacementGroup")
+    if chosen_pg and chosen_pg.get("Enabled") and not (chosen_pg.get("Name") or chosen_pg.get("Id")):
+        return True
+    return False
+
+
+def is_managed_placement_group_deletion(change, patch):
+    base_q_networking = get_q_from_config(change, patch.base_config).get("Networking", {})
+    base_cr_networking = get_cr_from_config(change, patch.base_config).get("Networking", {})
+    target_q_networking = get_q_from_config(change, patch.target_config).get("Networking", {})
+    target_cr_networking = get_cr_from_config(change, patch.target_config).get("Networking", {})
+    return is_placement_group_managed_for_compute_resource(
+        base_q_networking, base_cr_networking
+    ) and not is_placement_group_managed_for_compute_resource(target_q_networking, target_cr_networking)
 
 
 def is_slurm_scheduler(patch):
@@ -167,6 +230,14 @@ def fail_reason_queue_update_strategy(change, _):
     return reason
 
 
+def fail_reason_managed_placement_group(change, patch):
+    if is_managed_placement_group_deletion(change, patch):
+        reason = "All compute nodes must be stopped for a managed placement group deletion"
+    else:
+        reason = fail_reason_queue_update_strategy(change, patch)
+    return reason
+
+
 def condition_checker_queue_update_strategy(change, patch):
     result = not patch.cluster.has_running_capacity()
     # QueueUpdateStrategy can override UpdatePolicy of parameters under SlurmQueues
@@ -179,6 +250,14 @@ def condition_checker_queue_update_strategy(change, patch):
             != QueueUpdateStrategy.COMPUTE_FLEET_STOP.value
         )
 
+    return result
+
+
+def condition_checker_managed_placement_group(change, patch):
+    if is_managed_placement_group_deletion(change, patch) and patch.cluster.has_running_capacity():
+        result = False
+    else:
+        result = condition_checker_queue_update_strategy(change, patch)
     return result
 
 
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,68 @@ def is_slurm_queues_change(change):
...
+    if cr_type and cr_name:
+        queue = get_q_from_config(change, config)
+        for compute_resource in queue.get(cr_type, {}):
+            if compute_resource.get("Name", None) == cr_name:
+                return compute_resource
+    return None
+
+
+def is_placement_group_managed_for_compute_resource(queue_networking, compute_resource_networking):
+    cr_pg = compute_resource_networking.get("PlacementGroup", None)
```
hanwen-pcluste: The `None` default is optional

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,68 @@ def is_slurm_queues_change(change):
...
+    cr_pg_enabled = compute_resource_networking.get("PlacementGroup", {}).get("Enabled")
+    q_pg_enabled = queue_networking.get("PlacementGroup", {}).get("Enabled")
+    if (cr_pg is None and q_pg_enabled) or cr_pg_enabled:
+        return True
+    return False
```
hanwen-pcluste: The code can be simplified:
```
chosen_pg =  compute_resource_networking.get("PlacementGroup") or queue_networking.get("PlacementGroup")
if chosen_pg and chosen_pg.get("Enabled"):
    return True
return False
```

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,68 @@ def is_slurm_queues_change(change):
...
+    if base_q and target_q and base_cr and target_cr:
+        if is_placement_group_managed_for_compute_resource(
+            base_q.get("Networking", {}), base_cr.get("Networking", {})
+        ) and not is_placement_group_managed_for_compute_resource(
+            target_q.get("Networking", {}), target_cr.get("Networking", {})
+        ):
+            print("return a")
+            return True
+    else:
+        if base_q.get("Networking", {}).get("PlacementGroup", {}).get("Enabled") and not target_q.get(
```
hanwen-pcluste: We can copy some utility functions from https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/utils.py#L416 to here

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,66 @@ def is_slurm_queues_change(change):
...
+def is_placement_group_managed_for_compute_resource(queue_networking, compute_resource_networking):
+    cr_pg = compute_resource_networking.get("PlacementGroup", None)
+    cr_pg_enabled = compute_resource_networking.get("PlacementGroup", {}).get("Enabled")
+    q_pg_enabled = queue_networking.get("PlacementGroup", {}).get("Enabled")
+    if (cr_pg is None and q_pg_enabled) or cr_pg_enabled:
+        return True
+    return False
+
+
+def is_managed_placement_group_deletion(change, patch):
```
hanwen-pcluste, line 181: Does the logic work for the following config changes:
Base config:
```
Queues:
  - PlacementGroup: Enabled
    ComputeResources: <No placement group specification>
```
target config:
```
Queues:
  - <No placement group specification>
    ComputeResources: <Every compute resource has PlacementGroup: Enabled>
```

We should expect no errors generated from the changes above, right?

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,66 @@ def is_slurm_queues_change(change):
...
+        ) and not is_placement_group_managed_for_compute_resource(
+            target_q.get("Networking", {}), target_cr.get("Networking", {})
+        ):
+            return True
+    else:
+        if base_q.get("Networking", {}).get("PlacementGroup", {}).get("Enabled") and not target_q.get(
+            "Networking", {}
+        ).get("PlacementGroup", {}).get("Enabled"):
+            return True
+    return False
```
hanwen-pcluste: I don't fully understand the logic. Can you add more comments in the code?

Alternatively, can you refactor the code to:
```
base_chosen_placement_group = base_cr.get("Networking", {}).get("PlacementGroup") or base_q.get("Networking", {}).get("PlacementGroup")
target_chosen_placement_group = target_cr.get("Networking", {}).get("PlacementGroup") or target_q.get("Networking", {}).get("PlacementGroup")
return is_placement_group_managed_for_compute_resource(base_chosen_placement_group) and not is_placement_group_managed_for_compute_resource(target_chosen_placement_group)
```

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Thanks, I think it makes sense.

cli/src/pcluster/config/update_policy.py
dreambeyondorange: the update process will evaluate changes at the queue level separately from the compute resource.  If we try to evaluate all the possible combinations we will end up with tons of redundant update checks.  So in this case for example, we would have a change for the queue level pg as well as a change for the compute resource level pg's.  We would then be checking every compute resource whenever there is a change to the queue level.  It seems like too much coupling and going against the pattern of the update process to be using the data from a different change.  So I decided to separate the checks and if the user removes the managed placement group from the queue level regardless of what they do at the compute resource level, they will have to stop the fleet, even if they compensate at the compute resource level.

So the logic in this case would be
1. Queue level managed group removal, stop required
2. Compute resource managed group addition, not required.
3. Result - stop required, even though it may not strictly be necessary

I think it would be good to discuss this in more depth offline

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Thanks for pointing that one out!  It's not quite the same as I am interested in the value as well is whether it exists, since it could be false, so I think I'll leave this as is, but I am going to take a closer look at those utility functions!

Actually, after some refactoring, I don't need this nesting anyway

cli/src/pcluster/config/update_policy.py
dreambeyondorange: I will clean this up!  Basically, the else block ignores the compute resource since it is not part of the specific change being evaluated

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Thanks for the tip!

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,54 @@ def is_slurm_queues_change(change):
...
+            if compute_resource.get("Name", None) == cr_name:
+                return compute_resource
+    return {}
+
+
+def is_placement_group_managed_for_compute_resource(queue_networking, compute_resource_networking):
+    chosen_pg = compute_resource_networking.get("PlacementGroup") or queue_networking.get("PlacementGroup")
+    if chosen_pg and chosen_pg.get("Enabled"):
+        return True
+    return False
```
hanwen-pcluste: This can be corrected simplified:
`return chosen_pg and chosen_pg.get("Enabled") and chosen_pg.get("Id") is None and chosen_pg.get("Name") is None`

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,54 @@ def is_slurm_queues_change(change):
...
+        queue = get_q_from_config(change, config)
+        for compute_resource in queue.get(cr_type, {}):
+            if compute_resource.get("Name", None) == cr_name:
+                return compute_resource
+    return {}
+
+
+def is_placement_group_managed_for_compute_resource(queue_networking, compute_resource_networking):
+    chosen_pg = compute_resource_networking.get("PlacementGroup") or queue_networking.get("PlacementGroup")
+    if chosen_pg and chosen_pg.get("Enabled"):
```
hanwen-pcluste: What is `Enabled` is specified together with `Id` or `Name`? Then the placement group is not managed, right?

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,54 @@ def is_slurm_queues_change(change):
     return any(path.startswith("SlurmQueues[") for path in change.path)
 
 
+def extract_type_and_name_from_path(path):
+    obj_type = re.search(".+(?=\\[)", path).group(0)
+    obj_name = re.search("(?<=\\[).+(?=\\])", path).group(0)
+    return obj_type, obj_name
+
+
+def get_q_from_config(change, config):
```
EddyMM, line 145: Let's add a sample string of the `change` to better understand how the regex parsing?/matching is happening.

cli/src/pcluster/config/update_policy.py
```diff
@@ -125,6 +134,54 @@ def is_slurm_queues_change(change):
...
+        if re.search("Queues\\[", path):
+            q_type, q_name = extract_type_and_name_from_path(path)
+    if q_type and q_name:
+        for queue in config.get("Scheduling", {}).get(q_type, {}):
+            if queue.get("Name", None) == q_name:
+                return queue
+    return {}
+
+
+def get_cr_from_config(change, config):
```
EddyMM, line 159: Let's add a sample string of the `change` to better understand how the regex parsing?/matching is happening.

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Leave for now with a decision doc to follow for future improvement

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Will update the logic here and double check for enabled + name/id combinations!

cli/src/pcluster/config/update_policy.py
dreambeyondorange: will do!

cli/src/pcluster/config/update_policy.py
dreambeyondorange: will do!

cli/src/pcluster/config/update_policy.py
dreambeyondorange: will update, good catch!

cli/src/pcluster/config/update_policy.py
dreambeyondorange: I want this function to always return a boolean, so I will keep the if but add the check for the Name/Id

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -807,3 +814,415 @@ def test_shared_storage_update_policy_condition_checker(
...
+                is_list=False,
+            ),
+            True,
+            True,
+            False,
+            "All compute nodes must be stopped for a managed placement group deletion",
+        ),
+    ],
+)
+def test_condition_checker_managed_placement_group(
```
EddyMM, line 1279: Can we test a scenario where the `base_config` has more than one queue, and the `target_config` includes `PlacementGroup` changes in two or more of the queues?

dreambeyondorange: I can add a unit test to tackle this scenario, but I'd like to avoid the explosion of possibilities when adding more queues and more changes.  What is the main function you want to see tested here?  The queue and compute resource selection process?

EddyMM: > The queue and compute resource selection process?

Exactly 👍.

hanwen-pcluste (dismissed):
LGTM!

cli/tests/pcluster/config/test_update_policy.py
dreambeyondorange: sounds good!  Added!

EddyMM (approved):

PR #4395, opened 2022-10-03
[Release 3.2] Fix benchmark result check in case of osu_bibw test

francesco-giordano:
### Description of changes
* Fix benchmark result check in case of osu_bibw test

### DIFF
--- a/tests/integration-tests/tests/efa/test_efa.py
+++ b/tests/integration-tests/tests/efa/test_efa.py
@@ -238,26 +238,34 @@ def _check_osu_benchmarks_results(test_datadir, instance, mpi_version, benchmark
     logging.info(output)
     # Check avg latency for all packet sizes
     failures = 0
-    for packet_size, latency in re.findall(r"(\d+)\s+(\d+)\.", output):
+    for packet_size, value in re.findall(r"(\d+)\s+(\d+)\.", output):
         with open(
             str(test_datadir / "osu_benchmarks" / "results" / instance / mpi_version / benchmark_name), encoding="utf-8"
         ) as result:
             previous_result = re.search(rf"{packet_size}\s+(\d+)\.", result.read()).group(1)
 
-            # Use a tolerance of 10us for 2 digits values.
-            # For 3+ digits values use a 20% tolerance, except for the higher-variance latency benchmark.
-            if len(previous_result) <= 2:
-                accepted_tolerance = 10
+            if benchmark_name == "osu_bibw":
+                # Invert logic because osu_bibw is in MB/s
+                tolerated_value = float(previous_result) - (float(previous_result) * 0.2)
+                is_failure = int(value) < tolerated_value
             else:
-                multiplier = 0.3 if benchmark_name == "osu_latency" else 0.2
-                accepted_tolerance = float(previous_result) * multiplier
-            tolerated_latency = float(previous_result) + accepted_tolerance
+                # Use a tolerance of 10us for 2 digits values.
+                # For 3+ digits values use a 20% tolerance, except for the higher-variance latency benchmark.
+                if len(previous_result) <= 2:
+                    accepted_tolerance = 10
+                else:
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -238,31 +238,49 @@ def _check_osu_benchmarks_results(test_datadir, instance, mpi_version, benchmark
...
+                # Use a tolerance of 10us for 2 digits values.
+                # For 3+ digits values use a 20% tolerance, except for the higher-variance latency benchmark.
+                if len(previous_result) <= 2:
+                    accepted_tolerance = 10
+                else:
+                    multiplier = 0.3 if benchmark_name == "osu_latency" else 0.2
+                    accepted_tolerance = float(previous_result) * multiplier
+                tolerated_value = float(previous_result) + accepted_tolerance
+
+                message = (
```
enrico-usai: this might be put in common.

tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -238,31 +238,49 @@ def _check_osu_benchmarks_results(test_datadir, instance, mpi_version, benchmark
...
 
-            message = (
-                f"{mpi_version} - {benchmark_name} - packet size {packet_size}: "
-                f"tolerated: {tolerated_latency}, current: {latency}"
-            )
-            if int(latency) > tolerated_latency:
-                failures = failures + 1
-                logging.error(message)
+                # Invert condition when benchmark_name is osu_bibw
+                if int(value) > tolerated_value:
```
enrico-usai: here you can define a variable like:
```
is_failure = int(value) > tolerated_value
```
and use it outside the if. In the other branch you can have the opposite condition.

This will avoid to duplicate message and logging part

tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -238,31 +238,49 @@ def _check_osu_benchmarks_results(test_datadir, instance, mpi_version, benchmark
...
+                    failures = failures + 1
+                    logging.error(message)
+                else:
+                    logging.info(message)
             else:
-                logging.info(message)
+                # Invert logic becasue osu_bibw is in MB/s
+                if benchmark_name == "osu_bibw":
```
enrico-usai: Why another additional if inside? This is always true

tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -238,31 +238,49 @@ def _check_osu_benchmarks_results(test_datadir, instance, mpi_version, benchmark
...
+                    f"{mpi_version} - {benchmark_name} - packet size {packet_size}: "
+                    f"tolerated: {tolerated_value}, current: {value}"
+                )
 
+                # Invert condition when benchmark_name is osu_bibw
+                if int(value) < tolerated_value:
+                    failures = failures + 1
+                    logging.error(message)
+                else:
+                    logging.info(message)
```
enrico-usai: You can avoid the logging duplication by doing:
```python
 log_level = logging.ERROR if is_failure else logging.INFO
 logger.log(log_level, message)
```

francesco-giordano: Keeping the previous implementation since the number of line is the same and the code seems more readable.

tests/integration-tests/tests/efa/test_efa.py
francesco-giordano: Done

tests/integration-tests/tests/efa/test_efa.py
francesco-giordano: Done

tests/integration-tests/tests/efa/test_efa.py
francesco-giordano: Done

enrico-usai (dismissed):

enrico-usai (approved):

gmarciani (approved):

PR #4402, opened 2022-10-05
[DFSM] Make security group rules attached to file systems have a DeletionPolicy that is consistent with the DeletionPolicy set for the file systems

gmarciani:
### Description of changes
Make security group rules attached to file systems have a DeletionPolicy that is consistent with the DeletionPolicy set for the file systems.

#### Assumptions
The SG attached to a EFS/FSx file system must have the following rules:
1. allow traffic within itself
2. allow traffic with the head node
3. allow traffic with the compute node

#### User Experience without this change
If the user wants to _retain_ the file system, then the following resources are retained:
1. file system
2. SG attached to the file system

All the above rules will be deleted. 
That's correct for rules 2 and 3 because the lifecycle of the file system is now decoupled from the cluster. 
That's not correct for rule 1, because its lifecycle should be coupled with the file system lifecycle.

#### User Experience with this change
If the user wants to _retain_ the file system, then the following resources are retained:
1. file system
2. SG attached to the file system
3. Rule 1

Rules 2 and 3 will be deleted (as it currently is).

### Tests
1. Unit tests

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -516,7 +516,7 @@ class ClusterCdkStack(Stack):
                 )
 
     def _allow_all_ingress(self, description, source_security_group_id, group_id):
-        ec2.CfnSecurityGroupIngress(
+        return ec2.CfnSecurityGroupIngress(
             self,
             description,
             ip_protocol="-1",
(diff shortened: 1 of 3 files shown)


enrico-usai:
I'd like to have a more clear description of the reason why we need this change. 

Is it a user-facing issue? is it something impacting some specific corner case? Is it a new feature?

enrico-usai (approved):
Thanks for the detailed description!

PR #4403, opened 2022-10-05
[DFSM] Make integration test test_overwrite_sg consistent with changes in the security group attached to file systems.

gmarciani:
### Description of changes
Make integration test test_overwrite_sg consistent with changes in the security group attached to file systems.
The test assumed that the SG attached to EFS/FSx file systems is equal to the one specified in `HeadNode.Networking.SecurityGroups`. This is not true anymore because with DFSM we introduce a dedicated SG for the file system.

### Tests
1. Locally launched the test: test_securuty_groups::test_overwrite_sg

### DIFF
--- a/tests/integration-tests/tests/networking/test_security_groups.py
+++ b/tests/integration-tests/tests/networking/test_security_groups.py
@@ -72,9 +72,19 @@ def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_
         fsx_security_groups = ec2_client.describe_network_interfaces(NetworkInterfaceIds=[network_interface_id])[
             "NetworkInterfaces"
         ][0]["Groups"]
-        logging.info("Asserting the network interface of FSx has and only has the custom security group")
-        assert_that(fsx_security_groups[0]["GroupId"]).is_equal_to(custom_security_group_id)
+        logging.info(
+            "Asserting the network interface of FSx has only the dedicated security group "
+            "allowing all traffic to itself and to the custom security group"
+        )
         assert_that(fsx_security_groups).is_length(1)
+        fsx_security_group = fsx_security_groups[0]
+        fsx_security_group_id = fsx_security_group["GroupId"]
+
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -72,9 +72,25 @@ def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_
...
         assert_that(fsx_security_groups).is_length(1)
+        fsx_security_group = fsx_security_groups[0]
+        fsx_security_group_id = fsx_security_group["GroupId"]
+        # We expect the FSx Security Group to have exactly 4 rules:
+        #  * 2 rules (ingress and egress) allowing traffic to/from the FSx Security Group itself
+        #  * 2 rules (ingress and egress) allowing traffic to/from the custom Security Group
+        _assert_security_group_rules(
+            ec2_client,
+            fsx_security_group_id,
+            [
```
enrico-usai: This loop might be moved into the `_assert_security_group_rules` function and you can just pass the two security groups to loop for.

The `_assert_security_group_rules` test is very specific to egress so we can move the entire logic there and rename it to `_assert_egress_security_group_rules`.

enrico-usai: The comment might be moved as help of the function as well.

gmarciani: > This loop might be moved into the _assert_security_group_rules function and you can just pass the two security groups to loop for.

I agree, thank you.

> The _assert_security_group_rules test is very specific to egress so we can move the entire logic there and rename it to _assert_egress_security_group_rules.

The assertion covers both egress (IsEgress: True) and ingress (IsEgress: False), so I'd keep the current name without moving the entire logic within it.

enrico-usai (dismissed):

enrico-usai (approved):

PR #4405, opened 2022-10-06
[DFSM] Use UpdatePolicy.IGNORED for Ebs/Efs/FSxLustreSettings.

gmarciani:
### Description of changes
Use `UpdatePolicy.IGNORED` for `Ebs/Efs/FSxLustreSettings`.

We want this update policy because the setting `Ebs/Efs/FSxLustreSettings.DeletionPolicy` supports the updates.
With this change we will support the following update, that we should provide to our users:

From:

```
  - MountDir: /opt/shared/efs/managed/1
    Name: shared-efs-managed-1
    StorageType: Efs
```

to:

```
  - MountDir: /opt/shared/efs/managed/1
    Name: shared-efs-managed-1
    StorageType: Efs
    EfsSettings:
      DeletionPolicy: Retain|Delete
```

### Tests
1. Manually tested the above scenario and that update containing not updatable settings is rejected.
2. Unit tests

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -482,9 +482,9 @@ class SharedStorageSchema(BaseSchema):
         validate=validate.OneOf(["Ebs", FSX_LUSTRE, FSX_OPENZFS, FSX_ONTAP, "Efs"]),
         metadata={"update_policy": UpdatePolicy.UNSUPPORTED},
     )
-    ebs_settings = fields.Nested(EbsSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    efs_settings = fields.Nested(EfsSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    fsx_lustre_settings = fields.Nested(FsxLustreSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
+    ebs_settings = fields.Nested(EbsSettingsSchema, metadata={"update_policy": UpdatePolicy.IGNORED})
+    efs_settings = fields.Nested(EfsSettingsSchema, metadata={"update_policy": UpdatePolicy.IGNORED})
+    fsx_lustre_settings = fields.Nested(FsxLustreSettingsSchema, metadata={"update_policy": UpdatePolicy.IGNORED})
     fsx_open_zfs_settings = fields.Nested(
         FsxOpenZfsSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED}
     )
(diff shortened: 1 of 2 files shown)


chenwany (dismissed):
LGTM!

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -482,9 +482,9 @@ class SharedStorageSchema(BaseSchema):
         validate=validate.OneOf(["Ebs", FSX_LUSTRE, FSX_OPENZFS, FSX_ONTAP, "Efs"]),
         metadata={"update_policy": UpdatePolicy.UNSUPPORTED},
     )
-    ebs_settings = fields.Nested(EbsSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    efs_settings = fields.Nested(EfsSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    fsx_lustre_settings = fields.Nested(FsxLustreSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
+    ebs_settings = fields.Nested(EbsSettingsSchema, metadata={"update_policy": UpdatePolicy.IGNORED})
```
enrico-usai, line 485: It would be good to add a unit test to verify that this new update path is supported. I mean from:
```
  - MountDir: /opt/shared/efs/managed/1
    Name: shared-efs-managed-1
    StorageType: Efs
```
to:
```
  - MountDir: /opt/shared/efs/managed/1
    Name: shared-efs-managed-1
    StorageType: Efs
    EfsSettings:
      DeletionPolicy: Retain|Delete
```

gmarciani: Agree. Added

cli/tests/pcluster/config/test_config_patch.py
```diff
@@ -505,6 +505,43 @@ def _test_more_target_sections(base_conf, target_conf):
     target_storage["EbsSettings"]["Iops"] = 20
     target_storage["EbsSettings"]["VolumeType"] = "gp2"
 
+    # Check the update of Ebs,Efs,FsxLustreSettings
+    # with both updatable and not updatable keys
+    for storage_type in ("Ebs", "Efs", "FsxLustre"):
```
enrico-usai: Thanks for adding these unit tests!

The `_test_more_target_sections` function was used to verify that a generic case on which you were adding a section is working as expected.

I think we can move these new tests outside of `_test_more_target_sections` to keep the test more self contained and you can have a new generic `_test_storages`.

In this way we can continue to extend storage specific unit tests in this new function. What do you think?

gmarciani: Done

enrico-usai (approved):

PR #4419, opened 2022-10-10
Support requiring IMDSv2 in cluster and build image config

ermanno:
#### Notes
This patch implements support for specifying a section like the following:

```
Imds:
  RequireImdsV2: true
```

at the top level of the cluster config file, or in the `Build` section of the build image file.  If omitted, it defaults to `false` but we show the customer a warning clarifying that we will change the default in the future.

#### Tests
* Unite tests.
* Manual tests: created clusters with Slurm, AWS Batch and BYOS schedulers, verified both the default case and the case when specifying the parameter to `true`, checking the metadata of the EC2 cluster instances (head and compute nodes) and validating that the `http_token` was `optional` and `required`, respectively. Did the same with the Build instance launched in the course of the `build-image` command.
* Changes to integration tests:
  * If not specified in the cluster or build image config for the integration test, we set the `RequireImdsV2` to `true`
  * We change the `test_create.py::test_create_imds_secured` to take an additional `require_imds_v2` parameter, and launch it with `true` and `false` to test the positive and negative case, asserting that all the nodes of the cluster have the desired status (`"required"` or `"optional"`, respectively).
  * We change the `test_build_image` and `test_kernel4_build_image_run_cluster` in `test_createami.py` to assert that the build instance has the required IMDSv2 setting.
* Run integration tests with the following configuration:
```
---
scheduler-plugins:
  slurm_plugin:
    scheduler-definition: "../../scheduler_plugins/slurm/utils/upload_artifacts.sh"
    scheduler-commands: "tests.common.schedulers_common.SlurmCommands"
    requires-sudo: true
test-suites:
  schedulers:
    test_slurm.py::test_slurm:
      dimensions:
        - regions: ["eu-central-1"]
          oss: ["alinux2"]
          schedulers: ["slurm_plugin"]
  createami:
    test_createami.py::test_build_image:
      dimensions:
        - regions: ["eu-central-1"]
          oss: ["alinux2"]
    test_createami.py::test_kernel4_build_image_run_cluster:
      dimensions:
        - regions: ["eu-central-1"]
          schedulers: ["slurm"]
          oss: ["alinux2"]
  create:
    test_create.py::test_create_imds_secured:
      dimensions:
        - regions: [ "eu-central-1" ]
          oss: ["alinux2"]
          schedulers: [ "slurm" ]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
   to support storage retention on deletion.
 - Enable server-side encryption for the EcrImageBuilder SNS topic created when deploying ParallelCluster API and used to notify on docker image build events.
 - Add support for on-demand capacity reservations.
+- Add support for requiring IMDSv2 in cluster configurations.
 - Add support for Slurm Accounting.
 
 **CHANGES**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ CHANGELOG
   to support storage retention on deletion.
 - Enable server-side encryption for the EcrImageBuilder SNS topic created when deploying ParallelCluster API and used to notify on docker image build events.
 - Add support for on-demand capacity reservations.
-- Add support for requiring IMDSv2 in cluster configurations.
+- Add support for requiring IMDSv2 in cluster and build image configurations via the `Imds > RequireImdsV2` property.
 - Add support for Slurm Accounting.
 
 **CHANGES**
--- a/cli/src/pcluster/config/common.py
+++ b/cli/src/pcluster/config/common.py
@@ -276,6 +276,19 @@ class BaseDevSettings(Resource):
             self._register_validator(UrlValidator, url=self.aws_batch_cli_package)
 
 
+class Imds(Resource):
+    """
+    Represent the IMDS configuration shared between the ImageBuilder and Cluster.
+
+    It represents the Imds element that can be either at top level in the cluster config file,
+    or in the Build section of the build image config file.
+    """
+
+    def __init__(self, require_imds_v2: bool = None, **kwargs):
+        super().__init__(**kwargs)
+        self.require_imds_v2 = Resource.init_param(require_imds_v2, default=False)
+
+
 # ------------ Common attributes class between ImageBuilder an Cluster models ----------- #
 
 
--- a/cli/src/pcluster/templates/awsbatch_builder.py
+++ b/cli/src/pcluster/templates/awsbatch_builder.py
@@ -151,6 +151,19 @@ class AwsBatchConstruct(Construct):
         self._code_build_notification_rule = self._add_code_build_notification_rule()
         self._manage_docker_images_custom_resource.add_depends_on(self._code_build_notification_rule)
 
+    def _launch_template(self):
+        launch_template = ec2.CfnLaunchTemplate(
+            self.stack_scope,
+            "PclusterComputeEnvironmentLaunchTemplate",
+            launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens="required")
+            ),
+        )
+        return batch.CfnComputeEnvironment.LaunchTemplateSpecificationProperty(
+            launch_template_id=launch_template.ref,
+            version=launch_template.attr_latest_version_number,
+        )
+
     def _add_compute_env(self):
         return batch.CfnComputeEnvironment(
             self.stack_scope,
@@ -169,6 +182,7 @@ class AwsBatchConstruct(Construct):
                 instance_role=self._iam_instance_profile.ref,
                 bid_percentage=self.compute_resource.spot_bid_percentage,
                 spot_iam_fleet_role=self._spot_iam_fleet_role.attr_arn if self._spot_iam_fleet_role else None,
+                launch_template=self._launch_template() if self.config.imds.require_imds_v2 else None,
                 tags={
                     **get_default_instance_tags(
                         self.stack_name,
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -882,6 +882,9 @@ class ClusterCdkStack(Stack):
                 iam_instance_profile=ec2.CfnLaunchTemplate.IamInstanceProfileProperty(
                     name=self._head_node_instance_profile
                 ),
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens="required")
+                if self.config.imds.require_imds_v2
+                else None,
                 user_data=Fn.base64(
                     Fn.sub(
                         get_user_data_content("../resources/head_node/user_data.sh"),
@@ -1484,6 +1487,9 @@ class ComputeFleetConstruct(Construct):
                 instance_market_options=instance_market_options,
                 instance_initiated_shutdown_behavior="terminate",
                 capacity_reservation_specification=capacity_reservation_specification,
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens="required")
+                if self._config.imds.require_imds_v2
+                else None,
                 user_data=Fn.base64(
                     Fn.sub(
                         get_user_data_content("../resources/compute_node/user_data.sh"),
--- a/cli/src/pcluster/validators/imagebuilder_validators.py
+++ b/cli/src/pcluster/validators/imagebuilder_validators.py
@@ -54,3 +54,18 @@ class SecurityGroupsAndSubnetValidator(Validator):
                     "Subnet id {0} is specified, security groups is required.".format(subnet_id),
                     FailureLevel.ERROR,
                 )
+
+
+class RequireImdsV2Validator(Validator):
+    """IMDSv2 enforcement validator."""
+
+    def _validate(self, require_imds_v2):
+        if not require_imds_v2:
+            self._add_failure(
+                "The current build image configuration does not disable IMDSv1, "
+                "which we plan to disable by default in future versions. If you do "
+                "not explicitly need to use IMDSv1 enforce IMDSv2 usage by setting the "
+                "RequireImdsV2 configuration parameter to true (see documentation at: "
+                "https://docs.aws.amazon.com/parallelcluster/latest/ug/Build-v3.html)",
+                level=FailureLevel.INFO,
+            )
--- a/cli/tests/pcluster/models/test_imagebuilder.py
+++ b/cli/tests/pcluster/models/test_imagebuilder.py
@@ -48,7 +48,11 @@ from tests.pcluster.test_utils import FAKE_NAME
                             {"key": "date", "value": "2022.1.1"},
                         ],
                     },
-                    "build": {"parent_image": "ami-0185634c5a8a37250", "instance_type": "c5.xlarge"},
+                    "build": {
+                        "imds": {"require_imds_v2": True},
+                        "parent_image": "ami-0185634c5a8a37250",
+                        "instance_type": "c5.xlarge",
+                    },
                 }
             },
             {
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -600,6 +600,9 @@ def inject_additional_config_settings(  # noqa: C901
     with open(cluster_config, encoding="utf-8") as conf_file:
         config_content = yaml.safe_load(conf_file)
 
+    if not dict_has_nested_key(config_content, ("Imds", "RequireImdsV2")):
+        dict_add_nested_key(config_content, True, ("Imds", "RequireImdsV2"))
+
     if request.config.getoption("custom_chef_cookbook") and not dict_has_nested_key(
         config_content, ("DevSettings", "Cookbook", "ChefCookbook")
     ):
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -574,6 +574,9 @@ def inject_additional_image_configs_settings(image_config, request):
     with open(image_config, encoding="utf-8") as conf_file:
         config_content = yaml.load(conf_file, Loader=yaml.SafeLoader)
 
+    if not dict_has_nested_key(config_content, ("Build", "Imds", "RequireImdsV2")):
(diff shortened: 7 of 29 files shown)


cli/src/pcluster/config/common.py
```diff
@@ -276,6 +276,17 @@ def _register_validators(self):
...
 
+class Imds(Resource):
+    """Represent the IMDS configuration shared between the ImageBuilder and Cluster.
+
+    It represents the Imds element that can be either at top level in the cluster config file,
+    or in the Build section of the build image config file."""
+
+    def __init__(self, require_imds_v2: bool = None, **kwargs):
+        super().__init__(**kwargs)
+        self.require_imds_v2 = Resource.init_param(require_imds_v2, default=False)
```
gmarciani, line 289: Maybe I'm missing something, but why the default should be False if we would like to enforce the usage of Imds v2?

cli/src/pcluster/templates/awsbatch_builder.py
```diff
@@ -151,6 +151,19 @@ def _add_resources(self):
         self._code_build_notification_rule = self._add_code_build_notification_rule()
         self._manage_docker_images_custom_resource.add_depends_on(self._code_build_notification_rule)
 
+    def _launch_template(self):
+        launch_template = ec2.CfnLaunchTemplate(
+            self.stack_scope,
+            "PclusterComputeEnvironmentLaunchTemplate",
+            launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens="required")
```
gmarciani, line 159: Unit tests should cover the setting of this metadata.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1484,6 +1487,9 @@ def _add_compute_resource_launch_template(
                 instance_market_options=instance_market_options,
                 instance_initiated_shutdown_behavior="terminate",
                 capacity_reservation_specification=capacity_reservation_specification,
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens="required")
```
gmarciani, line 1490: HEre for compute and above for head node: unit tests should cover the setting of this metadata.

cli/src/pcluster/config/common.py
ermanno: The idea was that we will set it to `false` because we have customers that currently do IMDSv1 calls in the post install scripts, etc., so we don't want to require `IMDSv2` now, however we want to notify them that they will need to change it because we will switch the default sometime down the line (see [Notes section of the PR description](https://github.com/aws/aws-parallelcluster/pull/4419#issue-1403098826)).

cli/src/pcluster/templates/awsbatch_builder.py
ermanno: At this moment afaik we do these tests by asserting on the produced Cfn we generate with CDK (see `test_required_imds_v2`)

cli/src/pcluster/templates/cluster_stack.py
ermanno: At this moment afaik we do these tests by asserting on the produced Cfn we generate with CDK (see `test_required_imds_v2`)

gmarciani:
In this change we are introducing a new configuration property `RequireImdsV2` that seems something temporary because (i) in the next releases IMDSv2 will be mandatory and (ii) we are embedding a version number (v2) in a property name.

This new property is going to be removed sooner or later. Afaik in pcluster we've always tried to avoid temporary customer-facing properties.

What about adding this property in DevSettings rather than at the root level?
Advantages of this approach would be: (i) pcluster team can use the IMDSv2 enforcement (ii) customers interested in using the enforcement before the release that will enforce it by default can use it (iii) we can remove properties from DevSettings whenever we want since DevSettings is an experimental section.

ermanno:
> In this change we are introducing a new configuration property `RequireImdsV2` that seems something temporary because (i) in the next releases IMDSv2 will be mandatory and (ii) we are embedding a version number (v2) in a property name.
> 
> This new property is going to be removed sooner or later. Afaik in pcluster we've always tried to avoid temporary customer-facing properties.

Afaik this is **not** going to be a temporary property, in next releases we will change the *default*, my understanding is that we are not going to remove the property itself. `RequireImdsV2` has been chosen because it is in line with what other AWS products such as [CDK](https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_autoscaling/AutoScalingGroup.html) are using.

> What about adding this property in DevSettings rather than at the root level? Advantages of this approach would be: (i) pcluster team can use the IMDSv2 enforcement (ii) customers interested in using the enforcement before the release that will enforce it by default can use it (iii) we can remove properties from DevSettings whenever we want since DevSettings is an experimental section.

This was the first solution we considered, however afaik @demartinofra proposed to expose it outside `DevSettings` in a structure similar to the one that we already have for the `HeadNode` section (`Imds`), this would allow us more flexibility in the future if we want to add more granularity (i.e. allow to set this for individual compute resources, etc.). One other issue I see with the `DevSettings` approach is that my understanding is that those properties are not publicly documented.

gmarciani:
Let's consider my comment about putting the property in DevSettings as resolved.
We discussed about it offline and agreed on the fact that such property is not going to be temporary, so it makes sense to have it customer-facing at the root level.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
   to support storage retention on deletion.
 - Enable server-side encryption for the EcrImageBuilder SNS topic created when deploying ParallelCluster API and used to notify on docker image build events.
 - Add support for on-demand capacity reservations.
+- Add support for requiring IMDSv2 in cluster and build image configurations.
```
gmarciani: I think we should mention the specific property: `Imds > RequireImdsV2 `

cli/tests/pcluster/models/test_imagebuilder.py
```diff
@@ -48,7 +48,11 @@
                             {"key": "date", "value": "2022.1.1"},
                         ],
                     },
-                    "build": {"parent_image": "ami-0185634c5a8a37250", "instance_type": "c5.xlarge"},
+                    "build": {
+                        "imds": {"require_imds_v2": True},
```
gmarciani, line 52: Do we cover the IMDSv1 scenario somewhere else?
I think it would be better to cover both cases: IMDSv1 and IMDSv2, especially because with this solution we are going to cover the non default scenario.
I know that currently these unit tests do not support templatization of cluster config, but we should add it sooner or later. May you please keep track of this improvement in our backlog?

CHANGELOG.md
ermanno: I am okay with this change, however note that we didn't do this for the immediately preceding item `Add support for on-demand capacity reservations.`

cli/tests/pcluster/models/test_imagebuilder.py
ermanno: `test_imagebuilder_kms_key_id_encrypted_validator_and_ami_volume_size_validator` is for the `build-image` path, and tests the key id encrypted and AMI volume size validations, the reason we added this bit of configuration was to avoid the warning for IMDSv2 since testing that behaviour was outside the scope of the test.

CHANGELOG.md
ermanno: Changed in the last revision.

cli/src/pcluster/validators/imagebuilder_validators.py
```diff
@@ -54,3 +54,18 @@ def _validate(self, security_group_ids: list, subnet_id: str):
...
+
+    def _validate(self, require_imds_v2):
+        if not require_imds_v2:
+            self._add_failure(
+                "The current build image configuration does not disable IMDSv1, "
+                "which we plan to disable by default in future versions. If you do "
+                "not explicitly need to use IMDSv1 enforce IMDSv2 usage by setting the "
+                "RequireImdsV2 configuration parameter to true (see documentation at: "
+                "https://docs.aws.amazon.com/parallelcluster/latest/ug/Build-v3.html)",
+                level=FailureLevel.WARNING,
```
gmarciani: Here and at cluster config. I suppose that the warning message has been already approved at product level. Nevertheless, I'm a bit concerned about it because most of the customers will always see the warning message at every creation/update. A warning message expresses that something not blocking is going wrong.

tests/integration-tests/conftest.py
```diff
@@ -574,6 +574,9 @@ def inject_additional_image_configs_settings(image_config, request):
     with open(image_config, encoding="utf-8") as conf_file:
         config_content = yaml.load(conf_file, Loader=yaml.SafeLoader)
 
+    if not dict_has_nested_key(config_content, ("Build", "Imds", "RequireImdsV2")):
```
gmarciani, line 577: In this way we will have all our integration tests covering the case with IMDSv2 even if we already know that most of our customers will still be using IMDSv1. This is a risk because we are testing the non default scenario leaving uncovered the default one.

I'd propose to have the other way around: most of the tests covering the default scenario that is the most widely adopted by customers and a subset of them with the non default one; then switch the two when we will make IMDSv2 the default.

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -420,6 +430,29 @@ def test_build_image_custom_components(
...
+
+    for reservations in describe_response.get("Reservations"):
+        for instance in reservations.get("Instances"):
+            instance_id = instance.get("InstanceId")
+            instance_name = instance.get("Tags").get("Name")
+            imds_v2_status = instance.get("MetadataOptions").get("HttpTokens")
+
+            logging.info(f"Instance {instance_id} ({instance_name}) has IMDSv2 {imds_v2_status}")
+
+            assert_that(imds_v2_status).is_equal_to(status)
```
gmarciani: This assertion logic is duplicated. what about defining it in a single common place and reuse it?

tests/integration-tests/tests/createami/test_createami/test_build_image/image.config.yaml
```diff
@@ -7,6 +7,8 @@ Image:
         Encrypted: True
 
 Build:
+    Imds:
+        RequireImdsV2: true
```
gmarciani, line 11: Just for my better understanding: since we are enforcing IMDSv2 at conftest level, why should we explicitly set this to true here?

tests/integration-tests/conftest.py
ermanno: I am relatively agnostic with regards to this, with @demartinofra we decided on the other way around, but I don't have strong feelings on this. 

However, note that not all the tests will run with IMDSv2, we do have tests that use IMDSv1 (the code above sets IMDSv2 by default only if it is not otherwise specified in the test configuration template).

I believe the point is:
- We have integration tests covering both the positive and negative case.
- The other integration tests don't have to do with IMDSv2 and therefore it is acceptable to set our preferred default *provided we have some tests exercising both cases* (see previous point).

Another point I wanted to raise was that, as explained [in the description of the PR](https://github.com/aws/aws-parallelcluster/pull/4419#issue-1403098826), with @demartinofra we decided to add assertions to existing tests rather than writing dedicated ones, in light of the time taken by the overall test suite (in particular by the `build-image` tests).

ermanno: However, note that not all the tests will run with IMDSv2, we do have tests than use IMDSv1 (the code above sets IMDSv2 by default only if it is not otherwise specified in the test configuration template).

tests/integration-tests/tests/createami/test_createami/test_build_image/image.config.yaml
ermanno: Correct, it is not needed, I did it to be explicit that we were testing the positive case.

tests/integration-tests/tests/createami/test_createami.py
ermanno: Yes, in this case I had a slightly different logging message because the ImageBuilder instances have the "Test"/"Build" embedded in the name, but it can be refactored to extract this common logic. Is this a blocking issue, or can it be done in a follow up PR?

cli/src/pcluster/validators/imagebuilder_validators.py
ermanno: This is what we want to express. We don't want to block their operation, but we want to *warn* the customer that we intend to change this default in the future and they should fix their scripts and migrate to IMDSv2. I am not unsympathetic to the default/warning point, but this was a point I already raised and discussed at length with @demartinofra, the SDM and the rest of the team., and this was the solution that was agreed upon.

gmarciani: Ok, if there is already an agreement on that, let's proceed this way.

**My non blocking opinion**
I confirm my concern on this: the risk is that we pollute our response with too many messages. Example: in this release we already added additional info messages related to storage.
As an alternative, we could use the changelog to advise for upcoming disruptive deprecation.

tests/integration-tests/conftest.py
gmarciani: Ok, I got the point. Thanks

tests/integration-tests/tests/createami/test_createami.py
gmarciani: I suggest to remove the code duplication in this PR to not forget about it.
In any case this is not a blocking comment.

gmarciani (dismissed):

tests/integration-tests/conftest.py
demartinofra: The main reason why I was in favor of running all tests with IMDSv2 is that we guarantee that no pcluster functionality ever relies on IMDSv1 being enabled. Also note that when IMDSv1 is enabled also IMDSv2 works, they are not mutually exclusive. 
At the same time having a single test that checks the default setting should give us enough confidence that the IMDSv1 works as expected.

tests/integration-tests/tests/createami/test_createami.py
ermanno: Agreed, sending change in this PR.

cli/src/pcluster/validators/imagebuilder_validators.py
ermanno: After sync up with @demartinofra  and @gmarciani offline, we agreed to make this an `INFO`

gmarciani (approved):

PR #4422, opened 2022-10-10
[Parsing] Make YAML parser fail on duplicate keys.

gmarciani:
### Description of changes
1. Make YAML parser fail on duplicate keys.
2. Move YAML parsing logic into a single place to avoid code duplication.
3. Added unit test covering YAML parsing with and without duplicate keys.

With this change, a cluster configuraiton like the following one:

```
...
Scheduling:
  Scheduler: slurm
  SlurmSettings:
    QueueUpdateStrategy: DRAIN
  SlurmSettings:
    QueueUpdateStrategy: TERMINATE
...
```

makes the pcluster create-cluster and update-cluster fail with message:

```
{
  "message": "Bad Request: Configuration must be a valid YAML document. Duplicate key found: SlurmSettings\n  in \"<unicode string>\", line 17, column 3:\n      SlurmSettings:\n      ^"
}

```

In `pcluster.log` there will be the parsing failure detail:

```
2022-10-11 13:31:59,106 - ERROR - common.py:67:parse_config() - Failed when parsing the configuration due to invalid YAML document: Duplicate key found: SlurmSettings
  in "<unicode string>", line 17, column 3:
      SlurmSettings:
      ^
```

Without this change, the above configuration would be accepted with `SlurmSettings.QueueUpdateStrategy=TERMINATE` that is unexpected.

### Tests
1. Unit tests
4. Manually tested that a cluster config with [without] duplicate keys in cluster config file makes the command fail [succeed]

### DIFF
--- a/cli/src/pcluster/utils.py
+++ b/cli/src/pcluster/utils.py
@@ -29,6 +29,9 @@ import dateutil.parser
 import pkg_resources
 import yaml
 from pkg_resources import packaging
+from yaml import SafeLoader
+from yaml.constructor import ConstructorError
+from yaml.resolver import BaseResolver
 
 from pcluster.aws.common import get_region
 from pcluster.constants import SUPPORTED_OSES_FOR_ARCHITECTURE, SUPPORTED_OSES_FOR_SCHEDULER
@@ -312,7 +315,7 @@ def get_url_scheme(url):
 def load_yaml_dict(file_path):
     """Read the content of a yaml file."""
     with open(file_path, encoding="utf-8") as conf_file:
-        yaml_content = yaml.safe_load(conf_file)
+        yaml_content = yaml_load(conf_file)
 
     # TODO use from cfn_flip import load_yaml
     return yaml_content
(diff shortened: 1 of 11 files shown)


cli/src/pcluster/utils.py
```diff
@@ -333,3 +336,24 @@ def get_attr(obj, attributes, default=None):
...
+        tag=BaseResolver.DEFAULT_MAPPING_TAG, constructor=yaml_no_duplicates_constructor, Loader=SafeLoader
+    )
+    return yaml.safe_load(s)
+
+
+def yaml_no_duplicates_constructor(loader, node, deep=False):
+    """Check for duplicate keys."""
+
+    mapping = {}
+    for key_node, value_node in node.value:
```
enrico-usai, line 351: By using this approach we're not able to surface the "Duplicate key found" message to the user (only in the log file).

What about using instead a `@validates_schema` defined in the `BaseSchema` class?
This kind of validation errors should surface to the user.

gmarciani: Duplicate keys are invalid, as per [YAML 1.2 specs](https://yaml.org/spec/1.2.2/), so this is something not related to pcluster schema but to YAML itself. 

Currently, we handle parsing errors at the PyYAML level, not in pcluster schema so currently all syntax errors are not surfaced to the user but are stored in the log.

I agree with you that it would be better to surface the error so my proposal is to report the parsing error message within the surfaced Bad Request message, but to keep the duplicate detection at the PyYAML level, i.e.:

```
{
  "message": "Bad Request: Configuration must be a valid YAML document. DETAILED ERROR HERE"
}
```

What do you think?

enrico-usai: I like the idea, thanks!

enrico-usai (dismissed):

cli/src/pcluster/utils.py
gmarciani: Clarified offline. We'll go with the proposal of surfacing the error message from the parser but keeping the logic at PyYAML level

gmarciani:
> I like the idea but unfortunately the path is not accessible when running the code from API. Let's keep the code as is. thanks!

Clarified offline: there was a misunderstanding here. My proposal is to include the error message from the YAML parser into the surfaced error message, not to include any cluster config path there.

enrico-usai (dismissed):

enrico-usai (approved):

PR #4424, opened 2022-10-11
Add parameters for CIDR blocks in Slurm accounting database CFN template

jdeamicis:
### Description of changes
* Add parameters for CIDR blocks in Slurm accounting database CFN template
* Set rules to cause the stack creation to fail if VPC or subnets creation is requested and the default CIDR blocks are not modified.

### Tests
* Ran manual tests with the new template about:
    * [x] failing assertions if default CIDR blocks are not modified when requesting creation of:
       * [x] VPC
       * [x] subnets
    * [x] usage of provided CIDR blocks when creating:
       * [x] VPC
       * [x] subnets

### Checklist

### DIFF
--- a/cloudformation/database/serverless-database.yaml
+++ b/cloudformation/database/serverless-database.yaml
@@ -17,6 +17,9 @@ Metadata:
           - Vpc
           - DatabaseClusterSubnetOne
           - DatabaseClusterSubnetTwo
+          - VpcCidrBlock
+          - Subnet1CidrBlock
+          - Subnet2CidrBlock
     ParameterLabels:
       Vpc:
         default: "The VPC to use for the database cluster."
@@ -32,6 +35,12 @@ Metadata:
         default: "The first subnet to use for the database cluster."
       DatabaseClusterSubnetTwo:
         default: "The second subnet to use for the database cluster."
+      VpcCidrBlock:
+        default: "The CIDR block to be used for the VPC if its creation is requested."
+      Subnet1CidrBlock:
+        default: "The CIDR block to be used for the first Subnet if its creation is requested."
+      Subnet2CidrBlock:
+        default: "The CIDR block to be used for the second Subnet if its creation is requested."
 Parameters:
   AdminPasswordSecretString:
     Description: >-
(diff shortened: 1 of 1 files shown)


cloudformation/database/serverless-database.yaml
```diff
@@ -42,7 +51,7 @@ Parameters:
     NoEcho: true
     MinLength: 8
     # Only allow 'Medium' or better strength passwords according to https://dev.mysql.com/doc/refman/8.0/en/validate-password.html
-    AllowedPattern: (?=^.{8,}$)(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\w\s])(?!.*[@/'"])^.*
+    AllowedPattern: '(?=^.{8,}$)(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\w\s])(?!.*[@/''"])^.*'
```
davprat: Does this work?

cloudformation/database/serverless-database.yaml
```diff
@@ -113,6 +140,35 @@ Rules:
...
+        - ''
+    Assertions:
+      - Assert: !Not
+          - !Equals
+            - !Ref Subnet1CidrBlock
+            - '0.0.0.0/24'
+      - Assert: !Not
+          - !Equals
+            - !Ref Subnet2CidrBlock
+            - '0.0.0.0/24'
```
davprat: I think it would be helpful to have a description for when the assertions fail so the user knows what went wrong.

cloudformation/database/serverless-database.yaml
```diff
@@ -17,6 +17,9 @@ Metadata:
           - Vpc
           - DatabaseClusterSubnetOne
           - DatabaseClusterSubnetTwo
+          - VpcCidrBlock
+          - Subnet1CidrBlock
+          - Subnet2CidrBlock
```
davprat, line 22: Would it make more sense to have the Cidr blocks for each resource placed right after the resource id field?

davprat (changes requested):
I think we should add a description to the assertions to better explain what the user needs to fix.

cloudformation/database/serverless-database.yaml
jdeamicis: Restored previous pattern.

cloudformation/database/serverless-database.yaml
jdeamicis: Done.

cloudformation/database/serverless-database.yaml
jdeamicis: I thought about it, and in my opinion having all CIDR block next to each other forces users to think about possible overlaps or inconsistencies in the blocks definition. What do you think?

davprat (dismissed):

davprat (approved):

PR #4429, opened 2022-10-13
2.x: Modify update policy of tags property

enrico-usai:
### Description of changes
* The update policy has been modified to COMPUTE_FLEET_STOP in ParallelCluster 2.10.0 but it never worked, because a change in the tags requires a change in the LaunchTemplate of the head node and this triggers a head node replacement.

With this change the update of the tags is no longer supported during a cluster update.

### Tests
* Created a cluster with no tags and tried to update it --> Update is aborted.
```
Retrieving configuration from CloudFormation for cluster test2x...
Validating configuration file ...
Found Configuration Changes:

#    parameter          old value    new value
---  -----------------  -----------  ------------------------------
     [cluster alinux2]
01*  tags               {}           {'key': 'value', 'key2': 'v...

Validating configuration update...
The requested update cannot be performed. Line numbers with an asterisk indicate updates requiring additional actions. Please review the details below:

#01
Update actions are not currently supported for the 'tags' parameter
How to fix:
Restore 'tags' value to '{}'. If you need this change, please consider creating a new cluster instead of updating the existing one.

In case you want to override these checks and proceed with the update please use the --force flag. Note that the cluster could end up in an unrecoverable state.
Update aborted.
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 CHANGELOG
 =========
 
-2.x.x
+2.11.8
 -----
+
 **CHANGES**
-Upgrade Python runtime used by Lambda functions in AWS Batch integration to python3.9.
+- Upgrade Python runtime used by Lambda functions in AWS Batch integration to python3.9.
+
+**BUG FIXES**
+- Prevent cluster tags to be changed during an update because not supported.
 
 2.11.7
 -----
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 CHANGELOG
 =========
 
-2.x.x
+2.11.8
 -----
 **CHANGES**
-Upgrade Python runtime used by Lambda functions in AWS Batch integration to python3.9.
+- Upgrade Python runtime used by Lambda functions in AWS Batch integration to python3.9.
+- Removed possibility to change tags when performing a `pcluster update`.
```
demartinofra: I'd call this as a bug fix, and slightly rephrase to something like "Preventing cluster tags to be changed during an update because not supported"

demartinofra (approved):

PR #4433, opened 2022-10-14
Add integration test for db server authentication

davprat:
Signed-off-by: David Pratt <davprat@amazon.com>

### Description of changes
* This change adds an integration test to make sure enabling MySQL database server authentication on slurmdbd works.
* Serverless template will be removed from this change as it is in another PR.

### Tests
* Manually ran integration test for slurm accounting.

### Checklist

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
@@ -56,14 +56,11 @@ def _require_server_identity(remote_command_executor, test_resources_dir, region
     )
 
 
-@retry(stop_max_attempt_number=3, wait_fixed=seconds(10))
-def _assert_accounting_is_enabled(remote_command_executor):
-    assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
-
-
 def _test_require_server_identity(remote_command_executor, test_resources_dir, region):
     _require_server_identity(remote_command_executor, test_resources_dir, region)
-    _assert_accounting_is_enabled(remote_command_executor)
+    retry(stop_max_attempt_number=3, wait_fixed=seconds(10))(_is_accounting_enabled)(
+        remote_command_executor,
+    )
 
 
 def _test_slurmdb_users(remote_command_executor, scheduler_commands, test_resources_dir):
--- a/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
@@ -40,6 +40,28 @@ def _is_accounting_enabled(remote_command_executor):
     return remote_command_executor.run_remote_command("sacct", raise_on_error=False).ok
 
 
+def _require_server_identity(remote_command_executor, test_resources_dir, region):
+    ca_url = f"https://truststore.pki.rds.amazonaws.com/{region}/{region}-bundle.pem"
+    remote_command_executor.run_remote_script(
+        str(test_resources_dir / "require_server_identity.sh"),
+        args=[
+            ca_url,
+            f"{region}-bundle.pem",
+        ],
+        run_as_root=True,
+    )
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=10 * 1000)
+def _assert_accounting_is_enabled(remote_command_executor):
+    assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
+
+
+def _test_require_server_identity(remote_command_executor, test_resources_dir, region):
+    _require_server_identity(remote_command_executor, test_resources_dir, region)
+    _assert_accounting_is_enabled(remote_command_executor)
+
+
 def _test_slurmdb_users(remote_command_executor, scheduler_commands, test_resources_dir):
     logging.info("Testing Slurm Accounting Users")
     expected_users = _get_expected_users(remote_command_executor, test_resources_dir)
@@ -130,6 +152,7 @@ def test_slurm_accounting(
     _test_successful_startup_in_log(remote_command_executor)
     _test_slurmdbd_log_exists_in_log_group(cluster)
     _test_slurmdb_users(remote_command_executor, scheduler_commands, test_resources_dir)
+    _test_require_server_identity(remote_command_executor, test_resources_dir, region)
     _test_jobs_get_recorded(scheduler_commands)
 
 
--- a/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import re
 
 import boto3
@@ -6,6 +7,7 @@ import pytest
 from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
 from retrying import retry
+from time_utils import seconds
 
 from tests.cloudwatch_logging import cloudwatch_logging_boto3_utils as cw_utils
 
@@ -30,7 +32,9 @@ def _get_slurm_database_config_parameters(database_stack_outputs):
 
 
 def _get_expected_users(remote_command_executor, test_resources_dir):
-    users = remote_command_executor.run_remote_script(str(test_resources_dir / "get_accounting_users.sh")).stdout
+    users = remote_command_executor.run_remote_script(
+        os.path.join(str(test_resources_dir), "get_accounting_users.sh")
+    ).stdout
     for user in users.splitlines():
         logging.info("  Expected User: %s", user)
     return users.splitlines()
@@ -43,7 +47,7 @@ def _is_accounting_enabled(remote_command_executor):
 def _require_server_identity(remote_command_executor, test_resources_dir, region):
     ca_url = f"https://truststore.pki.rds.amazonaws.com/{region}/{region}-bundle.pem"
     remote_command_executor.run_remote_script(
-        str(test_resources_dir / "require_server_identity.sh"),
+        os.path.join(str(test_resources_dir), "require_server_identity.sh"),
         args=[
             ca_url,
             f"{region}-bundle.pem",
@@ -52,7 +56,7 @@ def _require_server_identity(remote_command_executor, test_resources_dir, region
     )
 
 
-@retry(stop_max_attempt_number=3, wait_fixed=10 * 1000)
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(10))
 def _assert_accounting_is_enabled(remote_command_executor):
     assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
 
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -40,6 +40,28 @@ def _is_accounting_enabled(remote_command_executor):
     return remote_command_executor.run_remote_command("sacct", raise_on_error=False).ok
 
 
+def _require_server_identity(remote_command_executor, test_resources_dir, region):
+    ca_url = f"https://truststore.pki.rds.amazonaws.com/{region}/{region}-bundle.pem"
+    remote_command_executor.run_remote_script(
+        str(test_resources_dir / "require_server_identity.sh"),
```
jdeamicis: Is this correct? Shouldn't it be `str(test_resources_dir "/require_server_identity.sh)`?

Alternatively you may use [`os.path.join`](https://docs.python.org/3/library/os.path.html#os.path.join).

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -40,6 +40,28 @@ def _is_accounting_enabled(remote_command_executor):
...
+        args=[
+            ca_url,
+            f"{region}-bundle.pem",
+        ],
+        run_as_root=True,
+    )
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=10 * 1000)
```
jdeamicis: For better readability, I would write `wait_fixed=seconds(10)`.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -40,6 +40,28 @@ def _is_accounting_enabled(remote_command_executor):
...
+def _assert_accounting_is_enabled(remote_command_executor):
+    assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
```
jdeamicis: This is a one-liner, and `assert_that(_is_accounting_enabled(...))` has already been used directly in other functions. I would not wrap this into a new function.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -40,6 +40,28 @@ def _is_accounting_enabled(remote_command_executor):
...
+
+
+def _test_require_server_identity(remote_command_executor, test_resources_dir, region):
+    _require_server_identity(remote_command_executor, test_resources_dir, region)
+    _assert_accounting_is_enabled(remote_command_executor)
+
```
jdeamicis: See previous comment.

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
```diff
@@ -0,0 +1,21 @@
...
+CA_URL=$1
+CA_NAME=$2
+CA_PATH="/tmp/slurm"
+CA_FILE_PATH="${CA_PATH}/${CA_NAME}"
+SLURMDBD_CONFIG_FILE="/opt/slurm/etc/slurmdbd.conf"
+
+wget ${CA_URL} -nH -x --cut-dirs=1 -P ${CA_PATH}
+
+echo "Clearing previous CA Setting"
+sed -i "s/^StorageParameters=.*$//g" "${SLURMDBD_CONFIG_FILE}"
```
jdeamicis: You should be able to delete the line with:
```
sed -i "/^StorageParameters\s*=/d"
```

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "StorageParameters=" >> "${SLURMDBD_CONFIG_FILE}"
+
+echo "Setting 'StorageParameters' SSL_CA=${CA_FILE_PATH}"
+sed -i "s:^StorageParameters=.*$:StorageParameters=SSL_CA=${CA_FILE_PATH}:g" "${SLURMDBD_CONFIG_FILE}"
```
jdeamicis: You can join these two by doing:
```
echo "StorageParameters=SSL_CA=${CA_FILE_PATH}" >> "${SLURMDBD_CONFIG_FILE}"
```

jdeamicis (commented):
Overall the logic is fine, I propose some changes to some of the statements in the tests.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Yeah, os.path.join is better - I was just following the pattern I saw in test_slurm.py, which does have the `/` operator: https://docs.python.org/3/library/pathlib.html#operators

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Hmm... except there is a bit more going on here in that the _assert_accounting_is_enabled uses the retry decorator to automatically retry the assertion if it fails to allow for the service to have some time to startup. Is there a clean way to do this without wrapping it in a function?

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
davprat: I remember I had a reason for doing it this way, but I don't remember why. I will change it to your version.

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
davprat: Done

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
davprat: Done

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Done

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Done

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
jdeamicis: I'm not sure if it's clean enough, but you don't necessarily have to decorate the target function when defining it: you can call the decorator directly when using the target function as done [here](https://github.com/aws/aws-parallelcluster/blob/c8248dfa53c88b10a16742808e963da270ab520f/tests/integration-tests/tests/schedulers/test_slurm.py#L838-L842):

```python
    retry(wait_fixed=seconds(20), stop_max_delay=minutes(5))(assert_lines_in_logs)(
        remote_command_executor,
        ["/var/log/slurmctld.log"],
        ["Nodes {} not responding, setting DOWN".format(",".join(dynamic_nodes))],
    )
```

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
jdeamicis: Oh, I didn't know about pathlib. Thanks!

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Sounds good - I've updated to code to work like this.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -30,7 +32,9 @@ def _get_slurm_database_config_parameters(database_stack_outputs):
 
 
 def _get_expected_users(remote_command_executor, test_resources_dir):
-    users = remote_command_executor.run_remote_script(str(test_resources_dir / "get_accounting_users.sh")).stdout
+    users = remote_command_executor.run_remote_script(
+        os.path.join(str(test_resources_dir), "get_accounting_users.sh")
```
jdeamicis, line 36: Minor: isn't `test_resources_dir` a string already?

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -43,7 +47,7 @@ def _is_accounting_enabled(remote_command_executor):
 def _require_server_identity(remote_command_executor, test_resources_dir, region):
     ca_url = f"https://truststore.pki.rds.amazonaws.com/{region}/{region}-bundle.pem"
     remote_command_executor.run_remote_script(
-        str(test_resources_dir / "require_server_identity.sh"),
+        os.path.join(str(test_resources_dir), "require_server_identity.sh"),
```
jdeamicis, line 50: Minor: same as above.

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -e
+
+CA_URL=$1
+CA_NAME=$2
+CA_PATH="/tmp/slurm"
+CA_FILE_PATH="${CA_PATH}/${CA_NAME}"
+SLURMDBD_CONFIG_FILE="/opt/slurm/etc/slurmdbd.conf"
+
+wget ${CA_URL} -nH -x --cut-dirs=1 -P ${CA_PATH}
```
jdeamicis: I didn't notice this earlier, but the options `-nH -x --cut-dirs=1` here are useless because they are canceling each other out. Please remove them and keep the command as:
```
wget ${CA_URL} -P ${CA_PATH}
```

jdeamicis (commented):
I left a couple of minor comments, plus one that I would like to be addressed before merging.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: It's a path object I think.

tests/integration-tests/tests/schedulers/test_slurm_accounting.py
davprat: Same as above.

tests/integration-tests/tests/schedulers/test_slurm_accounting/resources/require_server_identity.sh
davprat: Good catch. Removed.

jdeamicis (approved):

PR #4454, opened 2022-10-19
[Test] Add fixture to serialize the execution of marked tests according to the adopted instance type

gmarciani:
### Description of changes
1. Add fixture `serial_execution_by_instance` to serialize the execution of marked tests when they are executed with instances c5n.18xlarge or p4d.24xlarge.
2. Applied the above fixture to tests: test_efa, test_fabric and test_intel_hpc.

### How to use this fixture
```
@pytest.mark.usefixtures("serial_execution_by_instance")
def test_1(...):
   ...

@pytest.mark.usefixtures("serial_execution_by_instance")
def test_2(...):
   ...
```

When the above tests are executed with instance equal to c5n.18xlarge or p4d.24xlarge, then theior execution is serialized via file locking, otherwise, the desired parallelism is applied.

### Tests
1. Tested locally with fake tests to validate enforcement of serial execution on the desired tests.
4. Tested serial execution of tests test_efa, test_fabric and test_intel_hpc with parallelism 16, even with for some of them I still got ICEd

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -1223,6 +1223,24 @@ def pytest_runtest_makereport(item, call):
             logging.error("Failed when generating config for failed tests: %s", e, exc_info=True)
 
 
+@pytest.fixture(scope="class")
+def serial_execution_by_instance(request, instance):
+    """Enforce serial execution of tests, according to the adopted instance."""
+    if instance in ["c5n.18xlarge", "p4d.24xlarge"]:
+        logging.info("Enforcing serial execution for instance %s", instance)
+        outdir = request.config.getoption("output_dir")
+        lock_file = f"{outdir}/{instance}.lock"
+        lock = FileLock(lock_file=lock_file)
+        logging.info("Acquiring lock file %s", lock.lock_file)
+        with lock.acquire(poll_interval=15, timeout=7200):
+            yield
+        logging.info("Releasing lock file %s", lock.lock_file)
+        lock.release()
+    else:
+        logging.info("Ignoring serial execution for instance %s", instance)
+        yield
+
+
 def update_failed_tests_config(item):
     out_dir = Path(item.config.getoption("output_dir"))
     if not str(out_dir).endswith(".out"):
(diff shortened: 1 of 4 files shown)


tests/integration-tests/conftest.py
```diff
@@ -1223,6 +1223,24 @@ def pytest_runtest_makereport(item, call):
             logging.error("Failed when generating config for failed tests: %s", e, exc_info=True)
 
 
+@pytest.fixture(scope="class")
+def serial_execution_by_instance(request, instance):
+    """Enforce serial execution of tests, according to the adopted instance."""
+    if instance in ["c5n.18xlarge", "p4d.24xlarge"]:
+        logging.info("Enforcing serial execution for instance %s", instance)
+        outdir = request.config.getoption("output_dir")
+        lock_file = f"{outdir}/{instance}.lock"
```
enrico-usai, line 1232: Is it output_dir a shared dir for all the tests? or for the single region?
Can you add in the comment an example of output_dir ?

I'm asking this because otherwise we should add the region in the lockfile name to be sure to not block other tests in other regions.

tests/integration-tests/conftest.py
```diff
@@ -1223,6 +1223,24 @@ def pytest_runtest_makereport(item, call):
...
+        lock = FileLock(lock_file=lock_file)
+        logging.info("Acquiring lock file %s", lock.lock_file)
+        with lock.acquire(poll_interval=0.1):
```
enrico-usai: I think this poll interval is too strict, we're talking about a test execution so maybe we can check every 30seconds?

Then what about adding a timeout (2h) to the acquire method?

tests/integration-tests/conftest.py
gmarciani: output_dir is the regional folder for tests.

tests/integration-tests/conftest.py
gmarciani: Very good catch, thank you!
I'd go with a poll interval of 15 seconds and a 2h timeout on acquire

enrico-usai (dismissed):

enrico-usai (approved):

PR #4460, opened 2022-10-19
Fix InstanceType config in Slurm accounting integration tests

davprat:
### Description of changes
* Change to Instances schema broke configs used for Slurm integration unit tests.
* Fixed configs to conform to the correct schema.

### Tests
* Manually ran integration tests for Slurm accounting.

### Checklist


lukeseawalker:
> * Change to Instances schema broke configs used for Slurm integration unit tests.

I don't understand why, the test should work with the new schema as well, can we followup on this?

enrico-usai:
I agree on Luca's comment, the integration tests should pass with both the configuration. Both of them must be supported. 

Could you add in the description which is the issue we're seeing in the integration tests without this patch?

PR #4466, opened 2022-10-20
Integ-tests: Reduce sporadic failures of test_scontrol_reboot

hanwen-pcluste:

### References
* Here is a similar PR https://github.com/aws/aws-parallelcluster/pull/4217/files

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -443,6 +443,7 @@ def test_scontrol_reboot(
         slurm_commands,
         ["queue1-dy-t2micro-1", "queue1-dy-t2micro-2"],
         "idle",
+        stop_max_delay_secs=330,
     )
 
     # Test that idle static and dynamic nodes can be rebooted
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -443,6 +443,7 @@ def test_scontrol_reboot(
         slurm_commands,
         ["queue1-dy-t2micro-1", "queue1-dy-t2micro-2"],
         "idle",
+        stop_max_delay_secs=330,
```
jdeamicis, line 446: If we are really seeing issues with this first job, let's increase this to 600 seconds to be on the safe side. If 600 seconds are not enough, then there's a problem with EC2 I would say.

jdeamicis (approved):
Added one minor comment.

tests/integration-tests/tests/schedulers/test_slurm.py
hanwen-pcluste: We are indeed seeing failures here. I put 330 to make sure the nodes finish the job within reasonable timing.
I am ok with increasing from 300 to 330 because usually we wait node state by state:
1. assert job running
2. assert job finishing
3. assert nodes idle
Here we are asserting nodes idle directly, so 30 more seconds should cover all the delays in the other two steps

PR #4484, opened 2022-10-25
Add missing fields to launch template validators

ermanno:
### Notes
This patch adds the missing fields to the launch template validators. For the head node, these are the `BlockDeviceMappings` and `KeyName`, while for the compute resources these are the `BlockDeviceMappings`, `InstanceMetadataOptions` and `CapacityReservationSpecification`. Created an abstract class holding the common logic, subclassed by two concrete classes generating CDK object and dictionaries for the desired launch template resources, by implementing abstract methods that are called in the template methods defined in the superclass.

### Tests
- Unit tests
- Manual test with the following configuration (note that root volumes are encrypted by default), verifying that it failed with the old code, and passes with my patch (because of the additional fields passed in the run instance dry-run that is executed by the validators):
```
Region: eu-west-1
Image:
  Os: alinux2
HeadNode:
  InstanceType: t2.micro
  Networking:
    SubnetId: <subnet id>
  Ssh:
    KeyName: <ssh key>
  Iam:
    S3Access:
      - BucketName: <bucket name>
Scheduling:
  Scheduler: slurm
  SlurmQueues:
  - Name: queue1
    CapacityType: SPOT
    ComputeResources:
    - Name: t2micro
      InstanceType: c5.xlarge
      MinCount: 1
      MaxCount: 10
    Networking:
      SubnetIds:
      - <subnet id>
```

on an account with the following policy settings disabling the creation of and run instance with unencrypted volumes:

```
{
  "Effect": "Deny",
  "Action": "ec2:CreateVolume",
  "Resource": "*",
  "Condition": {
    "Bool": {
      "ec2:Encrypted": "false"
    }
  }
},
{
  "Effect": "Deny",
  "Action": "ec2:RunInstances",
  "Resource": "arn:aws:ec2:*:*:volume/*",
  "Condition": {
    "Bool": {
      "ec2:Encrypted": "false"
    }
  }
}
```

### DIFF
--- a/cli/src/pcluster/launch_template_utils.py
+++ b/cli/src/pcluster/launch_template_utils.py
@@ -1,0 +1,56 @@
+from abc import ABC, abstractmethod
+
+from pcluster.constants import OS_MAPPING
+
+
+class _LaunchTemplateBuilder(ABC):
+    """Abstract class with methods with the common logic for launch template builders."""
+
+    def get_block_device_mappings(self, root_volume, image_os):
+        """Return a list of block device mappings."""
+        block_device_mappings = []
+        for _, (device_name_index, virtual_name_index) in enumerate(zip(list(map(chr, range(97, 121))), range(0, 24))):
+            device_name = "/dev/xvdb{0}".format(device_name_index)
+            virtual_name = "ephemeral{0}".format(virtual_name_index)
+            block_device_mappings.append(self._block_device_mapping_for_virt(device_name, virtual_name))
+
+        block_device_mappings.append(
+            self._block_device_mapping_for_ebs(OS_MAPPING[image_os]["root-device"], root_volume)
+        )
+        return block_device_mappings
+
+    def get_instance_market_options(self, queue, compute_resource):
+        """Return the instance market options for spot instances."""
+        instance_market_options = None
+        if queue.is_spot():
+            instance_market_options = self._instance_market_option(
+                market_type="spot",
+                spot_instance_type="one-time",
+                instance_interruption_behavior="terminate",
+                max_price=None if compute_resource.spot_price is None else str(compute_resource.spot_price),
+            )
+        return instance_market_options
+
+    def get_capacity_reservation(self, queue, compute_resource):
+        """Return the capacity reservation if a target is defined at the queue or compute resource level."""
+        capacity_reservation = None
+        cr_target = compute_resource.capacity_reservation_target or queue.capacity_reservation_target
+        if cr_target:
+            capacity_reservation = self._capacity_reservation(cr_target)
+        return capacity_reservation
+
+    @abstractmethod
+    def _block_device_mapping_for_virt(self, device_name, virtual_name):
+        pass
+
+    @abstractmethod
+    def _block_device_mapping_for_ebs(self, device_name, volume):
+        pass
+
+    @abstractmethod
+    def _instance_market_option(self, market_type, spot_instance_type, instance_interruption_behavior, max_price):
+        pass
+
+    @abstractmethod
+    def _capacity_reservation(self, cr_target):
+        pass
--- a/cli/src/pcluster/utils.py
+++ b/cli/src/pcluster/utils.py
@@ -361,3 +361,8 @@ def yaml_no_duplicates_constructor(loader, node, deep=False):
 def get_http_tokens_setting(imds_support):
     """Get http tokens settings for supported IMDS version."""
     return "required" if imds_support == "v2.0" else "optional"
+
+
+def remove_none_values(original_dictionary):
+    """Return a dictionary without entries with None value."""
+    return {key: value for key, value in original_dictionary.items() if value is not None}
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -35,7 +35,13 @@ from pcluster.constants import (
     SUPPORTED_REGIONS,
     SUPPORTED_SCHEDULERS,
 )
-from pcluster.utils import get_installed_version, get_supported_os_for_architecture, get_supported_os_for_scheduler
+from pcluster.launch_template_utils import _LaunchTemplateBuilder
+from pcluster.utils import (
+    get_installed_version,
+    get_supported_os_for_architecture,
+    get_supported_os_for_scheduler,
+    remove_none_values,
+)
 from pcluster.validators.common import FailureLevel, Validator
 
 # pylint: disable=C0302
@@ -941,6 +947,10 @@ class MixedSecurityGroupOverwriteValidator(Validator):
 class _LaunchTemplateValidator(Validator):
     """Abstract class to contain utility functions used by head node and queue LaunchTemplate validators."""
 
+    def __init__(self):
+        super().__init__()
+        self._launch_template_builder = DictLaunchTemplateBuilder()
+
     @staticmethod
     def _build_launch_network_interfaces(
         network_interfaces_count, use_efa, security_group_ids, subnet, use_public_ips=False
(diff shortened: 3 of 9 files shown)


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1202,3 +1220,50 @@ def _validate(self, scheduler):
...
+            k: v
+            for k, v in {
+                "Encrypted": root_volume.encrypted,
+                "VolumeType": root_volume.volume_type,
+                "Iops": root_volume.iops,
+                "Throughput": root_volume.throughput,
+                "DeleteOnTermination": root_volume.delete_on_termination,
+                "VolumeSize": root_volume.size,
+            }.items()
+            if v is not None
```
enrico-usai: we can create a `_remove_none_values` utility function in this module and reuse it here and in the `get_instance_market_options_template` function .

cli/src/pcluster/launch_template_utils.py
```diff
@@ -0,0 +1,46 @@
+from pcluster.constants import OS_MAPPING
+
+
+def get_block_device_mappings_template(create_block_device_mapping_for_virt, create_block_device_mapping_for_ebs):
```
enrico-usai: After looking at the changes I think it's better to use the class approach you were using in the first commit. I mean defining an `AbstractLaunchTemplateBuilder` here and have the two different implementation `CdkLaunchTemplateBuilder` and `DictLaunchTemplateBuilder`

```python
class AbstractLaunchTemplateBuilder(ABC):

    def get_block_device_mappings(root_volume, image_os):
        ...
        for ...
            ...
            block_device_mappings.append(self._create_block_device_mapping_for_virt(device_name, virtual_name))

        block_device_mappings.append(
            self._create_block_device_mapping_for_ebs(OS_MAPPING[image_os]["root-device"], root_volume)
        )
        return block_device_mappings

    @abstractmethod
    def _create_block_device_mapping_for_virt(self, ...):
        pass

    @abstractmethod
    def _create_block_device_mapping_for_ebs(self, ...):
        pass

    def get_instance_market_options(queue, compute_resource):
        ... self._create_instance_market_options(...)

    @abstractmethod
    def _create_instance_market_options()

    ...get_capacity_reservation ... etc
```

In this way:
* we're encapsulating all the Launch Template related logic into a single class (that in future we can extend to build the full LaunchTemplate)
* the responsibility of the implementation will be incapsulated in the Cdk and Dict builder classes and not in the caller of the function
* we don't need to pass lambda functions every time we need to call these functions, so we can reuse them in multiple places
* we can remove the function `get_block_device_mappings` from `cdk_builder_utils` and instead instantiate `CdkLaunchTemplateBuilder` and `DictLaunchTemplateBuilder` in the places we need them

See: https://refactoring.guru/design-patterns/template-method/python/example

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1143,6 +1155,12 @@ def _test_compute_resource(
             NetworkInterfaces=network_interfaces,
             DryRun=True,
             TagSpecifications=self._generate_tag_specifications(tags),
+            InstanceMarketOptions=get_instance_market_options(queue, compute_resource),
+            CapacityReservationSpecification=get_capacity_reservation(
+                queue,
+                compute_resource,
```
enrico-usai: nit: if you remove the comma after compute_resource and execute autoformat this will fit into a single line.

cli/tests/pcluster/templates/test_cdk_builder_utils.py
```diff
@@ -10,10 +10,238 @@
...
+                        throughput=20,
+                        delete_on_termination=True,
+                    ),
+                ),
+            ],
+            id="test with all root volume fields populated",
+        ),
+        pytest.param(
+            RootVolume(
+                encrypted=True,
```
enrico-usai: let's modify in this tests the values for all the parameters: `encrypted=False`, `iops=13`, `throughput=30`, `delete_on_termination=False`, etc just to pass from another path.

cli/tests/pcluster/templates/test_cdk_builder_utils.py
```diff
@@ -10,10 +10,238 @@
...
+                volume_type="mockVolumeType",
+                iops=15,
+                throughput=20,
+                delete_on_termination=True,
+            ),
+            "centos7",
```
enrico-usai: let's try with another os, in this way will validate that `device_name` for the root volume is a different one.
We should have a separate unit tests to verify all the OSes but we can cover something with this switch.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1116,6 +1126,221 @@ def test_generate_tag_specifications(input_tags):
...
+                        "Throughput": 20,
+                        "DeleteOnTermination": True,
+                    },
+                },
+            ],
+            id="test with all root volume fields populated",
+        ),
+        pytest.param(
+            RootVolume(
+                encrypted=True,
```
enrico-usai: same here, let's try other values for the parameters.

cli/src/pcluster/launch_template_utils.py
ermanno: I don't have strong feelings on this, the functional implementation (2nd commit) was because I had later read in Effective Java 3rd edition the suggestion to use functional interfaces and composition instead of inheritance like in the traditional GoF design pattern (I realize that this is Python, but the general point being made seemed relevant since it was in the context of the introduction of functional interfaces, and Python supports higher order functions). I will proceed to revert that commit and, as per your comment above, change it to have a builder for each "kind" of launch template (the CDK one and the EC2 run instance params that we need in the launch template validators to simulate the launch template).

>we don't need to pass lambda functions every time we need to call these functions, so we can reuse them in multiple places

To be clear, the function that was called was the `get_instance_market_options` which was created starting from the template one (in 2nd commit called `get_market_options_template`, maybe a better name would have been `make_market_options_getter`), which is functionally equivalent to subclassing in that respect (or, in the case of Effective Java's suggestion, a static factory method that took the two functions as input). In other words, the caller does not need to pass the lambdas each time: in canonical GoF implementation, the functions are implemented in the subclass, in the functional one they are passed to the function that generates the CDK/Dictionary function, and then the generated function, `get_instance_market_options`, is the one that is called in the launch templates generator for the head node and the compute resources, for example.

cli/src/pcluster/validators/cluster_validators.py
ermanno: I'll create a function in `utils.py` which I will reuse both here and in the `_LaunchTemplateValidator`.

cli/tests/pcluster/templates/test_cdk_builder_utils.py
ermanno: Agree, however note that these pass through the same path (we just take the values from the volume object, we don't process them in any way, in terms of our logic, so we don't branch on those conditions).

cli/tests/pcluster/templates/test_cdk_builder_utils.py
ermanno: Agreed, however note that this is not something that influences our logic (we use the values defined in our `constants.py` file for the `OS_MAPPING`).

cli/src/pcluster/utils.py
```diff
@@ -361,3 +361,8 @@ def yaml_no_duplicates_constructor(loader, node, deep=False):
 def get_http_tokens_setting(imds_support):
     """Get http tokens settings for supported IMDS version."""
     return "required" if imds_support == "v2.0" else "optional"
+
+
+def dictionary_without_entries_with_none_values(original_dictionary):
```
enrico-usai: what about: `remove_none_values(dictionary)` ?

cli/src/pcluster/launch_template_utils.py
```diff
@@ -0,0 +1,139 @@
...
+    @abstractmethod
+    def _instance_market_option(self, market_type, spot_instance_type, instance_interruption_behavior, max_price):
+        pass
+
+    @abstractmethod
+    def _capacity_reservation(self, cr_target):
+        pass
+
+
+class CdkLaunchTemplateBuilder(_LaunchTemplateBuilder):
```
enrico-usai: We could move this into `cdk_builder_utils.py` to have all CDK related logic into `templates` folder.

enrico-usai (approved):

PR #4493, opened 2022-11-01
Change Node version validator to use 'nvm current'

davprat:
### Description of changes
* Use `nvm current` to get the current version of Node.js if `node --version` fails
* Raise an exception if `nvm current' returns a major version is greater than a max major version asking the user to install the latest release of 'max major version'.

### Tests
* Added test to make sure that the fallback to `nvm current` succeeds if the call to `node --version` fails.

### Checklist


cli/src/pcluster/api/util.py
```diff
@@ -177,19 +177,46 @@ def _assert_node_executable():
...
+        except Exception:
+            message = "Unable to check Node.js version"
+            LOGGER.critical(message)
+            raise Exception(message)
+        if node_version.major > packaging.version.parse(NODEJS_MAX_MAJOR_VERSION).major:
+            message = (
+                f"Node.js version {node_version_string} may not work on this platform. Use the Node"
+                f" Version Manager (nvm) to install and use the latest release of Node.js version"
+                f" {NODEJS_MAX_MAJOR_VERSION}."
+            )
```
demartinofra: I'd prefer if possible not to hardcode NODEJS_MAX_MAJOR_VERSION in our CLI. Can we just suggest to make sure node is properly installed and functional if we detect that node is available but returning some errors?

davprat: Removed and reworked.

cli/src/pcluster/api/util.py
```diff
@@ -177,19 +177,47 @@ def _assert_node_executable():
...
+            #   untrusted source
+            node_version_string = subprocess.check_output(  # nosec
+                ["nvm", "current"], stderr=subprocess.STDOUT, shell=False, encoding="utf-8"
+            )
+            LOGGER.debug("Found Node.js version '%s' in use", node_version_string)
+        except Exception:
+            message = "Unable to check Node.js version"
+            LOGGER.critical(message)
+            raise Exception(message)
+        if node_version_string == "none":
```
demartinofra, line 205: [optional] might be useful to tell in a comment what command we expect to return the "none" string. is it `nvm current`? do we need a test case for this?

demartinofra (dismissed):
LGTM, left a minor comment

cli/src/pcluster/api/util.py
davprat: I think this [test case](https://github.com/aws/aws-parallelcluster/pull/4493/files#diff-9350f46969010c1d42950e607a45080fda852e9c43b1939bd821b40212586734R59) covers the none response? Let me know if I'm missing something though.

davprat: Added comment.

demartinofra (approved):

PR #4520, opened 2022-11-08
Allow to specify a VcpConfig for PCluster Lambdas

ermanno:
### Description of changes
This patch adds the possibility to specify a VPC configuration to use for ParallelCluster Lambda Functions (which ParallelCluster uses to implement CloudFormation Custom Resources). This patch also adds a validator that will be used to validate the ParallelCluster Lambda functions' VPC configuration. It checks that the configuration's subnets and security groups are present in the account and all have the same VPC.

### Tests
- Unit tests.
- Manual tests with:
  - Private subnet with NATGW.
  - Private subnet isolated from the internet configured with VPC endpoints ([ParallelCluster documentation](https://docs.aws.amazon.com/parallelcluster/latest/ug/network-configuration-v3.html#aws-parallelcluster-in-a-single-public-subnet-no-internet-v3)).

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/awsbatch_builder.py
+++ b/cli/src/pcluster/templates/awsbatch_builder.py
@@ -617,6 +617,7 @@ class AwsBatchConstruct(Construct):
                         )
                     ),
                 ],
+                has_vpc_config=self.config.lambda_functions_vpc_config,
             )
 
         return PclusterLambdaConstruct(
@@ -680,6 +681,7 @@ class AwsBatchConstruct(Construct):
                         )
                     )
                 ],
+                has_vpc_config=self.config.lambda_functions_vpc_config,
             )
 
         return PclusterLambdaConstruct(
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -77,8 +77,8 @@ class SingleSubnetValidator(Validator):
             self._add_failure("The SubnetId used for all of the queues should be the same.", FailureLevel.ERROR)
 
 
-class LambdaVpcConfigValidator(Validator):
-    """Validator of a Lambda function's VPC configuration."""
+class LambdaFunctionsVpcConfigValidator(Validator):
+    """Validator of Pcluster Lambda functions' VPC configuration."""
 
     def _validate(self, security_group_ids: List[str], subnet_ids: List[str]):
         existing_security_groups = AWSApi.instance().ec2.describe_security_groups(security_group_ids)
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -75,3 +75,51 @@ class SingleSubnetValidator(Validator):
         subnet_ids = {tuple(set(queue_subnets)) for queue_subnets in queues_subnets}
         if len(subnet_ids) > 1:
             self._add_failure("The SubnetId used for all of the queues should be the same.", FailureLevel.ERROR)
+
+
+class LambdaVpcConfigValidator(Validator):
+    """Validator of a Lambda function's VPC configuration."""
+
+    def _validate(self, security_group_ids: List[str], subnet_ids: List[str]):
+        existing_security_groups = AWSApi.instance().ec2.describe_security_groups(security_group_ids)
+        existing_subnets = AWSApi.instance().ec2.describe_subnets(subnet_ids)
+
+        self._validate_all_security_groups_exist(existing_security_groups, security_group_ids)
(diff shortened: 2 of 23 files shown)


cli/src/pcluster/templates/awsbatch_builder.py
```diff
@@ -617,6 +617,7 @@ def _add_manage_docker_images_lambda(self):
                         )
                     ),
                 ],
+                has_vpc_config=self.config.lambda_functions_vpc_config,
```
francesco-giordano, line 620: Should `self.config.lambda_functions_vpc_config` be `self.config.deployment_settings.lambda_functions_vpc_config` in every configuration?

francesco-giordano (commented):
Are the integration tests covered in future change?

ermanno:
> Are the integration tests covered in future change?

Yes, they will be pushed in a separate PR

cli/src/pcluster/templates/awsbatch_builder.py
ermanno: Yes, in this case the two structures are at the same level (top level) for the build and cluster configurations (this needs not necessarily be the case for shared structures, for example in the case of `Imds` we have it at top level for the cluster config and under the `Build` element in the build image one, despite it being the same structure).

This is defined as a property (because Python does not have optional chaining/null propagation operator, i.e. like `.?` in JS, plus if we make a change in the naming in the future, we can change the "path" only in the property):

```
    @property
    def lambda_functions_vpc_config(self):
        """Return the vpc config of the PCluster Lambda Functions or None."""
        return self.deployment_settings.lambda_functions_vpc_config if self.deployment_settings else None
```

This is similar to what we do in other cases such as:

```
    @property
    def custom_node_package(self):
        """Return custom node package value or None."""
        return self.dev_settings.node_package if self.dev_settings else None

    @property
    def custom_aws_batch_cli_package(self):
        """Return custom custom aws batch cli package value or None."""
        return self.dev_settings.aws_batch_cli_package if self.dev_settings else None
```

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -75,3 +75,51 @@ def _validate(self, queues_subnets):
...
+        self._validate_all_security_groups_exist(existing_security_groups, security_group_ids)
+        self._validate_all_subnets_exist(existing_subnets, subnet_ids)
+        self._validate_all_resources_belong_to_the_same_vpc(existing_security_groups, existing_subnets)
+
+    def _validate_all_resources_belong_to_the_same_vpc(self, existing_security_groups, existing_subnets):
+        group_vpc_ids = {group["VpcId"] for group in existing_security_groups}
+        subnet_vpc_ids = {subnet["VpcId"] for subnet in existing_subnets}
+        if len(group_vpc_ids) > 1:
+            self._add_failure(
+                "The security groups associated to the Lambda are required to be in the same VPC.", FailureLevel.ERROR
```
enrico-usai, line 96: minor: you could avoid duplication creating a format string to be used in every case:
```python
message = "The {} associated to the Lambda are required to be in the same VPC."
```
then you can use:
```python
message.format("security groups")
message.format("subnets")
message.format("security groups and subnets")
```
in the different cases.

cli/tests/pcluster/utils.py
```diff
@@ -46,3 +49,33 @@ def get_resources(
...
+            assert_that(_get_vpc_config(lambda_function)).is_none()
+            assert_that(_get_managed_policy_arns(role)).does_not_contain(LAMBDA_VPC_ACCESS_MANAGED_POLICY)
+
+
+def _get_vpc_config(lambda_function):
+    return lambda_function.get("Properties").get("VpcConfig")
+
+
+def _get_role_name(lambda_function):
+    return lambda_function.get("Properties").get("Role").get("Fn::GetAtt")[0]
```
enrico-usai, line 73: Note: when you have multiple getters you can avoid KeyError or None exceptions in this way and fail according to the returned value.
```python
return lambda_function.get("Properties", {}).get("Role", {}).get("Fn::GetAtt")[0]
```

enrico-usai (approved):

cli/tests/pcluster/utils.py
ermanno: Yes, however in this case I expect there to be there because these are the unit tests whose content I control (it would be a problem if these were empty).

cli/src/pcluster/validators/networking_validators.py
ermanno: I don't have strong opinions on this, but imho it wouldn't really make the code clearer/more explicit.

PR #4523, opened 2022-11-08
[3.x] Integ-tests: improve test_multiple_efs

hanwen-pcluste:
1. Check EFS file systems are mounted with `amazon-efs-utils`
2. Check EFS file systems are auto mounted after instances reboot.
3. Minor refactor of the test code. 

To Do: this change assumes `tls` and `iam` are turned off. After making CLI change to support the new parameters, we should enrich the test to test enabling `tls` and `iam`

### Tests
* test_multiple_efs has been passed

### References
* This test can pass only after https://github.com/aws/aws-parallelcluster-cookbook/pull/1588 is merged

### Checklist

### DIFF
--- a/tests/integration-tests/tests/storage/storage_common.py
+++ b/tests/integration-tests/tests/storage/storage_common.py
@@ -161,18 +161,27 @@ def _write_user_data(efs_id, random_file_name):
         """  # noqa: E501
 
 
-def test_efs_correctly_mounted(remote_command_executor, mount_dir):
-    logging.info("Testing ebs {0} is correctly mounted".format(mount_dir))
-    result = remote_command_executor.run_remote_command("df | grep '{0}'".format(mount_dir))
+def test_efs_correctly_mounted(remote_command_executor, mount_dir, tls=False):
+    # TODO: Add iam. The value of the two parameters should be set according to cluster configuration parameters.
+    logging.info("Checking efs {0} is correctly mounted".format(mount_dir))
+    # Following EFS instruction to check https://docs.aws.amazon.com/efs/latest/ug/encryption-in-transit.html
+    result = remote_command_executor.run_remote_command("mount | column -t | grep '{0}'".format(mount_dir))
     assert_that(result.stdout).contains(mount_dir)
-
-    result = remote_command_executor.run_remote_command("cat /etc/fstab")
-    assert_that(result.stdout).matches(
-        (
-            r".* {mount_dir} nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,"
-            r"timeo=30,retrans=2,noresvport,_netdev 0 0"
-        ).format(mount_dir=mount_dir)
+    if tls:
+        logging.info("Checking efs {0} enables in-transit encryption".format(mount_dir))
+        assert_that(result.stdout).contains("127.0.0.1:/")
+    logging.info("Checking efs {0} is successfully mounted in efs mount log".format(mount_dir))
+    result = remote_command_executor.run_remote_command(
+        f'grep -E "Successfully mounted.*{mount_dir}" /var/log/amazon/efs/mount.log'
     )
+    assert_that(result.stdout).contains(mount_dir)
+    # Check fstab content according to https://docs.aws.amazon.com/efs/latest/ug/automount-with-efs-mount-helper.html
+    logging.info("Checking efs {0} is correctly configured in fstab".format(mount_dir))
+    result = remote_command_executor.run_remote_command("cat /etc/fstab")
+    if tls:  # TODO: Add a another check when tls and iam are enabled together
+        assert_that(result.stdout).matches(rf".* {mount_dir} efs _netdev,noresvport,tls 0 0")
+    else:
+        assert_that(result.stdout).matches(rf".* {mount_dir} efs _netdev,noresvport 0 0")
 
 
 # for FSX
--- a/tests/integration-tests/tests/storage/test_efs.py
+++ b/tests/integration-tests/tests/storage/test_efs.py
@@ -20,8 +20,8 @@ from troposphere import Ref, Template, ec2
 from troposphere.efs import MountTarget
 from utils import generate_stack_name, get_compute_nodes_instance_ips, get_vpc_snakecase_value
 
+from tests.common.utils import reboot_head_node
 from tests.storage.storage_common import (
-    get_efs_ids,
     test_efs_correctly_mounted,
     verify_directory_correctly_shared,
     write_file_into_efs,
@@ -95,6 +95,7 @@ def test_multiple_efs(
         num_existing_efs = 20
     else:
         num_existing_efs = 2
+    # TODO: create an additional EFS with file system policy to prevent anonymous access
     existing_efs_ids = efs_stack_factory(num_existing_efs)
     if request.config.getoption("benchmarks"):
         mount_target_stack_factory(existing_efs_ids)
@@ -109,7 +110,7 @@ def test_multiple_efs(
     new_efs_mount_dirs = ["/shared"]  # OSU benchmark relies on /shared directory
 
     _assert_subnet_az_relations(region, vpc_stack, expected_in_same_az=False)
-
+    # TODO: change cluster configuration file to test different tls and iam settings to EFS.
     cluster_config = pcluster_config_reader(
         existing_efs_mount_dirs=existing_efs_mount_dirs,
         existing_efs_ids=existing_efs_ids,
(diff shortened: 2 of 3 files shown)


tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -161,18 +161,24 @@ def _write_user_data(efs_id, random_file_name):
         """  # noqa: E501
 
 
-def test_efs_correctly_mounted(remote_command_executor, mount_dir):
-    logging.info("Testing ebs {0} is correctly mounted".format(mount_dir))
-    result = remote_command_executor.run_remote_command("df | grep '{0}'".format(mount_dir))
+def test_efs_correctly_mounted(remote_command_executor, mount_dir, tls=False):
```
gmarciani, line 164: Why having a parameter for in-transit encryption but not for IAM authentication?

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -161,18 +161,24 @@ def _write_user_data(efs_id, random_file_name):
...
-
-    result = remote_command_executor.run_remote_command("cat /etc/fstab")
-    assert_that(result.stdout).matches(
-        (
-            r".* {mount_dir} nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,"
-            r"timeo=30,retrans=2,noresvport,_netdev 0 0"
-        ).format(mount_dir=mount_dir)
+    if tls:
+        logging.info("Checking efs {0} enables in-transit encryption".format(mount_dir))
+        assert_that(result.stdout).contains("127.0.0.1:/")
```
gmarciani, line 172: I'm not getting why `127.0.0.1:/` would mean that in-transit encryption is enabled?

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -161,18 +161,24 @@ def _write_user_data(efs_id, random_file_name):
...
+    logging.info("Checking efs {0} is successfully mounted in efs mount log".format(mount_dir))
+    result = remote_command_executor.run_remote_command(
+        f'grep -E "Successfully mounted.*{mount_dir}" /var/log/amazon/efs/mount.log'
     )
+    assert_that(result.stdout).contains(mount_dir)
+    logging.info("Checking efs {0} is correctly configured in fstab".format(mount_dir))
+    result = remote_command_executor.run_remote_command("cat /etc/fstab")
+    if tls:
+        assert_that(result.stdout).matches(rf".* {mount_dir} efs tls,iam 0 0")
```
gmarciani: Why in-transit encryption should also enable IAM authorization?
Isn't it the contrary?

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -119,25 +120,38 @@ def test_multiple_efs(
...
-    if os == "alinux2" and scheduler == "slurm":
-        logging.info("Checking EFS utils on Amazon linux 2")
-        _test_efs_utils(
-            remote_command_executor, scheduler_commands, cluster, region, all_mount_dirs, get_efs_ids(cluster, region)
+    if scheduler == "slurm":  # Only Slurm supports compute nodes reboot
+        remote_command_executor, scheduler_commands = _check_efs_after_nodes_reboot(
+            all_mount_dirs, cluster, remote_command_executor, scheduler_commands_factory
         )
 
+    run_benchmarks(remote_command_executor, scheduler_commands)
```
gmarciani, line 134: Juts out of my curiosity: are we actually using the benchmarks executed here?

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -119,25 +120,38 @@ def test_multiple_efs(
...
+
+
+def _check_efs_after_nodes_reboot(all_mount_dirs, cluster, remote_command_executor, scheduler_commands_factory):
+    reboot_head_node(cluster, remote_command_executor)
```
gmarciani, line 138: Does this method return after the node is actually rebooted or it simply request the reboot and returns immediately?

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -119,25 +120,38 @@ def test_multiple_efs(
...
+    for compute_node in compute_nodes:
+        scheduler_commands.reboot_compute_node(compute_node)
+    scheduler_commands.wait_nodes_status("idle", compute_nodes)
+    _check_efs_correctly_mounted_and_shared(all_mount_dirs, remote_command_executor, scheduler_commands)
+    return remote_command_executor, scheduler_commands
+
+
+def _check_efs_correctly_mounted_and_shared(all_mount_dirs, remote_command_executor, scheduler_commands):
+    for mount_dir in all_mount_dirs:
+        test_efs_correctly_mounted(remote_command_executor, mount_dir)
```
gmarciani, line 152: Minor: Let's be consistent with method names: I think here we should have `_test_efs_correctly_mounted` right?

tests/integration-tests/tests/storage/storage_common.py
hanwen-pcluste: Because this integration test will be improved by another PR. In other integration test PR will be created after the support on cluster configuration file parameter is done is done

tests/integration-tests/tests/storage/storage_common.py
hanwen-pcluste: I will add a description in the comment
https://docs.aws.amazon.com/efs/latest/ug/encryption-in-transit.html

tests/integration-tests/tests/storage/storage_common.py
hanwen-pcluste: Fixed

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: No. This code is usable but has been sleeping for some time

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: This is a public function shared with other files

tests/integration-tests/tests/storage/storage_common.py
gmarciani: I'm not getting why we are currently blocked to add the IAM part but we are not for the TLS.
May you please clarify this point?

tests/integration-tests/tests/storage/storage_common.py
gmarciani: Yes, let's add a comment in the code for this check

tests/integration-tests/tests/storage/test_efs.py
gmarciani: If you think we should removed from here, let's add at least a TODO

tests/integration-tests/tests/storage/storage_common.py
gmarciani: I see from the TODO in PR description that you're going to improve the test once the actual changes in cookbook will be in place.
Nevertheless, since both TLS and IAM are actually not in place, why adding TLS and not IAM?

tests/integration-tests/tests/storage/storage_common.py
hanwen-pcluste: I added the comment on line 167

tests/integration-tests/tests/storage/storage_common.py
hanwen-pcluste: The tls was there because there are code testing encryption. What is the reason you prefer either to have both `TLS` and `IAM` logic or none of them?

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: We should keep it here. We can run the benchmark anytime on our testing infrasttructure

tests/integration-tests/tests/storage/test_efs.py
hanwen-pcluste: The function wait two minutes before return

tests/integration-tests/tests/storage/storage_common.py
gmarciani: It's not a concern, it's because I'm not getting why we want one and not the other

gmarciani: Clarified offline. That's fine. It' sjuts because we already have the logic to check for TLS but we don't for IAM

tests/integration-tests/tests/storage/test_efs.py
gmarciani: Rethinking about it: it makes sense to have these sleeping benchmark here since this is a storage related inte test.

gmarciani (approved):

PR #4524, opened 2022-11-09
[3.x] Add a Resource-Prefix Config Option to CLI

himani2411:
### Description of changes

* Providing Resource Prefix config option so that user is able to customize the IAM resource's path prefix and role based prefix.
* Changing schema `ClusterIamSchema` and resource `ClusterIam` to enable user to give ResourcePrefix config option.
* Using IAM naming conventions to create a Regex which will validate the input.

### Tests
* Manual Tests

      1. Created a cluster with wrong Resource Prefix format to see the validation ERROR
      2. Created a cluster with wrong Resource Prefix format and suppressed validators to see the no validation ERROR
      3. Created a Cluster with correct Resource Prefix format (this wont be used for resource creation yet)

* Automated Tests

      1. Unit test cases to validate the Regex handles both positive and negative test scenarios.

### References
* [IAM Naming Conventions](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html)

### Checklist


cli/tests/pcluster/schemas/test_schema_validators.py
```diff
@@ -589,6 +589,7 @@
                     contain = True
         assert_that(contain).is_true()
     else:
+        print(section_dict)
```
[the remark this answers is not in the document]
gmarciani: We should remove the above `print`. Probably put there for debugging during development

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,33 @@ def _validate(self, policy: str):
             self._add_failure(str(e), FailureLevel.ERROR)
 
 
+class IamResourcePrefixValidator(Validator):
+    """
+    Iam Resource Prefix Validator.
+
+    Verify if the Resource Prefix belongs in the permission boundary if given.
```
gmarciani: I think this comment is wrong. The goal of this validator is to verify that the ResourcePrefix is compliant with the expected format

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,33 @@ def _validate(self, policy: str):
...
+    IAM_ROLE_NAME_PREFIX = r"^[a-zA-Z0-9_.,+@=-]+$"
+
+    def _validate(self, resource_prefix: str):
+        try:
+            if not (
+                re.match(IamResourcePrefixValidator.IAM_PATH_NAME_REGEX, resource_prefix)
+                or re.match(IamResourcePrefixValidator.IAM_ROLE_NAME_PREFIX, resource_prefix)
+            ):
+                self._add_failure(
+                    f"Resource Prefix {resource_prefix} provided does not fall under the accepted pattern",
```
gmarciani: Is the error message approved by PM?
I'd rephrase with something like "Unsupported format for ResourcePrefix {resource_prefix} " or "ResourcePrefix {resource_prefix} does not comply with expected format."

Furthermore, I'd provide some suggestions to our customers about the expected format.
For example, we could suggest them to have a look at IAM documentation

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,33 @@ def _validate(self, policy: str):
...
+                    FailureLevel.ERROR,
+                )
+        except Exception as e:
```
gmarciani: What exception do you expect here?

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,33 @@ def _validate(self, policy: str):
...
+            self._add_failure(
+                f"Resource Prefix {resource_prefix} provided does not fall under the accepted pattern",
```
gmarciani: This error message is specifically meant for ResourcePrefix being in an invalid format.
If we have an exception here, it does not necessarily mean that. Right?

gmarciani (commented):
There are some failing checks. May you please fix? Thanks

cli/src/pcluster/validators/iam_validators.py
himani2411: No the message wasnt approved. I will take your suggestion to change it.

cli/src/pcluster/validators/iam_validators.py
himani2411: I have removed the Exception section it was to handle one of the test cases where I had kept the resource prefix as None, however that condition is handled even before we call the Validation.

cli/src/pcluster/validators/iam_validators.py
himani2411: I have removed the Exception section it was to handle one of the test cases where I had kept the resource prefix as None, however that condition is handled even before we call the Validation.

cli/src/pcluster/validators/iam_validators.py
himani2411: The IAM documentation would be confusing right now. I was thinking of maybe later adding the link to our documentation where we mention the exact requirements.

gmarciani: > I was thinking of maybe later adding the link to our documentation where we mention the exact requirements.

I agree. Let's invite the user to refer to our docs. 
In general, I think we should avoid putting specific links in error message because we would couple our code to something that could change

gmarciani: I propose the message: "ResourcePrefix {resource_prefix} is not valid. Please, refer to our public docs for further details."

cli/src/pcluster/validators/iam_validators.py
himani2411: I will change the Comment.

cli/src/pcluster/validators/iam_validators.py
himani2411: Changed the Failure message as requested.

gmarciani:
In this PR we are introducing a new cluster config parameter.
This type of change should be reported into our CHANGELOG.
May you please add it?
Thanks

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,28 @@ def _validate(self, policy: str):
...
+    IAM_ROLE_NAME_PREFIX = r"^[a-zA-Z0-9_.,+@=-]+$"
+
+    def _validate(self, resource_prefix: str):
+        if not (
+            re.match(IamResourcePrefixValidator.IAM_PATH_NAME_REGEX, resource_prefix)
+            or re.match(IamResourcePrefixValidator.IAM_ROLE_NAME_PREFIX, resource_prefix)
+        ):
+            self._add_failure(
+                f"Unsupported format for ResourcePrefix {resource_prefix} "
+                f"Please, refer to our public docs for further details.",
```
gmarciani: 1. I'd say "our official documentation" instead of "public docs" because the latter sounds too "internal".
2. Missing full stop at the end of the first sentence.

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
```diff
@@ -235,6 +235,7 @@ SharedStorage:
       StorageType: HDD  # HDD | SSD
 Iam:
   PermissionsBoundary: arn:aws:iam::aws:policy/boundary
+  ResourcePrefix: /boundary/role-prefix
```
gmarciani: **General best practice**
I suggest to replace `/boundary/role-prefix` with something like `/path/role-prefix` or `/whatever-path/whatever-role-prefix` because we should not suggest the idea that the prefix used in `ResourcePrefix` should match the prefix used by the `PermissionBoundary`.

Let's apply this best practices in every other test parameter in this PR.

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
     mocker.patch("pcluster.aws.iam.IamClient.get_policy", side_effect=expected_get_policy_side_effect)
     actual_failures = AdditionalIamPolicyValidator().execute(policy=policy_arn)
     assert_failure_messages(actual_failures, expected_message)
+
+
+@pytest.mark.parametrize(
+    "resource_prefix, expected_message,expected_failure_level",
```
gmarciani, line 121: We always expect the failure level to be `FailureLevel.ERROR` if a failure occurs. 
so, we can avoid the parameter `expected_failure_level` and assert `FailureLevel.ERROR` in the logic directly.

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+    [
+        (
+            r"\pathprefix\\",
+            "Unsupported format for ResourcePrefix",
```
gmarciani: If possible, it's always better to test the entire error message that is returned, not just a part of it.

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+        ("/prefix/", None, None),
+        ("/somepath/pathprefix/roleprefix", None, None),
+        ("/prefix/", None, None),
+        ("roleprefix", None, None),
+        ("/pathprefix/parallelcluster/", None, None),
+        ("/parallelcluster/pathprefix/", None, None),
+        ("/pathprefix/", None, None),
+        ("role-prefix", None, None),
+        ("@roleprefix", None, None),
+        ("role-prefix", None, None),
```
gmarciani: duplicated case

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+            FailureLevel.ERROR,
+        ),
+        ("/prefix/", None, None),
+        ("/somepath/pathprefix/roleprefix", None, None),
+        ("/prefix/", None, None),
+        ("roleprefix", None, None),
+        ("/pathprefix/parallelcluster/", None, None),
+        ("/parallelcluster/pathprefix/", None, None),
+        ("/pathprefix/", None, None),
+        ("role-prefix", None, None),
```
gmarciani: We can reduce the number of cases to the significative ones.
It would make sense to have one case including all the special characters that are accepted.
For example: `role-prefix@-+=._,`

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+        ("/pathprefix/+roleprefix", None, None),
+        ("/pathprefix/=roleprefix", None, None),
+        ("/pathprefix/role.prefix", None, None),
+        ("/pathprefix/role_prefix", None, None),
+        ("/pathprefix/role,prefix", None, None),
+        ("/path_prefix/@roleprefix/", None, None),
+        ("/path-prefix/role-prefix", None, None),
+        ("/path@prefix/+roleprefix/", None, None),
+        (
+            "/path+prefix/=roleprefix",
```
gmarciani: Minor: why the below cases not in single lines?

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+            "Unsupported format for ResourcePrefix",
+            FailureLevel.ERROR,
+        ),
+        (
+            "#$/pathprefix/anything/",
+            "Unsupported format for ResourcePrefix ",
+            FailureLevel.ERROR,
+        ),
+        (
+            "&&&roleprefix",
```
gmarciani: It's not just role name prefix but also policy name prefix, right?

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -113,3 +115,125 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
+        (
+            ";)*/anything/pathprefix/",
+            "Unsupported format for ResourcePrefix",
+            FailureLevel.ERROR,
+        ),
+        (
+            "#$/pathprefix/anything/",
+            "Unsupported format for ResourcePrefix ",
+            FailureLevel.ERROR,
+        ),
```
gmarciani: We can collapse these cases into a single one

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -57,6 +59,28 @@ def _validate(self, policy: str):
...
 
+class IamResourcePrefixValidator(Validator):
+    """
+    Iam Resource Prefix Validator.
+
+    Verify if the Resource Prefix is compliant with IAM naming conventions
+    """
+
+    IAM_PATH_NAME_REGEX = r"^((/[a-zA-Z0-9_.,+@=-]+)+)/"
+    IAM_ROLE_NAME_PREFIX = r"^[a-zA-Z0-9_.,+@=-]+$"
```
gmarciani: It's not just a role name prefix, but also policy name prefix, right?

cli/src/pcluster/validators/iam_validators.py
himani2411: Sure

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
himani2411: I will change it to /path-prefix/role-prefix

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Changing this string would lead me to make changes on the Function level as well, which would give an unresolved error as it wont compare with the parameter we are giving but the defined Enum we have.

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: No, the user can below inputs
-> Both Path prefix and role name prefix
-> Just Path prefix
-> Just role name prefix

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: This case was giving an exception while comparing due to the the escape character given and which is part of the expected message.

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Inputs

Both: ( Start with / but Shouldnt end with / )
/pathprefix/roleprefix 
Just Path prefix: ( Start and end with /)
/pathprefix/
Just Rolename: (No /)
roleprefix

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
gmarciani: Not clear to me: when we put `ResourcePrefix` is `/blablabla/tiptoe`, then `blablabla` will be the path or the path prefix?

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Will change that

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Sure

cli/src/pcluster/validators/iam_validators.py
himani2411: https://github.com/aws/aws-parallelcluster/pull/4524#discussion_r1019246026

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Will rename this is to name prefix

cli/src/pcluster/validators/iam_validators.py
himani2411: Will rename IAM_ROLE_NAME_PREFIX to IAM_NAME_PREFIX_REGEX

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Will add a comment for TO DO

cli/tests/pcluster/validators/test_all_validators/test_scheduler_plugin_all_validators_are_called/scheduler_plugin_1.yaml
himani2411: Will rename to /path-prefix/name-prefix

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: After a call, we have resolved to keep the test function as it is.

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Sure

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Remved Duplication

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 CHANGELOG
 =========
+3.4.0
+-----
+
+**ENHANCEMENTS**
+- Add new configuration parameter `ResourcePrefix` in `Iam` to specify a Path prefix and Name prefix for IAM resources.
```
gmarciani: According to our best practices the line should be something like:

"Add new configuration parameter in `Iam/ResourcePrefix` to specify a prefix for path and name of IAM resources created by ParallelCluster."

himani2411: Made changes

hanwen-pcluste (approved):

PR #4541, opened 2022-11-14
[Multi-AZ] Modify networking validators impacted by Multi-AZ

jdeamicis:
### Description of changes
* Modify Networking validators impacted by the Multi-AZ functionality in the Slurm and Scheduler Plugin scenarios:
  * Add better description for `SubnetsValidator`.
  * Add new `QueueSubnetsValidator` to check
    * if there are duplicate IDs in `subnet_ids`;
    * if all subnets in a queue are from different AZs;
  * Modify `SingleSubnetValidator` to:
    * remove check on unique subnet across all compute queues;   
    * check if compute resources with single instance type have one single subnet;
  * Make the implementation of `compute_subnet_ids` more readable

### Tests
* Added or modified unit tests for the validator impacted by this PR
* No change was introduced in the core functioning of the product, so except for the validation phase the behavior of the product is the same.

### Checklist

### DIFF
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -77,7 +77,7 @@ class QueueSubnetsValidator(Validator):
         # Test if there are duplicate IDs in subnet_ids
         if len(set(subnet_ids)) < len(subnet_ids):
             self._add_failure(
-                "The list of subnet IDs specified in queue {0} contains duplicate IDs.".format(queue_name),
+                "SubnetIds specified in queue {0} contains duplicate subnet IDs.".format(queue_name),
                 FailureLevel.ERROR,
             )
 
@@ -91,8 +91,8 @@ class QueueSubnetsValidator(Validator):
 
             if len(az_set) < len(subnet_ids):
                 self._add_failure(
-                    "Some of the subnet IDs specified in queue {0} are in the same AZ. Please make sure all subnets "
-                    "are in different AZs.".format(queue_name),
+                    "SubnetIds specified in queue {0} contains multiple subnets in the same AZ. "
+                    "Please make sure all subnets are in different AZs.".format(queue_name),
                     FailureLevel.ERROR,
                 )
 
@@ -117,9 +117,10 @@ class SingleSubnetValidator(Validator):
     def _validate(self, queue_name, subnet_ids):
         if len(subnet_ids) > 1:
             self._add_failure(
-                "At least one compute resource in queue {0} use a single instance type. "
-                "Multi AZ configuration does not support single instance type. "
-                "Please make sure to use the multiple instance type allocation.".format(queue_name),
+                "At least one compute resource in queue {0} uses a single instance type. "
+                "Multiple subnets configuration is not supported for single instance type, "
+                "please use the Instances configuration parameter for multiple instance type "
+                "allocation.".format(queue_name),
                 FailureLevel.ERROR,
             )
 
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -112,12 +112,16 @@ class ElasticIpValidator(Validator):
 
 
 class SingleSubnetValidator(Validator):
-    """Validate all queues reference the same subnet."""
+    """Validate only one subnet is used for compute resources with single instance type."""
 
-    def _validate(self, queues_subnets):
-        subnet_ids = {tuple(set(queue_subnets)) for queue_subnets in queues_subnets}
+    def _validate(self, queue_name, subnet_ids):
         if len(subnet_ids) > 1:
-            self._add_failure("The SubnetId used for all of the queues should be the same.", FailureLevel.ERROR)
+            self._add_failure(
+                "At least one compute resource in queue {0} use a single instance type. "
+                "Multi AZ configuration does not support single instance type. "
+                "Please make sure to use the multiple instance type allocation.".format(queue_name),
+                FailureLevel.ERROR,
+            )
 
 
 class MultiAzPlacementGroupValidator(Validator):
--- a/cli/src/pcluster/validators/networking_validators.py
+++ b/cli/src/pcluster/validators/networking_validators.py
@@ -67,11 +67,21 @@ class QueueSubnetsValidator(Validator):
     """
     Queue Subnets validator.
 
+    Check that there is no duplicate subnet id in the subnet_ids list.
     Check that subnets in a queue belong to different AZs (EC2 Fleet requests do not support multiple subnets
     in the same AZ).
     """
 
     def _validate(self, queue_name, subnet_ids: List[str]):
+
+        # Test if there are duplicate IDs in subnet_ids
+        if len(set(subnet_ids)) < len(subnet_ids):
+            self._add_failure(
+                "The list of subnet IDs specified in queue {0} contains duplicate IDs.".format(queue_name),
+                FailureLevel.ERROR,
+            )
+
+        # Test if the subnets are all in different AZs
         try:
             subnets = AWSApi.instance().ec2.describe_subnets(subnet_ids=subnet_ids)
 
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/validators/networking_validators.py
```diff
@@ -57,6 +63,43 @@ def _validate(self, subnet_ids: List[str]):
...
+        # Test if there are duplicate IDs in subnet_ids
+        if len(set(subnet_ids)) < len(subnet_ids):
+            self._add_failure(
+                "SubnetIds specified in queue {0} contains duplicate subnet IDs.".format(queue_name),
+                FailureLevel.ERROR,
+            )
+
+        # Test if the subnets are all in different AZs
+        try:
+            subnets = AWSApi.instance().ec2.describe_subnets(subnet_ids=subnet_ids)
```
EddyMM: This logic of getting the Subnet AZs can moved into the `ec2.py` file for re-use or extend the existing implementation [get_subnet_avail_zone](https://github.com/EddyMM/aws-parallelcluster/blob/50264bbdbf1c540dbe916db7e007b8c7c0e7b6f1/cli/src/pcluster/aws/ec2.py#L113-L113).

I've seen a similar case in the EFA Validator: [az_list](https://github.com/aws/aws-parallelcluster/pull/4544/files#diff-bd4426f9ee10414432866ef60f1b6a41dab292da908dc94ded28ffee6cc66cb3R1912)

EddyMM: I think it would be a better experience if the customer was informed about which subnets are in the same AZ, especially since we have this information within this code block.

For duplicate subnets, it's easy for the user to see the duplicate (since there are few AZs/Subnets that can be included).

jdeamicis: > This logic of getting the Subnet AZs can moved into the `ec2.py` file for re-use or extend the existing implementation [get_subnet_avail_zone](https://github.com/EddyMM/aws-parallelcluster/blob/50264bbdbf1c540dbe916db7e007b8c7c0e7b6f1/cli/src/pcluster/aws/ec2.py#L113-L113).
> 
> I've seen a similar case in the EFA Validator: [az_list](https://github.com/aws/aws-parallelcluster/pull/4544/files#diff-bd4426f9ee10414432866ef60f1b6a41dab292da908dc94ded28ffee6cc66cb3R1912)

This has been addressed in #4547, on top of which this branch has been rebased.

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -57,6 +63,39 @@ def _validate(self, subnet_ids: List[str]):
...
+                FailureLevel.ERROR,
+            )
+
+        # Test if the subnets are all in different AZs
+        try:
+            az_set = {subnet_id_az_mapping[subnet_id] for subnet_id in subnet_ids}
+
+            if len(az_set) < len(subnet_ids):
+                self._add_failure(
+                    "SubnetIds specified in queue {0} contains multiple subnets in the same AZ. "
```
EddyMM: From previous revision

> I think it would be a better experience if the customer was informed about which subnets are in the same AZ, especially since we have this information within this code block.

It could be cumbersome for a customer to check each subnet one by one again to find out which ones are in the same AZ.

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -69,12 +108,17 @@ def _validate(self, elastic_ip: Union[str, bool]):
 
 
 class SingleSubnetValidator(Validator):
-    """Validate all queues reference the same subnet."""
+    """Validate only one subnet is used for compute resources with single instance type."""
 
-    def _validate(self, queues_subnets):
-        subnet_ids = {tuple(set(queue_subnets)) for queue_subnets in queues_subnets}
+    def _validate(self, queue_name, subnet_ids):
```
EddyMM, line 128: A validator should be able to determine the correct/incorrect state using only the information it is provided.
This validator already assumes that the queue is being used with a Single Instance Type (since it's only registered when there is a Single Instance Type usage).

I think validation logic should be local to the validator. The list of ComputeResources can be passed to this validator and it should then determine what's valid/invalid.

EddyMM: This will also allow the validator to communicate which Compute Resource is invalid.

cli/src/pcluster/validators/networking_validators.py
jdeamicis: > I think it would be a better experience if the customer was informed about which subnets are in the same AZ, especially since we have this information within this code block.
> 
> For duplicate subnets, it's easy for the user to see the duplicate (since there are few AZs/Subnets that can be included).

I added both lists, for more clarity.

cli/src/pcluster/validators/networking_validators.py
jdeamicis: I addressed this (see [above](https://github.com/aws/aws-parallelcluster/pull/4541#discussion_r1025403078))

cli/src/pcluster/validators/networking_validators.py
jdeamicis: I had to do this to avoid a circular import between `cluster_config.py` and `networking_validators.py`. We can discuss about this.

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -57,6 +64,58 @@ def _validate(self, subnet_ids: List[str]):
...
+            )
+
+        # Test if the subnets are all in different AZs
+        else:
+            try:
+                az_set = {subnet_id_az_mapping[subnet_id] for subnet_id in subnet_ids}
+                if len(az_set) < len(subnet_ids):
+
+                    # Find the AZs with multiple subnets
+                    azs_with_multiple_subnets = {}
```
EddyMM: Could we move this logic into a separate function, named `get_subnets_sharing_an_az` perhaps.

Also, could it be simplified to avoid the nested loop? Would this logic work for this use-case?
```
def get_subnets_sharing_an_az(subnet_ids):
    subnets_sharing_az = set()
    visited_subnet = {}

    for subnet_id in subnet_ids:
        az = subnet_id_az_mapping[subnet_id]

        if visited_subnet.get(subnet_id) == az:
            subnets_sharing_az.add(subnet_id)
        else:
            visited_subnet[subnet_id] = az

    return list(subnets_sharing_az)
```

jdeamicis: We discussed about this offline and decided to implement a new property `az_subnet_ids_mapping` in the `_QueueNetworking` class. This property is a dictionary containing the list of subnets belonging to each AZ used in the queue.

EddyMM (dismissed):

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -57,6 +64,55 @@ def _validate(self, subnet_ids: List[str]):
...
+            self._add_failure(
+                "The following subnet ids are specified multiple times "
+                "in queue {0}: {1}.".format(
+                    queue_name,
+                    ", ".join(duplicate_ids),
+                ),
+                FailureLevel.ERROR,
+            )
+
+        # Test if the subnets are all in different AZs
```
NSsirena, line 94: Are we sure these tests are mutually exclusive ?

It seems possible to have duplicated ID in the list of subnets and also have more than one subnet-id in the same AZ.

jdeamicis: No, they are not mutually exclusive. I just don't want the duplicate subnet IDs to trigger also the second part of the validator. Without this `if ... else` the duplicate subnet IDs would also be counted as multiple subnets in the same AZ, which to me doesn't make sense.

jdeamicis: But I see what you mean. I would need to remove the duplicate subnet IDs and pass this intermediate list of subnet IDs to the next phase of the validator.

Is this worth the effort? The validation would fail anyway.

jdeamicis: Good catch @NSsirena. I changed implementation so that now it's possible to have both validation messages at the same time.

EddyMM (approved):

PR #4544, opened 2022-11-15
EFA MultiAZ Validator (3.x)

NSsirena:
### Description of changes
* Added two properties to _CommonQueue
  * az_list that returns the list of AZ (without duplicates) associated to the subnets defined in the networking section 
  * multi_az_enabled that is True if the Queue extends over more than one AZ
* Added a new Validator that will fail if both MultiAZ and EFA are enabled at the same time
* Changed the EfaPlacementGroupValidator to suppress suggestions about PlacementGroups when MultiAZ is enabled 

### Tests
* Unit tests
* Dryrun pcluster create with configurations where
    * Both EFA and MultiAZ were enabled (Failed) - No suggestions for PlacementGroups with EFA
    * EFA was enabled but in a Single AZ queue (Success)
    * EFA was disabled and MultiAZ was enabled (Success)

### References
-

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -77,6 +77,7 @@ from pcluster.validators.cluster_validators import (
     DeletionPolicyValidator,
     DuplicateMountDirValidator,
     DuplicateNameValidator,
+    EfaMultiAzValidator,
     EfaOsArchitectureValidator,
     EfaPlacementGroupValidator,
     EfaSecurityGroupValidator,
@@ -1929,6 +1930,7 @@ class _CommonQueue(BaseQueue):
         self.capacity_reservation_target = capacity_reservation_target
         self.compute_resources = compute_resources
         self.networking = networking
+        self.__az_list = None
 
     @property
     def instance_role(self):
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1929,6 +1930,7 @@ def __init__(
         self.capacity_reservation_target = capacity_reservation_target
         self.compute_resources = compute_resources
         self.networking = networking
+        self.__az_list = None
```
jdeamicis, line 1933: Do we really need this dunder attribute here? We are not going to re-define `az_list` in children classes, are we?

Edit: oh, you are using it as a sort of cache. Just beware that the double underscore in front of an attribute in Python has a [specific meaning](https://www.hacksoft.io/blog/underscores-dunders-and-everything-nice#name-mangling). We could call it `_az_list` if we want to keep it.

Edit2: maybe you want to make it completely unaccessible/uneditable?

jdeamicis (commented):
I left one comment about a dunder attribute defined in this PR.

cli/src/pcluster/config/cluster_config.py
jdeamicis: OK, we discussed about this and I'm fine with this implementation.

jdeamicis (dismissed):

jdeamicis (approved):

PR #4547, opened 2022-11-15
Add subnet_id -> az mapping property at queue level

jdeamicis:
### Description of changes
* Add subnet_id -> az mapping property at queue level to be used for various validators in Multi-AZ scenarios.

### Tests
* Add unit test to cover the newly implemented `get_subnets_az_mapping` method.
* Modify `get_describe_subnets_mocked_request` to allow it to return information about the subnet's AZ.

### References
* This PR is needed by #4541 #4544 

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -625,7 +625,7 @@ class _QueueNetworking(_BaseNetworking):
         self.subnet_ids = Resource.init_param(subnet_ids)
 
     @property
-    def queue_subnet_id_az_mapping(self):
+    def subnet_id_az_mapping(self):
         """Map queue subnet ids to availability zones."""
         return AWSApi.instance().ec2.get_subnets_az_mapping(self.subnet_ids)
 
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -624,6 +624,11 @@ class _QueueNetworking(_BaseNetworking):
         self.assign_public_ip = Resource.init_param(assign_public_ip)
         self.subnet_ids = Resource.init_param(subnet_ids)
 
+    @property
+    def queue_subnet_id_az_mapping(self):
+        """Map queue subnet ids to availability zones."""
+        return AWSApi.instance().ec2.get_subnets_az_mapping(self.subnet_ids)
+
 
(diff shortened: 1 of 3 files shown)


NSsirena (dismissed):
LGTM

cli/src/pcluster/config/cluster_config.py
```diff
@@ -622,6 +622,11 @@ def __init__(self, subnet_ids: List[str], assign_public_ip: str = None, **kwargs
         self.assign_public_ip = Resource.init_param(assign_public_ip)
         self.subnet_ids = Resource.init_param(subnet_ids)
 
+    @property
+    def queue_subnet_id_az_mapping(self):
```
EddyMM: NIT: Following the naming convention we use, this should be `subnet_id_az_mapping` since it's already in the context of "Queue" Networking.

cli/tests/pcluster/aws/test_ec2.py
```diff
@@ -334,10 +334,14 @@ def test_run_instances_dryrun(boto3_stubber, error_code, raise_exception):
         Ec2Client().run_instances(**kwargs)
 
 
-def get_describe_subnets_mocked_request(subnets, state):
+def get_describe_subnets_mocked_request(subnets, state, avail_zones: dict):
     return MockedBoto3Request(
```
EddyMM, line 338: Optional:
For this mock, it seems we return subnet information. We could have a namedtuple `SubnetResponse(subnet_id, state, availability_zone)`
This way we can pass the `subnets` and a map `{"subnet_id": SubnetResponse(subnet_id, state, availability_zone), ...}`. This is just to avoid passing the state as a string, then using a dict/map for availability_zone.

cli/src/pcluster/config/cluster_config.py
jdeamicis: Done.

cli/tests/pcluster/aws/test_ec2.py
jdeamicis: I see what you mean and I like the idea of the namedtuple.

I had a quick try and realized that this would probably require a bit of thought about the `test_describe_subnets_cache` function, which relies on the fact that you can do multiple requests for the same subnet (e.g. with different states).

If it's fine to you I would leave it as it is for now.

EddyMM (approved):

PR #4548, opened 2022-11-15
[3.x] Add Resource Prefix in Cluster Creation for IAM resources

himani2411:
### Description of changes
Changing Cluster Stack Creation to add Path and Name Prefix when creating IAM Resources and Lambda Resources 
(Role, Policy and Instance Profile)

### Tests
* Unit Test cases to Check Slurm Policies, Instance Profiles,S3 Policies, Parallel Cluster Policies, Roles and Lambda Roles have Paths and Name prefix.
* Manual Tests to check unit test cases and revert to Original Role Names when No Resource Prefix is provided.

### Checklist


cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
     return queue_security_groups
 
 
+def add_cluster_iam_resource_prefix(stack_name, config, name: str, iam_type: str):
+    """Return a path and Name prefix from the Resource prefix config option."""
+    full_resource_path = None
+    full_resource_name = None
+    if config.iam and config.iam.resource_prefix:
+        iam_path, iam_name_prefix = get_path_n_name_prefix_from_iam_resource_prefix(config.iam.resource_prefix)
```
dreambeyondorange: it might be easier to just call this function `split_resource_prefix`

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+        if iam_name_prefix:
+            if iam_type in ["AWS::IAM::Role", "AWS::IAM::InstanceProfile"]:
+                full_resource_name = iam_name_prefix + name + "-"
```
dreambeyondorange: You should break this up to get all the individual pieces first and assign it once to `full_resource_name` e.g. 

```
part_a = x
part_b = y
part_c = z + a + f
full_name = part_a + part_b + part_c
```

I think this will be easier to digest

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+                # Add a region for making the hash and Name Unique Globally
+                full_resource_name = (
+                    full_resource_name
+                    + hashlib.sha256((full_resource_name + stack_name + iam_type + config.region).encode("utf-8"))
+                    .hexdigest()[:12]
+                    .capitalize()
+                )
+
+            elif iam_type == "AWS::IAM::Policy":
```
dreambeyondorange: Why are we doing this for Policies?  It seems odd not to specify a name for the policy.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+    )
+    _, policy_name_prefix = add_cluster_iam_resource_prefix(
+        scope.config.cluster_name, scope.config, None, iam_type="AWS::IAM::Policy"
+    )
+    if role_path:
+        lambda_path = role_path
+    if policy_name_prefix:
+        policy_name = policy_name_prefix + "-" + policy_name
+    if role_name:
+        role_id = f"{function_id}Role"
```
dreambeyondorange: instead of reassigning the `role_id` we should only assign a value once with an if/else statement

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+    policy_name = "LambdaPolicy"
+    # Need to reduce calls to this function
+    role_path, role_name = add_cluster_iam_resource_prefix(
+        scope.config.cluster_name, scope.config, name=f"{function_id}Role", iam_type="AWS::IAM::Role"
+    )
+    _, policy_name_prefix = add_cluster_iam_resource_prefix(
+        scope.config.cluster_name, scope.config, None, iam_type="AWS::IAM::Policy"
+    )
+    if role_path:
+        lambda_path = role_path
```
dreambeyondorange: instead of reassigning the `lambda_path` we should only assign a value once with an if/else statement

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+    if policy_name_prefix:
+        policy_name = policy_name_prefix + "-" + policy_name
```
dreambeyondorange: instead of reassigning the `policy_name` we should only assign a value once with an if/else statement

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,59 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+
+    return full_resource_path, full_resource_name
+
+
 def add_lambda_cfn_role(scope, function_id: str, statements: List[iam.PolicyStatement], has_vpc_config: bool):
     """Return a CfnRole to be used for a Lambda function."""
+    role_id = f"{function_id}FunctionExecutionRole"
+    lambda_path = IAM_ROLE_PATH
+    policy_name = "LambdaPolicy"
+    # Need to reduce calls to this function
```
dreambeyondorange: Remove the comment if you aren't planning to do something about it in the PR.  You could also consider breaking the function up into two, since you only need half of it for the policy name

cli/src/pcluster/templates/cdk_builder_utils.py
dreambeyondorange: Also, why are we assigning the full policy name here instead of in the function `add_cluster_iam_resource_prefix`?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -324,25 +367,44 @@ def _add_role_and_policies(self, node: Union[HeadNode, BaseQueue], name: str):
...
             assume_role_policy_document=get_assume_role_policy_document("ec2.{0}".format(Stack.of(self).url_suffix)),
+            role_name=role_name,
         )
 
     def _add_pcluster_policies_to_role(self, role_ref: str, name: str):
+        policy_name = "parallelcluster"
+        _, policy_name_prefix = add_cluster_iam_resource_prefix(
+            self._config.cluster_name, self._config, name, iam_type="AWS::IAM::Policy"
+        )
+        if policy_name_prefix:
```
dreambeyondorange: use if/else to avoid reassigning the variable

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -397,12 +459,19 @@ def _add_s3_access_policies_to_role(self, node: Union[HeadNode, BaseQueue], role
                 else:
                     read_only_s3_resources.append(arn)
 
+        policy_name = "S3Access"
+        _, policy_name_prefix = add_cluster_iam_resource_prefix(
+            self._config.cluster_name, self._config, name, iam_type="AWS::IAM::Policy"
+        )
+        if policy_name_prefix:
```
dreambeyondorange: use if/else

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -422,8 +491,10 @@ def _add_s3_access_policies_to_role(self, node: Union[HeadNode, BaseQueue], role
                 )
             )
 
-    def _cluster_scoped_iam_path(self):
+    def _cluster_scoped_iam_path(self, iam_path=None):
         """Return a path to be associated IAM roles and instance profiles."""
```
dreambeyondorange: can you fix this comment?  `associated with IAM`

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -422,8 +491,10 @@ def _add_s3_access_policies_to_role(self, node: Union[HeadNode, BaseQueue], role
...
+        if iam_path:
```
dreambeyondorange, line 475: It's not required, but I think using if/else here is clearer

cli/src/pcluster/templates/slurm_builder.py
```diff
@@ -131,6 +134,8 @@ def _add_policies_to_compute_node_role(self, node_name, role):
             },
         ]
         policy_name = "parallelcluster-slurm-compute"
+        if policy_name_prefix:
```
dreambeyondorange: please review all instances where you assign a variable and then do an if check to re-assign it to make it an if/else

cli/src/pcluster/utils.py
```diff
@@ -293,6 +293,23 @@ def get_resource_name_from_resource_arn(resource_arn):
...
 
 
+def get_path_n_name_prefix_from_iam_resource_prefix(resource_prefix):
+    if resource_prefix:
+        if resource_prefix.endswith("/"):
+            # No Name Prefix
+            return resource_prefix, None
+        elif resource_prefix.startswith("/") and not resource_prefix.endswith("/"):
+            # Contains Both
+            if resource_prefix.rfind("/") != -1:
```
dreambeyondorange: we should write this so that we don't need so many if statements.  rfind will return -1 if there is no slash right?  and resource_prefix[:split_index] will return an empty string.  If its empty you can return None.  So I think it should be possible to do an rfind first and return the split result (assigning None when the split returns an empty string)

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -79,6 +79,12 @@ def _validate(self, resource_prefix: str):
                 f"Please refer to our official documentation for further details.",
                 FailureLevel.ERROR,
             )
+        if len(resource_prefix) > 31:
```
dreambeyondorange: why a hardcoded 31?  We should use shared variables to define the limits so that other modules can access the same info.

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -153,10 +153,21 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
...
             FailureLevel.ERROR,
         ),
+        (
+            "/path_.,+@=-prefix/name_.,+@=-prefix",
+            "Length of ResourcePrefix /path_.,+@=-prefix/name_.,+@=-prefix should be less than 30 characters."
+            " Please refer to our official documentation for further details.",
+            FailureLevel.ERROR,
+        ),
+        (
+            "/longerPath/pathprefix/nameprefix",
```
dreambeyondorange: Why is the path included in the limit for the name prefix?  This path won't be reflected in the IAM names, so should not result in an error

cli/tests/pcluster/validators/test_iam_validators.py
```diff
@@ -153,10 +153,21 @@ def test_additional_iam_policy_validator(mocker, policy_arn, expected_get_policy
             "Unsupported format for ResourcePrefix //",
             FailureLevel.ERROR,
         ),
+        (
+            "/path_.,+@=-prefix/name_.,+@=-prefix",
```
dreambeyondorange: The path should not affect the name limit

dreambeyondorange (changes requested):
Great work Himani!  I left a few comments.

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Sure

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Thank you for point thing this.
I have change the logic and the error message to reflect the Length Limit on Name prefix.

cli/tests/pcluster/validators/test_iam_validators.py
himani2411: Thank you for point thing this.
I have change the logic and the error message to reflect the Length Limit on Name prefix.

cli/src/pcluster/validators/iam_validators.py
himani2411: I have used a constant a Limit of 30

dreambeyondorange: What I mean is, this should be a variable defined in common.py or something similar and referenced from there instead of a hardcoded number here

cli/src/pcluster/utils.py
himani2411: Done used the rfind rather than using startswith and endswith

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Made it more readable friendly

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Made the naming convention for all the IAM resources similar, rather than differentiating using Type

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Assigning the value just once

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Assigning the value just once. And changed the implementation to have similar naming convention for all IAM resources

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Assigning the value just once

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Removed the comment. However, using the same function with same naming convention for all IAM resource

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Assigning the value just once

cli/src/pcluster/templates/slurm_builder.py
himani2411: Assigning the value just once

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: avoided reassigning the variable

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Change the function to have a consistent styling

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,54 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+        scope.config.cluster_name, scope.config, name=f"{function_id}Role", iam_type="AWS::IAM::Role"
+    )
+    _, policy_name_prefix = add_cluster_iam_resource_prefix(
+        scope.config.cluster_name, scope.config, "LambdaPolicy", iam_type="AWS::IAM::Policy"
+    )
+    if role_path:
+        lambda_path = role_path
+    else:
+        lambda_path = IAM_ROLE_PATH
+    if policy_name_prefix:
```
dreambeyondorange: similar ternary option here

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -250,17 +251,54 @@ def get_queue_security_groups_full(managed_compute_security_group: ec2.CfnSecuri
...
+        policy_name = policy_name_prefix
+    else:
+        policy_name = "LambdaPolicy"
+    if role_name:
```
dreambeyondorange: ternary as well

cli/src/pcluster/constants.py
```diff
@@ -200,3 +200,6 @@
 DEFAULT_COMPUTE_CONSOLE_LOGGING_MAX_SAMPLE_SIZE = 100
 
 LAMBDA_VPC_ACCESS_MANAGED_POLICY = "arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
+
+IAM_NAME_PREFIX_LENGTH_LIMIT = 30
+IAM_PATH_PREFIX_LENGTH_LIMIT = 512
```
dreambeyondorange: I would probably just call this the path rather than the path prefix.  It's the whole path right?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -396,13 +442,16 @@ def _add_s3_access_policies_to_role(self, node: Union[HeadNode, BaseQueue], role
...
+            self._config.cluster_name, self._config, "S3Access", iam_type="AWS::IAM::Policy"
+        )
 
         s3_access_policy = iam.CfnPolicy(
             Stack.of(self),
             name,
             policy_document=iam.PolicyDocument(statements=[]),
             roles=[role_ref],
-            policy_name="S3Access",
+            policy_name=policy_name_prefix or "S3Access",
```
dreambeyondorange: I wonder if we should be appending the prefix to the defaults if they exist?  Seems useful to always have `S3Access` in the name for example

cli/src/pcluster/utils.py
```diff
@@ -293,6 +293,24 @@ def get_resource_name_from_resource_arn(resource_arn):
     return resource_arn.rsplit("/", 1)[-1] if resource_arn else ""
 
 
+def split_resource_prefix(resource_prefix):
```
dreambeyondorange, line 296: nice!

cli/src/pcluster/validators/iam_validators.py
```diff
@@ -79,6 +80,19 @@ def _validate(self, resource_prefix: str):
...
+        iam_path_prefix, iam_name_prefix = split_resource_prefix(resource_prefix)
+        if iam_name_prefix and (len(iam_name_prefix) > IAM_NAME_PREFIX_LENGTH_LIMIT):
+            self._add_failure(
+                f"Length of Name Prefix {iam_name_prefix} should be upto {IAM_NAME_PREFIX_LENGTH_LIMIT} characters. "
+                f"Please refer to our official documentation for further details.",
+                FailureLevel.ERROR,
+            )
+        if iam_path_prefix and (len(iam_path_prefix) > IAM_PATH_PREFIX_LENGTH_LIMIT):
+            self._add_failure(
+                f"Length of Path {iam_path_prefix} should be upto {IAM_PATH_PREFIX_LENGTH_LIMIT} characters. "
```
dreambeyondorange: `must be less than` instead of `should be upto`

dreambeyondorange (dismissed):

cli/src/pcluster/constants.py
himani2411: Sure

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: i will change the Variable name as it is no longer a prefix but full policy name that I am returning due to implementation change for add_cluster_iam_resource_prefix()

cli/src/pcluster/validators/iam_validators.py
himani2411: Done

cli/src/pcluster/utils.py
himani2411: Thanks

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: Or do you mean not keeping the prefix to policy name at all.

dreambeyondorange (approved):

PR #4554, opened 2022-11-17
Add failure reason to describe-cluster output when cluster creation fail with headnode wait condition creation fail

chenwany:
### Description of changes
* Add failure_reason to describe-cluster output when cluster creation fail 
* The failure reason is retrieved from headnode wait condition, when headnode wait condition creation fail, the failure reason is going to be displayed in `describe-cluster` result.

### Tests
* Unit test
* Manual test
```
{
  "creationTime": "2022-11-09T18:14:35.452Z",
  "headNode": {
    "launchTime": "2022-11-09T18:17:51.000Z",
    "instanceId": "...",
    "publicIpAddress": "...",
    "instanceType": "c5.xlarge",
    "state": "running",
    "privateIpAddress": "192.168.54.215"
  },
  "version": "3.3.0",
  "clusterConfiguration": ...,
  "tags": [
    {
      "value": "3.3.0",
      "key": "parallelcluster:version"
    },
    {
      "value": "dfsm-test",
      "key": "parallelcluster:cluster-name"
    }
  ],
  "cloudFormationStackStatus": "ROLLBACK_IN_PROGRESS",
  "clusterName": "dfsm-test",
  "computeFleetStatus": "RUNNING",
  "failureReason": "WaitCondition received failed message: 'This AMI was created with aws-parallelcluster-cookbook-3.3.0, but is trying to be used with aws-parallelcluster-cookbook-3.4.0b1. Please either use an AMI created with aws-parallelcluster-cookbook-3.4.0b1 or change your ParallelCluster to aws-parallelcluster-cookbook-3.3.0' for uniqueId: i-01fdb3c2b73182693",
  "cloudformationStackArn": "xxx",
  "lastUpdatedTime": "2022-11-09T18:14:35.452Z",
  "region": "us-east-1",
  "clusterStatus": "CREATE_FAILED",
  "scheduler": {
    "type": "slurm"
  }
}
```

### References

### Checklist


cli/src/pcluster/models/cluster_resources.py
```diff
@@ -74,6 +74,22 @@ def batch_compute_environment(self):
...
+        wait_conditiaon_failure_events = None
+        for chuck in stack_events:
+            if wait_conditiaon_failure_events:
+                break
+            for event in chuck:
+                if (
+                    event.get("ResourceType") == "AWS::CloudFormation::WaitCondition"
+                    and event.get("ResourceStatus") == "CREATE_FAILED"
+                ):
+                    wait_conditiaon_failure_events = event
```
davprat: Typo in `wait_conditiaon_failure_events`.

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -74,6 +74,22 @@ def batch_compute_environment(self):
         """Return Batch compute environment."""
         return self._get_output("BatchComputeEnvironmentArn")
 
+    def get_failure_reason(self):
+        """Reason of the failure when the cluster_status is in CREATE_FAILED."""
+        stack_events = get_all_stack_events(self.name)
+        wait_conditiaon_failure_events = None
+        for chuck in stack_events:
```
davprat: Is `chuck` meant to be `chunk`? I do like `chuck` better though.

davprat (dismissed):
Other than a couple of potential typos, it looks good to me.

cli/src/pcluster/models/cluster_resources.py
charlesg3: What about just returning when you doing the right event?

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -74,6 +74,22 @@ def batch_compute_environment(self):
...
+            if wait_conditiaon_failure_events:
+                break
+            for event in chuck:
+                if (
+                    event.get("ResourceType") == "AWS::CloudFormation::WaitCondition"
```
charlesg3: I would break out this predicate so that the logic for determining the correct event is separate (as a local function) which will provide better self-documenting code

charlesg3: If you break out the predicate you can use the following to functionally find the first match:

```
found = next(filter(is_failed_wait, chunk), None)
…
```

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -74,6 +74,22 @@ def batch_compute_environment(self):
         """Return Batch compute environment."""
         return self._get_output("BatchComputeEnvironmentArn")
 
+    def get_failure_reason(self):
+        """Reason of the failure when the cluster_status is in CREATE_FAILED."""
+        stack_events = get_all_stack_events(self.name)
```
charlesg3: You can use https://docs.python.org/3/library/itertools.html#itertools.chain to avoid dealing with the chunks by lazily flattening the result at this call site

cli/src/pcluster/models/cluster_resources.py
chenwany: Fixed the typo.
>What about just returning when you doing the right event?

Good idea!

cli/src/pcluster/models/cluster_resources.py
chenwany: oops, fix

cli/src/pcluster/models/cluster_resources.py
chenwany: this is a good idea! Changed

cli/src/pcluster/models/cluster_resources.py
chenwany: 👍 lesson learned

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -74,6 +75,21 @@ def batch_compute_environment(self):
...
+
+        def _is_failed_wait(event):
+            if (
+                event.get("ResourceType") == "AWS::CloudFormation::WaitCondition"
+                and event.get("ResourceStatus") == "CREATE_FAILED"
+            ):
+                return True
+            return False
+
+        stack_events = list(itertools.chain.from_iterable(get_all_stack_events(self.name)))
```
davprat, line 89: Question (non-blocking): Does this need to be realized into a list? I'm not sure, but it seems like you could pass the result of `from_iterable` directly to `filter`?

davprat (approved):

charlesg3 (approved):

cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -231,8 +233,9 @@ def describe_cluster(cluster_name, region=None):
         cloudformation_stack_arn=cfn_stack.id,
         last_updated_time=to_utc_datetime(cfn_stack.last_updated_time),
         region=os.environ.get("AWS_DEFAULT_REGION"),
-        cluster_status=cloud_formation_status_to_cluster_status(cfn_stack.status),
+        cluster_status=cluster_status,
         scheduler=Scheduler(type=cluster.stack.scheduler, metadata=cluster.get_plugin_metadata()),
+        failure_reason=cfn_stack.get_failure_reason() if cluster_status == ClusterStatus.CREATE_FAILED else None,
```
eantonin, line 238: question, non blocking: is this actually the first time we use `failure_reason` field? Am I right saying that before this change that field would never be populated?

PR #4555, opened 2022-11-17
Add MultiAzPlacementGroupValidator that fails if both MultiAZ and PlacementGroup are enabled

NSsirena:
Fails if both MultiAZ and PlacementGroup are enabled.

Signed-off-by: Nicola Sirena <nssirena@amazon.com>

### Description of changes
* Added a new MultiAzPlacementGroupValidator that fails if both MultiAZ and PlacementGroup are enabled.
  * A PlacementGroup is enabled if Enabled is true or a Name or Id are specified.

### Tests
* Unit tests
* pcluster create ... --dryrun enabled where
  * both where disabled -> success
  * either MultiAz or PG where enabled (by flag, name or Id) -> success
  * both where enabled -> failure 

### References
-

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -164,6 +164,7 @@ from pcluster.validators.kms_validators import KmsKeyIdEncryptedValidator, KmsKe
 from pcluster.validators.monitoring_validators import ComputeConsoleLoggingValidator
 from pcluster.validators.networking_validators import (
     ElasticIpValidator,
+    MultiAzPlacementGroupValidator,
     SecurityGroupsValidator,
     SingleSubnetValidator,
     SubnetsValidator,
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -2014,6 +2015,12 @@ def _register_validators(self, context: ValidatorContext = None):
                 compute_resource_name=compute_resource.name,
                 compute_resource_efa_enabled=compute_resource.efa.enabled,
             )
+            placement_group = self.get_chosen_placement_group_setting_for_compute_resource(compute_resource)
+            self._register_validator(
+                MultiAzPlacementGroupValidator,
+                multi_az_enabled=self.multi_az_enabled,
+                placement_group_enabled=placement_group.enabled or placement_group.assignment is not None,
```
EddyMM: I think the PlacementGroup Resource should have a property to get the general status of whether a placement group was specified or not. This is probably a common use-case.

```
@property
def enabled_or_assigned(self):
    return self.enabled or self.assignment is not None
```

NSsirena: Good catch.
I agree and apply the changes.

EddyMM (approved):

PR #4501, opened 2022-11-03
[3.x] Add new OnNodeUpdated config parameter - CLI

eantonin:
### Description of changes
* This will add a new on_node_updated parameter to the CLI
* This PR should be merged together with the other  on_node_updated PRs

### Tests
* unit testing
* created a cluster with the parameter and verified the dna.json file was containing the passed value

### References
* Cookbook PR https://github.com/aws/aws-parallelcluster-cookbook/pull/1600

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [ ] Check if documentation is impacted by this change. -> Documentation will be treated separately

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4563, opened 2022-11-18
[OnNodeUpdate] Add new integration test to verify custom action behaviour

enrico-usai:
### Description of changes
The test verifies:
* the `postupdate.sh` script is NOT executed at cluster creation time
* the `postupdate.sh` script is executed at update time with the right argument (`UPDATE-ARG1`)
* the custom bootstrap script and the arguments are updatable
* the `updated_postupdate.sh` is executed during second update with the right argument (`UPDATE-ARG2`)

### Tests
* Tested creating a custom ami with cookbook changes and running `test_update_slurm`.

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1600
* https://github.com/aws/aws-parallelcluster/pull/4501

### Checklist

### DIFF
--- a/tests/integration-tests/tests/update/test_update.py
+++ b/tests/integration-tests/tests/update/test_update.py
@@ -60,7 +60,7 @@ def test_update_slurm(region, pcluster_config_reader, s3_bucket_factory, cluster
         bucket.upload_file(str(test_datadir / script), f"scripts/{script}")
 
     # Create cluster with initial configuration
-    init_config_file = pcluster_config_reader(resource_bucket=bucket_name, bucket=bucket_name)
+    init_config_file = pcluster_config_reader(resource_bucket=bucket_name)
     cluster = clusters_factory(init_config_file)
 
     # Check update hook is NOT executed at cluster creation time
@@ -141,7 +141,6 @@ def test_update_slurm(region, pcluster_config_reader, s3_bucket_factory, cluster
     additional_policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonAppStreamServiceAccess"
     updated_config_file = pcluster_config_reader(
         config_file="pcluster.config.update.yaml",
-        bucket=bucket_name,
         resource_bucket=bucket_name,
         additional_policy_arn=additional_policy_arn,
     )
--- a/tests/integration-tests/tests/update/test_update.py
+++ b/tests/integration-tests/tests/update/test_update.py
@@ -10,6 +10,7 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
+import os.path
 import re
 import time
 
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/update/test_update.py
```diff
@@ -48,20 +49,33 @@ def test_update_slurm(region, pcluster_config_reader, s3_bucket_factory, cluster
...
+        "updated_postupdate.sh",
+    ]:
         bucket.upload_file(str(test_datadir / script), f"scripts/{script}")
 
     # Create cluster with initial configuration
     init_config_file = pcluster_config_reader(resource_bucket=bucket_name, bucket=bucket_name)
     cluster = clusters_factory(init_config_file)
 
+    # Check update hook is NOT executed at cluster creation time
+    assert_that(os.path.exists("/tmp/postupdate_out.txt")).is_false()
```
ermanno, line 67: Non-blocking: for assertions like this, rather than adding a comment saying what we are checking, it might be better to create a method that says what the test is doing (i.e. `_check_update_hook_has_not_run`).

ermanno (dismissed):

tests/integration-tests/tests/update/test_update.py
enrico-usai: I agree in general,
in this case I'd say we can keep it as is, to avoid to introduce a one-liner function.

ermanno: Imho nothing wrong with a one-liner function if it helps make the code self-documenting 😉, in any case it's not blocking, for me the PR can be merged as soon as we have merged the CLI patch and run the test :thumbsup:.

tests/integration-tests/tests/update/test_update/test_update_slurm/pcluster.config.update.yaml
```diff
@@ -18,6 +18,11 @@ HeadNode:
         EnableWriteAccess: true
     AdditionalIamPolicies:
       - Policy: {{ additional_policy_arn }}
+  CustomActions:
+    OnNodeUpdated:
+      Script: s3://{{ bucket }}/scripts/updated_postupdate.sh
+      Args:
+        - UPDATE-ARG2
```
eantonin, line 25: suggestion, not-blocking: Why don't we check a situation with no_args and one with multiple args too?

tests/integration-tests/tests/update/test_update/test_update_slurm/updated_postupdate.sh
```diff
@@ -0,0 +1,14 @@
...
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+echo "updated_postupdate-$1" > /tmp/updated_postupdate_out.txt
+echo "updated_postupdate run OK"
```
eantonin, line 15: question, not blocking: do we check this too?

eantonin (commented):
question, non-blocking, should we test removal of the script

tests/integration-tests/tests/update/test_update.py
enrico-usai: We can discuss about one-liner functions: https://mytwoocents.com/2017/03/25/one-line-methods-can-easily-become-an-antipattern/ ;)

tests/integration-tests/tests/update/test_update/test_update_slurm/updated_postupdate.sh
enrico-usai: This is just a print to see a line in the cloud-init logs

tests/integration-tests/tests/update/test_update/test_update_slurm/pcluster.config.update.yaml
enrico-usai: All these combinations and the updatability of all of them are already unit tested.
Here we're just checking the script is really executed, with the right updated parameters, we can avoid testing every combination and just focus the test on the "integration" part (CLI param + cookbook scripts + update).

enrico-usai:
> question, non-blocking, should we test removal of the script

We could, the point is that we need to perform another update and this requires 15-30min. This combination can be covered (and is already covered) by unit tests.

eantonin (approved):
LGTM

PR #4567, opened 2022-11-19
[3.x] fix easyconfig when the AWS Account contains ipv6-only subnets

jsynack:
The configuration generator silently crashes if it encountered an IPv6-only subnet in an AWS VPC within the account.

This would take place during the VPC/subnet probe process and take place if it was in the same AWS account (Before you can select a VPC to use for pcluster).

This takes place in `_extract_subnet_size()` as `CidrBlock` is only present for IPv4-enabled subnets.

This adds a filter to the `describe_subnets()` call to ignore IPv6-only subnets for now (`ipv6-native` is false).

This will *not* impact dual-stack subnets (IPv4 and IPv6). As `CidrBlock` is present in these subnets and they will continue to be displayed / allowed to be used.


hanwen-pcluste (dismissed):
Thank you for creating the PR. We are going to merge it!

hanwen-pcluste (approved):

PR #4573, opened 2022-11-22
Add Unmanaged FSx multiple AZ validator.

NSsirena:
Unmanaged FSx can be associated to AZs that differ from the ones configured in the queues. In such cases we inform the customers that they may incur in increased latency and costs.

Signed-off-by: Nicola Sirena <nssirena@amazon.com>

### Description of changes
* Add few properties in BaseSharedFsx to 
    * retrieve the list of subnets
    * know if the FS is unmanaged
* Add Unmanaged FSx multiple AZ validator that
    * compares the set of subnets of the queue and the subnets of the FS
    * informs of possible increased latency and costs in case of mismatch

### Tests
* Unit tests.

### References
None

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -103,6 +103,7 @@ from pcluster.validators.cluster_validators import (
     SchedulerValidator,
     SharedStorageMountDirValidator,
     SharedStorageNameValidator,
+    UnmanagedFsxMultiAzValidator,
 )
 from pcluster.validators.common import ValidatorContext
 from pcluster.validators.database_validators import DatabaseUriValidator
@@ -372,16 +373,37 @@ class BaseSharedFsx(Resource):
         self._register_validator(SharedStorageNameValidator, name=self.name)
 
     @property
+    def is_unmanaged(self):
+        """Return true if using existing FSx."""
+        return self.file_system_id is not None
+
+    @property
     def file_system_data(self):
         """Return filesystem information if using existing FSx."""
-        if not self.__file_system_data and self.file_system_id:
+        if not self.__file_system_data and self.is_unmanaged:
             self.__file_system_data = AWSApi.instance().fsx.get_file_systems_info([self.file_system_id])[0]
         return self.__file_system_data
 
     @property
+    def file_system_subnets(self):
+        """Return list of subnets associated to existing FSx."""
+        return self.file_system_data.subnet_ids if self.is_unmanaged else []
+
+    @property
+    def file_system_availability_zones(self):
+        """Return list of AZ associated to existing FSx."""
+        availability_zones = []
+        if self.is_unmanaged:
+            mapping = AWSApi.instance().ec2.get_subnets_az_mapping(self.file_system_subnets)
+            for availability_zone in mapping.values():
+                availability_zones.append(availability_zone)
+
+        return availability_zones
+
+    @property
     def existing_dns_name(self):
         """Return DNSName if using existing FSx filesystem."""
-        return self.file_system_data.dns_name if self.file_system_id else ""
+        return self.file_system_data.dns_name if self.is_unmanaged else ""
 
 
 class SharedFsxLustre(BaseSharedFsx):
@@ -645,6 +667,10 @@ class _QueueNetworking(_BaseNetworking):
             for subnet_id, _az in self.subnet_id_az_mapping.items():
                 self._az_subnet_ids_mapping[_az].add(subnet_id)
         return self._az_subnet_ids_mapping
+
+    @property
+    def az_list(self):
+        return list(self.az_subnet_ids_mapping.keys())
 
 
 class SlurmQueueNetworking(_QueueNetworking):
@@ -1357,7 +1383,9 @@ class BaseClusterConfig(Resource):
             )
 
             self._validate_max_storage_count(ebs_count, existing_storage_count, new_storage_count)
-            self._validate_new_storage_multiple_subnets(self.compute_subnet_ids, new_storage_count)
+            self._validate_new_storage_multiple_subnets(
+                self.scheduling.queues, self.compute_subnet_ids, new_storage_count
+            )
 
         self._validate_mount_dirs()
 
@@ -1373,12 +1401,19 @@ class BaseClusterConfig(Resource):
             local_mount_dir_list=list(self.local_mount_dir_instance_types_dict.keys()),
         )
 
-    def _validate_new_storage_multiple_subnets(self, compute_subnet_ids, new_storage_count):
+    def _validate_new_storage_multiple_subnets(self, queues, compute_subnet_ids, new_storage_count):
         self._register_validator(
             ManagedFsxMultiAzValidator,
             compute_subnet_ids=compute_subnet_ids,
             new_storage_count=new_storage_count,
         )
+        for storage in self.shared_storage:
+            if isinstance(storage, (SharedFsxLustre, ExistingFsxOpenZfs, ExistingFsxOntap)) and storage.is_unmanaged:
+                self._register_validator(
+                    UnmanagedFsxMultiAzValidator,
+                    queues=queues,
+                    fsx_az_list=storage.file_system_availability_zones,
+                )
 
     def _validate_max_storage_count(self, ebs_count, existing_storage_count, new_storage_count):
         for storage_type in ["EFS", "FSx", "RAID"]:
@@ -1929,7 +1964,6 @@ class _CommonQueue(BaseQueue):
         self.capacity_reservation_target = capacity_reservation_target
         self.compute_resources = compute_resources
         self.networking = networking
-        self.__az_list = None
 
     @property
     def instance_role(self):
@@ -1950,22 +1984,9 @@ class _CommonQueue(BaseQueue):
             return None
 
     @property
-    def az_list(self):
-        """Return the list of unique AvailabilityZoneID associated to the subnets defined in the Networking section."""
-        # initialize the hidden property just once
-        if self.__az_list is None:
-            az_set = set()
-            for _key, val in self.networking.subnet_id_az_mapping.items():
-                az_set.add(val)
-
-            self.__az_list = list(az_set)
-
-        return self.__az_list
-
-    @property
     def multi_az_enabled(self):
         """Return true if more than one AZ are defined in the queue Networking section."""
-        return len(self.az_list) > 1
+        return len(self.networking.az_list) > 1
 
     def get_managed_placement_group_keys(self) -> List[str]:
         managed_placement_group_keys = []
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -701,6 +701,28 @@ class ManagedFsxMultiAzValidator(Validator):
             )
 
 
+class UnmanagedFsxMultiAzValidator(Validator):
+    """
+    Unmanaged FSx Storage Vs Multiple Subnets validator.
+
+    Unmanaged FSx volumes can exist in AZ that are different from the ones defined in queues configuration.
+    In these cases we notify customers that they may incur in increased latency and costs.
+    """
+
+    def _validate(self, queues, fsx_az_list):
+        for queue in queues:
+            queue_az_set = set(queue.networking.az_list)
+            fs_az_set = set(fsx_az_list)
+            # we want to ensure that all the az defined in the queue are supported by the FS
+            if not queue_az_set.issubset(fs_az_set):
+                self._add_failure(
+                    "Your configuration for Queue '{0}' includes multiple subnets and external shared storage "
+                    "configuration. Accessing a shared storage from different AZs can lead to increased "
+                    "latency and costs.".format(queue.name),
+                    FailureLevel.INFO,
+                )
+
+
 class EfsIdValidator(Validator):  # TODO add tests
     """
     EFS id validator.
--- a/cli/tests/pcluster/validators/test_cluster_validators.py
+++ b/cli/tests/pcluster/validators/test_cluster_validators.py
@@ -17,6 +17,7 @@ from pcluster.config.cluster_config import (
     BaseQueue,
     CapacityReservationTarget,
     RootVolume,
+    SchedulerPluginQueueNetworking,
     SlurmComputeResource,
     SlurmQueue,
     SlurmQueueNetworking,
@@ -58,6 +59,7 @@ from pcluster.validators.cluster_validators import (
     SchedulerOsValidator,
     SharedStorageMountDirValidator,
     SharedStorageNameValidator,
+    UnmanagedFsxMultiAzValidator,
     _LaunchTemplateValidator,
 )
 from pcluster.validators.common import FailureLevel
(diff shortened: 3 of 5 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,17 +372,27 @@ def __init__(self, mount_dir: str, name: str):
...
-        if not self.__file_system_data and self.file_system_id:
+        if not self.__file_system_data and self.unmanaged:
             self.__file_system_data = AWSApi.instance().fsx.get_file_systems_info([self.file_system_id])[0]
         return self.__file_system_data
 
+    @property
+    def file_system_subnets(self):
+        """Return list of subnets associated to FSx."""
+        return self.file_system_data.subnet_ids if self.unmanaged else []
+
```
jdeamicis: This implementation/name might cause confusion in the future: here you are returning the subnet IDs only if the FSx is unmanaged. Is it wise to create this difference?

jdeamicis: I see you are probably using a pre-existing pattern here. Feel free to dismiss this comment.

jdeamicis: Also, is a new property absolutely necessary, since this information already exists in `file_system_data`?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -701,6 +701,25 @@ def _validate(self, queues, new_storage_count):
...
+    """
+    Unmanaged FSx Storage Vs Multiple Subnets validator.
+
+    Unmanaged FSx volumes can exist in AZ that are different from the ones defined in queues configuration.
+    In these cases we notify customers that they may incur in increased latency and costs.
+    """
+
+    def _validate(self, queues, fsx_subnets):
+        for queue in queues:
+            if not set(queue.networking.subnet_ids) == set(fsx_subnets):
```
jdeamicis: Are we sure ONTAP is multi-zonal in the sense that it supports different mount targets on different AZs? If not, would this be the right logic for a multizonal ONTAP?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1724,6 +1725,176 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                        subnet_ids=["subnet-1", "subnet-2"],
+                    ),
+                ),
+            ],
+            ["subnet-1", "subnet-2"],
+            None,
+            "",
+        ),
+    ],
+)
```
jdeamicis, line 1976: Can you please make one of these cases use the `SchedulerPluginQueueNetworking` class?

jdeamicis (commented):
Left some comments to discuss about.

cli/src/pcluster/config/cluster_config.py
NSsirena: I'd prefer not to ..however `file_system_data` is retrieved only if the FS is unmanaged. This is why I put the check here. 
Moreover I created a property to make it explicit since the previous check wasn't so clear.

I thought about generalizing this also for the managed case, but this was out-of-scope for this PR and also I didn't want to break existing behaviors.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,17 +372,27 @@ def __init__(self, mount_dir: str, name: str):
     def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
         self._register_validator(SharedStorageNameValidator, name=self.name)
 
+    @property
+    def unmanaged(self):
```
lukeseawalker: TIP - tipically we use "is_" prefix to describe that a boolean value is returned.

NSsirena: Got it.
I'll update the property name.

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: Sure, thanks for catching this.

cli/src/pcluster/validators/cluster_validators.py
jdeamicis: Additional comment (more important): this poses the "constraint" (although it's just an INFO) to use the same subnets between the queue and the FSx. We should check the AZs, not the subnets.

cli/src/pcluster/config/cluster_config.py
NSsirena: Discussed this with Jacopo and agreed to proceed with this approach since it's extensively used.

cli/src/pcluster/validators/cluster_validators.py
NSsirena: I totally agree.
I fixed the check and now uses AZ list instead of subnet list.

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: Fixed

cli/src/pcluster/config/cluster_config.py
NSsirena: Fixed

cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,17 +372,38 @@ def __init__(self, mount_dir: str, name: str):
...
+        """Return list of subnets associated to existing FSx."""
+        return self.file_system_data.subnet_ids if self.is_unmanaged else []
+
+    @property
+    def file_system_azs(self):
+        """Return list of AZ associated to existing FSx."""
+        azs = []
+        if self.is_unmanaged:
+            mapping = AWSApi.instance().ec2.get_subnets_az_mapping(self.file_system_subnets)
+            for _subnet_id, availability_zone in mapping.items():
```
lukeseawalker: if `_subnet_id` is not used, you can use
```
for _, availability_zone in mapping.items():
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,17 +372,38 @@ def __init__(self, mount_dir: str, name: str):
...
         return self.__file_system_data
 
+    @property
+    def file_system_subnets(self):
+        """Return list of subnets associated to existing FSx."""
+        return self.file_system_data.subnet_ids if self.is_unmanaged else []
+
+    @property
+    def file_system_azs(self):
+        """Return list of AZ associated to existing FSx."""
```
lukeseawalker: since there is an az_list method, what about calling this `file_system_az_list` for coherence?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1379,6 +1405,13 @@ def _validate_new_storage_multiple_subnets(self, compute_subnet_ids, new_storage
             compute_subnet_ids=compute_subnet_ids,
             new_storage_count=new_storage_count,
         )
+        for storage in self.shared_storage:
+            if isinstance(storage, (SharedFsxLustre, ExistingFsxOpenZfs, ExistingFsxOntap)) and storage.is_unmanaged:
+                self._register_validator(
+                    UnmanagedFsxMultiAzValidator,
+                    queues=queues,
+                    fsx_subnets=storage.file_system_subnets,
```
lukeseawalker: isn't `UnmanagedFsxMultiAzValidator` expecting a list of AZ instead of a list of subnets?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -701,6 +701,25 @@ def _validate(self, compute_subnet_ids, new_storage_count):
...
+    """
+    Unmanaged FSx Storage Vs Multiple Subnets validator.
+
+    Unmanaged FSx volumes can exist in AZ that are different from the ones defined in queues configuration.
+    In these cases we notify customers that they may incur in increased latency and costs.
+    """
+
+    def _validate(self, queues, fsx_az_list):
+        for queue in queues:
+            if not set(queue.networking.az_list) == set(fsx_az_list):
```
lukeseawalker: In principle, they way az_list list are built, there should not be duplicated elements, right?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1790,6 +1792,153 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                ),
+            ],
+            {"subnet-1": "us-east-1a", "subnet-2": "us-east-1b"},
+            ["us-east-1b"],
+            FailureLevel.INFO,
+            "Your configuration for Queue multi-az-queue-partial-match includes multiple subnets and external "
+            "shared storage configuration. Accessing a shared storage from different AZs can lead to increased "
+            "latency and costs.",
+        ),
+        (
```
lukeseawalker, line 1912: duplicated test

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1790,6 +1792,153 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                ),
+            ],
+            {"subnet-1": "us-east-1a", "subnet-2": "us-east-1b"},
+            ["us-east-1a", "us-east-1b"],
+            None,
+            "",
+        ),
+    ],
+)
+def test_unmanaged_fsx_multiple_az_validator(
```
lukeseawalker, line 1977: can you add a test with len(SlurmQueue) > 1?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1379,6 +1405,13 @@ def _validate_new_storage_multiple_subnets(self, compute_subnet_ids, new_storage
             compute_subnet_ids=compute_subnet_ids,
             new_storage_count=new_storage_count,
         )
+        for storage in self.shared_storage:
+            if isinstance(storage, (SharedFsxLustre, ExistingFsxOpenZfs, ExistingFsxOntap)) and storage.is_unmanaged:
+                self._register_validator(
+                    UnmanagedFsxMultiAzValidator,
+                    queues=queues,
```
lukeseawalker, line 1414: `queues` is not defined

cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,17 +372,38 @@ def __init__(self, mount_dir: str, name: str):
...
+        """Return list of subnets associated to existing FSx."""
+        return self.file_system_data.subnet_ids if self.is_unmanaged else []
+
+    @property
+    def file_system_availability_zones(self):
+        """Return list of AZ associated to existing FSx."""
+        availability_zones = []
+        if self.is_unmanaged:
+            mapping = AWSApi.instance().ec2.get_subnets_az_mapping(self.file_system_subnets)
+            for _, availability_zone in mapping.items():
```
lukeseawalker:
```
for availability_zone in mapping.values():
```

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1373,12 +1401,19 @@ def _validate_mount_dirs(self):
...
             ManagedFsxMultiAzValidator,
             compute_subnet_ids=compute_subnet_ids,
             new_storage_count=new_storage_count,
         )
+        for storage in self.shared_storage:
+            if isinstance(storage, (SharedFsxLustre, ExistingFsxOpenZfs, ExistingFsxOntap)) and storage.is_unmanaged:
+                self._register_validator(
+                    UnmanagedFsxMultiAzValidator,
+                    queues=queues,
+                    fsx_subnets=storage.file_system_availability_zones,
```
lukeseawalker: wrong param name, subnet != az

cli/src/pcluster/config/cluster_config.py
NSsirena: 👍 ..I'll fix it

cli/src/pcluster/config/cluster_config.py
NSsirena: oops! ...good catch!
wonder why it was working...

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -701,6 +701,25 @@ def _validate(self, compute_subnet_ids, new_storage_count):
...
+                self._add_failure(
+                    "Your configuration for Queue {0} includes multiple subnets and external shared storage "
```
lukeseawalker: quote the values, i.e. `'{0}'`

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1790,6 +1792,167 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                    compute_resources=[],
+                    networking=SlurmQueueNetworking(
+                        subnet_ids=["subnet-1"],
+                    ),
+                ),
+            ],
+            [{"subnet-2": "us-east-1b"}, {"subnet-1": "us-east-1a"}],
+            ["us-east-1a"],
+            FailureLevel.INFO,
+            "Your configuration for Queue single-az-different-subnet-queue includes multiple subnets and external "
```
lukeseawalker: add quoting to the value, i.e. `'single-az-different-subnet-queue'`

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1790,6 +1792,167 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                    networking=SchedulerPluginQueueNetworking(
+                        subnet_ids=["subnet-1", "subnet-2"],
+                    ),
+                ),
+            ],
+            [{"subnet-1": "us-east-1a", "subnet-2": "us-east-1b"}],
+            ["us-east-1a", "us-east-1b"],
+            None,
+            "",
+        ),
```
lukeseawalker, line 1974: missing test case with more than 1 queue where none of the queue is in the same AZ of the FS. This to trigger the path of multiple failure messages

cli/src/pcluster/validators/cluster_validators.py
NSsirena: I had quotes. I removed them since in the messages reported in the doc the values were unquoted.

Anyway I'll bring back the quotes.

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: Same as above.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1790,6 +1792,167 @@ def test_new_storage_multiple_subnets_validator(queues, new_storage_count, failu
...
+                    compute_resources=[],
+                    networking=SlurmQueueNetworking(
+                        subnet_ids=["subnet-1", "subnet-2"],
+                    ),
+                ),
+            ],
+            [{"subnet-1": "us-east-1a"}, {"subnet-1": "us-east-1a", "subnet-2": "us-east-1b"}],
+            ["us-east-1a", "us-east-1b"],
+            FailureLevel.INFO,
+            "Your configuration for Queue 'one-az-match-queue' includes multiple subnets and external "
```
lukeseawalker: there should be no error in this case.
FS is in AZ ["us-east-1a", "us-east-1b"], so cluster will be able to mount the FS in the same AZ

NSsirena: This is true IF and OLNY IF the data in each AZ is an exact replica and so a compute node in `us-east-1a` will connect to the FS mount point in `us-east-1a` and all the data will be local.

If, somehow, this is not true data will be transferred from one AZ to another and this will have an impact on the latency while accessing the data and possibly cost.

NSsirena: We discussed this topic in a call a decided to slightly change the logic of the validator.

lukeseawalker (approved):

PR #4571, opened 2022-11-21
3.x: Allow to identify a deprecated AMI as the official one

awslitvit:
### Description of changes
As per [EC2 documentation](https://aws.amazon.com/about-aws/whats-new/2022/03/amazon-machine-images-public-visibility-two-years/), all public AMIs are now deprecated after 2 years at most after their creation. In ParallelCluster context, it means that after 2 years from a ParallelCluster release its official AMIs are going to be deprecated.

This change aims at making them still discoverable after the deprecation, even if not supported, in order to restore the behavior prior to the change by EC2.

Specifically:
* If there are active (not deprecated) suitable AMIs, the active AMI with the latest creation date will be picked as the official AMI
* If all the suitable AMIs are deprecated, the one with the latest creation date will be picked as the official AMI even if deprecated

#### Unit tests
A specific Unit Test has been added to verify the correct AMI is picked as the official one, according to the rules above.

#### Manual tests
In order to simulate AMI deprecation, private images with names matching ParallelCluster official image names were created in `ACCOUNT_A` and shared with `ACCOUNT_B`.

`DevSettings > AmiSearchFilters > Owner` in ParallelCluster config file was set to ACCOUNT_A and test performed on ACCOUNT_B.

The behavior prior to the patch verified:
* AMI1 created and shared
* create-cluster: AMI1 was picked as the official image
* AMI1 deprecated 
* create-cluster: received the response message "Cannot find official ParallelCluster AMI"

After applying the patch:
* create-cluster: deprecated AMI1 was picked as the official image
* AMI2 created and shared
* create-cluster: active AMI2 was picked as the official image because it was the only active image
* AMI2 deprecated 
* create-cluster: deprecated AMI2 was picked as the official AMI because of a later CreationDate
* Deprecation removed from AMI1
* create-cluster: AMI1 was picked as the official AMI despite the earlier CreationDate, as the only active image
* Deprecation removed from AMI2
* create-cluster: AMI2 was picked as the official AMI because of a later CreationDate

### References
* [Amazon EC2 now reduces visibility of public Amazon Machine Images (AMIs) older than two years](https://aws.amazon.com/about-aws/whats-new/2022/03/amazon-machine-images-public-visibility-two-years/)

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
[...]

PR #4584, opened 2022-11-22
3.x: Include deprecated AMIs into list-official-images

awslitvit:
### Description of changes
As per [EC2 documentation](https://aws.amazon.com/about-aws/whats-new/2022/03/amazon-machine-images-public-visibility-two-years/), all public AMIs are now deprecated after 2 years at most after their creation. In ParallelCluster context, it means that after 2 years from a ParallelCluster release its official AMIs are going to be deprecated.

This change aims at making them still listed by `list-official-images` after the deprecation, even if not supported, in order to restore the behavior prior to the change by EC2.

Only one AMI per OS and architecture will be returned at most.
- If there are active (not deprecated) suitable AMIs for OS/architecture, the active AMI with the latest creation date will be returned
- If all the suitable AMIs for OS/architecture are deprecated, the one with the latest creation date will be returned even if deprecated

### Unit tests
Unit Tests have been added to verify the correct AMI for OS and architecture is returned, according to the rules above.

### Manual tests
In order to simulate AMI deprecation, two private images with names matching ParallelCluster official image names were created in ACCOUNT_A and shared with ACCOUNT_B.

DevSettings > AmiSearchFilters > Owner in ParallelCluster config file was set to ACCOUNT_A and test performed on ACCOUNT_B.

Prior to the change, deprecated images weren't listed in response to `list-official-images`.
After the change, only one of them was listed, since they are both suitable for the same OS/architecture. 
The listed image was correct according to the rules above.

### References
* [Amazon EC2 now reduces visibility of public Amazon Machine Images (AMIs) older than two years](https://aws.amazon.com/about-aws/whats-new/2022/03/amazon-machine-images-public-visibility-two-years/)
* Previous PR: [3.x: Allow to identify a deprecated AMI as the official one](https://github.com/aws/aws-parallelcluster/pull/4571)

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -8,6 +8,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+import itertools
 import re
 from datetime import datetime
 from typing import Any, List, Tuple
@@ -288,6 +289,9 @@ class Ec2Client(Boto3Client):
     def _is_image_deprecated(image):
         return "DeprecationTime" in image and utils.to_iso_timestr(datetime.now()) >= image["DeprecationTime"]
 
+    def _find_valid_official_image(self, images):
+        return max(images, key=lambda image: ("0" if self._is_image_deprecated(image) else "1") + image["CreationDate"])
+
     @AWSExceptionHandler.handle_client_exception
     @Cache.cached
     def get_official_image_id(self, os, architecture, filters=None):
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/aws/ec2.py
```diff
@@ -300,21 +304,22 @@ def get_official_image_id(self, os, architecture, filters=None):
...
+    @Cache.cached
     def get_official_images(self, os=None, architecture=None):
         """Get the list of official images, optionally filtered by os and architecture."""
-        try:
-            owners = ["amazon"]
-            name = f"{self._get_official_image_name_prefix(os, architecture)}*"
-            filters = [{"Name": "name", "Values": [name]}]
-            return self.describe_images(ami_ids=[], owners=owners, filters=filters)
-        except ImageNotFoundError:
-            return []
```
hanwen-pcluste, line 317: Does this deletion change the behavior of ParallelCluster CLI when the image cannot be found?

awslitvit: No.
Both images created in `ACCOUNT_A` (see test description) are for `alinux2`, there are no images for `centos7`.
This is the test I performed:

```
.../python .../entrypoint.py list-official-images -r eu-west-1 --os centos7
{
  "images": []
}
```

`ImageNotFoundError` was thrown by `EC2Client.describe_images` which is not called anymore, because it wraps EC2 response into `ImageInfo` class, hiding information about `DeprecationTime`. EC2 client is now called directly, and it returns an empty list if there are no images. This is why I removed the exception handling.

hanwen-pcluste (approved):

PR #4606, opened 2022-11-24
3.x: Use the original official AMI when updating the cluster

awslitvit:
### Description of changes
As of today, ParallelCluster does not support head node replacement. 
For this reason, if the official AMI is upgraded after the cluster creation, a later cluster update would break.
This change avoids the issue, reusing the original AMI instead of picking the latest one.

### Tests
Manual tests:
  * New cluster created with the official AMI, CFN parameter `OfficialAmi` is set correctly
  * During cluster update, the official AMI is not looked up but is taken from the CFN parameter

### References
* https://github.com/aws/aws-parallelcluster/pull/4571
* https://github.com/aws/aws-parallelcluster/pull/4584

### Checklist

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -172,7 +172,7 @@ class Cluster:
         self.template_body = None
         self.__config = None
         self.__s3_artifact_dir = None
-
+        self.__official_ami = None
         self.__has_running_capacity = None
         self.__running_capacity = None
 
@@ -181,6 +181,7 @@ class Cluster:
         """Return the ClusterStack object."""
         if not self.__stack:
             self.__stack = ClusterStack(AWSApi.instance().cfn.describe_stack(self.stack_name))
+            self.__official_ami = self.__stack.official_ami
         return self.__stack
 
     @property
@@ -434,6 +435,8 @@ class Cluster:
             LOGGER.info("Validating cluster configuration...")
             Cluster._load_additional_instance_type_data(cluster_config_dict)
             config = self._load_config(cluster_config_dict)
+            config.official_ami = self.__official_ami
+
             validation_failures = config.validate(validator_suppressors, context)
             if any(f.level.value >= FailureLevel(validation_failure_level).value for f in validation_failures):
                 raise ConfigValidationError("Invalid cluster configuration.", validation_failures=validation_failures)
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/models/cluster.py
```diff
@@ -434,6 +434,7 @@ def _validate_and_parse_config(
             LOGGER.info("Validating cluster configuration...")
             Cluster._load_additional_instance_type_data(cluster_config_dict)
             config = self._load_config(cluster_config_dict)
+            config.official_ami = self.__stack.official_ami if self.__stack else None
```
demartinofra: don't you need to access `stack` property to have the ClusterStack object properly initialized? Also what happens if the call to CFN fails? do we handle the failure properly?

awslitvit: > don't you need to access stack property to have the ClusterStack object properly initialized?

This code runs both on `cluster-create` and on `cluster-update`. The reason of this ugly if is that `self.__stack` doesn't exist during cluster creation. I didn't find any better way to check and I didn't want to change the code more than required for this specific change. I would appreciate any suggestion.

> Also what happens if the call to CFN fails? do we handle the failure properly?

At this point, if we are on cluster creation path `self.__stack` is not set and if we are on cluster update path we rely on the fact it is already set. I agree it is not robust. However, right now I don't see how a CFN failure can affect this specific line of code: it would break before and the error handling doesn't change.

demartinofra: >This code runs both on cluster-create and on cluster-update. The reason of this ugly if is that self.__stack doesn't exist during cluster creation. I didn't find any better way to check and I didn't want to change the code more than required for this specific change. I would appreciate any suggestion.

The problem I see here is that if no-one calls the self.stack property this code will not be executed `self.__stack = ClusterStack(AWSApi.instance().cfn.describe_stack(self.stack_name))`. Which means that accessing the AMI through `self.__stack.official_ami` relies on the assumption that at some point self.stack property has been already invoked and the ClusterStack object initialized. (if I'm not missing something :))

>However, right now I don't see how a CFN failure can affect this specific line of code: it would break before and the error handling doesn't change.

This is true if we assume that the stack object has been already initialized (which in a way is a weak assumption as you are saying). I'm ok if we feel like we want to leave it as is.

awslitvit: Code refactored.

enrico-usai (approved):

PR #4624, opened 2022-11-29
Adds MultiAzEBSVolumeValidator

NSsirena:
### Description of changes
* Adds MultiAzEBSVolumeValidator that 
  * fails the cluster creation if the EBS Volume is created in an AZ different from the HeadNode
  * informs customers of risks of increased latency and costs if a queue has resources in an AZ different from the EBS volume

* Adds some property to SharedEBS

### Tests
* Unit test

### References
None

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -474,6 +474,7 @@ class Ec2Client(Boto3Client):
         )
 
     @AWSExceptionHandler.handle_client_exception
+    @Cache.cached
     def describe_volume(self, volume_id):
         """Describe a volume."""
         return self._client.describe_volumes(VolumeIds=[volume_id]).get("Volumes")[0]
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -120,6 +120,7 @@ from pcluster.validators.ebs_validators import (
     EbsVolumeThroughputIopsValidator,
     EbsVolumeThroughputValidator,
     EbsVolumeTypeSizeValidator,
+    MultiAzEbsVolumeValidator,
     SharedEbsVolumeIdValidator,
 )
 from pcluster.validators.ec2_validators import (
@@ -310,6 +311,19 @@ class SharedEbs(Ebs):
         )
         self._register_validator(EbsVolumeSizeSnapshotValidator, snapshot_id=self.snapshot_id, volume_size=self.size)
         self._register_validator(DeletionPolicyValidator, deletion_policy=self.deletion_policy, name=self.name)
+
+    @property
+    def is_managed(self):
+        """Return True if the volume is managed."""
+        return self.volume_id is None
+
+    @property
+    def availability_zone(self):
+        """Return the availability zone of an existing EBS volume."""
+        if not self.is_managed:
+            return AWSApi.instance().ec2.describe_volume(self.volume_id)["AvailabilityZone"]
+        else:
+            return ""
 
 
 class SharedEfs(Resource):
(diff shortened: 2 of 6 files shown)


cli/src/pcluster/aws/ec2.py
```diff
@@ -473,10 +473,15 @@ def describe_network_interfaces(self, network_interface_ids):
             self._paginate_results(self._client.describe_network_interfaces, NetworkInterfaceIds=network_interface_ids)
         )
 
+    @AWSExceptionHandler.handle_client_exception
+    def describe_volumes(self, volume_ids):
+        """Describe volumes."""
+        return list(self._paginate_results(self._client.describe_volumes, VolumeIds=[volume_ids]))
```
lukeseawalker: if `volume_ids` is a list, the value passed to `VolumeIds` will be a list of list, which I think it's wrong.
A unit test would have caught this.

NSsirena: Good catch!
I did this in a hurry to fix an issue with the mock_api and missed this.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1986,6 +1990,185 @@ def test_unmanaged_fsx_multiple_az_validator(
...
+                        subnet_ids=["subnet-2"],
+                    ),
+                ),
+            ],
+            [{"subnet-1": "us-east-1a"}, {"subnet-2": "us-east-1b"}],
+            FailureLevel.ERROR,
+            [
+                "Your configuration includes an EBS volume created in a different availability zone than the "
+                + "Head Node. The volume and instance must be in the same Availability Zone.",
+                "Your configuration for Queue 'different-az-queue-2' includes multiple subnets and external "
```
lukeseawalker, line 2119: not sure if the info message is useful in this case, since it comes from an invalid configuration. What about returning only the error message?

cli/src/pcluster/aws/ec2.py
```diff
@@ -473,10 +473,15 @@ def describe_network_interfaces(self, network_interface_ids):
             self._paginate_results(self._client.describe_network_interfaces, NetworkInterfaceIds=network_interface_ids)
         )
 
+    @AWSExceptionHandler.handle_client_exception
+    def describe_volumes(self, volume_ids):
```
lukeseawalker: what about caching the outputs?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -311,6 +312,19 @@ def _register_validators(self, context: ValidatorContext = None):
...
+    @property
+    def is_managed(self):
+        """Return True if the volume is managed."""
+        return self.volume_id is None
+
+    @property
+    def availability_zone(self):
+        """Return the availability zone of an existing EBS volume."""
+        if not self.is_managed:
+            return AWSApi.instance().ec2.describe_volume(self.volume_id)["AvailabilityZone"]
```
lukeseawalker, line 324: nit - if output of the describe is cached, this can be simplified into
```
return AWSApi.instance().ec2.describe_volume(self.volume_id).get("AvailabilityZone","")
```

cli/src/pcluster/aws/ec2.py
NSsirena: I thought about it ..but since I added this method there is no code calling it directly and it seemed a premature optimization.

I would leave it for when we have a real need for this.

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: It is a valid information and we should raise the point to the customers as soon as we are aware of it.
They may ignore it, but if they want to act on it they will not be forced to make two attempts to discover it.

cli/src/pcluster/aws/ec2.py
NSsirena: fixed

cli/src/pcluster/config/cluster_config.py
NSsirena: As said in the previous comment. Adding caching now seems like a premature optimization since there are basically no other callers for this method than the one I just created.

lukeseawalker (dismissed):

lukeseawalker (dismissed):

lukeseawalker (dismissed):

lukeseawalker (approved):

PR #4636, opened 2022-11-30
Add policies to cluster nodes to allow mounting EFS with EFS utils

hanwen-pcluste:
1. If IamAuthorization is used when mounting EFS, the clients of the mount (EC2 instances) need to have additional policies:
```
{
            "Action": [
                "elasticfilesystem:ClientWrite",
                "elasticfilesystem:ClientMount",
                "elasticfilesystem:DescribeMountTargets",
            ],
            "Resource": <EFS file system ARN>,
            "Effect": "Allow",
            "Sid": "Efs"
        },
```
Therefore, we add the policies whenever an EFS file system uses IamAuthorization. This EFS file system can be either managed or unmanaged. I manually tested both cases. We don't add the policies by default to keep cluster policies minimal
2. Because these policies apply to both head and compute nodes, they are added in the parent class NodeIamResourcesBase, instead of the subclass HeadNodeIamResources or ComputeNodeIamResources
3. Because now the policy resource depends on shared EFS resource, the logic of creating shared storage is moved ahead of Iam resource
4. This commit also fixes a bug. Before the fix, the information of EFS were passed to cookbook dna.json as "True,False...". After the change, we pass the value to cookbook using lower case "true, false..."

## Test
Manually created a cluster with the following cluster configuration file:
```
Region: us-east-1
Image:
  Os: alinux2
  CustomAmi: ami-xxx
HeadNode:
  InstanceType: m6g.xlarge
  Networking:
    SubnetId: subnet-xxx
  Ssh:
    KeyName: jan2022
Scheduling:
  Scheduler: slurm
  SlurmQueues:
  - Name: queue1
    ComputeResources:
    - Name: m6gxlarge
      InstanceType: m6g.xlarge
      MinCount: 1
      MaxCount: 13
    Networking:
      SubnetIds:
      - subnet-xxx
SharedStorage:
  - MountDir: efs1
    Name: efs1
    StorageType: Efs
    EfsSettings:
      FileSystemId: fs-xxx # With file system policy
      EncryptionInTransit: true
      IamAuthorization: true
  - MountDir: efs2
    Name: efs2
    StorageType: Efs
    EfsSettings:
      FileSystemId: fs-xxx # Without file system policy
      EncryptionInTransit: true
      IamAuthorization: true
  - MountDir: efs3
    Name: efs3
    StorageType: Efs
    EfsSettings:
      EncryptionInTransit: true
      IamAuthorization: true

```

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -118,8 +118,12 @@ def get_directory_service_dna_json_for_head_node(config: BaseClusterConfig) -> d
     )
 
 
-def to_comma_separated_string(list):
-    return ",".join(str(item) for item in list)
+def to_comma_separated_string(list, use_lower_case=False):
+    result = ",".join(str(item) for item in list)
+    if use_lower_case:
+        return result.lower()
+    else:
+        return result
 
 
 def get_shared_storage_ids_by_type(shared_storage_infos: dict, storage_type: SharedStorageType):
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -380,18 +392,46 @@ def _add_node_role(self, node: Union[HeadNode, BaseQueue], name: str):
...
+                        sid="Efs",
+                        actions=[
+                            "elasticfilesystem:ClientMount",
+                            "elasticfilesystem:ClientRootAccess",
+                            "elasticfilesystem:ClientWrite",
+                        ],
+                        effect=iam.Effect.ALLOW,
+                        resources=efs_with_iam_authorization_arns,
+                    ),
+                )
```
francesco-giordano, line 414: What about add this logic in the method _build_policy?

hanwen-pcluste: Adding this logic in _build_policy requires duplication for head and compute nodes.
Head node https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/cdk_builder_utils.py#L488
Compute nodes https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/cdk_builder_utils.py#L756

So the current place should be the right place to put common policies

francesco-giordano (approved):

PR #4652, opened 2022-12-05
[3.x] Add Integration tests for Resource Prefix

himani2411:
### Description of changes
Adding Integration Tests Cases for addition of Resource Prefix, where I am creating a different user role with a permission boundary which restricts a user from creating IAM resources without a specific path or name prefix.

-   Add Integration test cases and config file for Iam Resource Prefix in test_iam.py 
-   Add user-role-rp.cfn.yaml to provide a User Role for creating test clusters 
-   Add the test case in pcluster3.yaml config for Jenkins tests.

### Tests
* Manually tested the scenarios covered in integration Tests.

### Checklist


tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+    test_datadir,
+    scheduler_commands_factory,
+    s3_bucket_factory,
+    s3_bucket,
+    iam_resource_prefix,
+):
+    bucket_name = s3_bucket
+    cfn_client, _, iam_client, _ = _create_boto3_clients(region)
+
+    create_config, update_config = _get_config_create_and_update(test_datadir)
```
gmarciani: The `update_config` variable is never used. 
Is it because the PR is actually a draft and we expect to use it later on?
If this is the case, it would be better to mark the PR as a draft

tests/integration-tests/tests/iam/test_iam/test_iam_resource_prefix/pcluster.config.yaml
```diff
@@ -0,0 +1,49 @@
...
+Iam:
+  ResourcePrefix: {{ iam_resource_prefix}}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Iam:
+    S3Access:
```
gmarciani, line 12: Here and below: why does the cluster need access to a S3 bucket?

tests/integration-tests/tests/iam/test_iam/test_iam_resource_prefix/pcluster.config.yaml
```diff
@@ -0,0 +1,49 @@
...
+        S3Access:
+          - BucketName: {{ bucket }}
+            KeyName: read_only/
+            EnableWriteAccess: false
+          - BucketName: {{ bucket }}
+            KeyName: read_and_write/
+            EnableWriteAccess: true
+DevSettings:
+  Timeouts:
+    HeadNodeBootstrapTimeout: 1234  # timeout in seconds
```
gmarciani: Why do we need to set the headnode bootstrap timeout for these tests?

tests/iam_policies/user-role-rp.cfn.yaml
```diff
@@ -0,0 +1,796 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
```
gmarciani: I suppose `user-role-rp.cfn.yaml` is the copy of [user-role.cfn.yaml](https://github.com/aws/aws-parallelcluster/blob/develop/tests/iam_policies/user-role.cfn.yaml) with the only difference that in `user-role-rp.cfn.yaml` some permissions are restricted to resources (roles and policies) with the expected resource prefix used by the integration tests.

If this is the case I suggest to avoid such duplication because every change to the pcluster user role should then be reflected in two files that must be kept aligned.

What about reading the original yaml file within the test and apply the resource prefix changes at runtime?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+    region,
+    pcluster_config_reader,
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+    s3_bucket_factory,
+    s3_bucket,
+    iam_resource_prefix,
+):
+    bucket_name = s3_bucket
```
gmarciani: Why do you need a bucket in this test?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+        if "arn:aws:iam:" in resource:
+            if iam_path:
+                assert_that(resource).contains(iam_path)
+            else:
+                assert_that(resource).contains("/parallelcluster/")
+        if iam_name_prefix:
+            assert_that(resource).contains(iam_name_prefix)
+
+
+def _test_iam_resource_in_cluster(cfn_client, iam_client, stack_name, iam_resource_prefix):
```
gmarciani, line 398: Here you're verifying that the IAM resources are created with the expected ResourcePrefix.

I think that these are verifications that could and should be executed at the unit testing level, not at the integration testing level.

What do you think?

tests/integration-tests/tests/iam/test_iam/test_iam_resource_prefix/pcluster.config.yaml
himani2411: I need this as S3 bucket has a few Policies. As you suggested I will test them for Updates to check the updated policies.

tests/integration-tests/tests/iam/test_iam/test_iam_resource_prefix/pcluster.config.yaml
himani2411: Will remove them.

tests/integration-tests/tests/iam/test_iam.py
himani2411: I need this as S3 bucket has a Policy. As you suggested I will test them for Updates to check the updated policies.

tests/iam_policies/user-role-rp.cfn.yaml
himani2411: I had to make changes by adding a lot of different permissions I the policies as well as the permission boundary as well as removed the Policies related to Image Builder. If I keep these Image Builder policies my File goes over 50Kb and I wont be able to run this template and would have to upload the template through S3 bucket for my Stack to be created.

tests/integration-tests/tests/iam/test_iam.py
himani2411: Will check for creation of the cluster.

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+    cluster_config = pcluster_config_reader(
+        config_file=create_config,
+        min_count=1,
+        bucket=bucket_name,
+        iam_resource_prefix=iam_resource_prefix,
+    )
+    cluster = clusters_factory(cluster_config)
+    _test_iam_resource_in_cluster(cfn_client, iam_client, cluster.name, iam_resource_prefix)
+
+def _split_resource_prefix(resource_prefix):
```
dreambeyondorange, line 366: I wonder if we should extract this method from the core implementation so it can be reused here?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+            if iam_path:
+                assert_that(resource).contains(iam_path)
+            else:
+                assert_that(resource).contains("/parallelcluster/")
+        if iam_name_prefix:
+            assert_that(resource).contains(iam_name_prefix)
+
+
+def _test_iam_resource_in_cluster(cfn_client, iam_client, stack_name, iam_resource_prefix):
+    """Test IAM resources ( Roles, policy and Instance profiles)."""
```
dreambeyondorange: Could you elaborate on what exactly we are testing for here?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+            resource_arn_list.append(
+                iam_client.get_instance_profile(InstanceProfileName=resource["PhysicalResourceId"])["InstanceProfile"][
+                    "Arn"
+                ]
+            )
+    _check_iam_resource_prefix(resource_arn_list, iam_resource_prefix)
+
+
+@pytest.fixture(scope="class")
+def initialize_resource_prefix_cli_creds(request):
```
dreambeyondorange, line 428: This function can be combined with the existing initialize_cli_creds() by adding a parameter or two with default values.  For example
```
stack_template_path = os.path.join("..", "iam_policies", "user-role-rp.cfn.yaml")
```
can be parameterized with `user_role_config_file`

and 
```
stack_name = generate_stack_name(
                    "integ-tests-iam-rp-user-role", request.config.getoption("stackname_suffix")
                )
```
can be parameterized with `base_stack_name`

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +336,134 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+        yield cli_creds
+
+        if not request.config.getoption("no_delete"):
+            stack_factory.delete_all_stacks()
+        else:
+            logging.warning("Skipping deletion of CFN stacks because --no-delete option is set")
+
+
+@pytest.fixture(scope="class")
+def register_resource_prefix_cli_credentials(initialize_resource_prefix_cli_creds):
```
dreambeyondorange: this function is identical to the existing `register_cli_credentials`.  Is there a reason you can't just use that function?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -10,16 +10,20 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
+import os
```
dreambeyondorange: looks like you are importing os twice?

tests/integration-tests/tests/iam/test_iam.py
dreambeyondorange: Passing parameters from a test to a fixture seems overly complicated but can be done using indirect parametrization https://docs.pytest.org/en/7.1.x/example/parametrize.html#indirect-parametrization

tests/integration-tests/tests/iam/test_iam.py
himani2411: I did that originally but during one of the issues with Jenkins which Hanwen helped for solving he suggested I move this in the tests section as the Integration tests should be independent of the core implementation.

tests/integration-tests/tests/iam/test_iam.py
himani2411: The scope of existing initialize_cli_creds() was session and autouse set to True which is why I didnt want to reuse them as the changes I could make have a bigger blast radius and affect the other test cases CLI creds.

tests/integration-tests/tests/iam/test_iam/test_iam_resource_prefix/pcluster.config.yaml
gmarciani: Got it, thanks

tests/iam_policies/user-role-rp.cfn.yaml
gmarciani: > If I keep these Image Builder policies my File goes over 50Kb and I wont be able to run this template and would have to upload the template through S3 bucket for my Stack to be created.

Still it's possible to parse the original yaml file and remove the unnecessary resources at runtime to avoid duplication.
Nevertheless, let's proceed iteratively as we agreed on.

the first iteration on integ tests should provide the following values:
1. verify that the feature does not introduce regressions, e.g. having a ResourcePrefix with the standard user role does not break the cluster
1. verify that the ResourcePrefix works in a basic case

tests/integration-tests/tests/iam/test_iam.py
gmarciani: Got it thanks

tests/integration-tests/tests/iam/test_iam.py
gmarciani: Great and please remember to keep track in a task of the necessity to add a unit tests to cover the generated resources.

tests/integration-tests/tests/iam/test_iam.py
gmarciani: I agree that the long term solution would be to make the session-scoped initialize_cli_creds more flexible to be able to also cover this scenario.

Nevertheless, let's proceed iteratively. Let's start simple with a fixture dedicated to our case and improve it in later steps.

tests/integration-tests/tests/iam/test_iam.py
himani2411: Will resolve this.

tests/integration-tests/tests/iam/test_iam.py
himani2411: The reason is same as [above](https://github.com/aws/aws-parallelcluster/pull/4652#discussion_r1042501173) mentioned

tests/integration-tests/tests/iam/test_iam.py
himani2411: I am testing if the IAM resources have the expected path and names in their ARN or name.
Apart from that a new change is to add a test for checking the Cluster status, which is what Giacomo suggested [here](https://github.com/aws/aws-parallelcluster/pull/4652#discussion_r1040914842) 

Will add this in the Comments as well.

tests/integration-tests/tests/iam/test_iam.py
dreambeyondorange: good point about the session scope!  I don't really think indirect parametrization will work in that case

tests/integration-tests/clusters_factory.py
```diff
@@ -417,10 +424,12 @@ def create_cluster(self, cluster, log_error=True, raise_on_error=True, **kwargs)
 
         # create the cluster
         logging.info("Creating cluster {0} with config {1}".format(name, cluster.config_file))
+        self.custom_cli_credentials = kwargs.get("custom_cli_credentials")
```
dreambeyondorange: Can you not just reference the credentials from the __init__ function?  Why reset this value here?

dreambeyondorange: Either we should define the credentials for the factory or use the credentials provided to the create method, but not do both

dreambeyondorange: I'm leaning toward the latter, since the factory can then be independent of the cluster's credentials and it seems that the factory only uses the credentials argument (and the run_pcluster_command) in the creation of the cluster

tests/integration-tests/framework/credential_providers.py
```diff
@@ -25,15 +25,20 @@ def register_cli_credentials_for_region(region, iam_role):
...
+    if kwargs.get("custom_cli_credentials"):
+        with sts_credential_provider(region, credential_arn=kwargs.get("custom_cli_credentials")):
             return run_command(*args, **kwargs)
     else:
-        return run_command(*args, **kwargs)
+        if region in cli_credentials:
+            with sts_credential_provider(region, cli_credentials[region]):
+                return run_command(*args, **kwargs)
+        else:
+            return run_command(*args, **kwargs)
```
dreambeyondorange: I think this switching would be better done on the credentials argument itself and avoid multiple return statements.  It should be possible to do this with one assignment
```
credentials_arg = kwargs.get("custom_cli_credentials") or cli_credentials.get(region)
```
the `or` should do the switching for you, i.e. if kwargs.get("custom_cli_credentials") returns None, `credentials_arg` will be the region credentials

I think we should also refactor the cli_credentials[region] to use the get() method to avoid missing key errors.

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -332,3 +334,140 @@ def test_s3_read_write_resource(region, pcluster_config_reader, s3_bucket_factor
...
+
+    yield _create_resource_prefix_cli_creds
+
+    if not request.config.getoption("no_delete"):
+        stack_factory.delete_all_stacks()
+    else:
+        logging.warning("Skipping deletion of CFN stacks because --no-delete option is set")
+
+
+def _inject_resource_in_config(cluster_config, resource_value, resource_keys):
```
dreambeyondorange: this looks like a candidate for a more general utility module.  Is this something that is specific to this test?

tests/integration-tests/utils.py
```diff
@@ -80,7 +80,16 @@ def retry_if_subprocess_error(exception):
...
-def run_command(command, capture_output=True, log_error=True, env=None, timeout=None, raise_on_error=True, shell=False):
+def run_command(
+    command,
+    capture_output=True,
+    log_error=True,
+    env=None,
+    timeout=None,
+    raise_on_error=True,
+    shell=False,
+    custom_cli_credentials=None,
```
dreambeyondorange: It doesn't look like this argument is used?

tests/integration-tests/clusters_factory.py
himani2411: Since the clusters_factory(scope class) fixture in conftest is calling the Clustersfactory and if I setup the custom_cli_credentials value there it will always take the default None value. As in test_iam class any class which comes first will setup the ClusterFactory object. So when I use cluster_factory fixture in my test case I would have to send my custom_credentials where it will reset the object member in my test case.

tests/integration-tests/utils.py
himani2411: Run_command() definition didnt originally had variable arguments, so the kwargs that its being sent from run_pcluster_command now has custom_cli_credentials as one of the arguments. So either I mention the variable I want to send or just add **kwargs in the run_command definition to avoid an error. I preferred adding my own custom_cli_credentials in way of avoiding sending unnecessary variables through kwargs.

tests/integration-tests/tests/iam/test_iam.py
himani2411: I was planning on using this function for injecting my Resource for Iam_resource_prefix in the user-role.cfn.yaml which is discussed [here](https://github.com/aws/aws-parallelcluster/pull/4652#discussion_r1040909717). though I am planning on doing this in another PR.

tests/integration-tests/framework/credential_providers.py
himani2411: Will make the changes.

tests/integration-tests/utils.py
himani2411: Will change this after part of our discussion

tests/integration-tests/tests/iam/test_iam.py
himani2411: Will remove this as not being used in this PR but will include in another PR.

tests/integration-tests/clusters_factory.py
dreambeyondorange: I'm not sure I follow.  It doesn't make sense to me to do both of these things
```
self.custom_cli_credentials = custom_cli_credentials
...
self.custom_cli_credentials = kwargs.get("custom_cli_credentials")
```
It seems like you are setting the class variable and getting the context from the create_cluster arguments.  So, the first time you set self.custom_cli_credentials is essentially dead code since it's replaced by the second call.  I think in this case since the clusters factory fixture is creating the Cluster object directly and bypassing the ClustersFactory class (a design flaw IMO) we should remove the credentials member variable from the ClustersFactory class and use the credentials provided to the create_cluster method.

himani2411: okay

tests/iam_policies/user-role-rp.cfn.yaml
himani2411: Will put this changes in another PR as its non-blocking improvement.

tests/integration-tests/framework/credential_providers.py
```diff
@@ -25,15 +25,19 @@ def register_cli_credentials_for_region(region, iam_role):
...
 
-    if region in cli_credentials:
-        with sts_credential_provider(region, cli_credentials[region]):
+        if region in cli_credentials:
+            with sts_credential_provider(
+                region, credential_arn=kwargs.get("custom_cli_credentials") or cli_credentials.get(region)
+            ):
+                kwargs.pop("custom_cli_credentials", None)
+                return run_command(*args, **kwargs)
+        else:
```
dreambeyondorange: did you mean for this to be indented under the `if not region`?

himani2411: Reverted the changes

dreambeyondorange (approved):

PR #4692, opened 2022-12-14
[MultiAZ] Shared Storage Integration Tests: External FSx in a MultiAZ cluster

EddyMM:
### Description of changes
- Multi AZ is supported in clusters with Unmanaged FSx FileSystems
- This test creates a Multi-AZ cluster with Unmanaged Ontap, OpenZFS and Lustre FSx FileSystems
- Each partition writes to the FileSystem and reads files written by all partitions
- The test also tries to use a configuration with a Managed FSx FileSystem
    - In this case there is an assertion to check if there is a `ManagedFsxMultiAzValidator` validation error

### Tests
* Creates a MultiAZ Cluster with FSx
* Write from each partition/queue and read all files from each queue (and HeadNode)

### References
* N/A

### Checklist

### DIFF
--- a/tests/integration-tests/tests/storage/test_fsx_lustre.py
+++ b/tests/integration-tests/tests/storage/test_fsx_lustre.py
@@ -269,8 +269,8 @@ def test_multiple_fsx(
     bucket_name = s3_bucket_factory()
     bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
     bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
-    import_path = "s3://{0}".format(bucket_name)
-    export_path = "s3://{0}/export_dir".format(bucket_name)
+    import_path, export_path = "s3://{0}".format(bucket_name), "s3://{0}/export_dir".format(bucket_name)
+
     partition = utils.get_arn_partition(region)
     num_new_fsx_lustre = 1
     num_existing_fsx_ontap_volumes = 2 if partition != "aws-cn" else 0  # China does not have Ontap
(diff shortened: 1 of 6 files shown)


tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -342,6 +321,105 @@ def test_multiple_fsx(
...
+    test_datadir,
+):
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
+    import_path = "s3://{0}".format(bucket_name)
+    export_path = "s3://{0}/export_dir".format(bucket_name)
+
+    partition = utils.get_arn_partition(region)
+    num_existing_fsx_ontap_volumes = 2 if partition != "aws-cn" else 0  # China does not have Ontap
```
lukeseawalker: do we really need to test multiple FS of the same type here? I think this is already covered by the other test

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -342,6 +321,105 @@ def test_multiple_fsx(
...
+    existing_fsx_lustre_fs_ids = _create_fsx_lustre_volume_ids(
+        num_existing_fsx_lustre, fsx_factory, import_path, export_path
+    )
+    fsx_on_tap_volume_ids = _create_fsx_on_tap_volume_ids(num_existing_fsx_ontap_volumes, fsx_factory, svm_factory)
+    fsx_open_zfs_volume_ids = _create_fsx_open_zfs_volume_ids(
+        num_existing_fsx_ontap_volumes, fsx_factory, open_zfs_volume_factory
+    )
+
+    cluster_config = pcluster_config_reader(
+        config_file="pcluster-managed-fsx.config.yaml",
```
lukeseawalker: is this correct? I think it should point to `pcluster.config.yaml`

tests/integration-tests/tests/storage/test_fsx_lustre.py
EddyMM: I agree, we can use one type (Lustre probably)

tests/integration-tests/tests/storage/test_fsx_lustre.py
EddyMM: Good catch. Left over from a local test. Fixing this ...

NSsirena: Since the test here is checking different behavior between managed and unmanaged fsx renaming this config into pcluster-unmanaged-fsx.config.yaml (and accordingly the local variable) would make the code more readable

lukeseawalker (approved):

PR #4704, opened 2022-12-15
develop: Limit SSH connection to the cluster from EC2 instance running the test

enrico-usai:
### Description of changes
By default the `AllowedIps` value is set to `0.0.0.0/0` with this patch we're limiting the access to the head node by default from instance running the test, if it is an EC2 instance.

This SSH access is used by using the SSH key to execute actions in the head node (e.g. job submission)
and by using the SSH user/password for example by AD tests to verify AD setup is working as expected.

### Tests
* Manual execution of test_ad_integration from local PC and from EC2 instance.
* I see `2022-12-15 17:07:59,685 - INFO - 8978 - test_ad_integration[eu-west-1-c5.xlarge-alinux2-slurm-benchmarks0-SimpleAD-ldaps-True] - conftest - Limiting AllowedIps rule to IP: 52.48.85.144` message when running it in an EC2 instance.

### Checklist

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -74,6 +74,7 @@ from utils import (
     get_architecture_supported_by_instance_type,
     get_arn_partition,
     get_instance_info,
+    get_metadata,
     get_network_interfaces_count,
     get_vpc_snakecase_value,
     random_alphanumeric,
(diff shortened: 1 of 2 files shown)


lukeseawalker (dismissed):

tests/integration-tests/conftest.py
```diff
@@ -608,6 +609,11 @@ def inject_additional_config_settings(  # noqa: C901
     with open(cluster_config, encoding="utf-8") as conf_file:
         config_content = yaml.safe_load(conf_file)
 
+    if not dict_has_nested_key(config_content, ("HeadNode", "Ssh", "AllowedIps")):
+        # Permit SSH connection from Jenkins instance only
+        instance_public_ip = get_metadata("public-ipv4")
```
lukeseawalker: what if test is launched from a local machine (not an instance)?

enrico-usai: Fixed

lukeseawalker (dismissed):

tests/integration-tests/conftest.py
```diff
@@ -608,6 +609,12 @@ def inject_additional_config_settings(  # noqa: C901
     with open(cluster_config, encoding="utf-8") as conf_file:
         config_content = yaml.safe_load(conf_file)
 
+    if not dict_has_nested_key(config_content, ("HeadNode", "Ssh", "AllowedIps")):
+        # If the test is running in an EC2 instance limit SSH connection access from instance running the test
+        instance_public_ip = get_metadata("public-ipv4", raise_error=False)
+        if instance_public_ip:
+            dict_add_nested_key(config_content, f"{instance_public_ip}/32", ("HeadNode", "Ssh", "AllowedIps"))
```
gmarciani: Are we sure that public ips are always assigned to the calling instance?
If not, we should use private IP that we know is always assigned, right?

gmarciani (dismissed):
Approved with the non blocking comment about using private IPs rather than public IPs.

gmarciani (approved):

PR #4740, opened 2022-12-20
[Fix] Passing the expected cluster stack scope to apply permissions boundary to the cluster.

gmarciani:
### Description of changes
1. Fix the cluster stack scope passed to apply the permissions boundary to the cluster;
2. Add a new unit test to cover the fixed flow;

#### User Experience
The change fixes a failing path that is triggered whenever a user attempts to create/update a cluster applying a permissions boundary. For example, without this change, the following configuration:

```
...
Iam:
  PermissionsBoundary: "arn:aws:iam::REDACTED:policy/AdminPermissionsBoundary"
...
```

would trigger the following customer facing error:

```
{
  "message": "Cluster update failed.\ntype of argument scope must be constructs.Construct; got pcluster.templates.cluster_stack.ClusterCdkStack instead"
}
```

### Tests
1. Unit tests (existing and the newly added to reproduce the failure we are fixing with this change)

### DIFF
--- a/cli/tests/pcluster/templates/test_iam.py
+++ b/cli/tests/pcluster/templates/test_iam.py
@@ -1,0 +1,43 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from assertpy import assert_that
+
+from pcluster.schemas.cluster_schema import ClusterSchema
+from pcluster.templates.cdk_builder import CDKTemplateBuilder
+from pcluster.utils import load_yaml_dict
+from tests.pcluster.aws.dummy_aws_api import mock_aws_api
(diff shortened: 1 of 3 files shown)


cli/tests/pcluster/templates/test_iam.py
```diff
@@ -0,0 +1,44 @@
...
+
+    generated_template = CDKTemplateBuilder().build_cluster_template(
+        cluster_config=cluster_config, bucket=dummy_cluster_bucket(), stack_name="clustername"
+    )
+
+    role_head_node = get_resources(generated_template, type="AWS::IAM::Role", name="RoleHeadNode").get("RoleHeadNode")
+
+    assert_that(role_head_node).is_not_none()
+
+    assert_that(role_head_node["Properties"]["PermissionsBoundary"]).is_equal_to(permissions_boundary)
```
awslitvit, line 43: Shouldn't we check keys exist? The way it is implemented right now, if `PermissionsBoundary` does not exist we would get a less speaking error.

gmarciani: I agree

gmarciani: I think the following could save us from the need of asserting key existence:

1. we can assume that Properties always exists since we are asserting on CDK resources;
2. we can retrieve "PermissionsBoundary" with the get, so if that does not exist then the assertion would mention it clearly saying that it is None

awslitvit: It would work. However, in order to concatenate several gets without relying on keys existence, you should be able to use `map.get("key1", {}).get("key2", {})`

gmarciani: Experimentally checked and agreed that key indexing would trigger a more understandable error than the retrieval via get(), so I will revert to the first proposal

awslitvit (approved):

PR #4748, opened 2022-12-21
[3.x] Add unit tests to increase cluster CFN template test coverage

awslitvit:
In order to facilitate unit testing, sids have been added to production code.

### Description of changes
Add unit tests for CloudFormation template generation code parts which were not covered.
Also, in order to enable testing sids have been added to production code.

### Tests
* Existing and added unit tests
* Since I didn't find any official documentation on Sid max length, I manually created a cluster using a sid which was longer than any of the newly introduced ones.

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -695,6 +695,7 @@ class HeadNodeIamResources(NodeIamResourcesBase):
             if capacity_reservation_ids:
                 policy.append(
                     iam.PolicyStatement(
+                        sid="AllowRunningReservedCapacity",
                         actions=["ec2:RunInstances"],
                         effect=iam.Effect.ALLOW,
                         resources=[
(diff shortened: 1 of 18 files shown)


cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -695,6 +695,7 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
             if capacity_reservation_ids:
                 policy.append(
                     iam.PolicyStatement(
+                        sid="AllowRunningReservedCapacity",
```
gmarciani, line 698: [Minor non blocking] we have two statements related to reserved capacity and both are focused on *running* the capacity:
1. one by reservation id, so the appropriate Sid would be "AllowRunningReservedCapacityByReservationId"
2. another one (line 715) by resource group, so the appropriate Sid would be "AllowRunningReservedCapacityByResourceGroup"

I'm telling this just because "AllowManagingReservedCapacity" does not give the sense of the actual permissions within it.

gmarciani (approved):

cli/src/pcluster/templates/cdk_builder_utils.py
awslitvit: Since this is just a way to run unit tests, I am going to merge the PR and to move the discussion offline. Another PR will be submitted to change sid, if required.

PR #4749, opened 2022-12-21
[3.x] test non-happy ontap and openfsx mounting flow

Nbot07A:
### Description of changes
* test to ensure failure of mounting when mounting should fail

### Tests
* added and passed test_fsx.py::test_non_happy_describe_volumes and test_shared_storage.py::test_non_happy_ontap_and_openzfs_mounting

### Checklist


cli/src/pcluster/validators/fsx_validators.py
```diff
@@ -222,3 +222,16 @@ def _validate(self, auto_import_policy, import_path):
             bucket = get_bucket_name_from_s3_url(import_path)
             if AWSApi.instance().s3.get_bucket_region(bucket) != get_region():
                 self._add_failure("FSx auto import is not supported for cross-region buckets.", FailureLevel.ERROR)
+
+
+class FsxDescribeVolumesValidator(Validator):
```
gmarciani: The goal of the PR is to implement a unit test to prove that when the user does not have the permissions to describe the volumes, then the generation of the cluster template would fail.

Why are we adding a new validator?

cli/src/pcluster/validators/fsx_validators.py
```diff
@@ -222,3 +222,16 @@ def _validate(self, auto_import_policy, import_path):
...
+    """Backup id validator."""
```
gmarciani: Why Backup?

cli/src/pcluster/validators/fsx_validators.py
```diff
@@ -222,3 +222,16 @@ def _validate(self, auto_import_policy, import_path):
...
+
+    def _validate(self, volume_ids):
+        try:
+            AWSApi.instance().fsx.describe_volumes(volume_ids)
```
gmarciani: I assume that you're adding this AWS call to trigger an interaction with FSx, but our goal here is to test the existent interaction.

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -406,6 +412,7 @@ def test_scheduler_plugin_all_validators_are_called(test_datadir, mocker):
                 "CapacityReservationResourceGroupValidator",
                 "DatabaseUriValidator",
                 "InstanceTypePlacementGroupValidator",
+                "FsxDescribeVolumesValidator",
```
gmarciani: We do not want to add new validators in the production code

gmarciani (changes requested):

cli/src/pcluster/validators/fsx_validators.py
Nbot07A: This approach is much more straightforward than attempting to mock the failure of `describe_volumes` from within `test_shared_storage.py`

cli/src/pcluster/validators/fsx_validators.py
Nbot07A: Sorry, this is a remnant from a reference. So, I'll change this.

cli/src/pcluster/validators/fsx_validators.py
Nbot07A: Yes, however mocking the existent interaction even within a unit test seems to be much more involved than taking this approach.

cli/tests/pcluster/validators/test_all_validators.py
Nbot07A: I am also concerned about this even though this exact line actually skips the check of the new validator. It seems that this approach to truly be complete could introduce risks, but the alternative approach e.g. from `test_shared_storage.py` is unclear. So, I chose this way of adding a validator, but not having it enforced as a potential middle ground between the riskier and the convoluted approaches.

cli/src/pcluster/validators/fsx_validators.py
gmarciani: In general we should always pursue the simplicity of a solution, but the first goal is the correctness of the solution.
Here, we want to mock the existent interaction; we do not want to introduce new validation logic in production.

cli/src/pcluster/validators/fsx_validators.py
gmarciani: It could be considered more straightforward, but still it's not correct: here, we want to mock the existent interaction; we do not want to introduce new validation logic in production.

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -87,11 +88,11 @@ def instance_storage_supported(self):
 
 
 class _DummyAWSApi(AWSApi):
-    def __init__(self):
+    def __init__(self, non_happy=False):
```
gmarciani: I have a strong concern about this approach.
A user of this constructor cannot understand the implications of setting `non_happy`:
1. setting `non_happy` at the constructor level would make me think that it would imply non happy flows in every client, instead it does it only on FSx;
2. what does `non_happy` mean? which calls are affected by this settings?

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -196,8 +197,9 @@ def get_efs_mount_target_security_groups(self, target_id):
 
 
 class _DummyFSxClient(FSxClient):
-    def __init__(self):
+    def __init__(self, non_happy=False):
         """Override Parent constructor. No real boto3 client is created."""
+        self.non_happy = non_happy
         pass
```
gmarciani: Unnecessary pass

cli/tests/pcluster/validators/test_all_validators.py
gmarciani: Let's remove the validator: we want to test an existing behaviour leveraging mocking, not introducing new validation logic into the production code.

cli/tests/pcluster/validators/test_fsx_validators.py
```diff
@@ -438,3 +439,15 @@ def test_auto_import_policy_validator(mocker, auto_import_policy, cluster_region
 
     actual_failures = FsxAutoImportValidator().execute(auto_import_policy, "s3://test/test1/test2")
     assert_failure_messages(actual_failures, expected_message)
+
+
+def test_fsx_non_happy_ontap_openfsx_mounting(mocker):
```
gmarciani: Let's remember to remove this unit tests covering the newly added validator, as this validator should not be added to production.

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -337,8 +342,8 @@ def describe_secret(self, secret_arn):
         }
 
 
-def mock_aws_api(mocker, mock_instance_type_info=True):
+def mock_aws_api(mocker, mock_instance_type_info=True, non_happy=False):
     """Mock AWS Api."""
-    mocker.patch("pcluster.aws.aws_api.AWSApi.instance", return_value=_DummyAWSApi())
+    mocker.patch("pcluster.aws.aws_api.AWSApi.instance", return_value=_DummyAWSApi(non_happy))
```
gmarciani: You added the new parameter `non_happy` to `_DummyAwsApi` to inject the unhappy behaviour into the FSx client. The problems with this approach are:
1. the `non_happy` parameter should inject the unhappy behaviour into the FSx client only, but your are setting at the overall API level, so from the outside one could expect that the unhappy behaviour would affect also other APIs; In case this would be the correct approach (but it is not), the name of the parameter should have been more specific to the failing API, e.g. `non_happy_fsx`. 
2. in this unit test we want to inject an unhappy behaviour specifically into the FSx API, without any other impact on other APIs, so I would expect the `mocker.patch` method to patch the specific FSx function, e.g. `pcluster.aws.fsx.FSxClient.describe_volumes` (or whatever other API call you want to mock);

cli/src/pcluster/validators/fsx_validators.py
Nbot07A: Ok, that makes sense.

cli/tests/pcluster/validators/test_all_validators.py
Nbot07A: Ok

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: I could rename the parameter to `test_non_happy_flow` for clarity. This can be applied to any test needing a mock for a non-happy flow. Also, as it is a keyword argument it has no impact on existing code and can facilitate any future testing of non-happy flows.

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: Thanks, I'll remove this

cli/tests/pcluster/validators/test_fsx_validators.py
Nbot07A: Yes

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: I agree with your first point, however, I disagree about it's impact to other API calls as the argument is optional and therefore does not affect any existing code. Furthermore, this approach enables faster implementation of similar non-happy tests in the future. As for your second point, I agree that would be ideal, however that does not work as expected. Using the exact example you gave fails because the actual `FSxClient` from `fsx` is not used due to being replaced in the mocking logic. Furthermore, a similar approach fails i.e. attempting to modify the exact mocked `describe_volumes` function because the testing code, as seen in the validator approach's implementation, requires an actual function from source code to mock. So, passing the argument for non-happy flows through the mocking architecture is the simplest, correct, solution that I've found.

cli/tests/pcluster/aws/dummy_aws_api.py
gmarciani: I think we should avoid adding this parameter. What if we want to inject failure in N API calls? With the approach you're proposing we should add N parameters to the `DummyAwsApi` constructor and it would not scale.

what about something like the following?

```
def your_unit_test(mocker):
    dummy_api = _DummyAWSApi()
    dummy_api._fsx.make_describe_volumes_fail_with_error(True, error=UnauthorizedException)
    mocker.patch("pcluster.aws.aws_api.AWSApi.instance", return_value=dummy_api)

   ... your unit tests logic ...
```

cli/tests/pcluster/aws/dummy_aws_api.py
gmarciani: I understand the challenge you're facing.
We can overcome this blocker with the approach I proposed in the above comment:

```
def your_unit_test(mocker):
    dummy_api = _DummyAWSApi()
    dummy_api._fsx.make_describe_volumes_fail_with_error(True, error=UnauthorizedException)
    mocker.patch("pcluster.aws.aws_api.AWSApi.instance", return_value=dummy_api)

   ... your unit tests logic ...
```

cli/tests/pcluster/templates/test_shared_storage.py
```diff
@@ -202,6 +203,21 @@ def assert_sg_rule(
     assert_that(sg_rules).is_length(1)
 
 
+def test_non_happy_ontap_and_openfsx_mounting(mocker, test_datadir):
+    dummy_api = _DummyAWSApi()
+    dummy_api._fsx.set_non_happy(True)
+    mocker.patch("pcluster.aws.aws_api.AWSApi.instance", return_value=dummy_api)
+    mocker.patch("pcluster.aws.ec2.Ec2Client.get_instance_type_info", side_effect=_DummyInstanceTypeInfo)
```
gmarciani, line 212: is this patching required for this unit test?
If not, let's remove this line.

cli/tests/pcluster/templates/test_shared_storage.py
```diff
@@ -202,6 +203,21 @@ def assert_sg_rule(
     assert_that(sg_rules).is_length(1)
 
 
+def test_non_happy_ontap_and_openfsx_mounting(mocker, test_datadir):
+    dummy_api = _DummyAWSApi()
+    dummy_api._fsx.set_non_happy(True)
```
gmarciani: We want to be more specific here: how can I understand from `_fsx.set_non_happy(True)` that you're injecting a faulty behaviour for the describe_volume API? It should rather be something more specific like `mock_describe_volume_failure(True, AWSClientError)`

cli/tests/pcluster/aws/test_fsx.py
```diff
@@ -124,3 +125,20 @@ def test_describe_volumes(boto3_stubber):
...
+def get_non_happy_describe_volumes(volume_ids):
+    return MockedBoto3Request(
+        method="describe_volumes",
+        response="Error",
+        generate_error=True,
+        expected_params={"VolumeIds": volume_ids},
+    )
+
+
+def test_non_happy_describe_volumes(boto3_stubber):
```
gmarciani: Why do we need this unit tests when we already have the below one `test_non_happy_ontap_and_openzfs_mounting`?

cli/tests/pcluster/templates/test_shared_storage.py
```diff
@@ -202,6 +203,21 @@ def assert_sg_rule(
     assert_that(sg_rules).is_length(1)
 
 
+def test_non_happy_ontap_and_openfsx_mounting(mocker, test_datadir):
```
gmarciani: Typo: it's not `openfsx`, but `openzfs`

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -208,8 +209,14 @@ def get_filesystem_info(self, fsx_fs_id):
             },
         }
 
+    def set_non_happy(self, non_happy):
```
gmarciani: `non_happy` is too generic. How could someone understand that this setting refers specifically to the describe volume API call?

cli/tests/pcluster/aws/test_fsx.py
Nbot07A: This serves as an example of the underlying failure and a test to demonstrate it. However, it could be omitted if it doesn't provide much value beyond `test_non_happy_ontap_and_openzfs_mounting` does.

cli/tests/pcluster/templates/test_shared_storage.py
Nbot07A: Thanks, I'll update this.

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: It is currently only being used in the describe volume call, but I will rename this more descriptively so it also clear without any knowledge of the implementation.

cli/tests/pcluster/templates/test_shared_storage.py
Nbot07A: Yes, I agree. I'll rename it.

cli/src/pcluster/validators/fsx_validators.py
Nbot07A: this was removed

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: I like that approach

cli/tests/pcluster/templates/test_shared_storage.py
```diff
@@ -202,6 +203,21 @@ def assert_sg_rule(
     assert_that(sg_rules).is_length(1)
 
 
+def test_non_happy_ontap_and_openzfs_mounting(mocker, test_datadir):
+    dummy_api = _DummyAWSApi()
+    dummy_api._fsx.set_non_happy_describe_volumes(True)
```
gmarciani: set_non_happy_describe_volumes(True) says that we are injecting a failure in the describe volumes API call, but it does not make any specification about the nature of such failure. Internally you’re injecting a client error saying unauhtorized exception, but you’re not in control of that

So, we have two options:
rename set_non_happy_describe_volumes  in a way that reflects the fact that you’re forcing an unauthorized error
add a parameter to set_non_happy_describe_volumes  to specify the nature of the raised error

cli/tests/pcluster/aws/dummy_aws_api.py
Nbot07A: Ok

cli/tests/pcluster/templates/test_shared_storage.py
Nbot07A: Yes, it is necessary as the Ec2Client expects to have instance_type_info. Otherwise, the test fails.

cli/tests/pcluster/templates/test_shared_storage.py
Nbot07A: Ok, this is addressed here [add error parameter for set_non_happy_describe_volumes](https://github.com/aws/aws-parallelcluster/pull/4749/commits/784f24a26faefe0a7fa76fab0e2e498aa1ca71b5) and [remove boolean parameter from set_non_happy_describe_volumes](https://github.com/aws/aws-parallelcluster/pull/4749/commits/d7da791e830d9555bbb9ce52d84174302efead58)

cli/tests/pcluster/templates/test_shared_storage.py
gmarciani: Thank you for the clarification.

gmarciani (approved):

PR #4754, opened 2022-12-22
[Test] Include testing for detached retained managed storage to the integration tests covering the dynamic file system mounting.

gmarciani:
### Description of changes
* Included testing for detached retained managed storage to the integration tests covering the dynamic file system mounting. In particular we want to verify the following use case: 
    * attach to the cluster all types of managed storage with `DeletionPolicy` set to `Delete`;
    * update the cluster to set the `DeletionPolicy` to `Retain` for every managed storage;
    * update the cluster to detach the managed storage;
    * verify that the managed storage is retained detached from the cluster;
    * delete the retained managed storage to avoid dangling resources;
* Collecting storage related utilities in dedicated folder `common/storage`
* Added some log lines to the original test to better understand the phase of the test
* Specified the config output file to make the same template reusable.

### Tests
1. Integ test `test_update.py::test_dynamic_file_systems_update` locally launched in personal account, both with and without `--no-delete` option, and succeeded

### DIFF
--- a/tests/integration-tests/tests/common/storage/efs_utils.py
+++ b/tests/integration-tests/tests/common/storage/efs_utils.py
@@ -1,0 +1,86 @@
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from retrying import retry
+from time_utils import seconds
+
+from tests.common.networking.security_groups import delete_security_group
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(5))
+def describe_efs_filesystem(region: str, file_system_id: str):
+    logging.info(f"Describing EFS File System {file_system_id}")
+    try:
+        return _efs(region).describe_file_systems(FileSystemId=file_system_id)["FileSystems"][0]
+    except Exception as e:
+        if isinstance(e, ClientError) and "FileSystemNotFound" in str(e):
+            return None
+        else:
+            logging.error(f"Cannot describe EFS File System {file_system_id}: {e}")
+            raise e
+
+
+@retry(stop_max_attempt_number=10, wait_fixed=seconds(30))
+def delete_efs_filesystem(region: str, file_system_id: str, delete_dependent_resources: bool = True):
+    logging.info(f"Deleting EFS File System {file_system_id}")
+    try:
+        if delete_dependent_resources:
+            mount_targets = describe_efs_mount_targets(region, file_system_id)
+            security_group_ids = set()
+            for mount_target in mount_targets:
+                mount_target_id = mount_target["MountTargetId"]
+                security_groups = describe_mount_target_security_groups(region, mount_target_id)
+                for security_group_id in security_groups:
+                    security_group_ids.add(security_group_id)
+                delete_efs_mount_target(region, mount_target_id)
+            logging.info(
+                "The following Security Groups will be deleted as part of "
+                f"the deletion for the EFS File System {file_system_id}: {security_group_ids}"
+            )
+            for mount_target in mount_targets:
+                mount_target_id = mount_target["MountTargetId"]
+                wait_for_efs_mount_target_deletion(region, file_system_id, mount_target_id)
+            for security_group_id in security_group_ids:
+                delete_security_group(region, security_group_id)
+        _efs(region).delete_file_system(FileSystemId=file_system_id)
+    except Exception as e:
+        if isinstance(e, ClientError) and "FileSystemNotFound" in str(e):
+            logging.warning(f"Cannot delete EFS File System {file_system_id} because it does not exist")
+        else:
+            logging.error(f"Cannot delete EFS File System {file_system_id}: {e}")
+            raise e
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(5))
+def describe_efs_mount_targets(region: str, file_system_id: str):
+    logging.info(f"Describing Mount Targets for EFS File System {file_system_id}")
+    return _efs(region).describe_mount_targets(FileSystemId=file_system_id).get("MountTargets", [])
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(5))
+def describe_mount_target_security_groups(region: str, mount_target_id: str):
+    logging.info(f"Describing Security Groups for EFS Mount Target {mount_target_id}")
+    return _efs(region).describe_mount_target_security_groups(MountTargetId=mount_target_id).get("SecurityGroups", [])
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(5))
+def delete_efs_mount_target(region: str, mount_target_id: str):
+    logging.info(f"Deleting EFS Mount Target {mount_target_id}")
+    _efs(region).delete_mount_target(MountTargetId=mount_target_id)
+
+
+@retry(stop_max_attempt_number=10, wait_fixed=seconds(60))
+def wait_for_efs_mount_target_deletion(region: str, file_system_id: str, mount_target_id: str):
+    logging.info(f"Waiting for deletion of EFS Mount Target {mount_target_id} in EFS File System {file_system_id}")
+    mount_targets = describe_efs_mount_targets(region, file_system_id)
+    mount_target_ids = [mt["MountTargetId"] for mt in mount_targets]
+    if mount_target_id in mount_target_ids:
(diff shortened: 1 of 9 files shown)


tests/integration-tests/tests/common/storage/fsx_utils.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from retrying import retry
+from time_utils import seconds
+
+
+@retry(stop_max_attempt_number=3, wait_fixed=seconds(5))
+def describe_fsx_filesystem(region: str, file_system_id: str):
```
himani2411, line 15: This function seems like a duplication of ```describe_efs_filesystem()```. Can we not make a generic or common function?

tests/integration-tests/tests/common/storage/fsx_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        _fsx(region).delete_file_system(FileSystemId=file_system_id)
+    except Exception as e:
+        if isinstance(e, ClientError) and "FileSystemNotFound" in str(e):
+            logging.warning(f"Cannot delete FSx File System {file_system_id} because it does not exist")
+        else:
+            logging.error(f"Cannot delete FSx File System {file_system_id}: {e}")
+            raise e
+
+
+def _fsx(region):
```
himani2411, line 72: Another duplication though can be ignored is,  ```_efs(region), _ec2(region) and _fsx(region)``` can be combined as one single function where we just pass a parameter of the client we want.

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
```diff
@@ -54,12 +54,13 @@ SharedStorage:
...
         Type: 0
         NumberOfVolumes: 5
+      DeletionPolicy: {{ new_raid_deletion_policy }}
   - MountDir: {{ new_ebs_mount_dir }}
     Name: /manage-ebs
     StorageType: Ebs
     EbsSettings:
       VolumeType: gp3
-      DeletionPolicy: Delete
+      DeletionPolicy: {{ new_ebs_deletion_policy }}
```
himani2411, line 61: If the Deletion policy for all the storage types is same, then why not keep a single variable like ``` storage_deletion_policy```.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -997,6 +1045,28 @@ def test_dynamic_file_systems_update(
...
+            region, StorageType.STORAGE_EFS, filesystem_id, should_exist=True, expected_states=["available"]
+        )
+        delete_storage_on_teardown(StorageType.STORAGE_EFS, filesystem_id)
+
+    # Verify that detached FSx File Systems have been retained and delete them.
+    for filesystem_id in retained_fsx_filesystem_ids:
+        assert_storage_existence(
+            region, StorageType.STORAGE_FSX, filesystem_id, should_exist=True, expected_states=["AVAILABLE"]
+        )
+        delete_storage_on_teardown(StorageType.STORAGE_FSX, filesystem_id)
```
himani2411: These 3 for loops can be combined to form a function which can be used used with different parameters for EBS, EFS and Fsx

tests/integration-tests/tests/common/storage/efs_utils.py
```diff
@@ -0,0 +1,67 @@
...
+            mount_targets = describe_efs_mount_targets(region, file_system_id)
+            security_group_ids = set()
+            for mount_target in mount_targets:
+                mount_target_id = mount_target["MountTargetId"]
+                security_groups = describe_mount_target_security_groups(region, mount_target_id)
+                for security_group_id in security_groups:
+                    security_group_ids.add(security_group_id)
+                delete_efs_mount_target(region, mount_target_id)
+            for security_group_id in security_group_ids:
+                delete_security_group(region, security_group_id)
```
hanwen-pcluste, line 45: How does ParallelCluster manage security group of retained storage?

gmarciani: When an EFS file system is created by ParallelCluster, the following security group are created along with it:
1. a SG to allow in/outbound traffic within the file system SG itself
2. a SG to allow in/out bound traffic with the head node
3. a SG to allow in/out bound traffic with the compute nodes

When the file system is detached from the cluster with DeletionPolicy=Retain, only the SecurityGroup (1) is retained with the file system. It is like this because that one is the only security group that is strictly related to the file system regardless the specific cluster. The other twos are specific to the cluster so it makes sense to delete them when the file system is detached.

gmarciani: Your comment made me think that the FSx file system too should have the equivalent security group to be removed. Doing that

tests/integration-tests/tests/common/storage/fsx_utils.py
gmarciani: I understand your point. Nevertheless, in that case it would be equivalent to directly use `boto3("service-name", region)`.

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
gmarciani: That's a good question. The reason is that having different variables makes the test more flexible because in this way it would be ready to use different deletion policies for different storage types.

tests/integration-tests/tests/update/test_update.py
gmarciani: Yes, I was really tempted to do it before submitting the PR. The reason why I didn't do it is that I was afraid that we would lose readability. Nevertheless, I think we could reduce the duplication with the following solution and IO actually like it. What do you think?

```
    retained_storage = {
        StorageType.STORAGE_EBS: dict(ids=retained_ebs_volume_ids, expected_states=["available"]),
        StorageType.STORAGE_EFS: dict(ids=retained_efs_filesystem_ids, expected_states=["available"]),
        StorageType.STORAGE_FSX: dict(ids=retained_fsx_filesystem_ids, expected_states=["AVAILABLE"]),
    }
    for storage_type in retained_storage:
        for storage_id in retained_storage[storage_type]["ids"]:
            assert_storage_existence(region, storage_type, storage_id, should_exist=True, expected_states=retained_storage[storage_type]["expected_states"])
            delete_storage_on_teardown(storage_type, storage_id)
```

himani2411: Yes this is a good solution

tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/pcluster.config.update_drain.yaml
himani2411: Got it. Thanks

tests/integration-tests/tests/common/storage/fsx_utils.py
himani2411: Yes, I believe so.

tests/integration-tests/tests/common/storage/fsx_utils.py
```diff
@@ -0,0 +1,73 @@
...
+        else:
+            logging.error(f"Cannot describe FSx File System {file_system_id}: {e}")
+            raise e
+
+
+@retry(stop_max_attempt_number=10, wait_fixed=seconds(30))
+def delete_fsx_filesystem(region: str, file_system_id: str, delete_dependent_resources: bool = True):
+    logging.info(f"Deleting FSx File System {file_system_id}")
+    try:
+        security_group_ids = set()
```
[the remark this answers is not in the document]
gmarciani: fixing

hanwen-pcluste (dismissed):

tests/integration-tests/tests/common/storage/fsx_utils.py
gmarciani: Yes we could, but I think we should apply the 3strikes rule: implement a common logic and refactor the exisinting one if the common logic would be reused at least three times.

hanwen-pcluste (approved):

PR #4767, opened 2022-12-28
[integ-tests-3.4.0] Moving IMDSv1 test in MXP

francesco-giordano:
### Description of changes
* Move `test_kernel4_build_image_run_cluster` test to MXP to have all IMDSv1 test there

### Tests
* Running one of the tests with test-runner

### Checklist

### DIFF
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -154,7 +154,7 @@ createami:
         oss: ["alinux2", "ubuntu2004", "centos7"]
   test_createami.py::test_kernel4_build_image_run_cluster:
     dimensions:
-      - regions: ["ca-central-1"]
+      - regions: ["eu-south-1"]
         instances: {{ common.INSTANCES_DEFAULT_X86 }}
         schedulers: ["awsbatch", "slurm"]
         oss: ["alinux2"]


tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.update.yaml
```diff
@@ -1,3 +1,5 @@
+Imds:
+  ImdsSupport: {{ imds_support }}
```
ermanno: I have a question regarding the necessity to explicitly set the `ImdsSupport` field in these templates. This also applies to the patches linked in the description, and in particular to [the one for the "develop" branch](https://github.com/aws/aws-parallelcluster/pull/4745). Afaik [this code should inject this property in build image configs](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/conftest.py#L585) and [this in cluster configs](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/conftest.py#L625), could you explain (maybe in the description to this PR) why it's necessary to also add it to the single templates, so that it's documented?

Actually, are we sure that we need this? Afaik `inject_additional_image_configs_settings` and `inject_additional_config_settings` were added exactly to handle these kinds of situations, where we have a setting that we want to apply by default to all configs (unless explicitly overridden)... I don't particularly like having it explicitly in the template, in case in the future we mistakenly add a new template without this setting.

ermanno (approved):

PR #4774, opened 2023-01-03
[3.x] Create Runtime CFN template for Custom CLI Credentails

himani2411:
### Description of changes
* Creating custom CLI credentials to run Resource prefix integration test case through a runtime change in existing CFN template user-role.cfn.yaml
* Removing template [user-role-iam-resource-prefix.cfn.yaml](https://github.com/aws/aws-parallelcluster/compare/develop...himani2411:aws-parallelcluster:wip/runtime-iam-role#diff-6c891a2944891d4a7fea672dba07bc0ef430cf2e316f196359616003b04d3123) which was used for creating CLI credentials.

### Tests
* Running the test_iam_resource_prefix test case locally
* Running jenkins pcluster3 test cases

### Checklist


tests/iam_policies/user-role.cfn.yaml
```diff
@@ -823,10 +823,12 @@ Resources:
               - dynamodb:BatchWriteItem
             Resource: !Sub arn:${AWS::Partition}:dynamodb:${AWS::Region}:${AWS::AccountId}:table/parallelcluster-*
             Effect: Allow
-          - Action: ec2:DescribeInstances
```
dreambeyondorange, line 826: Why do we need to convert all these to lists?  Does this affect current tests?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -363,6 +369,233 @@ def test_iam_resource_prefix(
             _test_iam_resource_in_cluster(cfn_client, iam_client, cluster.name, iam_resource_prefix)
 
 
+def _remove_from_config(config, item_list):
```
dreambeyondorange: Seems like you are building some general purpose utilities to update the config files.  Can you add descriptions and explanations for the functions you've created here?  Its a bit hard to follow how they all fit together.

tests/iam_policies/user-role.cfn.yaml
himani2411: No this doesnt affect the test. I just converted it to a list so that when I am trying to parse this through a template generator and make my changes, I can use a generic functionality rather than creating another function to handle it as all the other policy statements have been a list so far.

himani2411: I hav already tested the changes using jenkins

tests/integration-tests/tests/iam/test_iam.py
himani2411: Sure.

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -363,6 +369,247 @@ def test_iam_resource_prefix(
...
+    """Remove a list of sections from Config template."""
+    if isinstance(config, dict):
+        for item in section_list:
+            config.pop(item)
+    elif isinstance(config, list):
+        for item in section_list:
+            config.remove(item)
+
+
+def _replace_resource(config_string, list_of_dict):
```
dreambeyondorange, line 382: Is this replacing all the different substrings with one?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -363,6 +369,247 @@ def test_iam_resource_prefix(
             _test_iam_resource_in_cluster(cfn_client, iam_client, cluster.name, iam_resource_prefix)
 
 
+def _remove_from_config(config, section_list):
```
dreambeyondorange, line 372: It seems like you are doing the replacement manually.  Why not use a tool like jinja for this?

tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -363,6 +369,247 @@ def test_iam_resource_prefix(
...
+        ):
+            permission_boundary_statements.remove(statement)
+
+
+def _find_resource_action(resource_type, action_list):
+    """Check if a specific resource type exists in the Policy statement's Actions."""
+    return any(resource_type in action for action in action_list)
+
+
+def _create_iam_user_role_config(cluster_config, iam_resource_prefix):
```
gmarciani, line 542: #### Major
1. In general, we should always  use Jinja to make runtime changes to a cloudformation template instead of ad-hoc "manual changes".
2. We can consider the following alternative approach instead of making direct changes to the user role template: implement a fixture that creates a permissions boundary policy according to the given IAM prefix and applies it to the user role

#### Minor
The value returned by this function is a stack template, not a generic config. I suggest to rename this function "_create_iam_user_role_template"

gmarciani (changes requested):

tests/integration-tests/tests/iam/test_iam.py
himani2411: Creating a new [PR](https://github.com/aws/aws-parallelcluster/pull/4782) to address new design changes

PR #4778, opened 2023-01-04
[3.x] Pre-import cdk libraries at the beginning of cluster create and update

awslitvit:
### Description of changes
Import of cdk libraries at lambda cold start takes 12 to 16 seconds, sometimes causing lambda timeout. 
Importing them in a separate thread, started at the beginning of create and update operations and executed concurrently with stack validation allows to reduce the time to complete the operation.
The amount of the execution time gain depends on how long the validation takes.
Longer is the validation more impactful this optimization gets.

### Tests
Manual test via API
* create-cluster dryrun
  * no import performed
* create-cluster no dryrun
  * import starts 1 msec after the lambda execution starts
  * validation takes ~ 5.5 sec.
  * total time to import cdk: ~ 12 sec., waiting time post validation before starting cdk generation: ~ 6.5 sec.
* update-cluster dryrun
  * no import performed
* update-cluster no dryrun
  * import starts 1 msec after the lambda execution starts
  * validation takes ~ 6 sec.
  * total time to import cdk: ~ 12 sec., waiting time post validation before starting cdk generation: ~ 6 sec.
Manual test via CLI
* create-cluster in dryrun mode and real cluster creation worked as expected
* update-cluster considered redundant

### References
* [Open issue on cdk import slowness](https://github.com/aws/jsii/issues/3389)

### Checklist

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -66,6 +66,7 @@ from pcluster.models.compute_fleet_status_manager import ComputeFleetStatus, Com
 from pcluster.models.s3_bucket import S3Bucket, S3BucketFactory, S3FileFormat, create_s3_presigned_url, parse_bucket_url
 from pcluster.schemas.cluster_schema import ClusterSchema
 from pcluster.templates.cdk_builder import CDKTemplateBuilder
+from pcluster.templates.import_cdk import start as start_cdk_import
 from pcluster.utils import (
     datetime_to_epoch,
     generate_random_name_with_prefix,
@@ -352,6 +353,7 @@ class Cluster:
         raises ClusterActionError: in case of generic error
         raises ConfigValidationError: if configuration is invalid
         """
+        start_cdk_import()
         creation_result = None
         artifact_dir_generated = False
         try:
@@ -887,6 +889,7 @@ class Cluster:
         raises ConfigValidationError: if configuration is invalid
         raises ClusterUpdateError: if update is not allowed
         """
+        start_cdk_import()
         try:
             target_config, changes, ignored_validation_failures = self.validate_update_request(
                 target_source_config, validator_suppressors, validation_failure_level, force
--- a/cli/src/pcluster/templates/import_cdk.py
+++ b/cli/src/pcluster/templates/import_cdk.py
@@ -1,0 +1,27 @@
+import logging
+import time
+from threading import Thread
+
+LOGGER = logging.getLogger(__name__)
+
+
+def import_cdk():
+    """Import cdk libraries."""
+    LOGGER.info("Start importing")
+    begin = time.time()
+    from aws_cdk.core import App  # noqa: F401 pylint: disable=import-outside-toplevel
+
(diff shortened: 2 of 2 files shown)


cli/src/pcluster/api/controllers/import_cdk.py
```diff
@@ -0,0 +1,27 @@
+import logging
```
demartinofra: is there a reason why this file is in the controllers dir?

demartinofra (commented):
2 questions:
1. what happens if the main thread reaches the cdk import before the warm up thread has completed?can it break due to some race condition?
2. what happens if I exit the cli with ctrl+c while the background thread is running? will it hang?what is present in logs?

cli/src/pcluster/api/controllers/import_cdk.py
awslitvit: Thank you for pointing this out. Moving to `templates`.

awslitvit:
> what happens if the main thread reaches the cdk import before the warm up thread has completed?can it break due to some race condition?

That's exactly the case I have tested. It doesn't break, both threads continue in parallel and the import completes simultaneously in both. I ran this test several times and it never broke.

> what happens if I exit the cli with ctrl+c while the background thread is running? will it hang?what is present in logs?

cli exits without hanging, logging:
```
2023-01-05 10:29:41,985 - INFO - import_cdk.py:10:import_cdk() - Start importing
2023-01-05 10:29:43,273 - INFO - entrypoint.py:244:main() - Received KeyboardInterrupt. Exiting.
```

awslitvit:
Multiple concurrent cdk imports tested:
```
import time
from pcluster.templates.import_cdk import start

print("Starting")
threads = []
for i in range(1, 21):
    threads.append(start())
    time.sleep(0.2)

print("Waiting")

for t in threads:
    t.join()

print("Completed")
```

Output:
```
Starting
Start importing (x20)
Waiting
Import complete in 5.7333149909973145 seconds
Import complete in 5.53127908706665 seconds
Import complete in 5.330389976501465 seconds
Import complete in 5.128042936325073 seconds
Import complete in 4.927767992019653 seconds
Import complete in 4.726654052734375 seconds
Import complete in 4.526315689086914 seconds
Import complete in 4.3224451541900635 seconds
Import complete in 4.120244026184082 seconds
Import complete in 3.9184842109680176 seconds
Import complete in 3.7149658203125 seconds
Import complete in 3.5131521224975586 seconds
Import complete in 3.3125240802764893 seconds
Import complete in 3.1082780361175537 seconds
Import complete in 2.9072659015655518 seconds
Import complete in 2.7062909603118896 seconds
Import complete in 2.5047547817230225 seconds
Import complete in 2.301931619644165 seconds
Import complete in 2.101245880126953 seconds
Import complete in 1.9005579948425293 seconds
Completed
```

Same result with 1000 iterations and delay of 0.01 sec.

awslitvit:
Following code exited with ctrl+c without hanging:

```
from pcluster.templates.import_cdk import start

print("Starting")
thread = start()
print("Waiting")
thread.join()  # ctrl+c here
print("Completed")
```

demartinofra (approved):

PR #4797, opened 2023-01-13
3.x Use more aggressive timeouts for s3api call when downloading custom cookbooks

davprat:
### Description of changes
* Set connection timeout for AWS CLI call to s3api to timeout after 2 minutes with 3 attempts to connect.
* This is done to fail compute node launch faster if there is no connectivity from the compute node to AWS services.
* Before this change it would make 5 attempts with each attempt taking ~8minutes for a total time to failure of ~40m.
* With this change the TTF should be ~6m
* Reason for s3 failure is now included in the user_data.sh exit reason.

### Tests
* Deployed a cluster with a custom cookbook where the compute fleet was deployed into an isolated subnet.
* Compute nodes successfully self-terminated after ~6m.

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/compute_node/user_data.sh
+++ b/cli/src/pcluster/resources/compute_node/user_data.sh
@@ -149,14 +149,21 @@ write_files:
       [ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh
       custom_cookbook=${CustomChefCookbook}
       export _region=${AWS::Region}
+      cookbook_retry_strategy=--retry 3
       s3_url=${AWS::URLSuffix}
       if [ "${!custom_cookbook}" != "NONE" ]; then
         if [[ "${!custom_cookbook}" =~ ^s3://([^/]*)(.*) ]]; then
-          bucket_region=$(aws s3api get-bucket-location --bucket ${!BASH_REMATCH[1]} | jq -r '.LocationConstraint')
+          export AWS_RETRY_MODE=standard
+          export AWS_MAX_ATTEMPTS=2
+          # S3API_RESULT=$(aws s3api get-bucket-location --debug --cli-connect-timeout 60 --bucket ${!BASH_REMATCH[1]} 2>&1 | tee -a /var/log/aws.log) || error_exit "${!S3API_RESULT}"
+          S3API_RESULT=$(aws s3api get-bucket-location --cli-connect-timeout 15 --bucket ${!BASH_REMATCH[1]} 2>&1) || error_exit "${!S3API_RESULT}"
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -149,10 +149,23 @@ write_files:
       [ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh
       custom_cookbook=${CustomChefCookbook}
       export _region=${AWS::Region}
+
+      # The default curl setting will timeout the connect attempt at 5 minutes, so
+      # having the retry count set to 3 will result in 15m to failure if there is
+      # no internet connectivity from the compute node.
+      cookbook_url_retry_strategy=--retry 3
```
chenwany, line 156: It looks like we are not changing the numbers of retry in this PR?  The original retry is also 3 times

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -149,10 +149,23 @@ write_files:
...
       s3_url=${AWS::URLSuffix}
       if [ "${!custom_cookbook}" != "NONE" ]; then
         if [[ "${!custom_cookbook}" =~ ^s3://([^/]*)(.*) ]]; then
-          bucket_region=$(aws s3api get-bucket-location --bucket ${!BASH_REMATCH[1]} | jq -r '.LocationConstraint')
+          export AWS_RETRY_MODE=standard
```
chenwany, line 160: `standard` is the default `AWS_RETRY_MODE` for aws CLI V2.
 I think we are using AWS CLI v1, is `standard` the default `AWS_RETRY_MODE`?

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -149,10 +149,23 @@ write_files:
...
+          export AWS_MAX_ATTEMPTS=2
```
chenwany, line 161: With export `AWS_MAX_ATTEMPTS` variable here, is this also going to impact the numbers of retries for AWS CLI commands to use somewhere else? Do we want to change the retries to 2 for all the AWS CLI commands? Or do we need to unset AWS_MAX_ATTEMPTS after running `aws s3api` command?

cli/src/pcluster/resources/compute_node/user_data.sh
davprat: I believe the default for v1 is `legacy`: https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-retries.html

PR #4800, opened 2023-01-17
[Isolated Regions] Add support for isolated regions: us-iso-* and us-isob-*

gmarciani:
### Description of changes
Added support for isolated regions: us-iso-* and us-isob-*

### Tests
* Unit tests
* Manual test creating a cluster
* Full regression tests will be done by the authoritative pipeline


dreambeyondorange (dismissed):

cli/src/pcluster/utils.py
```diff
@@ -36,18 +36,48 @@
 
 LOGGER = logging.getLogger(__name__)
 
+DEFAULT_PARTITION = "aws"
+PARTITION_MAP = {
+    "cn": "aws-cn",
```
lukeseawalker: why not `cn-?`

cli/src/pcluster/utils.py
```diff
@@ -36,18 +36,48 @@
...
+    "us-gov": "aws-us-gov",
```
lukeseawalker: why not`us-gov-`?

cli/src/pcluster/utils.py
```diff
@@ -36,18 +36,48 @@
...
+    "us-iso-": "aws-iso",
+    "us-isob": "aws-iso-b",
```
lukeseawalker: why not `us-isob-`?

cli/src/pcluster/resources/batch/docker/scripts/modify_yum_vars.sh
```diff
@@ -41,14 +41,22 @@ main() {
...
+    local REGION="${1}"
+    local AWS_DOMAIN="amazonaws.com"
+    if [[ ${REGION} == cn-* ]]; then
+        AWS_DOMAIN="amazonaws.com.cn"
+    elif [[ ${REGION} == us-iso-* ]]; then
+        AWS_DOMAIN="c2s.ic.gov"
+    elif [[ ${REGION} == us-isob-* ]]; then
+        AWS_DOMAIN="sc2s.sgov.gov"
+    fi
+    if [[ ${REGION} == cn-* ]] || [[ ${REGION} == us-iso-* ]] || [[ ${REGION} == us-isob-* ]]; then
```
lukeseawalker: instead of duplicating the comparison logic here, why not just check if AWS_DOMAIN is different from the default then enter the if

cli/src/pcluster/utils.py
gmarciani: No specific reason. I agree to use `cn-` instead of `cn`.

cli/src/pcluster/utils.py
gmarciani: No specific reason. I agree to use `us-gov-` instead of `us-gov`.

cli/src/pcluster/utils.py
gmarciani: No specific reason. I agree to use `us-isob-` instead of `us-isob`.

cli/src/pcluster/resources/batch/docker/scripts/modify_yum_vars.sh
gmarciani: Good point, doing

lukeseawalker (approved):
LGTM

PR #4802, opened 2023-01-17
[3.x] Create THIRD-PARTY-LICENSES.txt file using create-attribution-doc.sh

himani2411:
### Description of changes
* Add a Bash script to create THIRD_PARTY_LICENSES.txt 
* Added Python License depending on the version where you run this script

### Tests
* Local tests to create a THIRD_PARTY_LICENSES.txt 

### Checklist

### DIFF
--- a/util/create-attribution-doc.sh
+++ b/util/create-attribution-doc.sh
@@ -1,12 +1,15 @@
 #!/bin/bash
 
 set -ex
+# Install the python version if it doesnt exist
+if test ! -d ${PYENV_ROOT}/versions/3.9.10;
+then 
+  env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.9.10
+fi
 
-pyenv virtualenv attribution-doc-env
+pyenv virtualenv 3.9.10 attribution-doc-env
 # switch to a specific virtual env
 source ${PYENV_ROOT}/versions/attribution-doc-env/bin/activate
-
-
 
 # Update Pip
 pip3 install --upgrade pip
@@ -22,10 +25,10 @@ pip3 install -e cli
 #pip3 install -r requirements.txt
 
 final_license_file=THIRD-PARTY-LICENSES.txt
-python_license_file=python_license.txt
-aws_cdk_license_file=aws-cdk-license.txt
+
 # Create a list of aws cdk Sub-packages we want to ignore 
 aws_cdk_ignore_subpackages=$(pip list | grep cdk | awk '{print $1}') 
+aws_cdk_version=$(pip list | grep cdk | awk '{print $2}'| head -n 1)
 # Create a pip License document
 pip-licenses -i $aws_cdk_ignore_subpackages aws-parallelcluster pip-licenses --format=plain-vertical --with-license-file --with-urls --no-license-path --with-authors --output-file=$final_license_file
 
@@ -34,13 +37,15 @@ pip-licenses -i $aws_cdk_ignore_subpackages aws-parallelcluster pip-licenses --f
 cpy_version=$(python -V |  grep -Eo '([0-9]+)(\.?[0-9]+)' | head -1) 
 
 
-# Add other License in the final file
-# Get Aws-cdk License file
-echo "aws-cdk; https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE" >> $final_license_file
+# Appending aws-cdk and Python License to the THIRD-PARTY-LICENSES file
+# Adding a header for aws-cdk version and License URL to THIRD-PARTY-LICENSES file
+echo "aws-cdk $aws_cdk_version; https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE" >> $final_license_file
+# Appending aws-cdk License to THIRD-PARTY-LICENSES file
 curl https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE >> $final_license_file
-# Download Python License file
-echo -e "\nPython $cpy_version " >> $final_license_file
-curl $python_license_file https://raw.githubusercontent.com/python/cpython/$cpy_version/LICENSE >> $final_license_file
+# Adding a header for Python version and License URL to THIRD-PARTY-LICENSES file
+echo -e "\nPython $cpy_version; https://raw.githubusercontent.com/python/cpython/$cpy_version/LICENSE " >> $final_license_file
+# Appending Python License to THIRD-PARTY-LICENSES file
+curl https://raw.githubusercontent.com/python/cpython/$cpy_version/LICENSE >> $final_license_file
 
 
 deactivate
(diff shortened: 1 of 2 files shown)


util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -ex
+
+pyenv virtualenv attribution-doc-env
```
dreambeyondorange: should we specify the python version here?  This may not be python 3.

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+# switch to a specific virtual env
+source ${PYENV_ROOT}/versions/attribution-doc-env/bin/activate
+
+
+
+# Update Pip
+pip3 install --upgrade pip
```
dreambeyondorange, line 15: If we are using pyenv, shouldn't this be just `pip`?

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+# install pcluster via source
+cd ..
+pip3 install -e cli
+#pip3 install -r requirements.txt
+
+final_license_file=THIRD-PARTY-LICENSES.txt
+python_license_file=python_license.txt
+aws_cdk_license_file=aws-cdk-license.txt
+# Create a list of aws cdk Sub-packages we want to ignore 
+aws_cdk_ignore_subpackages=$(pip list | grep cdk | awk '{print $1}')
```
dreambeyondorange, line 30: Can we safely assume all the packages use the same license?

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+pip3 install -e cli
+#pip3 install -r requirements.txt
+
+final_license_file=THIRD-PARTY-LICENSES.txt
+python_license_file=python_license.txt
+aws_cdk_license_file=aws-cdk-license.txt
+# Create a list of aws cdk Sub-packages we want to ignore 
+aws_cdk_ignore_subpackages=$(pip list | grep cdk | awk '{print $1}') 
+# Create a pip License document
+pip-licenses -i $aws_cdk_ignore_subpackages aws-parallelcluster pip-licenses --format=plain-vertical --with-license-file --with-urls --no-license-path --with-authors --output-file=$final_license_file
```
dreambeyondorange, line 33: Does this also capture transitive dependencies?  I.e. the dependencies of our dependencies?

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+
+
+#Getting python version
+cpy_version=$(python -V |  grep -Eo '([0-9]+)(\.?[0-9]+)' | head -1) 
+
+
+# Add other License in the final file
```
dreambeyondorange: I would suggest clarifying this comment a bit more.  What exactly are we doing?

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+# Get Aws-cdk License file
+echo "aws-cdk; https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE" >> $final_license_file
```
dreambeyondorange: can you get this from the pyenv instead of fetching online?

util/create-attribution-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+curl https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE >> $final_license_file
+# Download Python License file
```
dreambeyondorange: Is this also available locally?

dreambeyondorange (commented):
Nice work!

util/create-attribution-doc.sh
himani2411: Initially i did keep the python version as 3.9.10. But, when I run this script from another virtual environment it takes the Python version used in that environment. We can keep a comment to run this in the latest python version.  Or I could keep this at the latest python 3 version.

util/create-attribution-doc.sh
himani2411: Yes, all f the sub-packages use the same version and the same license which is why in order to remove the duplication, I have added the aws-cdk license once, which is what we do in the older version of 3rd party license too

util/create-attribution-doc.sh
himani2411: yes, it does. For example, https://github.com/raimon49/pip-licenses/issues/41

util/create-attribution-doc.sh
himani2411: The Python License file location could differ might change for others, I found getting it online much more straitforward.

util/create-attribution-doc.sh
himani2411: I would have to take into account the packages and the versions for finding this. I could find this locally, but isnt it straightforward to get it online. Is there a reason I shouldnt?

util/create-attribution-doc.sh
himani2411: I did add other comments below these, but will elaborate on them on a high level.

util/create-attribution-doc.sh
himani2411: If I keep the latest python version I think we wouldnt have to worry about this

util/create-attribution-doc.sh
dreambeyondorange: I don't think it makes sense to set this up without specifying the version of python.  I'm not sure I understand `when I run this script from another virtual environment it takes the Python version used in that environment`.  Aren't you creating the virtual environment here?

util/create-attribution-doc.sh
dreambeyondorange: I don't think we should be using the pyenv and pip3 at the same time.  Either we set up the virtual env and use the package manager for it, or we require 3.x and use pip3

util/create-attribution-doc.sh
dreambeyondorange: how do you guarantee that what you are fetching online is the same version you have locally?

util/create-attribution-doc.sh
dreambeyondorange: similarly, how do you know it matches what you have?

util/create-attribution-doc.sh
himani2411: I have specified the python version to 3.9.10

util/create-attribution-doc.sh
himani2411: I have specified the python version to 3.9.10. So should resolve the use of pip3

util/create-attribution-doc.sh
himani2411: I am taking the Python version we have on pyenv and using that to get the LICENSE file directly

util/create-attribution-doc.sh
himani2411: I took the URL from the pip-license command which is https://github.com/aws/aws-cdk. But for concatenating I needed the raw content.

dreambeyondorange (approved):

PR #4812, opened 2023-01-20
[Isolated Regions] Add validators to check features support by region

gmarciani:
### Description of changes
Add validators to check features support by region. In particular, the following features are set as unsupported:
* AWS Batch in ap-northeast-3 and us-iso* regions
* NICE DCV in us-iso* regions
* FSx Lustre in us-iso* regions
* FSx ONTAP in us-iso* regions
* FSx OpenZfs in us-iso* regions

#### Customer Experience
When a customer attempts to create a cluster with unsupported features the following error message is returned for every unsupported feature attempted (example with FSxLustre pretending it's not supported in eu-west-1):

```
{
    {
      "level": "ERROR",
      "type": "FeatureRegionValidator",
      "message": "FSx Lustre is not supported in region 'eu-west-1'."
    }
  ],
}
```

#### Developer Experience
When a developer wants to modify the support for a given feature in some regions, he/she needs to modify `UNSUPPORTED_FEATURES_MAP` in `cli/src/pcluster/constants.py` accordingly, where we have region prefixes mapped to unsupported features.

```
UNSUPPORTED_FEATURES_MAP = {
    Feature.BATCH: ["ap-northeast-3", "us-iso"],
    Feature.DCV: ["us-iso"],
    Feature.FSX_LUSTRE: ["us-iso"],
    Feature.FSX_ONTAP: ["us-iso"],
    Feature.FSX_OPENZFS: ["us-iso"],
}
```

and register the corresponding validator `FeatureRegionValidator`, if it's not registered yet:

```
self._register_validator(FeatureRegionValidator, feature=Feature.FSX_LUSTRE, region=self.region)
```

### Tests
* Unit test
* Manual test pretending that some features where not supported in eu-west-1 to verify error messages.


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1368,12 +1372,18 @@ def _register_storage_validators(self):
                         existing_fsx.add(storage.file_system_id)
                     else:
                         new_storage_count["fsx"] += 1
+                    self._register_validator(FsxLustreRegionValidator, region=self.region)
```
enrico-usai: what about generalizing the validator: `FeatureRegionValidator` and pass 2 parameters, the feature and the region, so: 
```python
self._register_validator(FeatureRegionValidator, feature=Feature.FSX_LUSTRE, region=self.region)
```

This will permit to add more feature validators or to remove them without the need to create new code.

cli/src/pcluster/constants.py
```diff
@@ -206,3 +207,36 @@
...
+    """
+
+    BATCH = "AWS Batch"
+    DCV = "Nice DCV"
+    FSX_LUSTRE = "FSx Lustre"
+    FSX_ONTAP = "FSx ONTAP"
+    FSX_OPENZFS = "FSx OpenZfs"
+
+
+UNSUPPORTED_FEATURES_MAP = {
```
enrico-usai, line 234: what about having this map by feature, rather than by region?

```python
UNSUPPORTED_FEATURES_MAP = {
  Feature.BATCH: ["ap-northeast-3", "us-iso"]
  Feature.DCV: [...]
```

cli/src/pcluster/utils.py
```diff
@@ -32,11 +32,29 @@
 from yaml.resolver import BaseResolver
 
 from pcluster.aws.common import get_region
-from pcluster.constants import SUPPORTED_OSES_FOR_ARCHITECTURE, SUPPORTED_OSES_FOR_SCHEDULER
+from pcluster.constants import SUPPORTED_OSES_FOR_ARCHITECTURE, SUPPORTED_OSES_FOR_SCHEDULER, UNSUPPORTED_FEATURES_MAP
 
 LOGGER = logging.getLogger(__name__)
 
 
+def is_feature_supported(feature: str, region: str = None):
```
enrico-usai: If we create the generic `FeatureRegionValidator` this code can be moved there, removing this utils function.

cli/src/pcluster/config/cluster_config.py
enrico-usai: This corresponds to what you're doing with the utility function.
Then this permits to disable the region validators at once, by disabling the FeatureRegionValidator.

cli/src/pcluster/constants.py
enrico-usai: This corresponds that the validator is for the feature. We are registering the validator when the feature is enabled in the config, so the feature is our starting point to see the limitations.

cli/src/pcluster/utils.py
gmarciani: In we agree on having the validator that way, I'll do that. Let's refer to the discussion in https://github.com/aws/aws-parallelcluster/pull/4812#discussion_r1082311502

cli/src/pcluster/constants.py
gmarciani: I like it, will do

cli/src/pcluster/config/cluster_config.py
gmarciani: I like it and I was thinking to do that since the beginning to reduce code duplication.
The only reason why I did not do that is consistency with the way we developed other validators: we have always preferred to decouple them and since validator names are visible to customers we have always preferred to return specific names to do ad-hoc overrides.

Nevertheless, I'm re.thinking about it thanks to your comment and I think that there is no reason to do an ad hoc override of this type of validators, so we can do the refactoring you suggest.

cli/src/pcluster/utils.py
enrico-usai: Let's do it!

cli/src/pcluster/config/cluster_config.py
enrico-usai: I'd say we can start simple, with a single class, then in future we can decide if we want to split into multiple validators.

enrico-usai (dismissed):

lukeseawalker (approved):
LGTM

PR #4818, opened 2023-01-23
[Isolated Regions] Add controllers assertion to check operations support by region

gmarciani:
### Description of changes
Add assertion on operations controllers to fail fast with a BadRequest when the requested operation is not supported in the given region. In particular, the following operations are set as unsupported both on CLI and API
 * `build-image` us-iso* regions
 * `delete-image` us-iso* regions
 * `describe-image` us-iso* regions
 * `list-images` us-iso* regions
 * `export-image-logs` us-iso* regions
 * `get-image-log-events` us-iso* regions
 * `get-image-stack-events` us-iso* regions
 * `list-image-log-streams` us-iso* regions

#### Customer Experience
When a customer attempts to execute an operation that is unsupported in the target region, the following error message is returned:

```
{
  "message": "Bad Request: The operation 'build-image' is not supported in region 'us-iso-east-1'."
}
```

#### Developer Experience
When a developer wants to modify the support for a given operation in some regions, he/she needs to modify `UNSUPPORTED_FEATURES_MAP` in `cli/src/pcluster/constants.py` accordingly, where we have region prefixes mapped to unsupported features.

```
UNSUPPORTED_OPERATIONS_MAP = {
    Operation.BUILD_IMAGE: ["us-iso"],
    Operation.DELETE_IMAGE: ["us-iso"],
    Operation.DESCRIBE_IMAGE: ["us-iso"],
    Operation.LIST_IMAGES: ["us-iso"],
    Operation.EXPORT_IMAGE_LOGS: ["us-iso"],
    Operation.GET_IMAGE_LOG_EVENTS: ["us-iso"],
    Operation.GET_IMAGE_STACK_EVENTS: ["us-iso"],
    Operation.LIST_IMAGE_LOG_STREAMS: ["us-iso"],
}
```

and add the corresponding assertion within the operation controller, if it's not there yet:

```
assert_supported_operation(operation=Operation.BUILD_IMAGE, region=region)
```

### Tests
* Unit test
* Manual test pretending that an operation is not supported in eu-west-1 to verify error messages.

### Description of changes

### Tests

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/api/controllers/common.py
+++ b/cli/src/pcluster/api/controllers/common.py
@@ -27,9 +27,9 @@ from pcluster.api.errors import (
     NotFoundException,
     ParallelClusterApiException,
 )
-from pcluster.aws.common import BadRequestError, LimitExceededError, StackNotFoundError
+from pcluster.aws.common import BadRequestError, LimitExceededError, StackNotFoundError, get_region
 from pcluster.config.common import AllValidatorsSuppressor, TypeMatchValidatorsSuppressor, ValidatorSuppressor
-from pcluster.constants import SUPPORTED_REGIONS
+from pcluster.constants import SUPPORTED_REGIONS, UNSUPPORTED_OPERATIONS_MAP, Operation
 from pcluster.models.cluster import Cluster
 from pcluster.models.common import BadRequest, Conflict, LimitExceeded, NotFound, parse_config
 from pcluster.utils import get_installed_version, to_utc_datetime
@@ -177,3 +177,13 @@ def get_validator_suppressors(suppress_validators: Optional[List[str]]) -> Set[V
         validator_suppressors.add(TypeMatchValidatorsSuppressor(validator_types_to_suppress))
 
     return validator_suppressors
+
+
+def assert_supported_operation(operation: Operation, region: str):
+    """Raise a BadRequestException if the operation is not supported in the region."""
+    _region = get_region() if region is None else region
+    prefixes_of_unsupported_regions = UNSUPPORTED_OPERATIONS_MAP.get(operation, [])
+    if any(_region.startswith(region_prefix) for region_prefix in prefixes_of_unsupported_regions):
+        message = f"The operation '{operation.value}' is not supported in region '{region}'."
+        LOGGER.critical(message)
+        raise BadRequestException(message)
--- a/cli/src/pcluster/cli/entrypoint.py
+++ b/cli/src/pcluster/cli/entrypoint.py
@@ -215,7 +215,15 @@ def _run_operation(model, args, extra_args):
             error_encoded = encoder.JSONEncoder().encode(message)
             raise APIOperationException(json.loads(error_encoded))
     else:
-        return args.func(args, extra_args)
+        try:
+            return args.func(args, extra_args)
+        except pcluster.api.errors.ParallelClusterApiException as e:
+            # Format exception messages in the same manner as the api
+            message = pcluster.api.errors.exception_message(e)
+            error_encoded = encoder.JSONEncoder().encode(message)
+            raise APIOperationException(json.loads(error_encoded))
+        except Exception as e:
+            raise e
 
 
 def run(sys_args, model=None):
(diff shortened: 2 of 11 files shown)


cli/src/pcluster/api/controllers/common.py
```diff
@@ -177,3 +178,12 @@ def get_validator_suppressors(suppress_validators: Optional[List[str]]) -> Set[V
         validator_suppressors.add(TypeMatchValidatorsSuppressor(validator_types_to_suppress))
 
     return validator_suppressors
+
+
+def assert_supported_operation(operation: Operation, region: str):
+    """Raise a BadRequestException if the operation is not supported in the region."""
+    validation_failures = OperationRegionValidator().execute(operation=operation, region=region)
```
enrico-usai: I think what you implemented is not a Validator.

The Validators are semantic validators that checks configuration parameters (and that can be "suppressed", if needed). They are automatically called by internal methods, when checking the cluster config.

Here we're using Validators logic to do instead a command check. We're forcing the execution of the `execute` method and then we're parsing the output to retrieve the `message` of the first item in the validation failures.

I think in this case it's more clear to have the `assert_supported_operation` that simply executes the content of the `_is_operation_supported` and then raises a BadRequestException, without putting the `Validator` class in the middle.

cli/src/pcluster/constants.py
```diff
@@ -206,3 +207,37 @@
...
+    """
+
+    BUILD_IMAGE = "build-image"
+    DELETE_IMAGE = "delete-image"
+    DESCRIBE_IMAGE = "describe-image"
+    LIST_IMAGES = "list-images"
+    EXPORT_IMAGE_LOGS = "export-image-logs"
+    GET_IMAGE_LOG_EVENTS = "get-image-log-events"
+    GET_IMAGE_STACK_EVENTS = "get-image-stack-events"
+    LIST_IMAGE_LOG_STREAMS = "list-image-log-streams"
```
enrico-usai, line 275: Should we add all the operations here?
Is there a way to check if we're missing some operations from the API? (e.g. with a unit test?)

gmarciani: > Should we add all the operations here?

We can for completeness, but it's not strictly required. 
An entry here is required only if we are going to set it as unsupported in some regions.
I agree on adding all of theme for completeness.

> Is there a way to check if we're missing some operations from the API? (e.g. with a unit test?)

Not now. To do so we must:
1. add all the operations into the enumeration, and
2. extend the current unit test to cover them;

I will do.

cli/src/pcluster/api/controllers/common.py
gmarciani: I agree, thanks

enrico-usai (dismissed):

cli/src/pcluster/cli/entrypoint.py
```diff
@@ -215,7 +215,13 @@ def _run_operation(model, args, extra_args):
...
             raise APIOperationException(json.loads(error_encoded))
     else:
-        return args.func(args, extra_args)
+        try:
+            return args.func(args, extra_args)
+        except Exception as e:
+            # format exception messages in the same manner as the api
+            message = pcluster.api.errors.exception_message(e)
+            error_encoded = encoder.JSONEncoder().encode(message)
+            raise APIOperationException(json.loads(error_encoded))
```
gmarciani: This change may be considered too invasive because it moves the failure output of the commands not included in API model from plaintext to the JSON style provided by the API. 

Currently: the following commands are not defined in the API model, so they would fall into this except block in case of an Exception:

1. configure
1. dcv-connect
1. export-cluster-logs
1. export-image-logs
1. ssh
1. version

gmarciani: I decided the reduce the impact of the change by formatting the output as JSON only on the specific exception  `pcluster.api.errors.ParallelClusterApiException`. I think this is a good tradeoff because:
1. it allows commands excluded from API model to return exceptions in the API JSON style.
2. when a dev uses `ParallelClusterApiException` within a command outside of the API model it's self-explanatory from the name that the exception is expected to be outputted with an API style.

That said, I think that in future iterations we must align all the CLI commands to the API style, i.e. return eveything as a JSON.

lukeseawalker (dismissed):
LGTM

lukeseawalker (approved):

PR #4822, opened 2023-01-25
[3.x]Update THIRD-PARTY-LICENSES.txt

himani2411:
### Description of changes
Making changes with respect to Opens source Distribution for adding specified URL's and License files.

### Tests
* No tests. Running Locally

### Checklist


util/create-attribution-doc.sh
```diff
@@ -29,23 +29,80 @@ final_license_file=THIRD-PARTY-LICENSES.txt
...
-# Adding a header for aws-cdk version and License URL to THIRD-PARTY-LICENSES file
-echo "aws-cdk $aws_cdk_version; https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE" >> $final_license_file
-# Appending aws-cdk License to THIRD-PARTY-LICENSES file
-curl https://raw.githubusercontent.com/aws/aws-cdk/main/LICENSE >> $final_license_file
-# Adding a header for Python version and License URL to THIRD-PARTY-LICENSES file
-echo -e "\nPython $cpy_version; https://raw.githubusercontent.com/python/cpython/$cpy_version/LICENSE " >> $final_license_file
-# Appending Python License to THIRD-PARTY-LICENSES file
-curl https://raw.githubusercontent.com/python/cpython/$cpy_version/LICENSE >> $final_license_file
+# Function to Append Licenses of different Packages to the THIRD-PARTY-LICENSES file
+append_to_final_license_file(){
```
dreambeyondorange: so is this function not actually adding the content of the license file, just the URL?  Don't we need to include the full license text?

dreambeyondorange: I think we should be specific about this in the name about what exactly we are appending

util/create-attribution-doc.sh
```diff
@@ -29,23 +29,80 @@ final_license_file=THIRD-PARTY-LICENSES.txt
...
+# Packaging
+append_to_final_license_file "packaging" $packaging_version "pache Software License; BSD License" "https://github.com/pypa/packaging" "https://raw.githubusercontent.com/pypa/packaging/main/LICENSE" "https://raw.githubusercontent.com/pypa/packaging/main/LICENSE.APACHE" "https://raw.githubusercontent.com/pypa/packaging/main/LICENSE.BSD"
+
+# pyinstaller-hooks-contrib
+# Adding a header for pyinstaller-hooks-contrib version and License URL 
+echo -e "\n\n\npyinstaller-hooks-contrib \n2022.15\nApache License 2.0; GNU General Public License \nhttps://github.com/pyinstaller/pyinstaller-hooks-contrib/archive/refs/tags/2022.15.tar.gz." >> $final_license_file
+# Appending pyinstaller-hooks-contrib 3 License files 
+wget -c https://github.com/pyinstaller/pyinstaller-hooks-contrib/archive/refs/tags/2022.15.tar.gz -O - | tar -xz
+
+cat pyinstaller-hooks-contrib-2022.15/LICENSE >> $final_license_file
```
dreambeyondorange: why only do this for the hooks package and not the others we are ignoring with pip-licenses?

util/create-attribution-doc.sh
```diff
@@ -29,23 +29,80 @@ final_license_file=THIRD-PARTY-LICENSES.txt
...
+cat pyinstaller-hooks-contrib-2022.15/LICENSE.APL.txt >> $final_license_file
+cat pyinstaller-hooks-contrib-2022.15/LICENSE.GPL.txt >> $final_license_file
+
+rm -rf pyinstaller-hooks-contrib-2022.15
+
+# pyinstaller
+# Adding a header for pyinstaller version and License URL 
+echo -e "\n\n\npyinstaller \n5.7.0\nGNU General Public License v2 (GPLv2) \nhttps://github.com/pyinstaller/pyinstaller/archive/refs/tags/v5.7.0.tar.gz." >> $final_license_file
```
dreambeyondorange: should we use the function for this?

util/create-attribution-doc.sh
```diff
@@ -29,23 +29,80 @@ final_license_file=THIRD-PARTY-LICENSES.txt
...
+wget -c https://github.com/pyinstaller/pyinstaller/archive/refs/tags/v5.7.0.tar.gz -O - | tar -xz
+# Appending pyinstaller License file
+cat pyinstaller-5.7.0/COPYING.txt >> $final_license_file
```
dreambeyondorange: is this file not available in the other archive above?

util/create-attribution-doc.sh
himani2411: Its appending a header which is Package Name, Package Version, License Type , URL for package to the final License file as well as the whole License file. I have included comments inside the function

util/create-attribution-doc.sh
himani2411: The distribution team provided this specific link for the tar file, which is why I need to unzip it and add to the final file.  For others its a direct link which I can use.

util/create-attribution-doc.sh
himani2411: Sure

util/create-attribution-doc.sh
himani2411: I could directly use this link https://github.com/pyinstaller/pyinstaller-hooks-contrib/tree/2022.15, But I thought we had already decided on going with the approach of using the tar file and unzipping it.

util/create-attribution-doc.sh
himani2411: No I couldnt find this ( COPYING.txt) for other packages.

util/create-attribution-doc.sh
himani2411: I have changed the link and directly taking from online using  https://github.com/pyinstaller/pyinstaller-hooks-contrib/tree/2022.15, so that I can reuse the same function

dreambeyondorange: What I mean is it seems like we are just putting a link for the license instead of the actual license content for these packages, whereas with the hooks package we put the license content into the file.  Is that correct?

himani2411: No we are putting the license content for all the packages into the final file as well as the License link

dreambeyondorange (approved):

PR #4823, opened 2023-01-25
Change get of node attributes to use /etc/chef/node_attributes.json

francesco-giordano:
### Description of changes
* To make more robust the getting of attributes from the cookbook this commit exploit the newly created  attribute file `/etc/chef/node_attributes.json`

### Tests
* Tested in Jenkins

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1679

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster_test.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster_test.yaml
@@ -5,7 +5,7 @@ schemaVersion: 1.0
 constants:
   - CookbookDefaultFile:
       type: string
-      value: /etc/chef/cookbooks/aws-parallelcluster/attributes/default.rb
+      value: /etc/chef/node_attributes.json
 
 phases:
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster_test.yaml
```diff
@@ -55,7 +55,7 @@ phases:
           commands:
             - |
               set -v
-              PATTERN=$(grep -F "default['cluster']['python-version']" {{ CookbookDefaultFile }})
+              PATTERN=$(jq '.default.cluster."python-version"' {{ CookbookDefaultFile }})
```
awslitvit, line 58: Nitpick. Looks like we should change from `python-version` to `python_version`. By the way, how do we understand which config values are externally facing? Honestly, I would create (in a future PR) a json with only those values, so that it is clear what is exposed.

awslitvit (approved):

PR #4838, opened 2023-01-30
Add integration test for interaction of scontrol reboot with EC2 status checks

jdeamicis:
### Description of changes
* Add an integration test to check the logic to allow compute nodes rebooted via Slurm not to be drained due to EC2 status checks failures (not yet added to the daily integ tests).

### Tests
* The test covers the two ways a node rebooted via `scontrol reboot` may be treated as unhealthy by `clustermgtd` if EC2 status checks fail during the reboot process:
  1. the EC2 status checks report failure while the node is still rebooting (in `REBOOT_ISSUED` state);
  2. the EC2 status checks report failure after the node has already restarted Slurmd, due to the time it takes EC2 to publish the EC2 status check state.

### References
* https://github.com/aws/aws-parallelcluster-node/pull/490

### Checklist
- [ ] Check all commits' messages are clear, describing what and why vs how

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -559,7 +559,6 @@ def test_scontrol_reboot_ec2_health_checks(
 
     # 2 iterations to cover the two scenarios described in the test description
     for scenario in range(1, 3):
-
         if scenario == 2:
             # Remove delay in startup of slurmd on the compute nodes.
             # TODO: generalize it for multiple nodes (it requires a parallel ssh utility such as clush)
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -33,6 +33,7 @@ from tests.common.assertions import (
     assert_num_instances_in_cluster,
     assert_scaling_worked,
     wait_for_num_instances_in_cluster,
+    wait_for_slurm_rebooted_nodes,
 )
 from tests.common.hit_common import (
     assert_compute_node_reasons,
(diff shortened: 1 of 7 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -487,6 +488,112 @@ def test_scontrol_reboot(
...
+            remote_command_executor.run_remote_command(f'ssh {str(compute_nodes[0])} "./{script}"')
+
+        # Reboot static compute nodes
+        for node in compute_nodes:
+            slurm_commands.reboot_compute_node(node, asap=True)
+
+        # Wait for compute nodes to reply to slurmctld after rebooting
+        wait_for_slurm_rebooted_nodes(compute_nodes, remote_command_executor, stop_max_delay_secs=600)
+
+        # Check that the nodes are not marked as unhealthy in the following minutes
```
chenwany, line 580: Can we add a check to verify the node fail health check?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -487,6 +488,112 @@ def test_scontrol_reboot(
...
+        time.sleep(300)
+        assert_no_msg_in_logs(
+            remote_command_executor,
+            ["/var/log/parallelcluster/clustermgtd"],
+            ["Setting nodes failing health check type ec2_health_check to DRAIN"],
```
chenwany, line 585: What about add a check to check if this is in log: These nodes were recently restarted and they were not set to DRAIN

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: I agree with you, but here there's an annoying issue: it's not guaranteed that the EC2 checks will fail, because that depends on the time it takes for the reboot to occur. Sometimes a reboot is very quick and the EC2 status checks won't fail, and some other times the reboot might take longer, which causes the EC2 checks to fail.

Adding an assertion here may cause the test to fail.

In the end I decided to just verify at the end of the iteration that the node had not been drained during the reboot process.

tests/integration-tests/tests/schedulers/test_slurm.py
jdeamicis: See above.

demartinofra (dismissed):

demartinofra (dismissed):

gmarciani (approved):

PR #4845, opened 2023-02-01
`[API]` Update to use hosted policies, lambda layer, new integ-tests

charlesg3:
### Description of changes
* In latest version of ParallelCluster the policies have been factored out to live on their own, this PR updates the API stack to leverage these policies as a nested sub-stack.
- Update integration tests to match new artifacts

### Tests
- 

### Checklist


api/docker/awslambda/Dockerfile
demartinofra, line 1: Shall we keep publishing the pcluster docker image. It might still be useful to some use cases (e.g. run CLI in a docker container or deploy API outside of Lambda) and the entire CI/CD is already pretty much automated?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -259,23 +262,61 @@ Resources:
...
-              - [!Select [2, !Split ['/', !Ref EcrImage]], !Select [3, !Split ['/', !Ref EcrImage]]]
+        - Key: 'parallelcluster:resource'
+          Value: api
+        - Key: 'parallelcluster:version'
+          Value: !FindInMap [ParallelCluster, Constants, Version]
+      Runtime: python3.9
+      Handler: index.handler
+      Layers:
+        - !Ref PclusterLayer
+      Code:
```
demartinofra, line 288: what is the reason why we can't simply use the existing entrypoint rather then inlining it here? is it a limitation of using layers?

api/docker/awslambda/Dockerfile
eantonin: I agree with Francesco, we should leave this part for the moment.
Deleting will mean also fixing various CI/CD parts, and we should not work on this now

api/infrastructure/deploy-api.sh
```diff
@@ -90,53 +66,29 @@ case $key in
...
-
-echo "Building docker image"
-"${SCRIPT_DIR}/../docker/awslambda/docker-build.sh"
-
-echo "Pushing docker image to ${ECR_ENDPOINT}/${ECR_REPO}"
-aws ecr get-login-password | docker login --username AWS --password-stdin "${ECR_ENDPOINT}"
-docker tag pcluster-lambda:latest "${ECR_ENDPOINT}/${ECR_REPO}:latest"
-docker push "${ECR_ENDPOINT}/${ECR_REPO}:latest"
+POLICIES_S3_URI="s3://${S3_BUCKET}/stacks/parallelcluster-policies.yaml"
+POLICIES_TEMPLATE_URI="http://${S3_BUCKET}.s3.${AWS_DEFAULT_REGION}.amazonaws.com/stacks/parallelcluster-policies.yaml"
```
eantonin, line 77: question, non-blocking, I do not get why do you need both lines here

api/infrastructure/deploy-api.sh
```diff
@@ -90,53 +66,29 @@ case $key in
...
     --s3-bucket "${S3_BUCKET}" \
     --s3-prefix "api/" \
-    --parameter-overrides ApiDefinitionS3Uri="${S3_UPLOAD_URI}" "${OVERRIDE_IMAGE_URI_PRM}"="${ECR_ENDPOINT}/${ECR_REPO}:latest" \
+    --parameter-overrides ApiDefinitionS3Uri="${S3_UPLOAD_URI}" \
+                          PoliciesTemplateUri="${POLICIES_TEMPLATE_URI}" \
                           EnableIamAdminAccess="${ENABLE_IAM_ADMIN}" CreateApiUserRole="${CREATE_API_USER}" \
-                          ImageBuilderVpcId="${IMAGE_BUILDER_VPC_ID}" \
-                          ImageBuilderSubnetId="${IMAGE_BUILDER_SUBNET_ID}" \
     --capabilities CAPABILITY_NAMED_IAM
-
```
eantonin, line 134: question, non-blocking, should we add the layer to the lambda too?

tests/integration-tests/README.md
```diff
@@ -42,7 +42,7 @@ usage: test_runner.py [-h] --key-name KEY_NAME --key-path KEY_PATH [-n PARALLELI
                       [--custom-cookbook-url CUSTOM_COOKBOOK_URL] [--createami-custom-cookbook-url CREATEAMI_CUSTOM_COOKBOOK_URL] [--createami-custom-node-url CREATEAMI_CUSTOM_NODE_URL] [--custom-awsbatchcli-url CUSTOM_AWSBATCHCLI_URL]
                       [--pre-install PRE_INSTALL] [--post-install POST_INSTALL] [--instance-types-data INSTANCE_TYPES_DATA] [--custom-ami CUSTOM_AMI] [--pcluster-git-ref PCLUSTER_GIT_REF] [--cookbook-git-ref COOKBOOK_GIT_REF]
                       [--node-git-ref NODE_GIT_REF] [--ami-owner AMI_OWNER] [--benchmarks] [--benchmarks-target-capacity BENCHMARKS_TARGET_CAPACITY] [--benchmarks-max-time BENCHMARKS_MAX_TIME]
-                      [--api-definition-s3-uri API_DEFINITION_S3_URI] [--api-infrastructure-s3-uri API_INFRASTRUCTURE_S3_URI] [--public-ecr-image-uri PUBLIC_ECR_IMAGE_URI] [--api-uri API_URI] [--vpc-stack VPC_STACK] [--cluster CLUSTER]
```
eantonin, line 45: question, should we add `--policies-uri` here?

tests/integration-tests/test_runner.py
```diff
@@ -331,14 +331,7 @@ def _init_argparser():
...
     api_group.add_argument(
-        "--public-ecr-image-uri",
-        help="S3 URI of the ParallelCluster API spec",
-        default=TEST_DEFAULTS.get("public_ecr_image_uri"),
-    )
-    api_group.add_argument(
-        "--api-uri",
-        help="URI of an existing ParallelCluster API",
-        default=TEST_DEFAULTS.get("api_uri"),
+        "--api-uri", help="URI of an existing ParallelCluster API", default=TEST_DEFAULTS.get("api_uri")
```
eantonin, line 334: question, same as above, why don't we need the policy parameter?

tests/integration-tests/tests/pcluster_api/test_api_infrastructure.py
```diff
@@ -106,7 +101,6 @@ def _assert_parallelcluster_lambda(lambda_name, lambda_arn, lambda_image_uri):
     assert_that(lambda_configuration["MemorySize"]).is_equal_to(2048)
     assert_that(lambda_resource["Tags"]).contains("parallelcluster:version")
     assert_that(lambda_resource["Tags"]).contains("parallelcluster:resource")
-    assert_that(lambda_resource["Code"]["ImageUri"]).is_equal_to(lambda_image_uri)
```
eantonin, line 109: question, should we add an assert on the layer?

eantonin (changes requested):

api/infrastructure/deploy-api.sh
charlesg3: The bucket is used for uploading the file in this script while the template URI is in the format needed in the CFN template.

api/infrastructure/deploy-api.sh
charlesg3: Not quite sure what you are referring to here -- do you mean to also add a parameter to specify the custom bucket?

api/infrastructure/deploy-api.sh
eantonin: Just wondering why `S3_UPLOAD_URI ` which should follow the same logic, does not need two different variables

api/infrastructure/deploy-api.sh
eantonin: I just got confused `parallelcluster-api.yaml` itself adds the layer to the lambda

Anyway there is still a problem. From this PR I see `parallelcluster-api.yaml` as deleted

api/infrastructure/deploy-api.sh
charlesg3: Ah -- the openapi yaml file isn't a nested template so we're simply uploading it and then providing that value. For the policies it is so we need to upload it and then specify the template url.

charlesg3: (the ECR_ENDPOINT is similar in that we were able to push to that repo in this script and then reference it in the template as-is because it follows the same format in the push command as where we consume it in the template)

api/infrastructure/deploy-api.sh
eantonin: My fault here, was confused by GH

api/docker/awslambda/test-events/event.json
eantonin, line 1: just a note, I think this may actually be still needed to test events for a Lambda. But I'm ok deleting it cause I don't think we are using it

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -39,10 +34,10 @@ Parameters:
     Type: String
     Default: ''
 
-  PublicEcrImageUri:
-    Description: When specified, the URI of the Docker image for the Lambda of the ParallelCluster API
+  PoliciesTemplateUri:
+    Description: S3 URI of the ParallelCluster Policies Template
     Type: String
-    Default: public.ecr.aws/parallelcluster/pcluster-api:3.6.0
```
eantonin, line 45: improvement: This lets to remove a line from `bump_version.sh`

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -39,10 +34,10 @@ Parameters:
...
+    Default: s3://<REGION>-aws-parallelcluster/parallelcluster/3.5.0/templates/policies/policies.yaml
```
eantonin: This should be `3.6.0` and will be updated with `bump_version.sh`

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -149,39 +139,58 @@ Globals:
...
+    Type: AWS::CloudFormation::Stack
+    Condition: CreateIamResources
+    Properties:
+      TemplateURL: !If
+        - UseCustomPoliciesTemplateUri
+        - !Ref PoliciesTemplateUri
+        - !Sub
+          - "s3://${AWS::Region}-aws-parallelcluster/parallelcluster/${Version}/templates/policies/policies.yaml"
+          - { Version: !FindInMap [ParallelCluster, Constants, Version] }
+      TimeoutInMinutes: 10
```
eantonin, line 171: question, non-blocking, are we really sure that 10 minutes are always enough if we do not have problems?
(same for others below)

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -309,719 +320,6 @@ Resources:
...
-            Sid: CloudWatchLogs
-          - Action:
-              - resource-groups:ListGroupResources
-              - resource-groups:GetGroupConfiguration
-            Resource: '*'
-            Effect: Allow
-            Sid: ResourceGroupRead
-
-
-  ### IMAGE ACTIONS POLICIES
```
eantonin, line 753: question, we should be able to remove all the image policies from the parallelcluster policies file too, right?

eantonin (approved):

api/infrastructure/parallelcluster-api.yaml
charlesg3: There shouldn't be anything outstanding that prevents the policies from being created in time (it should be on the order of 1-2 minutes)

api/infrastructure/parallelcluster-api.yaml
charlesg3: I believe many are also needed for the `build-image` operation. It might be worth a second pass (separately) to make sure they are scoped down as well now that we're not using this pathway for creating the API.

PR #4857, opened 2023-02-07
Improve Existing EFS, FSx Security Groups Validation

hanwen-pcluste:
In addition to existing checks on port and CIDR, this commit adds checks on allowed security groups. This improvement will fix false warning during `pcluster update-cluster`. 

In other words, there are three ways to specify src/dst in a security group rule: CIDR, security groups, prefix list. Before this commit, only CIDR was properly checked. This commit adds proper checks for security groups. This commit does not change the behavior of prefix list .

This commit also refactors the code for clarity and reduces redundant EFS error messages

### Tests
* Unit tests have been created and modified
#### Manual testing:
When creating clusters with bad EFS and FSx security groups, the error messages are correctly displayed and only appear once:
```
{
    {
      "level": "ERROR",
      "type": "EfsIdValidator",
      "message": "There is an existing Mount Target fsmt-xxx in the Availability Zone us-east-1b for EFS fs-xxx, but it does not have a security group that allows inbound and outbound rules to support NFS. Please modify the Mount Target's security group, to allow traffic on port 2049."
    },
    {
      "level": "ERROR",
      "type": "ExistingFsxNetworkingValidator",
      "message": "The current security group settings on file system 'fs-xxx' does not satisfy mounting requirement. The file system must be associated to a security group that allows inbound and outbound TCP traffic through ports [988]. Missing ports: [988]"
    }
  ],
}
```
When updating cluster with EFS and FSx with security groups allowing connections from cluster security groups, the validator no longer shows false error messages
```
{
  "message": "Request would have succeeded, but DryRun flag is set.",
  "changeSet": [
    {
      "parameter": "SharedStorage",
      "requestedValue": {
        "MountDir": "/efs1",
        "Name": "efs1",
        "StorageType": "Efs",
        "EfsSettings": {
          "FileSystemId": "fs-01cf0078e64e83b54"
        }
      },
      "currentValue": "-"
    },
    {
      "parameter": "SharedStorage",
      "requestedValue": {
        "MountDir": "/fsx1",
        "Name": "fsx1",
        "StorageType": "FsxLustre",
        "FsxLustreSettings": {
          "FileSystemId": "fs-01fef606ec594d3b3"
        }
      },
      "currentValue": "-"
    }
  ]
}

```

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -37,6 +37,7 @@ class StackInfo:
         self._params = self._stack_data.get("Parameters", [])
         self.tags = self._stack_data.get("Tags", [])
         self.outputs = self._stack_data.get("Outputs", [])
+        self.__resources = None
 
     @property
     def id(self):
@@ -47,6 +48,15 @@ class StackInfo:
     def name(self):
         """Return the name of the stack."""
         return self._stack_data.get("StackName")
+
+    @property
+    def resources(self):
+        """Return the resources of the stack."""
+        if not self.__resources:
+            from pcluster.aws.aws_api import AWSApi  # noqa: F401 pylint: disable=import-outside-toplevel
+
+            self.__resources = AWSApi.instance().cfn.describe_stack_resources(self.name)
+        return self.__resources
 
     @property
     def status(self):
@@ -89,6 +99,14 @@ class StackInfo:
         """
         param_value = next((par["ParameterValue"] for par in self._params if par["ParameterKey"] == key_name), None)
         return None if param_value is None else param_value.strip()
+
+    def get_resource_physical_id(self, resource_logical_id: str):
+        """Return the resource information."""
+        resource = self.resources.get(resource_logical_id)
+        if resource:
+            return resource["PhysicalResourceId"]
+        else:
+            return None
 
 
 class InstanceInfo:
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -136,7 +136,7 @@ class Ec2Client(Boto3Client):
     @Cache.cached
     def get_subnet_cidr(self, subnet_id):
         """Return cidr block  of the given subnet."""
-        subnets = self._client.describe_subnets(SubnetIds=[subnet_id]).get("Subnets")
+        subnets = self.describe_subnets([subnet_id])
         if subnets:
             return subnets[0].get("CidrBlock")
         raise AWSClientError(function_name="describe_subnets", message=f"Subnet {subnet_id} not found")
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -440,6 +440,9 @@ class Cluster:
             Cluster._load_additional_instance_type_data(cluster_config_dict)
             config = self._load_config(cluster_config_dict)
             config.official_ami = self.__official_ami
+            if context.during_update:
+                config.managed_head_node_security_group = self.stack.get_resource_physical_id("HeadNodeSecurityGroup")
+                config.managed_compute_security_group = self.stack.get_resource_physical_id("ComputeSecurityGroup")
 
             validation_failures = config.validate(validator_suppressors, context)
             if any(f.level.value >= FailureLevel(validation_failure_level).value for f in validation_failures):
@@ -851,7 +854,7 @@ class Cluster:
             validator_suppressors=validator_suppressors,
             validation_failure_level=validation_failure_level,
             config_text=target_source_config,
-            context=ValidatorContext(head_node_instance_id=self.head_node_instance.id),
+            context=ValidatorContext(head_node_instance_id=self.head_node_instance.id, during_update=True),
         )
         changes = self._validate_patch(force, target_config)
 
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -12,7 +12,7 @@ import math
 import re
 from collections import defaultdict
 from enum import Enum
-from ipaddress import ip_network
+from ipaddress import collapse_addresses, ip_network
 from itertools import combinations, product
 from typing import List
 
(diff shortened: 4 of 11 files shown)


cli/src/pcluster/aws/ec2.py
```diff
@@ -136,7 +136,7 @@ def get_subnet_vpc(self, subnet_id):
     @Cache.cached
     def get_subnet_cidr(self, subnet_id):
         """Return cidr block  of the given subnet."""
-        subnets = self._client.describe_subnets(SubnetIds=[subnet_id]).get("Subnets")
+        subnets = self.describe_subnets([subnet_id])
```
hanwen-pcluste, line 139: This is a minor improvement to better share the cache mechanism

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -48,6 +49,15 @@ def name(self):
         """Return the name of the stack."""
         return self._stack_data.get("StackName")
 
+    @property
+    def resources(self):
+        """Return the resources of the stack."""
+        if not self.__resources:
+            from pcluster.aws.aws_api import AWSApi  # noqa: F401 pylint: disable=import-outside-toplevel
```
gmarciani, line 56: why do we need to import locally instead of top level?

cli/src/pcluster/models/cluster.py
```diff
@@ -851,7 +854,7 @@ def validate_update_request(
             validator_suppressors=validator_suppressors,
             validation_failure_level=validation_failure_level,
             config_text=target_source_config,
-            context=ValidatorContext(head_node_instance_id=self.head_node_instance.id),
+            context=ValidatorContext(head_node_instance_id=self.head_node_instance.id, during_update=True),
```
gmarciani, line 857: Minor:
It's fine to add the new parameter `during_update`.
Only consider that it would be enough to have the head node id not None to prove that the context is for an update.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -483,6 +519,23 @@ def _check_sg_rules_for_port(rule, port_to_check, protocol):
     return False
 
 
+def _are_subnets_covered_by_cidrs(ip_ranges, subnets):
```
gmarciani, line 532: This function should be unit tested but it's not.
Furthermore, if you unit test this function, then you can mock its behaviour when unit testing the whole validator

cli/src/pcluster/aws/aws_resources.py
hanwen-pcluste: Importing on top leads to circular reference between Python files

cli/src/pcluster/models/cluster.py
hanwen-pcluste: Either is fine. Having it explicitly might be more readable

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -417,42 +417,78 @@ def _validate(
...
+
+
+def _are_ip_ranges_and_sg_accessible(security_groups_by_nodes, allowed_ip_ranges, allowed_security_groups, subnets):
+    # For all cluster nodes, at least one of the security groups attached need to be in the UserIdGroupPairs.
+    return all(
+        node_security_groups & allowed_security_groups for node_security_groups in security_groups_by_nodes
+    ) or _are_subnets_covered_by_cidrs(allowed_ip_ranges, subnets)
+
+
+def _populate_allowed_src_or_dst(rule, ip_ranges, allowed_security_groups):
```
gmarciani, line 477: a function called `_populate_...` is unexpected to return a boolean value.
Is returning a boolean value required at all?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -760,7 +810,9 @@ def _validate(self, efs_id, avail_zones_mapping: dict, are_all_security_groups_c
             if head_node_target_id:
                 # Get list of security group IDs of the mount target
                 sg_ids = AWSApi.instance().efs.get_efs_mount_target_security_groups(head_node_target_id)
-                if not _check_in_out_access(sg_ids, port=2049, is_cidr_optional=are_all_security_groups_customized):
+                if not _is_access_allowed(
+                    sg_ids, subnets, port=2049, security_groups_by_nodes=security_groups_by_nodes
```
gmarciani: Let's avoid magic n umbers if possible. Since 2049 is a notable port, let's define a self-explanatory constant for that

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: The minor inconsistency comes from the special case of `PrefixListIds`. I would keep the name because the majority of the function is doing population. The boolean return is to handle a special case

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: I added a unit test with a new commit.
I didn't change other test mocking behavior to not accidentally reduce any coverage

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: Where would you put the constant? In the method, the file, or in constants.py?

cli/src/pcluster/aws/aws_resources.py
gmarciani: And this is a bad red flag for us.
I agree that we can consider this as out of scope for this specific PR, but we must address this issue in a follow up activity.

cli/src/pcluster/models/cluster.py
gmarciani: Sure, just to mention :-)

cli/src/pcluster/validators/cluster_validators.py
gmarciani: Thank you

cli/src/pcluster/validators/cluster_validators.py
gmarciani: May you please add a brief comment on the function to explain the behaviour, especially about the meaning of the boolean value?

cli/src/pcluster/validators/cluster_validators.py
gmarciani: I suggest to put it in `constants.py` for the time being.
I think that `constants.py` must be split in smaller files with single responsibilities, but it is the best candidate for now.

gmarciani (commented):
Gentle reminder: add a changelog entry mentioning this improvement.

cli/src/pcluster/validators/cluster_validators.py

cli/src/pcluster/validators/cluster_validators.py
hanwen-pcluste: Done

gmarciani (approved):

PR #4869, opened 2023-02-08
[headnode metric graphs] add disk and memory graphs for headnode CloudWatch dashboard

delongmeng:
### Description of changes
* Add new graphs `Disk Used Percent` and `Memory Used Percent` to CloudWatch Dashboard of a cluster to monitor memory usage and disk usage of the root volume on the head node.

### Tests
* Manual test: 
  * After code change, create a new cluster, and verify that new graphs `Disk Used Percent` and `Memory Used Percent` (within the `Head Node EC2 Metrics` - `Head Node Instance Metrics` section) show up in CloudWatch Dashboard as expected. 
  * For a head node of c5.large Nitro instance, "reboot" of the head node didn't affect the disk usage metric and graph.
  * Update the cluster to have Monitoring - Dashboards - CloudWatch disabled using the following configuration section, then observed that the dashboard was deleted; update the cluster again with default configuration (Enabled: true), and observed that the dashboard showed up again.
```
Monitoring:
  Dashboards:
    CloudWatch:
      Enabled: false  # true
```
* Unit tests added to verify the dashboard and graphs exist in the CloudFormation template when `Monitoring` - `Dashboards` - `CloudWatch` - `Enabled` is set to `true` (default) in the cluster configuration file.

### References
* PR that adds the disk and memory metrics: https://github.com/aws/aws-parallelcluster-cookbook/pull/1732
* PR that adds the disk and memory alarms: https://github.com/aws/aws-parallelcluster/pull/4877

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [x] Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4877, opened 2023-02-10
[headnode alarms] add headnode disk and memory alarms

delongmeng:
### Description of changes
* Add new CloudWatch alarms to monitor memory usage and root volume disk usage of the head node. Alarms will be triggered when either `mem_used_percent` or `disk_used_percent` exceeds the threshold of 90 (%) for 1 datapoint within 1 minute.
* Add CloudWatch Alarm related IAM policies: `PutMetricAlarm`, `DeleteAlarms`, and `DescribeAlarms`.

### Tests
- Manual test: After code change, create a new cluster, and verify that new alarms show up in CloudWatch Alarms as expected:
  - `{cluster-name}_MemAlarm_HeadNode` with the condition `mem_used_percent > 90 for 1 datapoints within 1 minute`. State is "OK".
  - `{cluster-name}_DiskAlarm_HeadNode` with the condition `disk_used_percent > 90 for 1 datapoints within 1 minute`. State is "OK".
  - Test the "In alarm" state: set the threshold of disk_used_percent to 40 and observed that the state of the alarm is "In alarm".
  - For a head node of c5.large Nitro instance, "reboot" of the head node didn't affect the disk usage metric and alarm.
  - Update the cluster to have Monitoring - Dashboards - CloudWatch disabled using the following configuration section, then observed that the alarms were deleted; update the cluster again with default configuration (Enabled: true), and observed that the alarms showed up again.
```
Monitoring:
  Dashboards:
    CloudWatch:
      Enabled: false  # true
```
- Unit tests added to verify the alarms exist in the CloudFormation template when Monitoring - Dashboards - CloudWatch - Enabled is set to true (default) in the cluster configuration file.

### References
* PR that adds the disk and memory metrics in CloudWatch: https://github.com/aws/aws-parallelcluster-cookbook/pull/1732
* PR that adds the disk and memory graphs in the CloudWatch dashboard: https://github.com/aws/aws-parallelcluster/pull/4869

### Checklist


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -299,6 +301,44 @@ def _add_resources(self):
...
+    def _add_alarms(self):
+        self.alarms = []
+
+        metrics_for_alarms = {
+            "Mem": cloudwatch.Metric(
+                namespace="CWAgent",
+                metric_name="mem_used_percent",
+                dimensions_map={"InstanceId": self.head_node_instance.ref},
+                statistic="Maximum",
+                period=Duration.seconds(60),
```
chenwany: Could you have this value as a constant in a single place `constants.py` and it can be reused for both memory and disk alarm?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -299,6 +301,44 @@ def _add_resources(self):
...
+        for metric_key, metric in metrics_for_alarms.items():
+            alarm_id = f"Pcluster{metric_key}Alarm"
+            alarm_name = f"{metric_key}Alarm_{self.stack.stack_name}_{self.head_node_instance.ref}"
+            self.alarms.append(
+                cloudwatch.Alarm(
+                    scope=self.stack,
+                    id=alarm_id,
+                    metric=metric,
+                    evaluation_periods=1,
+                    threshold=90,
```
chenwany: Could you set the threshold in  constant in constant?

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -118,6 +118,33 @@ def _generate_template(cluster, capsys):
     return generated_template
 
 
+@pytest.mark.parametrize(
+    "config_file_name",
+    [
+        "centos7.slurm.full.yaml",
```
chenwany: Instead of create new config here, can we utilize the config in example_config: see example in `test_cluster_lambda_functions_vpc_config` 
input_yaml = load_yaml_dict(f"{EXAMPLE_CONFIGS_DIR}/{config_file_name}")
We can see if we can change one of the cw setting in the example config to be true/false

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -299,6 +301,44 @@ def _add_resources(self):
...
+                namespace="CWAgent",
+                metric_name="disk_used_percent",
+                dimensions_map={"InstanceId": self.head_node_instance.ref},
+                statistic="Maximum",
+                period=Duration.seconds(60),
+            ),
+        }
+
+        for metric_key, metric in metrics_for_alarms.items():
+            alarm_id = f"Pcluster{metric_key}Alarm"
```
chenwany: What about adding HeadNode in both the ID and alarm_name, the reason is that in the future we can add more alarms for compute nodes

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -299,6 +301,44 @@ def _add_resources(self):
...
+            alarm_name = f"{metric_key}Alarm_{self.stack.stack_name}_{self.head_node_instance.ref}"
```
chenwany: Instead of having {self.head_node_instance.ref} in the alarm name, I think it would be nice to replace it with "HeadNode". Considering in the future we may want to add alarms for Computes, what do you think?

chenwany (commented):
Were you able to test the update update behavior to enable/disable monitoring for metrics

cli/src/pcluster/templates/cluster_stack.py
delongmeng: added default values for alarm setting in `constants.py`

cli/src/pcluster/templates/cluster_stack.py
delongmeng: added default values for alarm setting in `constants.py`

cli/src/pcluster/templates/cluster_stack.py
delongmeng: sure

cli/src/pcluster/templates/cluster_stack.py
delongmeng: yes sure

cli/tests/pcluster/templates/test_cluster_stack.py
delongmeng: Within cluster_stack.py, we build the dashboard and add alarms within the same branch:
```
        if self.config.is_cw_dashboard_enabled:
            self.cloudwatch_dashboard = CWDashboardConstruct(......)
            self._add_alarms()
```
The dashboard building is tested by `test_cw_dashboard_builder.py` so I copied the same config files for that test to here to be consistent.

As discussed, we will use config files in EXAMPLE_CONFIGS_DIR and change the `Dashboards - CloudWatch - Enabled` setting to be false in one of them (the default is true).

delongmeng:
> Were you able to test the update update behavior to enable/disable monitoring for metrics

Cluster Update was tested. Please refer to the Tests section in PR description.

chenwany (dismissed):
LGTM!

chenwany (approved):

chenwany (dismissed):
LGTM!

chenwany (dismissed):

chenwany:
Did you validate all these policies are required for alarms? Could you share the source of where mention these are required policies?

delongmeng:
> Did you validate all these policies are required for alarms? Could you share the source of where mention these are required policies?

These are just alarm related policies according to this documentation https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/iam-identity-based-access-control-cw.html

They might not be all required for the integration test purpose. For example, if we do an update on Cloudwatch-Dashboard-enabled that involves deletion of existing alarms, then the `cloudwatch:DeleteAlarms` policy would probably be needed.

Is the requirement here the minimum set of policies that ensure the current integration tests don't break? I can test that by adding or removing them one by one, if needed.

chenwany:
> Is the requirement here the minimum set of policies that ensure the current integration tests don't break? I can test that by adding or removing them one by one, if needed

Here we limit the minimum IAM policies to reduce the risk of unauthorized access to the resources. 
We need to add here the minimum set of policies here that make sure the alarm functionality work, like creating cluster with alarm, deleting cluster alarm, updating cluster...  
It is not just for integration test, but also for the API role.

delongmeng:
> Here we limit the minimum IAM policies to reduce the risk of unauthorized access to the resources.
> We need to add here the minimum set of policies here that make sure the alarm functionality work, like creating cluster with alarm, deleting cluster alarm, updating cluster...
> It is not just for integration test, but also for the API role.

Tested and confirmed that these 3 policies are required: `PutMetricAlarm`, `DeleteAlarms`, and `DescribeAlarms`.

`DescribeAlarmsForMetric` and `DescribeAlarmHistory` may not be needed, and have been removed.

chenwany (approved):

Issue #3425, opened 2021-11-01
Cross-account Hosted Zone for v3.0


gallanik:
I am trying to create a cluster with pre-existing hosted zone, but in a different account. The hosted zone is public and the IAM role attached to the EC2 machine which I am using to start a PC has route53:Get* permissions on all resources (*). Since Route53 does not support resource-based policy, there is nothing like whitelisting the account ID in R53 Hosted Zone.
Does v3.0 support cross-account public hosted zone? and If yes, what changes have to be made in order to get this to work.

cluster-config.yaml
```
Region: us-east-1
Image:
  Os: alinux2
HeadNode:
  InstanceType: t2.micro
  Networking:
    SubnetId: subnet-xxxx1111
  Ssh:
    KeyName: my-keypair
Scheduling:
  Scheduler: slurm
  SlurmSettings:
    ScaledownIdletime: 10
    Dns:
      HostedZoneId: xx12345
  SlurmQueues:
  - Name: def-q
    ComputeResources:
    - Name: t2micro
      InstanceType: t2.micro
      MinCount: 0
      MaxCount: 2
    Networking:
      SubnetIds:
      - subnet-xxxx1111
  - Name: spot-q
    ComputeResources:
    - Name: t2micro
      InstanceType: t2.micro
      MinCount: 0
      MaxCount: 2
    Networking:
      SubnetIds:
      - subnet-xxxx1111
```

gallanik:
We are an enterprise dealing with a huge estate of AWS Accounts and hundreds of projects/teams utilizing these accounts. We have a central services account, which has the public hosted zone which can be used by anyone within the company. Having a separate private hosted zone per account will create unnecessary complexity. 

I tried to disable DNS (DisableManagedDns: true) in the cluster config, and the cluster was created successfully, but now it fails while creating the compute nodes. Do I have to do anything else?

```
[2021-11-02T11:59:16.872] Requeuing JobId=2
--
[2021-11-02T11:59:16.872] update_node: node def-q-dy-t2micro-1 state set to DOWN
[2021-11-02T11:59:16.896] error: get_addr_info: getaddrinfo() failed: Name or service not known
[2021-11-02T11:59:16.896] error: slurm_set_addr: Unable to resolve "def-q-dy-t2micro-1"
[2021-11-02T11:59:16.896] error: fwd_tree_thread: can't find address for host def-q-dy-t2micro-1, check slurm.conf
[...]

PR #4887, opened 2023-02-10
Cross Account Enhancements: VPC and Route53

BrentSouza:
### Description of changes

Our organization controls VPC and Route53 from a central account. For VPCs, we create subnets and share those to application specific accounts. For Route53, we manage a hosted zone and create delegate roles with assume role policy in place to allow the application accounts to control creation of records.

These scenarios are not supported today with the way the parallelcluster configuration is interpreted/instrumented. This PR suggests a couple of patterns that could be used to allow the following:

* Specify Role or Profile names per client to override which IAM Credentials are used (in this case just Route53 is setup)
* Support full ARNs in relevant configuration fields where an Id is being specified and converted to an ARN

This would address #3425 and is similar to some changes made in #4881 but solve for that problem in a more broadly applicable way.

### Tests
* We'll be trying to run with these changes in our environment, but I'm uncertain about where we should include some formal tests in the current integration-test suite. Direction here would be appreciated.

### Checklist


gmarciani:
Hi @BrentSouza , 

thanks for your contribution and to keep the light on this use case!

Regarding the specific PR: the proposed approach injects the custom role in the Route53 client used by pcluster CLI, but it is the `clustermgtd` agent implemented in https://github.com/aws/aws-parallelcluster-node that is responsible for managing the DNS.

We added this feature request in our backlog for further discussions and prioritization.

BrentSouza:
We ended up going a simpler route for the shared subnets. We're testing internally now and if it works I'll open another PR with just those changes.

PR #4891, opened 2023-02-11
[develop] Refactor AZ setup for region and misc fix

lukeseawalker:

### Description of changes
* Refactor AZ setup for region

  Refactor AZ setup, so that if a default AZ is set for a given region, return it, otherwise return a random AZ

*  Get subnets using VPC stack methods

    Get subnets using VPC stack methods instead of retrieving them from CFN outputs. This because CFN outputs for subnets varies depending on the AZ where the subnets are created.

* Use CIDR from the defined list

  Use CIDR from the defined list so to avoid conflicts.

### Tests
n/a

### References
n/a

### Checklist

### DIFF
--- a/tests/integration-tests/conftest_networking.py
+++ b/tests/integration-tests/conftest_networking.py
@@ -136,6 +136,18 @@ def get_availability_zones(region, credential):
     return az_list
 
 
+def get_az_setup_for_region(region: str, credential: list):
+    """Return a default AZ ID and its name, the list of all AZ IDs and names."""
+    # TODO Region can be AZ ID, in this case convert it to Region
+    # TODO remove DEFAULT_AVAILABILITY_ZONE
+    az_id_to_az_name_map = get_az_id_to_az_name_map(region, credential)
+    az_ids = list(az_id_to_az_name_map)  # cannot be a dict_keys
+    default_az_id = random.choice(DEFAULT_AVAILABILITY_ZONE.get(region, az_ids))
+    default_az_name = az_id_to_az_name_map.get(default_az_id)
+
+    return default_az_id, default_az_name, az_id_to_az_name_map
+
+
 def get_az_id_to_az_name_map(region, credential):
     """Return a dict mapping AZ IDs (e.g, 'use1-az2') to AZ names (e.g., 'us-east-1c')."""
     # credentials are managed manually rather than via setup_sts_credentials because this function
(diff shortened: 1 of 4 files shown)


tests/integration-tests/conftest_networking.py
```diff
@@ -136,6 +136,19 @@ def get_availability_zones(region, credential):
     return az_list
 
 
+def get_az_setup_for_region(region: str, credential: list):
+    """Return a default AZ ID and its name, the list of all AZ IDs and names."""
+    # TODO Region can be AZ ID, in this case convert it to Region
+    # TODO remove DEFAULT_AVAILABILITY_ZONE
+    az_id_to_az_name_map = get_az_id_to_az_name_map(region, credential)
```
NSsirena, line 143: Since below you are transforming the results into pairs .. 

for index, (az_id, az_name) in enumerate(zip(az_ids, az_names)):

why not do it here ?

az_pairs = [(id,name) for id,name in az_id_to_az_name_map] ?

tests/integration-tests/tests/schedulers/conftest.py
```diff
@@ -104,8 +102,8 @@ def _create_database_stack(stack_factory, request, region, vpc_stack_for_databas
             {"ParameterKey": "ClusterName", "ParameterValue": cluster_name},
             {"ParameterKey": "Vpc", "ParameterValue": vpc_stack_for_database.cfn_outputs["VpcId"]},
             {"ParameterKey": "AdminPasswordSecretString", "ParameterValue": admin_password},
-            {"ParameterKey": "Subnet1CidrBlock", "ParameterValue": "192.168.8.0/23"},
-            {"ParameterKey": "Subnet2CidrBlock", "ParameterValue": "192.168.4.0/23"},
+            {"ParameterKey": "Subnet1CidrBlock", "ParameterValue": CIDR_FOR_CUSTOM_SUBNETS[-1]},
+            {"ParameterKey": "Subnet2CidrBlock", "ParameterValue": CIDR_FOR_CUSTOM_SUBNETS[-2]},
```
NSsirena, line 106: Out of curiosity ... why are you picking CIDRs in reverse order ?

lukeseawalker: quick trick to make sure they aren't already taken

tests/integration-tests/conftest_networking.py
lukeseawalker: it's because I was using the two lists later on. If you think it's cleaner, I'd rather return the dict

NSsirena: There is a slight computational overhead in how the separated lists are created, but this is more a nitpick.

The concern here is "are we assuming that those lists have the same order ?" ...that is the first element of the az_id[1] -> az_name[1] ... and so on ?

If yes, we should keep them together to make the rest of the code more robust.

tests/integration-tests/tests/schedulers/conftest.py
NSsirena: mmm ... how long it will hold ?
seems risky

NSsirena (approved):

tests/integration-tests/tests/schedulers/conftest.py
lukeseawalker: tests only use first element of CIDR_FOR_CUSTOM_SUBNETS
In the next refactoring I think we may want to drop CIDR_FOR_CUSTOM_SUBNETS completely and rather of using a subnet in same AZ different CIDR, we may want to use a different subnet in different AZ

PR #4895, opened 2023-02-13
[Isolated Regions] Fix support for US isolated regions.

gmarciani:
*This PR makes use of cookbook changes in https://github.com/aws/aws-parallelcluster-cookbook/pull/1763*

### Changes
Fix support for US isolated regions.

In particular:

1. Explicitly configure the CloudFormation endpoint used by cfn-hup by adding the 'url' parameter into `/etc/cfn/cfn-hup.conf`.
1. Add a step to configure cluster nodes when running in US isolated regions. The following conditional actions have been added to the user data for both head and compute nodes: a. execute the script `/opt/parallelcluster/scripts/patch-iso-instance.sh`. b. reload the user profile so that 'root' can execute the other user data action with the new/updated environment variables provided by the patching script.
1. Set default EBS volume type to `gp2` is US isolated regions. The default volume type is reflected in: 
    a. head node root volume. 
    b. compute resource root volume. 
    c. shared EBS volumes.

Note: Not adding a new entry in the CHANGELOG because the feature is in pending.

### Tests
Manual tests (US ISO TESTS ARE STILL ONGOING) on release branch 3.5:

1. Cluster creation (executed both in Commercial and US Isolated)
1. Cluster update with dynamic file system mounting (executed both in Commercial and US Isolated)
1. Cluster update with compute nodes min-count change (executed both in Commercial and US Isolated)
1. Submission of job that requires compute fleet autoscaling (executed both in Commercial and US Isolated)

A full regression tests will be done on the authoritative pipeline once the change will be merged.


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1059,6 +1059,8 @@ def _add_head_node(self):
             indent=4,
         )
 
+        cloudformation_url = f"https://cloudformation.{self.stack.region}.{get_url_domain_suffix()}"
```
lukeseawalker, line 1062: how this was working in CN?
Before, we weren't setting the CloudFormationUrl for CN, now we are setting the URL based on the get_url_domain_suffix. Is this a change?

gmarciani: I think the change is safe.
CloudFormation endpoints in the supported China regions are:
1. cn-north-1: cloudformation.cn-north-1.amazonaws.com.cn
2. cn-northwest-1: cloudformation.cn-northwest-1.amazonaws.com.cn

gmarciani: Will execute a manual test (create+update) in China to double check my assumptions in any case and let you know.

PR #4897, opened 2023-02-14
Multi nic

Nbot07A:
### Description of changes
* test that ip is set to NIC 0 Device 0 primary ip to ensure IP alignment with Multi NIC instances

### Tests
* Modified and passed integration test test_trainium with new method _test_primary_ip

### Checklist


tests/integration-tests/tests/trainium/test_trainium.py
```diff
@@ -64,3 +66,14 @@ def _test_ccl_two_nodes(test_datadir, remote_command_executor, scheduler_command
 
     print(result.stdout)
     assert_that(result.stdout).contains("CCL(1)", "CCL(50)", "CCL(99)", "CCL(100)")
+
+
+def _test_primary_ip(test_datadir, remote_command_executor, scheduler_commands):
+    result = scheduler_commands.submit_script(str(test_datadir / "test-primary-ip.sh"), partition="queue-trn32")
+    job_id = scheduler_commands.assert_job_submitted(result.stdout)
+    scheduler_commands.wait_job_completed(job_id)
+    scheduler_commands.assert_job_succeeded(job_id)
```
francesco-giordano, line 75: **Issue:** as the current logic of the script, if there is a misalignment in the hosts, the test does not pass this point because the script return 1.

tests/integration-tests/tests/trainium/test_trainium/test_trainium/test-primary-ip.sh
```diff
@@ -0,0 +1,29 @@
...
+      if [[ $word == $mac_ip ]]
+      then
+        echo PASSED | tee -a output-primary-ip.txt
+        exit 0
+      fi
+    done
+  fi
+done
+echo Error: Route53 IP does not match host IP | tee -a output-primary-ip.txt
+exit 1
```
francesco-giordano: **Issue:** I think that the inner `for word in $IP_HOSTS` is not aligned on what we want to achieve. We want to check if we have the line below in the `/etc/hosts`
```
IP FQDN HOSTNAME
#Example:
10.0.136.84 compute1-st-queue1-i1-1.pc2trn1.pcluster compute1-st-queue1-i1-1
```

Below a sample on how to query Route54: in the `IP` variable you will have the value store in Route53 
```
DNS_SERVER=""
grep Ubuntu /etc/issue &>/dev/null && DNS_SERVER=$(resolvectl dns | awk '{print $4}' | sort -r | head -1)
IP="$(host $HOSTNAME $DNS_SERVER | tail -1 | awk '{print $4}')"
```

Here a fast way to get the domain of the instance
```
DOMAIN=$(jq .cluster.dns_domain /etc/chef/dna.json | tr -d \")
```

Here is the line that should be present in the `/etc/hosts`
```
$IP $HOSTNAME.${DOMAIN::-1} $HOSTNAME
```

Can you refactor the script to leverage the function above?

Nbot07A: resolvectl wasn't installed on an instance created from a pcluster ami. Also, there isn't a dna.json in /etc/chef on that instance either.  So, I'm not sure how to verify the above. Could you provide an example that works on an instance created from a pcluster ami?

francesco-giordano: The dna.json should be there when the instance is run in a cluster. Below an updated command to retrieve the information that work also on ubuntu18
```
DNS_SERVER=""
grep Ubuntu /etc/issue &>/dev/null && DNS_SERVER=$(systemd-resolve --status | grep "DNS Servers" | awk '{print $3}'` | sort -r | head -1)
IP="$(host $HOSTNAME $DNS_SERVER | tail -1 | awk '{print $4}')"
```

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster/pull/5143

PR #4913, opened 2023-02-15
Add RHEL8 to image builder components

enrico-usai:
### Description of changes
* This change doesn't impact the user experience of the cluster related commands. Users cannot create a cluster with RHEL8.
* This change enables the possibility to run the ImageBuilder components for RHEL8.  

### Tests
* Manually tested calling `pcluster build-image` for RHEL8.

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -57,7 +57,7 @@ phases:
               [ -n "${CfnParamCincInstaller}" ] && CINC_URL="${CfnParamCincInstaller}"
               echo "${!CINC_URL}"
 
-      # Check input base AMI OS and get OS information, the output should be like centos.7 | amzn.2 | ubuntu.18.04 | ubuntu.20.04
+      # Check input base AMI OS and get OS information, the output should be like centos.7 | amzn.2 | ubuntu.18.04 | ubuntu.20.04 | rhel.8.7
       - name: OperatingSystemRelease
         action: ExecuteBash
         inputs:
@@ -90,6 +90,8 @@ phases:
                 OS='ubuntu1804'
               elif [ `echo "${!RELEASE}" | grep '^ubuntu\.20'` ]; then
                 OS='ubuntu2004'
+              elif [ `echo "${!RELEASE}" | grep '^rhel\.8'` ]; then
+                OS='rhel8'
               else
                 echo "Operating System '${!RELEASE}' is not supported. Failing build."
                 exit {{ FailExitCode }}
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -144,15 +146,15 @@ phases:
...
                   echo "This component does not support '${!RELEASE}'. Failing build."
                   exit {{ FailExitCode }}
                 fi
 
-                # This component only supports aarch64 CPUs on Amazon Linux 2, Ubuntu1804, Ubuntu2004 and Centos7
+                # This component only supports aarch64 CPUs on Amazon Linux 2, Ubuntu1804, Ubuntu2004, Centos7 and RHEL8
                 ARCH=$(uname -m)
                 if [[ `echo ${!ARCH}` == 'aarch64' ]]; then
-                  if [ `echo "${!RELEASE}" | grep -Ev '^(amzn\.2|centos\.7|ubuntu\.18\.04|ubuntu\.20\.04)'` ]; then
+                  if [ `echo "${!RELEASE}" | grep -Ev '^(amzn\.2|centos\.7|ubuntu\.18\.04|ubuntu\.20\.04|rhel\.8)'` ]; then
```
eantonin, line 157: question, not blocking: Should this be 8.7?

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -75,15 +77,15 @@ phases:
...
                 echo "This component does not support '${!RELEASE}'. Failing build."
                 exit {{ FailExitCode }}
               fi
 
-              # This component only supports aarch64 CPUs on Amazon Linux 2, Ubuntu1804, Ubuntu2004 and Centos7
+              # This component only supports aarch64 CPUs on Amazon Linux 2, Ubuntu1804, Ubuntu2004, Centos7 and RHEL 8
               ARCH=$(uname -m)
               if [[ `echo ${!ARCH}` == 'aarch64' ]]; then
-                if [ `echo "${!RELEASE}" | grep -Ev '^(amzn\.2|centos\.7|ubuntu\.18\.04|ubuntu\.20\.04)'` ]; then
+                if [ `echo "${!RELEASE}" | grep -Ev '^(amzn\.2|centos\.7|ubuntu\.18\.04|ubuntu\.20\.04|rhel\.8)'` ]; then
```
eantonin, line 88: same as above

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
enrico-usai: I don't think we should be so strict to the minor version. 

If we add this strict check in the ImageBuilder components this means we need to release a new version of the CLI to support Rhel8.8 or 8.9. Instead by keeping this checks less strict probably we would be to support future minor versions out of the box.

enrico-usai: BTW this is aligned with CentOS.

eantonin (approved):

PR #4938, opened 2023-02-22
Remove obfuscation of Expiration field in the print of credentials

enrico-usai:
The obfuscate_credentials function cannot be executed for the Expiration element
because it is a datetime and not a string.

Test:
```
>>> now = datetime.datetime.now()
>>> now_str = str(now)
>>> now_str
'2023-02-22 10:08:28.791525'
```

### DIFF
--- a/tests/integration-tests/framework/credential_providers.py
+++ b/tests/integration-tests/framework/credential_providers.py
@@ -146,7 +146,10 @@ def obfuscate_credentials(creds_dict):
     obfuscated_dict = {}
     for key, value in creds_dict.items():
         if value:
-            obfuscated_dict[key] = value[0:3] + "*" * (len(value) - 3)
+            if key == "Expiration":
+                obfuscated_dict[key] = str(value)
+            else:
+                obfuscated_dict[key] = value[0:3] + "*" * (len(value) - 3)
         else:
             obfuscated_dict[key] = value
     return obfuscated_dict


tests/integration-tests/framework/credential_providers.py
```diff
@@ -58,7 +58,6 @@ def sts_credential_provider(region, credential_arn, credential_external_id=None,
 
     logging.info("Assuming STS credentials for region %s and role %s", region, credential_arn)
     aws_credentials = _retrieve_sts_credential(region, credential_arn, credential_external_id, credential_endpoint)
-    logging.info("Retrieved credentials %s", obfuscate_credentials(aws_credentials))
 
     try:
         logging.info("Unsetting current credentials %s", obfuscate_credentials(credentials_to_backup))
```
lukeseawalker: it will then break here, when the STS retrieved credentials will be the current ones

enrico-usai: nope, because `_get_current_credentials` return a dict without the expiration:
```
    credentials_to_backup = _get_current_credentials()
...
def _get_current_credentials():
    return {
        "AWS_ACCESS_KEY_ID": os.environ.get("AWS_ACCESS_KEY_ID"),
        "AWS_SECRET_ACCESS_KEY": os.environ.get("AWS_SECRET_ACCESS_KEY"),
        "AWS_SESSION_TOKEN": os.environ.get("AWS_SESSION_TOKEN"),
        "AWS_PROFILE": os.environ.get("AWS_PROFILE"),
    }
```

enrico-usai: Modified the logic to avoid obfuscating the expiration field

NSsirena (approved):

PR #4953, opened 2023-02-23
[Isolated Regions] Fix support for US isolated regions.

gmarciani:
*This PR makes use of cookbook changes in https://github.com/aws/aws-parallelcluster-cookbook/pull/1800*

### Description of changes
Fix support for US isolated regions.

In particular:

1. Explicitly configure the CloudFormation endpoint and region used by cfn-bootstrap tools (cfn-hup, cfn-init, cfn-signal).
1. Add a step to configure cluster nodes when running in US isolated regions. The following conditional actions have been added to the user data for both head and compute nodes: 
    a. execute the script '/opt/parallelcluster/scripts/patch-iso-instance.sh'. 
    b. reload the user profile so that 'root' can execute the other user data action with the new/updated environment variables provided by the patching script.
1. Set default EBS volume type to 'gp2' is US isolated regions. The default volume type is reflected in: 
    a. head node root volume. 
    b. compute resource root volume. 
    c. shared EBS volumes.

### Tests
Manual tests:
1. Cluster creation (executed both in Commercial and US Isolated)
3. Cluster update with dynamic file system mounting (executed both in Commercial and US Isolated)
4. Cluster update with compute nodes min-count change (executed both in Commercial and US Isolated)

A full regression tests will be done on the authoritative pipeline once the change will be merged.


cli/src/pcluster/config/cluster_config.py
```diff
@@ -208,7 +209,10 @@ def __init__(
     ):
         super().__init__(**kwargs)
         self.encrypted = Resource.init_param(encrypted, default=True)
-        self.volume_type = Resource.init_param(volume_type, default=EBS_VOLUME_TYPE_DEFAULT)
+        self.volume_type = Resource.init_param(
+            volume_type,
+            default=EBS_VOLUME_TYPE_DEFAULT_US_ISO if get_region().startswith("us-iso") else EBS_VOLUME_TYPE_DEFAULT,
```
enrico-usai, line 214: suggestion: it's better to move this logic into an utility function and have a map for `EBS_VOLUME_TYPE_DEFAULT` with us-iso and default values.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1059,6 +1062,8 @@ def _add_head_node(self):
             indent=4,
         )
 
+        cloudformation_url = f"https://cloudformation.{self.stack.region}.{get_url_domain_suffix()}"
```
enrico-usai: we should define a common function here to retrieve cloudformation_url and avoid duplication.

enrico-usai (dismissed):

cli/src/pcluster/config/cluster_config.py
gmarciani: I agree, will address this improvement in follow up PRs.

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I agree, will address this improvement in follow up PRs.

gmarciani: Addressing in this PR.

cli/src/pcluster/config/cluster_config.py
gmarciani: I agree with you: I think we should have something like a _region-based property store_.

Something like:

```
self.volume_type = get_property(property=Property.EBS_VOLUME_TYPE_DEFAULT, region=get_region())
```

If you agree with this approach I think it should be considered out of scope for this PR and be addressed on a dedicated iteration since it would be a general solution to apply also elsewhere.

enrico-usai (approved):

PR #4961, opened 2023-02-27
[develop] Add unmarshal_az to cli_dimension_filtering

NSsirena:
### Description of changes
When filtering the tests to run based on the input configuration, the `regions` dimensions used to populate the allowed values my contain a list of regions/az_id.
Here we apply the unmarshal_az functions to all the values to ensure that every override is properly handled and a valid list of regions is used to filter the tests.

### Tests
* Verified locally that when multiple dimensions are specified the overrides are properly handled.

### Checklist

### DIFF
--- a/tests/integration-tests/conftest_tests_config.py
+++ b/tests/integration-tests/conftest_tests_config.py
@@ -86,7 +86,12 @@ def apply_cli_dimensions_filtering(config, items):
     """Filter tests based on dimensions passed as cli arguments."""
     allowed_values = {}
     for dimension in DIMENSIONS_MARKER_ARGS:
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest_tests_config.py
```diff
@@ -86,7 +86,12 @@ def apply_cli_dimensions_filtering(config, items):
     """Filter tests based on dimensions passed as cli arguments."""
     allowed_values = {}
     for dimension in DIMENSIONS_MARKER_ARGS:
-        allowed_values[dimension] = config.getoption(dimension + "s")
+        value = config.getoption(dimension + "s")
```
lukeseawalker, line 89: values in place of value, since it is a list

PR #4985, opened 2023-03-02
[Multi User] Add support for AD password as SSM Parameter.

gmarciani:
### Description of changes
Add support for AD password as SSM Parameter.
With this extension we will be able to provide the multi-user feature in those regions where Secrets Manager is not supported, but SSM is.

#### Customer Experience
With this change the customer will be able to specify the AD password in [DirectoryService/PasswordSecretArn](https://docs.aws.amazon.com/parallelcluster/latest/ug/DirectoryService-v3.html#yaml-DirectoryService-PasswordSecretArn) both as a secret in Secrets Manager (current behaviour) and a parameter in SSM.

### Tests
* Cluster creation + AD user login with AD password specified as a secret in Secrets Manager.
* Cluster creation + AD user login with AD password specified as a parameter in SSM.
* Unhappy validation cases

### DIFF
--- a/cli/src/pcluster/aws/ssm.py
+++ b/cli/src/pcluster/aws/ssm.py
@@ -1,0 +1,28 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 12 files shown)


cli/src/pcluster/aws/ssm.py
```diff
@@ -0,0 +1,29 @@
...
+
+class SsmClient(Boto3Client):
+    """SSM Boto3 client."""
+
+    def __init__(self):
+        super().__init__("ssm")
+
+    @AWSExceptionHandler.handle_client_exception
+    # @Cache.cached
+    def get_parameter(self, name: str):
```
enrico-usai, line 21: Should we modify documentation and instance-policies used in tests to support this?

enrico-usai (dismissed):

cli/src/pcluster/aws/ssm.py
gmarciani: Yes, we should.
I'll include the change for the user role policy used by the test runner in this PR.
Regarding the documentation we will document the change later on.

enrico-usai (approved):

PR #5008, opened 2023-03-07
Common error metrics

chenwany:
### Description of changes
Added metrics filters to identify common errors from JSON formatting log events on compute nodes that occur during cluster running time. These filters are integrated into the ParallelCluster cloudwatch dashboard to allow users to monitor and visualize metrics related to node launch failures, unhealthy slurm nodes, instance failure EC2 healthy check and nodes failed with running custom script.
image

### Tests
* Unit tests
* Integration test

### References

### Checklist


cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -172,15 +179,29 @@ def _add_text_widget(self, markdown):
...
+                title=title,
+                left=metric_list,
+                region=self._stack_region,
+                width=self.graph_width,
+                height=self.graph_height,
+                left_annotations=[
+                    cloudwatch.HorizontalAnnotation(
+                        value=1, label="github.com/aws/aws-parallelcluster/wiki#known-issues-"
+                    ),
+                ],
```
davprat: Instead of creating the widget in two different different was in an if-else, would it be better (or even possible) to simple define the `left_annotation` using the ternary operator? Something like:
```
widget = cloudwatch.GraphWidget(
                title=title,
                left=metric_list,
                region=self._stack_region,
                width=self.graph_width,
                height=self.graph_height,
                left_annotations=[
                    cloudwatch.HorizontalAnnotation(
                        value=1, label="github.com/aws/aws-parallelcluster/wiki#known-issues-"
                    ),
                ] if error_label else None
            )
```

Wdyt?

cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -172,15 +179,29 @@ def _add_text_widget(self, markdown):
         self.cloudwatch_dashboard.add_widgets(text_widget)
         self._update_coord_after_section(d_y=1)
 
-    def _generate_graph_widget(self, title, metric_list):
+    def _generate_graph_widget(self, title, metric_list, error_label):
```
davprat: I see that the `error_label` parameter mostly gets passed as`False`, would it make sense to make that the default value for this parameter so most callers won't have to explicitly pass `False`? Also, I did a quick search but didn't find any case where we set this to `True`, is this expected or did I just miss it?

cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -30,11 +31,13 @@ def __init__(self, x_value: int, y_value: int):
 _PclusterMetric = namedtuple(
     "_PclusterMetric", ["title", "metrics", "supported_vol_types", "namespace", "additional_dimensions"]
 )
+_CustomMetricFilter = namedtuple("MetricFilter", ["metric_name", "filter_pattern", "metric_value"])
```
davprat: Should the typename parameter be `_CustomMetricFilter` as well to be consistent with the other named tuples here?

davprat (dismissed):
I just left some minor questions, but otherwise it looks good to me.

cli/src/pcluster/templates/cw_dashboard_builder.py
chenwany: yes, good point, changed

cli/src/pcluster/templates/cw_dashboard_builder.py
chenwany: Sounds good to me! Changed

cli/src/pcluster/templates/cw_dashboard_builder.py
chenwany: Yes, set the default error_label to false since in most if the case it is false. I updated the PR to add a todo in the code on whether we set the annotation to True or add widgets below the graph widgets to show the troubleshooting links. Will update the PR once the decision is made

davprat (dismissed):
Looks good to me

cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -556,3 +692,29 @@ def _new_filter(self, pattern=None, param=None):
...
+        widgets_list = []
+        for error in cluster_common_errors:
+            metric_list = []
+            for new_filter in error.metric_filters:
+                self.custom_pcluster_metric_filter(
+                    metric_name=new_filter.metric_name,
+                    filter_pattern=new_filter.filter_pattern,
+                    custom_namespace=custom_namespace,
+                    metric_value=new_filter.metric_value,
+                )
```
delongmeng, line 724: Just a question out of curiosity:
What does this `self.custom_pcluster_metric_filter()` step do?
I see the `custom_pcluster_metric_filter()` function returns a metric_filter, but here the returned metric filter does not get stored or used right? And it's also not an attribute of any object? How does it influence the `cloudwatch_metric` or `metric_list`?
But the way, this `custom_pcluster_metric_filter()` seems to be a method, should the name contain a verb, like `create_custom_pcluster_metric_filter()`?

delongmeng (commented):
A general question, for the metric filter to work on the time series data, how is the timestamp created here? We have a timestamp field in the json format, but where is it used to parse the time? Otherwise the timestamp will need to be added by CloudWatch in a batch mode and would be less accurate, right?

cli/src/pcluster/templates/cw_dashboard_builder.py
chenwany: Good question, the custom_pcluster_metric_filter here create a metric filter cfn resource with the log group. We do not return or used it because don't need to use any attribute of the metric_filter here. Metrics are generated with namespace, metric_name and dimension specified in metrics filters. 
Yes, we can rename the function.

cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -222,6 +229,150 @@ def _add_conditional_storage_widgets(
...
+                filter_pattern=_generate_metric_filter_pattern(
+                    "protected-mode-error-count", "dynamic-resume-timeout-error"
+                ),
+                metric_value=metric_value,
+            ),
+            _CustomMetricFilter(
+                metric_name="EC2MaintenanceEvent",
+                filter_pattern=_generate_metric_filter_pattern(
+                    "nodes-failing-health-check-count", "ec2_health_check", failure_name="health-check-type"
+                ),
```
davprat: Would it be easier for you if I changed the failure_name to be consistent with the node_launch_failure_count?

davprat (dismissed):

davprat (approved):

chenwany:
> A general question, for the metric filter to work on the time series data, how is the timestamp created here? We have a timestamp field in the json format, but where is it used to parse the time? Otherwise the timestamp will need to be added by CloudWatch in a batch mode and would be less accurate, right?

This is discussed and answered in our follows-up discussion. The timeframe for data scanning to get the match pattern is controlled on the metrics filter side. It is not decided by the timestamp filed set in the json format. We are not using that filed for the metrics filter

PR #5044, opened 2023-03-10
`[cloudformation/custom_resource]` Add CFN custom resource for cluster.

charlesg3:
### Description of changes
* Add a CloudFormation stack for a Cluster CloudFormation Custom Resource

### Tests
* Tests added in the `cloudformation` directory which exercise the new stack. The tests were run locally against an uploaded python layer with the version manually set to `3.5.0`.

Results:

```.sh
tests $ pytest ./test_cluster_custom_resource.py -m local
======================================== test session starts ========================================
platform darwin -- Python 3.9.6, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/cgruenwa/src/cgpc/aws-parallelcluster/cloudformation/tests, configfile: pytest.ini
plugins: xdist-3.1.0, html-3.2.0, mock-3.10.0, typeguard-2.13.3, cov-4.0.0, metadata-2.0.4, datadir-1
.4.1
collected 7 items

test_cluster_custom_resource.py .......                                                       [100%]

========================================= warnings summary ==========================================
test_cluster_custom_resource.py::test_cluster_update
test_cluster_custom_resource.py::test_cluster_update
  /Users/cgruenwa/Library/Python/3.9/lib/python/site-packages/connexion/decorators/validation.py:16:
DeprecationWarning: Accessing jsonschema.draft4_format_checker is deprecated and will be removed in a
 future release. Instead, use the FORMAT_CHECKER attribute on the corresponding Validator.
    from jsonschema import Draft4Validator, ValidationError, draft4_format_checker

test_cluster_custom_resource.py: 19 warnings
  /Users/cgruenwa/Library/Python/3.9/lib/python/site-packages/pcluster/cli/entrypoint.py:113: Depreca
tionWarning: 'JSONEncoder' is deprecated and will be removed in Flask 2.3. Use 'Flask.json' to provid
e an alternate JSON implementation instead.
    return dispatch_func(**kwargs)

test_cluster_custom_resource.py::test_cluster_delete_out_of_band
  /Users/cgruenwa/Library/Python/3.9/lib/python/site-packages/pcluster/cli/middleware.py:109: Depreca
tionWarning: 'JSONEncoder' is deprecated and will be removed in Flask 2.3. Use 'Flask.json' to provid
e an alternate JSON implementation instead.
    ret = func(**kwargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================ 7 passed, 22 warnings in 2720.18s (0:45:20) ============================
```

### Checklist

### DIFF
--- a/cloudformation/custom_resource/cluster.yaml
+++ b/cloudformation/custom_resource/cluster.yaml
@@ -99,7 +99,7 @@ Resources:
         - { StackIdSuffix: !Select [2, !Split ['/', !Ref 'AWS::StackId']] }
       TracingConfig:
         Mode: Active
-      MemorySize: 1024
+      MemorySize: 2048
       Timeout: 60
       Code:
         ZipFile: |
--- a/cloudformation/custom_resource/cluster.yaml
+++ b/cloudformation/custom_resource/cluster.yaml
@@ -33,7 +33,7 @@ Resources:
         - { StackIdSuffix: !Select [2, !Split ['/', !Ref 'AWS::StackId']] }
       Description: Library which contains aws-parallelcluster python package and dependencies
       Content:
-        S3Bucket: !Sub ${Region}-aws-parallelcluster
+        S3Bucket: !Sub ${AWS::Region}-aws-parallelcluster
         S3Key: !Sub
         - parallelcluster/${Version}/layers/aws-parallelcluster/lambda-layer.zip
         - { Version: !FindInMap [ParallelCluster, Constants, Version] }
@@ -121,14 +121,14 @@ Resources:
           from pcluster.api.errors import exception_message, NotFoundException
           import pcluster.lib as pc
 
-          crhelper_path = "/opt/python//pcluster/resources/custom_resources/custom_resources_code"
+          crhelper_path = "/opt/python/pcluster/resources/custom_resources/custom_resources_code"
           sys.path.insert(0, crhelper_path)
           from crhelper import CfnResource
           helper = CfnResource()
 
           def update_response(response):
-              keep_keys = {"version", "creattionTime", "clusterName", "clusterStatus", "cloudformationStackArn", "lastUpdatedTime", "region"}
-              helper.Data.update({k: v for k, v in response.items() if k in keep_keys})
+              logger.info(response)
+              helper.Data.update(response)
 
           def serialize(val):
               return utils.to_iso_timestr(val) if isinstance(val, datetime.date) else val
@@ -141,7 +141,9 @@ Resources:
               elif event["RequestType"].upper() == "UPDATE" and event["PhysicalResourceId"] != properties.get("ClusterName"):
                   raise ValueError("Cannot update the ClusterName in the properties.")
 
-              physical_resource_id = properties["ClusterName"]
+              cluster_name = properties["ClusterName"]
+              logger.info(f"{event['RequestType'].upper()}: {cluster_name}")
+              physical_resource_id = cluster_name
 
               try:
                   kwargs = {**{pcluster.utils.to_snake_case(k): serialize(v) for k, v in properties.items() if k not in {"ServiceToken"}}, "wait": False}
@@ -153,6 +155,7 @@ Resources:
                   message = pcluster.api.errors.exception_message(e)
                   str_msg = encoder.JSONEncoder().encode(message)
                   if not re.search(r"No changes found", str_msg):
+                    logger.info(f"No changes found to update: {cluster_name}")
                     raise ValueError(str_msg)
 
               return physical_resource_id
@@ -168,6 +171,7 @@ Resources:
           @helper.delete
           def delete(event, context):
               cluster_name = event["ResourceProperties"].get("ClusterName")
+              logger.info(f"Deleting: {cluster_name}")
               try:
                   update_response(pc.delete_cluster(cluster_name=cluster_name))
               except (ParameterException, NotFoundException): # cluster deleted or invalid name -- ignore here.
@@ -195,7 +199,9 @@ Resources:
               # only invalid parameter can be the name
               except (ParameterException, NotFoundException):
                   if event["RequestType"].upper() == "DELETE":
-                      return True
+                      # Returning a value here signifies that the delete is completed and we can stop polling
+                      # not returning a value here causes cfn resource helper to keep polling.
+                      return cluster_name
                   raise ValueError(f"{cluster_name} failed {event['RequestType'].upper()}. See substack for further details.")
 
           @helper.poll_create
--- a/cloudformation/custom_resource/cluster.yaml
+++ b/cloudformation/custom_resource/cluster.yaml
@@ -1,0 +1,225 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: AWS ParallelCluster CloudFormation Template
+
+Parameters:
+
+  CustomLambdaRole:
+    Description: Custom role to use for PC Lambda
+    Type: String
+    Default: ''
+
+  AdditionalIamPolicies:
+    Description: Comma-delimited list of additional IAM Policies to add to the cluster (only used if CustomLambdaRole isn't provided).
+    Type: CommaDelimitedList
+    Default: ''
+
+Mappings:
+  ParallelCluster:
+    Constants:
+      Version: 3.6.0  # major.minor.patch+alpha/beta_identifier
+
+Conditions:
+  CustomRoleCondition: !Not [!Equals [!Ref CustomLambdaRole, '']]
+  UsePCPolicies: !Not [!Condition CustomRoleCondition ]
+  UseAdditionalIamPolicies: !Not [!Equals [!Join ['', !Ref AdditionalIamPolicies ], '']]
+
+Resources:
+  PclusterLayer:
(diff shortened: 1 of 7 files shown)


cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,225 @@
...
+          from pcluster.api.errors import exception_message, NotFoundException
+          import pcluster.lib as pc
+
+          crhelper_path = "/opt/python//pcluster/resources/custom_resources/custom_resources_code"
+          sys.path.insert(0, crhelper_path)
+          from crhelper import CfnResource
+          helper = CfnResource()
+
+          def update_response(response):
+              keep_keys = {"version", "creattionTime", "clusterName", "clusterStatus", "cloudformationStackArn", "lastUpdatedTime", "region"}
```
demartinofra: typo `creattionTime`

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,225 @@
...
+
+          import pcluster.api.controllers.cluster_operations_controller
+          import pcluster.api.errors
+          import pcluster.utils
+          from pcluster.api import encoder
+          from pcluster.cli.exceptions import APIOperationException, ParameterException
+          from pcluster.api.errors import exception_message, NotFoundException
+          import pcluster.lib as pc
+
+          crhelper_path = "/opt/python//pcluster/resources/custom_resources/custom_resources_code"
```
demartinofra: typo: double `/`

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,225 @@
...
+          sys.path.insert(0, crhelper_path)
+          from crhelper import CfnResource
+          helper = CfnResource()
+
+          def update_response(response):
```
demartinofra, line 129: why do we need to filter out some keys?

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,225 @@
...
+                      return cluster_name
+                  elif status in {"CREATE_FAILED", "UPDATE_FAILED", "DELETE_FAILED"}:
+                      raise ValueError(f"{cluster_name} failed {event['RequestType'].upper()}. See substack for further details.")
+
+              # If create fails and we try to roll-back (e.g. delete),
+              # gracefully handle missing cluster. on the delete pathway, the
+              # only invalid parameter can be the name
+              except (ParameterException, NotFoundException):
+                  if event["RequestType"].upper() == "DELETE":
+                      return True
```
demartinofra: is it expected to return a bool here and a string above?

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,225 @@
...
+      Code:
+        ZipFile: |
+          import datetime
+          import json
+          import logging
+          import random
+          import re
+          import string
+          import sys
+          logger = logging.getLogger()
```
demartinofra, line 113: I suggest to add some log prints here and there otherwise in case of failure this becomes very hard to debug

cloudformation/tests/test_cluster_custom_resource.py
```diff
@@ -0,0 +1,245 @@
+"""Test cases for Cluster CloudFormation Custom Resource."""
```
demartinofra, line 1: these local tests are extremely useful but need to be executed manually. I'd suggest to also include an end to end test (even as part of a separate review) so that the same can be periodically executed to validate both the develop and the released artifacts. 
For example if a new functionality is added and we forget to update the IAM policies how do we make sure to catch such errors? Or any breaking change to a dependent resource will be hard to catch (e.g. Python 3.9 reaches EOL and is dropped by Lambda, a change in the crhelper breaks the integration, the policy substack renames one of its outputs or adds a parameter)

cloudformation/custom_resource/cluster.yaml
charlesg3: The delete poll is never supposed to return anything (the return value isn't checked) -- I removed the `True` to match this.

cloudformation/custom_resource/cluster.yaml
charlesg3: All of the CLI logs are output in cloudwatch and in the event of errors the strings provided propagate to the CFN stack (I had to do a lot of debugging to handle the various conditions). However your point is still valid so I added some more logging for the various events to assist further.

cloudformation/tests/test_cluster_custom_resource.py
charlesg3: Agree -- this is being tracked in our backlog and should be addressed.

cloudformation/custom_resource/cluster.yaml
charlesg3: Good question -- we need to return a value here to signify that polling is complete. I added a comment to signify this and change it to return `cluster_name` so that it would be the same type as the other returns.

cloudformation/custom_resource/cluster.yaml
charlesg3: I thought this was needed but double checked and it isn't, removed.

charlesg3:
@demartinofra  -- comments addressed and local tests re-run successfully.

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,231 @@
...
+
+  PclusterCfnFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      FunctionName: !Sub
+        - pcluster-cfn-${StackIdSuffix}
+        - { StackIdSuffix: !Select [2, !Split ['/', !Ref 'AWS::StackId']] }
+      TracingConfig:
+        Mode: Active
+      MemorySize: 1024
```
eantonin: question, non-blocking: We set this to 2048 for the API to reduce the impact of "heavy" library imports like CDK, do you think this make sense also here? 

At the end this is just a lambda doing the same work as the API or am I wrong?

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,231 @@
...
+      Timeout: 60
+      Code:
+        ZipFile: |
+          import datetime
```
eantonin, line 106: just food for thought? Should we add this code in a separate file/artifact, do you think we may benefit from that?

It's just smelling that we see it there, but I don't see strong reason to move it

cloudformation/custom_resource/cluster.yaml
charlesg3: Yes that's correct -- in the case of the API we are concerned about fitting in the 30s timeout of the API gateway, however in this case we are just running a lambda from CFN so we don't have the same timeout constraints. I can update to `2048` to match for consistency.

cloudformation/custom_resource/cluster.yaml
charlesg3: There are some interesting trade-offs there. It's a bit more convenient from an operational perspective to have this file self-contained. Certainly if there were more complexity or expected to be growth of the interface or churn on the logic I'd see more value of moving it out, but that's not a 1-way door and something we can consider doing if needed.

Personally I haven't found it too difficult to modify the code accordingly (really the logic rather than the amount of code is the challenge here) and don't see the additional complexity to be worth it (yet).

cloudformation/custom_resource/cluster.yaml
```diff
@@ -0,0 +1,231 @@
...
+        - PCLayer-${StackIdSuffix}
+        - { StackIdSuffix: !Select [2, !Split ['/', !Ref 'AWS::StackId']] }
+      Description: Library which contains aws-parallelcluster python package and dependencies
+      Content:
+        S3Bucket: !Sub ${AWS::Region}-aws-parallelcluster
+        S3Key: !Sub
+        - parallelcluster/${Version}/layers/aws-parallelcluster/lambda-layer.zip
+        - { Version: !FindInMap [ParallelCluster, Constants, Version] }
+      CompatibleRuntimes:
+        - python3.9
```
eantonin, line 41: question, non-blocking: is this actually the only compatible runtime? Afaik 3.10 should work too
I guess we are using 3.9 here as we define the lambda with 3.9 right?

cloudformation/tests/test_cluster_custom_resource.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def _delete_cluster(cluster_name):
+    """Late bind library and delete cluster."""
+    import pcluster.lib as pcluster
+
+    return pcluster.delete_cluster(cluster_name=cluster_name)
+
+
+def _random_id():
```
eantonin: nitpick, non-blocking, why aren't we using `_random_str()` from `conftest.py` again?

eantonin (dismissed):
LGTM, but I think this may require a changelog update, don't you?

cloudformation/custom_resource/cluster.yaml
charlesg3: It should be compatible with 3.10 AFAIK.

charlesg3:
> LGTM, but I think this may require a changelog update, don't you?

Yes -- I added an entry to `CHANGELOG.md`. :+1:

eantonin (approved):
LGTM

PR #5045, opened 2023-03-10
Fix bug in ImageBuilder resources tagging

jdeamicis:
### Description of changes
* Set the build tags also in the resources created by ImageBuilder.

### Tests
* Modified imagebuilder_stack unit tests accordingly
* Added assertion in `test_build_image()` and ran the integration test to verify that the new assertion works as expected.

### References
Fixes known issue: https://github.com/aws/aws-parallelcluster/issues/5089

### Checklist

### DIFF
--- a/cli/tests/pcluster/templates/test_imagebuilder_stack.py
+++ b/cli/tests/pcluster/templates/test_imagebuilder_stack.py
@@ -2580,6 +2580,9 @@ def test_imagebuilder_build_tags(mocker, resource, response, expected_imagebuild
         else:
             assert_that(resource.get("Properties").get("Tags")).is_equal_to(expected_imagebuilder_resource_tags)
 
+        if resource_name == "InfrastructureConfiguration":
+            assert_that(resource.get("Properties").get("ResourceTags")).is_equal_to(expected_imagebuilder_resource_tags)
+
 
 @pytest.mark.parametrize(
     "resource, response, expected_imagebuilder_subnet_id",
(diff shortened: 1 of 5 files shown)


cli/tests/pcluster/templates/test_imagebuilder_stack.py
```diff
@@ -2577,6 +2577,9 @@ def test_imagebuilder_build_tags(mocker, resource, response, expected_imagebuild
             or resource_name == "BuildNotificationTopic"
         ):
             assert_that(resource.get("Properties").get("Tags")).is_equal_to(expected_role_tags)
+        elif resource_name == "InfrastructureConfiguration":
```
lukeseawalker: instead of adding this elif, you can add an if separated, so that the check on "Tags" isn't repeated

jdeamicis: Addressing in the next push.

tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -141,6 +142,7 @@ def test_build_image(
     )
 
     _test_build_image_success(image)
+    _test_build_instances_tags(image, image.config["Build"]["Tags"], region)
```
lukeseawalker, line 145: I think this will be a noop, since the previous `_test_build_image_success` waits for BUILD_COMPLETE, there are no more instances at this execution point

lukeseawalker: unless the describe_instances is also returning the just terminated instances

jdeamicis: `describe_instances` *may* return recently terminated instances (see [API documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html)).

Currently in the test we are already running a very similar check on IMDSv2, and it relies on a describe_instances of the very same ImageBuilder instances that are terminated at the end of the build-image process. That test currently works, and also the new check I just introduced works for the moment.

lukeseawalker (approved):

PR #5058, opened 2023-03-13
Execute LaunchTemplateValidator only for the first queue

enrico-usai:
### Description of changes
* To speed up config validation we are now executing the run-instances dryrun with the queue LaunchTemplate only for the first queue.

### Tests
* Tested cluster-creation with a policy requesting encryption of the root volume and explicitly setting false to the second queue. Without the patch the creation was failing, with the patch the creation can proceed.


eantonin (approved):
LGTM

PR #5059, opened 2023-03-13
[IntegTests] Fix Multi-AZ create/update integration test failures

EddyMM:
### Description of changes
* The MultiAZ create/update tests rely on creation of capacity reservations in different AZs to ensure instances are launched in different subnets/AZs. One of public AZs created by the `vpc_stack` fixture may be chosen randomly, making it unstable for use in these tests.

* These changes ensure the MultiAZ tests use AZs chosen from a consistent/deterministic list of subnets/AZs.

### Tests
* Ran test locally after reproducing the issue

```
test-suites:
  update:
    test_update.py::test_multi_az_create_and_update:
      dimensions:
      - oss:
        - alinux2
        regions:
        - eu-west-2
        schedulers:
        - slurm
```

```
{
    "all": {
        "total": 1,
        "skipped": 0,
        "failures": 0,
        "errors": 0,
        "succeeded": 1
    },
```

### References
* N/A

### Checklist

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -1180,20 +1180,20 @@ def odcr_stack(request, region, placement_group_stack, cfn_stacks_factory, vpc_s
     odcr_template.set_version()
     odcr_template.set_description("ODCR stack to test open, targeted, and PG ODCRs")
     public_subnet = vpc_stack.get_public_subnet()
-    public_subnets = vpc_stack.get_all_public_subnets().copy()
-    public_subnets.remove(public_subnet)
-    availability_zone = boto3.resource("ec2").Subnet(public_subnet).availability_zone
-    availability_zone_2 = boto3.resource("ec2").Subnet(public_subnets[0]).availability_zone
+    public_subnets = vpc_stack.get_all_public_subnets()
+    default_public_az = boto3.resource("ec2").Subnet(public_subnet).availability_zone
+    availability_zone_1 = boto3.resource("ec2").Subnet(public_subnets[0]).availability_zone
+    availability_zone_2 = boto3.resource("ec2").Subnet(public_subnets[1]).availability_zone
     open_odcr = ec2.CapacityReservation(
         "integTestsOpenOdcr",
-        AvailabilityZone=availability_zone,
+        AvailabilityZone=default_public_az,
         InstanceCount=4,
         InstancePlatform="Linux/UNIX",
         InstanceType="m5.2xlarge",
     )
     target_odcr = ec2.CapacityReservation(
         "integTestsTargetOdcr",
-        AvailabilityZone=availability_zone,
+        AvailabilityZone=default_public_az,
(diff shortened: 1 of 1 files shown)


tests/integration-tests/conftest.py
```diff
@@ -1182,18 +1182,19 @@ def odcr_stack(request, region, placement_group_stack, cfn_stacks_factory, vpc_s
     public_subnet = vpc_stack.get_public_subnet()
     public_subnets = vpc_stack.get_all_public_subnets().copy()
     public_subnets.remove(public_subnet)
-    availability_zone = boto3.resource("ec2").Subnet(public_subnet).availability_zone
-    availability_zone_2 = boto3.resource("ec2").Subnet(public_subnets[0]).availability_zone
+    default_or_random_public_az = boto3.resource("ec2").Subnet(public_subnet).availability_zone
```
NSsirena: I'd call_this `default_public_az` since it corresponds to what the test will receive when asking for the `default_public_az` . 
It is randomized when vpc_stack is created (at each test initialization) but it will remain constant for the entire test lifecycle.

tests/integration-tests/conftest.py
```diff
@@ -1182,18 +1182,19 @@ def odcr_stack(request, region, placement_group_stack, cfn_stacks_factory, vpc_s
...
+    availability_zone_1 = boto3.resource("ec2").Subnet(public_subnets[0]).availability_zone
+    availability_zone_2 = boto3.resource("ec2").Subnet(public_subnets[1]).availability_zone
```
NSsirena, line 1186: [Nitpick] This whole method and how these resources are created and used should likely be redesigned. But this is out-of-scope for this issue.

tests/integration-tests/conftest.py
```diff
@@ -1182,18 +1182,19 @@ def odcr_stack(request, region, placement_group_stack, cfn_stacks_factory, vpc_s
     public_subnet = vpc_stack.get_public_subnet()
     public_subnets = vpc_stack.get_all_public_subnets().copy()
     public_subnets.remove(public_subnet)
```
NSsirena: We shouldn't do this.

The ODCR stack using the `default_az` is used by single_az tests;
while `public_subnets[0]` and `public_subnets[1]` are used in the MultiAZODCR that is used by multi_az tests.

So it is important that what we here consider `public_subnets[0]` is the same seen by the test while requesting `{{ public_subnets[0] }}`

EddyMM: #4898 
Seems like it was a fix applied for a problem encountered making improvements in the test_runner. It appears to ensure that `public_subnets` contains non-default AZs (but the exact reason for doing this is unclear).

NSsirena: Likely it could create failures in the code below where we initialize `multi_az_odcr_group` when the two values pointed to the same AZ

NSsirena (approved):

PR #4552, opened 2022-11-16
3.x Add IAM Policy to allow head node to retrieve compute node console output.

davprat:
### Description of changes
* Changes required for retrieving console output from failed static compute nodes.
   * Add IAM Policy to allow head node to retrieve compute node console output.
   * Add permission to parallelcluster-slurm-head-node policy to allow head node to use batch-get-item DynamoDB call.

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [x] Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5074, opened 2023-03-14
[Test] Add permission dynamodb:BatchGetItem to cluster-roles.cfn.yaml

gmarciani:
### Description of changes
Add permission `dynamodb:BatchGetItem` to `cluster-roles.cfn.yaml`, which is required to retrieve compute node console output. (#4552 )


davprat (approved):
lgtm

PR #5079, opened 2023-03-15
[Slurm] Add CustomSlurmSettings parameter for queue and compute resource

NSsirena:
### Description of changes
* Add optional parameter `CustomSlurmSettings` for queue and compute resource
* Add update policy for `CustomSlurmSettings` parameter
* Add validator for CustomSlurmSettings
* Add Unit Tests for CustomSlurmSettings

### Tests
* local dryrun with config 
```
  SlurmQueues:
  - Name: queue1
    CustomSlurmSettings:
      Nodes: Node1
      PartitionName: CustomPartition
    ComputeResources:
    - Name: t2micro
      CustomSlurmSettings:
        NodeName: CustomNodeName
        State: CustomState
```

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1935,9 +1935,14 @@ class FlexibleInstanceType(Resource):
 class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with Multiple Instance Types."""
 
-    def __init__(self, instances: List[FlexibleInstanceType], **kwargs):
+    def __init__(self,
+                 instances: List[FlexibleInstanceType],
+                 custom_slurm_settings: Dict = None,
+                 **kwargs,
+    ):
         super().__init__(**kwargs)
         self.instances = Resource.init_param(instances)
+        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
 
     @property
     def instance_types(self) -> List[str]:
@@ -1972,10 +1977,15 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
 class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
-    def __init__(self, instance_type, **kwargs):
+    def __init__(self,
+                 instance_type,
+                 custom_slurm_settings: Dict = None,
+                 **kwargs,
+    ):
         super().__init__(**kwargs)
         self.instance_type = Resource.init_param(instance_type)
         self.__instance_type_info = None
+        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
 
     @property
     def instance_types(self) -> List[str]:
@@ -2161,9 +2171,11 @@ class SlurmQueue(_CommonQueue):
     def __init__(
         self,
         allocation_strategy: str = None,
+        custom_slurm_settings: Dict = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
+        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
         if any(
             isinstance(compute_resource, SlurmFlexibleComputeResource) for compute_resource in self.compute_resources
         ):
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -191,6 +191,13 @@ from pcluster.validators.scheduler_plugin_validators import (
     SupportedVersionsValidator,
     UserNameValidator,
 )
+from pcluster.validators.slurm_settings_validator import (
+    SLURM_SETTINGS_COMPUTE_RESOURCE_DENY_LIST,
+    SLURM_SETTINGS_QUEUE_DENY_LIST,
+    SlurmCustomSettingLevel,
+    SlurmCustomSettingsValidator,
+    SlurmCustomSettingsWarning,
+)
 
 LOGGER = logging.getLogger(__name__)
 
@@ -1858,6 +1865,7 @@ class _BaseSlurmComputeResource(BaseComputeResource):
         schedulable_memory: int = None,
         capacity_reservation_target: CapacityReservationTarget = None,
         networking: SlurmComputeResourceNetworking = None,
+        custom_slurm_settings: Dict = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -1873,6 +1881,7 @@ class _BaseSlurmComputeResource(BaseComputeResource):
         self._instance_types_with_instance_storage = []
         self._instance_type_info_map = {}
         self.networking = networking or SlurmComputeResourceNetworking(implied=True)
+        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
 
     @staticmethod
     def fetch_instance_type_info(instance_type) -> InstanceTypeInfo:
@@ -1935,14 +1944,13 @@ class FlexibleInstanceType(Resource):
 class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with Multiple Instance Types."""
 
-    def __init__(self,
-                 instances: List[FlexibleInstanceType],
-                 custom_slurm_settings: Dict = None,
-                 **kwargs,
+    def __init__(
+        self,
+        instances: List[FlexibleInstanceType],
+        **kwargs,
     ):
         super().__init__(**kwargs)
         self.instances = Resource.init_param(instances)
-        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
 
     @property
     def instance_types(self) -> List[str]:
@@ -1977,15 +1985,14 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
 class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
-    def __init__(self,
-                 instance_type,
-                 custom_slurm_settings: Dict = None,
-                 **kwargs,
+    def __init__(
+        self,
+        instance_type,
+        **kwargs,
     ):
         super().__init__(**kwargs)
         self.instance_type = Resource.init_param(instance_type)
         self.__instance_type_info = None
-        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
 
     @property
     def instance_types(self) -> List[str]:
@@ -2226,6 +2233,14 @@ class SlurmQueue(_CommonQueue):
                 queue_name=self.name,
                 subnet_ids=self.networking.subnet_ids,
             )
+        if self.custom_slurm_settings:
+            self._register_validator(
+                SlurmCustomSettingsValidator,
+                custom_settings=self.custom_slurm_settings,
+                deny_list=SLURM_SETTINGS_QUEUE_DENY_LIST,
+                settings_level=SlurmCustomSettingLevel.QUEUE,
+            )
+            self._register_validator(SlurmCustomSettingsWarning)
         for compute_resource in self.compute_resources:
             self._register_validator(
                 EfaSecurityGroupValidator,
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jdeamicis: I wonder if we have to use 2023 now?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -191,6 +191,13 @@
     SupportedVersionsValidator,
     UserNameValidator,
 )
+from pcluster.validators.slurm_settings_validator import (
+    SLURM_SETTINGS_COMPUTE_RESOURCE_DENY_LIST,
+    SLURM_SETTINGS_QUEUE_DENY_LIST,
```
jdeamicis: Should we use `CUSTOM_SLURM_SETTINGS_*` instead of `SLURM_SETTINGS_*`?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1935,7 +1944,11 @@ def __init__(self, instance_type: str, **kwargs):
 class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with Multiple Instance Types."""
 
-    def __init__(self, instances: List[FlexibleInstanceType], **kwargs):
+    def __init__(
+        self,
+        instances: List[FlexibleInstanceType],
+        **kwargs,
+    ):
```
jdeamicis, line 1950: Nitpick: should we revert this change, since it doesn't do anything?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1972,7 +1985,11 @@ def max_network_interface_count(self) -> int:
 class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
-    def __init__(self, instance_type, **kwargs):
+    def __init__(
+        self,
+        instance_type,
+        **kwargs,
+    ):
```
jdeamicis, line 1991: Nitpick: should we revert this change, since it doesn't do anything?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2161,9 +2178,11 @@ class SlurmQueue(_CommonQueue):
     def __init__(
         self,
         allocation_strategy: str = None,
+        custom_slurm_settings: Dict = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
+        self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
```
jdeamicis, line 2184: I'm wondering when we need `Resource.init_param()` and when not...

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2214,6 +2233,14 @@ def _register_validators(self, context: ValidatorContext = None):
                 queue_name=self.name,
                 subnet_ids=self.networking.subnet_ids,
             )
+        if bool(self.custom_slurm_settings):  # if not empty register validator
```
jdeamicis: Nitpick: `bool()` should not be necessary here.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2231,6 +2258,14 @@ def _register_validators(self, context: ValidatorContext = None):
                 is False,
                 multi_az_enabled=self.multi_az_enabled,
             )
+            if bool(compute_resource.custom_slurm_settings):  # if not empty register validator
```
jdeamicis: Same as above.

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -0,0 +1,69 @@
...
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+
+from pcluster.validators.common import FailureLevel, Validator
+
+SLURM_SETTINGS_QUEUE_DENY_LIST = ["Nodes", "State", "PartitionName", "ResumeTimeout", "SuspendTime"]
+SLURM_SETTINGS_COMPUTE_RESOURCE_DENY_LIST = ["NodeName", "NodeHostName", "NodeAddr", "State", "Gres"]
```
jdeamicis: We have to double-check if the parameters in the `slurm.conf` are case-sensitive. If they aren't, we may have to do this validation case-insensitive.

jdeamicis: Second paragraph from the [`slurm.conf` documentation](https://slurm.schedmd.com/slurm.conf.html#SECTION_DESCRIPTION):

> The contents of the file are case insensitive except for the names of nodes and partitions.

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class SlurmCustomSettingLevel(str, Enum):
```
jdeamicis, line 23: What is the `str` here used for?

jdeamicis: So the object in the class inherit from both Enums and strings?

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -0,0 +1,69 @@
...
+    def _validate(self, custom_settings, deny_list, settings_level: SlurmCustomSettingLevel):
+        denied_settings = set()
+
+        for setting in deny_list:
+            if setting in custom_settings.keys():
+                denied_settings.add(setting)
+        if len(denied_settings) > 0:
+            settings = ",".join(sorted(denied_settings))
+            self._add_failure(
+                f"Setting or overriding the following custom Slurm settings at {settings_level} level is not allowed: "
```
jdeamicis: I would word it "Using the following custom Slurm settings...".

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -0,0 +1,69 @@
...
+
+    QUEUE = "Queue"
+    COMPUTE_RESOURCE = "ComputeResource"
+
+
+class SlurmCustomSettingsValidator(Validator):
+    """
+    Slurm Custom Settings validator.
+
+    Validate custom settings in Slurm ComputeResource and queue.
```
jdeamicis: Nitpick: "Queue" instead of "queue".

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -187,6 +188,45 @@ def test_cluster_name_validator_slurm_accounting(cluster_name, scheduling, shoul
...
+            SlurmCustomSettingLevel.COMPUTE_RESOURCE,
+            "",
+        ),
+        (
+            "Fails when custom settings are in the deny_list, invalid parameters are reported",
+            {"Denied1": "Value1", "Denied2": "Value2"},
+            ["Denied1", "Denied2"],
+            SlurmCustomSettingLevel.COMPUTE_RESOURCE,
+            "Setting or overriding the following custom Slurm settings at"
+            " ComputeResource level is not allowed: Denied1,Denied2",
```
jdeamicis: For better clarity, please put the full sentence here.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -187,6 +188,45 @@ def test_cluster_name_validator_slurm_accounting(cluster_name, scheduling, shoul
     assert_failure_messages(actual_failures, expected_message)
 
 
+@pytest.mark.parametrize(
```
jdeamicis, line 195: I'm not seeing a unit test on the generic warning. Could we add one?

jdeamicis (changes requested):
Good job. We need to sort out the case-insensitivity of the Slurm configuration files.

I left some minor comments around the code.

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Yes. 
In python 3.11 there is an StrEnum and you can use that.

Here to use a string as enum value we have to inherit from both.

cli/src/pcluster/config/cluster_config.py
NSsirena: They are `SLURM_SETTINGS` meaning that they are defined by Slurm and we are `denying` their use.

Using here `CUSTOM` gives the impression that they are defined by the user, like a tag or something.

cli/src/pcluster/config/cluster_config.py
NSsirena: I thought about it, but then I decided to keep them.
Having the same `init style` across all resources make the code more readable.

cli/src/pcluster/config/cluster_config.py
NSsirena: same as above.

cli/src/pcluster/config/cluster_config.py
NSsirena: Not 100% sure, my understanding is that we use it when we want to have a default value and avoid `None`

cli/src/pcluster/config/cluster_config.py
NSsirena: It is.
I had to add it ..without it doesn't work with empty dictionaries.

cli/src/pcluster/config/cluster_config.py
NSsirena: same

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Good catch!
I'll update the check!

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: I picked the working from the design doc, but I can update it if you think this version is better.
In this case I'd update also the doc.

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: Not sure I got your point.
It is the full sentence, I just had to split it in two lines to make the code-linter happy

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: It is a different validator ..not sure what to test about it apart from its presence/absence.
I'll try to figure out a way to test it.

cli/src/pcluster/config/cluster_config.py
jdeamicis: I'm fine with that.

Just had an idea: can we have one single `SLURM_SETTINGS_DENY_LIST` dictionary so that we can easily extend the structure in the future?

So we will have something like:

```
SLURM_SETTINGS_DENY_LIST = {
    "Queue": [ ... ],
    "ComputeResource": [ ... ],
    "SlurmConf": [ ... ],
    ...
}
```

Please feel free to propose improvements.

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Fixed.
I also added a specific unit test.

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Fixed

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Fixed

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: I added a test for the Warning

cli/tests/pcluster/validators/test_cluster_validators.py
NSsirena: With the rewording fits into one line, so it should be easier to check that it is the full sentence.

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Changed to 2023

cli/src/pcluster/config/cluster_config.py
NSsirena: Done. 
I've also added new values in the deny lists.

jdeamicis (dismissed):

cli/src/pcluster/config/cluster_config.py
jdeamicis: That's weird. I'm not getting this behavior locally:
```
In [1]: A = {}

In [2]: if A:
   ...:     print("yes")
   ...: 

In [3]: B = {"bla": "blu"}

In [4]: if B:
   ...:     print("yes")
   ...: 
yes

```

jdeamicis (approved):

PR #5083, opened 2023-03-16
[Test] Adaptations of integ tests for US isolated regions + minor fixes

gmarciani:
### Description of changes
1. Add log lines to have more info during test execution in tests: test_queue_parameters_update and test_create_imds_secured.
1. Fix import in metrics_reporter.
1. Update test config to use m5d.xlarge in 'test_head_node_stop'.
1. Increase sleep time waiting for head node reboot from 120 to 240 seconds when running in US isolated regions.
1. Avoid using flexible instance type in test_efa when running in US isolated regions as the instance types used by this test are not supported by ec2:CreateFleet API.
1. Use ONDEMAND capacity instead of SPOT when in US isolated regions in test 'test_update_slurm'.

### Tests
* Changes have been validated in us-isob-east-1.

### DIFF
--- a/tests/integration-tests/benchmarks/common/metrics_reporter.py
+++ b/tests/integration-tests/benchmarks/common/metrics_reporter.py
@@ -17,7 +17,7 @@ from time import sleep
 import boto3
 from retrying import RetryError, retry
 from time_utils import seconds
-from utils import _describe_cluster_instances
+from utils import describe_cluster_instances
 
 METRIC_WIDGET_TEMPLATE = """
     {{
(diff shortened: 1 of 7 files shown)


tests/integration-tests/tests/common/utils.py
```diff
@@ -239,7 +239,9 @@ def reboot_head_node(cluster, remote_command_executor=None):
     logging.info(f"result.failed={result.failed}")
     logging.info(f"result.stdout={result.stdout}")
     wait_head_node_running(cluster)
-    time.sleep(120)  # Wait time is required for the head node to complete the reboot
+    # Wait time is required for the head node to complete the reboot.
+    # We observed that headnode in US isolated regions may take more time to reboot.
+    time.sleep(240 if "us-iso" in cluster.region else 120)
```
dreambeyondorange, line 244: should we just make it 240 for all?  WE've got a lot of us-iso caveats creeping into the codebase

dreambeyondorange (dismissed):

tests/integration-tests/tests/common/utils.py
gmarciani: I agree that us-iso are introducing a lot of different cases.
Nevertheless, I think that the peculiarities of us-iso regions should not have an impact on the rest of the regions we support. Especially if this impact is relaxing constraints in integration tests.

tests/integration-tests/benchmarks/common/metrics_reporter.py
```diff
@@ -17,7 +17,7 @@
 import boto3
 from retrying import RetryError, retry
 from time_utils import seconds
-from utils import _describe_cluster_instances
+from utils import describe_cluster_instances
```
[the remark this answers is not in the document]
gmarciani: Need to fix this

dreambeyondorange (approved):

PR #5091, opened 2023-03-17
Add default cluster tags and user's tags to LaunchTemplate CFN resources

enrico-usai:
### Description of changes
* TagSpecificationProperty specifies the tags to apply to the launch template during creation. To tag the launch template, ResourceType must be launch-template.
* This is useful for users where there are strict policies requiring all the resources to be tagged or to limit resource creations to specific tags

### Tests
* Tested manual creation of a cluster with a modified CLI.
* Verified the tags are present in both head node and compute launch templates.

### Checklist


enrico-usai:
This change might be an issue for the update because if we attach custom tags to the launch template an update of the tags will trigger an update of the instances (and this is an issue in particular for the head node).

PR #5129, opened 2023-03-27
Merge commits related to release 3.5.1

gmarciani:
### Description of changes
Merge commits related to release 3.5.1, taken from the release branch `release-3.5` into `develop`.

### Tests
1. Changes from the release branch have been tested in `release-3.5`.
2. Integ tests passed: test_mpi, test_awsbatch, test_slurm


tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -31,26 +31,35 @@
...
 
     bastion = vpc_stack.cfn_outputs["BastionUser"] + "@" + vpc_stack.cfn_outputs["BastionIP"]
     assert_that(len(get_compute_nodes_instance_ids(cluster.cfn_name, region))).is_equal_to(1)
     remote_command_executor = RemoteCommandExecutor(cluster, bastion=bastion)
     scheduler_commands = scheduler_commands_factory(remote_command_executor)
-    _test_fsx_in_private_subnet(cluster, region, fsx_mount_dir, remote_command_executor, scheduler_commands)
-
+    _test_shared_storage_in_private_subnet(
+        cluster, region, storage_type, mount_dir, bastion_instance, remote_command_executor, scheduler_commands
+    )
```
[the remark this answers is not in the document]
gmarciani: Fixing...

tests/integration-tests/tests/networking/test_cluster_networking.py
```diff
@@ -31,26 +31,35 @@
 )
 from tests.common.osu_common import compile_osu
 from tests.common.schedulers_common import SlurmCommands
-from tests.common.utils import get_default_vpc_security_group
-from tests.storage.storage_common import assert_fsx_correctly_shared, assert_fsx_lustre_correctly_mounted, get_fsx_ids
+from tests.common.storage.constants import StorageType
+from tests.common.utils import get_default_vpc_security_group, get_route_tables, retrieve_latest_ami
```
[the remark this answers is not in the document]
gmarciani: Fixing..

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -155,7 +166,7 @@ write_files:
           # 60s, the call will timeout the connect attempt at 8m. Setting it to 15s, causes
           # each attempt to take 240s, so 2m * 3 attempts will result in a failure after 6
           # minutes.
-          S3API_RESULT=$(AWS_RETRY_MODE=standard aws s3api get-bucket-location --cli-connect-timeout 15 --bucket ${!BASH_REMATCH[1]} 2>&1) || error_exit "${!S3API_RESULT}"
+          S3API_RESULT=$(AWS_RETRY_MODE=standard aws s3api get-bucket-location --cli-connect-timeout 15 --bucket ${!BASH_REMATCH[1]} --region ${AWS::Region} 2>&1) || error_exit "${!S3API_RESULT}"
```
dreambeyondorange, line 169: Was this not working without the region flag in the test region?

CHANGELOG.md
```diff
@@ -20,6 +20,30 @@ CHANGELOG
 - Fix missing tagging of resources created by ImageBuilder during the `build-image` operation.
 - Fix Update policy for MaxCount to always perform numerical comparisons on MaxCount property.
 
+3.5.1
+-----
+**ENHANCEMENTS**
+- Add a new way to distribute ParallelCluster as a self-contained executable shipped with a dedicated installer.
+- Add support for US isolated region us-isob-east-1.
```
dreambeyondorange, line 28: Does this only support east-1 or will it support any isob region?  Or is it more because we were only able to test in one region?

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -73,13 +75,28 @@ def _validate(self, domain_name):
...
+        In particular, the ARN should be one of the following resources:
+         1. a readable secret in AWS Secrets Manager, which is supported in all regions but us-isob-east-1.
+         2. a readable parameter in SSM Parameter Store, which is supported only in us-isob-east-1.
+        """
         try:
             # We only require the secret to exist; we do not validate its content.
-            AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
+            arn_components = Arn.parse(password_secret_arn)
+            service, resource = arn_components.service, arn_components.resource
+            if service == "secretsmanager" and resource == "secret" and region != "us-isob-east-1":
```
dreambeyondorange, line 89: It seems that this block could provide more specific information for certain situations.  If the user is trying to use secretsmanager in the isob region, we could produce an error to inform them of that instead of a generic unsupported message.

dreambeyondorange (dismissed):

CHANGELOG.md
gmarciani: In 3.5.1 we will support only us-isob-east-1.

cli/src/pcluster/resources/compute_node/user_data.sh
gmarciani: In order to work in US-ISOB partition, the `AWS_DEFAULT_REGION` should be set system wide to `us-isob-east-1`.
At the time we developed this logic we did not set this system-wide environment variable and we ended up validating this solution.
Currently we set the region within `cloud-boothook` so I think it's redundant to specify the region with the `--region` option and the option may then be removed.

Since every change involving US-ISOB requires time consuming validation, I suggest to move forward with this validated change and remove the redundant option in a follow up activity.

cli/src/pcluster/validators/directory_service_validators.py
gmarciani: May you please provide a specific example? What additional infornmation are you thinking about?

dreambeyondorange: For example you could do something like

if us-isob and secret manager print an error saying those two are incompatible since we know why the chosen params are invalid but not really explaining why to users

gmarciani: You mean: saying to the customer that the secrets is not supported because SecretsManager is not supported in us-isob-east-1?
In this case, I think we can, but still I suggest to address this in a follow up activity.

enrico-usai (approved):

PR #5165, opened 2023-04-07
[3.x] update test_iam_image.py to poll for the image resource to exist

dreambeyondorange:
update test_iam_image.py to poll for the image resource to exist (usually a couple minutes) instead of a blind sleep to avoid issues with data availability during the check_roles call

### Description of changes
* Changed the integration test test_iam_roles to check for the image resource instead of sleeping

### Tests
* Ran integration test locally

### Checklist


tests/integration-tests/tests/iam/test_iam_image.py
```diff
@@ -88,6 +87,20 @@ def _wait_build_image_complete(image):
...
+@retry(wait_fixed=minutes(1), stop_max_delay=minutes(60))
+def _get_resources_with_image_resource(cfn_client, stack_name):
+    logging.info("Describe stack resources")
+    resources = cfn_client.describe_stack_resources(StackName=stack_name)["StackResources"]
+    image_resource_exists = False
+    logging.info("Checking image resource")
+    for resource in resources:
+        if resource["ResourceType"] == "AWS::ImageBuilder::Image":
+            image_resource_exists = True
+            logging.info("The image resource exists!")
```
enrico-usai, line 98: once you found the resource we can break the loop

tests/integration-tests/tests/iam/test_iam_image.py
```diff
@@ -88,6 +87,20 @@ def _wait_build_image_complete(image):
     assert_that(image.image_status).is_equal_to("BUILD_COMPLETE")
 
 
+@retry(wait_fixed=minutes(1), stop_max_delay=minutes(60))
```
enrico-usai: I'd reduce the `stop_max_delay` to 20-25 minutes, since 10 was enough most of the times.

tests/integration-tests/tests/iam/test_iam_image.py
```diff
@@ -106,9 +119,11 @@ def _check_roles(
             # Check the role is attached to the Lambda function
             lambda_function = lambda_client.get_function(FunctionName=resource["PhysicalResourceId"])["Configuration"]
             assert_that(lambda_function["Role"]).is_equal_to(lambda_cleanup_role)
+            logging.info("Lambda function role confirmed")
         if resource_type == "AWS::ImageBuilder::Image":
```
enrico-usai, line 124: We can move this entire logic into the `_get_resources_with_image_resource` renaming it to `_check_instance_profile_arn`.

I mean we're doing the `_get_resources_with_image_resource` to retrieve the image arn, then we're using in the check outside, it's better to put together the logic to wait for it and to assert for the policy, since the lambda assertion is a totally separate check.

tests/integration-tests/tests/iam/test_iam_image.py
dreambeyondorange: I can do that.  It usually succeeds on the second or third retry in my local testing

tests/integration-tests/tests/iam/test_iam_image.py
dreambeyondorange: doing it this way will ensure the lambda function resource is also there.  It is possible that neither exists.  I was actually considering also checking for the lambda resource, but the image resource will take longer to start creation

dreambeyondorange: We are also retrieving the resources on line 114, if I move the wait condition to the if block for the image resource, I will need to get the resources in two separate places

enrico-usai: ok let's keep it as is! Thanks for the clarification.

enrico-usai (approved):

PR #5169, opened 2023-04-07
Increase limit on number of compute resources per queue

EddyMM:
### Description of changes
* The maximum number of compute resources at the moment is 5 per queue. With the use of nested stacks, the CloudFormation templates can accommodate more resources before reaching the [limits](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html).
* These changes introduce the possibility of using multiple compute resources into a queue (as long as the cluster remains below the maximum compute resource per cluster limit). 
* To ensure stack deployments are traceable (i.e. each compute resource can be traced back to the queue stack), each compute resource should be placed in the same nested stack as its own queue, therefore there is still need to have a maximum number of compute resources per queue.
* An extra improvement as well is the possibility to have more queues in a cluster as long as the number of compute resources per cluster is within limits
* Major limits introduced in this PR:
```
* Maximum compute resources per queue: 40  # Ensures that each queue will share the same stack as its compute resources
* Maximum compute resources per cluster: 150  # Based on API timeout limitations
* Maximum queues per cluster: 100
```
* Previous limits:

```
* Maximum compute resources per queue: 5
* Maximum compute resources per cluster: Not Used
* Maximum queues per cluster: 40
```

* These changes also alter the behaviour of `pcluster configure`. `pcluster configure` uses a sequential approach to construct the cluster configuration file. The process assumes fixed limits on the number of compute resources and queues but does not account for a limit on the number of compute resources allowed per cluster. A workaround for this for now) is to use the new maximum compute resources per queue as a fallback but first calculate the maximum allowed compute resources after the customer has specified how many queues they intend to use. i.e. minimum value between the `MAXIMUM_COMPUTE_RESOURCES_PER_CLUSTER / NO_OF_ QUEUES_SPECIFIED`  and `MAX_COMPUTE_RESOURCES_PER_QUEUE `

### Tests
* Added a unit test to test the distribution of compute resources and queues into nested stacks
* Modified existing tests that check the maximum count of queues and compute resources used in the cluster
* Manually created cluster:
* With 100 queues / 1 compute resource --> Successful
* With 110 queues / 1 compute resource

```
    {
      "level": "ERROR",
      "type": "MaxCountValidator",
      "message": "Invalid number of SlurmQueues (110) specified. Currently only supports up to 100 SlurmQueues."
    }
```

* With 200 queues / 1 compute resource

```
    {
      "level": "ERROR",
      "type": "MaxCountValidator",
      "message": "Invalid number of ComputeResources per Cluster (200) specified. Currently only supports up to 150 ComputeResources per Cluster."
    }
```

* With 1 queue / 41 compute resource

```
    {
      "level": "ERROR",
      "type": "MaxCountValidator",
      "message": "Invalid number of ComputeResources per Queue (41) specified. Currently only supports up to 40 ComputeResources per Queue."
    },
```

### References
* #5099

### Checklist


enrico-usai (approved):

enrico-usai:
The patch looks good to me. Great job!

PR #5170, opened 2023-04-07
Add warning about VPC usage charges in automation

dreambeyondorange:
### Description of changes
* Adding a message to inform users that NAT Gateways incur charges outside of the free tier.

### Tests
* Tested the CLI locally to print the message during configure

### Checklist

### DIFF
--- a/cli/src/pcluster/cli/commands/configure/easyconfig.py
+++ b/cli/src/pcluster/cli/commands/configure/easyconfig.py
@@ -341,7 +341,18 @@ def _create_vpc_parameters(scheduler, head_node_instance_type, compute_instance_
     vpc_list = vpc_and_subnets["vpc_list"]
     if not vpc_list:
         print("There are no VPC for the given region. Starting automatic creation of VPC and subnets...")
-    if not vpc_list or prompt("Automate VPC creation? (y/n)", lambda x: x in ("y", "n"), default_value="n") == "y":
+    if (
+        not vpc_list
+        or prompt(
+            "The creation of the VPC will result in minimal charges for NAT gateway \n"
+            "creation using a public subnet that are not covered under the free tier.\n"
+            "Please refer to https://aws.amazon.com/vpc/pricing/ for more details.\n"
+            "Automate VPC creation? (y/n)",
+            lambda x: x in ("y", "n"),
+            default_value="n",
+        )
+        == "y"
+    ):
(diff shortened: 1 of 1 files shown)


himani2411 (dismissed):

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -341,7 +341,18 @@ def _create_vpc_parameters(scheduler, head_node_instance_type, compute_instance_
     vpc_list = vpc_and_subnets["vpc_list"]
     if not vpc_list:
         print("There are no VPC for the given region. Starting automatic creation of VPC and subnets...")
-    if not vpc_list or prompt("Automate VPC creation? (y/n)", lambda x: x in ("y", "n"), default_value="n") == "y":
+    if (
+        not vpc_list
+        or prompt(
+            "The creation of the VPC will result in minimal charges for NAT gateway \n"
```
davprat, line 347: Would it be better to say `standard` rather than `minimal` since minimal is subjective?

davprat (approved):
lgtm

cli/src/pcluster/cli/commands/configure/easyconfig.py
dreambeyondorange: That could be, but I don't want to alarm people with warning.  A NAT gateway cost ~$1 a day

cli/src/pcluster/cli/commands/configure/easyconfig.py
```diff
@@ -341,7 +341,18 @@ def _create_vpc_parameters(scheduler, head_node_instance_type, compute_instance_
...
+            "creation using a public subnet that are not covered under the free tier.\n"
```
enrico-usai, line 348: The warning message we're adding is about the "pcluster configure" command, so we need to take into consideration only two cases: public and public-private.
The NAT Gateway is created only in the public-private case, see templates used by pcluster configure here: https://github.com/aws/aws-parallelcluster/tree/develop/cloudformation/networking

So the message should be modified removing the "using a public subnet" sentence.


Then the NAT Gateway is not created at VPC creation time but at subnet creation time, so should be moved in the 
``` 
            or (prompt("Automate Subnet creation? (y/n)", lambda x: x in ("y", "n"), default_value="y") == "y")
```
only when selecting the public-private case.

PR #5171, opened 2023-04-07
Implement detailed monitoring configuration

delongmeng:
### Description of changes
* Implemented the "detailed monitoring" configuration under the "Monitoring" section.
* When enabled, the "detailed monitoring" configuration with only apply to compute nodes.
* The update policy is "COMPUTE_FLEET_STOP" for detailed monitoring.
* Change the update policy of "Monitoring" to "IGNORED" since its children layer has a mixture of update policies.
* Add validator for detailed monitoring to show a warning message that the cost might increase.

### Tests
* Manual tests
  * Three potential configurations:
    * not specified (the default is set to false)
    * explicitly set to false
    * explicitly set to true
  * Cluster creation with the above 3 configurations: worked as expected
  * Cluster update switching among these configurations: worked as expected
* Unit tests
  * Update current unit tests accordingly
  * Added unit tests for the new detailed monitoring validator

### Checklist


cli/src/pcluster/templates/queue_group_stack.py
```diff
@@ -271,6 +273,9 @@ def _add_compute_resource_launch_template(
                                 "IntelHPCPlatform": "true" if self._config.is_intel_hpc_platform_enabled else "false",
                                 "CWLoggingEnabled": "true" if self._config.is_cw_logging_enabled else "false",
                                 "LogRotationEnabled": "true" if self._config.is_log_rotation_enabled else "false",
+                                "DetailedMonitoringEnabled": "true"
```
chenwany: Why do we need to pass it to compute nodes userdata? It is only needed as an attribute in  laugh template, not need to pass to cookbook

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1039,6 +1039,7 @@ def _add_head_node(self):
                     "enable_intel_hpc_platform": "true" if self.config.is_intel_hpc_platform_enabled else "false",
                     "cw_logging_enabled": "true" if self.config.is_cw_logging_enabled else "false",
                     "log_rotation_enabled": "true" if self.config.is_log_rotation_enabled else "false",
+                    "detailed_monitoring_enabled": "true" if self.config.is_detailed_monitoring_enabled else "false",
```
chenwany: I think we don't need to pass it into dna.json, see reason in the comment below

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -102,6 +102,7 @@ write_files:
           "enable_intel_hpc_platform": "${IntelHPCPlatform}",
           "cw_logging_enabled": "${CWLoggingEnabled}",
           "log_rotation_enabled": "${LogRotationEnabled}",
+          "detailed_monitoring_enabled": "${DetailedMonitoringEnabled}",
```
chenwany: same comment here

chenwany (commented):
As we discussed before, what about adding a warning validator about the extra cost for enable detail monitoring?

cli/src/pcluster/templates/cluster_stack.py
delongmeng: If I removed this, `test_cluster_stack.py:test_head_node_dna_json` would fail.
I can remove the other ones since no unit tests are testing them (assuming that we don't expect this new parameter in userdata, rather than that we're missing a test).

cli/tests/pcluster/validators/test_monitoring_validators.py
```diff
@@ -10,8 +10,8 @@
 # limitations under the License.
 import pytest
 
-from pcluster.config.cluster_config import CloudWatchLogs, LogRotation, Logs
-from pcluster.validators.monitoring_validators import LogRotationValidator
+from pcluster.config.cluster_config import CloudWatchLogs, LogRotation, Logs, Monitoring
```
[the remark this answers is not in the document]
delongmeng: fixed

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ CHANGELOG
 - Track the longest dynamic node idle time in CloudWatch Dashboard.
 - Add new configuration section `HealthChecks/Gpu` for enabling the GPU Health Check in the compute node before job submission.
 - Add support for `Tags` in the `SlurmQueues` section.
+- Add support for the detailed monitoring configuration for Slurm compute fleet.
```
chenwany: What about: Add support for `EnableDetailedMonitoring` in the `Monitoring` section.

chenwany (commented):
When we add a new parameter, let's include it in 
[slurm.full.all_resources.yaml](https://github.com/aws/aws-parallelcluster/blob/a397014f21f5e514a640676e9581ee1671a8ed03/cli/tests/pcluster/templates/test_cluster_stack/test_cluster_config_limits/slurm.full.all_resources.yaml) file. Once included, you can use the yaml in the test output to update the snapshot ([slurm.full_config.snapshot.yaml](https://github.com/aws/aws-parallelcluster/blob/a397014f21f5e514a640676e9581ee1671a8ed03/cli/tests/pcluster/templates/test_cluster_stack/test_cluster_config_limits/slurm.full_config.snapshot.yaml)).

cli/src/pcluster/templates/cluster_stack.py
delongmeng: The unit test failed because "detailed monitoring" was added to the default dna json file. After removing it, there's no more problem with the unit tests.

CHANGELOG.md
delongmeng: sure. updated.

chenwany (approved):
lgtm!

PR #5172, opened 2023-04-10
Add Tags at compute resource level

chenwany:
### Description of changes
Add support for Tags section under compute resource level.
```
  SlurmQueues:
    - Name: compute
      AllocationStrategy: lowest-price
      ComputeResources:
        - Name: defaultcompute
          Instances:
            - InstanceType: c5.xlarge
          MinCount: 0
          MaxCount: 505
          Tags:
            - Key: mytag1
              Value: tagvalue13
```
The tags section under queue level support queue update strategy
* Add validator to make sure no duplicate tags within the same section
* Add warning to warn compute resource tags will override queue tags, queue tags will override cluster level tags when there are duplicated tag keys.

### Tests
* Manually test
  * Test cluster creation with Tags section in compute resources, 
    * verify duplicated tags in the Compute Resource section override Queue level
    * verify duplicated tags in the Compute Resource section override Cluster level
  * Test cluster update with Tags section in compute resources
  * Test cluster creation/update is blocked when there are duplicated Tags within the same section
* Unit test

### References

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@ CHANGELOG
 - Add support for customizing the cluster Slurm configuration via the ParallelCluster configuration YAML file.
 - Track the longest dynamic node idle time in CloudWatch Dashboard.
 - Add new configuration section `HealthChecks/Gpu` for enabling the GPU Health Check in the compute node before job execution.
-- Add support for `Tags` in the `SlurmQueues` section.
 - Add support for `DetailedMonitoring` in the `Monitoring` section.
+- Add support for `Tags` in the `SlurmQueues` and `SlurmQueues/ComputeResources` section.
 
 **CHANGES**
 - Increase the default `RetentionInDays` of CloudWatch logs from 14 to 180 days.
--- a/cli/src/pcluster/schemas/common_schema.py
+++ b/cli/src/pcluster/schemas/common_schema.py
@@ -65,6 +65,18 @@ def validate_no_reserved_tag(tags):
                 raise ValidationError(message=f"The tag key prefix '{PCLUSTER_PREFIX}' is reserved and cannot be used.")
 
 
+def validate_no_duplicate_tag(tags):
+    """Validate there is no duplicate tag keys in the same tag section."""
+    all_tags = set()
+    for tag in tags:
+        tag_key = tag.key
+        if tag_key in all_tags:
+            raise ValidationError(
+                f"Duplicate tag key ({tag_key}) detected. Tags keys should be unique within the Tags section."
+            )
+        all_tags.add(tag_key)
+
+
 def get_field_validator(field_name):
     allowed_values = ALLOWED_VALUES[field_name]
     return validate.OneOf(allowed_values) if isinstance(allowed_values, list) else validate.Regexp(allowed_values)
--- a/cli/src/pcluster/templates/queue_group_stack.py
+++ b/cli/src/pcluster/templates/queue_group_stack.py
@@ -134,12 +134,12 @@ class QueueGroupStack(NestedStack):
                     self._config.is_detailed_monitoring_enabled,
                 )
 
-    def _get_custom_compute_resource_tags(self, queue_config):
-        """Override Queue Tags value on Cluster level tags if there are duplicated keys."""
+    def _get_custom_compute_resource_tags(self, queue_config, compute_resource_config):
+        """Compute resource tags and Queue Tags value on Cluster level tags if there are duplicated keys."""
         tags = get_custom_tags(self._config, raw_dict=True)
         queue_tags = get_custom_tags(queue_config, raw_dict=True)
-        tags.update(queue_tags)
-        return dict_to_cfn_tags(tags)
+        compute_resource_tags = get_custom_tags(compute_resource_config, raw_dict=True)
+        return dict_to_cfn_tags({**tags, **queue_tags, **compute_resource_tags})
 
     def _add_compute_resource_launch_template(
         self,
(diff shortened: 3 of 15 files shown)


cli/src/pcluster/schemas/common_schema.py
```diff
@@ -65,6 +65,18 @@ def validate_no_reserved_tag(tags):
...
+    """Validate there is no duplicate tag keys in the same tag section."""
+    all_tags = set()
+    for tag in tags:
+        tag_key = tag.key
+        if tag_key in all_tags:
+            raise ValidationError(
+                f"Duplicate tag key ({tag_key}) detected. Tags should not have same tag key with the Tags section."
+            )
+        all_tags.add(tag_key)
+
```
delongmeng, line 78: Would it make sense to print out all duplicated keys instead of only the first one encountered?
In the text message, "within ... section" instead of "with ... section"?

cli/src/pcluster/templates/queue_group_stack.py
```diff
@@ -133,12 +133,12 @@ def _add_launch_templates(self):
...
+    def _get_custom_compute_resource_tags(self, queue_config, compute_resource_config):
         """Override Queue Tags value on Cluster level tags if there are duplicated keys."""
-        tags = get_custom_tags(self._config, raw_dict=True)
-        queue_tags = get_custom_tags(queue_config, raw_dict=True)
-        tags.update(queue_tags)
-        return dict_to_cfn_tags(tags)
+        tags = {**get_custom_tags(self._config, raw_dict=True)}
+        queue_tags = {**get_custom_tags(queue_config, raw_dict=True)}
+        compute_resource_tags = {**get_custom_tags(compute_resource_config, raw_dict=True)}
+        return dict_to_cfn_tags({**tags, **queue_tags, **compute_resource_tags})
```
delongmeng, line 142: Is the function description ("_Override Queue Tags value on Cluster level tags if there are duplicated keys_") still accurate? Now it looks like compute resource tags will also replace the existing keys in cluster tags and queue tags?
BTW, what's the difference between `xx_tags = {**get_custom_tags(...)}` and `xx_tags = get_custom_tags(...)` here given they are the only dictionary whose contents get unpacked there?

cli/src/pcluster/validators/tags_validators.py
```diff
@@ -14,18 +14,51 @@
...
+        overlapping_keys_list = sorted(list(overlapping_keys))
+        queue_cluster_overlapping_keys = sorted(
+            list(set(cluster_tags.keys()) & set(queue_tags.keys()) - overlapping_keys)
+        )
+        compute_resource_queue_overlapping_key = sorted(
+            list(set(queue_tags.keys()) & set(compute_resource_tags.keys()) - overlapping_keys)
+        )
+        cluster_compute_resource_overlapping_key = sorted(
+            list(set(cluster_tags.keys()) & set(compute_resource_tags.keys()) - overlapping_keys)
+        )
```
delongmeng, line 30: The names are a little inconsistent here:
- overlapping_key**s_list**
- queue_cluster_overlapping_key**s**
- compute_resource_queue_overlapping_key
- cluster_compute_resource_overlapping_key

Maybe we can simplify this part a little bit? Something like:
```
cluster_keys = {tag.key for tag in cluster_tags}
queue_keys = {tag.key for tag in queue_tags}
compute_resource_keys = {tag.key for tag in compute_resource_tags}
overlapping_keys = cluster_keys & queue_keys & compute_resource_keys
...
```
We only need the tag keys not values here right?

davprat (changes requested):
Other than the change to the warning message that @delongmeng pointed out ("with" -> "within"), looks good to me.

cli/src/pcluster/validators/tags_validators.py
chenwany: Yeah I was thinking mention the tag value in the validation errors, but we should be fine with just mentioning the tag keys, refactored the code here

cli/src/pcluster/templates/queue_group_stack.py
chenwany: since get_custom_tags only return one dict, we don't need to unpacked it here, we only need the unpack for dict_to_cfn_tags({**tags, **queue_tags, **compute_resource_tags})
Function description is updated, thanks

cli/src/pcluster/schemas/common_schema.py
chenwany: For detecting duplicated keys, it is checked in the same way as we check duplicated instance types, in the case we find duplicate tag, we can raise the error instead of going through all the tag keys, this can reduce the validation time a bit when there's duplicate keys and reduce the size of the validation message
The text message is updated

CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ CHANGELOG
 - Add support for customizing the cluster Slurm configuration via the ParallelCluster configuration YAML file.
 - Track the longest dynamic node idle time in CloudWatch Dashboard.
 - Add new configuration section `HealthChecks/Gpu` for enabling the GPU Health Check in the compute node before job execution.
-- Add support for `Tags` in the `SlurmQueues` section.
+- Add new configuration section `HealthChecks/Gpu` for enabling the GPU Health Check in the compute node before job submission.
+- Add support for `Tags` in the `SlurmQueues` and `SlurmQueues/ComputeResources` section.
```
delongmeng, line 17: Is the line "Add new configuration section `HealthChecks/Gpu` for enabling ..." duplicated here?

delongmeng (dismissed):
looks good to me!

delongmeng (approved):

davprat (approved):
Looks good

PR #5174, opened 2023-04-10
`[integ-tests/lambda-layer]` Parameterize lambda layer to copy.

charlesg3:
### Description of changes
* Add a parameter (`--lambda-layer-source`) that allows the user to specify an S3 URI to use to copy the lambda layer (rather than building it) when creating a resource bucket. If the parameter is specified and the integration tests will create a resource bucket (e.g. a resource bucket is _not_ specified) then it does an S3 copy rather than creating the lambda layer.

### Tests
* The `test_api_infrastructure.py::test_api_infrastructure_with_default_parameters` test was run with the following config:

```
test-suites:
  pcluster_api:
    test_api_infrastructure.py::test_api_infrastructure_with_default_parameters:
      dimensions:
        - regions: ["us-east-2"]
```

The test-runner was run both with and without the `--lambda-layer-source` parameter, e.g.:

```.sh
#!/bin/bash
python3 -m test_runner \
    -c configs/api.yaml \
    --key-name ${KEYNAME} \
    --key-path ${HOME}/keys/${KEYNAME}.pem \
    --lambda-layer-source s3://[personal-bucket]/parallelcluster/3.6.0/layers/aws-parallelcluster/lambda-layer.zip \
    --show-output \
    --vpc-stack "[redacted]" \
    --use-default-iam-credentials \
    --sequential
```

and

```.sh
#!/bin/bash
python3 -m test_runner \
    -c configs/api.yaml \
    --key-name ${KEYNAME} \
    --key-path ${HOME}/keys/${KEYNAME}.pem \
    --show-output \
    --vpc-stack "[redacted]" \
    --use-default-iam-credentials \
    --sequential
```

Both tests passed.

### Checklist


eantonin (approved):
LGTM

PR #5180, opened 2023-04-11
`[custom_resource]` Limit response size.

charlesg3:
### Description of changes
The response size is limited to 4k otherwise it throws an error.

This PR removes `tags.*`, the `clusterConfiguration.url` and the `scheduler.type` keys from the response as well as truncating the validation messages if they are too long (these are also available from the lamba output if they are truncated).

All of this data is provided from the cluster at creation time and is available by describing a cluster (either in the CLI or in a lambda in CFN) so therefore is not needed to be in the response.

### Tests
* An example cluster with custom storage and tags was failing to create before.
* Updating the code allowed the cluster to be created.

### Checklist

### DIFF
--- a/cloudformation/custom_resource/cluster.yaml
+++ b/cloudformation/custom_resource/cluster.yaml
@@ -159,13 +159,18 @@ Resources:
 
           def update_response(data):
               logger.info(data)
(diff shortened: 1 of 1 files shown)


eantonin (approved):
LGTM

PR #5181, opened 2023-04-12
Add redhat8 update cluster integration tests in develop

ermanno:
### Description of changes
This patch adds rhel8 to the common cluster update integration tests in develop, except for the awsbatch one, since awsbatch is not supported in rhel8, and the multi-az one, since our ocr_stack class-scoped fixture has hardcoded 'Linux/UNIX' as platform, while rhel8 required 'Red Hat Enterprise Linux'.

If in a follow up patch we decide to set the platform value conditionally based on the operating system, we ought to pay attention to what happens when running this tests for, i.e., alinux2 and rhel8, since this fixture is shared (class level scope) and the two OSs require different platform values.

### Tests
Run tests in personal pipeline.

### Checklist

### DIFF
--- a/tests/integration-tests/configs/common/common.yaml
+++ b/tests/integration-tests/configs/common/common.yaml
@@ -774,23 +774,23 @@ update:
     dimensions:
       - regions: ["eu-central-1"]
         instances: {{ common.INSTANCES_DEFAULT_X86 }}
-        oss: {{ common.OSS_COMMERCIAL_X86_NO_RHEL8 }}
+        oss: {{ common.OSS_COMMERCIAL_X86 }}
   test_update.py::test_update_compute_ami:
     dimensions:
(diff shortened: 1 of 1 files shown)


tests/integration-tests/configs/common/common.yaml
```diff
@@ -806,7 +806,7 @@ update:
     dimensions:
       - regions: [ "eu-west-2" ]
         schedulers: [ "slurm" ]
-        oss: ["alinux2"]
+        oss: ["alinux2", "rhel8"]
```
enrico-usai: let's replace alinux2 with rhel8 for all of them, since the update code is not OS specific, to avoid increasing the integ test time

enrico-usai (dismissed):

enrico-usai (approved):

PR #5185, opened 2023-04-13
Add integration test for detailed monitoring

delongmeng:
### Description of changes
* add integration test for detailed monitoring
* combine the integration test for detailed monitoring with that of dashboard and alarms, and rename the overall section as monitoring

### Tests
* Run the new integration test for the "monitoring" section locally

### References
* https://github.com/aws/aws-parallelcluster/pull/5171

### Checklist


tests/integration-tests/tests/monitoring/test_monitoring.py
```diff
@@ -190,3 +195,15 @@ def _verify_alarms(alarms, metric_name, instance_id):
         assert_that(alarms[0]["Dimensions"]).contains({"Name": "path", "Value": "/"}).contains(
             {"Name": "InstanceId", "Value": instance_id}
         )
+
+
+def _test_detailed_monitoring(ec2_client, compute_instance_ids):
+    ec2_response = ec2_client.describe_instances(InstanceIds=compute_instance_ids)
```
chenwany: the ec2 client is not used outside the function, what about we define it only in the function and pass the region in:
```
ec2_response = boto3.client("ec2", region_name=region).describe_instances(InstanceIds=InstanceIds[compute_instance_ids[0]])
```

tests/integration-tests/tests/monitoring/test_monitoring.py
```diff
@@ -190,3 +195,15 @@ def _verify_alarms(alarms, metric_name, instance_id):
...
+    monitoring_states = [
+        instance.get("Monitoring").get("State")
+        for reservation in ec2_response.get("Reservations")
+        for instance in reservation.get("Instances")
+    ]
+    assert_that(monitoring_states).is_not_empty()
+    assert_that(set(monitoring_states)).is_length(1)
+    assert_that(monitoring_states[0]).is_equal_to("enabled")
```
chenwany, line 208: This can be simplified to:
```
assert_that(ec2_response["Reservations"][0]["Instances"][0]["Monitoring"]["State"]).is_equal_to(""enabled"")
```

tests/integration-tests/tests/monitoring/test_monitoring.py
delongmeng: yeah that will work as well

tests/integration-tests/tests/monitoring/test_monitoring.py
delongmeng: Can we guarantee there is only 1 compute node? I wrote the function to work with any number of compute nodes and check they are all in the "enabled" state.

chenwany: In your code, you are asserting `assert_that(set(monitoring_states)).is_length(1)`, doesn't that mean the function only works for one compute node?

delongmeng: Not necessarily. There can be any non-zero number of compute nodes and here I'm checking that (1) they all have the same monitoring state; (2) that state is the expected value (here it's "enabled").

chenwany (approved):

PR #5186, opened 2023-04-13
Add Validator for total amount of tags on compute resources

chenwany:
### Description of changes
* Set MAX_TAGS_COUNT = 40 because tags are limited to 50, reserve some tags for parallelcluster specified tags

### Tests
* Unit tests
* Manually test

### References

### Checklist


davprat (dismissed):
Lgtm.

cli/tests/pcluster/validators/test_tags_validators.py
```diff
@@ -100,6 +100,49 @@
...
+            "The number of tags (41) associated with ComputeResource 'dummy_compute_resource' in queue 'dummy_queue' "
+            "has exceeded the limit of 40.",
+        ),
+        (
+            [Tag(f"key{i}", f"value{i}") for i in range(0, 41)],
+            None,
+            None,
+            "The number of tags (41) associated with ComputeResource 'dummy_compute_resource' in queue 'dummy_queue' "
+            "has exceeded the limit of 40.",
+        ),
```
delongmeng: What's the difference between the 2 test cases above?

cli/src/pcluster/validators/tags_validators.py
```diff
@@ -61,3 +63,10 @@ def _validate(self, queue_name, compute_resource_name, cluster_tags, queue_tags,
                 f" '{queue_name}'.",
                 FailureLevel.WARNING,
             )
+
+        if key_count > MAX_TAGS_COUNT:
+            self._add_failure(
+                f"The number of tags ({key_count}) associated with ComputeResource '{compute_resource_name}' in queue "
+                f"'{queue_name}' has exceeded the limit of {MAX_TAGS_COUNT}.",
+                FailureLevel.ERROR,
```
delongmeng, line 71: Is the tag limit on the queue level or cluster level? My understanding is it's the total number combining all 3 levels, if that's the case, how do we determine which queue to report here?

chenwany: it's the total number combining all 3 levels, we do the check for each queues, the validator is under cluster config

cli/tests/pcluster/validators/test_tags_validators.py
chenwany: They are the same, removed one

cli/src/pcluster/constants.py
```diff
@@ -306,3 +306,5 @@ class Operation(Enum):
     Operation.GET_IMAGE_STACK_EVENTS: ["us-iso"],
     Operation.LIST_IMAGE_LOG_STREAMS: ["us-iso"],
 }
+
+MAX_TAGS_COUNT = 40  # Tags are limited to 50, reserve some tags for parallelcluster specified tags
```
delongmeng, line 310: Is the tag number limitation on compute resource level or on the cluster level? Maybe clarify it in the comment?

cli/src/pcluster/validators/tags_validators.py
delongmeng: The error message here seems to imply the problem comes from the Compute resource level because of the wording "_the number of tags associated with ComputeResource xxx_". While I understand the "_tags associated with ComputeResource xxx_" include the tags from the cluster level, queue level, and compute resource level, so it's technically correct, I do feel there is better way to communicate the information to users. Something like this (note this is just an example)

"_The total number of tags ({key_count}) defined under `Tags`, `{queue_name}/Tags` and `{queue_name}/{compute_resource_name}/Tags` has exceeded the limit of {MAX_TAGS_COUNT}._"

This is just a suggestion, not a blocking comment, so feel free to go with your preferred solution.

chenwany: Thank you for your suggestion, however, {queue_name}/{compute_resource_name} is not a section name, usually we just write the section name, please see examples in other verification messages, I have indicated which queue or which computing resources in the verification message

cli/src/pcluster/constants.py
chenwany: This limitation is for all resources, see public documentation here: https://docs.aws.amazon.com/servicequotas/latest/userguide/sq-tagging.html

delongmeng: Thanks for sharing the link. I meant to clarify it in the comment part of the code line `MAX_TAGS_COUNT = 40`.
The current comment "Tags are limited to 50" is not very clear about which tags it's talking about, when this variable is defined in a common cluster-level script. Something like "Tags are limited to 50 for each compute resource" would be clearer in my opinion.

cli/src/pcluster/validators/tags_validators.py
delongmeng: That was just an example to express the idea. The point was to make clear the total number comes from all 3 levels. Is it better now:

"_The total number of tags ({key_count}) defined under `Tags`, `SlurmQueue/Tags` and `SlurmQueue/ComputeResources/Tags` has exceeded the limit of {MAX_TAGS_COUNT} for ComputeResource {compute_resource_name} in queue {queue_name}._" ?

cli/src/pcluster/constants.py
chenwany: Although the validator is implemented on the compute resource, the limitation is not only for the compute resource, this constant can be used in somewhere else. Thank you for your suggestions. After reviewing your comments and considering the goals of this task, I believe that the level of detail I provided in the PR is appropriate and sufficient. However, I’m open to discussing any specific concerns you may have or answering any questions you may need clarification on :）

cli/src/pcluster/validators/tags_validators.py
chenwany: The message you suggested can cause misunderstanding, it is possible they don't have queue tags, or they don't have compute resource tags, or they have tags under other queues.

Thank you for your suggestions, I believe that the level of detail provided in the current validation message is sufficient to achieve the desired outcome of preventing cluster creation in case of an error in this corner case. I will go with my preferred solution. Please reach out to me if you have any concerns

delongmeng (approved):

Issue #5151, opened 2023-04-04
(3.3.0-3.5.1) Cluster updates can break Slurm accounting functionality


enrico-usai:
## The issue

In ParallelCluster versions `3.3.0` up to `3.5.1`, if you enabled the [Slurm accounting](https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html#Scheduling-v3-SlurmSettings-Database) feature, a  `cluster-update` operation involving any change to settings under the [`Scheduling`](https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html) section, will break Slurm accounting configuration on the system.

Slurm accounting functionality will continue running unless a restart of the `slurmdbd` or, more in general, a reboot of the head node is executed.

## Mitigation

See https://github.com/aws/aws-parallelcluster/wiki/(3.3.0-3.5.1)-Cluster-updates-can-break-Slurm-accounting-functionality

enrico-usai:
Closing since all version > 3.5.0 contain a fix.

PR #5197, opened 2023-04-14
Modify test_slurm_accounting test to account for issue #5151

jdeamicis:
### Description of changes
* Modify test_slurm_accounting test to account for issue #5151
* *WARNING*: this may not the best place to cover the logic of the cookbook recipes. Evaluate if this is needed and acceptable, otherwise implement a check on the cookbook test infrastructure.

### Tests
* See above

### References
* Related to https://github.com/aws/aws-parallelcluster-cookbook/pull/2006

### Checklist

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm_accounting.py
@@ -123,6 +123,14 @@ def _test_that_slurmdbd_is_running(remote_command_executor):
     assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
 
 
+def _test_slurm_accounting_password(remote_command_executor):
+    storage_pass = remote_command_executor.run_remote_command(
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/schedulers/test_slurm_accounting.py
```diff
@@ -123,6 +123,14 @@ def _test_that_slurmdbd_is_running(remote_command_executor):
     assert_that(_is_accounting_enabled(remote_command_executor)).is_true()
 
 
+def _test_slurm_accounting_password(remote_command_executor):
```
NSsirena, line 126: As you said, this should be addressed on the cookbook since the behavior we are testing is that of a recipe.

However fixing the cookbook now is not strictly necessary, so it may be acceptable to start with this change in an existing test as long as
* we keep track of this need (with a comment here and a Task to move this test in the cookbook)
* we begin to review the integration tests and the cookbook

NSsirena (dismissed):

NSsirena (approved):

PR #5212, opened 2023-04-18
[FSxFileCache][3.x]Adding FsxFileCacheSettings/FileCacheId Config Option under SharedStorage

himani2411:
### Description of changes
* Adding Unmanaged [Fsx File Cache](https://docs.aws.amazon.com/fsx/latest/FileCacheGuide/what-is.html) Config Option under [SharedStorage section](https://docs.aws.amazon.com/parallelcluster/latest/ug/SharedStorage-v3.html)

### Tests
* Unit Tests for Validating the Config Options 
* Manual Local Tests for Validation of Config Options

### Checklist

### DIFF
--- a/cli/src/pcluster/aws/aws_resources.py
+++ b/cli/src/pcluster/aws/aws_resources.py
@@ -288,22 +288,28 @@ class InstanceTypeInfo:
         return self.instance_type_data.get("MemoryInfo", {}).get("SizeInMiB")
 
 
-class FsxFileSystemInfo:
-    """Data object wrapping the result of a describe_file_systems call."""
+class FsxStorageInfo:
+    """Data object wrapping the result of a describe_file_systems and describe_file_caches call."""
 
-    def __init__(self, file_system_data):
-        self.file_system_data = file_system_data
+    def __init__(self, file_storage_info):
+        self.file_storage_info = file_storage_info
 
     @property
-    def file_system_type(self):
+    def file_storage_type(self):
         """Return the type of FSx file system (LUSTRE, WINDOWS, ONTAP, or OPENZFS). WINDOWS is not supported."""
-        return self.file_system_data.get("FileSystemType")
+        return (
+            self.file_storage_info.get("FileSystemType")
+            if self.file_storage_info.get("FileSystemType")
+            else self.file_storage_info.get("FileCacheType")
+        )
 
     @property
     def mount_name(self):
         """Return MountName of the FSx Lustre file system."""
         return (
-            self.file_system_data.get("LustreConfiguration").get("MountName") if self.file_system_type == LUSTRE else ""
+            self.file_storage_info.get("LustreConfiguration").get("MountName")
+            if self.file_storage_type == LUSTRE
+            else ""
         )
 
     @property
@@ -313,27 +319,32 @@ class FsxFileSystemInfo:
 
         Lustre, OpenZFS have DNS name on file systems. Ontap has DNS name on storage virtual machines.
         """
-        return self.file_system_data.get("DNSName") if self.file_system_type in [LUSTRE, OPENZFS] else ""
+        return self.file_storage_info.get("DNSName") if self.file_storage_type in [LUSTRE, OPENZFS] else ""
 
     @property
     def file_system_id(self):
         """Return id of the file system."""
-        return self.file_system_data.get("FileSystemId")
+        return self.file_storage_info.get("FileSystemId")
+
+    @property
+    def file_cache_id(self):
+        """Return id of the file caches."""
+        return self.file_storage_info.get("FileCacheId")
 
     @property
     def vpc_id(self):
         """Return VPC id of the file system."""
-        return self.file_system_data.get("VpcId")
+        return self.file_storage_info.get("VpcId")
 
     @property
     def network_interface_ids(self):
         """Return network interface ids of the file system."""
-        return self.file_system_data.get("NetworkInterfaceIds")
+        return self.file_storage_info.get("NetworkInterfaceIds")
 
     @property
     def subnet_ids(self):
         """Return subnet ids of the file system."""
-        return self.file_system_data.get("SubnetIds")
+        return self.file_storage_info.get("SubnetIds")
 
 
 class ImageInfo:
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -621,7 +621,18 @@ class ExistingFsxFileCache(BaseSharedFsx):
     def __init__(self, file_cache_id: str, **kwargs):
         super().__init__(**kwargs)
         self.file_cache_id = file_cache_id
+        self.file_system_id = file_cache_id
         self.file_system_type = FILECACHE
+
+    @property
+    def existing_dns_name(self):
+        """Return DNSName of the existing FSx File Cache."""
+        return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0].dns_name
+
+    @property
+    def file_cache_mount_name(self):
+        """Return Mount Name of the existing FSx File Cache."""
+        return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0].mount_name
 
 
 # ---------------------- Networking ---------------------- #
@@ -1462,6 +1473,8 @@ class BaseClusterConfig(Resource):
                     self._register_validator(FeatureRegionValidator, feature=Feature.FSX_ONTAP, region=self.region)
                 elif isinstance(storage, ExistingFsxFileCache):
                     existing_storage_count["fsx"] += 1
+                    existing_fsx.add(storage.file_cache_id)
+                    self._register_validator(FeatureRegionValidator, feature=Feature.FSX_FILE_CACHE, region=self.region)
                 elif isinstance(storage, SharedEbs):
                     if storage.raid:
                         new_storage_count["raid"] += 1
@@ -1480,11 +1493,10 @@ class BaseClusterConfig(Resource):
                         new_storage_count["efs"] += 1
             self._register_validator(
                 ExistingFsxNetworkingValidator,
-                file_system_ids=list(existing_fsx),
+                file_storage_ids=list(existing_fsx),
                 subnet_ids=[self.head_node.networking.subnet_id] + self.compute_subnet_ids,
                 security_groups_by_nodes=self.security_groups_by_nodes,
             )
-
             self._validate_max_storage_count(ebs_count, existing_storage_count, new_storage_count)
             self._validate_new_storage_multiple_subnets(
                 self.scheduling.queues, self.compute_subnet_ids, new_storage_count
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -41,6 +41,7 @@ from pcluster.constants import (
     EBS_VOLUME_TYPE_DEFAULT,
     EBS_VOLUME_TYPE_DEFAULT_US_ISO,
     EBS_VOLUME_TYPE_IOPS_DEFAULT,
+    FILECACHE,
     LUSTRE,
     MAX_COMPUTE_RESOURCES_PER_QUEUE,
     MAX_EBS_COUNT,
@@ -612,6 +613,15 @@ class ExistingFsxOntap(BaseSharedFsx):
         return AWSApi.instance().fsx.describe_storage_virtual_machines([self.storage_virtual_machine_id])[0][
             "Endpoints"
         ]["Nfs"]["DNSName"]
+
+
+class ExistingFsxFileCache(BaseSharedFsx):
+    """Represent the shared FSX FileCache resource."""
+
+    def __init__(self, file_cache_id: str, **kwargs):
+        super().__init__(**kwargs)
+        self.file_cache_id = file_cache_id
+        self.file_system_type = FILECACHE
 
 
 # ---------------------- Networking ---------------------- #
@@ -1450,6 +1460,8 @@ class BaseClusterConfig(Resource):
                     existing_storage_count["fsx"] += 1
                     existing_fsx.add(storage.file_system_id)
                     self._register_validator(FeatureRegionValidator, feature=Feature.FSX_ONTAP, region=self.region)
+                elif isinstance(storage, ExistingFsxFileCache):
+                    existing_storage_count["fsx"] += 1
                 elif isinstance(storage, SharedEbs):
                     if storage.raid:
                         new_storage_count["raid"] += 1
@@ -1612,6 +1624,8 @@ class BaseClusterConfig(Resource):
                     storage_id = storage.file_system_id
                 elif isinstance(storage, (SharedEbs, ExistingFsxOpenZfs, ExistingFsxOntap)):
                     storage_id = storage.volume_id
+                elif isinstance(storage, ExistingFsxFileCache):
+                    storage_id = storage.file_cache_id
                 if storage_id:
                     storage_id_list.append(storage_id)
         return storage_id_list
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -245,6 +245,7 @@ class Feature(Enum):
     BATCH = "AWS Batch scheduler"
     DCV = "NICE DCV"
     FSX_LUSTRE = "FSx Lustre"
+    FSX_FILE_CACHE = "FSx FileCache"
     FSX_ONTAP = "FSx ONTAP"
     FSX_OPENZFS = "FSx OpenZfs"
     SLURM_DATABASE = "SLURM Database"
@@ -254,6 +255,7 @@ UNSUPPORTED_FEATURES_MAP = {
     Feature.BATCH: ["ap-northeast-3", "us-iso"],
     Feature.DCV: ["us-iso"],
     Feature.FSX_LUSTRE: ["us-iso"],
+    Feature.FSX_FILE_CACHE: ["us-iso"],
     Feature.FSX_ONTAP: ["us-iso"],
     Feature.FSX_OPENZFS: ["us-iso"],
     Feature.SLURM_DATABASE: ["us-iso"],
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -66,13 +66,17 @@ FSX_HDD_THROUGHPUT = [12, 40]
 LUSTRE = "LUSTRE"
 OPENZFS = "OPENZFS"
 ONTAP = "ONTAP"
+FILECACHE = "FILECACHE"
 
 FSX_LUSTRE = "FsxLustre"
 FSX_OPENZFS = "FsxOpenZfs"
 FSX_ONTAP = "FsxOntap"
+FSX_FILE_CACHE = "FsxFileCache"
 
 # https://docs.aws.amazon.com/fsx/latest/APIReference/API_DescribeVolumes.html#FSx-DescribeVolumes-request-VolumeIds.
 FSX_VOLUME_ID_REGEX = r"^fsvol-[0-9a-f]{17}$"
+# https://docs.aws.amazon.com/fsx/latest/APIReference/API_FileCacheCreating.html#FSx-Type-FileCacheCreating-FileCacheId:~:text=Pattern%3A-,%5E(fc%2D%5B0%2D9a%2Df%5D%7B8%2C%7D)%24,-Required%3A%20No
+FSX_FILE_CACHE_ID_REGEX = r"^(fc-[0-9a-f]{8,18})$"
 
 FSX_PORTS = {
     # Lustre Security group: https://docs.aws.amazon.com/fsx/latest/LustreGuide/limit-access-security-groups.html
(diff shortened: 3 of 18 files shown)


himani2411:
Looking at the Failing Tests/Checks

cli/src/pcluster/config/cluster_config.py
```diff
@@ -614,6 +615,15 @@ def existing_dns_name(self):
         ]["Nfs"]["DNSName"]
 
 
+class ExistingFsxFileCache(BaseSharedFsx):
```
gmarciani, line 618: *Observation to address in a separate PR, but worth starting the discussion for it.*

The file `cluster_config.py` is getting bigger and bigger and this is not a good practice.
We should consider to refactor it and put classes definition in specific files.

For example, we should move every classes related to the shared storage into a dedicated `shared_storage_config.py` file.

Let's think about it as a follow-up improvement.

cli/src/pcluster/constants.py
```diff
@@ -66,13 +66,16 @@
...
+FILECACHE = "FILECACHE"
 
 FSX_LUSTRE = "FsxLustre"
 FSX_OPENZFS = "FsxOpenZfs"
 FSX_ONTAP = "FsxOntap"
+FSX_FILE_CACHE = "FsxFileCache"
 
 # https://docs.aws.amazon.com/fsx/latest/APIReference/API_DescribeVolumes.html#FSx-DescribeVolumes-request-VolumeIds.
 FSX_VOLUME_ID_REGEX = r"^fsvol-[0-9a-f]{17}$"
+FSX_FILE_CACHE_ID_REGEX = r"^fc-[0-9a-f]{17}$"
```
gmarciani: Minor: if possible, let's add a comment with the link to the FSx public docs where the regexp is defined.

cli/src/pcluster/constants.py
```diff
@@ -66,13 +66,16 @@
 LUSTRE = "LUSTRE"
 OPENZFS = "OPENZFS"
 ONTAP = "ONTAP"
+FILECACHE = "FILECACHE"
```
gmarciani, line 69: *Observation to address in a separate PR, but worth starting the discussion for it.*

I understand that we need to add the constant `FILECACHE` and `FSX_FILE_CACHE` because we have always done this way for other storage types.

Nevertheless, we should improve this by requiring just one constant.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -490,7 +504,7 @@ class SharedStorageSchema(BaseSchema):
     name = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
     storage_type = fields.Str(
         required=True,
-        validate=validate.OneOf(["Ebs", FSX_LUSTRE, FSX_OPENZFS, FSX_ONTAP, "Efs"]),
+        validate=validate.OneOf(["Ebs", FSX_LUSTRE, FSX_OPENZFS, FSX_ONTAP, "Efs", FSX_FILE_CACHE]),
```
gmarciani, line 507: *Observation to address in a separate PR, but worth starting the discussion for it.*

Why "Efs" is the only one without a corresponding constant. We should fix this.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -552,6 +578,8 @@ def make_resource(self, data, **kwargs):
             return ExistingFsxOpenZfs(**shared_volume_attributes)
         elif storage_type == FSX_ONTAP:
             return ExistingFsxOntap(**shared_volume_attributes)
+        elif storage_type == FSX_FILE_CACHE:
+            return ExistingFsxFileCache(**shared_volume_attributes)
```
gmarciani, line 582: why do we need to inject shared_volume_attributes in File Cache?
In FileCache we do not have the concept of volume...

gmarciani (dismissed):

gmarciani:
I suggest to use the following prefix for every PR title and commits related to this feature: `[FSxFileCache]`.
In this way it will be easier to search for PRs and commits related to the feature when we are going to release it or look for issues to address

cli/src/pcluster/schemas/cluster_schema.py
himani2411: shared_volume_attributes has mount_dir, Name, which are not Volume specific.

cli/src/pcluster/config/cluster_config.py
himani2411: As discussed, we can add this in a follow up activity.

cli/src/pcluster/constants.py
himani2411: Will add.

cli/src/pcluster/constants.py
himani2411: we have 2 different constants for each to improve readability of the code. I would have to use mostly to_pascal_case() or to_snake_case(),etc at multiple places in the code.

cli/src/pcluster/schemas/cluster_schema.py
himani2411: As discussed, we can add this in a follow up activity.

cli/src/pcluster/constants.py
```diff
@@ -66,13 +66,17 @@
...
 
 FSX_LUSTRE = "FsxLustre"
 FSX_OPENZFS = "FsxOpenZfs"
 FSX_ONTAP = "FsxOntap"
+FSX_FILE_CACHE = "FsxFileCache"
 
 # https://docs.aws.amazon.com/fsx/latest/APIReference/API_DescribeVolumes.html#FSx-DescribeVolumes-request-VolumeIds.
 FSX_VOLUME_ID_REGEX = r"^fsvol-[0-9a-f]{17}$"
+# https://docs.aws.amazon.com/fsx/latest/APIReference/API_FileCacheCreating.html#FSx-Type-FileCacheCreating-FileCacheId:~:text=Pattern%3A-,%5E(fc%2D%5B0%2D9a%2Df%5D%7B8%2C%7D)%24,-Required%3A%20No
+FSX_FILE_CACHE_ID_REGEX = r"^(fc-[0-9a-f]{8,18})$"
```
gmarciani, line 79: Congrats with this point.
The public docs declares as regexp `^(fc-[0-9a-f]{8,})$` (not upper bound limit to 18). 
You did a step forward here by setting the limit to 21-3=18 as it should be according to a more in depth reading of the documentation. 
Thank you!

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: *Observation to address in a separate PR, but worth starting the discussion for it.*

the variable `shared_volume_settings` should be renamed in shared_storage_settings` because not all shared storage has the notion of volumes (sometimes we have volumes, sometimes we have file systems, other times we have caches)

cli/tests/pcluster/aws/dummy_aws_api.py
```diff
@@ -265,6 +265,13 @@ def describe_storage_virtual_machines(self, storage_virtual_machine_ids):
             result.append({"Endpoints": {"Nfs": {"DNSName": "abcd"}}})
         return result
 
+    def describe_file_caches(self, file_cache_ids):
+        """Describe FSx File cache."""
+        result = []
+        for _ in file_cache_ids:
+            result.append({"DNSName": "abcd", "LustreConfiguration": {"MountName": "/abcd"}})
```
gmarciani: *Minor: Code Style*
When settings mocked/testing values it's a good practice to not suggest relations between values that does not exists in the real case. In this specific line you are whispering to the developer that there exists a relation between the DNS name and the mount name, but this is not the case. I suggest to use something like "a.dns.name" and "/amoutname" instead of "abcd" and "/abcd". This is a good practice also because it make it easier to read the debugger outputs in case we need to troubleshoot with it.

cli/tests/pcluster/aws/test_fsx.py
```diff
@@ -124,3 +124,38 @@ def test_describe_volumes(boto3_stubber):
...
+        get_describe_file_caches_mocked_request([file_cache_id], "CREATING"),
+        get_describe_file_caches_mocked_request([additional_file_cache_id], "AVAILABLE"),
+        get_describe_file_caches_mocked_request([file_cache_id], "AVAILABLE"),
+    ]
+    boto3_stubber("fsx", mocked_requests)
+    assert_that(AWSApi.instance().fsx.describe_file_caches([file_cache_id])[0]["Lifecycle"]).is_equal_to("CREATING")
+
+    # Second boto3 call with more FileCaches. The FC already cached should not be included in the boto3 call.
+    response = AWSApi.instance().fsx.describe_file_caches([file_cache_id, additional_file_cache_id])
+    assert_that(response).is_length(2)
```
gmarciani, line 152: Not clear to me why this assertion should prove that the cached file cache is not included in the boto3 call.
If your goal is to verify that a describe call is not execute for a cached FC, then you need to assert that the underlying boto3 call has not been called with that id as parameter.

cli/tests/pcluster/aws/test_fsx.py
```diff
@@ -124,3 +124,38 @@ def test_describe_volumes(boto3_stubber):
     # Fourth boto3 call after resetting the AWSApi instance. The latest fsx lifecycle should be retrieved from boto3
     AWSApi.reset()
     assert_that(AWSApi.instance().fsx.describe_volumes([volume])[0]["Lifecycle"]).is_equal_to("CREATED")
+
+
+def get_describe_file_caches_mocked_request(file_cache_ids, lifecycle):
```
gmarciani: Why do we need the lifecyle in the tests?
I'm asking because I do not expect the CLI to do any check on the status of the FC, but only to verify that it exists and that it can be descirbed.

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -590,6 +590,45 @@ def test_slurm_flexible_queue(mocker, config_dict, failure_message):
...
+            {
+                "StorageType": "FsxFileCache",
+                "Name": "name",
+                "MountDir": "mount/tmp",
+                "FsxFileCacheSettings": {"FileCacheId": "fc-123456789012345678"},
+                "EfsSettings": {"Encrypted": True},
+            },
+            "Multiple .*Settings sections cannot be specified in the SharedStorage items",
+        ),
+        ({"StorageType": "FsxFileCache", "MountDir": "mount/tmp"}, "Missing data for required field."),
```
gmarciani, line 604: *Minor/Code Style*
It's a good practice to document each test case with a description because:
1. it's easier for the reader to understand the goal of the test case when reading the code.
2. it's easier for a developer to understand what test case is failing when running the test locally.
You can see examples [here](https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/templates/test_directory_service.py#L26-L36).

An argument may be that: we are not using this habit constantly in our codebase and that's true.
My suggestion is to use it at least on new test cases.
If using in your test cases onyl will make the test fail, then we will come back with a refactoring activity on a separate PR, otherwise I suggest to do that.

cli/tests/pcluster/templates/test_shared_storage.py
```diff
@@ -177,6 +178,51 @@ def test_shared_storage_fsx(mocker, test_datadir, config_file_name, storage_name
         )
 
 
+@pytest.mark.parametrize(
+    "config_file_name",
+    [
+        ("unmanaged_config.yaml"),
+    ],
+)
+def test_unmanaged_shared_storage_fsx(mocker, test_datadir, config_file_name):
```
gmarciani, line 187: Very weel done!

cli/tests/pcluster/aws/dummy_aws_api.py
himani2411: Got it. I have been following the pattern that other Storage mockers have set.

cli/tests/pcluster/templates/test_shared_storage.py
himani2411: Thank You!

cli/tests/pcluster/aws/test_fsx.py
himani2411: I assumed while checking if the cache exists we will also check for the Lifecycle of the Cache like it should not be in Failed or Deleting phase while we are creating the cluster. If this is not required I can remove the Lifecycle check in these tests.

cli/src/pcluster/constants.py
himani2411: Thank you!

cli/tests/pcluster/schemas/test_cluster_schema.py
himani2411: I think adding comments for the newly added test case should suffice, as I dont want the description to be mistaken as part of the test values. if you agree on that I will go ahead with the change.

cli/tests/pcluster/aws/test_fsx.py
himani2411: I think this would be more complicated in terms of checking for underlying boto3 call. But I can check for the fc_cache contains values before and after.

cli/tests/pcluster/aws/test_fsx.py
gmarciani: I think it's not required especially because the FC status may change from the time you check it and the time the cluster will attempt to mount it.

cli/tests/pcluster/aws/test_fsx.py
gmarciani: This check would not have exactly the same value, but that's completely fine

cli/tests/pcluster/schemas/test_cluster_schema.py
gmarciani: Adding comments does not provide the same value because that approach does not provide advantage (2).
I prefer the pytest native approach, but I'm ok either ways.

cli/tests/pcluster/aws/test_fsx.py
himani2411: Sure, will remove the check

himani2411:
Working on solving the pycov checks

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -288,22 +288,28 @@ def ec2memory_size_in_mib(self):
...
 
-class FsxFileSystemInfo:
-    """Data object wrapping the result of a describe_file_systems call."""
+class FsxStorageInfo:
+    """Data object wrapping the result of a describe_file_systems and describe_file_caches call."""
 
-    def __init__(self, file_system_data):
-        self.file_system_data = file_system_data
+    def __init__(self, file_storage_data):
+        self.file_storage_data = file_storage_data
```
gmarciani: I think you're renaming `file_system_data` into `file_storage_data` to make this logic more generic to cover both file systems and file caches.  Nevertheless, the concept of `file_storage` is really uncommon. So I would suggest to rename it in `storage_info`. because it contains the information about the storage returned by FSx.

what do you think?

cli/src/pcluster/aws/aws_resources.py
```diff
@@ -288,22 +288,28 @@ def ec2memory_size_in_mib(self):
...
 
     @property
-    def file_system_type(self):
+    def file_storage_type(self):
```
gmarciani, line 298: `FileCacheType` always returns `LUSTRE`, which seems counterintuitive since you're using FileCache and not FsxLustre. Nevertheless, this is correct because FileCache is a cache on top of a Lustre file system.
So:
1. if your logic is expecting this property to return a value representing FileCache, then we must fix it
2. It's correct for this property to be named `file_system_type` because Lustre is actually the file system under FileCache, so the property does not need to be generalized.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -614,6 +615,28 @@ def existing_dns_name(self):
...
+        super().__init__(**kwargs)
+        self.file_cache_id = file_cache_id
+        self.file_system_id = file_cache_id
+        self.file_system_type = FILECACHE
+
+    @property
+    def existing_dns_name(self):
+        """Return DNSName of the existing FSx File Cache."""
+        return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0].dns_name
+        # return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0]["dns_name"]
```
gmarciani: This comment should be removed, right?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -614,6 +615,28 @@ def existing_dns_name(self):
...
+
+    @property
+    def file_cache_mount_name(self):
+        """Return Mount Name of the existing FSx File Cache."""
+        return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0].mount_name
+        # return AWSApi.instance().fsx.describe_file_caches([self.file_cache_id])[0]["mount_name"]
```
gmarciani: This comment should be removed, right?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1429,6 +1452,7 @@ def _register_storage_validators(self):  # noqa: C901 FIXME: function too comple
             )
 
             existing_fsx = set()
+            existing_fsx_file_cache_id = set()
```
gmarciani: Why FileCache needs a separated set while any other FSx storage insist on the same set?
If the reason is that you want to apply a validator only to the existing FC and not to all FSx storage, then I would have the following follow up questions:
1. why a validator would be necessary only for FC but not for the other FSx storage types?
2. what about having a unique set and filtering FC within this set, e.g. [fc_id for fc_id in existing_fsx if fc_id.starts_with("fc-")] rather than splitting the concepts.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1450,6 +1474,9 @@ def _register_storage_validators(self):  # noqa: C901 FIXME: function too comple
                     existing_storage_count["fsx"] += 1
                     existing_fsx.add(storage.file_system_id)
                     self._register_validator(FeatureRegionValidator, feature=Feature.FSX_ONTAP, region=self.region)
```
gmarciani, line 1473: According to the feature design we decided not to add any feature/region validation, so you're doing right.
Nevertheless, I'm seeing now that we are actually doing that for other storage types, so I suggest to have a consistent approac. If you agree with that, let's add the feature/region validator in a separate PR.

In case we dice to remove such validation, we should do that consistently for every storage type, but let's keeo the things aligned.

cli/src/pcluster/aws/aws_resources.py
himani2411: But all the storage by FSx are w.r.t File systems. So, I dont see how file_storage is uncommon

cli/src/pcluster/config/cluster_config.py
himani2411: Sure

cli/src/pcluster/config/cluster_config.py
himani2411: I am not removing validation for the other storages. There are a few paramters that are different for file systems and file cache, but I can incorporate something to make it simpler.

cli/src/pcluster/config/cluster_config.py
himani2411: Yes

cli/src/pcluster/config/cluster_config.py
himani2411: Yes. will remove it

cli/src/pcluster/aws/aws_resources.py
himani2411: No the value will return LUSTRE, but i renamed the property in leu of consistency for all the properties under the FSxStorageInfo class. I believe it would be easier to understand using `file_storage_type , storage_type or fsx_storage_type` rather than `file_system_type `. 

What do you think?

cli/src/pcluster/aws/aws_resources.py
himani2411: Will change it to file_storage_info

cli/src/pcluster/config/cluster_config.py
```diff
@@ -614,6 +615,26 @@ def existing_dns_name(self):
         ]["Nfs"]["DNSName"]
 
 
+class ExistingFsxFileCache(BaseSharedFsx):
+    """Represent the shared FSX FileCache resource."""
+
+    def __init__(self, file_cache_id: str, **kwargs):
+        super().__init__(**kwargs)
+        self.file_cache_id = file_cache_id
```
gmarciani, line 623: Why are we are duplicating the information?

```
self.file_cache_id = file_cache_id
self.file_system_id = file_cache_id
```

himani2411: This was a workaround, as the parent class uses file_system_id. However, I can change the design of the parent class, if you want.

gmarciani: Now I remember that we considered it out of scope for this specific PR to not make it bigger.
Let's go with your solution and fix it a a followup PR.
Thanks, approved.

gmarciani (approved):

cli/src/pcluster/config/cluster_config.py
himani2411: As discussed, will add resolution in a followup PR

PR #5229, opened 2023-04-21
Export SLURM_TASKS_PER_NODE on NCCL test to avoid Slurm 23.02 bug

enrico-usai:
Temporary workaround to solve a bug in Slurm 23.02: https://bugs.schedmd.com/show_bug.cgi?id=16426

### DIFF
--- a/tests/integration-tests/tests/efa/test_efa/test_efa/nccl_benchmarks/nccl_tests_submit_openmpi.sh
+++ b/tests/integration-tests/tests/efa/test_efa/test_efa/nccl_benchmarks/nccl_tests_submit_openmpi.sh
@@ -3,6 +3,8 @@
 #SBATCH --exclusive
 #SBATCH --ntasks-per-node=8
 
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/efa/test_efa/test_efa/nccl_benchmarks/nccl_tests_submit_openmpi.sh
```diff
@@ -3,6 +3,8 @@
 #SBATCH --exclusive
 #SBATCH --ntasks-per-node=8
 
+# Temporary workaround to solve a bug in Slurm 23.02: https://bugs.schedmd.com/show_bug.cgi?id=16426
+export SLURM_TASKS_PER_NODE="8(x2)"
```
jdeamicis, line 7: If we want we can also make this like:

```
export SLURM_TASKS_PER_NODE="${SLURM_NTASKS_PER_NODE}(x${SLURM_NNODES})"
```

But since we are not going to change the configuration for this benchmark, it doesn't matter if we leave it hard-coded (plus it's a temporary workaround).

jdeamicis (approved):

PR #5248, opened 2023-05-03
Move custom resource integ tests to develop config

ermanno:
### Description of changes
- **Move custom resource integration tests to develop config.**
- **Reuse session scoped factory fixture for custom resource integration tests:**
The resource bucket policy stack should be deleted after deleting the custom resource stack. Usually, when deleting all stacks, a factory would delete them in the inverse order in which they were created. However, the bucket policy and custom resource were created by different factories, so the order of deletion was not deterministic. Since we already have a session scoped factory fixture, we can reuse it in the custom resource tests instead of instantiating one ourselves. In this way the two stacks will be created by the same factory and the order of deletion will be respected.

### Tests
Launched custom resource tests in personal Jenkins.

### Checklist


eantonin (dismissed):
LGTM

charlesg3 (approved):

PR #5271, opened 2023-05-05
[3.x][FileCache] Add Fsx FileCache in DFSM and its own Integration Tests

himani2411:
### Description of changes
* Adding Fsx FileCache integration tests to test the mounting of the external Fsx FC, reading already existed files and write files in FC.
* Adding Fsx FC in DFSM integration tests.
* Adding unmanaged Fsx storage  in CW Dashboard unit tests.

### Tests
* Tested the changes Locally

### Checklist


tests/integration-tests/conftest.py
```diff
@@ -17,6 +17,7 @@
 import logging
 import os
 import re
+import time
```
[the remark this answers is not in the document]
himani2411: Making changes to remove unwanted imports.

himani2411:
Looking at failing tox checks

cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -467,6 +467,9 @@ def _add_fsx_widgets(self, storages_list):
             if isinstance(storage.config, SharedFsxLustre):
                 metrics = lustre_metrics
                 dimensions_map = {"FileSystemId": storage.id}
+            elif isinstance(storage.config, ExistingFsxFileCache):
```
gmarciani: This change is not related to integration tests, which according to the PR title seems to be the focus of the PR. *General hint:* let's decouple PRs in smaller pieces.
*Specific hint:* let's align PR title and description to the shipped changes.

cli/tests/pcluster/templates/test_cw_dashboard_builder/test_cw_dashboard_builder/alinux2.batch.head_node_log.yaml
```diff
@@ -44,6 +44,21 @@ SharedStorage:
     StorageType: Ebs
     EbsSettings:
       VolumeType: st1
+  - MountDir: /my/mount/fc1
+    Name: name6
+    StorageType: FsxFileCache
+    FsxFileCacheSettings:
+      FileCacheId: fc-12345678
+  - MountDir: /my/mount/fsx-ontap/1
```
gmarciani: Why are we adding ONTAP and OpenZFS? Shouldn't the focus of this PR be limited to FileCache?

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
+
+Description: This template to creates FSX Filecache storage that can be attached to a cluster for Integration Test.
```
gmarciani: Minor: typo `FSX Filecache`-> `FSx FileCache`.
Minor: `for Integration Test` the template does not contain integ test specific stuff, such as Jinja variables, and this is great! So it's a more general solution that a template for integ tests. In this case I suggest to remove the "for Integration Test" sentence. I'm saying this because in the long run I think we should deploy auxiliary infrastructures with the same CFN templates we provide in our public docs/tutorial, so something not specific for integ tests, but in general meant to deploy the integrations.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+    Description: Subnet ID.
+    Type: AWS::EC2::Subnet::Id
+    Default: ""
+
+Resources:
+
+  FsxFileCacheSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Allow TCP Traffic
```
gmarciani: I suggest to be more specific in the SG description, e.g. "Allow traffic for FSx File Cache"

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+
+                logger.info(f"fsx_file_cache_id: {file_cache_id}")
+                # provide outputs
+                response_data['FileCacheId'] = file_cache_id
+                response_data['FileCacheARN'] = file_cache_arn
+              else:
+                  physical_resource_id = event['PhysicalResourceId']
+            except Exception as e:
+                response_status = cfnresponse.FAILED
+                reason = str(e)
```
gmarciani, line 153: what about emitting an error log line with the exception message? In this way we will have the reason in lambda cw ogs.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+            securitygroup_id = event['ResourceProperties']['SecurityGroupId']
+            logger.info(f"securitygroup_id: {securitygroup_id}")
+
+
+            response_data = {}
+            reason = None
+            response_status = cfnresponse.SUCCESS
+
+            try:
+              if event['RequestType'] == 'Create':
```
gmarciani, line 121: The custom resource should take care also for the deletion of the resource, not only for the creation.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+          IpProtocol: "-1"
+          ToPort: -1
+      SecurityGroupIngress:
+        - CidrIp: 0.0.0.0/0
+          FromPort: 988
+          IpProtocol: tcp
+          ToPort: 988
+      VpcId: !Ref VpcId
+
+  CreateFsxFileCacheLambda:
```
gmarciani: Custom resource takes care of the entire resource lifecycle, not only the creation. So, the name and description should reflect this responsibility.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+                    {
+                      "FileCachePath": file_cache_path,
+                      "DataRepositoryPath": s3_bucket,
+                    },
+                  ],
+                ).get("FileCache")
+              )
+
+            except fsx.exceptions.InternalServerError as e:
+              logger.info(f"Cannot create Fsx FileCache because of InternalServerError.")
```
gmarciani: should log at error level, not info.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+        Statement:
+          - Action:
+              - sts:AssumeRole
+            Effect: Allow
+            Principal:
+              Service:
+                - lambda.amazonaws.com
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonFSxFullAccess
+        - arn:aws:iam::aws:policy/AmazonFSxConsoleFullAccess
```
gmarciani: Why is `AmazonFSxConsoleFullAccess ` required?

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+      Policies:
+        - PolicyName: LogOutput
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Action:
+                  - logs:CreateLogGroup
```
gmarciani: Permissions required by a Lambda function to post logs and metrics are included in the managed policy `arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole`. I suggest to always use AWS managed policy when possible.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+                  - logs:CreateLogStream
+                  - logs:PutLogEvents
+                  - s3:Get*
+                  - s3:List*
+                  - s3:PutObject
+                  - s3:AbortMultipartUpload
+                  - s3:DeleteObject
+                  - s3:PutBucketNotification
+                  - s3:PutBucketPolicy
+                  - s3:DeleteBucketPolicy
```
gmarciani: Why are these s3 permissions required?

tests/integration-tests/resources/file-cache-storage-cfn.yaml
```diff
@@ -0,0 +1,261 @@
...
+                Effect: Allow
+                Resource: '*'
+
+  CleanupLambda:
```
gmarciani: what is the advantage of separating creation and deletion of the same FileCache resource in two distinct functions?
I suggest to have a single Lambda function responsible for the whole resource lifecycle.

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -295,21 +295,45 @@ def check_fsx(
...
-                _test_import_path(remote_command_executor, mount_dir)
-                _test_export_path(remote_command_executor, mount_dir, bucket_name, region)
-                _test_data_repository_task(remote_command_executor, mount_dir, bucket_name, fsx_id, region)
+            mount_name = get_mount_name(fsx_id, region)
+            if fsx_id.startswith("fc-"):
+                assert_fsx_lustre_correctly_mounted(
+                    remote_command_executor,
+                    mount_dir,
+                    mount_name=mount_name,
+                    mount_options=fsx_lustre_mount_options + ",x-systemd.requires=network.service",
```
gmarciani, line 318: *Observation to cover in another PR*: it seems that the mount option `x-systemd.requires=network.service` is required only for FSx FileCache, but it seems reasonable to have it also in pother FSx File Systems.

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -295,21 +295,45 @@ def check_fsx(
...
+                    remote_command_executor,
+                    mount_dir,
+                    mount_name=mount_name,
+                    mount_options=fsx_lustre_mount_options,
+                    fsx_id=fsx_id,
+                )
+                if bucket_name:
+                    _test_import_path(remote_command_executor, mount_dir)
+                    _test_export_path(remote_command_executor, mount_dir, bucket_name, region)
+                    _test_data_repository_task(remote_command_executor, mount_dir, bucket_name, fsx_id, region)
```
gmarciani, line 336: Why is the check `_test_data_repository_task ` not required for FSx File Cache?

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -458,58 +485,98 @@ def _test_data_repository_task(remote_command_executor, mount_dir, bucket_name,
...
+            )
+        )
+        assert_that(export_result.stdout).is_equal_to(
+            "{mount_dir}{file_cache_path}file_to_export: (0x00000009) exists archived, archive_id:1".format(
+                mount_dir=mount_dir, file_cache_path=file_cache_path
+            )
+        )
+
+        remote_command_executor.run_remote_command(
+            "sudo aws s3 cp --region {region} s3://{bucket_name}/file_to_export ./file_to_export".format(
```
gmarciani, line 513: Why is `sudo` required?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -325,6 +327,50 @@ def test_multiple_fsx(
...
+    Test existing Fsx file cache
+
+    Check fsx_fc_id provided in config file can be mounted correctly
+    """
+    # Create s3 bucket and upload test_file
+    bucket_name = s3_bucket_factory()
+    bucket = boto3.resource("s3", region_name=region).Bucket(bucket_name)
+    bucket.upload_file(str(test_datadir / "s3_test_file"), "s3_test_file")
+    partition = utils.get_arn_partition(region)
+    num_existing_fsx_file_caches = (
```
gmarciani: What about exiting the test case when the partition is not "aws"?
Currently, if the partition is not "aws" the test would proceed attempting to create the cluster, but it would not be required.

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -390,6 +436,51 @@ def _create_fsx_open_zfs_volume_ids(num_existing_fsx_open_zfs_volumes, fsx_facto
     return []
 
 
+def _get_fsx_file_cache_id(num_existing_fsx_file_caches, create_fsx_file_cache, file_cache_path, bucket_name):
```
gmarciani: This function is expected to return a single FileCache id, so what about returning None rather than an empty list?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -390,6 +436,51 @@ def _create_fsx_open_zfs_volume_ids(num_existing_fsx_open_zfs_volumes, fsx_facto
...
+    return []
+
+
+@pytest.fixture
+def create_fsx_file_cache(request, region, vpc_stack: CfnVpcStack, cfn_stacks_factory):
+    """Create a FSx File Cache Stack with its required Networking resources."""
+    file_cache_config_file = os.path.join("resources", "file-cache-storage-cfn.yaml")
+
+    def _create_fsx_file_cache(file_cache_path, bucket_name):
+        if request.config.getoption("external_shared_storage_stack_name"):
```
gmarciani, line 446: This option currently refers to the stack deployed as part of DFSM integ tests, i.e. `/aws-parallelcluster/tests/integration-tests/tests/update/test_update/test_dynamic_file_systems_update/storage-stack.yaml`.

It's correct to support the reuse of the stack, but we should have a separated option because every option should refer to a specific stack to avoid ambiguity.

cli/tests/pcluster/templates/test_cw_dashboard_builder/test_cw_dashboard_builder/alinux2.batch.head_node_log.yaml
himani2411: These were tests related to CW Dashboard and there was a miss on tests for other existing storage so just added them.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: Got it.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: Got it.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: Thanks that would be easier.

tests/integration-tests/tests/storage/storage_common.py
himani2411: I thought you mentioned that we are not supposed to check the data repository association of FC and S3? I can add this check as well.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: If we are providing an S3 bucket as a data repository, FSx FC will make a few checks on those S3. if we dont provide these it gives 
```
Amazon FSx is unable to validate access to the S3 bucket.
```

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: The reason i separated the Creation and Deletion was it helped in debugging. But I see your point in combining the two as both the creation/deletion logic is stand alone anyway.

tests/integration-tests/tests/storage/test_fsx_lustre.py
himani2411: Got it.

tests/integration-tests/tests/storage/test_fsx_lustre.py
himani2411: I feel reusing the same option is better rather than creating a separate one. I am using the Fsx Fc stack in the external_shared_storage_stack ( as nested stack) for the DFSM tests too. So why not use this option, this option does handle an external storage.

Also, adding another option would add to the confusion. Since I have adapted the ```test_dynamic_file_systems_update/storage-stack.yaml``` to have a nested stack('tests/integration-tests/resources/file-cache-storage-cfn.yaml'). I have used the ```external_shared_storage_stack_name ```option here again as that stack itself will have a FSx FC ID too. 
NOTE: will add a comment in ```storage-stack.yaml``` pointing to the nested stack's template that we are uploading in S3

tests/integration-tests/tests/storage/storage_common.py
himani2411: So this change in other PR need to be added in cookbook.

tests/integration-tests/resources/file-cache-storage-cfn.yaml
himani2411: My stack was failing the creation of Fsx FC due to insufficient access. In order to unblock myself i added these, will remove them.

tests/integration-tests/tests/storage/storage_common.py
himani2411: As discussed offline, we dont need to test this for fsx FC.

tests/integration-tests/tests/storage/test_fsx_lustre.py
himani2411: I will make these changes as Initially the code was written so that i add Fsx FC in ```test_multiple_fsx``` rather than as a standalone.

tests/integration-tests/tests/storage/test_fsx_lustre.py
gmarciani: > external_shared_storage_stack ( as nested stack) for the DFSM tests too

Did not notice that, then we are fine.

tests/integration-tests/tests/storage/storage_common.py
himani2411: The user in different OS does not have access to upload the files. So we use sudo

gmarciani (approved):

PR #5294, opened 2023-05-10
[3.6] Remove the need of cdk arn class in PasswordSecretArnValidator

eantonin:
### Description of changes
* Remove the need of cdk arn class in PasswordSecretArnValidator

### Tests
* The JUnit excercising that part were already present and they are still passing
* Launched 8 clusters with the custom resource, before the fix this would have almost surely showed at least an error, while it is not showing any error now, all validations succeeded

### References
* The root cause is probably a race condition between the CDK import in PasswordSecretArnValidator and the one in the thread added in https://github.com/aws/aws-parallelcluster/pull/4778 which runs in parallel with the validation.
* PasswordSecretArnValidator was the only validator importing cdk

### Checklist

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -81,16 +81,17 @@ class PasswordSecretArnValidator(Validator):
          2. a readable parameter in SSM Parameter Store, which is supported only in us-isob-east-1.
         """
         try:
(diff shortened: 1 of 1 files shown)


gmarciani (approved):

enrico-usai (approved):
Very great job finding the root cause of the issue, nice explanation!

PR #5309, opened 2023-05-12
3.x Remove cluster health metrics from ADC regions

davprat:
### Description of changes
* Some CloudWatch CloudFormation resources are not supported in the ADC regions.
* Do not try to create these resources if the cluster is being created in one of the ADC regions.

### Tests
* Added test to validate resources don't get added to the cluster template if the region is an ADC region.

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cw_dashboard_builder.py
+++ b/cli/src/pcluster/templates/cw_dashboard_builder.py
@@ -16,6 +16,7 @@ from aws_cdk import aws_ec2 as ec2
 from aws_cdk import aws_logs as logs
 from aws_cdk.core import Construct, Duration, Stack
 
+from pcluster.aws.common import get_region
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/templates/cw_dashboard_builder.py
```diff
@@ -147,7 +152,7 @@ def _add_resources(self):
 
         # Head Node logs add custom metrics if cw_log and metrics are enabled
         if self.config.is_cw_logging_enabled:
-            if self.config.scheduling.scheduler == "slurm":
+            if self.config.scheduling.scheduler == "slurm" and is_region_supported(get_region()):
```
chenwany, line 155: Why do we use get_region here instead of _stack_region?

davprat: Stack region was the Cfn token representing the region of the stack I think?

chenwany (approved):
lgtm

PR #5314, opened 2023-05-13
Introduce Slurm compute node priority customization

jdeamicis:
### Description of changes
* Introduce StaticNodePriority and DynamicNodePriority configuration parameters at Compute Resource level.

### Tests

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2127

### Checklist


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1959,6 +1962,8 @@ def __init__(
         self.health_checks = health_checks or HealthChecks(implied=True)
         self.custom_slurm_settings = Resource.init_param(custom_slurm_settings, default={})
         self.tags = tags
+        self.static_node_priority = Resource.init_param(static_node_priority, default=1)
+        self.dynamic_node_priority = Resource.init_param(dynamic_node_priority, default=1000)
```
NSsirena, line 1966: [Nitpick] For coherence we should define these constants in `constants.py` (see DEFAULT_MAX_COUNT a few lines above)

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -144,3 +144,36 @@ def _validate(self, custom_settings: List[Dict], include_file_url: str):
...
+    """
+
+    def _validate(self, queue_name: str, compute_resources: List[Dict]):
+        static_priorities = {cr.name: cr.static_node_priority for cr in compute_resources}
+        dynamic_priorities = {cr.name: cr.dynamic_node_priority for cr in compute_resources}
+
+        bad_static_priorities = {
+            key: value
+            for key, value in static_priorities.items()
+            if value in range(min(dynamic_priorities.values()), max(dynamic_priorities.values()) + 1)
```
NSsirena: These functions are evaluated once for each `static priorities` and below once per `dynamic priority`.
Moreover we aren't interested in the full range but only in on boundary: `max_static` and `min_dynamic`
and we want to ensure that "all static are <= min_dynamic" and "all dynamic are >= max_static".

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -144,3 +144,36 @@ def _validate(self, custom_settings: List[Dict], include_file_url: str):
...
+            key: value
+            for key, value in dynamic_priorities.items()
+            if value in range(min(static_priorities.values()), max(static_priorities.values()) + 1)
+        }
+
+        if bad_static_priorities or bad_dynamic_priorities:
+            self._add_failure(
+                f"Some compute resources in queue {queue_name} have static nodes with higher or equal priority than "
+                f"other dynamic nodes in the same queue. "
+                f"Possible problematic static node priorities are {bad_static_priorities}. "
```
NSsirena: Following the previous comment, we can enrich this message reporting also the boundaries extracted above (max_static and min_dynamic). This would make it easier to review the "offending" values.

cli/src/pcluster/config/cluster_config.py
jdeamicis: I thought about it, and I didn't do it for the following reasons:
- there are several other places where the default is simply written in the parameter definition, without defining a corresponding constant (I agree we should have more consistency on this in the code);
- if a default is used only once in the config and is independent from other default values, I believe defining a constant is pointless and also slightly annoying, since it forces you to follow a redirection to understand what the default value is.

cli/src/pcluster/validators/slurm_settings_validator.py
jdeamicis: Oh, you mean calculating first the ranges and using them "statically" in the if condition?

Honest doubt here: wouldn't Python's interpreter try to optimize something here? But I agree we can be stricter.

cli/src/pcluster/validators/slurm_settings_validator.py
jdeamicis: The current message is already providing this information, although not explicitly as you are saying.

What is being printed at the moment are all the values of priority (both static and dynamic) that go against the "soft rule" of having all dynamic priorities higher than the static ones within the same queue. So the max static and min dynamic are already included in the printed values (they will be respectively the last value of static priority and the first value of dynamic priority being printed in the message).

cli/src/pcluster/validators/slurm_settings_validator.py
jdeamicis: Addressed, and we saved some vertical space in the code :)

cli/src/pcluster/validators/slurm_settings_validator.py
NSsirena: Yes. I saw the examples in the tests.
However if you have a long list of Compute Resources is may be confusing and not immediately obvious which value you have to pick from which list..because you have to mentally rebuild the logic behind the check.

While in the other way you can immediately tell the customers which is the "offending threshold" in both cases.
Also you can avoid defining the full range, since what you need is only the min or the max of the range .

Anyway, I'll approve the PR since it is a sort of Nitpick.

NSsirena (dismissed):

cli/src/pcluster/config/cluster_config.py
NSsirena: ok

cli/src/pcluster/validators/slurm_settings_validator.py
```diff
@@ -144,3 +145,34 @@ def _validate(self, custom_settings: List[Dict], include_file_url: str):
...
+    nodes and throws a warning if that's the case.
+    """
+
+    def _validate(self, queue_name: str, compute_resources: List[Dict]):
+        static_priorities = {cr.name: cr.static_node_priority for cr in compute_resources if cr.min_count > 0}
+        dynamic_priorities = {
+            cr.name: cr.dynamic_node_priority for cr in compute_resources if cr.max_count > cr.min_count
+        }
+        max_static = max(static_priorities.values()) if len(static_priorities) > 0 else MIN_SLURM_NODE_PRIORITY - 1
+        min_dynamic = min(dynamic_priorities.values()) if len(dynamic_priorities) > 0 else MAX_SLURM_NODE_PRIORITY + 1
```
NSsirena: Please add a comment that explains that we are initializing those values to `impossible to achieve` thresholds because we want to ensure the the warning is not triggered when either static or dynamic nodes are not present in the configuration.

jdeamicis: Addressed.

NSsirena (approved):

PR #5317, opened 2023-05-15
[Multi User] Enable support for Active Directory password in Secrets Manager for region us-isob-east-1.

gmarciani:
### Description of changes
Enable support for Active Directory password in Secrets Manager for region us-isob-east-1.

### Tests
1. PENDING Cluster creation succeeded in us-isob-east-1 with password stored in Secrets Manager.

### DIFF
--- a/cli/src/pcluster/validators/directory_service_validators.py
+++ b/cli/src/pcluster/validators/directory_service_validators.py
@@ -77,8 +77,9 @@ class PasswordSecretArnValidator(Validator):
         """Validate that PasswordSecretArn contains a valid ARN for the given region.
 
         In particular, the ARN should be one of the following resources:
-         1. a readable secret in AWS Secrets Manager, which is supported in all regions but us-isob-east-1.
-         2. a readable parameter in SSM Parameter Store, which is supported only in us-isob-east-1.
+         1. a readable secret in AWS Secrets Manager, which is supported in all regions.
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -88,7 +89,7 @@ def _validate(self, password_secret_arn: str, region: str):
             if service == "ssm":
                 resource = arn_components[5].split("/")[0]
 
-            if service == "secretsmanager" and resource == "secret" and region != "us-isob-east-1":
+            if service == "secretsmanager" and resource == "secret":
                 AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
             elif service == "ssm" and resource == "parameter" and region == "us-isob-east-1":
```
hanwen-pcluste, line 94: Why is the `and region == "us-isob-east-1"` required?

gmarciani: we are removing that check. It was required because in us.isob-east.1 we were not able to use SM, and now we are.

gmarciani: If you are referring to line 94, then the reason to keep it is backward compatibility since in 3.5.1 we provided the support for password stored in SSM Parameter

hanwen-pcluste (approved):

PR #5325, opened 2023-05-16
[release-3.6] Permit tags prefix validators to work if a tag is defined as a boolean

enrico-usai:
### Description of changes

If one tag in the config has a boolean value (e.g. `YES`), `tags` variable will be a list of dict. If all of them are strings, `tags` variable will be a list of `BaseTag` objects.

This patch improves the `validate_no_duplicate_tag` to be able to work in both the cases.

### Tests

Previously, when using a config like:
```
Tags:
   - Key: WrongTagValue
     Value: YES
```
the cluster creation was failing with an error:
```
"message": "Invalid cluster configuration: 'dict' object has no attribute 'key'"
```

I replicated the issue within the unit tests and then fixed it.

Now, at cluster creation time, the validator is executed correctly and the Schema parsing fails with a meaningful message:
```
    "type": "ConfigSchemaValidator",
    "message": "[('Tags', {1: {'Value': ['Not a valid string.']}})]"
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ CHANGELOG
 - Fix issue causing `cfn-hup` daemon to fail when it gets restarted.
 - Fix issue causing dangling security groups to be created when creating a cluster with an existing EFS.
 - Fix issue causing NVIDIA GPU compute nodes not to resume correctly after executing an `scontrol reboot` command.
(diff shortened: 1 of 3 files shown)


eantonin (dismissed):
LGTM

eantonin (approved):

PR #5344, opened 2023-05-23
Allow Memory Based Scheduling and FIT

NSsirena:
Change the Validator to allow Memory Based Scheduling and Flexible Instance Types to be used at the same time.

A warning is triggered if the Memory difference between the smallest and biggest instances is greater than an absolute threshold and a percentage threshold.

### Tests
* Unit test run locally

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
 **ENHANCEMENTS**
 - Allow configuration of static and dynamic node priorities in compute resources via the ParallelCluster configuration YAML file.
 - Add support for Ubuntu 22.
+- Allow memory-based scheduling when multiple instance types are specified for a Slurm Compute Resource.
 
 **CHANGES**
 
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -110,6 +110,10 @@ MAX_NUMBER_OF_COMPUTE_RESOURCES_PER_CLUSTER = MAX_COMPUTE_RESOURCES_PER_QUEUE = 
 MIN_SLURM_NODE_PRIORITY = 1
 MAX_SLURM_NODE_PRIORITY = 2**32 - 1  # max value of uint32_t
 
+# Thresholds used to trigger a warning if Memory Based Scheduling and Flexible Instance Types are used together
+MIN_MEMORY_ABSOLUTE_DIFFERENCE = 4096
+MIN_MEMORY_PRECENTAGE_DIFFERENCE = 0.20
+
 MAX_EBS_COUNT = 5
 MAX_NEW_STORAGE_COUNT = {"efs": 1, "fsx": 1, "raid": 1}
 MAX_EXISTING_STORAGE_COUNT = {"efs": 20, "fsx": 20, "raid": 0}
--- a/cli/src/pcluster/validators/instances_validators.py
+++ b/cli/src/pcluster/validators/instances_validators.py
@@ -13,6 +13,7 @@ from typing import Callable, Dict
 
 from pcluster.aws.aws_resources import InstanceTypeInfo
 from pcluster.config import cluster_config
+from pcluster.constants import MIN_MEMORY_ABSOLUTE_DIFFERENCE, MIN_MEMORY_PRECENTAGE_DIFFERENCE
 from pcluster.validators.common import FailureLevel, Validator
 
 
@@ -233,8 +234,13 @@ class InstancesAllocationStrategyValidator(Validator, _FlexibleInstanceTypesVali
             )
 
 
-class InstancesMemorySchedulingValidator(Validator, _FlexibleInstanceTypesValidatorMixin):
-    """Validate support for Memory-based Scheduling when using Flexible Instance Types."""
+class InstancesMemorySchedulingWarningValidator(Validator):
+    """
+    Memory-based Scheduling Warning Validator.
+
+    This validator checks if Memory-based Scheduling is enabled when also using Flexible Instance Types.
+    In this case warns the customer about possible unused resources due to the differences in the instance types.
+    """
 
     def _validate(
         self,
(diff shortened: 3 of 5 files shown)


cli/src/pcluster/validators/instances_validators.py
```diff
@@ -233,21 +234,41 @@ def _validate(self, compute_resource_name: str, capacity_type: Enum, allocation_
...
         compute_resource_name: str,
         instance_types_info: Dict[str, InstanceTypeInfo],
         memory_scheduling_enabled: bool,
         **kwargs,
     ):
-        """Memory-based scheduling is NOT supported for Compute Resources with multiple instance types."""
-        if memory_scheduling_enabled and len(instance_types_info.items()) > 1:
+        """Enabling Memory-based scheduling with multiple instance types may lead to unused resources."""
+        if memory_scheduling_enabled \
+                and len(instance_types_info.items()) > 1 \
```
jdeamicis: Nitpick: isn't it more _pythonic_ to open the parenthesis and avoid the backslashes?

cli/src/pcluster/validators/instances_validators.py
```diff
@@ -243,11 +249,28 @@ def _validate(
...
+
+    def memory_difference_exceeds_thresholds(self, instance_types_info):
+        """Return True only if both thresholds are exceeded."""
+        min_memory, max_memory = self._min_max_memory(instance_types_info)
+        absolute_diff = max_memory - min_memory
+        percent_diff = absolute_diff / min_memory
+
+        return absolute_diff > MIN_MEMORY_ABSOLUTE_DIFFERENCE and percent_diff > MIN_MEMORY_PRECENTAGE_DIFFERENCE
+
+    def _min_max_memory(self, instance_types_info):
```
jdeamicis: Nitpick: can you please add typing for the `instance_types_info` object there? I though it was a dictionary, but it looks it's a more complex object with some internal methods.

cli/tests/pcluster/validators/test_instances_validators.py
```diff
@@ -499,21 +499,72 @@ def test_instances_allocation_strategy_validator(
...
 
 
+# Memory-based scheduling is allowed for Compute Resource that use multiple instance type under 'Instances'
+# but a warning is triggered to inform customers of possible wasted resources.
 @pytest.mark.parametrize(
     "compute_resource_name, instance_types_info, memory_scheduling_enabled, expected_message",
     [
-        # Memory-based scheduling is supported for Compute Resource that use either 'InstanceType' or have a single
-        # instance type under 'Instances'
+        # Memory Diff exceeds both Absolute (2G) and Percentage (0.25) threshold, so Warning is triggered.
```
jdeamicis: Nitpick: the other day I found out about the [`pytest.param()`](https://github.com/aws/aws-parallelcluster/blob/947f0c6f6b54d11b19a011d668f415b149e35293/cli/tests/pcluster/validators/test_cluster_validators.py#L326) we already use in some unit tests, where we can pass an `id` attribute to describe the case instead of relying on comments. The `id` gets rendered in the stdout/stderr of pytest when running the tests.

cli/src/pcluster/validators/instances_validators.py
```diff
@@ -243,11 +249,28 @@ def _validate(
...
+                FailureLevel.WARNING,
             )
+
+    def memory_difference_exceeds_thresholds(self, instance_types_info):
+        """Return True only if both thresholds are exceeded."""
+        min_memory, max_memory = self._min_max_memory(instance_types_info)
+        absolute_diff = max_memory - min_memory
+        percent_diff = absolute_diff / min_memory
+
+        return absolute_diff > MIN_MEMORY_ABSOLUTE_DIFFERENCE and percent_diff > MIN_MEMORY_PRECENTAGE_DIFFERENCE
```
jdeamicis, line 270: Nitpick: maybe `>=` instead of `>`?

cli/src/pcluster/constants.py
```diff
@@ -108,6 +108,10 @@
 MIN_SLURM_NODE_PRIORITY = 1
 MAX_SLURM_NODE_PRIORITY = 2**32 - 1  # max value of uint32_t
 
+# Thresholds used to trigger a warning if Memory Based Scheduling and Flexible Instance Types are used together
+MIN_MEMORY_ABSOLUTE_DIFFERENCE = 2048
+MIN_MEMORY_PRECENTAGE_DIFFERENCE = 0.25
```
jdeamicis: Also, I would expect fairly large instances are more relevant for HPC, meaning that even a small relative difference can result in fairly high absolute differences. I would make this 20 %, maybe even less. To be discussed internally.

cli/src/pcluster/constants.py
```diff
@@ -108,6 +108,10 @@
 MIN_SLURM_NODE_PRIORITY = 1
 MAX_SLURM_NODE_PRIORITY = 2**32 - 1  # max value of uint32_t
 
+# Thresholds used to trigger a warning if Memory Based Scheduling and Flexible Instance Types are used together
+MIN_MEMORY_ABSOLUTE_DIFFERENCE = 2048
```
jdeamicis: I would probably put 4 GiB here.

jdeamicis (commented):
Left some comments, some of which to be discussed internally.

cli/src/pcluster/constants.py
NSsirena: I considered both thresholds are "placeholders" for now.
We should have an internal discussion to decide which are meaningful values.

Anyway I'll change them to the value you suggested and then we will iterate if needed.

cli/src/pcluster/constants.py
NSsirena: See comment above

cli/src/pcluster/validators/instances_validators.py
NSsirena: It's a threshold. 
I think a strict check it's better.

cli/src/pcluster/validators/instances_validators.py
NSsirena: It is a dictionary. 
I'll add the type to make it more clear.

cli/tests/pcluster/validators/test_instances_validators.py
NSsirena: Cool!

I'll definitely change this.

cli/src/pcluster/validators/instances_validators.py
NSsirena: It is a dictionary. 
I'll add the type to make it more clear.

cli/src/pcluster/validators/instances_validators.py
```diff
@@ -243,11 +249,28 @@ def _validate(
...
+            and self.memory_difference_exceeds_thresholds(instance_types_info)
+        ):
             self._add_failure(
-                "Memory-based scheduling is only supported for Compute Resources using either 'InstanceType' or "
-                f"'Instances' with one instance type. Compute Resource {compute_resource_name} has more than "
-                "one instance type specified.",
-                FailureLevel.ERROR,
+                f'Enabling Memory-based scheduling when a Compute Resource ("{compute_resource_name}") has more than'
+                " one instance type specified may lead to unused resources since only the minimum available memory"
+                " across all instance-types can be specified in the Slurm node definition.",
```
jdeamicis, line 260: I guess this text is also a "placeholder" and should be reviewed internally, right?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
 **ENHANCEMENTS**
 - Allow configuration of static and dynamic node priorities in compute resources via the ParallelCluster configuration YAML file.
+- Allow memory-based scheduling when multiple instance types are specified for a Compute Resource.
```
jdeamicis: Please specify "Slurm Compute Resource".

jdeamicis (dismissed):

cli/src/pcluster/validators/instances_validators.py
NSsirena: Yes

CHANGELOG.md
NSsirena: Fixed

jdeamicis (approved):

PR #5349, opened 2023-05-24
[Isolated Regions][Test] Make 'test_create_wrong_pcluster_version' executable in us-iso regions + fix regression on Jinja syntax impacting EFS tests

gmarciani:
### Description of changes
1. Make 'test_create_wrong_pcluster_version' executable in us-iso regions. the test was disabled in US ISO regions because it assumed the AMI for PC 2.8.1 to be available in the region, but this is not the case for these regions because the oldest AMI is 3.5.1 there.
1. Fix regression on 'test_efs_compute_az', 'test_multiple_efs' and 'test_multi_az_create_and_update' caused by Jinja syntax error.

### Tests
Will be validated in US iso regions.

### DIFF
--- a/tests/integration-tests/tests/create/test_create.py
+++ b/tests/integration-tests/tests/create/test_create.py
@@ -55,7 +55,7 @@ def test_create_wrong_pcluster_version(
 ):
     """Test error message when AMI provided was baked by a pcluster whose version is different from current version"""
     current_version = get_installed_parallelcluster_version()
-    wrong_version = "2.8.1"
+    wrong_version = "3.5.1"
     logging.info("Asserting wrong_version is different from current_version")
     assert_that(current_version != wrong_version).is_true()
     # Retrieve an AMI without 'aws-parallelcluster-<version>' in its name.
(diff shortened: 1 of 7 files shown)


tests/integration-tests/tests/create/test_create.py
```diff
@@ -55,7 +55,9 @@ def test_create_wrong_pcluster_version(
 ):
     """Test error message when AMI provided was baked by a pcluster whose version is different from current version"""
     current_version = get_installed_parallelcluster_version()
-    wrong_version = "2.8.1"
+    # ParallelCluster 3.5.1 is the first version released in US Isolated regions,
+    # whereas 2.8.1 is an old version in all the other regions.
+    wrong_version = "3.5.1" if "us-iso" in region else "2.8.1"
```
hanwen-pcluste: Instead of the conditional statement, let's just use 3.5.1 for all

gmarciani: If we do that way, then integ test for 3.5.1 in Commercial would fail

hanwen-pcluste: It is fine. 3.5.1 tests are not running anyway

hanwen-pcluste (approved):

PR #5350, opened 2023-05-24
Add integration test for MemoryBased Scheduling with Multiple Instance Types

NSsirena:
### Description of changes
* Add integration test for MemoryBased Scheduling with Multiple Instance Types
* Add SchedulableMemoryValidator also to FlexibleSlurmComputeResource

### Tests
* Successfully launched integ-test on my account
* Tested the validator for Flexible resources adding an invalid amount of Schedulable memory returned

```
    {
      "level": "ERROR",
      "type": "SchedulableMemoryValidator",
      "message": "SchedulableMemory cannot be larger than EC2 Memory for selected instance type c5.xlarge (8192 MiB)."
    },
```

### References
* [Allow Memory Based Scheduling and FIT](https://github.com/aws/aws-parallelcluster/pull/5344)

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2046,6 +2046,21 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
             for flexible_instance_type in self.instances  # pylint: disable=not-an-iterable
         ]
 
+    def _min_schedulable_memory_and_instance_type(self):
+        instances_and_memory = {t: info.ec2memory_size_in_mib() for t, info in self.instance_type_info_map.items()}
+        smallest_type = min(instances_and_memory, key=instances_and_memory.get)
+        return smallest_type, instances_and_memory[smallest_type]
+
+    def _register_validators(self, context: ValidatorContext = None):
+        super()._register_validators(context)
+        smallest_type, min_memory = self._min_schedulable_memory_and_instance_type()
+        self._register_validator(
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -2046,6 +2046,21 @@ def instance_types(self) -> List[str]:
             for flexible_instance_type in self.instances  # pylint: disable=not-an-iterable
         ]
 
+    def _min_schedulable_memory_and_instance_type(self):
+        instances_and_memory = {t: info.ec2memory_size_in_mib() for t, info in self.instance_type_info_map.items()}
+        smallest_type = min(instances_and_memory, key=instances_and_memory.get)
+        return smallest_type, instances_and_memory[smallest_type]
+
+    def _register_validators(self, context: ValidatorContext = None):
+        super()._register_validators(context)
```
jdeamicis, line 2055: Please double-check if this call to the super's `_register_validators()` is necessary.

jdeamicis (approved):
Left a minor comment about the usage of the parent class `_register_validators()`.

cli/src/pcluster/config/cluster_config.py
NSsirena: I saw it used in many Objects and assumed it was a best practice to add it.
I agree that it is not clear if/when it should be used but I'll consider it out-of-scope for this PR.

I'll create an investigation task for diving deep on this.

PR #5358, opened 2023-05-27
Add login nodes configuration parameters

hehe7318:
### Description of changes
* Create classes to cluster_config.py with full configuration parameters without update policy:
* Create above classes’ schema to cluster_schema.py with validators
* Create LoginNodeIam class in cluster_config.py and for better maintainability
        * Create _BaseIam class with the fields in common between all the Iams
        * Modify Iam class to become HeadNodeAndQueueIam inherited from _BaseIam class
        * Make LoginNodeIam class inherited from _BaseIam class
* Centralise AMI id Regexp in [constants](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/constants.py) file for better maintainability
* Create BaseIam related classes in cluster_config.py and for better maintainability
        * Make LoginNodesIam and HeadNodeIam inherit from BaseIam
* Modify DummySlurmClusterConfig to add support for login nodes
* Unit test
    * Test cluster config
        * test login node pool default value
    * Test cluster schema validators
        * Add unit test to make sure the custom ami is validate
        * Add unit test to make sure login node pool count must be greater than zero
        * Add unit test to make sure login node pool is validate
        * Add unit test to confirm for the MVP, only 1 pool can be under the LoginNodes section
        * Add unit test to confirm that the instance profile and instance role cannot be used together.
    * Modify yaml file to add support for login nodes

### Sample YAML configuration
```
LoginNodes:
  Pools: 
  - Name: test
    InstanceType: t2.micro
    Count: 1
    Image:
        CustomAmi: ami-12345678
    Networking:
      SubnetId: subnet-12345678901234567
    Ssh:
      KeyName: test-us-east-2
    Iam:
```

### Tests

* Unit test run locally

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2197,9 +2197,9 @@ class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
     def __init__(
-            self,
-            instance_type,
-            **kwargs,
+        self,
+        instance_type,
+        **kwargs,
     ):
         super().__init__(**kwargs)
         self.instance_type = Resource.init_param(instance_type)
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -114,8 +114,6 @@ from pcluster.validators.cluster_validators import (
     SharedStorageMountDirValidator,
     SharedStorageNameValidator,
     UnmanagedFsxMultiAzValidator,
-    PoolsValidator,
-    AvailabilityZoneValidator,
 )
 from pcluster.validators.common import ValidatorContext, get_async_timed_validator_type_for
 from pcluster.validators.database_validators import DatabaseUriValidator
@@ -184,6 +182,7 @@ from pcluster.validators.networking_validators import (
     SecurityGroupsValidator,
     SingleInstanceTypeSubnetValidator,
     SubnetsValidator,
+    AvailabilityZoneValidator,
 )
 from pcluster.validators.s3_validators import (
     S3BucketRegionValidator,
@@ -666,8 +665,6 @@ class _BaseNetworking(Resource):
 class HeadNodeNetworking(_BaseNetworking):
     """Represent the networking configuration for the head node."""
 
-    _instance = None
-
     def __init__(self, subnet_id: str, elastic_ip: Union[str, bool] = None, proxy: Proxy = None, **kwargs):
         super().__init__(**kwargs)
         self.subnet_id = Resource.init_param(subnet_id)
@@ -683,12 +680,6 @@ class HeadNodeNetworking(_BaseNetworking):
     def availability_zone(self):
         """Compute availability zone from subnet id."""
         return AWSApi.instance().ec2.get_subnet_avail_zone(self.subnet_id)
-
-    @staticmethod
-    def getInstance():
-        if HeadNodeNetworking._instance is None:
-            raise Exception("HeadNode must be instantiated first!")
-        return HeadNodeNetworking._instance
 
 
 class PlacementGroup(Resource):
@@ -924,19 +915,16 @@ class S3Access(Resource):
             return [self.bucket_name, f"{self.bucket_name}/*"]
 
 
-class Iam(Resource):
-    """Represent the IAM configuration for HeadNode and Queue."""
-
+class _BaseIam(Resource):
+    """Represent the base IAM configuration."""
     def __init__(
-        self,
-        s3_access: List[S3Access] = None,
-        additional_iam_policies: List[AdditionalIamPolicy] = (),
-        instance_role: str = None,
-        instance_profile: str = None,
-        **kwargs,
+            self,
+            additional_iam_policies: List[AdditionalIamPolicy] = (),
+            instance_role: str = None,
+            instance_profile: str = None,
+            **kwargs,
     ):
         super().__init__(**kwargs)
-        self.s3_access = s3_access
         self.additional_iam_policies = additional_iam_policies
         self.instance_role = Resource.init_param(instance_role)
         self.instance_profile = Resource.init_param(instance_profile)
@@ -981,6 +969,27 @@ class Iam(Resource):
             self._register_validator(RoleValidator, role_arn=self.instance_role)
         elif self.instance_profile:
             self._register_validator(InstanceProfileValidator, instance_profile_arn=self.instance_profile)
+
+
+class HeadNodeAndQueueIam(_BaseIam):
+    """Represent the IAM configuration for HeadNode and Queue."""
+
+    def __init__(
+        self,
+        s3_access: List[S3Access] = None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.s3_access = s3_access
+
+
+class LoginNodeIam(_BaseIam):
+    """Represent the IAM configuration for HeadNode and Queue."""
+    def __init__(
+        self,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
 
 
 class Imds(Resource):
@@ -1216,9 +1225,6 @@ class LoginNodeSsh(Ssh):
     ):
         super().__init__(key_name=key_name)
 
-    def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
-        super()._register_validators(context)
-
 
 class LoginNodeNetworking(_BaseNetworking):
     """Represent the networking configuration for the login node."""
@@ -1226,23 +1232,10 @@ class LoginNodeNetworking(_BaseNetworking):
     def __init__(
             self,
             subnet_id: str,
-            security_groups: List[str] = None,
-            additional_security_groups: List[str] = None,
             **kwargs,
     ):
-        super().__init__(security_groups=security_groups, additional_security_groups=additional_security_groups,
-                         **kwargs)
+        super().__init__(**kwargs)
         self.subnet_id = Resource.init_param(subnet_id)
-
-    def _register_validators(self, context: ValidatorContext = None):
-        super()._register_validators(context)
-        self._register_validator(AvailabilityZoneValidator, login_node_subnet_id=self.subnet_id,
-                                 head_node_subnet_id=HeadNodeNetworking.getInstance().subnet_id)
-
-    @property
-    def availability_zone(self):
-        """Compute availability zone from subnet id."""
-        return AWSApi.instance().ec2.get_subnet_avail_zone(self.subnet_id)
 
 
 class LoginNodePool(Resource):
@@ -1256,9 +1249,7 @@ class LoginNodePool(Resource):
             networking: LoginNodeNetworking = None,
             count: int = None,
             ssh: LoginNodeSsh = None,
-            iam: Iam = None,
-            gracetime_period: int = None,
-            admin_user: str = None,
+            iam: LoginNodeIam = None,
             **kwargs,
     ):
         super().__init__(**kwargs)
@@ -1268,9 +1259,7 @@ class LoginNodePool(Resource):
         self.networking = networking
         self.count = Resource.init_param(count, default=1)
         self.ssh = ssh
-        self.iam = iam or Iam(implied=True)
-        self.gracetime_period = Resource.init_param(gracetime_period, default=60)
-        self.admin_user = Resource.init_param(admin_user, default=getpass.getuser())
+        self.iam = iam or LoginNodeIam(implied=True)
 
     def _register_validators(self, context: ValidatorContext = None):
         self._register_validator(InstanceTypeValidator, instance_type=self.instance_type)
@@ -1285,10 +1274,7 @@ class LoginNodes(Resource):
             **kwargs,
     ):
         super().__init__(**kwargs)
-        self.pools = pools or []
-
-    def _register_validators(self, context: ValidatorContext = None):
-        self._register_validator(PoolsValidator, pools=self.pools)
+        self.pools = pools
 
 
 class HeadNode(Resource):
@@ -1303,7 +1289,7 @@ class HeadNode(Resource):
         local_storage: LocalStorage = None,
         dcv: Dcv = None,
         custom_actions: CustomActions = None,
-        iam: Iam = None,
+        iam: HeadNodeAndQueueIam = None,
         imds: Imds = None,
         image: HeadNodeImage = None,
     ):
@@ -1317,7 +1303,7 @@ class HeadNode(Resource):
         self.local_storage = local_storage or LocalStorage(implied=True)
         self.dcv = dcv
         self.custom_actions = custom_actions
-        self.iam = iam or Iam(implied=True)
+        self.iam = iam or HeadNodeAndQueueIam(implied=True)
         self.imds = imds or Imds(implied=True)
         self.image = image
         self.__instance_type_info = None
@@ -1416,7 +1402,6 @@ class BaseClusterConfig(Resource):
         head_node: HeadNode,
         scheduling=None,
         shared_storage: List[Resource] = None,
-        login_nodes: LoginNodes = None,
         monitoring: Monitoring = None,
         additional_packages: AdditionalPackages = None,
         tags: List[Tag] = None,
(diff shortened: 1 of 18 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1257,6 +1262,66 @@ def make_resource(self, data, **kwargs):
         return HeadNode(**data)
 
 
+class LoginNodeImageSchema(BaseSchema):
+    custom_ami = fields.Str(validate=validate.Regexp(r"^ami-[0-9a-z]{8}$|^ami-[0-9a-z]{17}$"))
```
EddyMM: The regular expression used to check if an AMI id is valid seems to be similar for the `HeadNodeImageSchema` and `QueueImageSchema`.
Let's centralise it in a [constants](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/constants.py) file for better maintainability.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1257,6 +1262,66 @@ def make_resource(self, data, **kwargs):
...
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return LoginNodeSsh(**data)
+
+
+class LoginNodeNetworkingSchema(BaseNetworkingSchema):
+    subnet_id = fields.Str(required=True)
+    elastic_ip = fields.Str()
+    proxy = fields.Str()
```
EddyMM: I don't think these parameters are being used (for now).

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1183,6 +1195,102 @@ def __init__(
...
+
+
+class LoginNodeNetworking(_BaseNetworking):
+    """Represent the networking configuration for the login node."""
+
+    def __init__(
+            self,
+            subnet_id: str,
+            security_groups: List[str] = None,
+            additional_security_groups: List[str] = None,
```
EddyMM: I think the `security_groups` and `additional_security_groups` arguments can be consideed to be part of the `kwargs` argument.
We can then let the base class [_BaseNetworking](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/config/cluster_config.py#L655) read them.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1470,3 +1470,22 @@ def _validate(self, queues):
                     f"public IPs can only be assigned to instances launched with a single network interface.",
                     FailureLevel.ERROR,
                 )
+
+
+class PoolsValidator(Validator):
```
EddyMM: This can be verified at Schema level using the [marshmallow Length validator](https://marshmallow.readthedocs.io/en/stable/marshmallow.validate.html#marshmallow.validate.Length)

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1183,6 +1195,102 @@ def __init__(
...
+                         **kwargs)
+        self.subnet_id = Resource.init_param(subnet_id)
+
+    def _register_validators(self, context: ValidatorContext = None):
+        super()._register_validators(context)
+        self._register_validator(AvailabilityZoneValidator, login_node_subnet_id=self.subnet_id,
+                                 head_node_subnet_id=HeadNodeNetworking.getInstance().subnet_id)
+
+    @property
+    def availability_zone(self):
```
EddyMM: Where is this property used? If it's unused then we can remove it.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1257,6 +1262,66 @@ def make_resource(self, data, **kwargs):
...
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return LoginNodeNetworking(**data)
+
+
+class LoginNodePoolSchema(BaseSchema):
+    name = fields.Str(required=True, validate=[validate.Length(min=1), validate.Regexp(r'^[a-z0-9]+$')])
```
EddyMM: The Length=1 validation should probably be in the `LoginNodesSchema`.
Also, let's remove this regexp for now (to be consistent with other naming checks).

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1257,6 +1262,66 @@ def make_resource(self, data, **kwargs):
...
+    instance_type = fields.Str(required=True)
+    image = fields.Nested(LoginNodeImageSchema)
+    networking = fields.Nested(LoginNodeNetworkingSchema, required=True)
+    count = fields.Int(required=True, validate=validate.Range(min=1))
+    ssh = fields.Nested(LoginNodeSshSchema, required=True)
+    iam = fields.Nested(IamSchema)
+    gracetime_period = fields.Int(validate=validate.Range(max=120))
+    admin_user = fields.Str(required=True)
```
EddyMM: Let's remove these parameters for now (let's sync on this).

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1183,6 +1195,102 @@ def __init__(
...
+        super().__init__(**kwargs)
+        self.name = Resource.init_param(name)
+        self.instance_type = Resource.init_param(instance_type)
+        self.image = image
+        self.networking = networking
+        self.count = Resource.init_param(count, default=1)
+        self.ssh = ssh
+        self.iam = iam or Iam(implied=True)
+        self.gracetime_period = Resource.init_param(gracetime_period, default=60)
+        self.admin_user = Resource.init_param(admin_user, default=getpass.getuser())
```
EddyMM: Let's remove these parameters for now.

EddyMM (changes requested):
Good job on your first PR in this repo!

I've left a few comments. Plus there are some failing unit tests that need to be fixed e.g. [CLI tests](https://github.com/aws/aws-parallelcluster/actions/runs/5097687429/jobs/9164363957?pr=5358)

cli/src/pcluster/config/cluster_config.py
hehe7318: Done! I have removed them.

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done! I have removed them.

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: I added Length(min=1) validation because I thought the length of the 'name' string should be longer than 0. But I have removed it. And also removed the regexp.

cli/src/pcluster/config/cluster_config.py
hehe7318: It was originally intended for availability_zone validator, but I modified the validator implementation so it's not needed now, and I forgot to remove it. Now removed.

cli/src/pcluster/validators/cluster_validators.py
hehe7318: Done. I have moved it to Schema level.

cli/src/pcluster/config/cluster_config.py
hehe7318: That make sense. My bad! I have modified it.

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: I forgot to remove them. Now they were removed.

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done!

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1308,6 +1416,7 @@ def __init__(
         head_node: HeadNode,
         scheduling=None,
         shared_storage: List[Resource] = None,
+        login_nodes: LoginNodes = None,
```
EddyMM: LoginNodes should be a section associated with the `SlurmClusterConfig` instead of `BaseClusterConfig` for now (to avoid it being included in the `AWSBatchConfig`).

cli/src/pcluster/config/cluster_config.py
```diff
@@ -678,6 +684,12 @@ def availability_zone(self):
         """Compute availability zone from subnet id."""
         return AWSApi.instance().ec2.get_subnet_avail_zone(self.subnet_id)
 
+    @staticmethod
+    def getInstance():
```
EddyMM: Is this logic included in order to get the head node's subnet ID from the `LoginNodeNetworking` resource?
If so, I would recommended registering the validator at a higher layer in the configuration heirarchy (In this case, the `SlurmClusterConfig` resource). That way you can access both subnets using the

EddyMM (changes requested):

cli/src/pcluster/config/cluster_config.py
hehe7318: Done! It's really a good idea!

cli/src/pcluster/config/cluster_config.py
hehe7318: Got blockers. Can we sync on this?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -668,6 +669,7 @@ def __init__(self, subnet_id: str, elastic_ip: Union[str, bool] = None, proxy: P
         self.subnet_id = Resource.init_param(subnet_id)
         self.elastic_ip = Resource.init_param(elastic_ip)
         self.proxy = proxy
+        HeadNodeNetworking._instance = self
```
EddyMM: I don't think this is being used.

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -15,6 +15,7 @@
 from abc import ABC, abstractmethod
 from datetime import datetime
 
+import pyperclip
```
EddyMM: This won't be needed in the test.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -971,6 +970,27 @@ def _register_validators(self, context: ValidatorContext = None):  # noqa: D102
...
+        self,
+        s3_access: List[S3Access] = None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.s3_access = s3_access
+
+
+class LoginNodeIam(_BaseIam):
+    """Represent the IAM configuration for HeadNode and Queue."""
```
EddyMM: `... IAM configuration for login node.`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1183,6 +1203,79 @@ def __init__(
...
+    def __init__(self, custom_ami: str, **kwargs):
+        super().__init__()
+        self.custom_ami = Resource.init_param(custom_ami)
+
+    def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
+        if self.custom_ami:
+            self._register_validator(CustomAmiTagValidator, custom_ami=self.custom_ami)
+
+
+class LoginNodeSsh(Ssh):
```
EddyMM: Let's handle this like the IAM scenario (i.e. Inheritance) since the `Ssh` section for `LoginNodes` doesn't have an Allowed IPs parameter for now.

The Allowed IPs parameter might be an implied value if we keep it as it is now but the schema should try to match the resource to ensure it's not unintentionally used downstream.

cli/src/pcluster/config/cluster_config.py
EddyMM: Add/Extend a unit test to confirm that the instance profile and instance role cannot be used together.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -971,6 +970,27 @@ def _register_validators(self, context: ValidatorContext = None):  # noqa: D102
             self._register_validator(InstanceProfileValidator, instance_profile_arn=self.instance_profile)
 
 
+class HeadNodeAndQueueIam(_BaseIam):
```
EddyMM: We can simply call it `Iam` instead of `HeadNodeAndQueueIam` to avoid confusing the name with what it does. Essentially, it's used in either the HeadNode or the Queue but the term `HeadNodeAndQueueIam` might give the impression that the IAM values apply to both the head node and compute nodes at the same time (which is not the case).

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -928,6 +931,14 @@ def no_coexist_role_policies(self, data, **kwargs):
                 "InstanceProfile, InstanceRole or AdditionalIamPolicies can not be configured together."
             )
 
+
+class HeadNodeAndQueueIamSchema(BaseIamSchema):
```
EddyMM: Same suggestion, use `IamSchema`.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1257,6 +1282,64 @@ def make_resource(self, data, **kwargs):
...
+        """Generate resource."""
+        return LoginNodePool(**data)
+
+
+class LoginNodesSchema(BaseSchema):
+    pools = fields.Nested(
+        LoginNodePoolSchema,
+        many=True,
+        required=True,
+        validate=validate.Length(min=1, max=1, error="has a minimum size of 1, "
```
EddyMM: The error message will probably be revised later but for now, let's go with a tentative message: "Only one pool can be specified when using login nodes."

EddyMM: Also, you can use equal instead of setting a min and max of 1 ([Length](https://marshmallow.readthedocs.io/en/stable/marshmallow.validate.html#marshmallow.validate.Length))

cli/src/pcluster/validators/networking_validators.py
```diff
@@ -198,3 +198,14 @@ def _validate_all_subnets_exist(self, existing_subnets, expected_subnet_ids):
...
             )
+
+
+class AvailabilityZoneValidator(Validator):
+
+    def _validate(self, login_node_subnet_id, head_node_subnet_id):
+        if AWSApi.instance().ec2.get_subnet_avail_zone(login_node_subnet_id) != \
+                AWSApi.instance().ec2.get_subnet_avail_zone(head_node_subnet_id):
+            self._add_failure(
+                "LoginNode Networking SubnetId must be in the same availability zone as the HeadNode.",
```
EddyMM: Messaging will need rewording but let's go for a tentative message: "Login nodes and head node must be in the same availability zone".

cli/tests/pcluster/config/test_cluster_config.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+import getpass
```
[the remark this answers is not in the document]
EddyMM: ^^^

EddyMM (changes requested):

cli/src/pcluster/validators/networking_validators.py
hehe7318: Done!

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done!

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done!

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py
hehe7318: I forgot to remove it. Done!

cli/tests/pcluster/config/test_cluster_config.py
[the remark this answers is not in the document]
hehe7318: Done!

cli/tests/pcluster/templates/test_cluster_stack.py
hehe7318: Forgot to remove it. Done

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py

EddyMM (approved):

PR #5371, opened 2023-06-05
[integ-tests] Fix test_build_image intermittent failures

hanwen-pcluste:
The failing point is where the code checks the output of `pcluster get_log_events`. It asserts the output same/different from the command with different input flags.

The use of Regex is both unnecessary and wrong. It is unnecessary because there are no Regex in logs. It is wrong because it mistakes some symbols (e.g. `()`) as Regex operator, cause failures like:
```
AssertionError: Expected <Launched instance (instance id: i-0fc16a5f4611601a5)> to match pattern <Launched instance (instance id: i-0fc16a5f4611601a5)>, but did not.
```

This commit fixes bug from https://github.com/aws/aws-parallelcluster/pull/3677

### Checklist

### DIFF
--- a/tests/integration-tests/tests/createami/test_createami.py
+++ b/tests/integration-tests/tests/createami/test_createami.py
@@ -289,10 +289,10 @@ def _test_get_image_log_events(image):
             assert_that(events).is_length(expect_count)
 
         if expect_first is True:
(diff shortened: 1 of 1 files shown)


chenwany (approved):
lgtm!

PR #5381, opened 2023-06-09
[develop] Extend test for fast failover to validate multiple instance type behavior

NSsirena:
### Description of changes
Fast Failover test wasn't taking into account the different behavior derived from create-fleet, used in case of Multiple Instance Types.
This PR 
* Remove mocked responses and make direct calls to EC2 APIs. Expected behaviors are triggered by properly overriding the configuration at runtime
* Add support for slurm sbatch `prefer` option (useful to target a CR while allowing requeueing)
* Add create_fleet override for fast-failover and multiple instance types
* Add multiple-instance-type CR for fast-failover test
* Extends test_fast_failover to validate create_fleet behavior 

This MUST be merged only after merging the PR with the changes in the node package.

### Tests
* End to End test launched successfully in my dev account.
  * This depends on the changes made in the PR linked below and for testing I used a custom AMI with those changes.

### References
* [Code changes in node package](https://github.com/aws/aws-parallelcluster-node/pull/528)

### Checklist

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -345,34 +345,68 @@ def test_fast_capacity_failover(
     remote_command_executor = RemoteCommandExecutor(cluster)
     clustermgtd_conf_path = _retrieve_clustermgtd_conf_path(remote_command_executor)
     scheduler_commands = scheduler_commands_factory(remote_command_executor)
-    # after the cluster is launched, apply the override patch to launch ice nodes
+
+    # after the cluster is launched, apply the override patch to launch ice for nodes in
+    # "ice-compute-resource" and "ice-cr-multiple" compute resources
     remote_command_executor.run_remote_script(str(test_datadir / "overrides.sh"), run_as_root=True)
+
+    # All nodes
     nodes_in_scheduler = scheduler_commands.get_compute_nodes("queue1", all_nodes=True)
     static_nodes, dynamic_nodes = get_partition_nodes(nodes_in_scheduler)
-    ice_dynamic_nodes = [node for node in dynamic_nodes if "ice-compute-resource" in node]
-    static_nodes_in_ice_compute_resource = [node for node in static_nodes if "ice-compute-resource" in node]
-    # test enable fast instance capacity failover
+
+    # Nodes in Single Instance Type CR
+    ice_single_dynamic_nodes = [node for node in dynamic_nodes if "ice-compute-resource" in node]
+    ice_single_static_nodes = [node for node in static_nodes if "ice-compute-resource" in node]
+
+    # Nodes in Multiple Instance Type CR
+    ice_multi_dynamic_nodes = [node for node in dynamic_nodes if "ice-cr-multiple" in node]
+    ice_multi_static_nodes = [node for node in static_nodes if "ice-cr-multiple" in node]
+
+    # Checks Fast Failover in case of Single Instance Type - using RunInstances API
     _test_enable_fast_capacity_failover(
         scheduler_commands,
         remote_command_executor,
         clustermgtd_conf_path,
-        static_nodes_in_ice_compute_resource,
-        ice_dynamic_nodes,
+        ice_single_static_nodes,
+        ice_single_dynamic_nodes,
+        target_compute_resource="ice-compute-resource",
     )
+
+    # Check observability logic
     structured_log_event_utils.assert_that_event_exists(
         cluster, r".+\.slurm_resume_events", "node-launch-failure-count"
     )
     test_cluster_health_metric(["InsufficientCapacityErrors"], cluster.cfn_name, region)
-    # remove logs from slurm_resume log and clustermgtd log in order to check logs after disable fast capacity fail-over
-    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/slurm_resume.log")
-    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/clustermgtd")
-    # test disable ice logic
+
+    # test behavior when Fast Failover is disabled
     _test_disable_fast_capacity_failover(
         scheduler_commands,
         remote_command_executor,
         clustermgtd_conf_path,
-        static_nodes_in_ice_compute_resource,
-        ice_dynamic_nodes,
+        ice_single_static_nodes,
+        ice_single_dynamic_nodes,
+        target_compute_resource="ice-compute-resource",
+    )
+
+    # Checks Fast Failover in case of Multiple Instance Types - using CreateFleet API
+    # test enable fast instance capacity failover for "ice-cr-multiple" cr
+    _test_enable_fast_capacity_failover(
+        scheduler_commands,
+        remote_command_executor,
+        clustermgtd_conf_path,
+        ice_multi_static_nodes,
+        ice_multi_dynamic_nodes,
+        target_compute_resource="ice-cr-multiple",
+    )
+
+    # test disable ffo logic
+    _test_disable_fast_capacity_failover(
+        scheduler_commands,
+        remote_command_executor,
+        clustermgtd_conf_path,
+        ice_multi_static_nodes,
+        ice_multi_dynamic_nodes,
+        target_compute_resource="ice-cr-multiple",
     )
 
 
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh
```diff
@@ -32,4 +32,24 @@ def run_instances(region, boto3_config, **run_instances_kwargs):
     else:
         ec2_client = boto3.client("ec2", region_name=region, config=boto3_config)
         return ec2_client.run_instances(**run_instances_kwargs)
+
+
+def create_fleet(region, boto3_config, **create_fleet_kwargs):
+    lt_spec = create_fleet_kwargs.get("LaunchTemplateConfigs", [])[0].get("LaunchTemplateSpecification")
+    if "ice-cr-multiple" in lt_spec.get("LaunchTemplateName"):
```
lukeseawalker: handle the case LaunchTemplateSpecification is None, e.g.

```
lt_name = create_fleet_kwargs.get("LaunchTemplateConfigs", [])[0].get("LaunchTemplateSpecification", {}).get("LaunchTemplateName")
```

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1938,12 +1978,18 @@ def _test_enable_fast_capacity_failover(
...
     static_nodes_in_ice_compute_resource,
     ice_dynamic_nodes,
+    target_compute_resource,
 ):
     # set insufficient_capacity_timeout to 180 seconds to quicker reset compute resources
     _set_insufficient_capacity_timeout(remote_command_executor, 180, clustermgtd_conf_path)
-    # trigger insufficient capacity
+
+    # clear slurm_resume and clustermgtd logs in order to start from a clean state
+    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/slurm_resume.log")
```
lukeseawalker: NIT - what about creating an utility function out of this, instead of repeating the code multiple times?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -345,34 +345,68 @@ def test_fast_capacity_failover(
...
+    # test enable fast instance capacity failover for "ice-cr-multiple" cr
+    _test_enable_fast_capacity_failover(
+        scheduler_commands,
+        remote_command_executor,
+        clustermgtd_conf_path,
+        ice_multi_static_nodes,
+        ice_multi_dynamic_nodes,
+        target_compute_resource="ice-cr-multiple",
+    )
+
```
lukeseawalker, line 411: should we "Check observability logic" again here for this other case?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1938,12 +1978,18 @@ def _test_enable_fast_capacity_failover(
...
-    # trigger insufficient capacity
+
+    # clear slurm_resume and clustermgtd logs in order to start from a clean state
+    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/slurm_resume.log")
+    remote_command_executor.run_remote_command("sudo truncate -s 0 /var/log/parallelcluster/clustermgtd")
+
+    # trigger insufficient capacity: we are using `prefer` to allow requeuing the job on a different CR
     job_id = scheduler_commands.submit_command_and_assert_job_accepted(
-        submit_command_args={"command": "sleep 30", "nodes": 2}
+        submit_command_args={"command": "sleep 30", "nodes": 2, "prefer": target_compute_resource}
```
lukeseawalker, line 2023: why this require different submission parameter from the ones used when submitting the other job
i.e.
```
    job_id = scheduler_commands.submit_command_and_assert_job_accepted(
        submit_command_args={
            "command": "sleep 30",
            "nodes": 2,
            "other_options": "--no-requeue",
            "constraint": target_compute_resource,
        }
    )
````
ref https://github.com/aws/aws-parallelcluster/pull/5381/files#diff-5d2b210a3df82d89d8bed3f1d7d0bc48e5998e3eb9358c07aaa75c5df2d5d44eR1930?

Comment (NSsirena) [reply tests/integration-tests/tests/schedulers/test_slurm.py:2023]: It took me a while to figure out.

The first job is expected to succeed ..and for this to happen some the failed nodes must be requeued into another CR ..and for this we need to use `prefer` that "attempt to" follow the constraint ..but falls back to other CRs

The second jobs instead is expected to fail ( `--no-requeue` ) ... `constraint` here is needed to target the expected CR (not sure if it may work also with prefer ...but I used constraint to make the intent of the test slightly more clear.

File: tests/integration-tests/tests/schedulers/test_slurm.py (hunk 3 above)
Comment (NSsirena) [reply tests/integration-tests/tests/schedulers/test_slurm.py:411]: The check is just verifying that the observability logic is working.

Repeating the check will not add much.

File: tests/integration-tests/tests/schedulers/test_slurm.py (hunk 2 above)
Comment (NSsirena) [reply tests/integration-tests/tests/schedulers/test_slurm.py]: Makes sense.

File: tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh (hunk 1 above)
Comment (NSsirena) [reply tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh]: Done

File: tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh [hunk 5]
```diff
@@ -14,22 +14,54 @@ slurm_plugin_path=$(sudo find / -iname slurm_plugin -print0|grep -FzZ 'node_virt
 cat > $slurm_plugin_path/overrides.py << EOF
 from botocore.exceptions import ClientError
 import boto3
+import logging
 
+logger = logging.getLogger(__name__)
 
 def run_instances(region, boto3_config, **run_instances_kwargs):
     if "ice-compute-resource" in run_instances_kwargs.get("LaunchTemplate", {}).get("LaunchTemplateName"):
+        run_instances_kwargs['Placement']= {'Tenancy': 'host'}
```
lukeseawalker, line 25: please add a comment about the purpose of adding this parameter.
also consider doing `run_instances_kwargs['Placement']['Tenancy']=` so to not override the whole Placement

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -345,34 +345,80 @@ def test_fast_capacity_failover(
...
+        ice_single_static_nodes,
+        ice_single_dynamic_nodes,
+        target_compute_resource="ice-compute-resource",
+        expected_error_code="InsufficientHostCapacity",
+    )
+
+    # Checks Fast Failover in case of Multiple Instance Types - using CreateFleet API
+    # Requires 1 static node in the CR, expects the job to be partially reallocated in a different CR and succeed
+    # CreateFleet will return an empty list of instances, that should trigger FFO behavior
+    _test_enable_fast_capacity_failover(
```
lukeseawalker, line 402: I was expecting a `_test_enable_fast_capacity_failover` also for the new queue `exception-cr-multiple`, so to test fast failover also for the case when createFleet call generate an exception, why it wasn't considered?

tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh
NSsirena: Ooops, I missed the comment. I'll add it.

As for doing `run_instances_kwargs['Placement']['Tenancy']=` this is applied only on the target CR that doesn't have a `Placement`. As such the previous code will fail and I considered overkill and out-of-scope creating a more generic way to override this value since we are not going to use it.

tests/integration-tests/tests/schedulers/test_slurm.py
NSsirena: Both cases are triggering the same code path. The check that both conditions (exception and empty instance list) trigger the same code path is tested as Unit Test in the node package.

Repeating this test in the Integration test will add nothing in terms of coverage but will consume many more minutes in terms of execution time.

tests/integration-tests/tests/schedulers/test_slurm/test_fast_capacity_failover/overrides.sh
NSsirena: Added the missing comment

lukeseawalker (approved):

PR #5384, opened 2023-06-10
Setup AWS Infrastructure for login nodes using CDK

hehe7318:
### Description of changes
* Add Login Nodes Stack related functions in cluster_stack.py
    * modify _add_resource function
        * add support to initialize Login Nodes
        * add support to managed login nodes security groups
    * modify _add_security_groups function
        * add support for adding Login Nodes Security Groups
    * modify _add_inbounds_to_managed_security_groups function
        * add rules for login nodes and connection with head nodes and compute nodes
    * create _add_login_nodes_resources function
        * add dependency on the Head Node construct
    * create _add_login_nodes_security_group function
        * implement a base security group for login nodes and attach it to the instances
        * accept connections only on port 22 from everywhere 
        * split _add_security_groups and _add_inbounds_to_managed_security_groups to 4 functions, one for all, and other three for Head Node, Login Nodes and compute resources

* Create a Login Nodes Stack using CDK in a new python file login_nodes_stack.py
    * add login node iam resources
    * add launch templates for every login nodes pool
    * add a target group for all of the login nodes 
    * add auto scaling groups for every login nodes pool and bond with target group
        * desired capacity = MinCount, minimum capacity = MinCount, maximum capacity = MinCount.
    * add a network load balancer for all of the login nodes 
    * use a Pool Construct to organise LoginNodesStack for better readability

* Add related function in cluster_config.py
    * get IAM instance profile for login node
    * get IAM role for login node
    * add a property to get login nodes ami dict in SlurmClusterConfig class

* Modify cdk_builder_utils.py to add support for login nodes
    * Add get_login_nodes_security_groups_full function
        * Return full security groups to be used for the login node, default plus additional ones
    * Add LoginNodesIamResources class
        * Construct defining IAM resources for a login node
    * Modify node Union

* Add unit tests for the Login Node Stack in test_cluster_stack.py
    * unit test for ASG, NLB, TG and NLB listener
    * unit test for LoginNodes Launch Template

* Make LoginNodesPool.networking.subnet_id become a list→subnet_ids[str]
    * modify related function and yaml files to adapt to this change
    * add new functions to get availability zone list of the subnet ids list
    * add validator in schema to make sure "In MVP, only one subnet_id can be specified."

* Update the NLB scheme to reflect the status of the subnet
    * Private Subnet --> Internal Scheme, Public Subnet --> Internet-facing Scheme
        * modify ec2.py
            * add is_subnet_public function to see if a subnet is public or not
            * add describe_route_tables to add support for is_subnet_public function
        * modify test_ec2.py
            * add unit test to confirm is_subnet_public function assigns the correct subnet visibility (private/public) based on mocked route tables. 
        * modify dummy_aws_api.py
            * override is_subnet_public to make no real boto3 client is created
        * add is_subnet_public in LoginNodesPool.networking configuration
        * update login nodes stack NLB schema to call is_subnet_public function

```
LoginNodes:
  Pools: 
  - Name: test
    InstanceType: t2.micro
    Count: 1
    Image:
        CustomAmi: ami-12345678
    Networking:
      SubnetIds: 
      - subnet-12345678
    Ssh:
      KeyName: test-us-east-2
    Iam:
```

### Checklist


cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,140 @@
...
+                groups=login_nodes_pool_lt_security_groups,
+                subnet_id=login_nodes_pool.networking.subnet_id
+            )
+        ]
+        return ec2.CfnLaunchTemplate(
+            self,
+            f"LoginNodeLaunchTemplate{login_nodes_pool.name}",
+            launch_template_name=f"{self.stack_name}-{login_nodes_pool.name}",
+            launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
+                image_id=login_nodes_pool.image.custom_ami,
```
EddyMM: Let's fallback to th official ParallelCluster AMI used in the head node if their is no custom AMI.
I recommend adding a property to the login node config class to retrun the custom ami if it exists or `self.official_ami`

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,140 @@
...
+                # user_data=
+                # block_device_mappings=
+            )
+        )
+
+    def _add_login_nodes_pool_auto_scaling_group(
+        self,
+        login_nodes_pool: LoginNodesPools,
+    ):
+        auto_scaling_group = autoscaling.AutoScalingGroup(
```
EddyMM: I recommend using an L1 Construct ([CfnAutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_autoscaling/CfnAutoScalingGroup.html)) since the ASG will use a LaunchTemplate to describe the login node instances.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done, I added a property to get login nodes ami dict in SlurmClusterConfig class.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,217 @@
...
+        return elbv2.NetworkTargetGroup(
+            self,
+            f"{self.stack_name}-TargetGroup",
+            health_check=elbv2.HealthCheck(
+                port="22",
+                protocol=elbv2.Protocol.TCP,
+            ),
+            port=22,
+            protocol=elbv2.Protocol.TCP,
+            target_group_name=f"{self.stack_name}-TargetGroup",
```
EddyMM: What if the stack name has several characters? Does the TargetGroup name have a limit on the number of characters?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -156,7 +157,7 @@ def get_custom_tags(config: Union[BaseClusterConfig, SlurmQueue, SlurmComputeRes
 def get_default_instance_tags(
     stack_name: str,
     config: BaseClusterConfig,
-    node: Union[HeadNode, BaseComputeResource],
+    node: Union[HeadNode, LoginNodesPools, BaseComputeResource],
```
EddyMM: Given how this is used, the class should be `LoginNodesPool` instead of `LoginNodesPools`.

[LoginNodesPools](https://github.com/hehe7318/aws-parallelcluster/blob/3791241b0178e77017d9fd30ed1b4bef920c5409/cli/src/pcluster/config/cluster_config.py#L1245)

> """Represent the configuration of a LoginNodePool."""

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -268,6 +272,9 @@ def _add_resources(self):
         if self._condition_is_slurm() and hasattr(self.scheduler_resources, "cleanup_route53_custom_resource"):
             self.head_node_instance.add_depends_on(self.scheduler_resources.cleanup_route53_custom_resource)
 
+        # Initialize Login Nodes
+        self._add_login_nodes_resources()
```
EddyMM, line 261: Login nodes will depend on the Head node completing the bootstrapping and configuration process.
I think the `LoginNodesStack` should have a dependency on the Head Node construct.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,217 @@
...
+        self._add_resources()
+
+    @property
+    def stack_name(self):
+        """Name of the CFN stack."""
+        return Stack.of(self.nested_stack_parent).stack_name
+
+    def _add_resources(self):
+        # get unique login nodes pools subnet id
+        subnet_ids = set()
```
EddyMM: Consider using set comprehension ([Sets](https://docs.python.org/3/tutorial/datastructures.html#sets)).
```
{pool.networking.subnet_id for pool in self._login_nodes.pools}
```

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,217 @@
...
+        for login_nodes_pool in self._login_nodes.pools:
+            subnet_ids.add(login_nodes_pool.networking.subnet_id)
+        self._pools_subnet_ids = list(subnet_ids)
+
+        # get unique login nodes pools availability_zone
+        availability_zones = set()
```
EddyMM: set comprehension ([Sets](https://docs.python.org/3/tutorial/datastructures.html#sets)).

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,217 @@
...
+    LoginNodesIamResources,
+    get_default_instance_tags,
+    get_default_volume_tags,
+    get_login_nodes_security_groups_full,
+)
+from pcluster.utils import get_http_tokens_setting
+
+
+class LoginNodesStack(NestedStack):
+    """Stack encapsulating a set of LoginNodes and the associated resources."""
```
EddyMM: General comment on the approach used here to manage the pool info, can we use a [Construct](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html) to organise this code for better readability?
```
class Pool(Construct):
    ...
    <pool related resources>
```

Then keep the LoginNodesStack responsible for creating a Pool construct for each pool.

```
class LoginNodesStack(NestedStack):
    ...
    for pool in pools:
        Pool(...)
    ...
```

Ref: [Writing your own constructs](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_author)

EddyMM (changes requested):

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Yes, the TargetGroup name have a limit on the number of characters, the limit is **maximum of 32 characters**.
:param target_group_name: The name of the target group. This name must be unique per region per account, can have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, and must not begin or end with a hyphen. Default: - Automatically generated.
The target group's name is f"{cluster.name}-TargetGroup", so if we don't change this format, we should add a validator to make the cluster.name has a maximum of 20 characters.
Or we should change the target_group_name format.
Which option is better? I prefer the second one.

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: That's a good reminding, I have changed all the LoginNodesPools to LoginNodesPool. Thank you Eddy!

cli/src/pcluster/templates/cluster_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Okay, I will try.

cli/src/pcluster/templates/cluster_stack.py
hehe7318: Done. I added a dependency in _add_login_nodes_resources() function.
```
# Add dependency on the Head Node construct
self.login_nodes_stack.node.add_dependency(self.head_node_instance)
```

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
EddyMM: If the target group name is not being used (in a static way) by any other resource, then I think we can let CDK autogenerate the name.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,229 @@
...
+    CdkLaunchTemplateBuilder,
+    LoginNodesIamResources,
+    get_default_instance_tags,
+    get_default_volume_tags,
+    get_login_nodes_security_groups_full,
+)
+from pcluster.utils import get_http_tokens_setting
+
+
+class Pool(Construct):
```
EddyMM, line 19: The target group and load balancer are associated with a pool.
Should we create them inside the pool instead of creating them from the login nodes stack?

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,229 @@
...
+            internet_facing=True,
+            vpc_subnets=ec2.SubnetSelection(
+                subnets=[
+                    ec2.Subnet.from_subnet_id(self, f"LoginNodesSubnet{i}", subnet_id)
+                    for i, subnet_id in enumerate(self._pools_subnet_ids)
+                ]
+            ),
+        )
+
+        listener = login_nodes_load_balancer.add_listener("LoginNodeListener", port=22)
```
EddyMM: If each pool will have a load balancer, should we use a name that won't conflict with other listeners?

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,229 @@
...
+        listener.add_target_groups("LoginNodeListenerTargets", target_group)
```
EddyMM: Same question, this listener is associated with a single pool. The `id` should avoid conflicts.

EddyMM (changes requested):
Let's also add unit tests for the Login Node Stack to verify that the proper resources are described for the corresponding cluster configuration file.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done! I format the id with pool.name.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done! I format the id with pool.name.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -877,6 +907,86 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
...
+                resources=["*"],
+            ),
+            iam.PolicyStatement(
+                sid="CloudWatchLogs",
+                actions=[
+                    "logs:GetLogEvents",
+                    "logs:FilterLogEvents",
+                ],
+                effect=iam.Effect.ALLOW,
+                resources=["*"],
```
EddyMM: I don't think the login nodes need all these policies for now.

hehe7318: okay, I will delete _build_policy function for now.

hehe7318: I delete most of them, but keep EC2 Policy, cause _build_policy is a abstract method, must have. And if I return an empty list, two unit test will fail. So for the convenience of adding later, I keep one for now.
```
>           raise RuntimeError(resp.error) from JavaScriptError(resp.stack)
E           RuntimeError: Error: Resolution error: Supplied properties not correct for "CfnPolicyProps"
E             policyDocument: required but missing.
```

cli/src/pcluster/aws/ec2.py
```diff
@@ -487,3 +487,26 @@ def run_instances(self, **kwargs):
...
                 raise
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_route_tables(self, filters=None):
+        """Describe EC2 route tables."""
+        kwargs = {"Filters": filters} if filters else {}
+        return self._paginate_results(self._client.describe_route_tables, **kwargs)
+
+    @AWSExceptionHandler.handle_client_exception
+    def is_subnet_public(self, subnet_id):
```
EddyMM, line 539: Let's add a unit test for this function to confirm it assigns the correct subnet visibility (private/public) based on mocked route tables.

Also add a sample/snippet output of the `describe_route_tables` call in the docstring of this function for better documentation.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1244,15 +1244,40 @@ class LoginNodesNetworking(_BaseNetworking):
...
+
+    @property
+    def az_list(self):
+        """Get LoginNodesPool availability zones list."""
+        return list(self.az_subnet_ids_mapping.keys())
+
+    @property
+    def is_subnet_public(self):
+        """Get if the subnet is public or private."""
+        return AWSApi.instance().ec2.is_subnet_public(self.subnet_ids[0])
```
EddyMM: These methods are the same as the ones in the queue networking.
Is it possible to move them into an abstract class that can be used by both Networking classes?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -3270,13 +3319,30 @@ def has_gpu_health_checks_enabled(self):
                     return True
         return False
 
+    @property
+    def login_nodes_subnet_ids(self):
+        """Return the list of all compute subnet ids in the cluster."""
```
EddyMM: "compute subnet ids"?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -3270,13 +3319,30 @@ def has_gpu_health_checks_enabled(self):
...
+        subnet_ids_list = []
+        for pool in self.login_nodes.pools:
```
EddyMM, line 2837: Can we use Sets to perform this operation?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1308,36 +1308,40 @@ def make_resource(self, data, **kwargs):
 class LoginNodesNetworkingSchema(BaseNetworkingSchema):
     """Represent the networking schema of LoginNodes."""
 
-    subnet_id = fields.Str(required=True)
+    subnet_ids = fields.List(
+        fields.Str(validate=get_field_validator("subnet_id")),
+        required=True,
+        validate=validate.Length(equal=1, error="In MVP, only one subnet_id can be specified."),
```
EddyMM: How about an error message: "Only one subnet can be associated with a login node pool"?

(Let's avoid using the term "MVP" since it's more of a project planning concept)

EddyMM (changes requested):
Are there unit tests we can add related to the stack e.g. The template configuration of the AutoScalingGroup / NLB / Target Group etc. based on the cluster configuration?

cli/src/pcluster/templates/cdk_builder_utils.py
EddyMM: I suggest checking the 2 failing unit tests to understand how they can be improved.
Adding policies as placeholders is a bit risky since they can be forgotten unintentionally.

cli/src/pcluster/config/cluster_config.py
hehe7318: sorry, it should be login nodes subnet ids. I have fixed this issue.

cli/src/pcluster/config/cluster_config.py
hehe7318: Sure, I will also modify the function for compute subnet ids for better performance. Cause that function also use list and I copied that code.

cli/src/pcluster/config/cluster_config.py
hehe7318: I made a new class inherit from _BaseNetworking, contains these methods, and let the login nodes networking class and queue networking class inherit from the new class. To avoid duplicates.
```
class _NetworkingWithSubnetMapping(_BaseNetworking):
    """Represent the networking configuration with subnet to AZ mapping for queue networking and login nodes networking."""
```

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done! I will pay attention on it in the future.

cli/src/pcluster/aws/ec2.py
hehe7318: I have added the unit test and sample output. Please take a review cause I don't know if the sample output meet requirement.

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: I found that if I leave this function like this the unit test could all pass.
```
def _build_policy(self) -> List[iam.PolicyStatement]:
        return [
            iam.PolicyStatement(),
        ]
```

hehe7318: I removed all the LoginNodesIamResource related code for now.

cli/src/pcluster/aws/ec2.py
```diff
@@ -487,3 +487,65 @@ def run_instances(self, **kwargs):
...
+        """
+        kwargs = {"Filters": filters} if filters else {}
+        return list(self._paginate_results(self._client.describe_route_tables, **kwargs))
+
+    @AWSExceptionHandler.handle_client_exception
+    def is_subnet_public(self, subnet_id):
+        """Check if a subnet is public."""
+        route_tables = self.describe_route_tables(filters=[{"Name": "association.subnet-id", "Values": [subnet_id]}])
+        if not route_tables:
+            return False
```
EddyMM: This should raise an Exception since lack of route tables is not an indication that the subnet is private.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -639,6 +639,34 @@ def _register_validators(self, context: ValidatorContext = None):  # noqa: D102
         self._register_validator(SecurityGroupsValidator, security_group_ids=self.additional_security_groups)
 
 
+class _NetworkingWithSubnetMapping(_BaseNetworking):
```
EddyMM: Using a base class for a small set of additional methods might be difficult to maintain E.g. If we have other shared methods, we'd need to either change the name of this class or subclass again.

I'd suggest using `Mixins` which allow sharing of method implementations across child classes without several hierarchies of sub-classing.

EddyMM (changes requested):

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,206 @@
...
+        ]
+        return ec2.CfnLaunchTemplate(
+            self,
+            f"LoginNodeLaunchTemplate{self._pool.name}",
+            launch_template_name=f"{self.stack_name}-{self._pool.name}",
+            launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
+                image_id=self._config.login_nodes_ami[self._pool.name],
+                instance_type=self._pool.instance_type,
+                key_name=self._pool.ssh.key_name,
+                # iam_instance_profile=ec2.CfnLaunchTemplate.IamInstanceProfileProperty(name=self._instance_profile),
```
EddyMM: Let's remove fields that are not used (instead of commenting).

cli/src/pcluster/aws/ec2.py
hehe7318: That makes sense. Done!

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -639,6 +639,33 @@ def _register_validators(self, context: ValidatorContext = None):  # noqa: D102
         self._register_validator(SecurityGroupsValidator, security_group_ids=self.additional_security_groups)
 
 
+class _SubnetsAZMappingMixin:
```
EddyMM: `SubnetsMixin`

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2796,13 +2828,29 @@ def has_gpu_health_checks_enabled(self):
...
+
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
         self._register_validator(
             MixedSecurityGroupOverwriteValidator,
             head_node_security_groups=self.head_node.networking.security_groups,
             queues=self.scheduling.queues,
         )
+        if self.login_nodes:
+            self._register_validator(
```
EddyMM: This Validator is registered in the `BaseClusterConfig` class. This might result in duplicate validator registration. We can remove this for now and apply the VPC check when adding validators.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -0,0 +1,205 @@
...
+                # user_data=
+                # block_device_mappings=
+                tag_specifications=[
+                    ec2.CfnLaunchTemplate.TagSpecificationProperty(
+                        resource_type="instance",
+                        tags=get_default_instance_tags(
+                            self.stack_name, self._config, self._pool, "LoginNode", self._shared_storage_infos
+                        )
+                        + [CfnTag(key=PCLUSTER_LOGIN_NODES_POOL_NAME_TAG, value=self._pool.name)]
+                        # + custom tags for instance
```
EddyMM: Let's remove it instead of a comment.

EddyMM (changes requested):

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done. BTW I also add new unit test for Security Groups.

EddyMM (approved):

PR #5407, opened 2023-06-16
Remove scheduler-plugin code

enrico-usai:
Searched for: (case insensitive)
* scheduler-plugin
* schedulerplugin
* scheduler: plugin
* scheduler_plugin
* scheduler_queues
* scheduler_settings
* plugin

### References
* Related to https://github.com/aws/aws-parallelcluster-cookbook/pull/2299


cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -353,7 +353,6 @@ def update_cluster(
                 force=force_update,
                 validator_suppressors=get_validator_suppressors(suppress_validators),
                 validation_failure_level=FailureLevel[validation_failure_level],
-                dry_run=dryrun,
```
NSsirena, line 356: Why are we removing this ?
is `True` the default value ?

NSsirena: The intent of this code is not clear.

We enter in this block if and only if `dryrun=True` but then we do not use the dryrun value while validating the request.

Also if we want to set the default value for `dryrun` to `False` why we set the default to `none` and then initialize the variable with `dryrun = dryrun is True` ?

cli/tests/pcluster/api/controllers/test_cluster_compute_fleet_controller.py
```diff
@@ -40,27 +39,6 @@ def _build_dynamodb_item(scheduler, status, last_status_updated_time=None):
...
-                }
-            }
-            if last_status_updated_time:
-                last_status_updated_time = str(last_status_updated_time)
-            else:
-                last_status_updated_time = str(datetime.now())
-
-            dynamodb_item["Item"][compute_fleet_status_manager.DB_DATA][
-                compute_fleet_status_manager.COMPUTE_FLEET_LAST_UPDATED_TIME_ATTRIBUTE
-            ] = last_status_updated_time
```
NSsirena: `get_manager(` still returns a `JsonComputeFleetStatusManager `
 so either we keep the test and adapt it to run without the SchedulerPluign or we change the `get_manager` and remove the option to get a  `JsonComputeFleetStatusManager ` if it is not used anymore.

cli/src/pcluster/api/controllers/cluster_operations_controller.py
enrico-usai: We are removing this parameter because there is no code using the dryrun variable anymore into the `validate_create_request` and `validate_update_request` methods.

The dryrun variable was used only for a "scheduler-plugin" specific branch of the code that has been removed as part of this patch.

See `validate_create_request` and `validate_update_request` changes in this patch.

We can improve the code if we want in future to execute a more precise dryrun in the validate methods but as part of this patch I'm just removing scheduler-plugin specific code, I don't want to change internal logic.

cli/tests/pcluster/api/controllers/test_cluster_compute_fleet_controller.py
enrico-usai: You're right. Code updated to keep testing that branch

NSsirena (approved):

PR #5408, opened 2023-06-16
Propagate cluster s3 bucket information to the compute nodes

enrico-usai:
This is required to fetch config from S3 if for some reason the config file is no longer available in the shared storage.

As part of cookbook refactoring I wrongly moved the fetch_config.rb before the mount of the shared storage. Since the config was not available from the shared storage, the compute node was trying to download it from s3 but the cluster/cluster_s3_bucket node attribute was unavailable.

I'm going to restore the order to avoid fetching from s3 every time but in any case it's useful to add robustness to the compute nodes fetch_config logic.


enrico-usai:
This patch is incomplete, s3 key is missing. 

This patch has been totally refactored as part of: https://github.com/aws/aws-parallelcluster-cookbook/pull/2345

PR #5412, opened 2023-06-19
[develop] Always call parameters by name while submitting a job for testing

NSsirena:
### Description of changes
* when submitting a job during a test always calls parameters by name to ensure the they are properly managed and formatted in a valid job submission command 

### Tests
* Tests are failing in develop after merging the fix for Fast Failover that added a parameter and broke the sequence of parameters

### References
* [Fix FFO](https://github.com/aws/aws-parallelcluster/commit/43bb9307c9d0eae1dffdd08df99c7f384f4c1596)

### Checklist

### DIFF
--- a/tests/integration-tests/tests/common/schedulers_common.py
+++ b/tests/integration-tests/tests/common/schedulers_common.py
@@ -280,14 +280,14 @@ class SlurmCommands(SchedulerCommands):
 
         return self._submit_batch_job(
             job_submit_command,
-            nodes,
-            slots,
-            host,
-            after_ok,
-            partition,
-            constraint,
-            prefer,
-            other_options,
+            nodes=nodes,
+            slots=slots,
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/common/schedulers_common.py
```diff
@@ -280,14 +280,14 @@ def submit_command(
...
             job_submit_command,
-            nodes,
-            slots,
-            host,
-            after_ok,
-            partition,
-            constraint,
-            prefer,
-            other_options,
+            nodes=nodes,
```
enrico-usai, line 283: let's use the same approach in the line 317 too

NSsirena: Done

enrico-usai (approved):

PR #5418, opened 2023-06-20
Add Test option for using run-instance API rather than create-fleet API for compute Fleet

himani2411:
### Description of changes
* Making changes for using instances, region and OS dimensions for TEST_CONFIG Yaml file.
* Added `--force-run-instance-fleet-manager` testing option for using run-instance API rather than create-fleet API for compute Fleet

### Tests
* Tested the changes Locally

### Checklist

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -179,6 +179,11 @@ def pytest_addoption(parser):
         help="Name of CFN stack providing database stack to be used for testing Slurm accounting feature.",
     )
     parser.addoption("--external-shared-storage-stack-name", help="Name of existing external shared storage stack.")
+    parser.addoption(
+        "--force-run-instance-fleet-manager",
+        help="Force usage of Run-instance-api instead of creat-fleet api for compute fleet",
+        action="store_true",
+    )
 
 
 def pytest_generate_tests(metafunc):
@@ -712,6 +717,20 @@ def inject_additional_config_settings(cluster_config, request, region, benchmark
                 if not compute_resource.get("MaxCount"):
                     # Use larger max count to support performance tests if not specified explicitly.
                     compute_resource["MaxCount"] = 150
+
+    # Forcing the usage of Run-instance API for creation of fleet when we use InstanceType instead of Instances/InstanceType
+    if (
+        scheduler != "awsbatch"
+        and dict_has_nested_key(config_content, ("Scheduling", "SlurmQueues"))
+        and request.config.getoption("force_run_instance_fleet_manager")
+    ):
+        for queues in config_content["Scheduling"]["SlurmQueues"]:
+            if dict_has_nested_key(queues, ["ComputeResources"]):
+                for compute_resources in queues["ComputeResources"]:
+                    if dict_has_nested_key(compute_resources, ["Instances"]):
+                        instance_type = compute_resources["Instances"][0]["InstanceType"]
+                    compute_resources.pop("Instances")
+                    compute_resources["InstanceType"] = instance_type
 
     with open(cluster_config, "w", encoding="utf-8") as conf_file:
         yaml.dump(config_content, conf_file)
--- a/tests/integration-tests/test_runner.py
+++ b/tests/integration-tests/test_runner.py
@@ -89,6 +89,7 @@ TEST_DEFAULTS = {
     "cluster_custom_resource_service_token": None,
     "resource_bucket": None,
     "lambda_layer_source": None,
+    "force_run_instance_fleet_manager": False,
 }
 
 
@@ -402,6 +403,13 @@ def _init_argparser():
         default=TEST_DEFAULTS.get("external_shared_storage_stack_name"),
     )
 
+    debug_group.add_argument(
+        "--force-run-instance-fleet-manager",
+        help="Forces the usage of EC2 run-instances API for creation of Fleet",
+        default=TEST_DEFAULTS.get("force_run_instance_fleet_manager"),
+        action="store_true",
+    )
+
     return parser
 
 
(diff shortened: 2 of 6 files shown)


himani2411:
Need to remove the [2nd commit](https://github.com/aws/aws-parallelcluster/pull/5418/commits/2b8ad6630859ca6034cae4b24156a6616ba81424) where I have hardcoded the instance type with ARM architecture.

tests/integration-tests/conftest.py
```diff
@@ -179,6 +179,11 @@ def pytest_addoption(parser):
         help="Name of CFN stack providing database stack to be used for testing Slurm accounting feature.",
     )
     parser.addoption("--external-shared-storage-stack-name", help="Name of existing external shared storage stack.")
+    parser.addoption(
+        "--force-run-instance-fleet-manager",
```
enrico-usai, line 183: I'd use `--force-run-instances`

tests/integration-tests/conftest.py
```diff
@@ -179,6 +179,11 @@ def pytest_addoption(parser):
...
+        help="Force usage of Run-instance-api instead of creat-fleet api for compute fleet",
```
enrico-usai, line 184: Force usage of run-instances boto3 API instead of create-fleet for compute fleet scaling up. Note: if there are multiple instances in the list, only the first will be used.

tests/integration-tests/test_runner.py
```diff
@@ -402,6 +403,13 @@ def _init_argparser():
         default=TEST_DEFAULTS.get("external_shared_storage_stack_name"),
     )
 
+    debug_group.add_argument(
+        "--force-run-instance-fleet-manager",
+        help="Forces the usage of EC2 run-instances API for creation of Fleet",
+        default=TEST_DEFAULTS.get("force_run_instance_fleet_manager"),
```
enrico-usai, line 409: the boto3 api is "run-instances" not run-instance, this is why I suggest to rename it.

tests/integration-tests/test_runner.py
```diff
@@ -432,13 +440,16 @@ def _is_url(value):
         raise argparse.ArgumentTypeError("'{0}' is not a valid url".format(value))
 
 
-def _test_config_file(value):
-    _is_file(value)
+def _test_config_file(test_config_file, test_config_args=None):
```
enrico-usai, line 443: nit: 
```
_test_config_file(file_path, config_args=None):
```

tests/integration-tests/test_runner.py
```diff
@@ -432,13 +440,16 @@ def _is_url(value):
...
+    _is_file(test_config_file)
     try:
-        config = read_config_file(value)
+        if test_config_args:
+            config = read_config_file(test_config_file, **test_config_args)
+        else:
```
enrico-usai, line 448: this if/else shouldn't be required since read_config_file is able to manage the args even if it's None.

tests/integration-tests/test_runner.py
```diff
@@ -688,6 +702,17 @@ def _run_parallel(args):
         job.join()
 
 
+def get_config_arguments(args):
```
enrico-usai, line 705: the internal convention is to use the `_` prefix for private functions, that are only used in this python module.

tests/integration-tests/test_runner.py
```diff
@@ -705,7 +730,11 @@ def _check_args(args):
         assert_that(args.schedulers).described_as("--schedulers cannot be empty").is_not_empty()
     else:
         try:
-            args.tests_config = _test_config_file(args.tests_config)
+            test_config_args = get_config_arguments(args)
+            if test_config_args:
+                args.tests_config = _test_config_file(args.tests_config, test_config_args)
```
enrico-usai, line 735: same here, you can always pass ahead the test_config_args even if it's None, because the method is able to understand when to use it.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -67,7 +67,9 @@ def test_slurm(
     Grouped all tests in a single function so that cluster can be reused for all of them.
     """
     scaledown_idletime = 3
-    gpu_instance_type = "g3.4xlarge"
+    # gpu_instance_type = "g3.4xlarge"
+    # TODO Revert to x86_64 gpu_instance_type
+    gpu_instance_type = "g5g.2xlarge"
```
enrico-usai: you can create a check to use the right instance type according to the architecture of the main instance type configured in the test.
```python
gpu_instance_type = "g3.4xlarge" if instance_architecture_is_arm else "g5g.2xlarge"
```

In this way you don't need to modify the test by hand.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -96,14 +98,15 @@ def test_slurm(
...
+        instance_type="c7g.medium",  # TODO revert to c5.xlarge
         cpu_per_instance=4,
     )
     _gpu_resource_check(
         slurm_commands, partition="gpu", instance_type=gpu_instance_type, instance_type_info=gpu_instance_type_info
     )
+    # TODO revert to c5.xlarge
     _test_cluster_limits(
-        slurm_commands, partition="ondemand", instance_type="c5.xlarge", max_count=5, cpu_per_instance=4
+        slurm_commands, partition="ondemand", instance_type="c7g.medium", max_count=5, cpu_per_instance=4
```
enrico-usai: same here, you can use a if/else check for every instance type in the test.

himani2411: As discussed, these are not testing the actual instance that we want. So have removed the commit

tests/integration-tests/tests/schedulers/test_slurm.py
himani2411: As discussed, these are not testing the actual instance that we want. So have removed the commit

tests/integration-tests/test_runner.py
himani2411: Will make changes.

tests/integration-tests/test_runner.py
himani2411: I would get an error like 
`** must be a mapping, not NoneType` for not putting the condition

tests/integration-tests/conftest.py
himani2411: Will add this

tests/integration-tests/test_runner.py
himani2411: Will make this change.

tests/integration-tests/conftest.py
himani2411: Done

enrico-usai:
Closed in favour of https://github.com/aws/aws-parallelcluster/pull/5427

PR #5429, opened 2023-06-23
Improve new_os.yaml template to always use the NEW_OS placeholder

enrico-usai:
This configuration file must be used to test a new OS, so it's useless to execute tests for other OSes.

Removed optin regions to be able to run them in a single AWS account (a part from china and govcloud)


dreambeyondorange (approved):
~~Just curious, why the changes in region?~~

See the optin comment now

PR #5435, opened 2023-06-27
Add support for LoginNodes in describe_cluster_instances

maurizp:
### Description of changes
* Add functionalities to support LoginNodes in the describe_cluster_instances call.
* In order to achieve this result I update the smithy models to support the new type.
* Add more policies in the ParallelClusterClusterPolicy since it is needed to execute the API calls.

### Tests
* Update existing unit tests
* Add a dedicated integration test
* Manually tested the CLI

### Checklist

### References
- Login Nodes infrastructure: https://github.com/aws/aws-parallelcluster/commit/c732ab5fd0576f00a0b92954571cc1bc95af13c7

### DIFF
--- a/cloudformation/policies/parallelcluster-policies.yaml
+++ b/cloudformation/policies/parallelcluster-policies.yaml
@@ -564,6 +564,26 @@ Resources:
             Resource: '*'
             Effect: Allow
             Sid: ResourceGroupRead
+          - Action:
+              - autoscaling:CreateAutoScalingGroup
+              - autoscaling:DeleteAutoScalingGroup
+              - autoscaling:DescribeAutoScalingGroups
+              - autoscaling:DescribeScalingActivities
+              - autoscaling:UpdateAutoScalingGroup
+              - elasticloadbalancing:AddTags
+              - elasticloadbalancing:CreateListener
+              - elasticloadbalancing:CreateLoadBalancer
+              - elasticloadbalancing:CreateTargetGroup
+              - elasticloadbalancing:DeleteListener
+              - elasticloadbalancing:DeleteLoadBalancer
+              - elasticloadbalancing:DeleteTargetGroup
+              - elasticloadbalancing:DescribeListeners
+              - elasticloadbalancing:DescribeLoadBalancers
+              - elasticloadbalancing:DescribeTargetGroups
+              - elasticloadbalancing:ModifyLoadBalancerAttributes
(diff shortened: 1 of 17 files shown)


NSsirena (approved):

cloudformation/policies/parallelcluster-policies.yaml
```diff
@@ -564,6 +564,26 @@ Resources:
...
+              - elasticloadbalancing:CreateLoadBalancer
+              - elasticloadbalancing:CreateTargetGroup
+              - elasticloadbalancing:DeleteListener
+              - elasticloadbalancing:DeleteLoadBalancer
+              - elasticloadbalancing:DeleteTargetGroup
+              - elasticloadbalancing:DescribeListeners
+              - elasticloadbalancing:DescribeLoadBalancers
+              - elasticloadbalancing:DescribeTargetGroups
+              - elasticloadbalancing:ModifyLoadBalancerAttributes
+            Resource: '*'
```
enrico-usai, line 584: Is `*` really needed, or can we limit access to a specific ASG and LB with a prefix (e.g. `pcluster-*`) ?

tests/integration-tests/tests/pcluster_api/test_api.py
```diff
@@ -219,6 +219,35 @@ def test_cluster_awsbatch(
     )
 
 
+@pytest.mark.usefixtures("os", "instance")
+def test_login_nodes(
```
enrico-usai, line 223: What about adding these checks  and the Login node pool to the `test_cluster_slurm` and avoid the creation of an additional stack/test?

enrico-usai (commented):
From this patch I cannot see where you modified the code to add specific tags to the Login Nodes or how are you creating them.

If there is an additional patch can you please add it as reference?

Can you also modify the patch title? 

Thanks!

cloudformation/policies/parallelcluster-policies.yaml
maurizp: In the current implementation there are no specific identifiers for these resources. We can restrict them later.

tests/integration-tests/tests/pcluster_api/test_api.py
maurizp: The idea here was to create a dedicated set of integration tests for the Login Nodes lifecycle workflow.

PR #5440, opened 2023-06-29
Add a LifeCycle hook to the login nodes AutoScaling Group

hehe7318:
### Description of changes
* Add gracetime_period in LoginNodesPool Configuration and Schema
  * time in minutes after which login nodes will be terminated after the administrator runs the stop command
  * fix Unit test to adapt new parameter gracetime_period
* Add LifeCycle hook related functions in cluster_stack.py
  * create add lifecycle hook function
    * set heartbeat_timeout to gracetime_period
* Add unit test for gracetime_period and lifecycle hook

```
LoginNodes:
  Pools: 
  - Name: test
    InstanceType: t2.micro
    Count: 1
    Image:
        CustomAmi: ami-12345678
    Networking:
      SubnetIds: 
      - subnet-12345678
    Ssh:
      KeyName: test-us-east-2
    Iam:
    GracetimePeriod: 60
```

### Checklist


cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -71,6 +134,58 @@ def _add_login_nodes_pool_launch_template(self):
...
             )
         ]
+
+        # User data to setup and run the daemon script
+        user_data = """#!/bin/bash
+cat << 'EOF' > /opt/parallelcluster/scripts/daemon_script.sh
+#!/bin/bash
+while true; do
+  TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+  LIFECYCLE_STATE=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" \
```
EddyMM: Check if this works when [Imds configuration](https://docs.aws.amazon.com/parallelcluster/latest/ug/Imds-cluster-v3.html) is set to `v1.0`

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -71,6 +134,58 @@ def _add_login_nodes_pool_launch_template(self):
...
+# Prevent further login/SSH attempts by updating ssh config
+echo "AllowUsers $DEFAULT_USER" >> /etc/ssh/sshd_config
+
+# Reload the ssh configuration
+systemctl reload sshd
+
+# Broadcast a message to all logged in users using the wall command
+MSG="System is going down for termination in {0} minutes!"
+wall "$MSG"
+while true; do
```
EddyMM: Should we send the message to users every minute?

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -71,6 +134,58 @@ def _add_login_nodes_pool_launch_template(self):
...
+  sleep 60
+  wall "System is going down for termination, please save your work!"
+done
+EOF
+
+chmod +x /opt/parallelcluster/scripts/*.sh
```
EddyMM: This expression is a bit too general. This could be a security risk. Let's specify the exact scripts.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -71,6 +71,54 @@ def _add_login_nodes_pool_launch_template(self):
                 subnet_id=self._pool.networking.subnet_ids[0],
             )
         ]
+
```
EddyMM, line 74: Let's maintain these scripts outside of this PR (since we'll manage them in the cookbook package).
Not using user_data.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -117,8 +166,19 @@ def _add_login_nodes_pool_auto_scaling_group(self):
...
         )
 
+        self._add_lifecycle_hook(auto_scaling_group)
+
         return auto_scaling_group
 
+    def _add_lifecycle_hook(self, auto_scaling_group):
+        return autoscaling.CfnLifecycleHook(
+            self,
+            "LoginNodesASGLifecycleHook",
```
EddyMM, line 128: Do we need any additional IAM policies to create/manage a LifeCycleHook resource?

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: It can work.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: After discussing with Mauri, it has been removed.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: That makes sense. Thanks for your reminding.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Done

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: I think we need:
autoscaling:PutLifecycleHook -> create the lifecycle hook

EddyMM: Thanks, let's add it to the list of permissions needed for the API as well.

EddyMM (approved):

PR #5446, opened 2023-07-05
[develop] Add user-data for LoginNodes

NSsirena:
### Description of changes
* Add a user-data.sh template for LoginNodes
* Add HeadNode ENI info to LoginNode Pool Construct
* Configure and add UserData to LoginNode launch template

### Tests
* a cluster with LoginNodes was manually launched (using custom cookbook branch) and login nodes basic functionalities were manually tested: 
  * sinfo, sbatch, scontrol 

### References
* [Cookbook recipe changes](https://github.com/aws/aws-parallelcluster-cookbook/pull/2345)

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -1,0 +1,215 @@
+Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+MIME-Version: 1.0
+
+--==BOUNDARY==
+Content-Type: text/cloud-boothook; charset="us-ascii"
+MIME-Version: 1.0
+
+#!/bin/bash -x
+
+which dnf 2>/dev/null; dnf=$?
+which yum 2>/dev/null; yum=$?
+
+if [ "${!dnf}" == "0" ]; then
+  echo "proxy=${DnfProxy}" >> /etc/dnf/dnf.conf
+elif [ "${!yum}" == "0" ]; then
+  echo "proxy=${YumProxy}" >> /etc/yum.conf
+else
+  echo "Not yum system"
+fi
+
+which apt-get && echo "Acquire::http::Proxy \"${AptProxy}\";" >> /etc/apt/apt.conf || echo "Not apt system"
+
+proxy=${ProxyServer}
+if [ "${!proxy}" != "NONE" ]; then
+  proxy_host=$(echo "${!proxy}" | awk -F/ '{print $3}' | cut -d: -f1)
+  proxy_port=$(echo "${!proxy}" | awk -F/ '{print $3}' | cut -d: -f2)
+  echo -e "[Boto]\nproxy = ${!proxy_host}\nproxy_port = ${!proxy_port}\n" >/etc/boto.cfg
+  cat >> /etc/profile.d/proxy.sh <<PROXY
+export http_proxy="${!proxy}"
+export https_proxy="${!proxy}"
+export no_proxy="localhost,127.0.0.1,169.254.169.254"
+export HTTP_PROXY="${!proxy}"
+export HTTPS_PROXY="${!proxy}"
+export NO_PROXY="localhost,127.0.0.1,169.254.169.254"
+PROXY
+fi
+
+--==BOUNDARY==
+Content-Type: text/cloud-config; charset=us-ascii
+MIME-Version: 1.0
+
+package_update: false
+package_upgrade: false
+repo_upgrade: none
+
+datasource_list: [ Ec2, None ]
+output:
+  all: "| tee -a /var/log/cloud-init-output.log | logger -t user-data -s 2>/dev/console"
+write_files:
+  - path: /tmp/dna.json
+    permissions: '0644'
+    owner: root:root
+    content: |
+      {
+        "cluster": {
+          "base_os": "${BaseOS}",
+          "cluster_name": "${ClusterName}",
+          "cluster_user": "${OSUser}",
+          "custom_node_package": "${CustomNodePackage}",
+          "custom_awsbatchcli_package": "${CustomAwsBatchCliPackage}",
+          "cw_logging_enabled": "false",
+          "directory_service": {
+            "enabled": "${DirectoryServiceEnabled}"
+          },
+          "ebs_shared_dirs": "${EbsSharedDirs}",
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -0,0 +1,233 @@
...
+bootcmd:
+  # Disable multithreading using logic from https://aws.amazon.com/blogs/compute/disabling-intel-hyper-threading-technology-on-amazon-linux/
+  # thread_siblings_list contains a comma (,) or dash (-) separated list of CPU hardware threads within the same core as cpu
+  # e.g. 0-1 or 0,1
+  # cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list
+  #     | tr '-' ','       # convert hyphen (-) to comma (,), to account that some kernels and CPU architectures use a hyphen instead of a comma
+  #     | cut -s -d, -f2-  # split over comma (,) and take the right part
+  #     | tr ',' '\n'      # convert remaining comma (,) into new lines
+  #     | sort -un         # sort and unique
+  - if [ "${DisableMultiThreadingManually}" = "true" ]; then for cpunum in $(cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list | tr '-' ',' | cut -s -d, -f2- | tr ',' '\n' | sort -un); do echo 0 > /sys/devices/system/cpu/cpu$cpunum/online; done; fi
```
EddyMM: I see `DisableMultiThreadingManually` is explicitly set to false in the CDK UserData interpolation. Since the user data has a size limit can we avoid having this code segment until we actually need it?

cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -0,0 +1,233 @@
...
+        curl --retry 3 -v -L -o /etc/chef/aws-parallelcluster-cookbook.tgz ${!cookbook_url}
+        vendor_cookbook
+      fi
+      cd /tmp
+
+      mkdir -p /etc/chef/ohai/hints
+      touch /etc/chef/ohai/hints/ec2.json
+
+      # measure start time
+      start=$(date +%s)
```
EddyMM: The `startup_time` metrics probably don't need to be published in this case. In any case the `start` and `end` variables are not being used.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -397,7 +397,7 @@ def _add_fleet_and_scheduler_resources(self, cleanup_lambda, cleanup_lambda_role
                 slurm_construct=self.scheduler_resources,
             )
 
-    def _add_login_nodes_resources(self):
+    def _add_login_nodes_resources(self, head_eni):
```
EddyMM: `_head_eni` is already an instance variable, I don't think we need to pass it as an argument. We can go ahead and use `self._head_eni`.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -177,14 +263,14 @@ def __init__(
...
-        self._add_resources()
+        self._add_resources(head_eni)
 
     @property
     def stack_name(self):
         """Name of the CFN stack."""
         return Stack.of(self.nested_stack_parent).stack_name
 
-    def _add_resources(self):
+    def _add_resources(self, head_eni):
```
EddyMM: `head_eni` is already an instance variable (and it's assigned a value). We can use `self.head_eni`.

cli/src/pcluster/templates/cluster_stack.py
NSsirena: Good catch!

I'll fix it

cli/src/pcluster/templates/login_nodes_stack.py
NSsirena: In this case it is not. It is a member of `class Pool(Construct)`.

However I can make it an object variable (in case we need also elsewhere) and use self._head_eni.

cli/src/pcluster/resources/login_node/user_data.sh
NSsirena: Yep. I agree.
I was initially focused on "making it work" and just adding/fixing what strictly necessary.
But since it is now working we can start to remove unused/unnecessary parts.

NSsirena: Fixed

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -83,6 +93,81 @@ def _add_login_nodes_pool_launch_template(self):
...
+                        get_user_data_content("../resources/login_node/user_data.sh"),
+                        {
+                            **{
+                                "RAIDSharedDir": to_comma_separated_string(
+                                    self._shared_storage_mount_dirs[SharedStorageType.RAID]
+                                ),
+                                "RAIDType": to_comma_separated_string(
+                                    self._shared_storage_attributes[SharedStorageType.RAID]["Type"]
+                                ),
+                                "DisableMultiThreadingManually": "false",
```
jdeamicis, line 122: We probably don't need this.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -83,6 +93,81 @@ def _add_login_nodes_pool_launch_template(self):
...
+                    Fn.sub(
+                        get_user_data_content("../resources/login_node/user_data.sh"),
+                        {
+                            **{
+                                "RAIDSharedDir": to_comma_separated_string(
+                                    self._shared_storage_mount_dirs[SharedStorageType.RAID]
+                                ),
+                                "RAIDType": to_comma_separated_string(
+                                    self._shared_storage_attributes[SharedStorageType.RAID]["Type"]
+                                ),
```
jdeamicis, line 121: I'm not sure this is needed.

jdeamicis: Let's open a follow-up task to clean up the dna.json for compute and login nodes. I'm not sure we need everything listed here.

jdeamicis (dismissed):
Let's open a follow-up task to clean up the dna.json for compute and login nodes. I'm not sure we need everything listed here.

jdeamicis (approved):

PR #5459, opened 2023-07-07
Force the remote command executor to use only the key provided and add ubuntu2204 to integ test utils

dreambeyondorange:
Added an additional log line to make it clear to the user what the parameters are for the integ test connection via SSH.

Forcing the use of only the key provided will prevent paramiko from trying to use another key that might be found on the system via the `look_for_keys` parameter.  https://docs.paramiko.org/en/latest/api/client.html#paramiko.client.SSHClient.connect

The integration test framework needs a user defined for each OS and an entry in the `OS_TO_OFFICIAL_AMI_NAME_OWNER_MAP` dictionary

### Tests
* Local integ tests and a full new os test suite on jenkins

### Checklist

### DIFF
--- a/tests/integration-tests/remote_command_executor.py
+++ b/tests/integration-tests/remote_command_executor.py
@@ -46,7 +46,10 @@ class RemoteCommandExecutor:
             "host": node_ip,
             "user": username,
             "forward_agent": False,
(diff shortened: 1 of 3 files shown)


enrico-usai (approved):
When doing this kind of changes is useful to describe "why" we're changing the logic.

There was an error? what was the issue?

It's also useful to reference official guide for the new settings we're adding.

PR #5474, opened 2023-07-12
Avoid to call describe login nodes if they are not in the cluster config

maurizp:
### Description of changes
* Avoid to call describe login nodes if they are not in the cluster config

### Tests
* Manual test with a cluster with and without login nodes pool. Verified with the debugger the sequence of the internal calls.

### References
* Fix to the commit https://github.com/aws/aws-parallelcluster/commit/d9f40324303c0b62c02f68ca40eaa19d1f40c660

### Checklist

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -285,7 +285,8 @@ class Cluster:
     def login_nodes_status(self):
         """Status of the login nodes pool."""
         login_nodes_status = LoginNodesStatus(self.stack_name)
-        login_nodes_status.retrieve_data()
+        if self.config.login_nodes:
+            login_nodes_status.retrieve_data()
         return login_nodes_status
 
     @property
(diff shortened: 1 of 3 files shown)


NSsirena (approved):

cli/src/pcluster/models/cluster.py
```diff
@@ -285,7 +285,8 @@ def compute_fleet_status(self) -> ComputeFleetStatus:
     def login_nodes_status(self):
         """Status of the login nodes pool."""
         login_nodes_status = LoginNodesStatus(self.stack_name)
-        login_nodes_status.retrieve_data()
+        if self.config.login_nodes:
+            login_nodes_status.retrieve_data()
         return login_nodes_status
```
jdeamicis, line 290: Do we need to return an empty LoginNodesStatus object if we don't use login nodes?

maurizp: Yes, basically the LoginNodesStatus exposes the information about the fact that there are no load balancers available.

PR #5482, opened 2023-07-13
[Login Nodes] Set UpdatePolicy for LoginNodes parameters.

gmarciani:
### Changes
Set UpdatePolicy for LoginNodes parameters.
In particular, all parameters require login nodes to be stopped, with the only exception for Count and GracetimePeriod.

### Tests
1. Existing unit tests
2. New unit tests
3. Manual verifications

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -950,7 +950,19 @@ class QueueIamSchema(IamSchema):
 class LoginNodesIamSchema(BaseIamSchema):
     """Represent the IAM schema of LoginNodes."""
 
-    instance_profile = fields.Str(validate=validate.Regexp("^arn:.*:instance-profile/"))
+    instance_role = fields.Str(
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP}, validate=validate.Regexp("^arn:.*:role/")
+    )
+
+    additional_iam_policies = fields.Nested(
+        AdditionalIamPolicySchema,
+        many=True,
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP, "update_key": "Policy"},
+    )
+
+    instance_profile = fields.Str(
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP}, validate=validate.Regexp("^arn:.*:instance-profile/")
+    )
 
     @post_load
     def make_resource(self, data, **kwargs):
@@ -1260,7 +1272,10 @@ class HeadNodeSchema(BaseSchema):
 class LoginNodesImageSchema(BaseSchema):
     """Represent the Image schema of LoginNodes."""
 
-    custom_ami = fields.Str(validate=validate.Regexp(PCLUSTER_AMI_ID_REGEX))
+    custom_ami = fields.Str(
+        validate=validate.Regexp(PCLUSTER_AMI_ID_REGEX),
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP},
+    )
 
     @post_load
     def make_resource(self, data, **kwargs):
@@ -1270,6 +1285,8 @@ class LoginNodesImageSchema(BaseSchema):
 
 class LoginNodesSshSchema(BaseSshSchema):
     """Represent the Ssh schema of LoginNodes."""
+
+    key_name = fields.Str(metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
 
     @post_load
     def make_resource(self, data, **kwargs):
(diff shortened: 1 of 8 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -950,7 +950,19 @@ class QueueIamSchema(IamSchema):
...
+
+    additional_iam_policies = fields.Nested(
+        AdditionalIamPolicySchema,
+        many=True,
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP, "update_key": "Policy"},
+    )
+
+    instance_profile = fields.Str(
+        metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP}, validate=validate.Regexp("^arn:.*:instance-profile/")
+    )
```
NSsirena: Since these regex are used in many places, it would be better to define them in `constants.py` and use the same constant everywhere.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1297,14 +1323,20 @@ def make_resource(self, data, **kwargs):
...
-    iam = fields.Nested(LoginNodesIamSchema)
-    gracetime_period = fields.Int(validate=validate.Range(min=1, max=120))
+    name = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
+    instance_type = fields.Str(required=True, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
+    image = fields.Nested(LoginNodesImageSchema, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
+    networking = fields.Nested(
+        LoginNodesNetworkingSchema, required=True, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP}
+    )
+    count = fields.Int(
+        required=True, validate=validate.Range(min=0), metadata={"update_policy": UpdatePolicy.SUPPORTED}
```
NSsirena: Should set default value to `0` with `missing=0`

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1297,14 +1323,20 @@ def make_resource(self, data, **kwargs):
...
+    )
+    ssh = fields.Nested(LoginNodesSshSchema, required=True, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
+    iam = fields.Nested(LoginNodesIamSchema, metadata={"update_policy": UpdatePolicy.LOGIN_NODES_STOP})
+    gracetime_period = fields.Int(
+        validate=validate.Range(min=1, max=120), metadata={"update_policy": UpdatePolicy.SUPPORTED}
```
NSsirena: Should set default value to `60` with `missing=60`

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1733,7 +1765,7 @@ def make_resource(self, data, **kwargs):
 class ClusterSchema(BaseSchema):
     """Represent the schema of the Cluster."""
 
-    login_nodes = fields.Nested(LoginNodesSchema, many=False)
+    login_nodes = fields.Nested(LoginNodesSchema, many=False, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
```
NSsirena: Not sure about the behavior defined here.

Should this prevent to remove/add a `login_nodes` section from the config ?
That is i'm allowed to create a cluster with or without login nodes, but once I make a decision I cannot change it ?

Is this the desired behavior ?

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -1652,3 +1652,115 @@ def test_condition_checker_managed_fsx(
     assert_that(condition_checker_managed_fsx(change, patch)).is_equal_to(expected_subnet_updated)
     assert_that(fail_reason_managed_fsx(change, patch)).is_equal_to(expected_fail_reason)
     assert_that(actions_needed_managed_fsx(change, patch)).is_equal_to(expected_action_needed)
+
```
NSsirena, line 1655: Are we allowed to `add` or `remove` a `LoginNodes` section from a running cluster ?

If the answer is yes we need a test for this.

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: Default values are not set in the schema but at the config resource level.
Currently it is set to 1 and it should be 0 so i will fix in the next iteration.
The reason why the default values are set at the resource level rather than in schema is that there are default values that depends on the environment and other configurations that are not accessible from the schema.

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: Same as above: we typically set default values in config resources rather than config schema. I'm not against setting it at the schema level, but i think we should keep consistency

cli/src/pcluster/schemas/cluster_schema.py
gmarciani: I think this is the expected behaviour in MVP. I will clarify with the feature owner

gmarciani: Discussed with feature owner: we should support the update

cli/tests/pcluster/config/test_update_policy.py
gmarciani: I agree. I've checked how to do that and unfortunately it's not enough to add a test case in this unit test, but require a more complex unit test. i'll address this comment in a follow up PR.

NSsirena (dismissed):

NSsirena (approved):

PR #5515, opened 2023-07-19
Add a complete lifecycle hook signal in LoginNodes user_data.sh and related support to avoid MITM waring

hehe7318:
### Description of changes
* Add a complete lifecycle hooksignal in LoginNodes user_data.sh
* Add a new lifecycle hook of launching
* Add IAM policy for complete lifecycle hook signal
* Modify related unit test to adopt new changes
* Change Id of pool stack to make ASG name and lifecycle hook name unique

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -218,7 +218,7 @@ else
 fi
 
 INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
-aws autoscaling complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id ${INSTANCE_ID} --lifecycle-hook-name "${LifecycleHookName}" --auto-scaling-group-name "${AutoScalingGroupName}"  --region ${AWS::Region}
+aws autoscaling complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id $INSTANCE_ID --lifecycle-hook-name "${LifecycleHookName}" --auto-scaling-group-name "${AutoScalingGroupName}"  --region ${AWS::Region}
 
 # End of file
 --==BOUNDARY==
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -90,7 +90,8 @@ write_files:
           "stack_name": "${AWS::StackName}",
           "stack_arn": "${AWS::StackId}",
           "use_private_hostname": "${UsePrivateHostname}",
-          "auto_scaling_group_name": "${AutoScalingGroupName}"
+          "auto_scaling_group_name": "${AutoScalingGroupName}",
+          "lifecycle_hook_name": "${LifecycleHookName}"
         }
       }
   - path: /etc/chef/client.rb
@@ -216,7 +217,8 @@ else
   timeout ${Timeout} /tmp/bootstrap.sh || error_exit
 fi
 
-cfn-signal --stack ${AWS::StackName} --resource ${AutoScalingGroupName} --region ${AWS::Region} --exit-code 0
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+aws autoscaling complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id ${INSTANCE_ID} --lifecycle-hook-name "${LifecycleHookName}" --auto-scaling-group-name "${AutoScalingGroupName}"  --region ${AWS::Region}
 
 # End of file
 --==BOUNDARY==
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -216,7 +216,7 @@ else
   timeout ${Timeout} /tmp/bootstrap.sh || error_exit
 fi
 
-cfn-signal --stack ${AWS::StackName} --resource ${AutoScalingGroupName} --region ${AWS::Region}
+cfn-signal --stack ${AWS::StackName} --resource ${AutoScalingGroupName} --region ${AWS::Region} --exit-code 0
 
 # End of file
 --==BOUNDARY==
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -89,7 +89,7 @@ write_files:
           "scheduler": "${Scheduler}",
           "stack_name": "${AWS::StackName}",
           "stack_arn": "${AWS::StackId}",
-          "use_private_hostname": "${UsePrivateHostname}"
+          "use_private_hostname": "${UsePrivateHostname}",
           "auto_scaling_group_name": "${AutoScalingGroupName}"
         }
       }
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -90,6 +90,7 @@ write_files:
           "stack_name": "${AWS::StackName}",
           "stack_arn": "${AWS::StackId}",
           "use_private_hostname": "${UsePrivateHostname}"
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -215,5 +216,7 @@ else
   timeout ${Timeout} /tmp/bootstrap.sh || error_exit
 fi
 
+cfn-signal --stack ${AWS::StackName} --resource ${AutoScalingGroupName} --region ${AWS::Region} --exit-code 0
```
gmarciani: `cfn-signal` is not the tool we should use here. We should use [aws autoscaling group complete-lifecycle-action](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/autoscaling/complete-lifecycle-action.html), instead.

`cfn-signal` is the CloudFormation tool that allows you to send a positive/negative signal to a CloudFormation stack for a given CloudFormation resource. It's meant to tell CloudFormation the outcome of the creation/update for a given resource.

This is not want we want to do here, even if it is similar to it.

Our goal here is to notify the service Autoscaling Group (not CloudFormation) that the given AutoScalingGroup (not CloudFormation resource) can complete its lifecycle hook.

gmarciani: The solution should be something like this:

```
aws autoscaling complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id $INSTANCE_ID --lifecycle-hook-name "${lifecycleHookName}" --auto-scaling-group-name "${autoScalingGroupName}"  --region ${AWS::Region}
```

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -872,6 +872,14 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
                     )
                 ],
             ),
+            iam.PolicyStatement(
+                sid="CloudFormation",
+                actions=[
+                    "cloudformation:SignalResource",
+                ],
+                effect=iam.Effect.ALLOW,
+                resources=["*"],
```
gmarciani: A wildcard resource is not a good solution because it makes the policy too permissive. Try restricting the permissions as much as possible.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -872,6 +872,14 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
                     )
                 ],
             ),
+            iam.PolicyStatement(
+                sid="CloudFormation",
+                actions=[
+                    "cloudformation:SignalResource",
```
gmarciani: this action is correct for cfn-signal, but since cfn-signal is the wrong approach you should instead replace it with the permissions required by the complete-lifecycle-action operation.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -232,19 +234,29 @@ def _add_login_nodes_pool_auto_scaling_group(self):
...
             heartbeat_timeout=self._pool.gracetime_period * 60,
         )
 
+    def _add_launching_lifecycle_hook(self, auto_scaling_group):
+        return autoscaling.CfnLifecycleHook(
+            self,
+            "LoginNodesASGLifecycleHookLaunching",
+            auto_scaling_group_name=auto_scaling_group.ref,
+            lifecycle_transition="autoscaling:EC2_INSTANCE_LAUNCHING",
+            default_result="ABANDON",
```
gmarciani, line 260: Better to set a timeout

gmarciani (changes requested):

cli/src/pcluster/resources/login_node/user_data.sh
hehe7318: Done! Thank you for correcting.

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Done!

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Done!

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Have set it to 600 seconds.

cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -215,5 +217,8 @@ else
   timeout ${Timeout} /tmp/bootstrap.sh || error_exit
 fi
 
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
```
gmarciani, line 220: This is an IMDSv1 call. We should use IMDSv2 instead.

```
TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 300")
INSTANCE_ID=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/instance-id)
```

cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -89,7 +89,9 @@ write_files:
           "scheduler": "${Scheduler}",
           "stack_name": "${AWS::StackName}",
           "stack_arn": "${AWS::StackId}",
-          "use_private_hostname": "${UsePrivateHostname}"
+          "use_private_hostname": "${UsePrivateHostname}",
+          "auto_scaling_group_name": "${AutoScalingGroupName}",
+          "lifecycle_hook_name": "${LifecycleHookName}"
```
gmarciani, line 94: we have two different lifecycle hook for this ASG: one on start and one for terminate. Better to be more exlicit.

cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -89,7 +89,9 @@ write_files:
           "scheduler": "${Scheduler}",
           "stack_name": "${AWS::StackName}",
           "stack_arn": "${AWS::StackId}",
-          "use_private_hostname": "${UsePrivateHostname}"
+          "use_private_hostname": "${UsePrivateHostname}",
+          "auto_scaling_group_name": "${AutoScalingGroupName}",
```
gmarciani, line 93: why do we need these in dna.json? If they are only used in user data they do not need to go in dna.json

gmarciani (changes requested):

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -192,6 +193,8 @@ def _add_login_nodes_pool_launch_template(self):
                                         NODE_BOOTSTRAP_TIMEOUT,
                                     )
                                 ),
+                                "AutoScalingGroupName": f"{self._login_nodes_stack_id}-AutoScalingGroup",
```
gmarciani, line 196: We should not redefine the ASG and the hook name, but instead retrieve them from the stack, otherwise we risk to have naming misalignments.

gmarciani (changes requested):

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -83,6 +83,7 @@ def _add_login_node_iam_resources(self):
             self._pool,
             self._shared_storage_infos,
             self._pool.name,
+            f"{self._login_nodes_stack_id}-AutoScalingGroup",
```
gmarciani, line 86: Why do we need this here?

gmarciani (changes requested):

cli/src/pcluster/templates/login_nodes_stack.py
gmarciani: oh i got it, nvm

gmarciani:
Closed in favor of https://github.com/aws/aws-parallelcluster/pull/5520

PR #5518, opened 2023-07-20
[Directory Service] Remove redundant tests, keeping only SimpleAD with LDAP and MicrosoftAD with LDAPS.

gmarciani:
### Description of changes
Remove redundant tests, keeping only:
1. SimpleAD with LDAP, which is the simplest case
2. MicrosoftAD with LDAPS, which is the most complete case

### Tests
Not needed.

### DIFF
--- a/tests/integration-tests/tests/ad_integration/test_ad_integration.py
+++ b/tests/integration-tests/tests/ad_integration/test_ad_integration.py
@@ -663,8 +663,8 @@ def _check_ssh_key(user, ssh_generation_enabled, remote_command_executor, schedu
     [
         ("SimpleAD", "ldap", False),
         # ("SimpleAD", "ldaps", False),
-        ("SimpleAD", "ldaps", True),
-        ("MicrosoftAD", "ldap", False),
+        # ("SimpleAD", "ldaps", True),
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -663,8 +663,8 @@ def _check_ssh_key(user, ssh_generation_enabled, remote_command_executor, schedu
     [
         ("SimpleAD", "ldap", False),
         # ("SimpleAD", "ldaps", False),
-        ("SimpleAD", "ldaps", True),
-        ("MicrosoftAD", "ldap", False),
+        # ("SimpleAD", "ldaps", True),
```
dreambeyondorange, line 666: should we just delete the tests instead of leaving a comment?

dreambeyondorange (approved):

tests/integration-tests/tests/ad_integration/test_ad_integration.py
gmarciani: I understand your point, but I prefer not to lose the track of the previous coverage

PR #5519, opened 2023-07-20
Add read_only user in LoginNode dna.json

NSsirena:
### Description of changes
* Add read_only user in LoginNode dna.json
  * This parameter is required to check the connection with the remote directory service and sync the user db.

### Tests
* N/A
* unit test will be added in a following PR

### References
* [Check AD connection](https://github.com/aws/aws-parallelcluster-cookbook/pull/2372)

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/login_nodes_stack.py
+++ b/cli/src/pcluster/templates/login_nodes_stack.py
@@ -132,6 +132,7 @@ class Pool(Construct):
                                 "CustomAwsBatchCliPackage": self._config.custom_aws_batch_cli_package or "",
                                 "CWLoggingEnabled": "true" if self._config.is_cw_logging_enabled else "false",
                                 "DirectoryServiceEnabled": str(ds_config is not None).lower(),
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -132,6 +132,7 @@ def _add_login_nodes_pool_launch_template(self):
                                 "CustomAwsBatchCliPackage": self._config.custom_aws_batch_cli_package or "",
                                 "CWLoggingEnabled": "true" if self._config.is_cw_logging_enabled else "false",
                                 "DirectoryServiceEnabled": str(ds_config is not None).lower(),
+                                "DirectoryServiceReadOnlyUser": ds_config.domain_read_only_user,
```
gmarciani: What if DirectorySerbvice configuration is missing? We should check for ds_config being None, right?

NSsirena: Yes!
Good catch !
A neuron in my brain thought about this but it was suppressed ! :D

gmarciani (approved):

PR #5520, opened 2023-07-21
[Login Nodes] Add lifecycle hook for node launch in login nodes ASG

gmarciani:
### Description of changes
Add lifecycle hook for node launch in login nodes ASG.

Original PR: https://github.com/aws/aws-parallelcluster/pull/5515

### Tests
1. Cluster creation both in happy and unhappy case

### DIFF
--- a/cli/src/pcluster/templates/login_nodes_stack.py
+++ b/cli/src/pcluster/templates/login_nodes_stack.py
@@ -83,6 +83,7 @@ class Pool(Construct):
             self._pool,
             self._shared_storage_infos,
             self._pool.name,
+            f"{self._login_nodes_stack_id}-AutoScalingGroup",
         )
         self._instance_profile = self._iam_resource.instance_profile
         self._instance_role = self._iam_resource.instance_role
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -225,27 +230,40 @@ def _add_login_nodes_pool_auto_scaling_group(self):
...
 
+    def _add_launching_lifecycle_hook(self, auto_scaling_group):
+        return autoscaling.CfnLifecycleHook(
+            self,
+            "LoginNodesASGLifecycleHookLaunching",
+            auto_scaling_group_name=auto_scaling_group.ref,
+            lifecycle_hook_name=f"{self._login_nodes_stack_id}-LoginNodesLaunchingLifecycleHook",
+            lifecycle_transition="autoscaling:EC2_INSTANCE_LAUNCHING",
+            default_result="ABANDON",
+            heartbeat_timeout=600,
```
NSsirena, line 264: Nitpick: do we want this hardcoded or controllable through a config parameter (even non customer facing) ?

This could help for tests...

gmarciani: If you're referring to the fact that we are hardwiring the lifecycle hook name: it requires paradoxically some refactoring because there is a circular dependency between ASG and launch template. Solving the circular dependency would require a refactoring that it's not worth it for the beta.

If you are referring to the customizability of the timeout I like your proposal, but it should be at most a dev settings and i would introduce this customizability in another PR not considering it beta blocking.

What do you think?

NSsirena: :+1:

I was referring to the timeout ..but I agree we can do it in another PR

NSsirena (approved):

PR #5525, opened 2023-07-25
Add Integration test for cluster Updates (add, remove LN pools)

hehe7318:
### Description of changes
* Add Integration test for updating cluster by adding and removing LoginNodes pools and related support

### Checklist


tests/integration-tests/tests/update/test_update.py
```diff
@@ -1482,3 +1482,67 @@ def _wait_until_instances_are_stopped(cluster):
...
+    # Start a cluster without login nodes
+    initial_config = pcluster_config_reader(config_file="pcluster_create_without_login_nodes.config.yaml")
+    cluster = clusters_factory(initial_config)
+
+    # Describe cluster, verify the response is without login node section
+    cluster_info = cluster.describe_cluster()
+    assert_that(cluster_info).is_not_none()
+    assert_that(cluster_info).does_not_contain("loginNodes")
+
+    # Update the cluster adding 3 head node
```
NSsirena: typo: adding 3 'login' nodes

tests/integration-tests/tests/update/test_update.py
```diff
@@ -1482,3 +1482,67 @@ def _wait_until_instances_are_stopped(cluster):
...
+    assert_that(cluster_info["loginNodes"]["address"]).is_not_none()
+    assert_that(cluster_info["loginNodes"]["healthyNodes"] + cluster_info["loginNodes"]["unhealthyNodes"]).is_equal_to(
+        0
+    )
+
+    # Describe cluster instances, verify the response doesn't contain login nodes
+    instances = cluster.get_cluster_instance_ids(node_type="LoginNode")
+    assert_that(len(instances)).is_equal_to(0)
+
+    # Update the cluster with count = 2
```
NSsirena: This is somewhat redundant .. we already tested that
* we can add a LoginNode pool in a running cluster launched without them
* we can increase the number of login nodes (0 -> 3) 
* we can decrease the number of login nodes (3 -> 0)

Increasing again to two instances is basically the same test we did when we changed from 0 to 3.

tests/integration-tests/tests/update/test_update.py
```diff
@@ -1482,3 +1482,67 @@ def _wait_until_instances_are_stopped(cluster):
...
+    cluster_info = cluster.describe_cluster()
+    assert_that(cluster_info).is_not_none()
+    assert_that(cluster_info).contains("loginNodes")
+    assert_that(cluster_info["loginNodes"]["healthyNodes"] + cluster_info["loginNodes"]["unhealthyNodes"]).is_equal_to(
+        2
+    )
+
+    # Describe cluster instances, verify the response contains two login nodes
+    instances = cluster.get_cluster_instance_ids(node_type="LoginNode")
+    assert_that(len(instances)).is_equal_to(2)
```
NSsirena: As a final step instead of increasing from 0 to 2 we could instead test that 
* we can remove a Login Node pool from a running cluster

so update the cluster using the initial configuration `"pcluster_create_without_login_nodes.config.yaml"`

tests/integration-tests/tests/update/test_update.py
hehe7318: Thank you for reminding. Have corrected it.

tests/integration-tests/tests/update/test_update.py
hehe7318: Makes sense!

tests/integration-tests/tests/update/test_update.py
hehe7318: I have modified it! Thank you!

NSsirena (dismissed):

NSsirena (dismissed):

NSsirena (approved):

Issue #546, opened 2018-09-15
I hope to move cnfcluster to China region, what suggestions?


frank622:
How to modify this file “cfncluster/cloudformation/cfncluster.cfn.json”
Any  suggestions?

demartinofra:
Hi,

Sorry for the late reply. We added support for China regions cn-north-1 and cn-northwest-1 as part of the last release of AWS ParallelCluster (2.1.1).

Francesco

PR #5534, opened 2023-07-26
Extend fast ICE failover to include nodes that are launched using job-level scaling

EddyMM:
### Description of changes
* Job-level scaling attempts to launch nodes exclusively allocated to a job (or otherwise terminate launched capacity quickly)
* These changes test the functionality of fast failover when applied to a queue that handles scaling using job-level strategy. * Jobs submitted to the `queue-job-level` partition will have nodes exclusively allocated to the job and will follow job-scaling logic
* Tests following the current/previous scaling are preserved in the other partition `queue-node-list` are covered by the same fast failover capacity tests.
* The test uses a case of single-job-multi-node which is enough for job-level scaling since it covers a path where not all the nodes are available for the job.

### Tests
* Ran locally

### References
* [Fast insufficient capacity fail-over](https://docs.aws.amazon.com/parallelcluster/latest/ug/slurm-short-capacity-fail-mode-v3.html)
* [Keep fast capacity failover consistent with current behavior #546](https://github.com/aws/aws-parallelcluster-node/pull/546)

### Checklist


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -440,76 +441,82 @@ def test_fast_capacity_failover(
...
     remote_command_executor.run_remote_script(str(test_datadir / "overrides.sh"), run_as_root=True)
 
-    # All nodes
-    nodes_in_scheduler = scheduler_commands.get_compute_nodes("queue1", all_nodes=True)
-    static_nodes, dynamic_nodes = get_partition_nodes(nodes_in_scheduler)
+    for strategy in scaling_strategies:
+        partition = f"queue-{strategy}"
+        # All nodes
+        nodes_in_scheduler = scheduler_commands.get_compute_nodes(partition, all_nodes=True)
+        static_nodes, dynamic_nodes = get_partition_nodes(nodes_in_scheduler)
```
NSsirena, line 448: It doesn't seems the strategy is involved here.

Are we changing the cluster ?
If not we can retrieve this set only once outside the for loop as done before.

NSsirena (approved):

PR #5539, opened 2023-07-27
Add Integration test of testing external FS(ebs and efs) access on LoginNodes -- release-3.7 branch

hehe7318:
### Description of changes
* Add Integration test of testing external FS(ebs) access on LoginNodes and related support
  * Add LoginNodes section in related yaml file when `use_login_nodes` is true and `scheduler == "slurm"`
  * Test ebs correctly mounted on LoginNodes
  * Test ebs correctly shared between LoginNode and ComputeNodes
  * Test ebs correctly shared between HeadNode and LoginNode
  
* Add Integration test of testing external FS(efs) access on LoginNodes and related support
  * Test efs correctly mounted on LoginNodes
  * Test efs correctly shared between LoginNode and ComputeNodes
  * Test efs correctly shared between HeadNode and LoginNode

### Checklist

### DIFF
--- a/tests/integration-tests/tests/storage/storage_common.py
+++ b/tests/integration-tests/tests/storage/storage_common.py
@@ -55,6 +55,21 @@ def get_cluster_subnet_ids_groups(cluster: Cluster, scheduler: str, include_head
 def test_directory_correctly_shared_between_ln_and_hn(
     remote_command_executor_head_node, remote_command_executor_login_node, mount_dir
 ):
+    """
+    This test verifies if a shared directory is correctly shared between the HeadNode and the LoginNode.
+
+    Preconditions:
+    - An EBS volume is expected to be mounted at <HN_folder> on the HeadNode and exported through NFS.
+    - The EBS volume is expected to be mounted at <LN_folder> on the LoginNode through NFS.
+
+    Test Steps:
+    1. Writes a file from the HeadNode to the shared volume.
+    2. Writes a file from the LoginNode to the shared volume.
+    3. Reads both files from both the HeadNode and the LoginNode to validate that the shared volume is working correctly.
+
+    Expected Result:
+    - Both files can be successfully read from both the HeadNode and the LoginNode, proving that the shared volume is correctly shared between the two nodes.
+    """
     logging.info("Testing FS correctly mounted on login nodes")
     head_node_file = random_alphanumeric()
     logging.info(f"Writing HeadNode File: {head_node_file}")
@@ -98,7 +113,7 @@ def verify_directory_correctly_shared(remote_command_executor, mount_dir, schedu
         "B" reads files: ["A-<random_alphanumeric_characters>", "B-<random_alphanumeric_characters>"]
     """
     executor_node_file = random_alphanumeric()
-    logging.info(f"Writing ExecutorNode File: {executor_node_file}")
+    logging.info(f"{remote_command_executor.get_target_host_type()}: Writing File: {executor_node_file}")
     remote_command_executor.run_remote_command(
         "touch {mount_dir}/{executor_node_file} && cat {mount_dir}/{executor_node_file}".format(
             mount_dir=mount_dir, executor_node_file=executor_node_file
@@ -126,7 +141,7 @@ def verify_directory_correctly_shared(remote_command_executor, mount_dir, schedu
         files_to_read=" ".join([f"{mount_dir}/{target_file}" for target_file in files_to_read]),
     )
     # Attempt reading files from executor node
-    logging.info(f"Reading Files: {files_to_read} from executor node")
+    logging.info(f"{remote_command_executor.get_target_host_type()}: Reading Files: {files_to_read}")
     remote_command_executor.run_remote_command(read_all_files_command)
 
     # Submit a "Read" job to each partition
--- a/tests/integration-tests/tests/storage/storage_common.py
+++ b/tests/integration-tests/tests/storage/storage_common.py
@@ -65,10 +65,11 @@ def test_directory_correctly_shared_between_ln_and_hn(
     Test Steps:
     1. Writes a file from the HeadNode to the shared volume.
     2. Writes a file from the LoginNode to the shared volume.
-    3. Reads both files from both the HeadNode and the LoginNode to validate that the shared volume is working correctly.
+    3. Reads both files from both the HeadNode and the LoginNode to validate that shared volume is working correctly.
 
     Expected Result:
-    - Both files can be successfully read from both the HeadNode and the LoginNode, proving that the shared volume is correctly shared between the two nodes.
+    - Both files can be successfully read from both the HeadNode and the LoginNode,
+      proving that the shared volume is correctly shared between the two nodes.
     """
     logging.info("Testing FS correctly mounted on login nodes")
     head_node_file = random_alphanumeric()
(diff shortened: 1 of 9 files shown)


tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -25,32 +26,56 @@
...
+    region,
+    os,
+    scheduler_commands_factory,
+    use_login_node,
 ):
     mount_dir = "ebs_mount_dir"
     kms_key_id = kms_key_factory.create_kms_key(region)
     cluster_config = pcluster_config_reader(
-        mount_dir=mount_dir, ec2_iam_role=kms_key_factory.iam_role_arn, ebs_kms_key_id=kms_key_id
+        mount_dir=mount_dir,
```
NSsirena, line 42: There is no need to duplicate code.
Since we want to avoid running the test if the scheduler is not `slurm` you can add an initial line like 

```
if use_login_node and scheduler != "slurm": 
  return
```

and leave the rest of the test unchanged.

This hold also for the other tests.

hehe7318: Makes sense, I will modify it.

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -52,6 +52,39 @@ def get_cluster_subnet_ids_groups(cluster: Cluster, scheduler: str, include_head
     return compute_subnet_ids
 
 
+def test_directory_correctly_shared_between_ln_and_hn(
+    remote_command_executor_head_node, remote_command_executor_login_node, mount_dir
+):
+    logging.info("Testing FS correctly mounted on login nodes")
```
NSsirena, line 74: Please add a brief comments that describes what the test expect to achieve and which preconditions it expects (if any) ... like
"
It expects an ebs volume to be mounted at <HN_folder> on the HeadNode and exported through NFS
it expects the ebs volume to be mounted at <LN_folder> on the LoginNode through NFS

it will write a file from the HeadNode and a file from the LoginNode in the shared volume

it expects to be able to read both files from the HeadNode and the LoginNode

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -64,16 +97,16 @@ def verify_directory_correctly_shared(remote_command_executor, mount_dir, schedu
         "A" reads files: ["A-<random_alphanumeric_characters>", "B-<random_alphanumeric_characters>"]
         "B" reads files: ["A-<random_alphanumeric_characters>", "B-<random_alphanumeric_characters>"]
     """
-    head_node_file = random_alphanumeric()
-    logging.info(f"Writing HeadNode File: {head_node_file}")
+    executor_node_file = random_alphanumeric()
+    logging.info(f"Writing ExecutorNode File: {executor_node_file}")
```
NSsirena: `ExecutorNode` is not very speaking ... while reading the logs we will not be able to figure out in which node the command was actually executed.

We can add a label in the RemoteCommandExecutor 

https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/remote_command_executor.py#L41

here we can add something like 

```
        if compute_node_ip:
            self.target = "ComputeNode"
...
        elif use_login_node:
            self.target = "LoginNode"
...
        else:
            self.target = "HeadNode"
```

and then add a `get_target_host_type()` method that will return `self.target`

and finally change the message in something like 

```
f"{remote_command_executor.get_target_host_type()}: Writing File: {executor_node_file}"
```

tests/integration-tests/tests/storage/storage_common.py
```diff
@@ -92,8 +125,8 @@ def verify_directory_correctly_shared(remote_command_executor, mount_dir, schedu
     read_all_files_command = "cat {files_to_read}".format(
         files_to_read=" ".join([f"{mount_dir}/{target_file}" for target_file in files_to_read]),
     )
-    # Attempt reading files from HeadNode
-    logging.info(f"Reading Files: {files_to_read} from HeadNode")
+    # Attempt reading files from executor node
+    logging.info(f"Reading Files: {files_to_read} from executor node")
```
NSsirena: Same as above...let's try to make it clear from where we are trying to read the files

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -61,7 +88,11 @@ def test_ebs_snapshot(
     clusters_factory,
     scheduler_commands_factory,
     scheduler,
+    use_login_node,
 ):
+    if use_login_node and scheduler != "slurm":
+        return
```
NSsirena: I just discovered we have a more idiomatic way to skip a test 

```
pytest.skip(skip_message)
```

that allows us to have a message in the logs that informs us why/when we are skipping a test.

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -162,7 +235,11 @@ def test_ebs_existing(
     snapshots_factory,
     clusters_factory,
     scheduler_commands_factory,
+    use_login_node,
 ):
+    if use_login_node and scheduler != "slurm":
+        return
```
NSsirena: See above

tests/integration-tests/tests/storage/test_ebs.py
```diff
@@ -20,37 +20,64 @@
 from tests.storage.kms_key_factory import KMSKeyFactory
 from tests.storage.storage_common import (
     assert_subnet_az_relations_from_config,
+    test_directory_correctly_shared_between_ln_and_hn,
     test_ebs_correctly_mounted,
     verify_directory_correctly_shared,
 )
 
 
+@pytest.mark.parametrize("use_login_node", [True, False])
```
NSsirena: General comment that holds for all tests: 

I initially hoped that we could reuse the existing test and simply parameterize it.
However seeing the changes you had to make this is not possible and you were forced to add a new section with custom operations to run on the login nodes.

This makes the parameterization counterproductive because we are running the test twice (use_login_node=[False, True] ) and
 - in the first run we are only executing the tests on the HeadNode
 - in the second run we are executing the test both on the HeadNode AND the LoginNodes
 
This means that we are running the test in the HeadNode twice (without changes) and that we are creating two clusters without a real advantage.

So I propose to remove the parameterization 

```
@pytest.mark.parametrize("use_login_node", [True, False])
```
from all the tests and simply leave the extra section `LoginNodes` in the configs and the extra checks for the LoginNodes at the end of the tests.

This will allow us to test both HeadNode and LoginNode with a single cluster.

tests/integration-tests/tests/storage/storage_common.py
hehe7318: Thanks for reminding. Done!

tests/integration-tests/tests/storage/storage_common.py
hehe7318: Done!

tests/integration-tests/tests/storage/test_ebs.py
hehe7318: This code is deleted. We consider `use_login_node` always be True.

tests/integration-tests/tests/storage/test_ebs.py
hehe7318: This code is deleted. We consider use_login_node always be True.

tests/integration-tests/tests/storage/test_ebs.py
hehe7318: Makes sense. Done!

NSsirena (dismissed):

NSsirena (approved):

PR #5543, opened 2023-08-01
Test node scaling for exclusive jobs with near all-or-nothing launch

EddyMM:
### Description of changes
* When partial or no capacity is available, the launched instances are quickly terminated to reduce IDLE time
* When full capacity is available, the launched instances are launched successfully
* The mechanism used to emulate an Insufficient Capacity Error has been moved to a common directory for reuse across tests

### Tests
* Ran locally

### References
* https://github.com/aws/aws-parallelcluster-node/pull/527
* https://github.com/aws/aws-parallelcluster-node/pull/541

### Checklist


tests/integration-tests/tests/common/scaling_common.py
```diff
@@ -156,3 +160,14 @@ def get_batch_ce_min_size(stack_name, region):
         .get("computeResources")
         .get("minvCpus")
     )
+
+
+def emulate_ice_in_cluster(cluster):
```
lukeseawalker: is this the setup of the override for EC2 launch API calls, so to emulate ICE? Can we find a better naming, so that is clear that it's not emulating any EC2 launch instance call, but only setting up things?

lukeseawalker: e.g. `setup_ec2_launch_override_to_emulate_ice` ?

tests/integration-tests/tests/common/scaling_common.py
```diff
@@ -156,3 +160,14 @@ def get_batch_ce_min_size(stack_name, region):
...
+    """
+    Includes an override file that emulates an ICE error in a cluster.
+
+    It applies the override patch to launch ice for nodes in
+    "ice-compute-resource" and "ice-cr-multiple" compute resources with an ICE error
```
lukeseawalker: NIT is it possible to have these names as input? So we don't have to remember for which queue the ice is emulated, and build the test config accordingly, but it's the other way around, i.e. the test config is built and than the override is setup based on the name of the queue set for the test

EddyMM: It should be possible to pass the target queue names as parameters for the overrides script. I'll run some tests and confirm.

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -87,6 +87,79 @@ def test_multiple_jobs_submission(
...
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = scheduler_commands_factory(remote_command_executor)
+
+    setup_ec2_launch_override_to_emulate_ice(
+        cluster,
+        single_instance_type_ice_cr="ice-compute-resource",
+        multi_instance_types_ice_cr="ice-cr-multiple",
+        multi_instance_types_exp_cr="exception-cr-multiple",
```
lukeseawalker: are "ice-compute-resource" and "exception-cr-multiple" needed for this test? I could not see these CRs in the pcluster config related to this test

EddyMM: Only "ice-cr-multiple" is needed. Good catch, updating.

lukeseawalker (dismissed):

lukeseawalker (approved):

PR #5553, opened 2023-08-02
Keypair check

dreambeyondorange:
### Description of changes

### Tests

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -759,9 +759,6 @@ class _BaseSsh(Resource):
         super().__init__(**kwargs)
         self.key_name = Resource.init_param(key_name)
 
-    def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
-        self._register_validator(KeyPairValidator, key_name=self.key_name)
-
 
 class HeadNodeSsh(_BaseSsh):
     """Represent the SSH configuration for HeadNode."""
@@ -1535,6 +1532,7 @@ class BaseClusterConfig(Resource):
             volume_size=root_volume_size,
             volume_iops=root_volume.iops,
         )
+        self._register_validator(KeyPairValidator, key_name=self.head_node.ssh.key_name, os=self.image.os)
 
     def _register_storage_validators(self):  # noqa: C901 FIXME: function too complex
         if self.shared_storage:
--- a/cli/src/pcluster/validators/ec2_validators.py
+++ b/cli/src/pcluster/validators/ec2_validators.py
@@ -171,10 +171,15 @@ class KeyPairValidator(Validator):
     Verify the given key pair is correct.
     """
 
-    def _validate(self, key_name: str):
+    def _validate(self, key_name: str, os: str):
         if key_name:
             try:
-                AWSApi.instance().ec2.describe_key_pair(key_name)
+                key_pair_data = AWSApi.instance().ec2.describe_key_pair(key_name)
+                if os == "ubuntu2204" and key_pair_data["KeyPairs"][0]["KeyType"] == "rsa":
+                    self._add_failure(
+                        "Ubuntu 22.04 does not support RSA keys.  Please generate and use an ed25519 key",
+                        FailureLevel.ERROR,
+                    )
             except AWSClientError as e:
                 self._add_failure(str(e), FailureLevel.ERROR)
         else:
--- a/cli/tests/pcluster/validators/test_ec2_validators.py
+++ b/cli/tests/pcluster/validators/test_ec2_validators.py
@@ -332,18 +332,29 @@ def test_instance_type_base_ami_compatible_validator(
     assert_failure_messages(actual_failures, expected_message)
 
 
+GoodKeyPairsDict = {"KeyPairs": [{"KeyType": "ed25519"}]}
+BadKeyPairsDict = {"KeyPairs": [{"KeyType": "rsa"}]}
+
+
 @pytest.mark.parametrize(
-    "key_pair, side_effect, expected_message",
+    "key_pair, os, side_effect, expected_message",
     [
-        ("key-name", None, None),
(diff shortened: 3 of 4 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -759,9 +759,6 @@ def __init__(self, key_name: str = None, **kwargs):
         super().__init__(**kwargs)
         self.key_name = Resource.init_param(key_name)
 
-    def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
```
enrico-usai, line 762: By removing the register_validators from here you're removing the check for LoginNodesSsh. You need to restore it.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -171,10 +171,15 @@ class KeyPairValidator(Validator):
     Verify the given key pair is correct.
     """
 
-    def _validate(self, key_name: str):
+    def _validate(self, key_name: str, os: str):
         if key_name:
             try:
-                AWSApi.instance().ec2.describe_key_pair(key_name)
+                key_pair_data = AWSApi.instance().ec2.describe_key_pair(key_name)
+                if os == "ubuntu2204" and key_pair_data["KeyPairs"][0]["KeyType"] == "rsa":
```
enrico-usai, line 178: rather than parsing the key_pair_data here we could define a `KeyPairInfo` class in the `cli/src/pcluster/aws/aws_resources.py` and encapsulate the internal parsing logic there. 

The `describe_key_pair` method should then return an instance of `KeyPairInfo` and here we can just retrieve the type with: `key_pair.key_type`, this permits to decouple validator/pcluster code from AWS Api internal dictionaries

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -171,10 +171,15 @@ class KeyPairValidator(Validator):
...
+                    self._add_failure(
+                        "Ubuntu 22.04 does not support RSA keys.  Please generate and use an ed25519 key",
```
enrico-usai, line 180: nit: extra space before "Please" word.

enrico-usai, line 180: you can think to generalize this logic, in future AL2023 will have to do the same check, so we could create a dictionary and do something like:

```python
os_key_type = {"ubuntu22": "ed25519"}
if os in os_key_type:
    required_key_type = os_key_type.get(os)
    if key_pair.key_type != required_key_type:
        self._add_failure(f"{os} requires {required_key_type} key. Please generate and use a {required_key_type} key",...
```

cli/tests/pcluster/validators/test_ec2_validators.py
```diff
@@ -332,18 +332,29 @@ def test_instance_type_base_ami_compatible_validator(
...
+            None,
+            AWSClientError(function_name="describe_key_pair", message="does not exist"),
+            "does not exist",
+        ),
+        (GoodKeyPairsDict, "ubuntu2204", None, None),
+        (BadKeyPairsDict, "ubuntu2204", None, "Ubuntu 22.04 does not support RSA keys"),
     ],
 )
-def test_key_pair_validator(mocker, key_pair, side_effect, expected_message):
+def test_key_pair_validator(mocker, key_pair, os, side_effect, expected_message):
```
enrico-usai, line 354: we should add the tests for the other oses as well, to be sure the logic is not executed.

tests/integration-tests/configs/common/common.yaml
```diff
@@ -394,16 +394,16 @@ scaling:
...
-    dimensions:
-      - regions: ["us-gov-west-1"]
-        instances: {{ common.INSTANCES_DEFAULT_X86 }}
-        oss: ["alinux2"]
-        schedulers: ["awsbatch"]
-      - regions: ["ap-northeast-1"]
-        instances: {{ common.INSTANCES_DEFAULT_ARM }}
-        oss: ["alinux2"]
-        schedulers: ["awsbatch"]
+#  test_awsbatch.py::test_awsbatch:
```
enrico-usai, line 397: memo: this should not be part of this patch

enrico-usai:
Memo: before marking it "Ready to review" let's remember to change the title and description

cli/src/pcluster/validators/ec2_validators.py
dreambeyondorange: Sentences are supposed to have two spaces after a period

tests/integration-tests/configs/common/common.yaml
dreambeyondorange: agreed, I was waiting to open this PR until that patch was merged

cli/src/pcluster/validators/ec2_validators.py
enrico-usai: ok there is an ongoing discussion about this: https://www.instructionalsolutions.com/blog/one-space-vs-two-after-period :D

cli/src/pcluster/config/cluster_config.py
dreambeyondorange: This is moving to the BaseClusterConfig for the OS context should should run for any cluster

cli/src/pcluster/validators/ec2_validators.py
dreambeyondorange: I'm OK with one space!

dreambeyondorange: All the keys we generate are ED25519 now, but that doesn't stop users from using different key types like ECDSA, right?  I think we should avoid this complexity for now and we can add the dict when we have more than one case to consider.  And we should probably do the opposite logic, to restrict using RSA rather than checking for the specific use of a single key type

cli/src/pcluster/validators/ec2_validators.py
dreambeyondorange: Thanks for the idea, will do this!

PR #5554, opened 2023-08-02
[Performance Test] Add scaling test to validate scale out to 1000 nodes.

gmarciani:
### Description of changes
Add an integration tests to validate the scale up of a cluster to 1K static nodes and 1K dynamic nodes.

### Tests
Test succeeded with the most recent 3.7.0b1 AMI in eu-west-1 alinux2

### DIFF
--- a/tests/integration-tests/tests/performance_tests/test_scaling.py
+++ b/tests/integration-tests/tests/performance_tests/test_scaling.py
@@ -1,0 +1,46 @@
+import logging
+
+import pytest
+from remote_command_executor import RemoteCommandExecutor
+
+
+@pytest.mark.parametrize(
+    "max_nodes",
+    [1000],
+)
+def test_scaling(
+    vpc_stack,
+    instance,
+    os,
+    region,
+    scheduler,
+    pcluster_config_reader,
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+    max_nodes,
+):
+    cluster_config = pcluster_config_reader(max_nodes=max_nodes)
+    cluster = clusters_factory(cluster_config)
+
+    logging.info("Cluster Created")
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = scheduler_commands_factory(remote_command_executor)
+
+    logging.info(f"Submitting an array of {max_nodes} jobs on {max_nodes} nodes")
+    job_id = scheduler_commands.submit_command_and_assert_job_accepted(
+        submit_command_args={
+            "command": "sleep 1800",
+            "partition": "queue-1",
+            "other_options": f"--array [1-{max_nodes}] --exclusive",
+        }
+    )
+
+    logging.info(f"Waiting for job to be running: {job_id}")
+    scheduler_commands.wait_job_running(job_id)
+    logging.info(f"Job {job_id} is running")
+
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/performance_tests/test_scaling.py
```diff
@@ -0,0 +1,55 @@
...
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = scheduler_commands_factory(remote_command_executor)
+
+    nodes_per_job = 2
+    num_jobs = int(max_nodes / nodes_per_job)
+
+    logging.info(f"Submitting {num_jobs} jobs on {max_nodes} dynamic nodes ({nodes_per_job} jobs/node)")
+    job_ids = []
+    for _ in range(0, num_jobs):
```
enrico-usai: have you considered using a job array rather than submitting multiple jobs? In this way you have a single job to monitor.

```
sbatch --wrap "sleep 100" --array 0-10 --exclusive
```

tests/integration-tests/tests/performance_tests/test_scaling/test_scaling/pcluster.config.yaml
```diff
@@ -0,0 +1,51 @@
...
+    AdditionalIamPolicies:
+      - Policy: arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore #Required to report patching status
+Scheduling:
+  Scheduler: slurm
+  SlurmQueues:
+    - Name: queue-0
+      ComputeResources:
+        - Name: compute-resource-0
+          InstanceType: {{ instance }}
+          MinCount: {{ max_nodes }}
```
enrico-usai, line 20: By having MinCount == max_nodes we can face bootstrap issues at cluster creation time. What about having it set to 0 and just consider the "scaling" case for dynamic node?

I'd remove queue-0 and queue-2 to avoid confusion.

gmarciani: I think the goal of this test is to detect issues both on large scale static and dynamic fleets.
I suggest to keep a static fleet of 1K nodes that, in case of issues, will make the test to fail at cluster creation time.
Why you think it's not necessary?

I would remove the third queue as it is unused

tests/integration-tests/tests/performance_tests/test_scaling.py
gmarciani: Great point thank you! I'll do it

tests/integration-tests/tests/performance_tests/test_scaling/test_scaling/pcluster.config.yaml
enrico-usai: The only thing I'm thinking is that if the creation fails it's not so easy to understand the reason, so we need to check the cluster logs. Anyway I'm fine keeping it.

enrico-usai (approved):

tests/integration-tests/tests/performance_tests/test_scaling/test_scaling/pcluster.config.yaml
EddyMM: [For the static nodes] A trade-off could be to start with a MinCount of `0` and run an update-cluster with `MinCount==max_nodes`.

PR #5596, opened 2023-08-08
Update tests to check if headnode is replaced when we update the SecurityGroups

himani2411:
Check if the HeadNode instance is replaced if the Security Group ID and AdditionalSecurity Groups are updated.

### Checklist

### DIFF
--- a/tests/integration-tests/tests/networking/test_security_groups.py
+++ b/tests/integration-tests/tests/networking/test_security_groups.py
@@ -22,63 +22,75 @@ from utils import check_head_node_security_group, create_hash_suffix, generate_s
 
 
 @pytest.mark.usefixtures("os", "scheduler", "instance")
-def test_additional_sg_and_ssh_from(region, custom_security_groups, pcluster_config_reader, clusters_factory):
+@pytest.mark.parametrize("assign_additional_security_groups", [True, False])
+def test_additional_sg_and_ssh_from(
+    region, custom_security_groups, pcluster_config_reader, clusters_factory, assign_additional_security_groups
+):
     """
     Test when additional_sg ssh_from are provided in the config file
 
     The additional security group should be added to the head and compute nodes. The
+    Test if HeadNode is replaced if we update the SecurityGroups and AdditionalSecurityGroups
     """
-    custom_security_group_id = custom_security_groups
-    ssh_from = "10.11.12.0/32"
-    cluster_config = pcluster_config_reader(additional_sg=custom_security_group_id, ssh_from=ssh_from)
-    cluster = clusters_factory(cluster_config)
-    ec2_client = boto3.client("ec2", region_name=region)
-    instances = _get_instances_by_security_group(ec2_client, custom_security_group_id)
-    logging.info("Asserting that head node and compute node has the additional security group")
-    assert_that(instances).is_length(2)
-    logging.info("Asserting the security group of pcluster is not overwritten by additional seurity group")
-    for instance in instances:
-        assert_that(
-            any(security_group["GroupName"].startswith(cluster.name) for security_group in instance["SecurityGroups"])
-        ).is_true()
-    logging.info("Asserting the security group of pcluster on the head node is aligned with ssh_from")
-    check_head_node_security_group(region, cluster, 22, ssh_from)
-
-
-@pytest.mark.parametrize("assign_additional_security_groups", [False, True])
-@pytest.mark.usefixtures("os", "instance")
-def test_overwrite_sg(
-    region,
-    scheduler,
-    custom_security_groups,
-    pcluster_config_reader,
-    clusters_factory,
-    assign_additional_security_groups,
-):
-    """Test vpc_security_group_id overwrites pcluster default sg on head and compute nodes, efs, fsx"""
     number_of_sgs = 2
 
     custom_security_group_ids = custom_security_groups(number_of_sgs=number_of_sgs)
     default_security_group_id = custom_security_group_ids[0]
 
-    fsx_name, efs_name = "fsx", "efs"
+    ssh_from = "10.11.12.0/32"
     cluster_config = pcluster_config_reader(
-        vpc_security_group_id=default_security_group_id,
-        fsx_name=fsx_name,
-        efs_name=efs_name,
+        default_security_group_id=default_security_group_id,
+        ssh_from=ssh_from,
         assign_additional_security_groups=assign_additional_security_groups,
     )
     cluster = clusters_factory(cluster_config)
+    ec2_client = boto3.client("ec2", region_name=region)
+    instances = _get_instances_by_security_group(ec2_client, default_security_group_id)
+    logging.info("Asserting that head node and compute node has the additional security group")
+    assert_that(instances).is_length(2)
+    if assign_additional_security_groups:
+        logging.info("Asserting the security group of pcluster is not overwritten by additional security group")
+        for instance in instances:
+            assert_that(
+                any(
+                    security_group["GroupName"].startswith(cluster.name)
+                    for security_group in instance["SecurityGroups"]
+                )
+            ).is_true()
+        logging.info("Asserting the security group of pcluster on the head node is aligned with ssh_from")
+        check_head_node_security_group(region, cluster, 22, ssh_from)
+
     head_node_instance_id = cluster.head_node_instance_id
     logging.info(f"HeadNode of the {cluster} is {head_node_instance_id}")
 
+    # Update the Security Groups of HeadNode check if the HeadNode instance is replaced.
+    updated_config_file = pcluster_config_reader(
+        config_file="pcluster.config.update_sg.yaml",
+        default_security_group_id=default_security_group_id,
+        updated_security_group_id=custom_security_group_ids,
+        ssh_from=ssh_from,
+        assign_additional_security_groups=assign_additional_security_groups,
+    )
+    cluster.update(str(updated_config_file), force_update="true")
+    logging.info("Verifying the HeadNode is not replaced after cluster update")
+    assert_that(head_node_instance_id).is_equal_to(cluster.head_node_instance_id)
+
+
+@pytest.mark.usefixtures("os", "instance")
+def test_overwrite_sg(region, scheduler, custom_security_groups, pcluster_config_reader, clusters_factory):
+    """Test vpc_security_group_id overwrites pcluster default sg on head and compute nodes, efs, fsx"""
+    custom_security_group_id = custom_security_groups()[0]
+    fsx_name, efs_name = "fsx", "efs"
+    cluster_config = pcluster_config_reader(
+        vpc_security_group_id=custom_security_group_id, fsx_name=fsx_name, efs_name=efs_name
+    )
+    cluster = clusters_factory(cluster_config)
     ec2_client = boto3.client("ec2", region_name=region)
-    if not assign_additional_security_groups:
-        instances = _get_instances_by_security_group(ec2_client, default_security_group_id)
-        logging.info("Asserting that head node and compute node has and only has the custom security group")
-        assert_that(instances).is_length(3 if scheduler == "slurm" else 2)
-        for instance in instances:
-            assert_that(instance["SecurityGroups"]).is_length(1)
+    instances = _get_instances_by_security_group(ec2_client, custom_security_group_id)
+    logging.info("Asserting that head node and compute node has and only has the custom security group")
+    assert_that(instances).is_length(3 if scheduler == "slurm" else 2)
+    for instance in instances:
+        assert_that(instance["SecurityGroups"]).is_length(1)
 
     # FSx is not supported in US isolated regions or when using AWS Batch as a scheduler
     if "us-iso" not in region and scheduler != "awsbatch":
@@ -102,7 +114,7 @@ def test_overwrite_sg(
         _assert_security_group_rules(
             ec2_client,
             fsx_security_group_id,
-            [fsx_security_group_id, default_security_group_id],
+            [fsx_security_group_id, custom_security_group_id],
         )
 
     logging.info("Collecting security groups of the EFS")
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -45,21 +45,37 @@ def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_conf
...
+def test_overwrite_sg(
+    region, scheduler, custom_security_group, pcluster_config_reader, clusters_factory, additional_security_groups
+):
     """Test vpc_security_group_id overwrites pcluster default sg on head and compute nodes, efs, fsx"""
-    custom_security_group_id = custom_security_group.cfn_resources["SecurityGroupResource"]
+    count_of_sgs = 2
+
+    custom_security_group_stack = custom_security_group(count_of_sgs=count_of_sgs)
+    custom_security_group_ids = []
+    for sg_indx in range(count_of_sgs):
```
enrico-usai: minor: let's use "sg_index" rather than "indx", it's just an extra character.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -160,56 +187,65 @@ def _check_connections_between_head_node_and_compute_nodes(cluster):
...
+
+    def _custom_security_group(count_of_sgs=1):
+        template = Template()
+        template.set_version("2010-09-09")
+        template.set_description("custom security group stack for testing additional_sg and vpc_security_group_id")
+        vpc_id = vpc_stack.cfn_outputs["VpcId"]
+        for sg_indx in range(count_of_sgs):
+            security_group = template.add_resource(
+                SecurityGroup(
+                    f"SecurityGroupResource{sg_indx}",
```
enrico-usai: Minor: Not added by your patch, but I'd remove the word "Resource" from the name. Everything we're creating is a resource.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -160,56 +187,65 @@ def _check_connections_between_head_node_and_compute_nodes(cluster):
...
-        vpc_cidr = cidr_block_association["CidrBlock"]
-        template.add_resource(
-            SecurityGroupIngress(
-                f"SecurityGroupIngressResource{index}",
-                IpProtocol="-1",
-                FromPort=0,
-                ToPort=65535,
-                CidrIp=vpc_cidr,
-                GroupId=Ref(security_group),
+def custom_security_group(vpc_stack, region, request):
```
enrico-usai: minor: should be renamed to "custom_security_groups" because now is able to create more than 1 security group.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -160,56 +187,65 @@ def _check_connections_between_head_node_and_compute_nodes(cluster):
...
+    stacks_factory = CfnStacksFactory(request.config.getoption("credential"))
+
+    def _custom_security_group(count_of_sgs=1):
```
enrico-usai: I'd rename the count_of_sgs to number_of_sgs, because it's not counting something, it's instead the number of SGs we're asking to create.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -45,21 +45,37 @@ def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_conf
...
 @pytest.mark.usefixtures("os", "instance")
-def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_reader, clusters_factory):
+def test_overwrite_sg(
+    region, scheduler, custom_security_group, pcluster_config_reader, clusters_factory, additional_security_groups
+):
     """Test vpc_security_group_id overwrites pcluster default sg on head and compute nodes, efs, fsx"""
-    custom_security_group_id = custom_security_group.cfn_resources["SecurityGroupResource"]
+    count_of_sgs = 2
+
+    custom_security_group_stack = custom_security_group(count_of_sgs=count_of_sgs)
```
enrico-usai: there is something strange on the fact that `custom_security_group` is returning a stack. 
I'd say this should return the list of security group ids instead, so we should move the logic to retrieve the ids into the custom_security_groups function.

In this way we'll encapsulate the management of the index into that function, and the caller doesn't need to know that the name of the resource is something in the form `"SecurityGroupResource{sg_indx}"`

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -45,21 +45,37 @@ def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_conf
...
+    custom_security_group_ids = []
+    for sg_indx in range(count_of_sgs):
+        custom_security_group_ids.append(custom_security_group_stack.cfn_resources[f"SecurityGroupResource{sg_indx}"])
+
     fsx_name, efs_name = "fsx", "efs"
     cluster_config = pcluster_config_reader(
-        vpc_security_group_id=custom_security_group_id, fsx_name=fsx_name, efs_name=efs_name
+        vpc_security_group_id=custom_security_group_ids[0],
```
enrico-usai: it's better to create a variable called for example `default_security_group = custom_security_group_ids[0]` and use it everywhere rather than using `custom_security_group_ids[0]` here and there.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -45,21 +45,37 @@ def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_conf
...
+        fsx_name=fsx_name,
+        efs_name=efs_name,
+        additional_security_groups=additional_security_groups,
```
enrico-usai: This variable should be `assign_additional_security_groups`.

the boolean parameters should be "actions" in this way it's more clear the meaning of True/False.
Instead using `additional_security_groups` it seems the parameter is accepting a list of security groups.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -106,19 +122,30 @@ def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_
...
             mount_target_security_group_id,
-            [mount_target_security_group_id, custom_security_group_id],
+            [mount_target_security_group_id, custom_security_group_ids[0]],
         )
-
     if scheduler == "slurm":
         logging.info("Checking SSH connection between cluster nodes before cluster update")
         _check_connections_between_head_node_and_compute_nodes(cluster)
+
+        # Update the Security Groups of HN to check if the HN instance is replaced.
```
enrico-usai: let's use HeadNode rather than HN. AFAIK the HN acronym is not used in other places and Head Node is more self-explicative.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -45,21 +45,37 @@ def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_conf
...
-    logging.info("Asserting that head node and compute node has and only has the custom security group")
-    assert_that(instances).is_length(3 if scheduler == "slurm" else 2)
-    for instance in instances:
-        assert_that(instance["SecurityGroups"]).is_length(1)
+    if not additional_security_groups:
+        instances = _get_instances_by_security_group(ec2_client, custom_security_group_ids[0])
+        logging.info("Asserting that head node and compute node has and only has the custom security group")
+        assert_that(instances).is_length(3 if scheduler == "slurm" else 2)
+        for instance in instances:
+            assert_that(instance["SecurityGroups"]).is_length(1)
```
enrico-usai: you could avoid the `if not additional_security_groups` by using `assert_that(instance["SecurityGroups"]).is_length(number_of_sgs)` in the check.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -106,19 +122,30 @@ def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_
...
+
+        # Update the Security Groups of HN to check if the HN instance is replaced.
+        updated_config_file = pcluster_config_reader(
+            config_file="pcluster.config.update_sg.yaml",
+            fsx_name=fsx_name,
+            efs_name=efs_name,
+            vpc_security_group_id=custom_security_group_ids,
+            additional_security_groups=additional_security_groups,
+        )
+        cluster.update(str(updated_config_file), force_update="true")
```
enrico-usai: do we really need an extra update step? every update will take 15/30mins, I'd unify the changes in the `pcluster.config.update_sg.yaml` in the already existing `pcluster.config.update.yaml` file.

We can perform the update just one time and do all the checks we need.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -106,19 +122,30 @@ def test_overwrite_sg(region, scheduler, custom_security_group, pcluster_config_
...
+        logging.info("Checking if the HeadNode is replaced after cluster update")
```
enrico-usai: the message should be: `Verifying the HeadNode is not replaced after cluster update` otherwise it seems we're expecting it has to be replaced.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -160,56 +187,65 @@ def _check_connections_between_head_node_and_compute_nodes(cluster):
...
+        stacks_factory.create_stack(stack)
+
+        return stack
 
-    yield stack
+    yield _custom_security_group
 
     if not request.config.getoption("no_delete"):
-        cfn_stacks_factory.delete_stack(stack.name, region)
+        stacks_factory.delete_all_stacks()
```
enrico-usai, line 254: do we really need a stack for each security group?
It's better to have the security group loop into the same CfnStack, they are logically related since used by the same tests and in this way it's easier to cleanup in case of failures/issues.

tests/integration-tests/tests/networking/test_security_groups/test_overwrite_sg/pcluster.config.update_sg.yaml
```diff
@@ -0,0 +1,49 @@
+Image:
+  Os: {{ os }}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_ids[0] }}
+    {% if additional_security_groups == False %}
+    SecurityGroups:
+      - {{ vpc_security_group_id[1] }}
+    {% else %}
```
enrico-usai: `SecurityGroups` and `AdditionalSecurityGroups` can be used together. 

So  if I'm not wrong we can have:
```
pcluster.config.yaml

  Networking:
        SecurityGroups:
          - {{ vpc_security_group_id[0] }}
```


```
pcluster.config.update.yaml


  Networking:
        SecurityGroups:
          - {{ vpc_security_group_id[1] }}
        AdditionalSecurityGroups:
          - {{ vpc_security_group_id[0] }}
```

In this way you can test at the same time what happens updating Security or AdditionalSecurityGroups.
feel free to find better alternatives anyway the goal is to avoid to execute 2 times the update command.

tests/integration-tests/tests/networking/test_security_groups.py
himani2411: I followed the same modularity which was being followed earlier w.r.t `custom_security_group`. But then after changing the nos of outputs that it gives, it makes sense to return a list rather than a stack. Will make the change as you suggest

tests/integration-tests/tests/networking/test_security_groups.py
himani2411: Initially I was creating two stacks which I thought was excessive! 
But I later changed to creating one stack with multiple Security groups and missed the cleanup function here.
So we are creating a single stack but with multiple securityGroups.

tests/integration-tests/tests/networking/test_security_groups/test_overwrite_sg/pcluster.config.update_sg.yaml
himani2411: I tried that initially. But apparently they both should not [co-exist](https://github.com/aws/aws-parallelcluster/blob/0b79cf32a1bc01d19a4295e8592a8b9cd4ae9dc7/cli/src/pcluster/schemas/cluster_schema.py#L653-L654)
https://github.com/aws/aws-parallelcluster/blob/0b79cf32a1bc01d19a4295e8592a8b9cd4ae9dc7/cli/src/pcluster/schemas/cluster_schema.py#L653-L654

tests/integration-tests/tests/networking/test_security_groups.py
himani2411: The reason for creating another yaml file itself, was because one of the checks in the `pcluster.config.update.yaml` is that the HeadNodes' managed sg allows access to Complte fleets custom SG

tests/integration-tests/tests/networking/test_security_groups.py
himani2411: I dont want to call the `_get_instances_by_security_group` if not necessary which is why I am using the `not additional_security_groups` in the check

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -160,56 +190,66 @@
...
-        )
-    )
-    stack = CfnStack(
-        name=generate_stack_name("integ-tests-custom-sg", request.config.getoption("stackname_suffix")),
-        region=region,
-        template=template.to_json(),
-    )
-    cfn_stacks_factory.create_stack(stack)
+        stacks_factory.create_stack(stack)
+        sg_stack_name = stack.name
```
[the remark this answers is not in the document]
himani2411: Working on this!

tests/integration-tests/tests/networking/test_security_groups.py
himani2411: Will make changes in the `test_additional_sg_and_ssh_from` tests for updating the cluster with a new sg_id for AdditionalSecurityGroups and SecurityGroups. However, this will not reduce the time for the tests a I would still need to update the cluster.

tests/integration-tests/tests/networking/test_security_groups.py
```diff
@@ -14,21 +14,21 @@
...
 @pytest.mark.usefixtures("os", "scheduler", "instance")
-def test_additional_sg_and_ssh_from(region, custom_security_group, pcluster_config_reader, clusters_factory):
+def test_additional_sg_and_ssh_from(region, custom_security_groups, pcluster_config_reader, clusters_factory):
     """
     Test when additional_sg ssh_from are provided in the config file
 
     The additional security group should be added to the head and compute nodes. The
     """
-    custom_security_group_id = custom_security_group.cfn_resources["SecurityGroupResource"]
+    custom_security_group_id = custom_security_groups
```
enrico-usai: `custom_security_groups` is always returning a list now, so this should be:
```python
custom_security_group_id = custom_security_groups[0]
```

himani2411: I just made the change!

enrico-usai (dismissed):

hanwen-pcluste (approved):

PR #5599, opened 2023-08-09
[develop] Add and test encryption of the LoginNodes root volume

hehe7318:
### Description of changes
* Retrieve the head_node.local_storage object and create a LocalStorage for the LoginNodesPool with the same values  but ensuring that encrypted=true
* LoginNodes Stack CFN LT block_device_mappings setup

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2641,27 +2641,20 @@ class SlurmClusterConfig(BaseClusterConfig):
         self.scheduling = scheduling
         self.login_nodes = login_nodes
         if self.login_nodes:
-            # create a LocalStorage for the LoginNodesPool with the same values as HeadNode LocalStorage.
-            # but ensuring that encrypted = true
+            # Create a LocalStorage for the LoginNodesPool and ensure that encrypted = true
             for pool in self.login_nodes.pools:
-                head_node_root_volume = self.head_node.local_storage.root_volume
                 pool.local_storage = LocalStorage(
                     implied=True,
                     root_volume=RootVolume(
-                        size=head_node_root_volume.size,
-                        delete_on_termination=head_node_root_volume.delete_on_termination,
                         encrypted=True,
-                        volume_type=head_node_root_volume.volume_type,
-                        iops=head_node_root_volume.iops,
-                        throughput=head_node_root_volume.throughput,
                     ),
                 )
-        if self.login_nodes:
-            for pool in self.login_nodes.pools:
+                # If there's no customer ssh key for LoginNodes pool, set a default value as HeadNode's ssh key
                 if pool.ssh and not pool.ssh.key_name:
                     pool.ssh.key_name = self.head_node.ssh.key_name
                 elif not pool.ssh:
                     pool.ssh = LoginNodesSsh(key_name=self.head_node.ssh.key_name)
+
         self.__image_dict = None
         # Cache capacity reservations information together to reduce number of boto3 calls.
         # Since this cache is only used for validation, if AWSClientError happens
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2641,12 +2641,7 @@ class SlurmClusterConfig(BaseClusterConfig):
         if self.login_nodes:
             # Create a LocalStorage for the LoginNodesPool and ensure that encrypted = true
             for pool in self.login_nodes.pools:
-                pool.local_storage = LocalStorage(
-                    implied=True,
-                    root_volume=RootVolume(
-                        encrypted=True,
-                    ),
-                )
+                pool.local_storage = LocalStorage(implied=True)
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -2640,6 +2640,22 @@ def __init__(
...
+                        size=head_node_root_volume.size,
+                        delete_on_termination=head_node_root_volume.delete_on_termination,
+                        encrypted=True,
+                        volume_type=head_node_root_volume.volume_type,
+                        iops=head_node_root_volume.iops,
+                        throughput=head_node_root_volume.throughput,
+                    ),
+                )
         if self.login_nodes:
             for pool in self.login_nodes.pools:
```
EddyMM, line 2643: Instead of looping through the pools twice, could we use this existing loop?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2640,6 +2640,22 @@ def __init__(
         super().__init__(cluster_name, **kwargs)
         self.scheduling = scheduling
         self.login_nodes = login_nodes
+        if self.login_nodes:
+            # create a LocalStorage for the LoginNodesPool with the same values as HeadNode LocalStorage.
+            # but ensuring that encrypted = true
+            for pool in self.login_nodes.pools:
+                head_node_root_volume = self.head_node.local_storage.root_volume
```
EddyMM: The head node's local storage configuration can be obtained outside of the loop.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -109,6 +111,10 @@ def _add_login_nodes_pool_launch_template(self):
             f"LoginNodeLaunchTemplate{self._pool.name}",
             launch_template_name=f"{self.stack_name}-{self._pool.name}",
             launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
+                block_device_mappings=self._launch_template_builder.get_block_device_mappings(
```
EddyMM, line 114: Let's add tests to confirm the expected configuration are applied to the template.

cli/src/pcluster/config/cluster_config.py
hehe7318: Thank you for reminding Eddy! Sure we can. I used 2 loops just for better readability. I can modify it to use one.

cli/src/pcluster/config/cluster_config.py
hehe7318: Makes sense! Thank you Eddy.

cli/src/pcluster/templates/login_nodes_stack.py
hehe7318: Okay!

hehe7318: Done! Just found there were some incorrect code in previous unit test. Have corrected them. All the modification also done for release-3.7 branch. BTW, thank you for helping me correct the PR title.

cli/src/pcluster/config/cluster_config.py
hehe7318: Removed this line of code for now. Only set `encrypted=True`.

cli/src/pcluster/config/cluster_config.py
hehe7318: Done!

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2639,11 +2639,20 @@ def __init__(
         self.scheduling = scheduling
         self.login_nodes = login_nodes
         if self.login_nodes:
+            # Create a LocalStorage for the LoginNodesPool and ensure that encrypted = true
             for pool in self.login_nodes.pools:
+                pool.local_storage = LocalStorage(
+                    implied=True,
+                    root_volume=RootVolume(
+                        encrypted=True,
```
EddyMM: The default is already `True`. We can use the default value for uniformity and avoid having to check what is set for each type of node.

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -370,32 +370,28 @@ def test_compute_launch_template_properties(
...
+        iam_instance_profile_name,
     ):
-        self.pool_name = pool_name
-        self.instance_type = instance_type
-        self.count = count
         self.subnet_ids = subnet_ids
-        self.key_name = key_name
-        self.security_groups = security_groups
-        self.gracetime_period = gracetime_period
+        self.root_volume = root_volume
```
EddyMM: This test is meant to confirm that the login node's root volume is encrpyted by default. Should we be passing the settings explicitly? Or can we simply go ahead and assert that the Ebs volume has `"Encrypted": True` in the launch template?

hehe7318: Done!

EddyMM (approved):

PR #5619, opened 2023-08-11
[develop] Add AD test for LoginNodes

NSsirena:
### Description of changes
Test AD integration in LoginNodes with SimpleAD and MicrosoftAD with certification validation.

Checks that AD users can log into the nodes, their home and ssh keys created and their posix permissions correctly set.

Note: I added a dedicated test (duplicating part of the configuration) because the existing tests does many subtle changes and rely on files and folders being in specific places (like the `shared` folder) that aren't always available on the LoginNodes. So I just extracted in a dedicated test the portion that was relevant for the LoginNodes

### Tests
* Tests was successfully executed in my personal account.

### References
* N/A
* 
### Checklist

### DIFF
--- a/tests/integration-tests/tests/ad_integration/test_ad_integration.py
+++ b/tests/integration-tests/tests/ad_integration/test_ad_integration.py
@@ -837,3 +837,107 @@ def _check_ssh_auth(user, expect_success=True):
             raise e
         else:
             logging.info(f"SSH access denied for user {user.alias}, as expected")
+
+
+@pytest.mark.parametrize(
+    "directory_type,directory_protocol,directory_certificate_verification",
+    [
+        ("SimpleAD", "ldap", False),
+        ("MicrosoftAD", "ldaps", True),
+    ],
+)
+@pytest.mark.usefixtures("os", "instance")
+def test_ad_integration_on_login_nodes(
+    region,
+    scheduler,
+    scheduler_commands_factory,
+    pcluster_config_reader,
+    directory_type,
+    directory_protocol,
+    directory_certificate_verification,
+    test_datadir,
+    directory_factory,
+    request,
+    store_secret_in_secret_manager,
+    clusters_factory,
+):
+    """
+    Verify AD integration works as expected.
+    In particular, it verifies that :
+    1. AD users can access to the login node, both with password and SSH key (if created);
+    2. SSH key for AD users is created when the property GenerateSshKeysForUsers is true;
+    3. AD users can submit workloads;
+    """
+    head_node_instance_type = "c5n.18xlarge" if request.config.getoption("benchmarks") else "c5.xlarge"
+    compute_instance_type_info = {"name": "c5.xlarge", "num_cores": 4}
+    config_params = {
+        "compute_instance_type": compute_instance_type_info.get("name"),
+        "head_node_instance_type": head_node_instance_type,
+    }
+    directory_stack_name, nlb_stack_name = directory_factory(
+        request.config.getoption("directory_stack_name"),
+        request.config.getoption("ldaps_nlb_stack_name"),
+        directory_type,
+        str(test_datadir),
+        region,
+    )
+    directory_stack_outputs = get_infra_stack_outputs(directory_stack_name)
+    ad_user_password = directory_stack_outputs.get("UserPassword")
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -837,3 +837,108 @@ def _check_ssh_auth(user, expect_success=True):
             raise e
         else:
             logging.info(f"SSH access denied for user {user.alias}, as expected")
+
+
+@pytest.mark.parametrize(
+    "directory_type,directory_protocol,directory_certificate_verification",
+    [
+        ("SimpleAD", "ldap", False),
+        ("MicrosoftAD", "ldaps", True),
```
gmarciani, line 846: I suggest to keep only SimpleAD because the integration between login nodes and AD is equivalent to the one in the head node, which is already covered by both SimpleAd and MsAD, so for login nodes should be enoiugh to test just one. I suggest to use SimpleAd because it takes less time to be deployed.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -837,3 +837,108 @@ def _check_ssh_auth(user, expect_success=True):
...
+    test_datadir,
+    directory_factory,
+    request,
+    store_secret_in_secret_manager,
+    clusters_factory,
+):
+    """
+    Verify AD integration works as expected.
+    In particular, it verifies that :
+    1. AD users can access to the head node, both with password and SSH key (if created);
```
gmarciani: can access to the login nodes, not head node

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -837,3 +837,108 @@ def _check_ssh_auth(user, expect_success=True):
...
+    cluster_config = pcluster_config_reader(**config_params)
+    cluster = clusters_factory(cluster_config)
+
+    certificate_secret_arn = nlb_stack_parameters.get("CertificateSecretArn")
+    certificate = boto3.client("secretsmanager").get_secret_value(SecretId=certificate_secret_arn)["SecretBinary"]
+    with open(test_datadir / "certificate.crt", "wb") as f:
+        f.write(certificate)
+
+    # Publish compute node count metric every minute via cron job
+    # TODO: use metrics reporter from the benchmarks module
```
gmarciani: We can remove this TODO because we are not going to use the benchmark module anymore in this kind of tests. We should remove the equivalent comment also in the original integ test

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration_on_login_nodes/NLB_SimpleAD.yaml
```diff
@@ -0,0 +1,118 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
gmarciani, line 1: What about reusing the stack defined for the original test rather than creating a new one?
Same for the other stack.

gmarciani (changes requested):

tests/integration-tests/tests/ad_integration/test_ad_integration.py
NSsirena: I agree, but I'd keep it.
The long term goal is to move all tests related to regular-users behavior and UX on the LoginNodes.
So this is a step in that direction (and long term we are going to remove the test on the HeadNode and enrich this one).

tests/integration-tests/tests/ad_integration/test_ad_integration.py
NSsirena: Good catch!
Fixing it..

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration_on_login_nodes/NLB_SimpleAD.yaml
NSsirena: I'll make a quick attempt to adapt the fixture ..but if it requires too much time I'll leave this to a future improvement.

gmarciani (approved):

PR #5632, opened 2023-08-15
[Release 3.7] Revert back to curl for test dcv, add a check for an ex…

dreambeyondorange:
…ception that gets thrown in the test

When changing to wget, the command was hanging on rhel8 and returning a error code This change reverts back to curl which exits cleanly and returns an error code only on Ubuntu 22.  This change allows the test to resolve properly whether an error code is returned or not.

### Tests
* Integ tests locally for Ubuntu 22 and rhel8

### Checklist


hanwen-pcluste (approved):
I would be better if you could reference the original PR

PR #5639, opened 2023-08-16
[develop]Test connectivity and basic Slurm interactions with a LN Pool in a public subnet 

hehe7318:
### Description of changes
* adapt `test_slurm` and `test_pmix` existing tests use `use_login_node=true`

### Checklist


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -80,10 +90,16 @@ def test_slurm(
...
     cluster = clusters_factory(cluster_config, upper_case_cluster_name=True)
-    remote_command_executor = RemoteCommandExecutor(cluster)
-    clustermgtd_conf_path = _retrieve_clustermgtd_conf_path(remote_command_executor)
+    remote_command_executor = RemoteCommandExecutor(cluster, use_login_node=use_login_node)
+    remote_command_executor_for_head_node = None
+    if use_login_node:
+        remote_command_executor_for_head_node = RemoteCommandExecutor(cluster)
+    clustermgtd_conf_path = _retrieve_clustermgtd_conf_path(
+        remote_command_executor if not use_login_node else remote_command_executor_for_head_node
+    )
```
NSsirena: Since 
 * `clustermgtd_conf_path` is only used in `_test_compute_node_bootstrap_timeout(` below in the code 
 
and 
* `_test_compute_node_bootstrap_timeout(`
*  `assert_no_errors_in_logs(` 

must be always run in the HeadNode I propose to move the logic that must be run in the head node below and keep it together. Also to highlight the fact that those tests MUST be run on the head node I'd create a dedicated command executor

```
head_node_command_executor = RemoteCommandExecutor(cluster)
...
 assert_no_errors_in_logs(head_node_command_executor)
 ...
_test_compute_node_bootstrap_timeout(..., head_node_command_executor)
 ```

while in the first part of the test always use a variable called `remote_command_executor` (that may be initialized differently depending on the `use_login_node` parameter) to make it more clear that this section can be run without changes both in a LoginNode or in the HeadNode.

File: tests/integration-tests/tests/schedulers/test_slurm/test_slurm_pmix/pcluster.config.yaml [hunk 2]
```diff
@@ -1,5 +1,16 @@
...
+{% if scheduler == "slurm" and use_login_node == True %}
+LoginNodes:
+  Pools:
+    - Name: testpool
+      InstanceType: {{ instance }}
+      Count: 2
+      Networking:
+        SubnetIds:
+          - {{ public_subnet_id }}
+      GracetimePeriod: 60
```
NSsirena: Do we need `GracetimePeriod: 60` for the test or as in the others config we can set it at `GracetimePeriod: 5` ?

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Makes sense! Thank you Nicola!

tests/integration-tests/tests/schedulers/test_slurm/test_slurm_pmix/pcluster.config.yaml
hehe7318: Yes, we can set it to 5. I will do it.

NSsirena (approved):
It seems that in  `_test_compute_node_bootstrap_timeout(` we are using `remote_command_executor` only to read the slurm config.

It's probably not worth to have to remote executor just because we can reach a file also from a login_node.

We can simplify this and assume the whole tests should run from the HeadNode.
However since this is a nitpick I'll approve the PR and we can improve this in a different PR.

PR #5642, opened 2023-08-17
[Develop] Test Slurm with LoginNodes in a private network

NSsirena:
### Description of changes
* Cluster class provides private IP
* RemoteCommandExecutor retrieves and uses  LoginNode private IP when bastion is defined
* Test common Slurm tasks with LoginNodes in a private network

### Tests
* Manual launch of the test was successful

### References
N/A

### Checklist

### DIFF
--- a/tests/integration-tests/remote_command_executor.py
+++ b/tests/integration-tests/remote_command_executor.py
@@ -45,9 +45,12 @@ class RemoteCommandExecutor:
             bastion = f"{username}@{cluster.head_node_ip}"
         elif use_login_node:
             self.target = "LoginNode"
-            node_ip = cluster.get_login_node_public_ip()
+            if bastion:
+                node_ip = cluster.get_login_node_private_ip()
+            else:
+                node_ip = cluster.get_login_node_public_ip()
             if node_ip is None:
-                raise RemoteCommandExecutionError("No healthy LoginNode found in the cluster.")
+                raise RemoteCommandExecutionError("Unable to retrieve a valid LoginNode IP Address.")
         else:
             self.target = "HeadNode"
             node_ip = cluster.head_node_ip
@@ -68,6 +71,7 @@ class RemoteCommandExecutor:
             )
             connection_kwargs["gateway"] = f"ssh -W %h:%p -A {bastion}"
             connection_kwargs["forward_agent"] = True
+            connection_kwargs["connect_kwargs"]["banner_timeout"] = 60
         logging.info(
             f"Connecting to {connection_kwargs['host']} as {connection_kwargs['user']} with "
             f"{connection_kwargs['connect_kwargs']['key_filename']}"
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -189,6 +189,72 @@ def test_slurm_ticket_17399(
...
+    pcluster_config_reader,
+    clusters_factory,
+    test_datadir,
+    architecture,
+    scheduler_commands_factory,
+    vpc_stack,
+):
+    """Test Slurm features in login nodes inside a private network."""
+
+    scaledown_idletime = 3
```
EddyMM, line 204: I see that test_slurm uses the same logic: [test_slurm](https://github.com/NSsirena/aws-parallelcluster/blob/e90af4b9c0680ba4ae45d328e258099bc5b1d59a/tests/integration-tests/tests/schedulers/test_slurm.py#L80-L130). The config files are also quite similar

Can we parametrize the test_slurm to include a scheme parameter?

This way we can ensure that the Slurm specific test is always done in both networking schemes. Especially if the Slurm test changes.

tests/integration-tests/remote_command_executor.py
```diff
@@ -59,6 +62,7 @@ def __init__(
             "connect_kwargs": {
                 "key_filename": [alternate_ssh_key if alternate_ssh_key else cluster.ssh_key],
                 "look_for_keys": False,
+                "banner_timeout": 60,
```
EddyMM: Does this timeout translate to time between attempts to check the SSH Banner?

In case it does, can we enable this option only when using a bastion? Just to avoid potentially longer wait times for remote command connections that can directly access the head node or login node.

tests/integration-tests/tests/schedulers/test_slurm.py
NSsirena: I thought about this and didn't do it for two reasons:
* I wanted to avoid a combinatorial growth of conditions and test cases 
* I wanted to avoid making our tests more and more complex and hard to decode

tests/integration-tests/remote_command_executor.py
NSsirena: Not sure if it does, 
but I agree it is better to add it only when the bastion is used.

tests/integration-tests/tests/schedulers/test_slurm.py
EddyMM: This might limit how the Slurm tests evolve since it won't be easy to remember updating them in multiple code blocks. I however agree that we should strike a balance on the combination of parameters used in each test. Since this test simply attempts to cover core functionality of using Slurm it should be fine (even if the other test might evolve).

tests/integration-tests/remote_command_executor.py
NSsirena: fixed

tests/integration-tests/tests/schedulers/test_slurm.py
NSsirena: We should (and are planning to) achieve the same result (test code reusability) with a different approach that doesn't require make each existing test more and more convoluted and hard to understand.

EddyMM (approved):

PR #5646, opened 2023-08-17
[release-3.7]Test connectivity and basic Slurm interactions with a LN Pool in a public subnet 

hehe7318:
### Description of changes
* adapt `test_slurm` and `test_pmix` existing tests use `use_login_node=true`

### Checklist


NSsirena (approved):
It seems that in `_test_compute_node_bootstrap_timeout` we are using `remote_command_executor` only to read the slurm config.

It's probably not worth to have to remote executor just because we can reach a file also from a login_node.

We can simplify this and assume the whole tests should run from the HeadNode.
However since this is a nitpick I'll approve the PR and we can improve this in a different PR.

PR #5662, opened 2023-08-21
Surface update failures to pcluster describe cluster output

ohchocolate:
### Description of changes
* Add failure codes and failure reasons for cluster update failures
* Refactor related controller to handle both the creation and update failures

### Tests
* Manual tests
* Add unit test

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/api/controllers/cluster_operations_controller.py
+++ b/cli/src/pcluster/api/controllers/cluster_operations_controller.py
@@ -239,7 +239,8 @@ def describe_cluster(cluster_name, region=None):
         region=os.environ.get("AWS_DEFAULT_REGION"),
         cluster_status=cluster_status,
         scheduler=Scheduler(type=cluster.stack.scheduler),
-        failures=_get_creation_failures(cluster_status, cfn_stack),
+        # failures=_get_creation_failures(cluster_status, cfn_stack),
+        failures=_get_failures(cluster_status, cfn_stack),
     )
 
     try:
@@ -470,9 +471,16 @@ def _create_message(failure_reason, action_needed):
     return message or "Error during update"
 
 
-def _get_creation_failures(cluster_status, cfn_stack):
-    """Get a list of Failure objects containing failure code and reason when cluster creation failed."""
-    if cluster_status != ClusterStatus.CREATE_FAILED:
+def _get_failures(cluster_status, cfn_stack):
+    """
+    Get a list of Failure objects containing failure code and reason
+    when cluster creation or update failed.
+    """
+    if cluster_status == ClusterStatus.CREATE_FAILED:
+        failure_code, failure_reason = cfn_stack.get_cluster_creation_failure()
+    elif cluster_status == ClusterStatus.UPDATE_FAILED:
+        failure_code, failure_reason = cfn_stack.get_cluster_update_failure()
+    else:
         return None
-    failure_code, failure_reason = cfn_stack.get_cluster_creation_failure()
+
     return [Failure(failure_code=failure_code, failure_reason=failure_reason)]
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/api/controllers/cluster_operations_controller.py
```diff
@@ -239,7 +239,8 @@ def describe_cluster(cluster_name, region=None):
         region=os.environ.get("AWS_DEFAULT_REGION"),
         cluster_status=cluster_status,
         scheduler=Scheduler(type=cluster.stack.scheduler),
-        failures=_get_creation_failures(cluster_status, cfn_stack),
+        # failures=_get_creation_failures(cluster_status, cfn_stack),
```
enrico-usai: should you remove this commented code?

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -172,15 +179,33 @@ class ClusterCreationFailure:
...
+        @dataclass
+        class ClusterUpdateFailure:
+            """Represent the object holding the data of ClusterCreationFailure."""
+
+            failure_code: str
+            api_failure_reason: str
+            cfn_failure_reason: str  # cloud formation matching string
+
+        cluster_update_failures = [
+            ClusterUpdateFailure("EbsMountFailure", "Failed to mount EBS volume.", "Failed to mount EBS volume"),
```
enrico-usai, line 195: Question: why we don't have the same kind of failures at creation time and we just have the general "ClusterCreationFailure" message?

cli/src/pcluster/api/controllers/cluster_operations_controller.py
ohchocolate: Sure！

cli/src/pcluster/models/cluster_resources.py
ohchocolate: I separated the failures for modularity and future additions. I anticipated that we may add more update failures in the future. But I'm open to refactor and combine them!

cli/src/pcluster/models/cluster_resources.py
```diff
@@ -172,15 +179,33 @@ class ClusterCreationFailure:
...
+            ClusterUpdateFailure("EfsMountFailure", "Failed to mount EFS.", "Failed to mount EFS"),
+            ClusterUpdateFailure("FsxMountFailure", "Failed to mount FSX.", "Failed to mount FSX"),
+            ClusterUpdateFailure("RaidMountFailure", "Failed to mount RAID array.", "Failed to mount RAID array"),
+        ]
+        general_failure = ClusterUpdateFailure("ClusterUpdateFailure", "Failed to update the cluster.", "")
+        return cluster_update_failures, general_failure
```
chenwany, line 201: Why do we change the return here?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1422,7 +1422,8 @@ def _add_head_node(self):
                             f" cfn-signal --exit-code=0 --reason='Update complete'"
                             f" --region {self.stack.region} --url {cloudformation_url}"
                             f" '{self.wait_condition_handle.ref}' ||"
-                            f" cfn-signal --exit-code=1 --reason='Update failed'"
+                            f' cfn-signal --exit-code=1 --reason="$(cat /var/log/parallelcluster/bootstrap_error_msg '
+                            f"2>/dev/null || echo 'Update failed')\""
```
chenwany: we will need to remove this file if it exist at the end in case the next update failure doesn't write to the file

cli/src/pcluster/models/cluster_resources.py
ohchocolate: There are two type of failures and each has its own general failure message.

cli/src/pcluster/templates/cluster_stack.py
ohchocolate: Added.

lukeseawalker: have you considered to handle the cases where the content of `bootstrap_error_msg` could be problematic? 
One I can think of is when the size is big enough to exceed the limit on the allowed parameter size.

ohchocolate: Good point! I think I can put a size limit of 4096 bytes.

chenwany:
This PR is not completed yet. We need to  rename or backup the bootstrap failure file after the update and surfacing the message so the message will not be showed up in the next update

PR #5669, opened 2023-08-27
Add exports to key Outputs from the serverless-database stack

mwvaughn:
This helps enable use of the `Fn::ImportValue` intrinsic function when launching a cluster from CloudFormation template.

### Description of changes
* I have added exports to the following stack outputs: ClusterName, DatabaseHost, DatabasePort, DatabaseAdminUser, DatabaseSecretArn, and DatabaseClusterSecurityGroup. They follow the template of `${AWS::StackName}-OUTPUTNAME`

### Tests
* For a stack named *teststack100*, I have confirmed the following exports: teststack100-ClusterName, teststack100-DatabaseHost, teststack100-DatabasePort, teststack100-DatabaseAdminUser, teststack100-DatabaseSecretArn, teststack100-DatabaseClusterSecurityGroup
* I have confirmed that the stack launches a valid database stack.
* I have confirmed that these exports can be imported into another CloudFormation stack and used to configure a working instance of ParallelCluster

### References
* The tutorial [Creating a cluster with AWS CloudFormation](https://docs.aws.amazon.com/parallelcluster/latest/ug/tutorials_09_cfn-custom-resource-v3.html) is good context for understanding this PR. If a customer wanted to extend the demonstration template to configure Slurm accounting, they could add a Parameter for the accounting database stack name. They could then import values from its exports when setting up `Scheduling.SlurmSettings.Database` and `HeadNode.Networking.AdditionalSecurityGroups`. 
    * Here is an example of importing the database URI: `Uri: {"Fn::ImportValue" : {"Fn::Sub" : "${DatabaseStackNameParameter}-DatabaseHost"}}`


cloudformation/database/serverless-database.yaml
```diff
@@ -338,24 +338,36 @@ Resources:
 Outputs:
   ClusterName:
     Value: !Ref ClusterName
+    Export:
```
EddyMM: Why is the target cluster trying to import the cluster name? Is this export assuming that the DB stack will be created before the Cluster?

EddyMM (commented):
I've removed the "release-3.8" in the title, we normally add it as a quick way of knowing which branch the PR is targeting. Right now we're targeting develop only so the title can be simple.

cloudformation/database/serverless-database.yaml
mwvaughn: The assumption is definitely that the DB stack will be created before the cluster, and should operate independently of it. But, you are right that the customer does not explicitly _need_ the ClusterName to configure accounting on their ParallelCluster deployment. Shall I take it out?

EddyMM (approved):

PR #5681, opened 2023-09-01
[develop] Setup custom munge key

hehe7318:
### Description of changes
* Add a DevSetting to specify a Secret Arn for the custom munge key and related functions
* Add related IAM policy
* Add a validator for custom munge key secret Arn
  * The CLI session does not have permission to read the content of the secret - warning
  * Generic error in accessing the secret from the CLI session - warning
  * The content of the secret is not a vaild Base64 encoded text - error

### Tests
* munge key secret validator related test
* creating cluster with custom munge key test

### Relate Pull Request
* See the PR of Setup custom munge key in cookbook: 
  * [[develop] Add custom munge key config logic](https://github.com/aws/aws-parallelcluster-cookbook/pull/2451)
  * [[develop] Add custom munge key update logic](https://github.com/aws/aws-parallelcluster-cookbook/pull/2452)
  * [[develop] Add custom munge key rotation script](https://github.com/aws/aws-parallelcluster-cookbook/pull/2453)

```
DevSettings:
  SlurmSettings:
    MungeKeySecretArn: Str
```

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1192,7 +1192,7 @@ class ClusterDevSettings(BaseDevSettings):
         ami_search_filters: AmiSearchFilters = None,
         instance_types_data: str = None,
         timeouts: Timeouts = None,
-        settings: SlurmSettingsForCustomMungeKey = None,
+        slurm_settings: SlurmSettingsForCustomMungeKey = None,
         compute_startup_time_metric_enabled: bool = None,
         **kwargs,
     ):
@@ -1204,7 +1204,7 @@ class ClusterDevSettings(BaseDevSettings):
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
-        self.settings = settings or SlurmSettingsForCustomMungeKey(implied=True)
+        self.slurm_settings = slurm_settings or SlurmSettingsForCustomMungeKey(implied=True)
 
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -178,7 +178,11 @@ from pcluster.validators.s3_validators import (
     S3BucketValidator,
     UrlValidator,
 )
-from pcluster.validators.secret_validators import MungeKeySecretArnValidator
+from pcluster.validators.secret_validators import (
+    MungeKeySecretArnExistsValidator,
+    MungeKeySecretArnServiceValidator,
+    MungeKeySecretSizeAndBase64Validator,
+)
 from pcluster.validators.slurm_settings_validator import (
     SLURM_SETTINGS_DENY_LIST,
     CustomSlurmNodeNamesValidator,
@@ -1177,9 +1181,17 @@ class SlurmSettingsForCustomMungeKey(Resource):
         super()._register_validators(context)
         if self.munge_key_secret_arn:
             self._register_validator(
-                MungeKeySecretArnValidator,
+                MungeKeySecretArnExistsValidator,
+                munge_key_secret_arn=self.munge_key_secret_arn,
+            )
+            self._register_validator(
+                MungeKeySecretArnServiceValidator,
                 munge_key_secret_arn=self.munge_key_secret_arn,
                 region=get_region(),
+            )
+            self._register_validator(
+                MungeKeySecretSizeAndBase64Validator,
+                munge_key_secret_arn=self.munge_key_secret_arn,
             )
 
 
@@ -1192,7 +1204,7 @@ class ClusterDevSettings(BaseDevSettings):
         ami_search_filters: AmiSearchFilters = None,
         instance_types_data: str = None,
         timeouts: Timeouts = None,
-        slurm_settings: SlurmSettingsForCustomMungeKey = None,
+        munge_key_settings: SlurmSettingsForCustomMungeKey = None,
         compute_startup_time_metric_enabled: bool = None,
         **kwargs,
     ):
@@ -1204,7 +1216,7 @@ class ClusterDevSettings(BaseDevSettings):
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
-        self.slurm_settings = slurm_settings or SlurmSettingsForCustomMungeKey(implied=True)
+        self.munge_key_settings = munge_key_settings or SlurmSettingsForCustomMungeKey(implied=True)
 
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -178,11 +178,7 @@ from pcluster.validators.s3_validators import (
     S3BucketValidator,
     UrlValidator,
 )
-from pcluster.validators.secret_validators import (
-    MungeKeySecretArnExistsValidator,
-    MungeKeySecretArnServiceValidator,
-    MungeKeySecretSizeAndBase64Validator,
-)
+from pcluster.validators.secret_validators import ArnServiceAndResourceValidator, MungeKeySecretSizeAndBase64Validator
 from pcluster.validators.slurm_settings_validator import (
     SLURM_SETTINGS_DENY_LIST,
     CustomSlurmNodeNamesValidator,
@@ -1181,13 +1177,11 @@ class SlurmSettingsForCustomMungeKey(Resource):
         super()._register_validators(context)
         if self.munge_key_secret_arn:
             self._register_validator(
-                MungeKeySecretArnExistsValidator,
-                munge_key_secret_arn=self.munge_key_secret_arn,
-            )
-            self._register_validator(
-                MungeKeySecretArnServiceValidator,
-                munge_key_secret_arn=self.munge_key_secret_arn,
+                ArnServiceAndResourceValidator,
+                arn=self.munge_key_secret_arn,
                 region=get_region(),
+                expected_service="secretsmanager",
+                expected_resource="secret",
             )
             self._register_validator(
                 MungeKeySecretSizeAndBase64Validator,
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1204,7 +1204,7 @@ class ClusterDevSettings(BaseDevSettings):
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
-        self.slurm_settings = slurm_settings or SlurmSettingsForCustomMungeKey(implied=True)
+        self.slurm_settings = slurm_settings
 
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -178,6 +178,7 @@ from pcluster.validators.s3_validators import (
     S3BucketValidator,
     UrlValidator,
 )
+from pcluster.validators.secret_validators import MungeKeySecretArnValidator
 from pcluster.validators.slurm_settings_validator import (
     SLURM_SETTINGS_DENY_LIST,
     CustomSlurmNodeNamesValidator,
@@ -1161,6 +1162,27 @@ class CapacityReservationTarget(Resource):
         self.capacity_reservation_resource_group_arn = Resource.init_param(capacity_reservation_resource_group_arn)
 
 
+class SlurmSettingsForCustomMungeKey(Resource):
+    """Represent the slurm settings for custom munge key test in dev settings."""
+
+    def __init__(
+        self,
+        munge_key_secret_arn: str = None,
+        **kwargs,
+    ):
+        super().__init__()
+        self.munge_key_secret_arn = Resource.init_param(munge_key_secret_arn)
+
+    def _register_validators(self, context: ValidatorContext = None):
+        super()._register_validators(context)
+        if self.munge_key_secret_arn:
+            self._register_validator(
+                MungeKeySecretArnValidator,
+                munge_key_secret_arn=self.munge_key_secret_arn,
+                region=get_region(),
+            )
+
+
 class ClusterDevSettings(BaseDevSettings):
     """Represent the dev settings configuration."""
 
@@ -1170,6 +1192,7 @@ class ClusterDevSettings(BaseDevSettings):
         ami_search_filters: AmiSearchFilters = None,
         instance_types_data: str = None,
         timeouts: Timeouts = None,
+        settings: SlurmSettingsForCustomMungeKey = None,
         compute_startup_time_metric_enabled: bool = None,
         **kwargs,
     ):
@@ -1181,6 +1204,7 @@ class ClusterDevSettings(BaseDevSettings):
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
+        self.settings = settings or SlurmSettingsForCustomMungeKey(implied=True)
 
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1204,7 +1204,7 @@ class ClusterDevSettings(BaseDevSettings):
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
-        self.slurm_settings = slurm_settings
+        self.slurm_settings = slurm_settings or SlurmSettingsForCustomMungeKey(implied=True)
 
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -91,6 +91,7 @@ from pcluster.config.cluster_config import (
     SlurmQueueNetworking,
     SlurmScheduling,
     SlurmSettings,
+    SlurmSettingsForCustomMungeKey,
     Timeouts,
 )
 from pcluster.config.common import BaseTag
@@ -1110,7 +1111,7 @@ class SlurmSettingsForCustomMungeKeySchema(BaseSchema):
     @post_load
     def make_resource(self, data, **kwargs):
         """Generate resource."""
-        return SlurmSettingsForCustomMungeKeySchema(**data)
+        return SlurmSettingsForCustomMungeKey(**data)
 
 
 class ClusterDevSettingsSchema(BaseDevSettingsSchema):
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1121,7 +1121,9 @@ class ClusterDevSettingsSchema(BaseDevSettingsSchema):
     instance_types_data = fields.Str(metadata={"update_policy": UpdatePolicy.SUPPORTED})
     timeouts = fields.Nested(TimeoutsSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
     compute_startup_time_metric_enabled = fields.Bool(metadata={"update_policy": UpdatePolicy.SUPPORTED})
-    slurm_settings = fields.Nested(SlurmSettingsForCustomMungeKeySchema, metadata={"update_policy": UpdatePolicy.IGNORED})
+    slurm_settings = fields.Nested(
+        SlurmSettingsForCustomMungeKeySchema, metadata={"update_policy": UpdatePolicy.IGNORED}
+    )
 
     @post_load
     def make_resource(self, data, **kwargs):
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1102,6 +1102,17 @@ class CapacityReservationTargetSchema(BaseSchema):
             )
 
 
+class SlurmSettingsForCustomMungeKeySchema(BaseSchema):
+    """Represent the schema of slurm settings for custom munge key test in dev settings."""
+
+    munge_key_secre_arn = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+
+    @post_load
+    def make_resource(self, data, **kwargs):
+        """Generate resource."""
+        return SlurmSettingsForCustomMungeKeySchema(**data)
+
+
 class ClusterDevSettingsSchema(BaseDevSettingsSchema):
     """Represent the schema of Dev Setting."""
 
@@ -1110,6 +1121,7 @@ class ClusterDevSettingsSchema(BaseDevSettingsSchema):
     instance_types_data = fields.Str(metadata={"update_policy": UpdatePolicy.SUPPORTED})
     timeouts = fields.Nested(TimeoutsSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
     compute_startup_time_metric_enabled = fields.Bool(metadata={"update_policy": UpdatePolicy.SUPPORTED})
+    slurm_settings = fields.Nested(SlurmSettingsForCustomMungeKeySchema, metadata={"update_policy": UpdatePolicy.IGNORED})
 
     @post_load
     def make_resource(self, data, **kwargs):
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1106,7 +1106,7 @@ class CapacityReservationTargetSchema(BaseSchema):
 class SlurmSettingsForCustomMungeKeySchema(BaseSchema):
     """Represent the schema of slurm settings for custom munge key test in dev settings."""
 
-    munge_key_secret_arn = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    munge_key_secret_arn = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_AND_LOGIN_NODES_STOP})
 
     @post_load
     def make_resource(self, data, **kwargs):
@@ -1122,7 +1122,7 @@ class ClusterDevSettingsSchema(BaseDevSettingsSchema):
     instance_types_data = fields.Str(metadata={"update_policy": UpdatePolicy.SUPPORTED})
     timeouts = fields.Nested(TimeoutsSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
     compute_startup_time_metric_enabled = fields.Bool(metadata={"update_policy": UpdatePolicy.SUPPORTED})
-    slurm_settings = fields.Nested(
+    munge_key_settings = fields.Nested(
         SlurmSettingsForCustomMungeKeySchema, metadata={"update_policy": UpdatePolicy.IGNORED}
     )
 
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1105,7 +1105,7 @@ class CapacityReservationTargetSchema(BaseSchema):
 class SlurmSettingsForCustomMungeKeySchema(BaseSchema):
     """Represent the schema of slurm settings for custom munge key test in dev settings."""
 
-    munge_key_secre_arn = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
+    munge_key_secret_arn = fields.Str(metadata={"update_policy": UpdatePolicy.COMPUTE_FLEET_STOP})
 
     @post_load
     def make_resource(self, data, **kwargs):
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -675,7 +675,7 @@ class HeadNodeIamResources(NodeIamResourcesBase):
                         sid="SecretsManager",
                         actions=["secretsmanager:GetSecretValue", "secretsmanager:DescribeSecret"],
                         effect=iam.Effect.ALLOW,
-                        resources=[self._config.dev_settings.slurm_settings.munge_key_secret_arn]
+                        resources=[self._config.dev_settings.slurm_settings.munge_key_secret_arn],
                     ),
                 ]
             )
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -583,10 +583,7 @@ class HeadNodeIamResources(NodeIamResourcesBase):
         policy = [
             iam.PolicyStatement(
                 sid="SecretsManager",
-                actions=[
-                    "secretsmanager:GetSecretValue",
-                    "secretsmanager:DescribeSecret"
-                ],
+                actions=["secretsmanager:GetSecretValue", "secretsmanager:DescribeSecret"],
                 effect=iam.Effect.ALLOW,
                 resources=[
                     self._format_arn(
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -585,14 +585,7 @@ class HeadNodeIamResources(NodeIamResourcesBase):
                 sid="SecretsManager",
                 actions=["secretsmanager:GetSecretValue", "secretsmanager:DescribeSecret"],
                 effect=iam.Effect.ALLOW,
-                resources=[
-                    self._format_arn(
-                        service="secretsmanager",
-                        resource="*",
-                        region=Stack.of(self).region,
-                        account=Stack.of(self).account,
-                    )
-                ],
+                resources=[self._config.dev_settings.slurm_settings.munge_key_secret_arn]
             ),
             iam.PolicyStatement(
                 sid="Ec2",
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -668,7 +668,11 @@ class HeadNodeIamResources(NodeIamResourcesBase):
             ),
         ]
 
-        if self._config.dev_settings and self._config.dev_settings.slurm_settings:
+        if (
+            self._config.dev_settings
+            and self._config.dev_settings.slurm_settings
+            and self._config.dev_settings.slurm_settings.munge_key_secret_arn
+        ):
             policy.extend(
                 [
                     iam.PolicyStatement(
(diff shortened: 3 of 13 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1120,6 +1121,27 @@ def __init__(self, capacity_reservation_id: str = None, capacity_reservation_res
         self.capacity_reservation_resource_group_arn = Resource.init_param(capacity_reservation_resource_group_arn)
 
 
+class SlurmSettingsForCustomMungeKey(Resource):
```
EddyMM, line 1165: Optional:
- We could use a more generic name here: `DevSlurmSettings` but since we'll likely remove this config eventually, it's an optional suggestion.

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,68 @@
...
+
+class MungeKeySecretArnValidator(Validator):
+    """MungeKeySecretArn validator."""
+
+    def _validate(self, munge_key_secret_arn: str, region: str):
+        try:
+            arn_components = munge_key_secret_arn.split(":")
+            service = arn_components[2]
+            resource = arn_components[5]
+            if service == "secretsmanager" and resource == "secret":
```
EddyMM: Some of the validators are similar to what we use for the [PasswordSecretArnValidator](https://github.com/aws/aws-parallelcluster/blob/1763c2a36f7bfe60f46acc81a02a86ff031eba78/cli/src/pcluster/validators/directory_service_validators.py#L86).

Is it possible to reuse the same validator?

cli/src/pcluster/config/cluster_config.py
hehe7318: Thanks Eddy!

cli/src/pcluster/validators/secret_validators.py
hehe7318: If you mean to reuse the common codes in a shared function, that's possible but seems not necessary. But to reuse the same validator is not possible I think. Some logics are not the same. This validator only need to check secretsmanager, the [PasswordSecretArnValidator](https://github.com/aws/aws-parallelcluster/blob/1763c2a36f7bfe60f46acc81a02a86ff031eba78/cli/src/pcluster/validators/directory_service_validators.py#L86) also check ssm. And this validator contains a logic of check the size of custom munge key and a logic of decoding custom munge key.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1170,6 +1192,7 @@ def __init__(
         ami_search_filters: AmiSearchFilters = None,
         instance_types_data: str = None,
         timeouts: Timeouts = None,
+        slurm_settings: SlurmSettingsForCustomMungeKey = None,
```
jdeamicis: Please call this internal variable something more precise, maybe `munge_key_settings`. Anyway, it's temporary.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1181,6 +1204,7 @@ def __init__(
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
+        self.slurm_settings = slurm_settings or SlurmSettingsForCustomMungeKey(implied=True)
```
jdeamicis: See previous comment.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1110,6 +1122,9 @@ class ClusterDevSettingsSchema(BaseDevSettingsSchema):
     instance_types_data = fields.Str(metadata={"update_policy": UpdatePolicy.SUPPORTED})
     timeouts = fields.Nested(TimeoutsSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
     compute_startup_time_metric_enabled = fields.Bool(metadata={"update_policy": UpdatePolicy.SUPPORTED})
+    slurm_settings = fields.Nested(
```
jdeamicis: See previous comment about the name of this `slurm_settings` parameter.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -668,6 +668,22 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
             ),
         ]
 
+        if (
+            self._config.dev_settings
+            and self._config.dev_settings.slurm_settings
+            and self._config.dev_settings.slurm_settings.munge_key_secret_arn
```
jdeamicis: Nitpick: I think you should do this with the dictionary `get()` method with a default value.

cli/src/pcluster/validators/secret_validators.py
jdeamicis: I agree with Eddy.

I propose the following. Let's create separate validators for each type of check:
- `SecretArnExistsValidator()` for the existence of the resource;
- `SecretArnServiceValidator()` for the check that the service of the ARN is the one we expect;
- `MungeSecretSizeValidator()` for the size of the secret;
- `MungeSecretBase64Validator()` for the `base64` encoding (maybe merged with the previous one and with all the checks that are relevant to Munge only);
- what else?

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+from pcluster.aws.common import AWSClientError
+from pcluster.validators.common import FailureLevel, Validator
+
+
+class MungeKeySecretArnValidator(Validator):
+    """MungeKeySecretArn validator."""
+
+    def _validate(self, munge_key_secret_arn: str, region: str):
+        try:
+            arn_components = munge_key_secret_arn.split(":")
```
jdeamicis: Are we 100% sure that all ARNs coming from all resources are structured in the same way?

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+            service = arn_components[2]
+            resource = arn_components[5]
+            if service == "secretsmanager" and resource == "secret":
```
jdeamicis: Separate comment on the same line: as a coding pattern, I would avoid nesting a lot of if conditions. Actually, I would do:

```python3
if not (service == "secretsmanager" and resource == "secret"):
    self._add_failure( ... )

# remaining logic in the main nesting level
...
```

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+                # Describe the secret to ensure it exists
+                AWSApi.instance().secretsmanager.describe_secret(munge_key_secret_arn)
+
+                # Get the actual secret value to check if it's valid Base64
+                secret_response = AWSApi.instance().secretsmanager.get_secret_value(munge_key_secret_arn)
+                secret_value = secret_response.get("SecretString")
+
+                if not secret_value:
+                    self._add_failure(
+                        f"The secret {munge_key_secret_arn} does not contain a valid secret string.",
```
jdeamicis: In our design we had the following text:

> The content of the secret {munge_key_secret_arn} is not a valid Base64 encoded string. Please refer to the ParallelCluster official documentation for more information.

Please use this.

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+                try:
+                    # Attempt to decode the secret value from Base64
+                    decoded_secret = base64.b64decode(secret_value)
+                    # Check if the decoded secret size is within the acceptable range
+                    decoded_secret_size_in_bits = len(decoded_secret) * 8
+                    if decoded_secret_size_in_bits < 256 or decoded_secret_size_in_bits > 8192:
+                        self._add_failure(
+                            f"The size of the decoded munge key in the secret {munge_key_secret_arn} "
+                            f"is {decoded_secret_size_in_bits} bits which is outside the allowed range [256-8192].",
```
jdeamicis: Please change this text to:

> The size of the decoded munge key in the secret {custom_munge_key_secret_arn} is {decoded_secret_size_in_bits} bits. Please use a key with a size between 256 and 8192 bits.

cli/tests/pcluster/templates/test_dev_settings.py
```diff
@@ -50,3 +50,36 @@ def test_custom_cookbook(mocker, test_datadir, config_file_name):
...
+    [
+        "config.yaml",
+    ],
+)
+def test_custom_munge_key(mocker, test_datadir, config_file_name):
+    mock_aws_api(mocker)
+    mock_bucket_object_utils(mocker)
+    mocker.patch(
+        "pcluster.aws.secretsmanager.SecretsManagerClient.get_secret_value",
+        return_value={"SecretString": "invalidBase64"},
```
jdeamicis: This line is confusing here.

This test verifies that when the munge key is used, the policy to retrieve the secret is attached to the node.

I would not call the string `invalidBase64`, since this makes the reader think that this test has something to do with the validation of an invalid secret.

cli/tests/pcluster/templates/test_dev_settings.py
```diff
@@ -50,3 +50,36 @@ def test_custom_cookbook(mocker, test_datadir, config_file_name):
...
     assert_that(actions).contains_only("s3:GetObject", "s3:GetBucketLocation")
+
+
+@pytest.mark.parametrize(
+    "config_file_name",
+    [
+        "config.yaml",
+    ],
+)
+def test_custom_munge_key(mocker, test_datadir, config_file_name):
```
jdeamicis: Let's call this test `test_custom_munge_key_iam_policy`.

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,133 @@
...
+from tests.pcluster.validators.utils import assert_failure_level, assert_failure_messages
+
+
+@pytest.mark.parametrize(
+    "munge_key_secret_arn, region, mock_response, expected_response, error_from_aws_service, expected_failure_level",
+    [
+        (
+            "arn:aws:secretsmanager:us-west-2:123456789012:secret:testSecret",
+            "us-west-2",
+            {"SecretString": "validBase64ValueMakeItInRange[256,8192]MakeItInRange[256,8192]"},
```
jdeamicis: Watch out! This string should lead to an error because it contains the `[`, `]` and `,` characters, which are not included in the `base64` alphabet defined by RFC 4648 (the wikipedia page has a [nice table](https://en.wikipedia.org/wiki/Base64#Base64_table_from_RFC_4648)).

It looks like Python's [`base64.b64decode()` method](https://docs.python.org/3/library/base64.html#base64.b64decode) by default discards any character that does not appear in the RFC.

We must set `validate=True` to make sure we are actually validating the original string and not a pre-processed version of it! Also, this is also consistent with what the `base64` bash utility does.

Also, please adapt the other cases of this unit test as required by the change.

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,133 @@
...
+            "The secret arn:aws:secretsmanager:us-west-2:123456789012:otherResource:testSecret is not supported"
+            " in region us-west-2.",
+            None,
+            FailureLevel.ERROR,
+        ),
+        (
+            "arn:aws:secretsmanager:us-west-2:123456789012:secret:testSecret",
+            "us-west-2",
+            {"SecretString": "invalidBase64MakeItInRange[256,8192]MakeItInRange[256,8192]"},
```
jdeamicis: Please put a comment about why "valid..." is a valid encoded secret and "invalid..." is an invalid one (and this has to do with the number of characters in the encoded string, including the number of padding chars).

jdeamicis (changes requested):
Great job!

There are some things to be changed:
- the main thing is that the validator on the base64 encoding of the secret is wrong due to the default behavior of the `base64` python module when decoding. Please address this.
- I would prefer more code reuse in the logic of the validators.
- Some validator messages must be changed.
- Other minor fixes.

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/config/cluster_config.py

cli/src/pcluster/schemas/cluster_schema.py
hehe7318: Done!

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Actually, it's a class, not a dictionary. The `get()` method doesn't fit.

cli/src/pcluster/validators/secret_validators.py
hehe7318: The secrets manager ARN follows following format:
`arn:aws:secretsmanager:region:account-id:secret:secret-name-abc123`
We check the format in the following codes:
`if service == "secretsmanager" and resource == "secret":`
If return False, that means it's not a valid secretsmanager ARN.

But for your question, the answer is no. Like the for the ssm service, it's ARN format is following:
`arn:aws:ssm:<region>:<account-id>:parameter/<parameter-name>`

cli/src/pcluster/validators/secret_validators.py
hehe7318: Done!

cli/src/pcluster/validators/secret_validators.py
hehe7318: Done!

cli/src/pcluster/validators/secret_validators.py
hehe7318: Done!

cli/src/pcluster/validators/secret_validators.py
hehe7318: BTW, I think you quote codes from wrong position. This codes is to check if the `secret_value` is `none`. 
I address your comment in correct position.
```
            except binascii.Error:
                self._add_failure(
                    f"The content of the secret {munge_key_secret_arn} is not a valid Base64 encoded string. "
                    f"Please refer to the ParallelCluster official documentation for more information.",
                    FailureLevel.ERROR,
                )
```

cli/tests/pcluster/templates/test_dev_settings.py
hehe7318: Thanks for reminding, have corrected this issue.

cli/tests/pcluster/templates/test_dev_settings.py
hehe7318: Done!

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: Done, also add a unit test for this. Thanks for reminding, Jacopo!

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: Done!

cli/src/pcluster/validators/secret_validators.py
hehe7318: Done! Now I am on refactoring unit tests for munge key related validators.

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -13,64 +13,78 @@
 
 from pcluster.aws.aws_api import AWSApi
 from pcluster.aws.common import AWSClientError
-from pcluster.validators.common import FailureLevel, Validator
+from pcluster.validators.common import FailureLevel, SecretArnValidator
 
 
-class MungeKeySecretArnValidator(Validator):
-    """MungeKeySecretArn validator."""
+class MungeKeySecretArnExistsValidator(SecretArnValidator):
```
jdeamicis: Nitpick: I believe this is generalizable: the validator could have another input parameter with a dictionary of pairs `{ service : secret_resource_type }`. In this case you could reuse this for the other validators. But that's low priority for now. Please put a TODO here to remind ourselves that we can generalize this.

hehe7318: Done in: `ArnServiceAndResourceValidator`

Also rewrite `_handle_arn_aws_client_error` function in `common.py`. In the future, if we want to call a API that needs a ARN, we can use this function to avoid most of duplicate codes.

cli/src/pcluster/validators/common.py
```diff
@@ -147,30 +148,29 @@ def __init__(self, head_node_instance_id: str = None, during_update: bool = None
...
-            self._add_failure(
-                f"Cannot validate secret {arn}. "
-                "Please refer to ParallelCluster official documentation for more information.",
-                FailureLevel.WARNING,
-            )
+def _get_arn_components(arn: str):
+    return arn.split(":")
+
+
+def _get_service_and_resource(arn: str):
```
jdeamicis: NitPick: call this function `_get_arn_service_and_resource()`

cli/src/pcluster/validators/common.py
```diff
@@ -147,30 +148,29 @@ def __init__(self, head_node_instance_id: str = None, during_update: bool = None
...
+            f"Cannot validate secret {arn} due to lack of permissions. "
+            "Please refer to ParallelCluster official documentation for more information.",
+            FailureLevel.WARNING,
+        )
+    else:
+        validator_instance._add_failure(
+            f"Cannot validate secret {arn}. "
+            "Please refer to ParallelCluster official documentation for more information.",
+            FailureLevel.WARNING,
+        )
```
jdeamicis, line 178: I'm not sure I understand what the functions above are now: are they common functions available outside this module? If so, please remove the underscore at the beginning of their names. Also, they would require some generalization before being placed in a common module, but I guess this is low priority for now.

cli/src/pcluster/validators/common.py
```diff
@@ -147,30 +148,29 @@ def __init__(self, head_node_instance_id: str = None, during_update: bool = None
...
+
+
+def _get_service_and_resource(arn: str):
+    arn_components = _get_arn_components(arn)
+    service = arn_components[2]
+    resource = arn_components[5]
+    return service, resource
+
+
+def _handle_arn_aws_client_error(e: AWSClientError, arn: str, validator_instance):
```
jdeamicis: Please put a TODO here to search for a better place for this function (maybe we could create an abstract class with abstract methods, of which this could be an implementation).

cli/src/pcluster/validators/common.py
```diff
@@ -11,13 +11,16 @@
 #
 # This module contains all the classes representing the Resources objects.
 # These objects are obtained from the configuration file through a conversion based on the Schema classes.
-#
+# pylint: disable=protected-access
```
jdeamicis, line 14: Why was this necessary?

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -82,37 +82,20 @@ def _validate(self, password_secret_arn: str, region: str):
...
         try:
-            # We only require the secret to exist; we do not validate its content.
-            arn_components = password_secret_arn.split(":")
-            service = arn_components[2]
-            resource = arn_components[5]
+            service, resource = _get_service_and_resource(password_secret_arn)
             if service == "ssm":
-                resource = arn_components[5].split("/")[0]
-
+                resource = password_secret_arn.split(":")[5].split("/")[0]
```
jdeamicis: why are we splitting again if we have introduced `_get_service_and_resource()`?

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -82,37 +82,20 @@ def _validate(self, password_secret_arn: str, region: str):
...
             if service == "secretsmanager" and resource == "secret":
                 AWSApi.instance().secretsmanager.describe_secret(password_secret_arn)
             elif service == "ssm" and resource == "parameter" and region == "us-isob-east-1":
-                parameter_name = arn_components[5].split("/")[1]
+                parameter_name = password_secret_arn.split(":")[5].split("/")[1]
```
jdeamicis: why are we splitting again if we have introduced `_get_service_and_resource()`?

cli/src/pcluster/validators/directory_service_validators.py
```diff
@@ -82,37 +82,20 @@ def _validate(self, password_secret_arn: str, region: str):
...
-                    "Please refer to ParallelCluster official documentation for more information.",
-                    FailureLevel.WARNING,
-                )
-            else:
-                self._add_failure(
-                    f"Cannot validate secret {password_secret_arn}. "
-                    "Please refer to ParallelCluster official documentation for more information.",
-                    FailureLevel.WARNING,
-                )
+            _handle_arn_aws_client_error(e, password_secret_arn, self)
```
jdeamicis: This should not be moved outside of this class into an external function, because this really depends on the type of API call that was performed in the caller method.

If you really want to avoid code duplication across multiple validators, the correct way is to have them as sibling classes (children of the same class, possibly an abstract one).

I would say, let's duplicate this code for the time being  (I have already created an internal task to refactor the various Secret validators we have in our code).

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1181,6 +1210,7 @@ def __init__(
         self.compute_startup_time_metric_enabled = Resource.init_param(
             compute_startup_time_metric_enabled, default=False
         )
+        self.munge_key_settings = munge_key_settings or SlurmSettingsForCustomMungeKey(implied=True)
```
jdeamicis, line 1213: Are you sure we need the `or SlurmSettingsForCustomMungeKey(implied=True)` bit? What does it do exactly? Doesn't it create a new object if the `munge_key_settings` object is missing? But if it's missing, it means that the custom munge key part of the config is not necessary, right? Or am I missing something here?

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+                        FailureLevel.ERROR,
+                    )
+            except binascii.Error:
+                self._add_failure(
+                    f"The content of the secret {munge_key_secret_arn} is not a valid Base64 encoded string. "
+                    f"Please refer to the ParallelCluster official documentation for more information.",
+                    FailureLevel.ERROR,
+                )
+        except AWSClientError as e:
+            _handle_arn_aws_client_error(e, munge_key_secret_arn, self)
```
jdeamicis: Are you sure this function (which was derived from the directory service `PasswordSecretArnValidator` can cope with the failures from the `get_secret_value()` API call?

cli/tests/pcluster/templates/test_dev_settings.py
```diff
@@ -50,3 +50,36 @@ def test_custom_cookbook(mocker, test_datadir, config_file_name):
...
+        "config.yaml",
+    ],
+)
+def test_custom_munge_key_iam_policy(mocker, test_datadir, config_file_name):
+    mock_aws_api(mocker)
+    mock_bucket_object_utils(mocker)
+    mocker.patch(
+        "pcluster.aws.secretsmanager.SecretsManagerClient.get_secret_value",
+        return_value={"SecretString": "validBase641234567890123456789012345678901234567"},
+    )
```
jdeamicis: Why is the mock for `get_secret_value` necessary to test the IAM policy? Doesn't it just use the Secret ARN? Have you tried commenting this mock out? Do we run the validators when generating the template?

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,157 @@
...
+        (
+            "arn:aws:secretsmanager:us-west-2:123456789012:bucket:my-bucket",
+            "us-west-2",
+            "validate_service",
+            "validate_resource",
+            "The arn:aws:secretsmanager:us-west-2:123456789012:bucket:my-bucket "
+            "is not supported in region us-west-2.",
+        ),
+    ],
+)
```
jdeamicis: Nitpick: instead of `validate_service` and `validate_resource`, please use `valid_service` and `valid_resource`.

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,157 @@
...
+    [
+        (
+            "arn:aws:validate_service:us-west-2:123456789012:validate_resource",
+            "us-west-2",
+            "validate_service",
+            "validate_resource",
+            None,
+        ),
+        (
+            "arn:aws:s3:us-west-2:123456789012:bucket:my-bucket",
```
jdeamicis: if you are using _abstract_ services and resources as input (i.e. `validate_service` and `validate_resource`), please use abstract services also in your ARNs when you are checking for failures.

cli/src/pcluster/validators/secret_validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+from pcluster.aws.aws_api import AWSApi
+from pcluster.aws.common import AWSClientError
+from pcluster.validators.common import FailureLevel, Validator, _get_service_and_resource, _handle_arn_aws_client_error
+
+
+class ArnServiceAndResourceValidator(Validator):
+    """Validate that Arn is a valid ARN in given region."""
+
+    def _validate(self, arn: str, region: str, expected_service: str, expected_resource: str):
```
jdeamicis, line 28: Please put a TODO to possibly extend to dictionaries of pairs `{"allowed_service" : "allowed_resource"}`.

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,157 @@
...
+            "arn:aws:secretsmanager:us-west-2:123456789012:secret:testSecret",
+            # In Base64 encoding:
+            # - Every 3 bytes of input are represented as 4 characters in the encoded string.
+            # - The length of the encoded string must be a multiple of 4.
+            # - Valid Base64 characters include upper/lowercase letters,
+            #   numbers, '+', '/', and possibly trailing '=' padding chars.
+            # - The '=' padding is used if the number of bytes being encoded is not divisible by 3.
+            # Given these rules:
+            # - "validBase64" is valid because its length is a multiple of 4 and uses valid Base64 characters.
+            # - "invalidBase64" is invalid because its length is not a multiple of 4.
```
jdeamicis, line 77: This is a BRILLIANT comment. Thanks for this!

jdeamicis (commented):
We need to address some issues in the PR.

cli/src/pcluster/validators/common.py
hehe7318: Done!

cli/src/pcluster/validators/common.py
hehe7318: > are they common functions available outside this module?

Yes, but not too common, they can used to handle aws client error and do some preprocessing when **checking a ARN**. I also use it in the existing `PasswordSecretArnValidator`

> remove the underscore

Done.

cli/src/pcluster/validators/common.py
hehe7318: Done!

cli/src/pcluster/validators/common.py
hehe7318: Or the code-linters will show:
************* Module src.pcluster.validators.common
```
src/pcluster/validators/common.py:165:8: W0212: Access to a protected member _add_failure of a client class (protected-access)
src/pcluster/validators/common.py:167:8: W0212: Access to a protected member _add_failure of a client class (protected-access)
src/pcluster/validators/common.py:173:8: W0212: Access to a protected member _add_failure of a client class (protected-access)
```

cli/src/pcluster/validators/directory_service_validators.py
hehe7318: Because the ssm ARN and the secretsmanager ARN have different format.

The secrets manager ARN follows following format:
`arn:aws:secretsmanager:region:account-id:secret:secret-name`

The ssm service ARN format is following:
`arn:aws:ssm:region:account-id:resource-type/resource-id`
`resource-type` is the type of SSM resource, eg: `parameter` or `document`
So The ARN for an SSM parameter might look like this:
`arn:aws:ssm:us-west-2:123456789012:parameter/parameter-name`

cli/src/pcluster/validators/directory_service_validators.py
hehe7318: See above. Because the ssm ARN and the secretsmanager ARN have different format.

cli/src/pcluster/validators/directory_service_validators.py
hehe7318: > have them as sibling classes (children of the same class, possibly an abstract one).

I tried, that's the way I implemented it before. But doesn't work. I use a father class inherit from `Validator` class. But because there's a unit test called `test_slurm_all_validators_are_called` and it consider this father class as a validator that should be called. So that this unit test will fail.

hehe7318: > This should not be moved outside of this class into an external function, because this really depends on the type of API call that was performed in the caller method.

But for now, it's the same as the newly added validator `MungeKeySecretSizeAndBase64Validator`. And I didn't change any logic in this `PasswordSecretArnValidator`. It does the same validate as before.

cli/src/pcluster/config/cluster_config.py
hehe7318: > What does it do exactly?

```
  munge_key_manager 'manage_munge_key' do
    munge_key_secret_arn lazy {
      node['cluster']['config'].dig(:DevSettings, :MungeKeySettings, :MungeKeySecretArn)
    }
  end
```
In the cookbook recipe, we will need `MungeKeySecretArn` as a input in some logic. eg: setup munge key and the rotation script, check if `MungeKeySecretArn` was updated function. 
So `SlurmSettingsForCustomMungeKey(implied=True)` Can ensure at least we have a
```
DevSettings:
  MungeKeySettings:
    MungeKeySecretArn: null
```
in the `cluster-config.yaml` in the shared directory.

Or it will not contain `MungeKeySettings` section. The `dig` method will raise `no method dig for nil class` error. If I don't remember wrong.

cli/src/pcluster/validators/secret_validators.py
hehe7318: I didn't change the logic inside it compared to before. Also didn't change the unit test for this Validator, and the unit test can pass. But anyway, I will do a manually test for it tomorrow.

cli/tests/pcluster/templates/test_dev_settings.py
hehe7318: > Have you tried commenting this mock out?

Just tried. You are correct! Thanks for your reminding! The test can pass with only keeping these two line.
```
    mock_aws_api(mocker)
    mock_bucket_object_utils(mocker)
```
and delete the the mock
```
    mocker.patch(
        "pcluster.aws.secretsmanager.SecretsManagerClient.get_secret_value",
        return_value={"SecretString": "validBase641234567890123456789012345678901234567"},
    )
```

> Do we run the validators when generating the template?

No, I don't think so.

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: Done! Thanks for reminding, Jacopo!

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: Done!

cli/src/pcluster/validators/secret_validators.py
hehe7318: Done!

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: Thank you Jacopo!

cli/src/pcluster/validators/common.py
jdeamicis: That's because `_add_failure()` was supposed to be used only from within the `Validator` class (or a subclass of it). 

Anyway we have a TODO at the top of `handle_arn_aws_client_error` to find a more OOP-oriented way to generalize the Secret validators.

cli/src/pcluster/validators/common.py
jdeamicis: We should probably refactor them in the future, but for the moment I guess they are fine.

cli/src/pcluster/validators/directory_service_validators.py
jdeamicis: Thank you for the explanation. Two comments/questions here:

- this maybe means that the function `get_arn_service_and_resource()` introduced above is already not general or useful enough? We will see that in the future.
- on line 92 I would do:
  ```python3
  resource_type = resource.split("/")[0]
  ```
  possibly differentiating between resource type and resource id. What do you think?

cli/src/pcluster/validators/directory_service_validators.py
jdeamicis: Same as above, I would do:
```python3
parameter_name = resource.split("/")[1]
```
what do you think?

cli/src/pcluster/templates/cdk_builder_utils.py
jdeamicis: Oh yes, here it's a class. Never mind then.

cli/src/pcluster/validators/secret_validators.py
jdeamicis: Yes, I guess I put the comment in the wrong place :)

cli/tests/pcluster/templates/test_dev_settings.py
jdeamicis: I see you totally removed the mocker. If this is really not necessary, even better.

cli/src/pcluster/validators/secret_validators.py
jdeamicis: See new comment. I believe this is fine for the time being.

cli/src/pcluster/config/cluster_config.py
jdeamicis: Exactly, but we can still deal with that logic by checking if the parent section of the dictionary exists.

Still, let's stick to what you implemented right now.

cli/src/pcluster/validators/directory_service_validators.py
jdeamicis: > But because there's a unit test called test_slurm_all_validators_are_called and it consider this father class as a validator that should be called. So that this unit test will fail.

There should be a way to exclude a validator class from that test.

Anyway, it's just a matter of refactoring this code into something more OOP-friendly. I will create a nice-to-have task for this.

cli/src/pcluster/validators/secret_validators.py
jdeamicis: Did you perform a test in then end?

cli/tests/pcluster/validators/test_secret_validators.py
```diff
@@ -0,0 +1,158 @@
...
+from pcluster.validators.common import FailureLevel
+from pcluster.validators.secret_validators import ArnServiceAndResourceValidator, MungeKeySecretSizeAndBase64Validator
+from tests.pcluster.validators.utils import assert_failure_level, assert_failure_messages
+
+
+@pytest.mark.parametrize(
+    "arn, region, expected_service, expected_resource, expected_response",
+    [
+        (
+            "arn:aws:valid_service:us-west-2:123456789012:valid_resource:valid_resource_name",
```
jdeamicis, line 24: If some ARNs are structured differently (like the ssm secrets, where component 5 is structured as `resource-type/resource-id`, are we sure the test cases implemented so far are enough?

Do we want to have a case with only `resource_type` as component 5? Do we want a case with `resource-type/resource-id` as component 5?

jdeamicis (dismissed):
Great! All the major issues have been addressed.

I left some minor comments about making some statements in the code a bit clearer and to possibly include some corner cases in some unit tests.

If you think they are not too important, please dismiss them.

cli/tests/pcluster/validators/test_secret_validators.py
hehe7318: This Validator for now can not handle this situation, let's leave it for future.

cli/src/pcluster/validators/directory_service_validators.py
hehe7318: Done!

jdeamicis (approved):

PR #5684, opened 2023-09-01
Removed dev settings from test config files.

hgreebe:
### Description of changes
Removed dev settings from config file for test_update_instance_list integration test.


enrico-usai:
Useful to link the patch where we wrongly added these changes as references, in this way the two patches will be merged together and this will help for example if we need to cherry pick it.

hanwen-pcluste (approved):

Issue #4499, opened 2022-11-02
Feature Request: Support S3 data import for FSx Persistent 2 File Systems


stefan-vaisala:
**Background:** 
We currently spin-up clusters using PCluster v3.2.1 once a day in an automated fashion. During this process, we spin up a Scratch 2 FSx file system that is connected to a S3 bucket and the data is imported from S3 using the lfs hsm_restore CLI. These clusters and all related resources are only up for ~2 hours; the entire stacks, including the created FSx volume are deleted. Currently, PCluster handles the creation of the Scratch 2 FSx file system and imports the data from S3 to the FSx via the "ImportPath" YAML option. Our post install procedures then run the lfs hsm_restore commands to load the data.

**Feature:**
With the new Persistent 2 FSx file system, the process of connecting the FSx file system to S3 and importing the data from S3 is way more complicated and time intensive than it should be. There are two additional tasks that need to be run to associate S3 to a Persistent 2 FSx file system: i) create a data repository association, and ii) running a data repository task to do a bulk import of the S3 metadata to the FSx volume.

We have added (i) and (ii) to our PCluster post-install bash scripts by using the recommended aws CLI calls (aws fsx create-data-repository-association and aws fsx create-data-repository-task). However, we quickly realized that creating the data repository association takes several minutes, but the aws CLI call returns immediately rather than waiting for the association to transition from "Creating" to "Available". As a result, any subsequent post install procedures that require the association, such as the required data repository task and lfs hsm_restore calls, will not succeed without waiting on this association. Overall, it is not ideal to have our post install script waiting on these AWS resources and tasks to complete/become available as these sort of tasks are "upstream" of anything post install procedures should handle. It seems like this could be handled directly after the FSx is created to save valuable time in cluster spin-up given how long the association takes to become available.

So, it would be great if PCluster could handle the steps described above for Persistent 2 FSx file systems - the creation of the data repository association and the initial data repository task to import the metadata from S3 to the FSx. This functionality already exists for the other FSx file system types supported by PCluster and would save us valuable time if was available for the Persistent 2 FSx as well. Thanks!

hanwen-pcluste:
Hi Stefan,

Thank you for the detailed description. I am going to mark this as a feature request and will discuss within our team.

Hanwen,

enrico-usai:
Released as part of release 3.8.0 

> Add support for Data Repository Associations when using PERSISTENT_2 as DeploymentType for a managed FSx for Lustre.

PR #5693, opened 2023-09-11
Added support for Data Repository Associations with Pcluster Managed FSx for Lustre

hgreebe:
### Description of changes
* FSx released Persistent_2 which no longer supports the use of ExportPath and ImportPath. Instead, it uses DRAs to configure S3 association. 
* The possibility to configure DRAs for managed FSx for Lustre file systems when configuring a cluster will now be supported.

### Tests
- Integration Tests
   -The addition, removal, and update of DRAs was tested
* Unit Tests
  -The new schema and validators were tested

### References
* https://github.com/aws/aws-parallelcluster/issues/4499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ CHANGELOG
 ------
 
 **ENHANCEMENTS**
-
+- Add support for Data Repository Associations when using PERSISTENT_2 as DeploymentType for a managed FSx for Lustre.
 **CHANGES**
 
 **BUG FIXES**
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -139,10 +139,10 @@ from pcluster.validators.ec2_validators import (
 from pcluster.validators.efs_validators import EfsMountOptionsValidator
 from pcluster.validators.feature_validators import FeatureRegionValidator
 from pcluster.validators.fsx_validators import (
-    DraValidator,
     FsxAutoImportValidator,
     FsxBackupIdValidator,
     FsxBackupOptionsValidator,
+    FsxDraValidator,
     FsxPersistentOptionsValidator,
     FsxS3Validator,
     FsxStorageCapacityValidator,
@@ -189,7 +189,6 @@ from pcluster.validators.slurm_settings_validator import (
 from pcluster.validators.tags_validators import ComputeResourceTagsValidator
 
 LOGGER = logging.getLogger(__name__)
-
 
 # pylint: disable=C0302
 
@@ -434,12 +433,14 @@ class BaseSharedFsx(Resource):
 
 
 class DataRepositoryAssociation(Resource):
+    """Represent the Data Repository Association resource."""
+
     def __init__(
         self,
-        name: str = None,
-        batch_import_metadata_on_create: bool = None,
-        data_repository_path: str = None,
-        file_system_path: str = None,
+        name: str,
+        data_repository_path: str,
+        file_system_path: str,
+        batch_import_meta_data_on_create: bool = None,
         imported_file_chunk_size: int = None,
         auto_export_policy: List[str] = None,
         auto_import_policy: List[str] = None,
@@ -447,10 +448,10 @@ class DataRepositoryAssociation(Resource):
     ):
         super().__init__(**kwargs)
         self.name = Resource.init_param(name)
-        self.batch_import_metadata_on_create = Resource.init_param(batch_import_metadata_on_create, default=False)
+        self.batch_import_meta_data_on_create = Resource.init_param(batch_import_meta_data_on_create, default=False)
         self.data_repository_path = Resource.init_param(data_repository_path)
         self.file_system_path = Resource.init_param(file_system_path)
-        self.imported_file_chunk_size = Resource.init_param(imported_file_chunk_size)
+        self.imported_file_chunk_size = Resource.init_param(imported_file_chunk_size, default=1024)
         self.auto_export_policy = Resource.init_param(auto_export_policy)
         self.auto_import_policy = Resource.init_param(auto_import_policy)
 
@@ -519,7 +520,7 @@ class SharedFsxLustre(BaseSharedFsx):
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
         self._register_validator(
-            DraValidator,
+            FsxDraValidator,
             data_repository_associations=self.data_repository_associations,
             import_path=self.import_path,
             export_path=self.export_path,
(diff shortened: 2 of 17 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ------
 
 **ENHANCEMENTS**
+- Add support for Data Repository Associations with Pcluster Managed FSx for Lustre.
```
enrico-usai: - Add support for Data Repository Associations when using `PERSISTENT_2` as `DeploymentType` for a managed FSx for Lustre.

cli/src/pcluster/config/config_patch.py
```diff
@@ -189,18 +189,19 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
...
-            if not base_nested_section.get("visited", False):
-                self.changes.append(
-                    Change(
-                        param_path,
-                        data_key,
-                        base_nested_section,
-                        None,
-                        change_update_policy,
-                        is_list=True,
+        if base_section:
```
hanwen-pcluste, line 192: Why this change?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -179,6 +180,49 @@ def test_fsx_lustre(
     )
 
 
+@pytest.mark.usefixtures("os", "instance", "scheduler")
+def test_fsx_lustre_dra(
```
hanwen-pcluste, line 184: Have you thought about merging this test to another test?
If merging the test is hard, we can keep this as separate. If we keep this as separate, let's add it to https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/configs/common/common.yaml so that the test will be included in the daily integration test

cli/src/pcluster/config/cluster_config.py
```diff
@@ -431,6 +432,38 @@ def existing_dns_name(self):
...
 
+class DataRepositoryAssociation(Resource):
+    """Represent the Data Repository Association resource."""
+
+    def __init__(
+        self,
+        name: str = None,
+        batch_import_meta_data_on_create: bool = None,
+        data_repository_path: str = None,
+        file_system_path: str = None,
```
hanwen-pcluste: Let's not put `None` defaults for "required" parameters. This is not a bug. But it is the custom in this file.

cli/src/pcluster/validators/fsx_validators.py
```diff
@@ -48,6 +48,28 @@ def _validate(
...
+    Verify compatibility of given S3 association options for FSX.
+    """
+
+    def _validate(self, data_repository_associations, import_path, export_path):
+        if data_repository_associations and (import_path or export_path):
+            self._add_failure(
+                "When specifying data repository associations, import and export path "
+                "can not be used on the same file system.",
+                FailureLevel.ERROR,
+            )
```
hanwen-pcluste, line 64: Is this validation already done in `validate_s3_association_unsupported_parameters`?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -179,6 +180,49 @@ def test_fsx_lustre(
...
+    region,
+    pcluster_config_reader,
+    s3_bucket_factory,
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+):
+    """Test Data Respository Associations for a FSx Lustre file system."""
+    mount_dir = "/fsx_mount_dir"
+    bucket_name = "integ-tests-knffdu7meov017vf"  # s3_bucket_factory()
```
hanwen-pcluste: Use `s3_bucket_factory()`

tests/integration-tests/tests/storage/test_fsx_lustre/test_fsx_lustre_dra/pcluster.config.yaml
```diff
@@ -0,0 +1,53 @@
...
+    KeyName: {{ key_name }}
+  Iam:
+    S3Access:
+      - BucketName: {{ bucket_name }}
+        EnableWriteAccess: False
+  Imds:
+    Secured: {{ imds_secured }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  {% if scheduler == "awsbatch" %}AwsBatchQueues:{% else %}SlurmQueues:{% endif %}
```
hanwen-pcluste: We don't need this conditional statement, because this this is for Slurm only anyway. Same comment for pcluster.config.update.yaml

tests/integration-tests/tests/storage/test_fsx_lustre/test_fsx_lustre_dra/pcluster.config.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - Name: queue-0
+      Iam:
+        S3Access:
+          - BucketName: {{ bucket_name }}
+            EnableWriteAccess: False
+      ComputeResources:
+        - Name: compute-resource-0
+          {% if scheduler == "awsbatch" %}
```
hanwen-pcluste: We don't need this conditional statement, because this this is for Slurm only anyway. Same comment for pcluster.config.update.yaml

tests/integration-tests/tests/storage/test_fsx_lustre.py
hgreebe: I considered adding it to test_fsx_luster but since that uses config files with import/export path it got a bit complicated because I would have to unmount the old file system and then mount a new one with dras

cli/src/pcluster/config/config_patch.py
hgreebe: because when the _test_storage from test_config_patch.py runs, it calls _compare_list on with a base_section of None, because it tries to go into data_repository_association since it is a nested attribute, but data_repository_association is None, which makes the test fail without this change.

tests/integration-tests/configs/common/common.yaml
```diff
@@ -540,6 +540,16 @@ storage:
...
+  test_fsx_lustre.py::test_fsx_lustre_dra:
+    dimensions:
+      - regions: [ "eu-west-2" ]
+        instances: {{ common.INSTANCES_DEFAULT_X86 }}
+        oss: [ "alinux2" ]
+        schedulers: [ "slurm" ]
+      - regions: [ "eu-north-1" ]
+        instances: {{ common.INSTANCES_DEFAULT_ARM }}
+        oss: [ "ubuntu2204" ]
+        schedulers: [ "slurm" ]
```
hanwen-pcluste: We can have the test run on a single region with a single os and instance type. Because this feature doesn't depends on os and instance type.
We can use eu-north-1

cli/src/pcluster/config/config_patch.py
hanwen-pcluste: Thank you.

hanwen-pcluste (approved):

PR #5713, opened 2023-09-21
[Disable Sudo] Add Config Option to disable Sudo access for default users in supported OSSes

himani2411:
### Description of changes
* Add Config Option so that users can disable sudo access for [default users](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/managing-users.html)
* Adding support for only Slurm schedulers
* Disable Sudo access using cloud-config section of cloud-init
      * Remove policy for Rocky8 ( only OS) which has a policy override in /etc/sudoers file
* Add Unit Tests to check the Output/content of Launch Templates for HeadNode, LoginNode and ComputeFleet.
* Add Integ test 

### Tests
* Added Unit Tests 
* Tested Integration Tests Manually

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ CHANGELOG
   - Critical Update for Intel oneAPI DPC++/C++ Compiler: 2023.2.1
   - Critical Update for Intel Fortran Compiler & Intel® Fortran Compiler Classic: 2023.2.1
   - Remove `CommunicationParameters` from the Custom Slurm Settings deny list.
+  - Add `DisableSudoAccessForDefaultUser` parameter to disable sudo access of default user in supported OSes.
 
 **CHANGES**
 - Upgrade Slurm to 23.11.1.
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1557,7 +1557,7 @@ class BaseClusterConfig(Resource):
         self.managed_compute_security_group = None
         self.instance_types_data_version = ""
         self._set_default_head_node_root_volume_size()
-        self.disable_sudo_access_default_user = Resource.init_param(disable_sudo_access_default_user)
+        self.disable_sudo_access_default_user = Resource.init_param(disable_sudo_access_default_user, default=False)
 
     def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
         self._register_validator(RegionValidator, region=self.region)
@@ -2027,11 +2027,6 @@ class BaseClusterConfig(Resource):
     def get_instance_types_data(self) -> dict:
         """Get instance type infos for all instance types used in the configuration file."""
         return {}
-
-    @property
-    def is_default_user_sudo_access_enabled(self):
-        """Return True if DisableSudoAccessForDefaultUser is True."""
-        return bool(self.disable_sudo_access_default_user)
 
 
 class AwsBatchComputeResource(BaseComputeResource):
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1522,6 +1522,7 @@ class BaseClusterConfig(Resource):
         additional_resources: str = None,
         dev_settings: ClusterDevSettings = None,
         deployment_settings: DeploymentSettings = None,
+        disable_sudo_access_default_user: bool = None,
     ):
         super().__init__()
         self.__region = None
@@ -1556,6 +1557,7 @@ class BaseClusterConfig(Resource):
         self.managed_compute_security_group = None
         self.instance_types_data_version = ""
         self._set_default_head_node_root_volume_size()
+        self.disable_sudo_access_default_user = Resource.init_param(disable_sudo_access_default_user)
 
     def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
         self._register_validator(RegionValidator, region=self.region)
@@ -1598,21 +1600,13 @@ class BaseClusterConfig(Resource):
                 os=self.image.os,
                 architecture=self.head_node.architecture,
             )
-        feature_requiring_additional_space = []
-        if self.additional_packages and self.additional_packages.intel_software:
-            if self.additional_packages.intel_software.intel_hpc_platform:
-                self._register_validator(IntelHpc2018OsValidator, os=self.image.os)
-                self._register_validator(IntelHpcArchitectureValidator, architecture=self.head_node.architecture)
-            if (
-                self.additional_packages.intel_software.one_api
-                and self.additional_packages.intel_software.one_api.base_toolkit
-            ):
-                feature_requiring_additional_space.append(Feature.INTEL_ONE_API_BASE_TOOLKIT)
-                self._register_validator(IntelOneApiToolkitsOsValidator, os=self.image.os)
-                self._register_validator(IntelOneApiToolkitsBootstrapTimeValidator)
-            if self.additional_packages.intel_software.python:
-                self._register_validator(IntelPythonOsValidator, os=self.image.os)
-
+        if (
+            self.additional_packages
+            and self.additional_packages.intel_software
+            and self.additional_packages.intel_software.intel_hpc_platform
+        ):
+            self._register_validator(IntelHpcOsValidator, os=self.image.os)
+            self._register_validator(IntelHpcArchitectureValidator, architecture=self.head_node.architecture)
         if self.custom_s3_bucket:
             self._register_validator(S3BucketValidator, bucket=self.custom_s3_bucket)
             self._register_validator(S3BucketRegionValidator, bucket=self.custom_s3_bucket, region=self.region)
@@ -1630,7 +1624,6 @@ class BaseClusterConfig(Resource):
             RootVolumeSizeValidator,
             root_volume_size=root_volume_size,
             ami_volume_size=ami_volume_size,
-            additional_space=self._additional_space_for_features(),
         )
         self._register_validator(
             EbsVolumeTypeSizeValidator, volume_type=root_volume.volume_type, volume_size=root_volume_size
@@ -1789,44 +1782,6 @@ class BaseClusterConfig(Resource):
         if volume_ids:
             AWSApi.instance().fsx.describe_volumes(volume_ids)
 
-    def _set_default_head_node_root_volume_size(self):
-        root_volume = self.head_node.local_storage.root_volume
-        if root_volume.size is None:
-            root_volume.size = Resource.init_param(
-                None,
-                default=AWSApi.instance().ec2.describe_image(self.head_node_ami).volume_size
-                + self._additional_space_for_features(),
-            )
-
-    def _additional_space_for_features(self):
-        feature_requiring_additional_space = []
-        if (
-            self.additional_packages
-            and self.additional_packages.intel_software
-            and self.additional_packages.intel_software.one_api
-            and self.additional_packages.intel_software.one_api.base_toolkit
-        ):
-            feature_requiring_additional_space.append(Feature.INTEL_ONE_API_BASE_TOOLKIT)
-        additional_space = 0
-        for feature in feature_requiring_additional_space:
-            additional_space += FEATURE_REQUIRING_ADDITION_SPACE[feature]
-        return additional_space
-
-    def additional_bootstrap_time_for_features(self):
-        """Calculate total additional bootstrap time for enabled features."""
-        feature_requiring_additional_bootstrap_time = []
-        if (
-            self.additional_packages
-            and self.additional_packages.intel_software
-            and self.additional_packages.intel_software.one_api
-            and self.additional_packages.intel_software.one_api.base_toolkit
-        ):
-            feature_requiring_additional_bootstrap_time.append(Feature.INTEL_ONE_API_BASE_TOOLKIT)
-        additional_bootstrap_time = 0
-        for feature in feature_requiring_additional_bootstrap_time:
-            additional_bootstrap_time += FEATURE_REQUIRING_ADDITION_BOOTSTRAP_TIME[feature]
-        return additional_bootstrap_time
-
     @property
     def region(self):
         """Retrieve region from environment if not set."""
@@ -1979,11 +1934,6 @@ class BaseClusterConfig(Resource):
         )
 
     @property
-    def are_alarms_enabled(self):
-        """Return False if Monitoring/Alarms/Enabled is False. True otherwise."""
-        return self.monitoring.alarms.enabled if self.monitoring and self.monitoring.alarms else True
-
-    @property
     def is_detailed_monitoring_enabled(self):
         """Return True if Detailed Monitoring is enabled."""
         return self.monitoring.detailed_monitoring
@@ -2077,6 +2027,11 @@ class BaseClusterConfig(Resource):
     def get_instance_types_data(self) -> dict:
         """Get instance type infos for all instance types used in the configuration file."""
         return {}
+
+    @property
+    def is_default_user_sudo_access_enabled(self):
+        """Return True if DisableSudoAccessForDefaultUser is True."""
+        return bool(self.disable_sudo_access_default_user)
 
 
 class AwsBatchComputeResource(BaseComputeResource):
@@ -2223,11 +2178,6 @@ class _BaseSlurmComputeResource(BaseComputeResource):
         self.static_node_priority = Resource.init_param(static_node_priority, default=1)
         self.dynamic_node_priority = Resource.init_param(dynamic_node_priority, default=1000)
 
-    @abstractmethod
-    def is_flexible(self) -> bool:
-        """Return True if the ComputeResource can contain multiple instance types, False otherwise."""
-        pass
-
     @staticmethod
     def fetch_instance_type_info(instance_type) -> InstanceTypeInfo:
         """Return instance type information."""
@@ -2298,7 +2248,11 @@ class FlexibleInstanceType(Resource):
 class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with Multiple Instance Types."""
 
-    def __init__(self, instances: List[FlexibleInstanceType], **kwargs):
+    def __init__(
+        self,
+        instances: List[FlexibleInstanceType],
+        **kwargs,
+    ):
         super().__init__(**kwargs)
         self.instances = Resource.init_param(instances)
 
@@ -2324,10 +2278,6 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
             ec2memory=min_memory,
             instance_type=smallest_type,
         )
-
-    def is_flexible(self):
-        """Return True because the ComputeResource can contain multiple instance types."""
-        return True
 
     @property
     def disable_simultaneous_multithreading_manually(self) -> bool:
@@ -2373,23 +2323,28 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
 class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
-    def __init__(self, instance_type=None, **kwargs):
+    def __init__(
+        self,
+        instance_type,
+        **kwargs,
+    ):
         super().__init__(**kwargs)
-        _instance_type = instance_type if instance_type else self._instance_type_from_capacity_reservation()
-        self.instance_type = Resource.init_param(_instance_type)
+        self.instance_type = Resource.init_param(instance_type)
         self.__instance_type_info = None
-
-    def is_flexible(self):
-        """Return False because the ComputeResource can not contain multiple instance types."""
-        return False
 
     @property
     def instance_types(self) -> List[str]:
         """List of instance types under this compute resource."""
         return [self.instance_type]
 
+    @property
+    def instance_type_info(self) -> InstanceTypeInfo:
+        """Return instance type information."""
+        return AWSApi.instance().ec2.get_instance_type_info(self.instance_type)
+
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
+        self._register_validator(ComputeResourceSizeValidator, min_count=self.min_count, max_count=self.max_count)
         self._register_validator(
             SchedulableMemoryValidator,
             schedulable_memory=self.schedulable_memory,
@@ -2427,19 +2382,7 @@ class SlurmComputeResource(_BaseSlurmComputeResource):
     @property
     def disable_simultaneous_multithreading_manually(self) -> bool:
         """Return true if simultaneous multithreading must be disabled with a cookbook script."""
-        return self.disable_simultaneous_multithreading and self._instance_type_info.default_threads_per_core() > 1
-
-    def _instance_type_from_capacity_reservation(self):
-        """Return the instance type from the configured CapacityReservationId, if any."""
-        instance_type = None
-        capacity_reservation_id = (
-            self.capacity_reservation_target.capacity_reservation_id if self.capacity_reservation_target else None
-        )
-        if capacity_reservation_id:
-            capacity_reservations = AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation_id])
-            if capacity_reservations:
-                instance_type = capacity_reservations[0].instance_type()
-        return instance_type
+        return self.disable_simultaneous_multithreading and self.instance_type_info.default_threads_per_core() > 1
 
 
 class _CommonQueue(BaseQueue):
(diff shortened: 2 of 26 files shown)


himani2411:
looking at failing tests

cli/src/pcluster/config/cluster_config.py
```diff
@@ -97,6 +97,7 @@
     RootVolumeEncryptionConsistencyValidator,
     RootVolumeSizeValidator,
     SchedulableMemoryValidator,
+    SchedulerDisableSudoAccessForDefaultUserValidator,
```
gmarciani, line 106: Why the validator is related to the scheduler? Isn't this disablement cluster-wide?

cli/src/pcluster/constants.py
```diff
@@ -320,3 +320,5 @@ class Operation(Enum):
 IAM_ROLE_REGEX = "^arn:.*:role/"
 IAM_INSTANCE_PROFILE_REGEX = "^arn:.*:instance-profile/"
 IAM_POLICY_REGEX = "^arn:.*:policy/"
+
+DISABLE_SUDO_ACCESS_FOR_DEFAULT_USER_CONFIG = {"system_info": {"default_user": {"sudo": ["ALL=(ALL) !ALL"]}}}
```
gmarciani, line 344: This constant represents the cloud-init configuration? If this is the case, let's mention CloudInit in the variable name or add a comment specifying it.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1846,6 +1846,9 @@ class ClusterSchema(BaseSchema):
     additional_resources = fields.Str(metadata={"update_policy": UpdatePolicy.SUPPORTED})
     dev_settings = fields.Nested(ClusterDevSettingsSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
     deployment_settings = fields.Nested(DeploymentSettingsSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
+    disable_sudo_default_user = fields.Bool(
```
gmarciani: *Minor*: something we name variable/classes like "disable_sudo_default_user" and other time "disable sudo *access*". I suggest to align the naming to avoid misunderstandings (even if really low probable in this case).

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -122,6 +124,10 @@ def get_directory_service_dna_json_for_head_node(config: BaseClusterConfig) -> d
     )
 
 
+def get_cloud_config_for_default_user(is_default_user_sudo_access_enabled):
```
gmarciani: I suggest to specify in a comment that having an empty configuration is equivalent to falling back to the default CloudInit configuration

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -121,6 +121,21 @@ def _validate(self, region):
...
+    """
+    Validator for DisableSudoAccessForDefaultUser and AWS Batch Scheduler.
+
+    Fail if using AWS Batch and DisableSudoAccessForDefaultUser is not supported.
+    """
+
+    def _validate(self, scheduler):
+        if scheduler == "awsbatch":
+            self._add_failure(
+                "DisableSudoAccessForDefaultUser is not supported when using AWS Batch as scheduler.",
```
gmarciani, line 138: Why the disablement cannot be applied when using AWS Batch? 
It should be applied in the head node and login node regardless the adopted scheduler.
Am I missing something here?

tests/integration-tests/tests/create/test_create.py
```diff
@@ -117,6 +119,39 @@ def test_create_imds_secured(
...
+    cluster_config = pcluster_config_reader(disable_sudo_access_default_user=disable_sudo_access_default_user)
+    cluster = clusters_factory(cluster_config, raise_on_error=True)
+
+    logging.info("Checking default user has disabled sudo access after cluster creation")
+    assert_head_node_is_running(region, cluster)
+    for node_type in NodeType:
+        assert_default_user_has_desired_sudo_access(
+            cluster, node_type, disable_sudo_access_default_user, region, scheduler_commands_factory
+        )
+    # logging.info("Updating Cluster to disable sudo access")
```
gmarciani: Is there value in keeping this commented out code?

gmarciani:
We should add an entry in the changelog for this change.

cli/src/pcluster/constants.py
himani2411: Its part of cloud-config section in cloud-init. The name is becoming super long which is why I added the `cloud-config` in function where I use it. Will add the comment as suggested.

cli/src/pcluster/templates/cdk_builder_utils.py
himani2411: will add the comment

cli/src/pcluster/validators/cluster_validators.py
himani2411: The reason I didnt add the change for Batch is that the change wouldnt be added in ComputeFleet, which would make the feature more confusing when we are able to disable the access in HeadNode and Login Node only for Batch and all the nodes when using Slurm.

I dont see value in a cluster where just HN and LN have disabled sudo access for default user but not the compute node.

tests/integration-tests/tests/create/test_create.py
himani2411: Just for me when I want to test after changing the  update policy of the feature. Will remove it and just add it in my local

cli/src/pcluster/config/cluster_config.py
himani2411: I am confused as why the validator should be cluster-wide? The validator I added is related to Disabling the feature for [AWSBatch](https://github.com/aws/aws-parallelcluster/pull/5713#discussion_r1335566242)

gmarciani: I think the validator should not be there because the configuration is cluster-wide so there is no reason to consider it as not appliable to a aws batch cluster (in a aws batch cluster you would apply the sudoer removal in the head node)

cli/src/pcluster/validators/cluster_validators.py
gmarciani: I think we can solve the potential confusion with public docs saying that in case of AWS Batch scheduler the removal of sudo access is applied only on the head node. I suggest this approach, but I suggest to sync with the PM on this.

himani2411: I re-checked the [code](https://github.com/aws/aws-parallelcluster/blob/2481f621bef8d4967a92b3b5dab53b12e63e3c1c/cli/src/pcluster/templates/awsbatch_builder.py#L159-L160) and I see that we can add the [user-data.sh for Ec2 user in AWSBatch](https://docs.aws.amazon.com/batch/latest/userguide/launch-templates.html). However, if you agree I can add this and [remove the validation for AWSBatch](https://github.com/aws/aws-parallelcluster/pull/5713#discussion_r1335554841) in the next PR,  as I would need to test the change for AWSBatch.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ CHANGELOG
   - Critical Update for Intel oneAPI DPC++/C++ Compiler: 2023.2.1
   - Critical Update for Intel Fortran Compiler & Intel® Fortran Compiler Classic: 2023.2.1
   - Remove `CommunicationParameters` from the Custom Slurm Settings deny list.
+- Add `DisableSudoAccessForDefaultUser` parameter to Disable Sudo Access of Default users in supported OSes.
```
gmarciani: Let's remove the uppercase letters (eg to "Disable Sudo Access" => "to disable sudo access").
Also, every cluster has one single default user, so "Default users" => "default user" (singular)

cli/src/pcluster/constants.py
```diff
@@ -339,3 +339,6 @@ class Operation(Enum):
 IAM_ROLE_REGEX = "^arn:.*:role/"
 IAM_INSTANCE_PROFILE_REGEX = "^arn:.*:instance-profile/"
 IAM_POLICY_REGEX = "^arn:.*:policy/"
+
+# Section of Cloud-Init(cloud-config) which will be added for Disabling the Sudo access of Default user
+DISABLE_SUDO_ACCESS_FOR_DEFAULT_USER_CONFIG = {"system_info": {"default_user": {"sudo": ["ALL=(ALL) !ALL"]}}}
```
gmarciani, line 344: *Not on this PR*, but an observation to be addressed in a follow up PR.
The `constants.py` file is getting bigger and bigger (more than 300 lines of constants) and I think we could start splitting it since most of the constants belongs to different categories. For this specific constant it could be put in a file `constants/cloud_init.py`

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -55,6 +55,10 @@ package_upgrade: false
 repo_upgrade: none
 
 datasource_list: [ Ec2, None ]
+
+${DisableSudoAccessForDefaultUserConfig}
```
gmarciani, line 63: This config references the `default_user` which is not defined within cloud-init config (it is in the CloudInit copnfig for Batch).
I suppose the default user is defined implicitly by CloudInit, but may you please confirm?

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -55,6 +55,10 @@ package_upgrade: false
...
+
+
```
gmarciani: Extra blank line

cli/src/pcluster/resources/head_node/user_data.sh
```diff
@@ -56,6 +56,9 @@ repo_upgrade: none
 
 datasource_list: [ Ec2, None ]
 
+${DisableSudoAccessForDefaultUserConfig}
+
+
```
gmarciani: Extra blank line

cli/src/pcluster/templates/awsbatch_builder.py
```diff
@@ -156,7 +157,10 @@ def _launch_template(self, http_tokens):
             self.stack_scope,
             "PclusterComputeEnvironmentLaunchTemplate",
             launch_template_data=ec2.CfnLaunchTemplate.LaunchTemplateDataProperty(
-                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens=http_tokens)
+                metadata_options=ec2.CfnLaunchTemplate.MetadataOptionsProperty(http_tokens=http_tokens),
+                user_data=Fn.base64(get_user_data_content("../resources/login_node/user_data.sh"))
```
gmarciani: Why login nodes user data? Shouldn't it be batch compute env user data?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -130,6 +132,11 @@ def get_directory_service_dna_json_for_head_node(config: BaseClusterConfig) -> d
...
+    """Return Cloud-Init Section ( cloud-config) in YAML format if DisableSudoAccessForDefaultUser is enabled."""
```
gmarciani: extra white space "( cloud-" => "(cloud-"

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -468,25 +489,41 @@ def assert_lt_properties(self, generated_template, resource_type):
...
         test_datadir,
     )
 
     asset_content = get_asset_content_with_resource_name(cdk_assets, "LaunchTemplate64e1c3597ca4c326")
     for lt_assertion in lt_assertions:
         lt_assertion.assert_lt_properties(asset_content, "LaunchTemplate64e1c3597ca4c326")
 
+    # Checking for DisableSudoAccessForDefaultUser
+    assert_that(
+        asset_content["Resources"]["LaunchTemplate64e1c3597ca4c326"]["Properties"]["LaunchTemplateData"]["UserData"][
```
gmarciani: The launch template logical id (LaunchTemplate64e1c3597ca4c326) is repeat three times, what about defining a constant?

cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -364,25 +366,40 @@ def get_generated_template_and_cdk_assets(
...
         test_datadir,
     )
 
     asset_content = get_asset_content_with_resource_name(cdk_assets, "LaunchTemplate64e1c3597ca4c326")
 
     for lt_assertion in lt_assertions:
         lt_assertion.assert_lt_properties(asset_content, "LaunchTemplate64e1c3597ca4c326")
+    # Checking for DisableSudoAccessForDefaultUser
+    assert_that(
+        asset_content["Resources"]["LaunchTemplate64e1c3597ca4c326"]["Properties"]["LaunchTemplateData"]["UserData"][
```
gmarciani: The launch template logical id (LaunchTemplate64e1c3597ca4c326) is repeat three times, what about defining a constant?

cli/tests/pcluster/templates/test_cluster_stack/test_compute_launch_template_properties/cluster-using-flexible-instance-types.yaml
```diff
@@ -23,3 +23,6 @@ Scheduling:
     Networking:
       SubnetIds:
       - subnet-12345678
+{% if disable_sudo_access_default_user == True %}
```
gmarciani: Why should we wrap the property within the if statement?
Wouldn't it be enough to do as follows?

```
DisableSudoAccessForDefaultUser: {{ disable_sudo_access_default_user }}
```

tests/integration-tests/configs/common/common.yaml
```diff
@@ -151,6 +151,12 @@ create:
         instances: {{ common.INSTANCES_DEFAULT_ARM }}
         schedulers: ["slurm"]
         oss: ["ubuntu2004"]
+  test_create.py::test_create_disable_sudo_access_for_default_user:
+    dimensions:
+      - regions: [ "ap-northeast-2" ]
+        instances: {{ common.INSTANCES_DEFAULT_X86 }}
+        schedulers: {{ common.SCHEDULERS_ALL }}
+        oss: [ "alinux2" ]
```
gmarciani: I suggest to test the case on at least two OSes. We are pretty confident that CloudInit does not change its behaviour with the Os, but with security sensitive features like this it's worth it to have more confidence.

Also I think that this test case could be a good candidate to be incorporated within the existing integ test on IMDS secured.

tests/integration-tests/tests/common/assertions.py
```diff
@@ -249,6 +256,43 @@ def assert_aws_identity_access_is_correct(cluster, users_allow_list, remote_comm
...
+    )
+    if node_type == NodeType.HEAD_NODE:
+        remote_command_executors.append(RemoteCommandExecutor(cluster))
+    elif node_type == NodeType.COMPUTE_NODES:
+        for compute_node_ip in get_compute_nodes_instance_ips(cluster.name, region):
+            remote_command_executors.append(RemoteCommandExecutor(cluster, compute_node_ip=compute_node_ip))
+    elif node_type == NodeType.LOGIN_NODES:
+        remote_command_executors.append(RemoteCommandExecutor(cluster, use_login_node=True))
+
+    command = "sudo cat /etc/sudoers.d/90-cloud-init-users"
```
gmarciani: If we just need to verify the sudo access, even a simple "sudo -v" should be enough

tests/integration-tests/tests/common/assertions.py
```diff
@@ -249,6 +256,43 @@ def assert_aws_identity_access_is_correct(cluster, users_allow_list, remote_comm
...
+    for node_index, remote_command_executor in enumerate(remote_command_executors):
+        result = remote_command_executor.run_remote_command(command, raise_on_error=False)
+        logging.info(f"Default user in {node_type} number {node_index} and result.failed={result.failed}")
```
gmarciani, line 284: Checking that the command failed is not enough as it could fail for many reasons. We should specifically verify that the command fails with the expected error

tests/integration-tests/tests/common/utils.py
```diff
@@ -272,13 +272,36 @@ def reboot_head_node(cluster, remote_command_executor=None):
...
 
+def reboot_login_node(cluster, remote_command_executor=None):
+    logging.info(f"Rebooting login node for cluster: {cluster.name}")
+    if not remote_command_executor:
+        remote_command_executor = RemoteCommandExecutor(cluster, use_login_node=True)
+    command = "sudo reboot"
+    result = remote_command_executor.run_remote_command(command, raise_on_error=False)
+    logging.info(f"result.failed={result.failed}")
+    logging.info(f"result.stdout={result.stdout}")
+    wait_login_node_running(cluster)
```
gmarciani, line 283: I think we should not only waiut for the instance to be running, but also for the instance to be in status ok. See https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/waiter/InstanceStatusOk.html

tests/integration-tests/tests/create/test_create.py
```diff
@@ -117,6 +119,29 @@ def test_create_imds_secured(
     assert_cluster_imds_v2_requirement_status(region, cluster, status)
 
 
+# FIXME: Move in tests_update after update policy is changed
+@pytest.mark.usefixtures("instance", "scheduler")
+def test_create_disable_sudo_access_for_default_user(
+    region, os, scheduler, pcluster_config_reader, clusters_factory, architecture, scheduler_commands_factory
+):
+    """
+    Verify that the cluster removes the Sudo access for Default user
```
gmarciani: Default => default

tests/integration-tests/tests/create/test_create.py
```diff
@@ -117,6 +119,29 @@ def test_create_imds_secured(
...
+    in all the nodes of the Cluster if the DisableSudoAccessForDefaultUser is enabled.
+    """
+    disable_sudo_access_default_user = True
+    cluster_config = pcluster_config_reader(disable_sudo_access_default_user=disable_sudo_access_default_user)
+    cluster = clusters_factory(cluster_config, raise_on_error=True)
+
+    logging.info("Checking default user has disabled sudo access after cluster creation")
+    assert_head_node_is_running(region, cluster)
+    for node_type in NodeType:
+        if scheduler == "awsbatch" and node_type.value == "LoginNodes":
```
gmarciani: Shouldn't it be singular: LoginNode?

cli/src/pcluster/templates/awsbatch_builder.py
himani2411: Thanks for the Catch

cli/src/pcluster/resources/compute_node/user_data.sh
himani2411: Yes it is. The below section that you saw in cloud-init config for Batch is the same for every node or ec2 instance.
```

system_info:
  default_user:
    sudo:
    - ALL=(ALL) !ALL
```

The `${DisableSudoAccessForDefaultUserConfig}` replaces the above section I had harcoded for Batch.

Its just that I hardcoded it in Batch as I wanted to avoid using the if else conditions in Batch. But sake of simplicity I will add that.

tests/integration-tests/tests/create/test_create.py
himani2411: Will not check for another Hardcoded Value

tests/integration-tests/tests/common/assertions.py
himani2411: Agreed but I want to check the Policy too, as thats the part which is being added as part of user_data.sh. I plan on using this assertion to test when we dont use `DisableSudoAccessDefaultUser` which is why I have added conditions for what the value of `disable_sudo_access_default_user` is True or False

tests/integration-tests/configs/common/common.yaml
himani2411: Hi,

Yes I have tested the feature with the other OSSes.  I wanted to test the assertion that I [mentioned](https://github.com/aws/aws-parallelcluster/pull/5713#discussion_r1434436840) where the default behaviour has not changed with IMDS tests, though I will add that in another PR.

tests/integration-tests/tests/common/assertions.py
himani2411: It asks for password for the user with a message which differs among the OSSes. Will check if it asks for password.

tests/integration-tests/tests/common/utils.py
himani2411: Thanks. Added

tests/integration-tests/tests/create/test_create.py
```diff
@@ -117,6 +119,27 @@
...
+    disable_sudo_access_default_user = True
+    cluster_config = pcluster_config_reader(disable_sudo_access_default_user=disable_sudo_access_default_user)
+    cluster = clusters_factory(cluster_config)
+
+    logging.info("Checking default user has disabled sudo access after cluster creation")
+    assert_head_node_is_running(region, cluster)
+    for node_type in NodeType:
+        assert_default_user_has_desired_sudo_access(
+            cluster, node_type, scheduler, disable_sudo_access_default_user, region, scheduler_commands_factory
+        )
```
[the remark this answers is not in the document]
himani2411: Removed

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2073,6 +2075,11 @@ def get_instance_types_data(self) -> dict:
         """Get instance type infos for all instance types used in the configuration file."""
         return {}
 
+    @property
+    def is_default_user_sudo_access_enabled(self):
+        """Return True if DisableSudoAccessForDefaultUser is True."""
+        return self.disable_sudo_access_default_user if self.disable_sudo_access_default_user else False
```
hanwen-pcluste, line 2081: Could be `return bool(self.disable_sudo_access_default_user)`

hanwen-pcluste: Why `is_default_user_sudo_access_enabled` and `disable_sudo_access_default_user` have the same boolean value?

cli/src/pcluster/resources/compute_node/user_data.sh
```diff
@@ -35,6 +35,10 @@ export NO_PROXY="localhost,127.0.0.1,169.254.169.254"
 PROXY
 fi
 
+if [ "${OSUser}" == "rocky" ]; then
```
hanwen-pcluste: Do we have to check the operating systems?
Can we just remove the `ALL=(ALL) NOPASSWD:ALL` from `/etc/sudoers` on any OSes?
Same question for other `user_data.sh`

hanwen-pcluste: Moreover, can we make this action only when `disable_sudo_access_default_user` is true?

tests/integration-tests/tests/common/utils.py
```diff
@@ -272,13 +272,45 @@ def reboot_head_node(cluster, remote_command_executor=None):
...
+        WaiterConfig={"Delay": 60, "MaxAttempts": 5},
+    )
+
+
+def wait_login_node_status_ok(cluster):
+    logging.info(f"Waiting for login node's Status to be Ok for cluster: {cluster.name}")
+    boto3.client("ec2", region_name=cluster.region).get_waiter("instance_status_ok").wait(
+        InstanceIds=cluster.get_cluster_instance_ids(node_type="LoginNode"),
+        WaiterConfig={"Delay": 60, "MaxAttempts": 5},
+    )
```
hanwen-pcluste, line 311: Minor: these two functions can be merged

cli/src/pcluster/resources/compute_node/user_data.sh
himani2411: > Moreover, can we make this action only when disable_sudo_access_default_user is true?
Will do
> Do we have to check the operating systems?
None of the other OS has that policy its just rocky

hanwen-pcluste: What if a custom AMI is used?

himani2411: If the base Ami has the policy then it would be required to be removed. Maybe removing the OS check would be okay as the command would work only if the policy string doesn’t exist

cli/tests/pcluster/config/test_cluster_config.py
```diff
@@ -327,6 +327,7 @@ def base_cluster_config(self):
         return BaseClusterConfig(
             cluster_name="clustername",
             image=Image("alinux2"),
+            disable_sudo_access_default_user=True,
```
hanwen-pcluste: Do we have to use this?
I believe we only need "required" parameters?

tests/integration-tests/tests/create/test_create/test_create_disable_sudo_access_for_default_user/pcluster.config.yaml
```diff
@@ -0,0 +1,41 @@
...
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Imds:
+    Secured: {{ imds_secured }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  {% if scheduler == "awsbatch" %}AwsBatchQueues:{% else %}SlurmQueues:{% endif %}
```
hanwen-pcluste: I believe only "SlurmQueues"  is needed, right?

himani2411: Thanks this also makes me realise I need to remove it from other YAMLs

cli/tests/pcluster/config/test_cluster_config.py
himani2411: Not really I think we can skip this

hanwen-pcluste (approved):
Great work!

PR #5722, opened 2023-09-26
Add integration test for scaling to 1000+ static nodes from an ODCR

EddyMM:
### Description of changes

- This integration test checks the following:
    - A cluster scales up to `1600` static nodes with the instances coming from a targeted ODCRs within specific time range and without over-scaling
    - Configuration of a cluster to use custom partitions without impacting scaling operations

- Steps
    - Cluster starts without no static nodes but targets a specific ODCR
    - Cluster is force updated to use 1600 static nodes
        - A check is done to confirm that all instances launched are from the ODCR and within a specific range of time (7 mins for now)
    - The cluster is then configured to have custom partitions
        - Simple job is submitted to the default partition and checked if it ran successfully
    - Cluster is then down sized to `1570` with force update and successful job execution is tested again
    - Cluster is then scaled up to `1600` again (this time the cluster has custom partitions configured);
        - A check is done to confirm that all instances launched are from the ODCR
        - A job is also submitted and checked if it was correctly executed

### Tests
* Ran the test in relevant accounts

### References
* N/A

### Checklist


tests/integration-tests/conftest.py
```diff
@@ -1193,17 +1193,119 @@ def placement_group_stack(cfn_stacks_factory, request, region):
...
+        scaling_odcr_b = ec2.CapacityReservation(
+            "integTestsScalingOdcrB",
+            AvailabilityZone=default_public_az,
+            InstanceCount=instances_count - first_odcr_instances_count,
+            InstancePlatform="Linux/UNIX",
+            InstanceType="c5.large",
+            InstanceMatchCriteria="targeted",
+        )
+
+        scaling_odcr_group = resourcegroups.Group(
```
NSsirena, line 1248: Mostly for my understanding: why are we splitting the ODCR in two reservations ?
There is a limit of some kind or there is a requirement related to the test ?

They seem to be equal in all respect.

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -158,8 +163,172 @@ def test_job_level_scaling(
     _submit_job_full_capacity(scheduler_commands, remote_command_executor)
 
 
+@pytest.mark.usefixtures("os", "instance", "scheduler")
+def test_scaling_special_cases(
+    region, pcluster_config_reader, clusters_factory, scaling_odcr_stack, scheduler_commands_factory, test_datadir
+):
+    full_cluster_size = 1600  # no of nodes after scaling the cluster up
+    downscaled_cluster_size = 1570  # no of nodes after scaling down the cluster
+    max_scaling_time = 7  # Chosen based on running the test multiple times
```
NSsirena: Can you add a comment section where you provide more information on 
* which "special cases" are we testing and why
* why we choose these values ? 1600 vs 1570 ?

Can we test the same condition with different values ?
Are those strictly required ?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -158,8 +163,172 @@ def test_job_level_scaling(
...
+        + "CapacityReservationSpecification.CapacityReservationTarget.CapacityReservationResourceGroupArn=="
+        + f"`{odcr_resource_group_arn}`"
+        + "].InstanceId[]"
+    )
+    instances_outside_odcr = [instance for instance in instances_outside_odcr_iter]
+    logging.info(f"Instances launched outside ODCR {odcr_resource_group_arn}: {instances_outside_odcr}")
+    assert_that(instances_outside_odcr).is_length(0)
+
+
+def _assert_cluster_update_scaling(
```
NSsirena: This can probably be split in two functions
* `_monitor_cluster_update` : that will call the cluster update and collect the time series , returning `ec2_capacity_time_series, compute_nodes_time_series, timestamps`
* `_check_scaling_results` : that will pick those time series and some conditions and will validate that the results are the ones we expected.

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -158,8 +163,172 @@ def test_job_level_scaling(
...
+    _assert_scaling_works(
+        ec2_capacity_time_series=ec2_capacity_time_series,
+        compute_nodes_time_series=compute_nodes_time_series,
+        expected_ec2_capacity=expected_ec2_capacity,
+        expected_compute_nodes=expected_compute_nodes,
+        min_for_scaledown=is_scale_down,
+    )
+
+
 def _assert_scaling_works(
```
NSsirena: This name is not very clear ...see my previous comment.
Maybe we can rename it `[verify|check]_scaling_[results|expectations]` ..or something similar ?

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -158,8 +163,172 @@ def test_job_level_scaling(
...
+        expected_ec2_capacity=(downscaled_cluster_size, full_cluster_size),
+        expected_compute_nodes=(downscaled_cluster_size, full_cluster_size),
+        max_monitoring_time=minutes(max_scaling_time),
+        is_scale_down=False,
+    )
+    _assert_compute_nodes_in_cluster_are_from_odcr(cluster, region, resource_group_arn)
+    _assert_simple_job_succeeds(scheduler_commands, full_cluster_size, partition="q1")
+
+
+def _assert_simple_job_succeeds(
```
NSsirena: This seems like a piece of code that we may want to reuse many times.

If we do not already have it, we may consider adding a method in `SlurmCommands` to submit 1+ "simple jobs" (I'd name them `simple_srun_job` or `simple_sbatch_job` ) and then have a function (in scheduler_commands or SlurmCommands) to launch and test the results.

tests/integration-tests/tests/scaling/test_scaling/test_scaling_special_cases/include_custom_partition_large.sh.jinja
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+set -e
```
NSsirena, line 3: This is very cryptic.

Can you add some comments here to explain which are the "design decisions" behind these custom partitions ?

Can we somehow "code" them ?
If we have to change something in the future ... manually checking if a specific ID in [1:1600] is in one partition or in another will be difficult

It is important to understand:
* what we want to test
* how many custom partitions do we need to be able to test it
* which are the required characteristics of each partition
* the mapping between each partition and a specific condition we want to check

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -175,75 +181,90 @@ def test_scaling_special_cases(
...
-        region,
-        cluster_config=str(upscale_cluster_config),
-        expected_ec2_capacity=(downscaled_cluster_size, full_cluster_size),
-        expected_compute_nodes=(downscaled_cluster_size, full_cluster_size),
-        max_monitoring_time=minutes(max_scaling_time),
-        is_scale_down=False,
-    )
-    _assert_compute_nodes_in_cluster_are_from_odcr(cluster, region, resource_group_arn)
-    _assert_simple_job_succeeds(scheduler_commands, full_cluster_size, partition="q1")
+    for launch_api in ["run_instances", "create_fleet"]:
```
NSsirena, line 206: Since we are going to create a new cluster anyway if we parameterize the test maybe we can speed up the test. Not sure if they may be executed in parallel

tests/integration-tests/tests/scaling/test_scaling.py
```diff
@@ -175,75 +181,90 @@ def test_scaling_special_cases(
...
+        scheduler_commands = scheduler_commands_factory(remote_command_executor)
+
+        upscale_cluster_config = pcluster_config_reader(
+            config_file="pcluster-upscale.config.yaml",
+            target_capacity_reservation_arn=resource_group_arn,
+            full_cluster_size=full_cluster_size,
+            launch_api=launch_api,
+        )
+
+        # Scale up cluster
```
NSsirena, line 224: Here and in the following occurrences... can we replace these comments with some more "validation oriented" descriptions ? Like

```
# Changing the cluster configuration to increase the number of static nodes from 0 to full_cluster_size 
# to verify that our logic is able to successfully complete the scale up in less than max_scaling_time minutes
```

and 

```
# Changing Slurm config file to create Custom Partitions because (??)
```
and
```
# Reducing the number of static nodes to YYY to validate (what ?)
```

tests/integration-tests/conftest.py
EddyMM: A single ODCR has a limit of 1000 instances so to support a test requiring more than that we need 2 ODCRs. The splitting was done mainly to distribute the needed capacity across the ODCRs.

An alternative could be to divide the ODCRs into "bins" of 1000 with the last ODCR potentially having <1000. I mostly went for splitting since we're not testing >2000 instances for now. It could be an improvement for later if needed.

I can also add a comment about the 1000 instances limit.

tests/integration-tests/tests/scaling/test_scaling.py
EddyMM: Sure, I can add a description of the special cases being tested.

> Can we test the same condition with different values ?

Not really, for the `1570`, the custom partitions refer to nodes up to `1565`. We want to avoid down scaling to a point where the custom partitions refer to nodes that no longer exist. This will result in slurmctld restart failing. I'll add a comment on this since it's relevant.

tests/integration-tests/tests/scaling/test_scaling.py
EddyMM: It should be possible. I'll try to add this minor refactor.

tests/integration-tests/tests/scaling/test_scaling/test_scaling_special_cases/include_custom_partition_large.sh.jinja
EddyMM: Let's sync on this.

tests/integration-tests/tests/scaling/test_scaling.py
EddyMM: We want to reuse the ODCR. Using parametrize would create the ODCR stack again.
Also, this test deploys a lot of resources so parallelism may come at the risk of hitting account quota limits easily.

tests/integration-tests/tests/scaling/test_scaling.py
EddyMM: I agree, it crossed my mind when writing it. The concern I had though is the nature of "a simple job". The nature of a simple job may vary depending on the test. Also should the job be executed on all nodes or a specific partition?

If I parameterise the command, we end up with a function that already exists i.e. a function to submit a job and wait for it to succeed.

---

What we can have is a `submit_and_assert_job_succeeded` function that can be used by any test (and it's up to the test to pass its idea of a simple job)

NSsirena (approved):

PR #5725, opened 2023-09-28
[develop] Add IAM policies for checking LoginNodes stopped script inside HeadNode

hehe7318:
### Description of changes
* Add IAM policies to HeadNode for checking LoginNodes status script inside HeadNode

### Checklist


cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -742,6 +742,17 @@ def _build_policy(self) -> List[iam.PolicyStatement]:
...
+                        sid="ElasticLoadBalancingDescribe",
+                        actions=[
+                            "elasticloadbalancing:DescribeLoadBalancers",
+                            "elasticloadbalancing:DescribeTags",
+                            "elasticloadbalancing:DescribeTargetGroups",
+                            "elasticloadbalancing:DescribeTargetHealth",
+                        ],
+                        effect=iam.Effect.ALLOW,
+                        resources=["*"],
+                    ),
```
jdeamicis: I would say this is the _wrong place_ for these policies. We should add these policies only if the custom munge key is used **and** login nodes are present.

Below in this very file there are a few examples of conditional policies for directory service and Slurm accounting (see line 774)..

jdeamicis (changes requested):

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Makes sense. Thank you Jacopo for your reminding. Done! Now the policies will be added only if LoginNodes and Custom Munge Key are used.

jdeamicis (approved):
Great job!

PR #5728, opened 2023-09-29
[Rocky8] Add rocky8 in Daily Integration tests

himani2411:
### Description of changes
* Add Rocky8 for daily Integration tests
* Update the Image Name for Rocky as per our Image Build
### Checklist

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -4,13 +4,14 @@ test-suites:
 {{- common.OSS_COMMERCIAL_ARM.append("centos7") or "" -}}
 {% filter indent(2) %}
 {% include 'common/common.yaml' %}
+{% include 'common/not_released_osses.yaml' %} # FIXME: Overrides sections from common.yaml for Rocky8
 {% endfilter %}
   efa:
     test_efa.py::test_efa:
       dimensions:
         - regions: ["euw1-az1"]  # do not move, unless capacity reservation is moved as well
           instances: ["c5n.18xlarge"]
-          oss: ["alinux2"]
+          oss: ["rocky8"]
           schedulers: ["slurm"]
         - regions: ["use1-az6"]   # do not move, unless capacity reservation is moved as well
           instances: ["p4d.24xlarge"]
@@ -43,7 +44,7 @@ test-suites:
       dimensions:
         - regions: ["sa-east-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux2"]
+          oss: ["rocky8"]
           schedulers: ["slurm"]
     test_api.py::test_cluster_awsbatch:
       dimensions:
@@ -93,7 +94,7 @@ test-suites:
           schedulers: ["slurm"]
         - regions: ["us-gov-east-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["rhel8"]
+          oss: ["rocky8"]
           schedulers: ["slurm"]
   trainium:
     test_trainium.py::test_trainium:
@@ -114,7 +115,7 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
     test_cluster_custom_resource.py::test_cluster_update:
       dimensions:
-        - oss: ["alinux2"]
+        - oss: ["rocky8"]
           regions: ["us-east-2"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
     test_cluster_custom_resource.py::test_cluster_update_invalid:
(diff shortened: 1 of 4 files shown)


tests/integration-tests/configs/common/common.yaml
```diff
@@ -56,7 +56,7 @@ cloudwatch_logging:
       # 2) run the test for all OSes with slurm
       - regions: ["ap-east-1"]
         instances: {{ common.INSTANCES_DEFAULT_ARM }}
-        oss: ["ubuntu2204", "rhel8"]
+        oss: ["ubuntu2204", "rocky8"]
```
enrico-usai: There is a problem with this change.

We cannot add rocky8 in common, because this test will also be used to test released version and we're not sure we'll release official pcluster rocky8 images, so these tests will fail.

If you need to enable these tests, you need to "override" them in develop.yaml

tests/integration-tests/tests/createami/test_createami/test_build_image/image.config.yaml
```diff
@@ -23,3 +23,7 @@ DeploymentSettings:
         - {{ private_subnet_id }}
         SecurityGroupIds:
         - {{ default_vpc_security_group_id }}
+
+DevSettings:
+    Cookbook:
+        ChefCookbook: https://github.com/himani2411/aws-parallelcluster-cookbook/tarball/wip/rocky/
```
enrico-usai: we shouldn't point to your remote, let's use develop instead

himani2411: Ohh I made these changes for testing. I forgot I had a PR open

tests/integration-tests/configs/common/common.yaml
himani2411: Sure

tests/integration-tests/configs/develop.yaml
```diff
@@ -4,13 +4,14 @@ test-suites:
...
 {% include 'common/common.yaml' %}
+{% include 'common/rocky8.yaml' %} # FIXME: Overrides sections from common.yaml for Rocky8
 {% endfilter %}
   efa:
     test_efa.py::test_efa:
       dimensions:
         - regions: ["euw1-az1"]  # do not move, unless capacity reservation is moved as well
           instances: ["c5n.18xlarge"]
-          oss: ["alinux2"]
+          oss: ["rocky8"]
```
enrico-usai, line 14: we cannot change test_efa because we have OSU baseline results to check performances only for alinux2 at the moment

tests/integration-tests/configs/develop.yaml
```diff
@@ -134,7 +135,7 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
     test_cluster_custom_resource.py::test_cluster_delete_retain:
       dimensions:
-        - oss: ["alinux2"]
+        - oss: ["rocky8"]
```
enrico-usai, line 138: Not related to your changes but it would be good to modify all the existing custom_resource tests to run with different oses. Previously they ran just with alinux2, now you correctly added rocky8, in future we should enable other oses too.

tests/integration-tests/configs/common/rocky8.yaml
```diff
@@ -0,0 +1,578 @@
+# FIXME: Remove this YAML once we officially release Rocky8 AMI
+# This duplicates some sections of common.yaml to include Rocky8 AMI in daily Integ tests
```
enrico-usai: In future we can think to do it more generic, calling it `not_released_oses.yaml` and define a new variable like `NOT_RELEASED_OSES`
```
{%- set NOT_RELEASED_OSES = ["rocky8"] -%}
```

tests/integration-tests/configs/develop.yaml
himani2411: The Baseline performance check is compared using the instance type like c5n.18xlarge

tests/integration-tests/configs/common/rocky8.yaml
himani2411: Added the same in this commit itself. However, It would be better if we manually change the `not_released_oses.yaml`  as we dont want to increase the number of tests. This file overrides the tests in `common.yaml` where I have replaced a few existing OS with rhel8.

tests/integration-tests/configs/develop.yaml
```diff
@@ -124,21 +125,21 @@ test-suites:
...
           regions: ["us-east-2"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
     test_cluster_custom_resource.py::test_cluster_create_with_custom_policies:
       dimensions:
-        - oss: ["alinux2"]
+        - oss: ["rhel8"]
```
dreambeyondorange, line 143: did you mean rocky8 here?

dreambeyondorange (commented):
How did you decide which tests to replace with rocky8?  It looks like alot of them were alinux2, was that on purpose?

tests/integration-tests/configs/develop.yaml
himani2411: This was to address the change Enrico Suggested. https://github.com/aws/aws-parallelcluster/pull/5728#discussion_r1345866850

himani2411:
> How did you decide which tests to replace with rocky8? It looks like alot of them were alinux2, was that on purpose?

It was random. That seems because the develop.yaml had majority of the tests using alinux2. However, `not_released_osses.yaml` was added to override the sections in common.yaml, which has a lot more OSSes compared to alinux2 and I have replaced other Osses too so as not to increase the number of tests.

dreambeyondorange (approved):

PR #5745, opened 2023-10-06
[Develop] Add the option to select an external /home filesystem for the cluster

dreambeyondorange:
### Tests
* Tested launching a cluster with /home mounted to EFS, FSx, and the default EBS head node root volume
* Tested launching a cluster with an external fs and not /home mounted on EFS, FSx and default EBS.

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2486

### Checklist


cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -1484,7 +1484,8 @@ def test_shared_storage_name_validator(name, expected_message):
         ("shared_ebs_1", None),
         ("shared", None),
         ("/shared", None),
-        ("home", "mount directory .* is reserved"),
+        ("home", None),
```
gmarciani, line 1487: Shouldn't this be `/home` rather than `home`?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
 - Add `Monitoring/Alarms/Enabled` parameter to toggle Amazon CloudWatch Alarms for the cluster.
 - Add head node alarms to monitor EC2 health checks, CPU utilization and the overall status of the head node.
 - Add the option to use EFS storage instead of NFS exports from the head node root volume for intra-cluster shared ParallelCluster, Intel, Slurm, and login node data.
+- Add the option to use EFS or FSx as external shared storage via the `SharedStorage` section of the config file for `/home`
```
gmarciani: I'd say "Allow to use" rather than "Add the option" because "Add the option" makes me think that we modified something in the cluster config schema.

dreambeyondorange: Ah good point

cli/tests/pcluster/validators/test_cluster_validators.py
dreambeyondorange: no it's not necessary, the validator adds the slash 
```
if not mount_dir.startswith("/"):
            mount_dir = "/" + mount_dir
```

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -892,7 +892,6 @@ def _validate(self, mount_dir: str):
             "/boot",
             "/dev",
             "/etc",
-            "/home",
```
gmarciani, line 895: In this way the user will be able to mount /home using EBS shared storage also.
Is this the desired behaviour?

gmarciani: Resolving this because according to https://github.com/aws/aws-parallelcluster-cookbook/pull/2486/files#r1350507989 we are going to support /home in shared EBS volume

gmarciani (approved):

cli/src/pcluster/validators/cluster_validators.py
dreambeyondorange: yea at first I thought it wouldn't be that useful to mount as EBS, but better to just provide the flexibility

PR #5732, opened 2023-10-04
[Head Node Reporting] Add head node alarms

gmarciani:
### Description of changes
1. Add the following new head node alarms: 
    1. EC2 Health Checks, 
    1. CPU Utilization
    1. a composite alarm including every head node alarm.
 1. Updated unit tests
 2. Updated integ tests, but added a TODO to evaluate replacement of such integ tests with unit/kitchen.

**Notes**
With reference to the feature design, the following items are out of scope for this PR:
1. alarms renaming
2. decouple alarms and dashboards
3. add alarms to dashboards

### Tests
* Manually verified creation of expected alarms
* Unit tests
* Integ tests: test_monitoring

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5747, opened 2023-10-09
[Head Node Reporting] Fix regression in test_monitoring causing cluster creation failure.

gmarciani:
### Description of changes
Fix regression in `test_monitoring` introduced by https://github.com/aws/aws-parallelcluster/pull/5732, that started causing cluster creation failure due to invalid cluster config.

The test was failing because the cluster config used by the test was expecting the variable `alarms_enabled` to be injected but it was not.

### Tests
Verified that cluster config is now valid in test_monitoring integ test.


enrico-usai:
Could you please describe in the description "why" it was failing? Why introducing this parameter is solving the issue?

enrico-usai (approved):

PR #5806, opened 2023-10-31
[develop] Rename LoginNodes Stack Target Group resource to specific schema

hehe7318:
### Description of changes
* Rename LoginNodes Stack Target Group resource to specific schema
* So that we can strict the IAM policy. Delete two related to the NLB and strict one related to TG.
* Also we will be able to modify the check_login_stopped.sh script to directly use TG name to query to reduce script running time

### Tests
* Manually tests done, passed successfully.

### References
* The cookbook modify the check_login_stopped.sh script to directly use TG name to query PR is [here](https://github.com/aws/aws-parallelcluster-cookbook/pull/2524).

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cdk_builder_utils.py
+++ b/cli/src/pcluster/templates/cdk_builder_utils.py
@@ -694,34 +694,16 @@ class HeadNodeIamResources(NodeIamResourcesBase):
             )
 
             if self._config.login_nodes:
-                target_group_name = _get_target_group_name(
-                    self._config.cluster_name,
-                    self._config.login_nodes.pools[0].name,
-                )
                 policy.extend(
                     [
                         iam.PolicyStatement(
                             sid="TargetGroupDescribe",
                             actions=[
                                 "elasticloadbalancing:DescribeTargetGroups",
-                            ],
-                            effect=iam.Effect.ALLOW,
-                            resources=["*"],
-                        ),
-                        iam.PolicyStatement(
-                            sid="TargetHealthDescribe",
-                            actions=[
                                 "elasticloadbalancing:DescribeTargetHealth",
                             ],
                             effect=iam.Effect.ALLOW,
-                            resources=[
-                                self._format_arn(
-                                    service="elasticloadbalancing",
-                                    resource=f"targetgroup/{target_group_name}/*",
-                                    region=Stack.of(self).region,
-                                    account=Stack.of(self).account,
-                                ),
-                            ],
+                            resources=["*"],
                         ),
                     ]
                 )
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -54,6 +54,15 @@
 PCLUSTER_LAMBDA_PREFIX = "pcluster-"
 
 
+def _get_target_group_name(cluster_name, pool_name):
```
EddyMM: - We can have some minimal unit tests for this function.
- This function seems to be relevant for the login nodes stack. We can keep it in the login nodes stack instead of having it in the `cdk_builder_utils` file which is used for reuseable code.
    - [nice-to-have] Alternatively make it generic e.g. the function accepts a list of strings and a pattern used to combine the strings to fit a desired length requirement.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -54,6 +54,15 @@
 PCLUSTER_LAMBDA_PREFIX = "pcluster-"
 
 
+def _get_resource_combination_name(resource_name_1, resource_name_2, partial_length, hash_length):
```
EddyMM: Seems restricted  to just 2 resource names, can we use something like this?

```
def _get_resource_combination_name(*resource_names, partial_length=6, hash_length=16):
  combined_name = "".join(resource_names)
  hash_value = sha256(combined_name.encode()).hexdigest()[:hash_length]
  prefix = "-".join(name[:partial_length] for name in resource_names)
  return f"{prefix}-{hash_value}"
```

Then call it like so ...
```
>>> _get_resource_combination_name("a", "b", partial_length=7, hash_length=16)
```

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Done!

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: That a brilliant point! Thanks Eddy!

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -54,6 +54,14 @@
 PCLUSTER_LAMBDA_PREFIX = "pcluster-"
 
 
+def _get_resource_combination_name(*resource_names, partial_length=7, hash_length=16):
```
EddyMM, line 57: NIT: We should remove the `_` since this is a public function, the underscore(`_`) is normally associated with a private function by convention. A more approriate name could also be `generate_fixed_size_resource_name`

EddyMM (approved):

PR #5817, opened 2023-11-03
[develop] Add support for Capacity Blocks for ML

enrico-usai:
### Description of changes
* Added support for `CapacityType=CAPACITY_BLOCK`.
  * To permit the usage of capacity block it's required to set `market_type="capacity-block"` in the Launch Template.
* Default value for `AllocationStrategy` is now set to `None` when using Capacity Blocks

#### Validators
* Extended `CapacityReservationValidator` to verify capacity-type vs reservation-type
  * CapacityReservation of type `capacity-block` have a new attribute `ReservationType`, we're using it to validate that configured `CapacityType` corresponds to it.
* Extended `CapacityReservationValidator` to have a single instance type when using capacity reservation id
   * Capacity Reservation is always associate to a single instance type.
* Extended `ComputeResourceSizeValidator` to validate min count when using CB
  * At the moment all the instances in Capacity Block reservations are managed as static instances, so min count must be equal to max count and this value should be > 0.
  * To perform this check we had to move ComputeResourceSizeValidator registration at queue level since capacity_type is specified at that level.
*Add new `CapacityReservationSizeValidator` to verify number of instances is not exceeded
  * In the cluster config we created a map to keep track of the number of instances associated through max-count to a given capacity reservation id. Then we're verifying that the number of instances is not exceeding the TotalInstanceCount.
  * This will be executed for every capacity reservation, but has a strong value when using Capacity Block reservation.
* Improve CapacityReservationValidator to always validate subnet
  * Previously the subnet was validated only when instance_type was set.

### Additional changes
* [Add missing requirement for test_lambda_handler unit test](https://github.com/aws/aws-parallelcluster/commit/481cccc966794928b07fd067b0acb8172e434dd5)
* Modify describe_capacity_reservations to return a list of CapacityBlockReservationInfo
  * From this moment we don't need to know the structure of the describe-capacity-reservations output in the entire code.  
  * We can use the `CapacityBlockReservationInfo` wrapper class that decouples boto3 structure from parallelcluster internal logic.
  * The code is now more robust and thanks to this change I found a bug in the code: the _validate_no_pg was returning no errors even if there were no capacity reservations without PG. Aligned the unit test for it.

### Tests
* Added unit tests to cover the new/modified code.

#### Manual test of validators

Instance not supporting capacity block: 
```
"type": "CapacityTypeValidator",
"message": "Usage type 'capacity-block' not supported with instance type 't3.large'"
```
Instance type not aligned with CR instance:
```
"type": "ComputeResourceLaunchTemplateValidator",
"message": "Unable to validate configuration parameters for instance type p5.48xlarge. Please double check your cluster configuration. Capacity Reservation's attribute does not match with requested instance parameter. Attribute: InstanceType, requested value: p5.48xlarge, expected value: t3.large"
...
"type": "CapacityReservationValidator",
"message": "Capacity reservation cr-0541d9b022c98faaa must have the same instance type as p5.48xlarge."
```

More instances in the same compute resource:
```
"type": "CapacityReservationSizeValidator",
"message": "Number of instances configured for Capacity Reservation cr-01d2ca74c0d1f1705: 4 is exceeding the total instances count available for the Capacity Reservation: 2"
```

more instances from multiple queues/compute resources, one with 2 and one with 1 from same CB
```
"type": "CapacityReservationSizeValidator",
"message": "Number of instances configured for Capacity Reservation cr-01d2ca74c0d1f1705: 3 is exceeding the total instances count available for the Capacity Reservation: 2"
```

Wrong size: MinCount: 0, MaxCount: 1
```
"type": "ComputeResourceSizeValidator",
"message": "Max count must be set to the same value of min count when using Capacity Block reservation."
...
"type": "ComputeResourceSizeValidator",
"message": "Min count must be a value > 0 when using Capacity Block reservation."
```

### References
* Cookbook changes: https://github.com/aws/aws-parallelcluster-cookbook/pull/2526
* Related patch for the node: https://github.com/aws/aws-parallelcluster-node/pull/591
* Slurm reservations: https://slurm.schedmd.com/reservations.html
* Capacity Blocks: https://docs.aws.amazon.com/en_us/AWSEC2/latest/UserGuide/ec2-capacity-blocks.html

### DIFF
--- a/cli/tests/pcluster/aws/test_ec2.py
+++ b/cli/tests/pcluster/aws/test_ec2.py
@@ -524,7 +524,7 @@ def get_describe_capacity_reservation_mocked_request(capacity_reservations, stat
                 for capacity_reservation in capacity_reservations
             ]
         },
-        expected_params={"CapacityReservationIds": capacity_reservations},  # TODO add: "ReservationType": None},
+        expected_params={"CapacityReservationIds": capacity_reservations},
     )
 
 
--- a/cli/tests/pcluster/aws/test_ec2.py
+++ b/cli/tests/pcluster/aws/test_ec2.py
@@ -524,7 +524,7 @@ def get_describe_capacity_reservation_mocked_request(capacity_reservations, stat
                 for capacity_reservation in capacity_reservations
             ]
         },
-        expected_params={"CapacityReservationIds": capacity_reservations},
+        expected_params={"CapacityReservationIds": capacity_reservations},  # TODO add: "ReservationType": None},
     )
 
 
@@ -540,7 +540,7 @@ def test_describe_capacity_reservations_cache(boto3_stubber):
         get_describe_capacity_reservation_mocked_request([capacity_reservation], "active"),
     ]
     boto3_stubber("ec2", mocked_requests)
-    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0]["State"]).is_equal_to(
+    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0].state()).is_equal_to(
         "pending"
     )
 
@@ -551,13 +551,13 @@ def test_describe_capacity_reservations_cache(boto3_stubber):
     assert_that(response).is_length(2)
 
     # Third boto3 call. The result should be from cache even if the state of the cr is different
-    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0]["State"]).is_equal_to(
+    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0].state()).is_equal_to(
         "pending"
     )
 
     # Fourth boto3 call after resetting the AWSApi instance. The latest cr state should be retrieved from boto3
     AWSApi.reset()
-    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0]["State"]).is_equal_to(
+    assert_that(AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation])[0].state()).is_equal_to(
         "active"
     )
 
--- a/cli/tests/pcluster/aws/test_ec2.py
+++ b/cli/tests/pcluster/aws/test_ec2.py
@@ -15,7 +15,7 @@ import pytest
 from assertpy import assert_that, soft_assertions
 
 from pcluster.aws.aws_api import AWSApi
-from pcluster.aws.aws_resources import ImageInfo, InstanceTypeInfo
+from pcluster.aws.aws_resources import CapacityReservationInfo, ImageInfo, InstanceTypeInfo
 from pcluster.aws.common import AWSClientError
 from pcluster.aws.ec2 import Ec2Client
 from pcluster.config.cluster_config import AmiSearchFilters, Tag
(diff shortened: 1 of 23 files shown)


lukeseawalker (dismissed):

cli/tests/pcluster/validators/test_ec2_validators.py
```diff
@@ -708,6 +713,29 @@ def test_capacity_reservation_validator(
     [
         (CapacityReservationInfo({"TotalInstanceCount": 1}), 1, []),
         (CapacityReservationInfo({"TotalInstanceCount": 2}), 1, []),
+        (
+            CapacityReservationInfo(
+                {
+                    "TotalInstanceCount": 0,
+                    "Tags": [
```
lukeseawalker, line 720: please add a test with "Tags" but with no CAPACITY_BLOCK_REQUESTED_QUANTITY_TAG_KEY "Key"

cli/tests/pcluster/aws/test_ec2.py
```diff
@@ -524,7 +524,7 @@ def get_describe_capacity_reservation_mocked_request(capacity_reservations, stat
                 for capacity_reservation in capacity_reservations
             ]
         },
-        expected_params={"CapacityReservationIds": capacity_reservations},
+        expected_params={"CapacityReservationIds": capacity_reservations},  # TODO add: "ReservationType": None},
```
lukeseawalker: missing TODO

cli/tests/pcluster/validators/test_all_validators/test_slurm_validators_are_called_with_correct_argument/slurm.yaml
```diff
@@ -52,11 +53,31 @@ Scheduling:
...
+          InstanceType: t2.xlarge
+          MinCount: 5
+          MaxCount: 5
+          CapacityReservationTarget:
+            CapacityReservationId: "cr-12345"
+        - Name: compute_resource2
+          InstanceType: t2.xlarge
+          MinCount: 3
+          MaxCount: 3
+          CapacityReservationTarget:
```
lukeseawalker, line 79: what about adding a "CapacityReservationTarget" at queue level? So to check validation are correctly called for that case too

cli/tests/pcluster/validators/test_instances_validators.py
```diff
@@ -476,25 +476,48 @@ def test_instances_networking_validator(
 
 
 @pytest.mark.parametrize(
-    "compute_resource_name, capacity_type, allocation_strategy, expected_message",
+    "capacity_type, allocation_strategy, expected_message",
     [
         # OnDemand Capacity type only supports "lowest-price" allocation strategy
```
lukeseawalker, line 481: why the comment for SPOT was removed?

cli/tests/pcluster/validators/test_instances_validators.py
```diff
@@ -476,25 +476,48 @@ def test_instances_networking_validator(
...
+        (
+            CapacityType.ONDEMAND,
+            None,
+            "Compute Resource TestComputeResource is using an OnDemand CapacityType but the Allocation Strategy "
+            "specified is not set. OnDemand CapacityType can only use 'lowest-price' allocation strategy.",
+        ),
+        # Spot Capacity type supports both "lowest-price" and "capacity-optimized" allocation strategy
+        (CapacityType.SPOT, AllocationStrategy.LOWEST_PRICE, ""),
+        (CapacityType.SPOT, AllocationStrategy.CAPACITY_OPTIMIZED, ""),
+        (CapacityType.SPOT, None, ""),
```
lukeseawalker, line 493: why this combination isn't generating a message like in the case for CapacityType.ONDEMAND / None ?

cli/tests/pcluster/validators/test_instances_validators.py
lukeseawalker: OK I see it now that it was moved

cli/tests/pcluster/validators/test_instances_validators.py
enrico-usai: Modified validator to not show the message in the on-demand case when the value is None.

This is an internal change only , because allocations-strategy has always a default value.

cli/tests/pcluster/validators/test_all_validators/test_slurm_validators_are_called_with_correct_argument/slurm.yaml
enrico-usai: done, moved it.

cli/tests/pcluster/aws/test_ec2.py
enrico-usai: removed. This is not implemented on ec2 API.

cli/tests/pcluster/validators/test_ec2_validators.py
enrico-usai: added and added more unit tests for CapacityReservationInfo methods.

lukeseawalker (approved):

PR #5834, opened 2023-11-09
[develop] Automatically retrieve instance type from capacity reservation

enrico-usai:
### Description of changes
Starting from now `InstanceType` is no longer a required configuration parameter.
This can be automatically retrieved from `CapacityReservationTarget/CapacityReservationId`
when it's specified at Compute Resource level.

### Tests
* Extended unit tests accordingly
* Manual test of cluster creation with and without instance type

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2215,11 +2215,7 @@ class FlexibleInstanceType(Resource):
 class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with Multiple Instance Types."""
 
-    def __init__(
-        self,
-        instances: List[FlexibleInstanceType],
-        **kwargs,
-    ):
+    def __init__(self, instances: List[FlexibleInstanceType], **kwargs):
         super().__init__(**kwargs)
         self.instances = Resource.init_param(instances)
 
@@ -2271,13 +2267,10 @@ class SlurmFlexibleComputeResource(_BaseSlurmComputeResource):
 class SlurmComputeResource(_BaseSlurmComputeResource):
     """Represents a Slurm Compute Resource with a Single Instance Type."""
 
-    def __init__(
-        self,
-        instance_type,
-        **kwargs,
-    ):
+    def __init__(self, instance_type=None, **kwargs):
         super().__init__(**kwargs)
-        self.instance_type = Resource.init_param(instance_type)
+        _instance_type = instance_type if instance_type else self._instance_type_from_capacity_reservation()
+        self.instance_type = Resource.init_param(_instance_type)
         self.__instance_type_info = None
 
     @property
@@ -2320,6 +2313,18 @@ class SlurmComputeResource(_BaseSlurmComputeResource):
     def disable_simultaneous_multithreading_manually(self) -> bool:
         """Return true if simultaneous multithreading must be disabled with a cookbook script."""
         return self.disable_simultaneous_multithreading and self._instance_type_info.default_threads_per_core() > 1
+
+    def _instance_type_from_capacity_reservation(self):
+        """Return the instance type from the configured CapacityReservationId, if any."""
+        instance_type = None
+        capacity_reservation_id = (
+            self.capacity_reservation_target.capacity_reservation_id if self.capacity_reservation_target else None
+        )
+        if capacity_reservation_id:
+            capacity_reservations = AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation_id])
+            if capacity_reservations:
+                instance_type = capacity_reservations[0].instance_type()
+        return instance_type
 
 
 class _CommonQueue(BaseQueue):
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2425,7 +2425,7 @@ class _CommonQueue(BaseQueue):
             if isinstance(compute_resource, SlurmComputeResource):
                 self._register_validator(
                     EfaValidator,
-                    instance_type=compute_resource.instance_type,
+                    instance_type=compute_resource.instance_types[0],
                     efa_enabled=compute_resource.efa.enabled,
                     gdr_support=compute_resource.efa.gdr_support,
                     multiaz_enabled=self.multi_az_enabled,
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -2285,11 +2285,6 @@ class SlurmComputeResource(_BaseSlurmComputeResource):
         """List of instance types under this compute resource."""
         return [self.instance_type]
 
-    @property
-    def instance_type_info(self) -> InstanceTypeInfo:
-        """Return instance type information."""
-        return AWSApi.instance().ec2.get_instance_type_info(self.instance_type)
-
     def _register_validators(self, context: ValidatorContext = None):
         super()._register_validators(context)
         self._register_validator(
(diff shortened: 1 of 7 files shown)


lukeseawalker:
>Starting from now InstanceType is no longer a required configuration parameter.
This can be automatically retrieved from CapacityReservationTarget/CapacityReservationId
when it's specified at Compute Resource level.

what about when CapacityReservationTarget/CapacityReservationId is specified at Queue level? Is InstanceType still required?

enrico-usai:
> > Starting from now InstanceType is no longer a required configuration parameter.
> > This can be automatically retrieved from CapacityReservationTarget/CapacityReservationId
> > when it's specified at Compute Resource level.
> 
> what about when CapacityReservationTarget/CapacityReservationId is specified at Queue level? Is InstanceType still required?

yes, it's still required. This is a limitation of this feature. The code is able to detect the instance type from capacity reservation ONLY if it's specified at compute resource level. This because at compute resource level we have both instance-type and capacity-target info so it's easy to choose between them, but if it's not specified in the compute resource we need to have the capacity-target from the queue and this requires to pass the value to the compute resources, this is an impacting change we can avoid for now.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2324,7 +2312,19 @@ def _instance_type_info(self) -> InstanceTypeInfo:
     @property
     def disable_simultaneous_multithreading_manually(self) -> bool:
         """Return true if simultaneous multithreading must be disabled with a cookbook script."""
-        return self.disable_simultaneous_multithreading and self.instance_type_info.default_threads_per_core() > 1
+        return self.disable_simultaneous_multithreading and self._instance_type_info.default_threads_per_core() > 1
+
+    def _instance_id_from_capacity_target(self):
```
lukeseawalker: what about _instance_type_from_capacity_reservation?

and modify the docstring accordingly

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2324,7 +2312,19 @@ def _instance_type_info(self) -> InstanceTypeInfo:
...
+        """Return the instance id from the configured CapacityReservationId, if any."""
+        instance_type = None
+        capacity_reservation_id = (
+            self.capacity_reservation_target.capacity_reservation_id if self.capacity_reservation_target else None
+        )
+        if capacity_reservation_id:
+            capacity_reservations = AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation_id])
```
lukeseawalker, line 2324: is this call cached?

enrico-usai: Short answer: yes,

Long answer: for this call there is an internal caching mechanism (unit tested), that was already in place.

```
        missed_capacity_reservations = []
        for capacity_reservation_id in capacity_reservation_ids:
            cached_data = self.capacity_reservations_cache.get(capacity_reservation_id)
            if cached_data:
                result.append(cached_data)
            else:
                missed_capacity_reservations.append(capacity_reservation_id)
        if missed_capacity_reservations:
            response = list(
                self._paginate_results(
                    self._client.describe_capacity_reservations, CapacityReservationIds=missed_capacity_reservations
                )
            )
```
this because there are multiple `describe` calls with different capacity_reservation_ids.

cli/src/pcluster/config/cluster_config.py
enrico-usai: yeah, I'm really not sure why I called it in this way.

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -486,7 +487,33 @@ def test_scheduling_schema(mocker, config_dict, failure_message):
                     }
                 ],
             },
-            "A Compute Resource needs to specify either InstanceType or Instances.",
+            "A Compute Resource needs to specify InstanceTypes, InstanceType or CapacityReservationId.",
+        ),
+        # Failing to specify either InstanceType or Instances should return a validation error
```
lukeseawalker: should this comment moved to the next params?

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -486,7 +487,33 @@ def test_scheduling_schema(mocker, config_dict, failure_message):
...
+                "Name": "Flex-Queue",
+                "Networking": {"SubnetIds": ["subnet-12345678"]},
+                "AllocationStrategy": "lowest-price",
+                "ComputeResources": [
+                    {"Name": "compute_resource1", "CapacityReservationTarget": {"CapacityReservationId": "id"}}
+                ],
+            },
+            None,
+        ),
+        (
```
lukeseawalker, line 504: what if both InstanceType and CapacityReservationId are specified? Or if Instances and CapacityReservationId are specified?
Can you add a test for these conditions?

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1509,14 +1509,22 @@ class SlurmComputeResourceSchema(_ComputeResourceSchema):
...
+    def instance_type_declaration(self, data, **kwargs):
+        """Validate that 'instance_type' is specified, directly or through capacity reservation id."""
+        if not data.get("instances"):
+            capacity_reservation_target = data.get("capacity_reservation_target")
+            capacity_reservation_id = (
+                capacity_reservation_target.capacity_reservation_id if capacity_reservation_target else None
+            )
+            if not data.get("instance_type") and not capacity_reservation_id:
+                raise ValidationError(
+                    "A Compute Resource needs to specify InstanceTypes, InstanceType or CapacityReservationId."
```
lukeseawalker: there is no InstanceTypes, but rather Instances/InstanceType

cli/src/pcluster/config/cluster_config.py
enrico-usai: updated and added unit tests for it.

cli/tests/pcluster/schemas/test_cluster_schema.py
enrico-usai: Added

cli/src/pcluster/schemas/cluster_schema.py
enrico-usai: nice catch!

cli/tests/pcluster/schemas/test_cluster_schema.py
```diff
@@ -486,7 +487,33 @@ def test_scheduling_schema(mocker, config_dict, failure_message):
...
-            "A Compute Resource needs to specify either InstanceType or Instances.",
+            "A Compute Resource needs to specify Instances, InstanceType or CapacityReservationId.",
+        ),
+        (
+            {
+                "Name": "Flex-Queue",
+                "Networking": {"SubnetIds": ["subnet-12345678"]},
+                "AllocationStrategy": "lowest-price",
+                "ComputeResources": [
+                    {"Name": "compute_resource1", "CapacityReservationTarget": {"CapacityReservationId": "id"}}
```
lukeseawalker, line 498: can you add a case when CapacityReservationTarget/CapacityReservationResourceGroupArn is specified in place of CapacityReservationTarget/CapacityReservationId ?

enrico-usai: it's already there:

```
{
                "Name": "Flex-Queue",
                "Networking": {"SubnetIds": ["subnet-12345678"]},
                "AllocationStrategy": "lowest-price",
                "ComputeResources": [
                    {
                        "Name": "compute_resource1",
                        "CapacityReservationTarget": {"CapacityReservationResourceGroupArn": "arn"},
                    }
                ],
            },
            "A Compute Resource needs to specify Instances, InstanceType or CapacityReservationId.",
```

lukeseawalker (approved):

PR #5837, opened 2023-11-10
[develop][integ-test] Add GetInstanceProfile permission for integration test permission boundary

hanwen-pcluste:
test_iam_resource_prefix started to fail a few days ago. It seems CloudFormation started to require GetInstanceProfile permission while creating an instance profile. Moreover, the action is ignores the path prefix. Therefore, this commit adds `iam:GetInstanceProfile` and relax the resource specification.

### Tests
test_iam_resource_prefix has been passed

### Checklist

### DIFF
--- a/tests/integration-tests/tests/iam/test_iam.py
+++ b/tests/integration-tests/tests/iam/test_iam.py
@@ -472,15 +472,13 @@ def _create_permission_boundary(permission_boundary_name):
                         "iam:AddRoleToInstanceProfile",
                         "iam:TagInstanceProfile",
                         "iam:UntagInstanceProfile",
+                        "iam:GetInstanceProfile",
                     ],
                     "Effect": "Allow",
                     "Resource": [
                         {"Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/${CustomIamPathPrefix}/*"},
                         {"Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/${CustomIamNamePrefix}*"},
-                        {
-                            "Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:instance-profile/"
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/iam/test_iam.py
```diff
@@ -472,15 +472,13 @@ def _create_permission_boundary(permission_boundary_name):
...
                     ],
                     "Effect": "Allow",
                     "Resource": [
                         {"Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/${CustomIamPathPrefix}/*"},
                         {"Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/${CustomIamNamePrefix}*"},
-                        {
-                            "Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:instance-profile/"
-                            "${CustomIamPathPrefix}/*"
-                        },
+                        {"Fn::Sub": "arn:${AWS::Partition}:iam::${AWS::AccountId}:instance-profile/*"},
```
himani2411, line 481: why are we relaxing the Resource specification? IIRC, We do add ResourcePrefix to the instance-profile as well

himani2411 (approved):

tests/integration-tests/tests/iam/test_iam.py
hanwen-pcluste: GetInstanceProfile requires to relax the Resource spec. I don't understand why. The best we can do is to split the permission to only relax the Resource for GetInstanceProfile, not for other actions.

gmarciani (approved):
In this PR we are introducing a new action `iam:GetInstanceProfile` that was not required until now and we are relaxing the permission removing the custom prefix from the resource. Let's be sure to understand why this became necessary. Approved because this is a readonly permission and the change only impacts test code.

PR #5862, opened 2023-11-17
[develop] Distinguish standard and flexible compute resource when validating ODCRs

enrico-usai:
`CapacityReservationId` is not supported by create-fleet (`FlexibleComputeResource`), independently by the number of instances.

### References

Issue introduced with: 2f25a09183bf7cff51bfbb185ab3909ac19f0761

### DIFF
--- a/cli/src/pcluster/validators/ec2_validators.py
+++ b/cli/src/pcluster/validators/ec2_validators.py
@@ -294,24 +294,31 @@ class CapacityReservationValidator(Validator):
         return capacity_type_to_reservation_type_class_map.get(capacity_type.value, capacity_type.value.lower())
 
     def _validate(
-        self, capacity_reservation_id: str, instance_types: List[str], subnet: str, capacity_type: CapacityType
+        self,
+        capacity_reservation_id: str,
+        instance_types: List[str],
+        is_flexible: bool,
+        subnet: str,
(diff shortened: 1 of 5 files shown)


cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -294,27 +294,42 @@ def _get_reservation_type_from_capacity_type(capacity_type: CapacityType):
...
-                        "A single instance type must be specified when using "
-                        f"Capacity reservation id: {capacity_reservation_id}"
+                        "CapacityReservationId parameter cannot be used with Instances parameter."
+                        "CapacityReservationId can only be used with the InstanceType parameter "
+                        "(https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html#yaml-"
+                        "Scheduling-SlurmQueues-ComputeResources-InstanceType)."
                     ),
                     FailureLevel.ERROR,
                 )
+                if len(instance_types) != 1:
```
lukeseawalker: remove, since doesn't make sense to suggest to set only 1 instance, since it's not supported

enrico-usai: done

lukeseawalker (approved):

PR #5874, opened 2023-11-22
[release-3.8] Use older versions of jsonschema and urllib3 for Lambda Layer

hanwen-pcluste:
Use older versions of jsonschema and urllib3 for Lambda Layer

Pcluster API execution fails if using the latest version of the packages.

* jsonschema introduced breaking changes in 4.18.0 in July 2023. Therefore, we use version 4.17.3. https://github.com/aws/aws-cdk/issues/26300
* urllib released 2.0 in May 2023. Pcluster API has been released with urllib 1.x. Therefore, we use version 1.x.

The root cause of both issues are because Lambda Function environment has older boto3/botocore versions than latest released versions on PyPi.
Specifically, when we prepare lambda layers, the dependencies are installed according to new boto3/botocore versions. When running the Lambda function, the older boto3/botocore are not compatible with the dependencies we provide in Lambda layers.

### Checklist

### DIFF
--- a/tests/integration-tests/conftest_resource_bucket.py
+++ b/tests/integration-tests/conftest_resource_bucket.py
@@ -40,7 +40,19 @@ def install_pc(basepath, pc_version):
     cli_dir = root / "cli"
     try:
         logger.info("installing ParallelCluster packages...")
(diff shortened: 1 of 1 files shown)


gmarciani (dismissed):

hanwen-pcluste:
Seems like Lambda Function is gradually upgrading their runtime. We may don't have to merge the PR

PR #5904, opened 2023-12-01
[Develop] Prevent updates to clusters that include changes to /home

dreambeyondorange:
Currently, PCluster does not support updates to a cluster's home directory.

### Tests
* Tested updates to clusters that have existing /home mounts in `SharedStorage` and those that do not have /home mounts

### Checklist

### DIFF
--- a/cli/src/pcluster/config/update_policy.py
+++ b/cli/src/pcluster/config/update_policy.py
@@ -227,7 +227,18 @@ def is_stop_required_for_shared_storage(change):
     return False
 
 
+def is_home_change(change):
+    if change.key == "SharedStorage":
+        if (change.old_value and change.old_value.get("MountDir") in ["home", "/home"]) or (
+            change.new_value and change.new_value.get("MountDir") in ["home", "/home"]
+        ):
+            return True
+    return False
+
+
 def fail_reason_shared_storage_update_policy(change, patch):
+    if is_home_change(change):
+        return "The /home directory cannot be changed during an update"
     if is_awsbatch_scheduler(change, patch):
         return f"Update actions are not currently supported for the '{change.key}' parameter"
     reason = "All login and compute nodes must be stopped"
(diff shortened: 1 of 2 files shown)


himani2411 (dismissed):

cli/src/pcluster/config/update_policy.py
```diff
@@ -227,7 +227,18 @@ def is_stop_required_for_shared_storage(change):
     return False
 
 
+def is_home_change(change):
+    if change.key == "SharedStorage":
+        if (change.old_value and change.old_value.get("MountDir") in ["home", "/home"]) or (
```
gmarciani, line 232: Minor: if you see that we are duplicating the list [home, /home] also elsewhere I suggest to define a constant to reduce duplication and reduce risks of errors (specifying only one of the two in other parts of the code)

gmarciani, line 232: With this condition we will prevent to change from home to /home and viceversa, even if /home and home are semantically equivalent for us. Putting this comment just to verify that this is expected.

cli/src/pcluster/config/update_policy.py
```diff
@@ -318,6 +329,8 @@ def condition_checker_managed_fsx(change, patch):
 
 
 def actions_needed_shared_storage_update(change, patch):
+    if is_home_change(change):
+        return "Please revert any changes to the /home mount"
```
gmarciani, line 333: What about capture the scenario with unit tests?

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Actually, I think changing from /home to home doesn't result in a change to evaluate

dreambeyondorange: Currently this is not used anywhere else in the codebase

cli/src/pcluster/config/update_policy.py
dreambeyondorange: Good point, I will add these!

dreambeyondorange:
Added unit tests!

gmarciani (approved):

PR #5906, opened 2023-12-04
[Performance Tests] Add coverage for all OSes supported by StarCCM+ and OpenFOAM with respective baselines.

gmarciani:
### Description of changes
1. Add coverage for all OSes supported by StarCCM+ and OpenFOAM
2. Add baselines for every covered OSes.
3. Distinguish between acceptable and unacceptable degradation in test logs.

### Tests
1. Perf tests executed in personal account.

### DIFF
--- a/tests/integration-tests/tests/performance_tests/test_starccm.py
+++ b/tests/integration-tests/tests/performance_tests/test_starccm.py
@@ -10,7 +10,14 @@ STARCCM_INSTALLATION_TIMEOUT = 1800
 STARCCM_JOB_TIMEOUT = 600
 STARCCM_LICENCE_SECRET = "starccm-license-secret"
 TASK_VCPUS = 36  # vCPUs are cut in a half because multithreading is disabled
-BASELINE_CLUSTER_SIZE_ELAPSED_SECONDS = {8: 64.475, 16: 33.1723, 32: 17.8983}
+BASELINE_CLUSTER_SIZE_ELAPSED_SECONDS = {
+    "alinux2": {8: 64.475, 16: 33.173, 32: 17.899},  # v3.1.3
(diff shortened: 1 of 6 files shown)


tests/integration-tests/tests/performance_tests/test_starccm.py
```diff
@@ -51,8 +57,16 @@ def test_starccm(
     number_of_nodes,
     test_datadir,
     scheduler_commands_factory,
+    s3_bucket_factory,
 ):
-    cluster_config = pcluster_config_reader(number_of_nodes=max(number_of_nodes))
+    # Create S3 bucket for custom actions scripts
+    bucket_name = s3_bucket_factory()
+    s3 = boto3.client("s3")
+    s3.upload_file(str(test_datadir / "dependencies.install.sh"), bucket_name, "scripts/dependencies.install.sh")
```
enrico-usai, line 65: to avoid the creation of an extra s3 bucket for a script of 1 line you can:
1. remove the `dependencies.install.sh` from the `OnNodeConfigured` and instead use the `remote_command_executor.run_remote_script` approach.
2. simply use `remote_command_executor.run_remote_command` given that it's a one-liner command

enrico-usai: In this way you can also avoid the usage of `install_extra_deps` in the cluster config file.

gmarciani: The deps installation is required in the whole cluster, not only on the head node.
In this case I think the best approach is to make use of custom actions. What do you think?

enrico-usai: In this case, I agree.

enrico-usai (approved):

PR #5918, opened 2023-12-05
Add performance test for startup time of compute nodes

hgreebe:
### Description of changes
* Add performance test for startup time of compute nodes
* Checks if the startup time of a compute node of a specific instance type and OS is within 2 standard deviations of the mean of a dataset of previously recorded start up times.

### Tests
* Ran performance test on personal Jenkins


tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
...
+from dateutil.relativedelta import relativedelta
+from utils import describe_cluster_instances
+
+MINIMUM_DATASET_SIZE = 2
+
+
+def evaluate_data(value, data):
+    standard_deviation = statistics.stdev(data)
+    mean = statistics.mean(data)
+    logging.info(f"Standard deviation: {standard_deviation}")
```
gmarciani, line 46: What about including also the mean within the log line?

tests/integration-tests/tests/performance_tests/test_startup_time/test_startup_time/pcluster.config.yaml
```diff
@@ -0,0 +1,57 @@
...
+HeadNode:
+  InstanceType: c5n.xlarge
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Iam:
+    AdditionalIamPolicies:
+      - Policy: arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore #Required to report patching status
+      - Policy: arn:aws:iam::aws:policy/CloudWatchFullAccess # Required to add performance testing widgets to the cluster dashboard
```
gmarciani: Do we expect the head node to create the cloudwatch widget?
Is the DevSetting `ComputeStartupTimeMetricEnabled` that makes the head node create the widget?

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if result["Datapoints"]:
+        value = result["Datapoints"][0].get("Average")
+    return value
+
+
+def test_startup_time(pcluster_config_reader, clusters_factory, test_datadir, region, instance, os, scheduler):
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    logging.info("Cluster Created")
```
gmarciani: Minor: I may be wrong but I think you already get the cluster creation log line from cluster_factory

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
...
+            {"Name": "ClusterName", "Value": cluster.name},
+            {"Name": "InstanceType", "Value": instance_type},
+        ]
+
+        value = get_metric(dimensions, cw_client)
+        logging.info(f"Value: {value}")
+
+        # get historical data
+        dimensions = [{"Name": "InstanceType", "Value": instance_type}, {"Name": "OS", "Value": os}]
+        data = get_data(dimensions, cw_client)
```
gmarciani: Clean Coding Hint: Separation of responsibilities.
Generating the dimension is not a responsibility of the test_startup_time, but it may be something done within an auxiliary function. 

The main responsibility of `test_startup_time` is to create cluster, retrieve data and make assertions on that; it should not be responsible of *how* to do that (Definition of dimensions). My suggestion is move the logic from line 93 to line 106 to a dedicated auxiliary function and leave here only the data retrieval

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
...
+    for instance in instances:
+        instance_type = instance["InstanceType"]
+        instance_id = instance["InstanceId"]
+        logging.info(f"Type: {instance_type}")
+
+        # get new startup time
+
+        dimensions = [
+            {"Name": "OS", "Value": os},
+            {"Name": "InstanceID", "Value": instance_id},
```
gmarciani: Specifying the instance id should be enough to retrieve the metric emitted by an ec2 instance.
why do we need to specify also the OS, cluster name and instance_type?

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
+import logging
+import statistics
+from datetime import datetime
+
+import boto3
+import pytest
+from dateutil.relativedelta import relativedelta
+from utils import describe_cluster_instances
+
+MINIMUM_DATASET_SIZE = 2
```
gmarciani: Why a sample size of 2. 
Is it possible to increase the dataset size to have more statistically powerful insights?

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,125 @@
...
+        # evaluate data
+        if len(data) > MINIMUM_DATASET_SIZE and value:
+            degradation = evaluate_data(value, data)
+            if degradation:
+                performance_degradation[instance_type] = value
+
+    if performance_degradation:
+        degraded_instances = performance_degradation.keys()
+        pytest.fail(
+            f"Performance test results show performance degradation for the following instances: {degraded_instances}"
```
gmarciani: I suggest to specify the magnitude of the degradation with the failure log.
If there is degradation, we would like to know how much is the degradation to make more informative decision. 

Example: If the degradation is really close to the level of tolerance we may end up saying that it's ok because the sample size is not enough to mitigate the variance.

tests/integration-tests/tests/performance_tests/test_startup_time.py
hgreebe: In order for the get_metric_statistics call to return the data point, you have to include all the dimensions. I tried only specifying the instance id, but the api call returned an empty dataset.

tests/integration-tests/tests/performance_tests/test_startup_time.py
hgreebe: I will increase it to 5

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    performance_degradation = {}
+
+    for instance in instances:
+        instance_type = instance["InstanceType"]
+        instance_id = instance["InstanceId"]
+        logging.info(f"Type: {instance_type}")
+
+        # get new startup time
```
gmarciani: Minor: we can remove this comment and name the `value` variable `startup_time_value`

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,126 @@
...
+
+        value = get_metric(os, cluster, instance_type, instance_id, cw_client)
+        logging.info(f"Value: {value}")
```
gmarciani: I suggest to be more specific on this log line and say "Observed Startup Time for instance ${instance_id} (${instance_type}) of cluster ${cluster_name}: ${value} seconds" (os can be avoided in the log line because it is reported in the test name that is the prefix of every log line). I assumed it is in seconds, if not let's use the appropriate unit of measure.

tests/integration-tests/tests/performance_tests/test_startup_time.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # get historical data
+        data = get_data(instance_type, os, cw_client)
+        if value in data:
+            data.remove(value)
+
+        logging.info(f"Data of {instance_type}: {data}")
+
+        # evaluate data
+        if len(data) > MINIMUM_DATASET_SIZE and value:
```
gmarciani: I `value` is None the test should fail because it was not able to collect the required data.
Furthermore, I think that the required sample size should be enforced on the number of compute nodes (eg: if we require a sample of size 10, then we should create a cluster with 10 compute nodes). With this enforcement is not even required to check if the length of data is greater than sample size, but we simply need to wait for the cluster to be in create complete.

Also, our tests should always yuse the same sample size, so we do not expect any case where data has length greater than the rewuired sample size.

himani2411:
In general, Is there a reason why we are getting these data points from a new test altogether? Is there a possibility to add these metrics in an existing test?

himani2411 (approved):

tests/integration-tests/tests/performance_tests/test_startup_time.py
gmarciani: Got it -- Having a look at the design I also understand the reasoning behind it. Let's go with this implementation.

PR #5923, opened 2023-12-06
Improvements to the external slurmdbd infrastructure

jdeamicis:
### Description of changes
* Add DBMS client security group from the DBMS stack to the slurmdbd instance;
* Make the user data for the slurmdbd instance modular (and disable some automatic package upgrades enabled by default by cloud-init);
* Fix cfn-init configuration for slurmdbd instance so that the cinc-client command can actually start on the slurmdbd instance.

### Tests
* Successful manual tests of the improvements.

### Checklist

### DIFF
--- a/cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
+++ b/cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
@@ -269,19 +269,20 @@ class ExternalSlurmdbdStack(Stack):
             type="String",
             description="The SSH key name to access the instance (for management purposes only)",
         )
-
         private_ip = CfnParameter(
             self,
             "PrivateIp",
             type="String",
             description="Static private IP address + prefix to assign to the slurmdbd instance",
         )
-
         private_prefix = CfnParameter(
             self,
             "PrivatePrefix",
             type="String",
             description="Subnet prefix to assign with the private IP to the slurmdbd instance",
+        )
+        dbms_client_sg_id = CfnParameter(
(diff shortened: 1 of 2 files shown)


cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
```diff
@@ -269,20 +270,21 @@ def _add_external_slurmdbd_launch_template(self):
...
             description="Static private IP address + prefix to assign to the slurmdbd instance",
         )
-
         private_prefix = CfnParameter(
             self,
             "PrivatePrefix",
             type="String",
             description="Subnet prefix to assign with the private IP to the slurmdbd instance",
         )
+        dbms_client_sg_id = CfnParameter(
```
EddyMM, line 285: NIT: We should have a central function (probably called from the constructor) to load all the CfnParameters we need for this stack.
It's difficult to trace all the `CfnParameters` associated with the stack when some are defined in the constructor, and others in lower depth functions.

EddyMM (approved):

cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
jdeamicis: I'm not yet fully sure about this. On one side I agree with you, on the other by defining `CfnParameters` deeper in the tree we can better highlight which parts of the stack the they are used for (especially if they are needed only in one single resource, such as the launch template).

I will create an internal task to deal with this later in the project.

PR #5924, opened 2023-12-06
[develop] Permit to pin kernel version while updating os packages

enrico-usai:
Sometimes it's required to pin the kernel version (e.g. if Lustre client is not avail for latest kernel) but at the same time permit to update all the other OSes packages.

Added a new `DevSettings/DisableKernelUpdate` parameter that is used only when `UpdateOsPackages` is set to true.


enrico-usai:
Closing in favour of #5929

PR #5928, opened 2023-12-11
Decouple OSU benchmarks from EFA integration tests

judysng:
### Description of changes
* Originally, OSU benchmark tests were nested inside EFA integration tests, so this PR makes a new test_osu.py and separated the OSU tests from EFA 
  * This is placed in the performance_tests folder, along with a new cluster config specifically using c5n.18xlarge and the benchmarks
* OSU benchmarks could also be run from other tests (test_ad_integration, test_multiple_efs, test_multiple_fsx, and test_hit_disable_hyperthreading) if the test config option "benchmarks" was set.
  * This PR also removes the benchmark config sections from those tests and deletes the extra code in those tests that uses the fixture and runs the test
  * The actual fixture and test config option is left though, just in case for future use

### Tests
* Ran integration tests for each of the tests modified (osu, efa, efs, ad_integration, fsx, and disable_hyperthreading)

### Checklist


tests/integration-tests/tests/efa/test_efa.py
```diff
@@ -44,11 +39,8 @@ def test_efa(
 
     Grouped all tests in a single function so that cluster can be reused for all of them.
     """
-    # We collected OSU benchmarks results for c5n.18xlarge only.
-    osu_benchmarks_instances = ["c5n.18xlarge"]
 
-    # 32 instances are required to see performance differences in collective OSU benchmarks.
-    max_queue_size = 32 if instance in osu_benchmarks_instances else 2
+    max_queue_size = 2
```
gmarciani: Minor: since this is now constant we can put it in the cluster config and remove the variable injection.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -670,13 +666,11 @@ def test_ad_integration(
...
     4. AD users can submit workloads;
     5. AD users filter out by LdapAccessFilter cannot access to the head node.
-
-    Optionally, it executes performance tests using OSU benchmarks.
     """
     if not is_directory_supported(region, directory_type):
         pytest.skip(f"Skipping the test because directory type {directory_type} is not supported in region {region}")
 
-    head_node_instance_type = "c5n.18xlarge" if request.config.getoption("benchmarks") else "c5.xlarge"
+    head_node_instance_type = "c5.xlarge"
```
gmarciani: Minor: since this is now constant we can put it in the cluster config and remove the variable injection.

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -0,0 +1,267 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Minor: copyright year should be 2023

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -0,0 +1,267 @@
...
+    clusters_factory,
+    test_datadir,
+    architecture,
+    network_interfaces_count,
+    mpi_variants,
+    scheduler_commands_factory,
+    request,
+):
+    # We collected OSU benchmarks results for c5n.18xlarge only.
+    osu_benchmarks_instances = ["c5n.18xlarge"]
```
gmarciani: Minor: this is a constant that could be defined outside the scope of the function

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if instance not in osu_benchmarks_instances:
+        raise Exception(f"OSU benchmarks can't be run on instance {instance}.")
```
gmarciani: Let's add to the log line what are the supported instances. In this way if a technician uses the wrong instance type he/she can know immediately what should be used.

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -0,0 +1,267 @@
...
+
+    # 32 instances are required to see performance differences in collective OSU benchmarks.
+    max_queue_size = 32
```
gmarciani: This is now a constant so we can put it directly into the clyster config file and avoid variable injection.

tests/integration-tests/tests/performance_tests/test_osu/test_osu/pcluster.config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  Imds:
+    Secured: {{ imds_secured }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  SlurmQueues:
+    - Name: efa-enabled
+      Networking:
+        PlacementGroup:
+          Enabled: true
+          Name: c5n_capacity_reservation
```
gmarciani, line 18: Another reason to hard wire also the compute nodes instance type: we are hard wiring the c5n PG (and it is correct), so it makes sense to also hard wire the c5n compute capacity rather than using a variable.

tests/integration-tests/tests/storage/test_efs.py
```diff
@@ -168,7 +163,7 @@ def test_multiple_efs(
     for i in range(num_existing_efs):
         existing_efs_mount_dirs.append(f"/existing_efs_mount_dir_{i}")
 
-    new_efs_mount_dirs = ["/shared"]  # OSU benchmark relies on /shared directory
+    new_efs_mount_dirs = ["/shared"]
```
gmarciani, line 166: According to the removed comment it seems that the /hsared folder was required only by OSU benchmarks. Is it still required if we remove them?

tests/integration-tests/tests/storage/test_fsx_lustre.py
```diff
@@ -325,14 +324,10 @@ def test_multiple_fsx(
...
-        # Only create more FSx when benchmarks are specified. Limiting OS to reduce cost of too many file systems
-        num_existing_fsx = 20
-    else:
-        # Minimal total existing FSx is the number of Ontap and OpenZFS plus one existing FSx Lustre
-        num_existing_fsx = num_existing_fsx_ontap_volumes + num_existing_fsx_open_zfs_volumes + 1
+    # Minimal total existing FSx is the number of Ontap and OpenZFS plus one existing FSx Lustre
+    num_existing_fsx = num_existing_fsx_ontap_volumes + num_existing_fsx_open_zfs_volumes + 1
     num_existing_fsx_lustre = num_existing_fsx - num_existing_fsx_ontap_volumes - num_existing_fsx_open_zfs_volumes
-    fsx_lustre_mount_dirs = ["/shared"]  # OSU benchmark relies on /shared directory
+    fsx_lustre_mount_dirs = ["/shared"]
```
gmarciani, line 330: Same as above: According to the removed comment it seems that the /hsared folder was required only by OSU benchmarks. Is it still required if we remove them?

tests/integration-tests/tests/performance_tests/test_osu.py
judysng: Makes sense, I put this and the other constants into their cluster config files. The only thing is that max_queue_size in test_osu is also referenced later, so I just have 32 floating there.

tests/integration-tests/tests/storage/test_efs.py
judysng: Yes, I originally tried to remove the "/shared", but the cluster configs use the shared mount dir ([test_efs here](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/tests/storage/test_efs/test_multiple_efs/pcluster.config.yaml#L46-L53) and [test_fsx_lustre here](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/tests/storage/test_fsx_lustre/test_multiple_fsx/pcluster.config.yaml#L38-L46)), and I wasn't sure if I should edit those cluster configs to delete them if the mount dir was used for something else, so I added the "/shared" back in. Should I try deleting them?

gmarciani: No, in this case better to keep it.

tests/integration-tests/tests/storage/test_fsx_lustre.py
gmarciani: Clarified in a similar comment. Better to keep this in this PR.

tests/integration-tests/tests/ad_integration/test_ad_integration/test_ad_integration/pcluster.config.yaml
```diff
@@ -1,7 +1,7 @@
 Image:
   Os: {{ os }}
 HeadNode:
-  InstanceType: {{ head_node_instance_type }}
+  InstanceType: "c5.xlarge"
```
gmarciani: I think I expressed my thought unclearly. what I meant with my previous comment was that we should not inject different instance types from the test body, but still we should use the instance type defined in the test config yaml file.

So for head node and compute fleet, we should use `{{ instance }}`

judysng: Got it, I changed it to `{{ instance }}`!

gmarciani: Thank you!

tests/integration-tests/tests/performance_tests/test_osu.py
gmarciani: Alright, let's keep whatever is referenced in the test body as the refactoring of the test body is not in scope for the PR

gmarciani (approved):

PR #5929, opened 2023-12-11
[develop] Pin Ubuntu 22 kernel version to LTS while updating OS packages

enrico-usai:
### Description of changes

Sometimes it's required to pin the kernel version (e.g. if Lustre client is not avail for latest kernel)
but at the same time permit to update all the other OSes packages.

Added a new `DevSettings/DisableKernelUpdate` parameter that is used only when `UpdateOsPackages` is set to true.

When `DISABLE_KERNEL_UPDATE` is set to true:
- for Ubuntu we're going to keep minor version (e.g. 5.15.0) but updating it to latest available patch version (through the  `linux-image-aws-lts-22.04` package)
- for other oses we're pinning the kernel version to the installed one, updating other packages, then removing the pin. 

In future we could pin to LTS for rhel OSes too and differentiate the behaviour by having 2 parameters (e.g. `DisableKernelUpdate` and `PinLtsKernel` or something like this).

### Ubuntu packaging details 
* linux-aws, linux-image-aws, linux-headers-aws are meta packages that permit to update to the latest kernel version (6.x) 
* linux-aws-lts-22.04, linux-image-aws-lts-22.04, linux-headers-aws-lts-22.04 are meta packages that permit to update to latest 5.15.0
* linux-aws-lts-22.04 depends on linux-image-aws-lts-22.04 and linux-headers-aws-lts-22.04
* linux-image-aws-lts-22.04 depends on linux-image-5.15.0-10xx-aws, that depends on linux-modules-5.15.0-10xx-aws
* linux-headers-aws-lts-22.04 depends on linux-headers-5.15.0-10xx-aws

### Tests
* Executed first stage build in personal Jenkins.


cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -107,6 +107,22 @@ phases:
...
+              set -v
+              PLATFORM='{{ build.PlatformName.outputs.stdout }}'
+              
+              # Wait for completion of os updates performed by cloud-init
+              if [[ ${!PLATFORM} == DEBIAN ]]; then
+                while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done
+                flock $(apt-config shell StateDir Dir::State/d | sed -r "s/.*'(.*)\/?'$/\1/")/daily_lock systemctl disable --now apt-daily.timer apt-daily.service apt-daily-upgrade.timer apt-daily-upgrade.service
+                sed "/Update-Package-Lists/s/\"1\"/\"0\"/; /Unattended-Upgrade/s/\"1\"/\"0\"/;" /etc/apt/apt.conf.d/20auto-upgrades > "/etc/apt/apt.conf.d/51pcluster-unattended-upgrades"
+                DEBIAN_FRONTEND=noninteractive apt-get -y update
+              fi
```
jdeamicis: Could we put some comments explaining what we are doing here and why? Even in a later PR.

jdeamicis: Oh, I see this was already present a bit below in the code.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -117,7 +133,97 @@ phases:
...
+              set -v
+              OS='{{ build.OperatingSystemName.outputs.stdout }}'
+              PLATFORM='{{ build.PlatformName.outputs.stdout }}'
+              DISABLE_KERNEL_UPDATE='{{ build.DisableKernelUpdate.outputs.stdout }}'
+              
+              if [[ ${!DISABLE_KERNEL_UPDATE} == true ]]; then
+                if [[ ${!PLATFORM} == RHEL ]]; then
+                  yum install -y yum-plugin-versionlock
+                  # listing all the packages because wildcard does not work as expected
+                  yum versionlock kernel kernel-core kernel-modules
```
jdeamicis, line 200: Don't we have `kernel-devel` at this stage? I guess not.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -149,13 +256,47 @@ phases:
                 else
                   package-cleanup -y --oldkernels --count=1
                 fi
+                # Install kernel-devel during OS update, so that headers are aligned with new kernel. 
+                # The same is done for Debian through `apt-get -y install linux-aws`
                 yum -y install kernel-devel
```
jdeamicis, line 261: I'm not sure this is always correct.

In RHEL derivatives if you pin the kernel to the version currently available in the AMI (via `DISABLE_KERNEL_UPDATE`), but the AMI is old enough, then this command may pull a version of `kernel-devel` newer than the current kernel, wouldn't it?

I'm afraid in this case for safety we should install a specific version via `yim install kernel-devel-${<current_version>}`, where `<current_version>` is derived from `uname -r`.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -149,13 +256,47 @@ phases:
...
-                sed "/Update-Package-Lists/s/\"1\"/\"0\"/; /Unattended-Upgrade/s/\"1\"/\"0\"/;" /etc/apt/apt.conf.d/20auto-upgrades > "/etc/apt/apt.conf.d/51pcluster-unattended-upgrades"
-                DEBIAN_FRONTEND=noninteractive apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --with-new-pkgs upgrade && apt-get --purge autoremove -y
-                apt-get -y install linux-aws linux-headers-aws linux-image-aws
+                DEBIAN_FRONTEND=noninteractive apt-get -y update
+                DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --with-new-pkgs upgrade
+                apt-get --purge autoremove -y
+
+                if [[ ${!DISABLE_KERNEL_UPDATE} != true ]]; then
+                   # already installed to LTS version
+                   apt-get -y install linux-aws linux-headers-aws linux-image-aws
```
jdeamicis, line 270: Why is this necessary? Wouldn't it be picked up by the previous upgrade? In case `$DISABLE_KERNEL_UPDATE != true` we shouldn't have removed them from the base AMI, should we?

jdeamicis (commented):
Left some comments with some questions to check together.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
enrico-usai: Adding a comment.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
enrico-usai: not yet, it's installed later

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
enrico-usai: This code was already there. I imagine the reason is that these packages (in particular linux-aws) are not always installed (expecially in case of custom AMIs). With this code we're sure we have them.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
jdeamicis: Agreed to implement in a separate PR.

jdeamicis (approved):

PR #5935, opened 2023-12-12
[develop] Install kernel-devel for Rocky Linux from vault 

enrico-usai:
### Description of changes

kernel-devel should be aligned to the installed version.
If there is kernel 8.8 we should have kernel-devel 8.8.

Without this patch we were installing latest kernel-devel package.
We cannot use `--releasever` to install 8.8 version because it is not more available in the repo.
Previous releases are moved into a vault area once a new minor release version is available for at least a week.

On RHEL8 instead all versions of the packages will remain available.

With this patch we're downloading this specific package from vault and installing it
if it's not available in the repository for the specific release version.

### References

Related to https://github.com/aws/aws-parallelcluster-cookbook/pull/2598

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
@@ -206,7 +206,7 @@ phases:
                   fi
               
                   if [[ ${!OS} == "rocky8" ]] ; then
-                    yum versionlock rocky-release
+                    yum versionlock rocky-release rocky-repos
                   elif [[ ${!OS} == "rhel8" ]] ; then
                     yum versionlock redhat-release
                   fi
--- a/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
@@ -234,7 +234,6 @@ phases:
               set -v
               OS='{{ build.OperatingSystemName.outputs.stdout }}'
               PLATFORM='{{ build.PlatformName.outputs.stdout }}'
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -256,15 +255,29 @@ phases:
...
+              OS='{{ build.OperatingSystemName.outputs.stdout }}'
+              PLATFORM='{{ build.PlatformName.outputs.stdout }}'
+              DISABLE_KERNEL_UPDATE='{{ build.DisableKernelUpdate.outputs.stdout }}'
 
+              if [[ ${!PLATFORM} == RHEL ]]; then
+                # Install kernel-devel during OS update, so that headers are aligned with new kernel. 
+                # The same is done for Debian through `apt-get -y install linux-aws`
+                yum -y install kernel-devel
+              
+              elif [[ ${!PLATFORM} == DEBIAN ]]; then
```
jdeamicis, line 295: If this is `InstallUpdatedHeaders`, I would expect only to be installing `linux-headers-aws` here. If you are adding this new step in the component, then let's add another one only for Debian derivatives where we install `linux-aws` and `linux-image-aws` and call it something like `InstallAWSDebianKernel` or something similar.

jdeamicis (commented):
I believe we should add a separate step for the installation of the kernel packages (other than headers) for debian-based distributions.

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
jdeamicis: Ok, I consider this addressed.

jdeamicis (dismissed):

jdeamicis (approved):

PR #5948, opened 2023-12-14
[Test][Isolated Regions] Restrict us-isob-east-1 test only in 2 AZs + ignore unsupported integ tests + increase monitoring time in us-iso regions.

gmarciani:
### Description of changes
1. skip integ test test_multi_az_create_and_update in isolated regions as it relies on unsupported resources and cannot be executed there.
2. restrict tests in us-isob-east-1 to be alunched in two AZs
3. increase monitoring time of scaling test from 5 to 10 minutes when running in isolated regions.

### DIFF
--- a/tests/integration-tests/conftest_networking.py
+++ b/tests/integration-tests/conftest_networking.py
@@ -56,6 +56,8 @@ AVAILABLE_AVAILABILITY_ZONE = {
     "eu-central-1": ["euc1-az2", "euc1-az3"],
     # FSx not available in cnn1-az4
     "cn-north-1": ["cnn1-az1", "cnn1-az2"],
+    # Should only consider supported AZs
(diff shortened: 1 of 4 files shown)


tests/integration-tests/conftest_networking.py
```diff
@@ -56,6 +56,8 @@
     "eu-central-1": ["euc1-az2", "euc1-az3"],
     # FSx not available in cnn1-az4
     "cn-north-1": ["cnn1-az1", "cnn1-az2"],
+    # Should only consider supported AZs
+    "us-isob-east-1": ["us-isob-east-1b", "us-isob-east-1c"],
```
himani2411, line 60: So do we no longer want to test the other AZ?

gmarciani: No, at least in the mid term

himani2411 (approved):

PR #5951, opened 2023-12-15
[Test] Make StarCCM+/OpenFOAM performance test use the capacity reser…

gmarciani:
### Description of changes
Make StarCCM+/OpenFOAM performance test use the capacity reservation.

### Tests
* [ONGOING] Executed StarCCM+ perf test (no need to execute OpenFOAM as the change is the same)


gmarciani:
Unnecessary. The test looked like not using the expected capacity reservation on RHEL8 because RHELK8 actually requires a diffeerent reservation. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-reservations.html#capacity-reservations-platforms

PR #5961, opened 2023-12-18
[develop][integ-tests] Avoid assuming IAM role itself.

hanwen-pcluste:
All new accounts prevent assuming IAM role itself by default (https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/). Therefore, we discovered that some integration tests assume a role multiple times. This commit avoids assuming IAM role itself

This bug is discovered during region expansion.

### Tests
* Ran tests in opt-in region and succeeded

### Checklist

### DIFF
--- a/tests/integration-tests/framework/credential_providers.py
+++ b/tests/integration-tests/framework/credential_providers.py
@@ -54,37 +54,49 @@ def aws_credential_provider(region, credential_arg):
 @contextmanager
 def sts_credential_provider(region, credential_arn, credential_external_id=None, credential_endpoint=None):
     """Assume a role through STS and set such creds in the environment for the duration of the context."""
-    credentials_to_backup = _get_current_credentials()
-
-    logging.info("Assuming STS credentials for region %s and role %s", region, credential_arn)
-    aws_credentials = _retrieve_sts_credential(region, credential_arn, credential_external_id, credential_endpoint)
-    logging.info("Retrieved credentials %s", obfuscate_credentials(aws_credentials))
-
-    try:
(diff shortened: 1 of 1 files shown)


tests/integration-tests/framework/credential_providers.py
```diff
@@ -54,37 +54,49 @@ def aws_credential_provider(region, credential_arg):
...
+    caller_arn_split = re.split(":|/", caller_arn)
+    credential_arn_split = re.split(":|/", credential_arn)
+    if caller_arn_split[4] == credential_arn_split[4] and caller_arn_split[6] == credential_arn_split[6]:
+        # If account id and role name are the same, the role do not need to / cannot assume itself
+        # Caller ARN can be a IAM role arn or STS session ARN.
+        # Therefore, we have to check the account id and role name are the same, instead of checking the whole string.
+        # For example: User:
+        #   arn:aws:sts::123456700000:assumed-role/JenkinsStsRole-il-central-1/il-central-1_integration_tests_session
+        #   is not authorized to perform: sts:AssumeRole on resource:
+        #   arn:aws:iam::123456700000:role/JenkinsStsRole-il-central-1
```
himani2411, line 74: Do we need to check the Expiry of that role? which identity does it take if the assumed role has expired? we might get an exception if it has expired

hanwen-pcluste: It is nice to have, but currently cannot be done:
1. There is no boto3 calls for checking the expiry time
2. We are not using this function to renew the expiry time intentionally. So I believe this change does not introduce any regression

himani2411 (approved):

PR #5982, opened 2023-12-27
Add 0 as an option for RetentionInDays cluster config param to not update CW log retention

judysng:
### Description of changes
* Allow 0 as an option for `RetentionInDays` cluster config parameter
  * Setting that value to be 0 does not change the CW log retention days and instead uses the [CW default, which is indefinite](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Working-with-log-groups-and-streams.html#SettingLogRetention).
    * Right now, the pcluster `RetentionInDays` [default value is 180 days without an option to never expire](https://docs.aws.amazon.com/parallelcluster/latest/ug/Monitoring-v3.html#yaml-Monitoring-Logs-CloudWatch-RetentionInDays). Even if the user doesn't specify `RetentionInDays`, pcluster will automatically make it 180 days. 
    * We want to allow this 0 value because some customers may have restrictions on updating log retentions
* `get_cloud_watch_logs_retention_days` is used when calling `logs.CfnLogGroup()` which passes in the return value for `retention_in_days` (such as in [awsbatch_builder.py](https://github.com/judysng/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/awsbatch_builder.py#L532)). `retention_in_days` is `None` by default in `logs.CfnLogGroup()`, so passing in `None` will use the default retention.

### Tests
* Created a cluster with a cluster config with RetentionInDays as 0
* Created a cluster with a cluster config with RetentionInDays as 1, then ran update-cluster with RetentionInDays as 0. The retention policy updated to never expire, the default.
* Ran cloud watch logs integration test with retention days as 0

### Checklist

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -816,7 +816,7 @@ class CloudWatchLogsSchema(BaseSchema):
 
     enabled = fields.Bool(metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
     retention_in_days = fields.Int(
-        validate=validate.OneOf([1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, 3653]),
+        validate=validate.OneOf([0, 1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, 3653]),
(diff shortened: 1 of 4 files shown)


hanwen-pcluste (dismissed):
Great work!

hanwen-pcluste (approved):

PR #6009, opened 2024-01-12
Add resource leak integration test

judysng:
### Description of changes
* Added integration test for checking that there isn't a file handler leak in computemgtd
  * Previously, there was a [file handler leak that was fixed](https://github.com/aws/aws-parallelcluster-node/pull/421), so we wanted to add a new test to make sure that we catch any leaks
* Added resource leak checking as callable functions. Called the functions in starccm test because the starccm runs for ~20 mins, which is long enough to notice the leak

### Tests
* Ran on personal Jenkins 
  * Current computemgtd's number of file descriptors doesn't change over time
  * Ran with a modified computemgtd that introduces the above file handler leak -- the test fails and number of open files increases as expected

### References
* https://github.com/aws/aws-parallelcluster-node/pull/421

### Checklist


dreambeyondorange (dismissed):

tests/integration-tests/tests/resource_leaks/test_resource_leaks.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from utils import get_compute_nodes_instance_ips
+
+from tests.common.assertions import assert_head_node_is_running
+
+
+@pytest.mark.usefixtures("instance", "os", "scheduler")
+def test_resource_leaks(
```
enrico-usai: can we avoid having a specific integration test for it? and instead add this logic to check file descriptors and resource leak by default in all the tests. I mean, we don't need to wait 30 minutes in every test, but we can do the check.

judysng: I updated the PR to add the logic in functions in utils.py and called it in the starccm test. Unfortunately I don't think we can add it by default to all the tests unless we add the assert in cluster creation/deletion, which I think we'd want to avoid

dreambeyondorange: How long does the starccm test run typically?

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,40 @@ def wait_process_completion(remote_command_executor, pid):
         raise Exception("The process is still running")
     else:
         return result.stdout.strip()
+
+
+def check_file_handler_leak(remote_command_executor, slurm_commands, cluster, region):
```
dreambeyondorange: I'm not sure this function name best describes what this is doing.  It seems like this just gets the number of open files.

tests/integration-tests/tests/resource_leaks/test_resource_leaks.py
judysng: ~20 mins, which is long enough for the resource leak test to fail. I tested running this starccm test with the resource leak issue [here](https://github.com/aws/aws-parallelcluster-node/pull/421) and it failed as expected.

tests/integration-tests/tests/common/utils.py
judysng: True, I renamed it to be `get_compute_ip_to_no_files`

dreambeyondorange: nit: I think the term `no` is less explicit than `num` to abbreviate the number of files.  `no` has more binary implications or boolean implications

dreambeyondorange (approved):

PR #6010, opened 2024-01-12
[external_slurmdbd] Allow traffic from external slurmdbd to slurmctld's

jdeamicis:
### Description of changes
Add ingress rule in the external dbd client security group to allow connections initiated from the slurmdbd.

Such connections may be established from the slurmdbd's side when the slurmdbd drops and then recovers: in this case it is the slurmdbd attempting to re-establish the connectivity between itself and all the slurmctld's it was previously connected to.

Remove useless egress rules from security groups (all outgoing traffic is enabled by default at the moment).

### Tests
* Manual test with POC external slurmdbd stack confirming that this works.

### Checklist

### DIFF
--- a/cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
+++ b/cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
@@ -204,20 +204,21 @@ class ExternalSlurmdbdStack(Stack):
             description="Allow SSH access to slurmdbd instance (server)",
             vpc=self.vpc,
         )
+
         client_sg = ec2.SecurityGroup(
             self,
             "SSHClientSecurityGroup",
             description="Allow SSH access to slurmdbd instance (client)",
             vpc=self.vpc,
         )
+
         server_sg.add_ingress_rule(
             peer=client_sg, connection=ec2.Port.tcp(22), description="Allow SSH access from client SG"
         )
-        client_sg.add_egress_rule(
-            peer=server_sg, connection=ec2.Port.tcp(22), description="Allow SSH access to server SG"
-        )
+
         return server_sg, client_sg
 
+    # FIXME: make the ingress rules more configurable
     def _add_slurmdbd_accounting_security_groups(self):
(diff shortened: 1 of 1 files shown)


cloudformation/external-slurmdbd/external_slurmdbd/external_slurmdbd_stack.py
```diff
@@ -239,10 +240,10 @@ def _add_slurmdbd_accounting_security_groups(self):
             description="Allow Slurm accounting traffic from the cluster head node",
         )
 
-        slurmdbd_client_sg.add_egress_rule(
+        slurmdbd_client_sg.add_ingress_rule(
             peer=slurmdbd_server_sg,
-            connection=ec2.Port.tcp(6819),
-            description="Allow Slurm accounting traffic to the slurmdbd instance",
+            connection=ec2.Port.tcp_range(6820, 6829),
```
NSsirena, line 245: The port can be configured in `slurm.conf`  ..if we hardcode here the port range we may impact customers that use different ports (i.e. `CustomSlurmSettings`).

Possibly we should retrieve these values from the config ..or at least document this limitation.

jdeamicis: The `slurm.conf` won't be available on the slurmdbd stack. The only thing we can do is to make the port range configurable via a CFN parameter: I put a TODO in the code, but maybe we want it to expose it immediately?

NSsirena: > The slurm.conf won't be available on the slurmdbd stack.

Yes .. it would be better to have a parameter for this to ensure that if customers use a different port range in the config they can have a simple way to align the slurmdbd stack.

jdeamicis: We discussed about this and decided not to include the customization of the port ranges for the time being (users would have other mechanisms to do so, such as editing the template or changing the ingress rule by hand via the EC2 console).

NSsirena (approved):

PR #6024, opened 2024-01-19
[DFSM][Test] Add check in integ 'test_update_slurm' to verify that compute nodes save the expected cluster config version on DynamoDB.

gmarciani:
### Description of changes
Add check in integ 'test_update_slurm' to verify that compute nodes save the expected cluster config version on DynamoDB.

### Tests
* Executed `test_update_slurm`


tests/integration-tests/tests/update/test_update.py
```diff
@@ -9,6 +9,7 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import json
```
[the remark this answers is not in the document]
gmarciani: Fixed!

tests/integration-tests/tests/common/assertions.py
```diff
@@ -341,3 +342,23 @@ def _assert_slurm_rebooted_nodes(compute_nodes, remote_command_executor):
             "/var/log/slurmctld.log",
             f"node {node} returned to service",
         )
+
+
+def assert_instance_config_version_on_ddb(cluster, expected_cluster_config_version: str):
+    """Verifies that every compute node stored on DynamoDB the expected cluster config version."""
+    compute_nodes = cluster.describe_cluster_instances(node_type="Compute")
+    logging.info(
+        f"Verifying that all {len(compute_nodes)} compute nodes stored the expected config version on DDB: "
```
dreambeyondorange: Not sure we need both logging statements here?

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,23 @@ def wait_process_completion(remote_command_executor, pid):
...
+    If 'compute_node_ip' is specified, the config version will be retrieved from the compute node;
+    otherwise, it will be retrieved from the head node.
+    """
+    command = "sudo cat /etc/chef/dna.json | jq -r '.cluster.cluster_config_version'"
+    rce = (
+        RemoteCommandExecutor(cluster, compute_node_ip=compute_node_ip)
+        if compute_node_ip
+        else RemoteCommandExecutor(cluster)
+    )
+    return rce.run_remote_command(command).stdout
```
dreambeyondorange: Should we make sure we get the data we expect here?

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,23 @@ def wait_process_completion(remote_command_executor, pid):
...
+
+
+@retry(wait_fixed=seconds(3), stop_max_delay=seconds(15))
+def get_ddb_item(region_name: str, table_name: str, item_key: dict):
+    table = boto3.resource("dynamodb", region_name=region_name).Table(table_name)
+    return table.get_item(Key=item_key)["Item"]
```
dreambeyondorange: Do we need a sanity check on the data here as well?

tests/integration-tests/tests/common/utils.py
gmarciani: Good point! I'll check that the returned value is the dna.json by checking that it is a JSON with the "cluster" field as first element

tests/integration-tests/tests/common/utils.py
gmarciani: Here I don't think we need because this function is very generic and we cannot make any assumption on the retrieved item.

tests/integration-tests/tests/common/assertions.py
gmarciani: In general I think your observation is right: we should always minimize the logs noise
In this case, I think it'd be useful to keep both initial and final lines because:

The initial log is useful if the operations made hereafter on DDB are generic: the function get_ddb_item is intentionally generic (it may be useful somewhere else) and if it fails the technician would see a DDB call failing without any context about why that call was happening. Instead, keeping the intial log line, the context would be immediately clear.
The final log is useful to put a milestone in our checks.
What do you think?

dreambeyondorange: Yup, I think it makes sense to provide information that will help with RCA, thanks!

tests/integration-tests/tests/common/utils.py
dreambeyondorange: We can atleast make sure it exists right?  Do we care if the returned item is empty?  Or leave that up to downstream processes?

gmarciani: I would let the downstream process to decide how to react on the absence of the item.
The point here that I should improve is that we should return None if it is missing (now it would fail with KeyErro). Iterating...

gmarciani: Addressed

dreambeyondorange (dismissed):

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,48 @@ def wait_process_completion(remote_command_executor, pid):
...
+
+    return dna_json["cluster"]["cluster_config_version"]
+
+
+def get_deployed_dna_json(cluster, compute_node_ip: str = None):
+    """Retrieves the dna.json from the cluster node
+    If 'compute_node_ip' is specified, it will be retrieved from the compute node;
+    otherwise, it will be retrieved from the head node.
+    """
+    command = "sudo cat /etc/chef/dna2.json"
```
enrico-usai: why the name is "dna2.json"? This does not seem a standard name. If it's a test specific name, let's pass it as parameter to the function with a default value "dna.json"

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,48 @@ def wait_process_completion(remote_command_executor, pid):
...
+    """Retrieves the cluster config version deployed on the cluster node from its dna.json
+    If 'compute_node_ip' is specified, the config version will be retrieved from the compute node;
+    otherwise, it will be retrieved from the head node.
+    """
+    dna_json = get_deployed_dna_json(cluster, compute_node_ip)
+
+    return dna_json["cluster"]["cluster_config_version"]
+
+
+def get_deployed_dna_json(cluster, compute_node_ip: str = None):
```
enrico-usai, line 416: if it's used only locally, let's add an underscore as prefix.

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,48 @@ def wait_process_completion(remote_command_executor, pid):
         raise Exception("The process is still running")
     else:
         return result.stdout.strip()
+
+
+def get_deployed_config_version(cluster, compute_node_ip: str = None):
+    """Retrieves the cluster config version deployed on the cluster node from its dna.json
```
enrico-usai, line 407: Suggestion: we could try to follow [multi-line doc-strings format](https://peps.python.org/pep-0257/#multi-line-docstrings).

In integ tests this is not enforced by our linters, but mentioning it because in the code it's forced.

tests/integration-tests/tests/common/assertions.py
```diff
@@ -341,3 +342,24 @@ def _assert_slurm_rebooted_nodes(compute_nodes, remote_command_executor):
...
+    for compute_node in compute_nodes:
+        instance_id = compute_node["instanceId"]
+        table_name = f"parallelcluster-{cluster.name}"
+        item_id = f"CLUSTER_CONFIG.{instance_id}"
+        item = get_ddb_item(cluster.region, table_name, {"Id": item_id})
+        assert_that(item).is_not_none()
+        cluster_config_version_on_ddb = item["Data"]["cluster_config_version"]
+        assert_that(cluster_config_version_on_ddb).is_equal_to(expected_cluster_config_version)
+    logging.info(
+        f"Verified that all {len(compute_nodes)} compute nodes stored the expected config version on DDB: "
```
enrico-usai: minor: we can create a variable for `len(compute_nodes)` because we're using it in 2 places.

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,48 @@ def wait_process_completion(remote_command_executor, pid):
...
+        if compute_node_ip
+        else RemoteCommandExecutor(cluster)
+    )
+
+    try:
+        result = rce.run_remote_command(command).stdout
+        dna_json = json.loads(result)
+    except RemoteCommandExecutionError as e:
+        raise RuntimeError(f"Cannot retrieve dna.json from cluster node ({rce.target}): {e}")
+    except ValueError as e:
```
enrico-usai: Here we should have a `json.JSONDecodeError`, rather than a `ValueError`.

tests/integration-tests/tests/common/utils.py
```diff
@@ -401,3 +401,48 @@ def wait_process_completion(remote_command_executor, pid):
...
+        raise ValueError(f"Returned value should be a dna.json, but it's not a valid JSON: {e}")
+
+    if "cluster" not in dna_json:
+        raise ValueError("Returned value should be a dna.json, but it does not contain the expect 'cluster' key")
```
enrico-usai: minor: the expected...

tests/integration-tests/tests/common/utils.py
gmarciani: typo. Made locally to temporarily test an unhappy path. Fixed in next rev.

tests/integration-tests/tests/common/utils.py
gmarciani: I've intentionally avoided the underscore, because i think it may be something generally useful as utility.

tests/integration-tests/tests/common/utils.py
gmarciani: Fixed

tests/integration-tests/tests/common/utils.py
gmarciani: Fixed

tests/integration-tests/tests/common/assertions.py
gmarciani: Fixed

enrico-usai (approved):

PR #6028, opened 2024-01-22
Add the config option DeploymentSettings/DefaultUserHome to change the default user's home directory

dreambeyondorange:
Add the config option `DefaultUserHome` under `DeploymentSettings` to change the default user's home directory

 Move the cluster user out of `/home` and into `/local/home` when the `DefaultUserHome` option is set to local

### Tests
* Created clusters locally using both options to ensure the directory was placed correctly

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2620

### Checklist


gmarciani:
May you please describe the change and how you tested it?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ------
 
 **ENHANCEMENTS**
+- Add the config option `DefaultUserHome` under `HeadNode` to allow users to move the default user's home directory to `/local/home` instead of `/home` (default)
```
gmarciani: * "config option" => "configuration parameter"
* Add full stop at the end of the sentence.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -1405,6 +1406,10 @@ def __init__(
             shared_storage_type,
             default="Ebs",
         )
+        self.default_user_home = Resource.init_param(
+            default_user_home,
+            default="Shared",
```
gmarciani: According to the changelog, the default user home is by default` /home`. Why is it `Shared` here?

gmarciani: Clarified below: that value must be one of Shared or Local. 
What about defining an enumeration for this value?
The advantage would be self documentation: a developer could immediately jump to valid values.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1247,6 +1247,7 @@ def _add_head_node(self):
                     "base_os": self.config.image.os,
                     "region": self.stack.region,
                     "shared_storage_type": self.config.head_node.shared_storage_type.lower(),
+                    "default_user_home": self.config.head_node.default_user_home.lower(),
```
gmarciani: Shouldn't this setting be reflected also in compute and login nodes?
I'm thinking about the case where the head node has `DefaultUserHome=Local`. Without reflecting the setting to compute/login nodes, these nodes would have the default user home directory in the usual /home whereas head node would have it in /local/home. Aren't there any side effect of this misalignment?

cli/tests/pcluster/templates/test_cluster_stack/test_head_node_dna_json/head_node_default.dna.json
```diff
@@ -13,6 +13,7 @@
     "dcv_enabled": "false",
     "dcv_port": "NONE",
     "ddb_table": "NONE",
+    "default_user_home": "shared",
```
gmarciani, line 16: What about testing also the other case (local)?

cli/src/pcluster/templates/cluster_stack.py
dreambeyondorange: Maybe it needs to be moved to a different section.  It should apply to all instance types in the cookbook already

dreambeyondorange: Moved to `DeploymentSettings`

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ------
 
 **ENHANCEMENTS**
+- Add the configuration parameter `DefaultUserHome` under `DeploymentSettings` to allow users to move the default user's home directory to `/local/home` instead of `/home` (default).
```
gmarciani: *Minor*
If you want to specify the section where the parameter is placed (good practice), we should write "configuration parameter `DeploymentSettings/DefaultUserHome`"

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1247,6 +1247,12 @@ def _add_head_node(self):
                     "base_os": self.config.image.os,
                     "region": self.stack.region,
                     "shared_storage_type": self.config.head_node.shared_storage_type.lower(),
+                    "default_user_home": self.config.deployment_settings.default_user_home.lower()
+                    if (
+                        self.config.deployment_settings is not None
+                        and self.config.deployment_settings.default_user_home is not None
+                    )
+                    else "shared",
```
gmarciani: What about having a constant that defines the default value for this setting using the enumeration and reuse such constant here and in the initialization of deployment settings in `cli/src/pcluster/config/common.py`?
In this way we would have a unique place where the default is define.

cli/src/pcluster/schemas/common_schema.py
```diff
@@ -262,6 +269,12 @@ def make_resource(self, data, **kwargs):
 class DeploymentSettingsSchema(BaseSchema):
     """Represent the common schema of DeploymentSettings for ImageBuilder and Cluster."""
 
+    default_user_home = fields.Str(
+        required=False,
+        metadata={"update_policy": UpdatePolicy.UNSUPPORTED},
+        validate=validate.OneOf([DefaultUserHomeType.SHARED.value, DefaultUserHomeType.LOCAL.value]),
```
gmarciani: What about listing all values from DefaultUserHomeType rather than explicitly listing each value?

cli/src/pcluster/templates/cluster_stack.py
gmarciani: Thanks for moving it to the more general `DeploymentSettings`.
I'm still not getting why we add this property only in the head node dna.json. If this applies to all cluster nodes it should be applied to every dna.json. How could it be applied to every cluster node if it is only defined within the head node dna.json?

On this topic, I'd agree that we should have a common dna.json where we define common properties just once, but since we do not have it atm i guess we should replicate the property in every dna.json.

dreambeyondorange: added to both queues and login nodes stacks!

cli/src/pcluster/templates/cluster_stack.py
dreambeyondorange: I don't see any precedent for this in the code, but I should use the Enum here!

cli/tests/pcluster/templates/test_login_nodes_stack/test_login_nodes_dna_json/dna-1.json
```diff
@@ -9,6 +9,7 @@
     "custom_awsbatchcli_package": "",
     "custom_node_package": "",
     "cw_logging_enabled": "true",
+    "default_user_home": "shared",
```
himani2411: Nitpick: Can we test with the other default value in one of the unit tests too. Like `dna-1.json` with `local` and`dna-2.json` with shared

dreambeyondorange: good idea, thanks!

cli/src/pcluster/config/common.py
```diff
@@ -386,8 +393,14 @@ class BaseDeploymentSettings(Resource):
     configuration file deploys some Cfn infrastructure as well.
     """
 
-    def __init__(self, lambda_functions_vpc_config: LambdaFunctionsVpcConfig = None, **kwargs):
+    def __init__(
+        self, default_user_home: str = None, lambda_functions_vpc_config: LambdaFunctionsVpcConfig = None, **kwargs
+    ):
         super().__init__(**kwargs)
+        self.default_user_home = Resource.init_param(
```
himani2411: I think we can move this to the ClusterDeploymentSettings as this  is just the cluster setting and not for both Cluster and Image configs

tests/integration-tests/tests/users/test_default_user_home.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+
+# flake8: noqa
+
+
+@pytest.mark.usefixtures("os", "scheduler", "instance")
+def test_default_user_local_home(
```
himani2411, line 22: Can we add this in an already existing test?

cli/tests/pcluster/templates/test_queues_stack/test_compute_nodes_dna_json/dna-2.json
```diff
@@ -9,6 +9,7 @@
     "custom_awsbatchcli_package": "",
     "custom_node_package": "",
     "cw_logging_enabled": "true",
+    "default_user_home": "shared",
```
himani2411: Can we make this "Local"?

tests/integration-tests/tests/users/test_default_user_home.py
dreambeyondorange: I tried to find a test to put this in, but there wasn't a good category to add this in.

cli/src/pcluster/config/common.py
dreambeyondorange: yes it should be there good catch!

cli/tests/pcluster/templates/test_queues_stack/test_compute_nodes_dna_json/dna-2.json
dreambeyondorange: I tested local in dna-1.json

himani2411: I see it being shared in `cli/tests/pcluster/templates/test_queues_stack/test_compute_nodes_dna_json/dna-1.json`

himani2411 (approved):

PR #6030, opened 2024-01-24
[Tests] Delete common.yaml integration tests config file

enrico-usai:
### Description of changes

We have two different set of tests for released and develop version but the differences are minimal because both develop.yaml and released.yaml are importing the same common.yaml.

With this commit we are just removing the common.yaml file and duplicating the tests definition in develop.yaml and released.yaml file. When specific dimensions are overridden in one of the two files I'm keeping the existing dimension.

In this way when adding a test for released version we’ll be forced to evaluate if it’s really useful and properly define the dimensions for the different branches.

In a future commit will differentiate the dimensions set of tests and dimensions.

I also put the test-suites in alphabetical order.

### Tests

* develop: pre: 266 tests, after: 266 tests
* released: pre: 258 tests, after: 258 tests


NSsirena (approved):
General comment: 

The goal of this change is to remove obsolete tests from the released versions and the overall weight of these tests.

However we are still targeting the same regions, this will force us to run these tests sequentially ..increasing the time required for a full validation.

If we reserve a set of test region for testing the develop branch and a non-overlapping set of regions for testing the released version we could run the checks in parallel, potentially cutting the runtime in half.

PR #6031, opened 2024-01-24
[DFSM] Add custom resource to check that all compute nodes have the expected cluster config version.

gmarciani:
### Description of changes
1. Add custom resource to check that all compute nodes have the expected cluster config version.
This custom resource will be added only when using SLURM scheduler because compute nodes updates will be supported only with SLURM.
1. Added more checks on the cluster config version in integ test `test_update_slurm`.
1. Added a TODO in cleanup_resource to remind that it could use a new function introduced in this PR to reduce code duplication.

The creation of such DDB records has been introduced in https://github.com/aws/aws-parallelcluster-cookbook/pull/2618

### Tests
* Manually created, empty update (forced), deleted a cluster: all operations succeeded and the custom resource was executed as expected.
* Executed existing unit tests.
* Executed integ test `test_update_slurm`.

### DIFF
--- a/cli/src/pcluster/resources/custom_resources/custom_resources_code/utils/ec2_utils.py
+++ b/cli/src/pcluster/resources/custom_resources/custom_resources_code/utils/ec2_utils.py
@@ -1,0 +1,67 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+#  the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+import boto3
+from constants import CLUSTER_NAME_TAG, NODE_TYPE_TAG
+from utils.retry_utils import retry
+
+
+@retry(max_retries=5, wait_time_seconds=3)
+def list_cluster_instance_ids_iterator(
+    cluster_name: str,
+    instance_state: [str] = ("pending", "running", "stopping", "stopped"),
+    node_type: [str] = None,
+    batch_size: int = 100,
+    ec2_client=None,
+):
+    """
+    Generate an iterator over batch of cluster instances.
+
+    :param cluster_name: name of the cluster.
+    :param instance_state: list of instance states to include in the filter; default is None.
+    :param node_type: list of node types to include in the filter; default is None.
+    :param batch_size: the maximum size of the batch returned by the iterator; default is 100.
+    :param ec2_client: the EC2 client to use in the requests; default is a plain EC2 client with default settings.
+    :return: the iterator to iterate over batch of cluster instances.
+    """
+    ec2 = ec2_client if ec2_client is not None else boto3.client("ec2")
+
+    filters = _get_filters_to_list_instances(cluster_name, instance_state, node_type)
+
+    pagination_config = {"PageSize": batch_size}
+    paginator = ec2.get_paginator("describe_instances")
+
+    for page in paginator.paginate(Filters=filters, PaginationConfig=pagination_config):
+        instances = []
+        for reservation in page.get("Reservations", []):
+            for instance in reservation.get("Instances", []):
+                instances.append(instance.get("InstanceId"))
+        yield instances
+
+
+def _get_filters_to_list_instances(cluster_name: str, instance_state: [str] = None, node_type: [str] = None):
+    """
+    Generate a list of filters to be used in EC2 requests to filter cluster instances.
+
+    :param cluster_name: name of the cluster.
+    :param instance_state: list of instance states to include in the filter; default is None.
+    :param node_type: list of node types to include in the filter; default is None.
+    :return: the list of filters.
+    """
+    filters = [{"Name": f"tag:{CLUSTER_NAME_TAG}", "Values": [cluster_name]}]
+
+    if node_type:
+        filters.append({"Name": f"tag:{NODE_TYPE_TAG}", "Values": list(node_type)})
+
+    if instance_state:
+        filters.append({"Name": "instance-state-name", "Values": list(instance_state)})
+
+    return filters
--- a/cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
+++ b/cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
@@ -1,0 +1,193 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+#  the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import string
+
+import boto3
+from botocore.config import Config
+from constants import CLUSTER_CONFIG_DDB_ID
+from crhelper import CfnResource
+from utils.ec2_utils import list_cluster_instance_ids_iterator
+from utils.retry_utils import retry
+
+helper = CfnResource(json_logging=False, log_level="INFO", boto_level="ERROR", sleep_on_delete=0)
+logger = logging.getLogger(__name__)
+BOTO3_CONFIG = Config(retries={"max_attempts": 60})
+BATCH_SIZE = 500
+
+
+@retry(max_retries=5, wait_time_seconds=3)
+def _get_cluster_config_records(table_name: str, instance_ids: [string]):
+    ddb = boto3.client("dynamodb", config=BOTO3_CONFIG)
+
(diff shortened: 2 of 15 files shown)


cli/src/pcluster/resources/custom_resources/custom_resources_code/utils/ec2_utils.py
```diff
@@ -0,0 +1,68 @@
...
+import boto3
+from utils.retry_utils import retry
+
+
+@retry(max_retries=5, wait_time_seconds=3)
+def list_cluster_instance_ids_iterator(
+    cluster_name: str,
+    instance_state: List[str] = ("pending", "running", "stopping", "stopped"),
+    node_type: List[str] = None,
+    batch_size: int = 100,
```
enrico-usai, line 22: I'd say we should increase the batch_size. What about 1000? or 500?

cli/src/pcluster/resources/custom_resources/custom_resources_code/utils/ec2_utils.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import List
+
+import boto3
+from utils.retry_utils import retry
+
+
+@retry(max_retries=5, wait_time_seconds=3)
+def list_cluster_instance_ids_iterator(
+    cluster_name: str,
+    instance_state: List[str] = ("pending", "running", "stopping", "stopped"),
```
enrico-usai: minor: it's defined as a list but the default is a tuple.

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    for instance_ids in list_cluster_instance_ids_iterator(
+        cluster_name=cluster_name, node_type=["Compute"], batch_size=100, ec2_client=ec2
+    ):
+        n_instance_ids = len(instance_ids)
+
+        if not n_instance_ids:
+            logger.warning("Found empty batch of compute nodes: nothing to check")
+            continue
```
enrico-usai, line 131: what about having a sleep of 5-10 secs in this case? To avoid calling too many times the api

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,150 @@
...
+
+    if not instance_ids:
+        logger.warning("No instances to check cluster config version for")
+        return
+
+    for instance_id in instance_ids:
+        key = f"CLUSTER_CONFIG.{instance_id}"
+        data = items_by_id.get(key)
+        if data is None:
+            raise RuntimeError(f"Check Failed: missing cluster config record for instance {instance_id}")
```
enrico-usai: Rather than raising the RuntimeError at the first error we could collect all the errors and complete the instance_ids loop.

We can differentiate the error by type: missing record, missing config info, wrong config info and at the end raise the RuntimeError, listing all the failed instances in the log and the numbers of errors by type.

This could help to understand if it's a sporadic issue or is systematic, for all the instances or if there are mixed issues.

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,150 @@
...
+
+@retry(max_retries=5, wait_time_seconds=3)
+def _get_cluster_config_records(table_name: str, instance_ids: [string]):
+    ddb = boto3.client("dynamodb", config=BOTO3_CONFIG)
+
+    if not instance_ids:
+        logger.warning("No instances to retrieve cluster config records for")
+        return []
+
+    item_ids = [f"CLUSTER_CONFIG.{instance_id}" for instance_id in instance_ids]
```
enrico-usai: I'd say we should create a constants.py file to add on it pcluster tags, this format "CLUSTER_CONFIG...." and other common staff.

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,150 @@
...
+    requested_keys = [{"Id": {"S": item_id}} for item_id in item_ids]
+
+    try:
+        response = ddb.batch_get_item(RequestItems={table_name: {"Keys": requested_keys}})
```
enrico-usai, line 40: In the patch description let's link the other patch where we're adding these items in the DB table.

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,150 @@
...
+        items = response.get("Responses", {}).get(table_name, [])
+    except Exception as e:
+        raise RuntimeError(f"Cannot read config versions due to DDB error: {e}")
+
+    return items
+
+
+def _check_cluster_config_items(instance_ids, items, expected_config_version):
+    # Transform DDB items to make it easier to search
+    items_by_id = {item["Id"]["S"]: item["Data"]["M"] for item in items}
```
enrico-usai: we can add this translation after the guard for `not instance_ids`  and avoid doing this if we're just returning.

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
enrico-usai: or let's start creating a local one for `CLUSTER_CONFIG.`

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
enrico-usai: it would be useful to have an example of Item to understand the original structure and how we're converting it).

cli/src/pcluster/resources/custom_resources/custom_resources_code/utils/ec2_utils.py
gmarciani: Let's go with 500

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
gmarciani: Why do you think we would call the API too many times?
The only case when the batch is empty is on the first page

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
gmarciani: Very good point. I'd say that we should apply this approach only at batch level, not for the whole check so that we respect the fail fast principle. Do you agree?

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
gmarciani: Creating constants file with all these values. Thanks

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
gmarciani: Addressed

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,186 @@
...
+    #     "cluster_config_version": { "HoqyEZYBkMig3gSxaMARUv0NGcG0rrso" },
+    #     "lastUpdateTime": { "2024-01-16 18:59:18 UTC" }
+    #   }
+    # }
+    items_by_id = {item["Id"]["S"]: item["Data"]["M"] for item in items}
+
+    missing = incomplete = wrong = []
+
+    for instance_id in instance_ids:
+        key = CLUSTER_CONFIG_DDB_ID.format(instance_id)
```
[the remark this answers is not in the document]
gmarciani: Addressed

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
```diff
@@ -0,0 +1,186 @@
...
+    try:
+        response = ddb.batch_get_item(RequestItems={table_name: {"Keys": requested_keys}})
+        items = response.get("Responses", {}).get(table_name, [])
+    except Exception as e:
+        raise RuntimeError(f"Cannot read config versions due to DDB error: {e}")
+
+    return items
+
+
+def _check_cluster_config_items(instance_ids: [str], items: [{}], expected_config_version: str):
```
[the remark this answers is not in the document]
gmarciani: Addressed

cli/src/pcluster/resources/custom_resources/custom_resources_code/wait_cluster_ready.py
enrico-usai: You're right!

cli/tests/pcluster/resources/custom_resources/custom_resources_code/test_wait_cluster_ready.py
```diff
@@ -0,0 +1,162 @@
...
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import patch
+
+import pytest, os
```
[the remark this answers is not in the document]
gmarciani: addressed

enrico-usai (approved):

PR #4043, opened 2022-05-17
Add tests for build-image with kernel 4 images

francesco-giordano:
### Description of changes
* Since the base ami for Amazon Linux are based on kernel 5.10, this test verifies that the build-image still work with Amazon Linux based on kernel 4
* Add tests to build kernel 4 images from image-build and test the newly created images with a cluster

### Tests
* Manual tested with:
  * slurm and awsbatch scheduler
  * eu-west-1
  * INSTANCES_DEFAULT_X86
* Running on automated tests

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [x] Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6033, opened 2024-01-24
[Tests] Remove useless and old tests

enrico-usai:
# Description of changes

This patch removes a set of tests that are useless because already covered by Kitchen tests, or because they were useful in past but less relevant in this moment.
Removed:
* 4 test_arm_pl tests (4 OSes) --> covered by kitchen
* 1 test_kernel4_build_image_run_cluster --> old kernel, replaced on 2021
* 1 test_scontrol_update_nodelist_sorting --> old Slurm version bug
* 1 test_slurm_reconfigure_race_condition --> old Slurm version bug
* 1 test_api - test_cluster_awsbatch
* 1 test_update_awsbatch

Details below. It's easy to look at the changes commit by commit.

## [Tests] Remove tests for ARM Performance Library

These checks have been moved to Kitchen tests for ARM PL chef custom resource
and executed at AMI build time. There is 100% overlapping.

`tag:install_arm_pl_installed`:
* verifies ARM PL library is correctly installed
* checks version of ARM PL according to the OS
* loads armpl module and gcc module and assert the module is loaded
* verifies EULA docs are correctly linked in the module output
* tests ARM PL examples

`tag:install_arm_pl_gcc_installed`:
* verifies GCC is correctly installed
* checks version of ARM PL and GCC according to the OS

### References

* https://github.com/aws/aws-parallelcluster-cookbook/blob/release-3.8/cookbooks/aws-parallelcluster-platform/test/controls/arm_pl_spec.rb

## [Tests] Remove test for build AMI with Kernel 4

This test has been introduced to verify the build-image with Amazon Linux based on kernel 4,
because the base AMI for Amazon Linux were based on kernel 5.10.

At the moment this test is no longer relevant,
kernel 5.10 in Amazon Linux 2 has been introduced on Nov 2021 and kernel 4.14 is now EOL.

### References

* https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-linux-2-ami-kernel-5-10/
* https://github.com/aws/aws-parallelcluster/pull/4043
* https://en.wikipedia.org/wiki/Linux_kernel_version_history

## [Tests] Remove test for dynamics nodes not powering down after an update

In Slurm 21.08 cloud nodes were not get powered-down after their
SuspendTime has expired if a cluster update is performed, which restarts the
slurmctld daemon and immediately performs a scontrol reconfigure.

This was causing a race condition between restart of slurmctld and scontrol reconfigure.

The issue has been fixed in Slurm 22.05.7 release, included in Pcluster 3.4.0, so this test is no longer useful.

### References

* https://github.com/aws/aws-parallelcluster/pull/4746
* https://bugs.schedmd.com/show_bug.cgi?id=13953

## [Tests] Remove test for scontrol update nodelist sorting bug

In Slurm 22.05.7 the `scontrol update node` logic was modified and a sorting routine was
introduced, which modified the order of the nodes in the nodelist,
this was causing mismatches between the Slurm nodenames and the assigned addresses.

The issue has been fixed in 22.05.8 release, included in Pcluster 3.5.0 so this test is no longer useful.

### References

* https://github.com/aws/aws-parallelcluster/pull/4786
* https://bugs.schedmd.com/show_bug.cgi?id=15731

## [Tests] Remove test for Pcluster API with AWS Batch as a scheduler

# Tests

* develop: pre: 266 tests, after: 262 tests
* released: pre: 258 tests, after: 255 tests

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -818,11 +818,6 @@ test-suites:
           schedulers: ["slurm"]
           oss: ["ubuntu2004"]
   update:
-    test_update.py::test_update_awsbatch:
-      dimensions:
-        - regions: ["eu-south-1"]
-          instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux2"]
     test_update.py::test_update_slurm:
       dimensions:
         - regions: ["eu-central-1"]
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -798,12 +798,6 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: {{ common.NOT_RELEASED_OSES }}
           schedulers: ["slurm"]
-    test_api.py::test_cluster_awsbatch:
-      dimensions:
-        - regions: ["sa-east-1"]
-          instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux2"]
-          schedulers: ["awsbatch"]
     test_api.py::test_custom_image:
       dimensions:
         - regions: ["sa-east-1"]
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -578,12 +578,6 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["ubuntu2004"]
           schedulers: ["slurm"]
-    test_slurm.py::test_scontrol_update_nodelist_sorting:
-      dimensions:
-        - regions: ["ca-central-1"]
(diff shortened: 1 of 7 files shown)


NSsirena:
Please do not remove tests from `Schedulers.yaml` this is an internal configuration file that is managed by the Schedulers workstream. We are using this configuration file to validate upgrades to Slurm versions and we want to keep here also "old" tests to check for possible regressions.

enrico-usai:
> Please do not remove tests from `Schedulers.yaml` this is an internal configuration file that is managed by the Schedulers workstream. We are using this configuration file to validate upgrades to Slurm versions and we want to keep here also "old" tests to check for possible regressions.

Thanks for the comment. I restored `test_slurm_reconfigure_race_condition` and `test_scontrol_update_nodelist_sorting` in schedulers.yaml file.

NSsirena (approved):

PR #6038, opened 2024-01-25
[Tests] Reduce tests dimensions for released version

enrico-usai:
### Description of changes

With this change we're reducing tests dimensions for released version.
* when the test is verifying different OSes or architecture dimensions we're reducing dimensions keeping a single test as "AWS Api canary" (AD, CW, DCV, EFA).
* when the test is verifying our code logic we are removing it because, once released our code and AMIs are stable so it's useless to test them.

Released version test changes (removes 30 tests):
- Reduce dimensions of AD integration tests (4 --> 1)
- Reduce dimensions of CloudWatch logging tests (6 --> 1)
- Remove test of pcluster configure with bad subnets (1 --> 0)
- Remove test of pcluster configure for a region without t2.micro (1 --> 0)
- Remove test of pcluster configure for EFA/PG settings (1 --> 0)
- Remove test of pcluster configure when EFA is not supported (1 --> 0)
- Remove DCV tests for different ports and access-from (1 --> 0)
- Remove cluster creation tests verifying internal logic (3 --> 0)
- Reduce dimensions of DCV tests (4 --> 1)
- Reduce EFA tests dimensions (4 --> 1)
- Remove GPU health check test (1 --> 0)
- Remove IAM Policy test(2 --> 0)
- Remove IAM Roles test (4 --> 0)
- Remove Intel HPC test (1 --> 0)

New instance type test changes:
- Reduce EFA tests dimensions when testing a new instance type (3 --> 1)

Details in the commit messages.

### Tests
* released: pre: 255 tests, after: 236 tests

### DIFF
--- a/tests/integration-tests/configs/new_instance_types.yaml
+++ b/tests/integration-tests/configs/new_instance_types.yaml
@@ -77,7 +77,7 @@ test-suites:
       dimensions:
         - regions: {{ NEW_REGIONS }}
           instances: {{ NEW_INSTANCE_TYPES }}
-          oss: ["alinux2", "rhel8", "ubuntu2004"]
+          oss: ["ubuntu2004"]
           schedulers: ["slurm"]
   configure:
     test_pcluster_configure.py::test_pcluster_configure:
--- a/tests/integration-tests/configs/released.yaml
+++ b/tests/integration-tests/configs/released.yaml
@@ -6,11 +6,7 @@ test-suites:
       dimensions:
         - regions: [ "ap-southeast-1" ]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux2", "rhel8"]
-          schedulers: ["slurm"]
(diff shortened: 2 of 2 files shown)


NSsirena (approved):
The comment report a difference of 30 test but in the Test section is reported 

```
released: pre: 258 tests, after: 239 tests
```

that amounts to 19 less test. So there is a difference of 11 tests.
Changes are in line with the decision doc so I assume this difference is either a typo or a weird behavior in the mechanism that counts the tests.

enrico-usai:
> Changes are in line with the decision doc so I assume this difference is either a typo or a weird behavior in the mechanism that counts the tests.

Yes, the numbers are collected by pytest but they are not correct. My guess is that it depends by test dimensions or something else. Anyway I'm using the `dryrun` execution just to be sure syntax of integ test config file is correct.

Just updated the values after the rebase:
* released: pre: 255 tests, after: 236 tests

PR #6052, opened 2024-01-30
[Tests] Add new configuration file to be used during release process

enrico-usai:
### Description of changes

With first test we're verifying that aws-parallelcluster package is available in the 3 partitions, that is possible to create a cluster and essential features are working.

With API test we're verifying that ECR image required for the API is available in the 3 partitions, and it's possible to communicate with it.

### DIFF
--- a/tests/integration-tests/configs/release-process.yaml
+++ b/tests/integration-tests/configs/release-process.yaml
@@ -1,0 +1,18 @@
+{%- import 'common.jinja2' as common with context -%}
+---
+test-suites:
+  basic:
+    # Verify aws-parallelcluster package is available in the 3 partitions,
+    # is possible to create a cluster and essential features are working
+    test_essential_features.py::test_essential_features:
+      dimensions:
+        - regions: ["ap-south-1", "cn-north-1", "us-gov-west-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["rhel8"]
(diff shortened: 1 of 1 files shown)


tests/integration-tests/configs/release-process.yaml
```diff
@@ -0,0 +1,14 @@
+{%- import 'common.jinja2' as common with context -%}
+---
+test-suites:
+  basic:
+    test_essential_features.py::test_essential_features:
```
NSsirena, line 7: Can we add a comment here explaining that 

`we're verifying that aws-parallelcluster package is available in the 3 partitions, that is possible to create a cluster and essential features are working`

tests/integration-tests/configs/release-process.yaml
```diff
@@ -0,0 +1,14 @@
...
+      dimensions:
+        - regions: ["ap-south-1", "cn-north-1", "us-gov-west-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["rhel8"]
+          schedulers: ["slurm"]
+  pcluster_api:
+    test_api_infrastructure.py::test_api_infrastructure_with_default_parameters:
```
NSsirena, line 16: and here 
`we're verifying that ECR image required for the API is available in the 3 partitions, and it's possible to communicate with it`

NSsirena (approved):

PR #6064, opened 2024-01-31
Allow AD stacks to be detected and used by another test execution

hgreebe:
### Description of changes
* Detects existing AD stacks based on tags so that they can be used by another test execution rather than creating a new AD
* Adds a --retain-ad-stack option that allows and ad stack and corresponding vpc stack to be retained after the integration test runs
* Uses the subnets from the VPC stack of the AD for the cluster
* Removed unused _generate_certificate function

### Tests
* Ran locally and on personal Jenkins


tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -39,8 +36,10 @@
 NUM_USERS_TO_CREATE = 5
 NUM_USERS_TO_TEST = 3
 
-ADMIN_PASSWORD = "".join(random.choices(string.ascii_letters + string.digits, k=60))
-USER_PASSWORD = "".join(random.choices(string.ascii_letters + string.digits, k=60))
+# ADMIN_PASSWORD = "".join(random.choices(string.ascii_letters + string.digits, k=60))
+# USER_PASSWORD = "".join(random.choices(string.ascii_letters + string.digits, k=60))
+ADMIN_PASSWORD = "TestPass123"
```
gmarciani: We can keep random password. To do that, we must retrieve the password secret from the stack output and read the password at runtime

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -253,15 +236,40 @@ def _directory_factory(
...
-                cfn_stacks_factory,
-                request,
-                directory_type,
-                region,
-                vpc_stack,
-            )
-            directory_stack_name = directory_stack.name
-            created_directory_stacks[region]["directory"] = directory_stack_name
+            directory_stack_name = ""
+            cfn_client = boto3.client("cloudformation", region_name=region)
```
gmarciani: Let's define the logic to retrieve a stack by tag in a dedicate function in a utility script that can be used by other tests. In the near future we may need to reuse this logic to retrieve other pre-existing auxiliary stacks.

tests/integration-tests/cfn_stacks_factory.py
```diff
@@ -307,6 +307,21 @@ def delete_all_stacks(self):
                     )
                 )
 
+    def delete_stacks_retain_ad(self):
+        """Destroy all created stacks."""
+        logging.debug("Destroying all cfn stacks except for AD and VPC stacks")
+        for value in reversed(OrderedDict(self.__created_stacks).values()):
+            if "vpc" in value.name or "SimpleAD" in value.name or "MicrosoftAD" in value.name:
```
gmarciani: Minor: in short we could do `if value.name in ["vpc", "SimpleAD", "MicrosoftAD"]`

tests/integration-tests/cfn_stacks_factory.py
```diff
@@ -307,6 +307,21 @@ def delete_all_stacks(self):
                     )
                 )
 
+    def delete_stacks_retain_ad(self):
+        """Destroy all created stacks."""
```
gmarciani: This function destroyes all stack but VPC, and AD stacks. The description should be aligned with this.

tests/integration-tests/conftest.py
```diff
@@ -156,6 +156,9 @@ def pytest_addoption(parser):
     parser.addoption(
         "--no-delete", action="store_true", default=False, help="Don't delete stacks after tests are complete."
     )
+    parser.addoption(
+        "--retain-ad-stack", action="store_true", default=False, help="Retain AD stack and corresponding VPC stack."
```
gmarciani, line 160: Let's include this new option in the README where all test runner options are documented.

tests/integration-tests/cfn_stacks_factory.py
```diff
@@ -307,6 +307,21 @@ def delete_all_stacks(self):
                     )
                 )
 
+    def delete_stacks_retain_ad(self):
```
gmarciani: I suggest to not have a separate function. Let us imagine if tomorrow we should do the same for N other auxiliary stacks to be retained: with this approach we would have N new functions.

I suggest to extend the logic of the original delete_all_stacks by adding an optional parameter, such as "excluded_stacks" (a list).

tests/integration-tests/utils.py
```diff
@@ -864,3 +864,24 @@ def retrieve_clustermgtd_conf_path(remote_command_executor):
 def disable_protected_mode(remote_command_executor):
     """Disable protected mode by setting protected_failure_count to -1."""
     set_protected_failure_count(remote_command_executor, -1)
+
+
+def find_stack_by_tag(tag, region, stack_prefix):
+    cfn_client = boto3.client("cloudformation", region_name=region)
```
gmarciani, line 871: Minor: let's document the function

tests/integration-tests/conftest.py
```diff
@@ -855,7 +858,11 @@ def cfn_stacks_factory(request):
     factory = CfnStacksFactory(request.config.getoption("credential"))
     yield factory
     if not request.config.getoption("no_delete"):
-        factory.delete_all_stacks()
+        if request.config.getoption("retain_ad_stack"):
```
gmarciani, line 862: Minor: we can simplify and make it ready for extensions.

```
excluded_stacks = []
if request.config.getoption("retain_ad_stack"):
   excluded_stacks.append(.....)
factory.delete_all_stacks(excdluded_stacks)
```

In this way you avoid repetition of the delate_all_Stacks function and when we will add a new retention it will only be a matter of adding elements in the excluded_stacks

tests/integration-tests/conftest.py
```diff
@@ -855,7 +858,11 @@ def cfn_stacks_factory(request):
     factory = CfnStacksFactory(request.config.getoption("credential"))
     yield factory
     if not request.config.getoption("no_delete"):
-        factory.delete_all_stacks()
+        excluded_stacks = []
+        if request.config.getoption("retain_ad_stack"):
+            excluded_stacks.append(["vpc", "SimpleAD", "MicrosoftAD"])
```
gmarciani: Warning: if you append the list, then excluded_stacks will be a list of lists. You have two alternatives:
1. my_array.append(element1) my_array.append(element2)
2. my_array += [element1, element2]
3. my_array.extend([element1, element_2])

gmarciani (approved):

PR #6091, opened 2024-02-09
Add DO-NOT-DELETE tag to VPC and AD stacks when retain-ad-stack option is selected

hgreebe:
### Description of changes
* Add DO-NOT-DELETE tag to VPC and AD stacks when retain-ad-stack option is selected
* Checks for DO-NOT-DELETE tag during cloudformation stack cleanup

### Tests
* Tested on personal Jenkins

### References
* Patch for: https://github.com/aws/aws-parallelcluster/pull/6064


tests/integration-tests/conftest.py
```diff
@@ -858,11 +858,7 @@ def cfn_stacks_factory(request):
     factory = CfnStacksFactory(request.config.getoption("credential"))
     yield factory
     if not request.config.getoption("no_delete"):
-        excluded_stacks = []
-        if request.config.getoption("retain_ad_stack"):
-            excluded_stacks.extend(["vpc", "SimpleAD", "MicrosoftAD"])
-            logging.warning("Skipping deletion of AD and VPC stacks because --retain-ad-stack option is set")
-        factory.delete_all_stacks(excluded_stacks)
+        factory.delete_all_stacks()
```
gmarciani: Not clear to me why we are removing the possibility to delete all stacks except fro the excluded ones.
Now that you've introduced the possibility to have the DO_NOT_DELETE tag, why not keeping the logic and include stacks with that tag in the excluded ones?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -195,14 +207,17 @@ def _create_directory_stack(cfn_stacks_factory, request, directory_type, region,
 
     with open(directory_stack_template_path) as directory_stack_template:
         stack_parameters = _get_stack_parameters(directory_type, vpc_stack, request.config.getoption("key_name"))
+        tags = [{"Key": "parallelcluster:integ-tests-ad-stack", "Value": directory_type}]
+        if request.config.getoption("retain_ad_stack"):
+            tags.append({"Key": "DO-NOT-DELETE", "Value": " "})
```
gmarciani: I suggest to use the value to specify the reason why we should not delete it. Could also be as simple as "Retained for integration testing". This would be helpful when someone not aware of your mechanism will see retained stack with this new tag

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -255,7 +270,8 @@ def _directory_factory(
                 )
                 directory_stack_name = directory_stack.name
                 created_directory_stacks[region]["directory"] = directory_stack_name
-
+                if request.config.getoption("retain_ad_stack"):
+                    add_tag_to_stack(vpc_stack.name, "DO-NOT-DELETE", " ")
```
gmarciani: Why do we need to update an existing stack with the tag?
I think this mechanism should be applied only to the brand new stack, not on existing ones.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -120,6 +120,18 @@ def _add_file_to_zip(zip_file, path, arcname):
         zip_file.writestr(zinfo, input_file.read())
 
 
+def add_tag_to_stack(stack_name, key, value):
+    cfn = boto3.resource("cloudformation")
+    stack = cfn.Stack(stack_name)
+    stack.update(
```
gmarciani: This implies a stack update operation that takes time. Better to avoid it just to add a tag if not strictly necessary. Why do we need to update a stack with tags. The mechanism should be applied only on brand new stack not on existing one, I guess.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hgreebe: the vpc stack is created as a fixture, so I does not have the tag. So if an existing active directory is not found, then a new one is created and a DO-NOT-DELETE tag is added to the corresponding vpc stack.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hgreebe: The tag is added to the new AD stacks when they are created, but the vpc stack has already been created without the tag, so that is why if we create new AD stacks, we need to update the vpc stack tags.

tests/integration-tests/conftest.py
```diff
@@ -1642,6 +1642,9 @@ def _fsx_factory(ports, ip_protocols, file_system_type, num=1, **kwargs):
 
         # Create security group. If using an existing file system
         # It must be associated to a security group that allows inbound TCP/UDP traffic to specific ports
+        if not vpc:
+            vpc = vpc_stack.cfn_outputs["VpcId"]
+            subnet = vpc_stack.get_public_subnet()
```
judysng, line 1647: Minor: I would suggest to split 
```
if not vpc:
            vpc = vpc_stack.cfn_outputs["VpcId"]
            subnet = vpc_stack.get_public_subnet()
```
into
```
if not vpc:
            vpc = vpc_stack.cfn_outputs["VpcId"]
if not subnet:
            subnet = vpc_stack.get_public_subnet()
```
just for consistency

judysng (approved):

PR #6111, opened 2024-02-21
Add wait time to allow time for job state to switch to pending

hgreebe:
### Description of changes
* Fix test_queue_parameters_update by adding 1 minute wait time to allow time for job state to switch to pending

### Tests
* Ran test on Jenkins


judysng (approved):

tests/integration-tests/tests/update/test_update.py
```diff
@@ -925,6 +925,7 @@ def _test_update_queue_strategy_with_running_job(
         # requeue job in queue2 to launch new instances for nodes
         remote_command_executor.run_remote_command(f"scontrol requeue {queue2_job_id}")
     elif queue_update_strategy == "TERMINATE":
+        time.sleep(60)
```
gmarciani, line 936: minor: it's always preferable to wrap a check within a retrying block than adding a fixed max time. For example here we could use a retry with 6 max attempts with 10 seconds delay. In this way you would wait at most 60seconds but you can potentially save 50s.

PR #6169, opened 2024-03-20
Remove wait for shutting-down instances

lukeseawalker:
### Description of changes
During cluster deletion, remove wait for shutting-down instances. The shutting-down state is not recoverable, and instances in this state are not billed. This change will fix the situation of cluster deletion failure when termination gets longer for some instance types

Ref. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html

### Tests
n/a

### References
n/a

### Checklist


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ CHANGELOG
 
 **BUG FIXES**
 - Fix DRA configuration to make `AutoExportPolicy` and `AutoImportPolicy` optional.
+- Consider Compute fleet clean-up completed during cluster deletion when instances are either in shutting-down or terminated state.
+  This to avoid cluster deletion failure for instance types with longer termination cycle.
```
dreambeyondorange: nit; "This is to" and "longer termination cycles"

cli/src/pcluster/resources/custom_resources/custom_resources_code/cleanup_resources.py
```diff
@@ -120,30 +120,12 @@ def _terminate_cluster_nodes(event):
             logger.info("Sleeping for 10 seconds to allow all instances to initiate shut-down")
             time.sleep(10)
```
dreambeyondorange: do we still need this?

lukeseawalker: I think it's to avoid throttling in between for loop cycle of terminate calls

lukeseawalker: no, I thought it was to cope with throttling, but that is already taken care by the for loop

dreambeyondorange (approved):

PR #6179, opened 2024-03-27
Output bootstrap errors in scaling test

judysng:
### Description of changes
* Adding a check for every scale up for any bootstrap errors in clustermgtd. If there are (very unlikely with the latest versions), get the latest console logs for that associated instance because it's possible that the bootstrap error would happen before the CWAgent has been configured (and thus no logs in Cloudwatch yet to check). The logs are visible in the test outputs in Jenkins
* There are some nuances with getting the IP address and then the associated instance id because we reuse the same cluster for multiple scale ups (and thus the info from previous scale ups will stay in clustermgtd when we query it for the bootstrap errors)
  * The node bootstrap error message in the clustermgtd gives the IP address, not the instance id. So then we need to handle getting the correct instance id from the IP address.
  * In the code, the bootstrap error code is ran for each scale up and the bootstrap error console logs are recorded for that IP address. Then if those logs are there, there will be no further checks on that IP address for future scale ups
    * This is to avoid getting the logs of an instance that doesn't have a bootstrap error (since the IP address will be flagged for every scale-up). The problem this poses would be if in one scale-up there is a bootstrap error with that IP address and then in a following scale-up there is another error with the same IP address so those console logs wouldn't be recorded, but this would be very rare.
  * The code gets the logs of the latest launched instance if there are multiple instances associated with the same IP address. 
    * This is to avoid getting the logs of an instance with the same IP address from a previous scale-up, but assumes that the latest launched instance is the one with the bootstrap error.
  * Since it is very rare to get a bootstrap error in the first place, these nuances should not matter much

### Tests
* Ran the test with a few P3 instances with v3.8.0 in order to get the known issue of the gdrcopy bootstrap failure. Logs are in the test artifacts in Jenkins
* Ran the test regularly and worked fine (but there were no bootstrap failures)

### DIFF
--- a/tests/integration-tests/tests/common/scaling_common.py
+++ b/tests/integration-tests/tests/common/scaling_common.py
@@ -11,6 +11,7 @@
 import datetime
 import json
 import logging
+import os
 import pathlib
 import time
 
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/common/scaling_common.py
```diff
@@ -64,6 +65,56 @@ def retry_if_scaling_target_not_reached(
...
+    for ip_address in ip_addresses_with_bootstrap_errors.splitlines():
+        # Since the same cluster is re-used for multiple scale up tests, the script may find the same bootstrap error
+        # multiple times and then get the wrong instance logs since the IP address would be attached to a new instance.
+        # Therefore, only write the compute node logs for the IP address if the file doesn't exist yet.
+        if _check_no_node_log_exists_for_ip_address(path, ip_address):
+            try:
+                logging.warning(f"Compute node with IP {ip_address} had bootstrap errors. Getting instance id...")
+                # Get the latest launched instance with the IP address since the most recent one should have the error
+                instance_id = _sort_instances_by_launch_time(
+                    client.describe_instances(Filters=[{"Name": "private-ip-address", "Values": [ip_address]}])
```
hanwen-pcluste: It is better to use pagination here to ensure all results are retrieved

hanwen-pcluste (dismissed):

hanwen-pcluste (approved):

PR #6201, opened 2024-04-09
[Tests] Move test_mpi to essential features

hgreebe:
### Description of changes
* Merge test_mpi with test_mpi_ssh
* Move test_mpi to test_essential_features

### Tests
* Ran integration test on personal Jenkins with the following config:
```
---
test-suites:
  basic:
      dimensions:
        - regions: ["us-east-1"]
          oss: ["alinux2"]
          schedulers: ["slurm"]
```


tests/integration-tests/tests/common/mpi_common.py
```diff
@@ -60,7 +60,7 @@ def _test_mpi(
 
     if verify_scaling:
         assert_scaling_worked(
-            scheduler_commands, region, stack_name, scaledown_idletime, expected_max=3, expected_final=0
+            scheduler_commands, region, stack_name, scaledown_idletime, expected_max=5, expected_final=2
```
himani2411, line 63: why is the Max and final node counts changing?

hgreebe: Because in the cluster config file there are two queues that have a MinCount of 1, so the expect_final count of compute nodes will never go below 2. So the increase in number of computes nodes stays the same (3) because 3 dynamic nodes were added, that is added on top of the 2 static nodes that already exist.

himani2411 (approved):

PR #6233, opened 2024-05-02
Print warning to stdout

stephenmsachs:
### Description of changes
* This enables the use of `pcluster dcv-connect -n <cluster> | pbcopy` to copy the url directly into macOS paste buffer.

### DIFF
--- a/cli/src/pcluster/cli/commands/dcv_connect.py
+++ b/cli/src/pcluster/cli/commands/dcv_connect.py
@@ -64,10 +64,11 @@ def _dcv_connect(args):
 
         try:
             url = _retry(_retrieve_dcv_session_url, func_args=[cmd, args.cluster_name, head_node_ip], attempts=4)
-            url_message = f"Please use the following one-time URL in your browser within 30 seconds:\n{url}"
+            url_message = f"Please use the following one-time URL in your browser within 30 seconds:"
 
             if args.show_url:
-                print(url_message)
+                print(url_message, file=sys.stderr)
+                print(f"{url}")
                 return
 
             try:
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/cli/commands/dcv_connect.py
```diff
@@ -64,10 +64,11 @@ def _dcv_connect(args):
 
         try:
             url = _retry(_retrieve_dcv_session_url, func_args=[cmd, args.cluster_name, head_node_ip], attempts=4)
-            url_message = f"Please use the following one-time URL in your browser within 30 seconds:\n{url}"
+            url_message = f"Please use the following one-time URL in your browser within 30 seconds:"
```
gmarciani, line 67: There is a test that should be changed accordingly: https://github.com/aws/aws-parallelcluster/blob/ea57725c46392c175aa66a2c8a8fcc9e8e7ec2a2/tests/integration-tests/tests/dcv/test_dcv.py#L76-L79

gmarciani, line 67: With this change we are removing an explicative message from stdout to make it easier to consume the url.
What if this command returns a JSON, such as:

```
{
    "message": "Please use the following one-time URL in your browser within 30 seconds",
    "url": "https://...."
}
```

This would be consistent with the other pcluster commands returning JSON and would make possible to consume the url easily with jq or, even better, with the --query option of the pcluster CLI.

What do you think?

stephenmsachs: Sounds good to me. Please feel free to change it.

PR #6243, opened 2024-05-20
[Test] Add integ tests to check that dynamic file systems updates does not cause data loss.

gmarciani:
### Description of changes
Add integ tests `test_dynamic_file_systems_update_data_loss` to check that dynamic file systems updates does not cause data loss.

The test executes the following steps:
    1. Creates the external shared storage stack
    2. Creates a cluster with all the external shared storage
    3. Writes a file in every external shared storage mount dir
    4. Updates the cluster to unmount all the external shared storage
    5. Updates the cluster to re-mount all the external shared storage
    6. Verifies that all files previously written where still there

### Tests
* Integ tests `test_dynamic_file_systems_update_data_loss` 

### References

### Checklist

### DIFF
--- a/tests/integration-tests/tests/update/test_update.py
+++ b/tests/integration-tests/tests/update/test_update.py
@@ -15,6 +15,7 @@ import os.path as os_path
 import re
 import time
 from collections import defaultdict
+from datetime import datetime
 
 import boto3
 import pytest
@@ -33,6 +34,7 @@ from utils import (
     get_arn_partition,
     get_root_volume_id,
     is_fsx_supported,
+    random_alphanumeric,
     retrieve_cfn_resources,
     wait_for_computefleet_changed,
 )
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/update/test_update.py
```diff
@@ -1529,6 +1533,113 @@ def test_dynamic_file_systems_update_rollback(
     )
 
 
+@pytest.mark.usefixtures("instance")
+def test_dynamic_file_systems_update_data_loss(
```
dreambeyondorange, line 1537: Instead of a new integ test, can this logic be added to an existing one?

gmarciani: This was my initial thought. I changed my mind for the following reasons:
1. **time:** the main DFSM test is already long (~2h); including this test would increase that time by ~15min. We recently invested effort in reducing the main DFSM integ test; including data loss test would go in the opposite direction
2. **complexity:** DFSM integ test is already complicated: iterating on new logic is frustrating because you need to comments out parts here and there to troubleshoot issues.

What do you think?

dreambeyondorange: There seem to be several competing priorities for integration tests and we don't have clear direction on what the right path is.  I think it makes sense to have separate tests since they will likely be run in parallel and eventually I think we want to use shared resources for these tests so they eventually consume less time overall.  But there was also a push to consolidate tests to have fewer of them, so it's not clear where the line is.  I think I will disagree and commit on this one and we need to take a long hard look at integ tests again to decide what our priorities really are.

dreambeyondorange (approved):

PR #6266, opened 2024-05-29
Removing usage of `/etc/profile.d/pcluster.sh`

himani2411:
### Description of changes
* Using cfn_bootstrap_virtualenv_path directly for cfn commands

### Tests
* COOKBOOK https://github.com/aws/aws-parallelcluster-cookbook/pull/2742

### References

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ CHANGELOG
   This is to avoid cluster deletion failure for instance types with longer termination cycles.
 - Allow cloudwatch dashboard to be enabled and alarms to be disabled in the `Monitoring` section of the cluster config.
 - Allow ParallelCluster Custom Resource to suppress validators using `PclusterCluster/SuppressValidators`.
+- Removing `/etc/profile.d/pcluster.sh` so that it's not executed at every user login and
+  `cfn_bootstrap_virtualenv` is not added in PATH environment variable.
 
 **CHANGES**
 - Upgrade Cinc Client to version to 18.4.12 from 18.2.7.
--- a/cli/src/pcluster/resources/head_node/user_data.sh
+++ b/cli/src/pcluster/resources/head_node/user_data.sh
@@ -70,7 +70,7 @@ function error_exit
   # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html
   cutoff=$(expr 4096 - $(stat --printf="%s" /tmp/wait_condition_handle.txt))
   reason=$(head --bytes=${!cutoff} /var/log/parallelcluster/bootstrap_error_msg 2>/dev/null) || reason="$1"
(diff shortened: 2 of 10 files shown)


cli/src/pcluster/constants.py
```diff
@@ -337,3 +337,5 @@ class Operation(Enum):
 IAM_ROLE_REGEX = "^arn:.*:role/"
 IAM_INSTANCE_PROFILE_REGEX = "^arn:.*:instance-profile/"
 IAM_POLICY_REGEX = "^arn:.*:policy/"
+
+CFN_BOOTSTRAP_VIRTUALENV_PATH = "/opt/parallelcluster/pyenv/versions/3.9.17/envs/cfn_bootstrap_virtualenv/bin"
```
enrico-usai: I don't like the fact we're adding in the cli the relation with the virtualenv created by the cookbook and with python version used by the cookbook too.

What we can do is to continue to the profile.sh file, renaming it to `cookbook_environment.sh` or something like this and rather than setting the `PATH`  into it, define an `export CFN_BOOTSTRAP_VIRTUALENV_PATH=something` on it.

So in the cli we need to source the `cookbook_environment.sh` and then simply use the `CFN_BOOTSTRAP_VIRTUALENV_PATH` that will be exported in the environment.

In this way the cookbook can change the path, the python version etc in a safe way, without impacting the cli code.

himani2411: I like this suggestion, will update and test accordingly

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
 - Consider Compute fleet clean-up completed during cluster deletion when instances are either in shutting-down or terminated state.
   This is to avoid cluster deletion failure for instance types with longer termination cycles.
 - Allow cloudwatch dashboard to be enabled and alarms to be disabled in the `Monitoring` section of the cluster config.
+- Renaming pcluster.sh profile to cookbook_environment.sh profile which will be used only for cfn commands.
```
gmarciani: Here we should explain the change from the user perspective. What's the consequence of this change for the user?

cli/src/pcluster/resources/head_node/user_data.sh
```diff
@@ -94,10 +94,10 @@ function vendor_cookbook
 # deploy config files
 export PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin
 # Load ParallelCluster environment variables
-[ -f /etc/profile.d/pcluster.sh ] && . /etc/profile.d/pcluster.sh
+[ -f /etc/profile.d/cookbook_environment.sh ] && . /etc/profile.d/cookbook_environment.sh && echo $CFN_BOOTSTRAP_VIRTUALENV_PATH
```
gmarciani: I think echoing the envar here was for debugging purpose. Can we remove the echo?

gmarciani (dismissed):

gmarciani (approved):

Issue #2337, opened 2020-12-28
Access points support in EFS


andrei-xdlab:
Hello,

I would like to replace slow EBS storage with EFS filesystem for hundreds of nodes pcluster. Do you support EFS Access points in 2.10.1? I have existing EFS filesystem and would like to create multiple access points (/home, /shared, /app etc) and mount shared on master/compute nodes. Do you plan to have a native support for EFS's POSIX user/permission?

Thanks

demartinofra:
Hi Andrei,

unfortunately at the moment ParallelCluster does not offer native support for EFS Access Points. I'm going to mark this as a feature request so that we take this into account when extending EFS functionalities in ParallelCluster.
Until then you could use a [custom post-install script](https://docs.aws.amazon.com/parallelcluster/latest/ug/pre_post_install.html) to mount an existing EFS with Access Points enabled.

Francesco

andrei-xdlab:
Hello Francesco,

Thank you for opening feature request. 
- ability to include existing access point IDs (fsap-xxx) in the template to mount multiple access points. 
- ability to create access EFS and access points and create directory permissions/ownership

[efs efscustom]
efs_fs_id = fs-xxxx
fsap_id1 = fsap-xxxx
shared_dir = /home
fsap_id2 = fsap-xxxx
shared_dir = /app

andrei-xdlab:
I created custom AMI (2.10.1) to include /home and /shared EFS access points in /etc/fstab. My pcluster create is failing due to nfs_export failure. (cfn-init.log below)

fs-xxxx /home efs _netdev,tls,accesspoint=fsap-0xxxx 0 0
fs-xxxx /shared efs _netdev,tls,accesspoint=fsap-0xxx 0 0

How do I disable creation of EBS volume and NFS exports on master node? I will have EFS access points mounted across master and all compute nodes.

On master instance I see that /home access point but /shared created from EBS volume

/dev/xvdb        20G   45M   19G   1% /shared
127.0.0.1:/     8.0E     0  8.0E   0% /home

 /var/log/cfn-init.log
 ================================================================================
    Error executing action `create` on resource 'nfs_export[/home]'
    ================================================================================
    
    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
[...]

PR #6359, opened 2024-07-19
Add EFS access point support (aws-parallelcluster#2337)

thheinen:
### Description of changes
Add support to specify EFS access points per issue aws/aws-parallelcluster#2337

### Tests
Tested mounting EFS with and without accesspoints from an aws-parallelcluster cluster.

**More tests to be done but opening a draft PR for discussion and early feedback**

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -371,7 +371,7 @@ class SharedEfs(Resource):
         deletion_policy: str = None,
         encryption_in_transit: bool = None,
         iam_authorization: bool = None,
-        accesspoint_id: str = None,
+        access_point_id: str = None,
     ):
         super().__init__()
         self.mount_dir = Resource.init_param(mount_dir)
@@ -388,7 +388,7 @@ class SharedEfs(Resource):
         )
         self.encryption_in_transit = Resource.init_param(encryption_in_transit, default=False)
         self.iam_authorization = Resource.init_param(iam_authorization, default=False)
-        self.accesspoint_id = Resource.init_param(accesspoint_id)
+        self.access_point_id = Resource.init_param(access_point_id)
 
     def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
         self._register_validator(SharedStorageNameValidator, name=self.name)
@@ -400,7 +400,7 @@ class SharedEfs(Resource):
             EfsMountOptionsValidator,
             encryption_in_transit=self.encryption_in_transit,
             iam_authorization=self.iam_authorization,
-            accesspoint_id=self.accesspoint_id,
+            access_point_id=self.access_point_id,
             name=self.name,
         )
 
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -371,6 +371,7 @@ class SharedEfs(Resource):
         deletion_policy: str = None,
         encryption_in_transit: bool = None,
         iam_authorization: bool = None,
+        accesspoint_id: str = None,
     ):
         super().__init__()
         self.mount_dir = Resource.init_param(mount_dir)
@@ -387,6 +388,7 @@ class SharedEfs(Resource):
         )
         self.encryption_in_transit = Resource.init_param(encryption_in_transit, default=False)
         self.iam_authorization = Resource.init_param(iam_authorization, default=False)
+        self.accesspoint_id = Resource.init_param(accesspoint_id)
 
     def _register_validators(self, context: ValidatorContext = None):  # noqa: D102 #pylint: disable=unused-argument
         self._register_validator(SharedStorageNameValidator, name=self.name)
@@ -398,6 +400,7 @@ class SharedEfs(Resource):
             EfsMountOptionsValidator,
             encryption_in_transit=self.encryption_in_transit,
             iam_authorization=self.iam_authorization,
+            accesspoint_id=self.accesspoint_id,
             name=self.name,
         )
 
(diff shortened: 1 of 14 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -321,6 +321,9 @@ class EfsSettingsSchema(BaseSchema):
     deletion_policy = fields.Str(
         validate=validate.OneOf(DELETION_POLICIES), metadata={"update_policy": UpdatePolicy.SUPPORTED}
     )
+    accesspoint_id = fields.Str(
+        validate=validate.Regexp(r"^fsap-[0-9a-z]{17}$"),
```
gmarciani: The property `accesspoint_id` should not support updates, at least for the first iteration on the feature. 
Please add `, metadata={"update_policy": UpdatePolicy.UNSUPPORTED}`.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -321,6 +321,9 @@ class EfsSettingsSchema(BaseSchema):
     deletion_policy = fields.Str(
         validate=validate.OneOf(DELETION_POLICIES), metadata={"update_policy": UpdatePolicy.SUPPORTED}
     )
+    accesspoint_id = fields.Str(
```
gmarciani: The property would be exposed in the cluster config as `AccessPointId`
Please renamed it as `access_point_id`, here and in the other files.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1109,6 +1109,7 @@ def _add_efs_storage(self, id: str, shared_efs: SharedEfs):
             shared_efs.encryption_in_transit
         )
         self.shared_storage_attributes[SharedStorageType.EFS]["IamAuthorizations"].append(shared_efs.iam_authorization)
+        self.shared_storage_attributes[SharedStorageType.EFS]["AccesspointIds"].append(shared_efs.accesspoint_id)
```
gmarciani: Here and in other files, we are assuming that more than one access point can be specified for a given file system (AccesspointsIds rather than AccesspointId).

It's true that a file system can support multiple access points (https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#efs-access-poiont-create), but the mount action is always done using a single access point (https://docs.aws.amazon.com/efs/latest/ug/mounting-access-points.html).

So please, let's assume to specify only one access point on the cluster config.

cli/src/pcluster/validators/efs_validators.py
```diff
@@ -18,7 +18,7 @@ class EfsMountOptionsValidator(Validator):
     IAM Authorization requires Encryption in Transit.
     """
 
-    def _validate(self, encryption_in_transit: bool, iam_authorization: bool, name: str):
+    def _validate(self, encryption_in_transit: bool, iam_authorization: bool, accesspoint_id: str, name: str):
```
gmarciani, line 21: This change should be captured in unit test here: https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/validators/test_efs_validators.py

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -335,6 +335,10 @@ def _add_compute_resource_launch_template(
                         self._shared_storage_attributes[SharedStorageType.EFS]["IamAuthorizations"],
                         use_lower_case=True,
                     ),
+                    "efs_accesspoint_ids": to_comma_separated_string(
```
gmarciani: This change should be captured in unit test here: https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/templates/test_queues_stack.py#L114

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -235,6 +235,10 @@ def _add_login_nodes_pool_launch_template(self):
                         self._shared_storage_attributes[SharedStorageType.EFS]["IamAuthorizations"],
                         use_lower_case=True,
                     ),
+                    "efs_accesspoint_ids": to_comma_separated_string(
```
gmarciani: This change should be captured in unit test here: https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/templates/test_login_nodes_stack.py#L23

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1288,6 +1289,10 @@ def _add_head_node(self):
                     "efs_iam_authorizations": to_comma_separated_string(
                         self.shared_storage_attributes[SharedStorageType.EFS]["IamAuthorizations"], use_lower_case=True
                     ),
+                    "efs_accesspoint_ids": to_comma_separated_string(
```
gmarciani: This change should be captured in unit test here: https://github.com/aws/aws-parallelcluster/blob/develop/cli/tests/pcluster/templates/test_cluster_stack.py#L901

cli/src/pcluster/validators/efs_validators.py
gmarciani, line 21: The validator should verify that the provided access point exists

gmarciani (changes requested):

cli/src/pcluster/config/cluster_config.py
```diff
@@ -371,6 +371,7 @@ def __init__(
         deletion_policy: str = None,
         encryption_in_transit: bool = None,
         iam_authorization: bool = None,
+        accesspoint_id: str = None,
```
gmarciani: this is a new customer facing feature, so a new entry in the CHANGELOG should be added.

gmarciani (changes requested):

gmarciani:
Hi @thheinen,
thanks a lot for your contribution!

Made a first round of review and unblocked PR checks.

cli/src/pcluster/templates/cluster_stack.py
dreambeyondorange: I think this is actually a per filesystem array, similar to the encryption in transit array

thheinen: Yes, it's actually an array of all the separate FS in config. I just aligned naming and structure to the other properties which are also named in plural.

PR #6377, opened 2024-08-02
[LoginNodes] Improve login node security posture

cjmakin:
### Description of changes
* Added `AllowedIps` configuration to login node pool config.
  * A user can now restrict SSH access to a login node pool with a specific CIDR or prefix-list. The default behavior is maintained, where the login node pool inherits the head node SSH restrictions if `AllowedIps` isn't specified for the login node pool.
 
* Added a managed security group to the login node pool load balancer.
  * The load balancer managed security group contains the same SSH restriction that applies to the login node managed security group.
  
* Aligned login node and load balancer security groups.
  * When a user attaches security groups to a login node pool with `SecurityGroups` or `AdditionalSecurityGroups`, the security groups are also applied to the pool’s load balancer. This change was made to address user’s having to manage multiple access points to the login nodes.
  
* Added inbound rule to the login node pool managed security group which references the load balancer managed security group.
  * This addresses a bug where a user restricts SSH access and the load balancer is unable to perform health checks on the login nodes and they are marked unhealthy. If a user defines their own security group (via `SecurityGroups` in the login node pool config), then the user will need to ensure access from the load balancer to the login nodes.
  
* Changed login node managed security groups to support multiple pools. 
  * Configured each pool to have its own managed security group. This is to allow flexibility in security posture of each pool. For example by having pools with dcv enabled or SSH restrictions.
  
* Updated CHANGELOG

#### Manual Testing:

* AllowedIPs:
    * Verified separate login node and head node `AllowedIps`.
    * Verified login nodes maintain head node `AllowedIPs` when not specified.
    * Verified update possible when login nodes are stopped.

* Load balancer security groups
    * Verified managed security group on load balancer.
    * Verified that load balancer security group is added as reference to managed login security group.
    * Verified shared load balancer security group has aligned SSH restriction with login node security group.
    * Verified `SecurityGroups` shared between login node pool and load balancer.
    * Verified `AdditionalSecurityGroups`:
        * Verified managed security group still exists for both login node pool and load balancer.
        * Verified `AdditionalSecurityGroups` shared between login node pool and load balancer.
    * Verified updates to security groups works correctly.

#### Unit tests

* Updated `test_cluster_stack`, `test_shared_storage`, and `test_cluster_config` to support new configuration. 

#### Integration tests

* Created `test_login_node_security_groups` to verify changed made to load balancer and login node security groups.

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -835,7 +835,7 @@ class LoginNodesSsh(_BaseSsh):
 
     def __init__(self, allowed_ips: str = None, **kwargs):
         super().__init__(**kwargs)
-        # If allowed IPs is not specified for the login node pool, will be aligned with the setting defined
+        # If AllowedIPs is not specified for the login node pool, then allowed_ips will default to the same settings
         # in the HeadNode to restrict SSH access from a specific CIDR or prefix-list.
         self.allowed_ips = Resource.init_param(allowed_ips)
 
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -1322,15 +1322,30 @@ class LoginNodesImage(Resource):
 class LoginNodesNetworking(_BaseNetworking, SubnetsMixin):
     """Represent the networking configuration for LoginNodes."""
 
-    def __init__(self, subnet_ids: List[str], proxy: Proxy = None, **kwargs):
+    def __init__(
+        self,
+        subnet_ids: List[str],
+        proxy: Proxy = None,
+        load_balancer_security_groups: List[str] = None,
+        load_balancer_additional_security_groups: List[str] = None,
+        **kwargs,
+    ):
         _BaseNetworking.__init__(self, **kwargs)
         SubnetsMixin.__init__(self, subnet_ids, **kwargs)
+        self.load_balancer_security_groups = Resource.init_param(load_balancer_security_groups)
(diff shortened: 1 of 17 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -1408,14 +1406,6 @@ def __init__(
         super().__init__(**kwargs)
         self.pools = pools
 
-    @property
```
hgreebe, line 1411: why are we removing this?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -574,11 +576,8 @@ def _add_head_eni(self):
 
     def _add_security_groups(self):
         head_node_security_groups, managed_head_security_group = self._head_security_groups()
-        (
-            login_security_groups,
-            managed_login_security_group,
-            custom_login_security_groups,
```
hgreebe, line 580: why are we removing custom_login_security_groups?

cli/src/pcluster/config/cluster_config.py
cjmakin: just need check if dcv is enabled at individual pool level now since each pool has own managed security group

cli/src/pcluster/templates/cluster_stack.py
cjmakin: it's only use was to grant access from the custom security groups to the managed security groups, but not necessary for pools to access each other

hgreebe (approved):

PR #6450, opened 2024-10-03
[integ-tests-framework] Change integration tests OS daily to increase coverage and avoid false failures when AMIs failed to build

hanwen-pcluste:
### Description of changes
Prior to this commit, OSes were specified for each integration test. For example, in [`develop.yaml`](https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/configs/develop.yaml)
```
test-suites:
  ad_integration:
    test_ad_integration.py::test_ad_integration:
      dimensions:
        - regions: [ "ap-southeast-1" ]
          oss: ["alinux2", "rhel8"]
          schedulers: ["slurm"]
        - regions: ["eu-west-1"]
          oss: ["ubuntu2204", "rhel9"]
          schedulers: ["slurm"]
  basic:
      dimensions:
        - regions: ["af-south-1"]
          oss: ["rhel8"]
          schedulers: ["slurm"]
  capacity_reservations:
    test_on_demand_capacity_reservation.py::test_on_demand_capacity_reservation:
      dimensions:
        - regions: ["us-west-2"]
          oss: ["alinux2023"]
```

This approach has the following shortcomings:
1. Functionality is not validated on all OSes
2. If AMIs are missing for some OSes, relevant integration tests will generate meaningless failures.

Therefore, this commit changes the test definition file to:
```
test-suites:
  ad_integration:
    test_ad_integration.py::test_ad_integration:
      dimensions:
        - regions: [ "ap-southeast-1" ]
          oss: [{{ common.OS_X86_1 }}]
          schedulers: ["slurm"]
  basic:
      dimensions:
        - regions: ["af-south-1"]
          oss: [{{ common.OS_X86_2 }}]
          schedulers: ["slurm"]
  capacity_reservations:
    test_on_demand_capacity_reservation.py::test_on_demand_capacity_reservation:
      dimensions:
        - regions: ["us-west-2"]
          oss: [{{ common.OS_X86_3 }}]
```
Where the OS is rotated each day. For example:
This new change is compatible with the old approach of hard-coding OSes. i.e. both approaches can appear in the same test configuration file.

### Tests
* dry run of develop.yaml is successful

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -8,35 +8,35 @@ test-suites:
       dimensions:
         - regions: [ "ap-southeast-1" ]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["alinux2", "rhel8"]
+          oss: [ {{ OS_X86_0 }}, {{ OS_X86_2 }}]
           schedulers: ["slurm"]
         - regions: ["eu-west-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["ubuntu2204", "rhel9"]
+          oss: [{{ OS_X86_4 }}, {{ OS_X86_6 }}]
           schedulers: ["slurm"]
   basic:
     test_essential_features.py::test_essential_features:
       dimensions:
         - regions: ["af-south-1"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["rhel8"]
+          oss: [{{ OS_X86_1 }}]
           schedulers: ["slurm"]
   capacity_reservations:
     test_on_demand_capacity_reservation.py::test_on_demand_capacity_reservation:
       dimensions:
         - regions: ["us-west-2"]
-          oss: ["alinux2023"]
+          oss: [{{ OS_X86_3 }}]
     test_capacity_blocks.py::test_capacity_blocks:
       dimensions:
         - regions: ["eu-west-1"]
-          oss: ["alinux2"]
+          oss: [{{ OS_X86_5 }}]
           schedulers: ["slurm"]
   cli_commands:
     test_cli_commands.py::test_slurm_cli_commands:
       dimensions:
         - regions: ["ap-northeast-2"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["ubuntu2204"]
+          oss: [{{ OS_X86_7 }}]
           schedulers: ["slurm"]
   cloudwatch_logging:
     test_cloudwatch_logging.py::test_cloudwatch_logging:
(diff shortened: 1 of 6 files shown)


tests/integration-tests/configs/develop.yaml
```diff
@@ -8,102 +8,98 @@ test-suites:
...
-          oss: ["alinux2"]
+          oss: [{{ OS_X86_5 }}]
           schedulers: ["slurm"]
   cli_commands:
     test_cli_commands.py::test_slurm_cli_commands:
       dimensions:
         - regions: ["ap-northeast-2"]
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
-          oss: ["ubuntu2204"]
+          oss: [{{ OS_X86_7 }}]
```
gmarciani, line 39: Correct me if I'm wrong: if only 2 x86 AMIs are available, the value OS_x86_7 will be set to the 2nd AMI available in alphabetical order?

tests/integration-tests/test_runner.py
```diff
@@ -656,6 +673,12 @@ def _set_ami_args(args, pytest_args):
     if args.ami_owner:
         pytest_args.extend(["--ami-owner", args.ami_owner])
 
+    if args.available_amis_oss_x86:
+        pytest_args.extend(["--available-amis-oss-x86", " ".join(args.available_amis_oss_x86)])
```
gmarciani, line 677: Why do we need to specify `available-amis-` in this option?
If I use an option like this I assume that the OSS I'm specifying are only those that are available.

tests/integration-tests/test_runner.py
```diff
@@ -61,6 +61,8 @@
     "cookbook_git_ref": None,
     "node_git_ref": None,
     "ami_owner": None,
+    "available_amis_oss_x86": [],
```
gmarciani, line 64: Why an empty list rather than a None?
It seems to me that the more correct values are:
1. SUPPORTED_OSES: in fact you declare that if the option is omitted, the default behaviour will be to use all OSes.
2. None: to represent you're omitting the value

gmarciani (changes requested):

tests/integration-tests/configs/develop.yaml
hanwen-pcluste: You are right

tests/integration-tests/test_runner.py
hanwen-pcluste: There is an existing `-oss` parameter in test_runner, which has a different meaning. I used a more verbose name to be distinguished

tests/integration-tests/test_runner.py
hanwen-pcluste: Default to `[]` is for the simplicity that in all usage, we can assume list type, and do list operations without checking `None` type

tests/integration-tests/test_runner.py
gmarciani: If I understood correctly, the difference between `--oss` and `--available-amis-oss-x86` is that the list of OSs for the second one is going to be rotated.

If this is the case, what about `--rotating-oss-x86` ?

tests/integration-tests/test_runner.py
gmarciani: I see, but, correct me if I'm wrong, here you're setting the default value for an option, and the default value for this option is actually the full list of OSes, so why not simply setting it to `SUPPORTED_OSES `?

gmarciani (approved):

PR #6462, opened 2024-10-11
[Draft] Add integ test for pyxis + enroot

hehe7318:
### Description of changes
* Add integ test for pyxis + enroot

### Tests
* Ongoing, will update when finish.

### Checklist

### DIFF
--- a/tests/integration-tests/tests/pyxis/test_pyxis.py
+++ b/tests/integration-tests/tests/pyxis/test_pyxis.py
@@ -1,0 +1,74 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import logging
+import pytest
+
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+from tests.common.schedulers_common import SlurmCommands
+
+
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+def test_pyxis(pcluster_config_reader, clusters_factory):
+    """
+    Test Enroot and Pyxis failure due to concurrent sed operations on shared filesystem.
+
+    This test creates a cluster with EFS as shared storage and 2000 dynamic compute nodes.
+    It submits a simple job to scale up the cluster, and then submits a Pyxis job which should fail
+    due to the known issue in version 3.11.0.
+
+    The test checks that the expected error messages appear in the job output.
+    """
+    mount_dir = "/fsx_mount_dir"
+    cluster_config = pcluster_config_reader(mount_dir=mount_dir)
+    cluster = clusters_factory(cluster_config)
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = SlurmCommands(remote_command_executor)
+
+    # Submit a simple job to scale up to 2000 nodes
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,64 @@
...
+from assertpy import assert_that
+from remote_command_executor import RemoteCommandExecutor
+
+from tests.common.schedulers_common import SlurmCommands
+
+
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+def test_pyxis(pcluster_config_reader, clusters_factory, test_datadir):
+    """
+    Test Enroot and Pyxis failure due to concurrent sed operations on shared filesystem.
```
gmarciani: The goal of the test is to validate that Pyxis use case works, not to validate that it fail on a specific condition.
That said, we are interested in checking that it now fails on a specific race condition, but this is an internal consideration, not the goal of the test

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,64 @@
...
+
+    job_id = slurm_commands.assert_job_submitted(result.stdout)
+    slurm_commands.wait_job_completed(job_id)
+
+    # Fetch the job output and check for the expected error messages
+    logging.info("Fetching job output")
+    slurm_out = remote_command_executor.run_remote_command("cat slurm.out").stdout
+
+    logging.info("Checking for expected error messages in job output")
+    assert_that(slurm_out).contains("Failed to open /opt/slurm/etc/plugstack.conf.d/")
```
gmarciani, line 61: We should do the other way around.
The test must verify that the job succeeds and in particular, that it does not contain this error in the output.

tests/integration-tests/tests/pyxis/test_pyxis/test_pyxis/pcluster.config.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - Name: compute-resource-0
+          Instances:
+            - InstanceType: t3.small
+          MinCount: 0
+          MaxCount: 1000
+      Networking:
+        SubnetIds:
+          - {{ private_subnet_id }}
+SharedStorage:
+  - MountDir: {{ mount_dir }}
```
gmarciani: The EFs storage should be used as internal shared storage to replicate the race condition.
See https://docs.aws.amazon.com/parallelcluster/latest/ug/HeadNode-v3.html#yaml-HeadNode-SharedStorageType

gmarciani (changes requested):

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: I know, this test is only for current version, when Himani merge the fix PR, I will make changes to adapt it.

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: I know, this test is only for current version, when Himani merge the fix PR, I will make changes to adapt it.

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: For now it's only testing the pyxis+enroot will cause errors in current version.

hehe7318: I will not merge the test with current discription and implementation. I create this draft only for better troubleshooting.

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,63 @@
...
+    assert_that(slurm_out).contains("Permission denied")
+    assert_that(slurm_out).contains("Plug-in initialization failed")
```
gmarciani, line 63: The first assertion is enough to prove the failure on the error message

tests/integration-tests/tests/pyxis/test_pyxis/test_pyxis/pcluster.config.yaml
hehe7318: Fixed

PR #6475, opened 2024-10-17
[Release-3.11] Add pyxis integration test

hehe7318:
### Description of changes
* Add test for Pyxis and Enroot functionality after configuration.
* This test creates a cluster with the necessary custom actions to configure Pyxis and Enroot.
* It submits two consecutive containerized jobs and verifies that they run successfully.

### Tests
* Passed

### Checklist

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -836,4 +836,3 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["ubuntu2204"]
           schedulers: ["slurm"]
-          scale_up_fleet: ["false"]
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -829,3 +829,11 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: ["ubuntu2004"]
           schedulers: ["slurm"]
+  pyxis:
+    test_pyxis.py::test_pyxis:
+      dimensions:
+        - regions: ["eu-west-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["ubuntu2204"]
+          schedulers: ["slurm"]
+          scale_up_fleet: ["false"]
--- a/tests/integration-tests/tests/pyxis/test_pyxis.py
+++ b/tests/integration-tests/tests/pyxis/test_pyxis.py
@@ -19,7 +19,8 @@ from remote_command_executor import RemoteCommandExecutor
 from tests.common.schedulers_common import SlurmCommands
 
 
-@pytest.mark.usefixtures("region", "os", "instance", "scheduler", "scale_up_fleet")
+@pytest.mark.parametrize("scale_up_fleet", [False])
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
 def test_pyxis(pcluster_config_reader, clusters_factory, test_datadir, s3_bucket_factory, region, scale_up_fleet):
     """
     Test Pyxis and Enroot functionality after configuration.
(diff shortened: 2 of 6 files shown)


tests/integration-tests/configs/develop.yaml
```diff
@@ -829,3 +829,11 @@ test-suites:
...
           oss: ["ubuntu2004"]
           schedulers: ["slurm"]
+  pyxis:
+    test_pyxis.py::test_pyxis:
+      dimensions:
+        - regions: ["eu-west-1"]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: ["ubuntu2204"]
+          schedulers: ["slurm"]
+          scale_up_fleet: ["false"]
```
gmarciani: isn't it necessary to declare the scale_up_fleet dimension as a pytest fixture in conftest?

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,94 @@
...
+
+    cluster_config = pcluster_config_reader(bucket_name=bucket_name, max_queue_size=max_queue_size)
+    cluster = clusters_factory(cluster_config)
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = SlurmCommands(remote_command_executor)
+
+    if scale_up_fleet:
+        job_id = slurm_commands.submit_command_and_assert_job_accepted(
+            submit_command_args={"command": "srun hostname", "nodes": 1000}
```
gmarciani: 1000 is the max queue size. let's use that variable rather than using magic numbers

gmarciani: why do we needto submit a standard job only in case of scale up?

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        slurm_commands.wait_job_completed(job_id, timeout=30)
+        slurm_commands.assert_job_succeeded(job_id)
+
+    # Submit the first containerized job
+    logging.info("Submitting first containerized job")
+
+    result = slurm_commands.submit_command(
+        command="srun --container-image docker://ubuntu:22.04 hostname",
+        nodes=3,
```
gmarciani: 3 is the max queue size. let's use that variable rather than using magic numbers

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,94 @@
...
+        other_options="-o slurm1.out",
```
gmarciani: nit: `-o` option is not needed. since you have the job id you can always retrieve the output for that job from `slurm-{job_id}`

tests/integration-tests/tests/pyxis/test_pyxis.py
```diff
@@ -0,0 +1,94 @@
...
+    slurm_out_1 = remote_command_executor.run_remote_command("cat slurm1.out").stdout
+
+    logging.info("Checking for expected messages in first job output")
+    assert_that(slurm_out_1).contains("pyxis: importing docker image: docker://ubuntu:22.04")
+
+    # Submit the second containerized job after the first one completes
+    logging.info("Submitting second containerized job")
+    result = slurm_commands.submit_command(
+        command="srun --container-image docker://ubuntu:22.04 hostname",
+        nodes=3,
```
gmarciani, line 72: use the variable here max_queue_size

gmarciani (changes requested):

tests/integration-tests/tests/pyxis/test_pyxis.py
gmarciani: I suppose it is this way to trigger the race condition.
This is great, however, I suggest to execute the containerized job at scaleif sclaeup is true.
In this way we can test also other race conditions

hehe7318: Done

hehe7318: Done

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: Second job is using a fixed 3 nodes, keep as 3.

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: Done

tests/integration-tests/tests/pyxis/test_pyxis.py
hehe7318: Done

gmarciani (approved):

PR #6488, opened 2024-10-21
[develop] Make export-logs able to download the logs using the default/custom PCluster bucket as staging bucket

hehe7318:
### Description of changes
- The CLI commands `export-cluster-logs` and `export-image-logs` can now by default export the logs to the default ParallelCluster bucket or to the CustomS3Bucket if specified in the config.
- If not specified `--bucket` option, cannot export logs to `parallelcluster/` as it is a protected folder.
- When by default using the pcluster default bucket(`parallelcluster-****************-v1-do-not-delete`), it will automatically update the bucket policy to the latest one to adapt the changes.
- Pcluster default bucket bootstrapped_file format, content and verify method changed.
- Unit tests and integration tests for the new changes.

### Tests
* Unit test fixed and passed. Added new unit tests for new functions. Also covered some old functions.
* Adapted changes for integ tests. Tests worked as expected.
* Manually test done and worked as expected

### Checklist


CHANGELOG.md
```diff
@@ -3,14 +3,12 @@ CHANGELOG
 3.12.0
 ------
 
+**CHANGES**
+- export-cluster-logs/export-image-logs now is able to download the logs using the default/custom PCluster bucket as staging bucket
```
gmarciani: I'd rephrase to "The CLI commands `export-cluster-logs` and `export-image-logs` can now export the logs to the default ParallelCluster bucket and the `CustomS3Bucket` if no other bucket is specified."

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -38,8 +38,10 @@ def register_command_args(self, parser: ArgumentParser) -> None:  # noqa: D102
         # Export options
         parser.add_argument(
             "--bucket",
-            required=True,
-            help="S3 bucket to export cluster logs data to. It must be in the same region of the cluster",
+            help=(
+                "S3 bucket to export cluster logs data to. It must be in the same region of the cluster. "
+                "If not specified, the default/custom ParallelCluster S3 bucket will be used."
```
gmarciani: I'd rephrase to "If not specified, the ParallelCluster default bucket or the CustomS3Bucket (if specified in the config) will be used."

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -43,8 +43,10 @@ def register_command_args(self, parser: ArgumentParser) -> None:  # noqa: D102
         # Export options
         parser.add_argument(
             "--bucket",
-            required=True,
-            help="S3 bucket to export image builder logs data to. It must be in the same region of the image",
+            help=(
+                "S3 bucket to export image builder logs data to. It must be in the same region of the image. "
+                "If not specified, the default ParallelCluster S3 bucket will be used."
```
gmarciani: Why can't we use the CustomS3Bucket bucket as well (if specified), the same way we do for clusters?

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -58,20 +60,35 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
         try:
             if args.output_file:
                 self._validate_output_file_path(args.output_file)
-            return self._export_image_logs(args, args.output_file)
+
+            is_cluster_bucket = False if args.bucket else True
```
gmarciani: The name `is_cluster_bucket` is misleading. What about `is_pcluster_bucket`?
This variable is used to determine if a prefix validation must be done or not; it just be done only if the pcluster bucket is used.

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -65,18 +67,33 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
         try:
             if args.output_file:
                 self._validate_output_file_path(args.output_file)
-            return self._export_cluster_logs(args, args.output_file)
+
+            is_cluster_bucket = False if args.bucket else True
```
gmarciani: The name `is_cluster_bucket` is misleading. What about `is_pcluster_bucket`?
This variable is used to determine if a prefix validation must be done or not; it just be done only if the pcluster bucket is used.

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -65,18 +67,33 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
+            return self._export_cluster_logs(args, args.output_file, is_cluster_bucket)
         except Exception as e:
             utils.error(f"Unable to export cluster's logs.\n{e}")
             return None
 
     @staticmethod
-    def _export_cluster_logs(args: Namespace, output_file: str = None):
+    def _validate_bucket_prefix(bucket_prefix: str) -> None:
+        if bucket_prefix:
+            if bucket_prefix.startswith("parallelcluster/") or bucket_prefix=="parallelcluster":
```
gmarciani: **CodeStyle** Magic strings should be avoided as much as magic numbers.
I suggest to define a constant `PCLUSTER_BUCKET_PROTECTED_PREFIXES` and use it for the checks

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -65,18 +67,33 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
+                raise ValueError(
+                    "Cannot export logs to the protected folder 'parallelcluster/'. Please use another folder."
```
gmarciani: The fact that some folders are protected when the pcluster bucket is used must be reflected in the option helper and the public doc.

Also the error message must be more specific and dynamic:  "Cannot export logs to `{bucket_prefix}` as it is a protected folder in {default bucket}. Please use another folder."

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -65,18 +67,33 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
+                )
+
+    @staticmethod
+    def _export_cluster_logs(args: Namespace, output_file: str = None, is_cluster_bucket: bool = False):
```
gmarciani: Why do we need to add `is_cluster_bucket`?
The line below may be 

`bucket=args.bucket if arg.bucket else cluster.bucket`

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -58,20 +60,35 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
+    def _export_image_logs(args: Namespace, output_file: str = None, is_cluster_bucket: bool = False):
         """Export the logs associated to the image."""
         LOGGER.debug("Beginning export of logs for the image: %s", args.image_id)
 
         # retrieve imagebuilder config and generate model
         imagebuilder = ImageBuilder(image_id=args.image_id)
         url = imagebuilder.export_logs(
-            bucket=args.bucket,
+            # imagebuilder.bucket will handle the bucket init including policy update
+            bucket=imagebuilder.bucket.name if is_cluster_bucket else args.bucket,
```
gmarciani: Why do we need to add `is_cluster_bucket`?
The line below may be 

`bucket=args.bucket if arg.bucket else image_builder.bucket`

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -58,20 +60,35 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
             utils.error(f"Unable to export image's logs.\n{e}")
             return None
 
     @staticmethod
-    def _export_image_logs(args: Namespace, output_file: str = None):
+    def _validate_bucket_prefix(bucket_prefix: str) -> None:
+        if bucket_prefix:
+            if bucket_prefix.startswith("parallelcluster/") or bucket_prefix=="parallelcluster":
+                raise ValueError(
+                    "Cannot export logs to the protected folder 'parallelcluster/'. Please use another folder."
```
gmarciani: Same comment made for the cluster export

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -154,6 +154,95 @@ def configure_s3_bucket(self):
...
+        return policy
+
+    def _get_bucket_policy_for_cloudwatch_logs(self):
+        """Generate the required bucket policy statements for CloudWatch Logs export."""
+        partition = self.partition
+        region = self.region
+        account_id = self.account_id
+        bucket_arn = f"arn:{partition}:s3:::{self.name}"
+        bucket_arn_with_wildcard = f"{bucket_arn}/*"
+        logs_service_principal = f"logs.{region}.{get_url_domain_suffix(partition)}"
```
gmarciani: This is a good candidate to be a common function because there are other places in our code where we create service principal strings. The function may be: `def get_service_principal(service_name: str, region: str, regional: bool = False)`.

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -154,6 +154,95 @@ def configure_s3_bucket(self):
...
+                "Sid": "AllowPutObjectForExportLogs",
+                "Action": "s3:PutObject",
+                "Effect": "Allow",
+                "Resource": bucket_arn_with_wildcard,
+                "Principal": {"Service": logs_service_principal},
+                "Condition": {
+                    "StringEquals": {"s3:x-amz-acl": "bucket-owner-full-control", "aws:SourceAccount": account_id},
+                    "ArnLike": {
+                        "aws:SourceArn": [
+                            f"arn:{partition}:logs:{region}:{account_id}:log-group:/aws/parallelcluster/*",
```
gmarciani: this is a good candidate for a common function. See `format_arn()` https://github.com/aws/aws-parallelcluster/blob/62fdb5eefbb866139180c7795ce539913506b218/cli/src/pcluster/templates/awsbatch_builder.py#L98-L99

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -154,6 +154,95 @@ def configure_s3_bucket(self):
...
+
+    def ensure_bucket_policy_for_logs(self):
+        """Ensure the bucket policy includes necessary statements for exporting logs."""
+        existing_policy = self._get_bucket_policy()
+
+        # Generate the required policy statements
+        required_statements = self._get_bucket_policy_for_cloudwatch_logs()
+
+        # Check which statements are missing
+        existing_sids = {stmt.get("Sid") for stmt in existing_policy.get("Statement", []) if "Sid" in stmt}
```
gmarciani: With this logic, SIDs cannot be changed in the bucket policy and also it's a a weak check: we are interested in statement content, not in statement ids. 
anyway, this check may not be necessary, I suggest to overwrite the bucket policy without checking the SIDs because we are using the pcluster bucket that we are fully in control of.

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -354,6 +443,13 @@ def _check_default_bucket(cls, service_name: str, artifact_directory: str, stack
             bucket.check_bucket_is_bootstrapped()
         except AWSClientError as e:
             cls._configure_bucket(bucket, e)
+        else:
+            # Bucket exists and is bootstrapped, ensure bucket policy is up-to-date
+            try:
+                bucket.ensure_bucket_policy_for_logs()
```
gmarciani: Improvement, that can be done in a follow up PR.

We should configure the bucket policy within `_configure_bucket()` where we are already setting the DenyHTTP statement https://github.com/aws/aws-parallelcluster/blob/8ac2c636f7941bc8983ee75f6abf202028b2d031/cli/src/pcluster/models/s3_bucket.py#L144-L156

At that point we must change the logic to check if the bucket is already bootstrapped:
1. change the bootstrap file body [here](https://github.com/aws/aws-parallelcluster/blob/8ac2c636f7941bc8983ee75f6abf202028b2d031/cli/src/pcluster/models/s3_bucket.py#L191) to reflect what are the bootstrapped features required, e.g. basic (what we already have up to Pc 3.11.1) and export-logs (introduced in PC 3.12.0). 
2. if the bootstrap file does not exist or it does not contain all the required features, then overwrite the bucket policy

gmarciani (changes requested):

CHANGELOG.md
hehe7318: Done!

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Done!

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: We can, but when doing manually tests I made some mistakes and it mislead me. Fixed

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Good catch! Thank you. Fixed

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Fixed

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Done

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Done

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Done

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Done. Also add msg in `--bucket-prefix` help panel.

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Done

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done.

cli/src/pcluster/models/s3_bucket.py
hehe7318: With the changes removing SIDs check, we will need this logic in this PR. Need some more time to implement

hehe7318: Done.

CHANGELOG.md
```diff
@@ -3,14 +3,12 @@ CHANGELOG
 3.12.0
 ------
 
+**CHANGES**
+- The CLI commands `export-cluster-logs` and `export-image-logs` can now export the logs to the default ParallelCluster bucket and the `CustomS3Bucket` if no other bucket is specified.
```
gmarciani: nit: Re-reading I think the line is confusing. 
We could replicate here the same sentence used in the helper: "to the default ParallelCluster bucket or to the `CustomS3Bucket` if specified in the config".

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -43,26 +43,48 @@ def register_command_args(self, parser: ArgumentParser) -> None:  # noqa: D102
...
+    @staticmethod
+    def _validate_bucket_prefix(bucket_prefix: str) -> None:
+        if bucket_prefix:
+            if (
+                bucket_prefix.startswith(PCLUSTER_BUCKET_PROTECTED_PREFIX)
+                or bucket_prefix == PCLUSTER_BUCKET_PROTECTED_FOLDER
+            ):
+                raise ValueError(
+                    f"Cannot export logs to {bucket_prefix} as it is a protected folder "
+                    f"in {PCLUSTER_BUCKET_PROTECTED_PREFIX}. Please use another folder."
```
gmarciani: as it is *within* a protected folder

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -43,26 +43,48 @@ def register_command_args(self, parser: ArgumentParser) -> None:  # noqa: D102
...
     def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102 #pylint: disable=unused-argument
         assert_supported_operation(operation=Operation.EXPORT_IMAGE_LOGS, region=args.region or get_region())
         try:
             if args.output_file:
                 self._validate_output_file_path(args.output_file)
+
+            is_pcluster_bucket = not args.bucket
+
+            if is_pcluster_bucket:
+                self._validate_bucket_prefix(args.bucket_prefix)
```
gmarciani, line 72: the prefix validation should be done both on the pcluster default bucket and on the CustomS3Bucket. The variable name should better reflect it, or at least add a comment here explaining when the validation is performed. Otherwise, it seems that the validation is done only for the default pcluster bucket.

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -184,19 +270,41 @@ def delete_s3_artifacts(self):
...
+                bucket_name=self.name,
+                key="/".join([self._root_directory, self._bootstrapped_file_name]),
+            )
+            file_content = response["Body"].read().decode("utf-8")
+            try:
+                bootstrapped_info = json.loads(file_content)
+                bootstrapped_features = bootstrapped_info.get("bootstrapped_features", [])
+            except json.JSONDecodeError:
+                LOGGER.info("Bootstrap file is in old format. Assuming basic features are bootstrapped.")
+                bootstrapped_features = ["basic"]
```
gmarciani: Let's define bootstrap features as constants

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -184,19 +270,41 @@ def delete_s3_artifacts(self):
...
+
+            for feature in self.REQUIRED_BOOTSTRAP_FEATURES:
+                if feature not in bootstrapped_features:
+                    return False
+            return True
+        except AWSClientError as e:
+            if e.error_code == "404":
+                LOGGER.info("Bootstrap file doesn't exist. S3 Bucket %s is not configured.", self.name)
```
gmarciani: This line should be a warning

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -184,19 +270,41 @@ def delete_s3_artifacts(self):
...
+            response = AWSApi.instance().s3.get_object(
+                bucket_name=self.name,
+                key="/".join([self._root_directory, self._bootstrapped_file_name]),
+            )
+            file_content = response["Body"].read().decode("utf-8")
+            try:
+                bootstrapped_info = json.loads(file_content)
+                bootstrapped_features = bootstrapped_info.get("bootstrapped_features", [])
+            except json.JSONDecodeError:
+                LOGGER.info("Bootstrap file is in old format. Assuming basic features are bootstrapped.")
```
gmarciani, line 297: Let's add a comment here explaining what's the old format like

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -184,19 +270,41 @@ def delete_s3_artifacts(self):
...
+            for feature in self.REQUIRED_BOOTSTRAP_FEATURES:
+                if feature not in bootstrapped_features:
+                    return False
+            return True
+        except AWSClientError as e:
+            if e.error_code == "404":
+                LOGGER.info("Bootstrap file doesn't exist. S3 Bucket %s is not configured.", self.name)
+                return False
+            else:
+                LOGGER.error("Unable to check S3 bucket %s is configured properly.", self.name)
```
gmarciani: typo: "...to check if..."

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -370,17 +479,13 @@ def _create_bucket(cls, bucket: S3Bucket, e: AWSClientError):
...
-            except AWSClientError as error:
-                LOGGER.error("Configure S3 bucket %s failed.", bucket.name)
-                raise error
-        else:
-            LOGGER.error("Unable to check S3 bucket %s is configured properly.", bucket.name)
-            raise e
+    def _configure_bucket(cls, bucket: S3Bucket):
+        try:
+            bucket.configure_s3_bucket()
+            bucket.upload_bootstrapped_file(bootstrapped_features=bucket.REQUIRED_BOOTSTRAP_FEATURES)
```
gmarciani: Why should the function `upload_bootstrapped_file` accept argument `bootstrapped_features`?
Is there any scenario where the list of features should be different from `REQUIRED_BOOTSTRAP_FEATURES `?
If the answer is no, then let's remove the argument.

cli/src/pcluster/constants.py
```diff
@@ -299,3 +299,6 @@ class Operation(Enum):
 IAM_ROLE_REGEX = "^arn:.*:role/"
 IAM_INSTANCE_PROFILE_REGEX = "^arn:.*:instance-profile/"
 IAM_POLICY_REGEX = "^arn:.*:policy/"
+
+PCLUSTER_BUCKET_PROTECTED_PREFIX = "parallelcluster/"
```
gmarciani: Let's interpolated PCLUSTER_BUCKET_PROTECTED_FOLDER rather than repeating `parallelcluster`

gmarciani (changes requested):

CHANGELOG.md
hehe7318: Done! Thanks!

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Done, also done for cluster_logs.py

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Done, added a comment to explain the scenario.

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done. BTW, it's a normal scenario when the cluster is created as first time.

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done. Added # The bootstrap file old format is a simple String "bucket is configured successfully."

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done, added `if`.

cli/src/pcluster/models/s3_bucket.py
hehe7318: Good catch. Thank you.

hehe7318: Done

cli/src/pcluster/constants.py
hehe7318: Done

cli/src/pcluster/cli/commands/image_logs.py
gmarciani: I mean `Cannot export logs to {bucket_prefix} as it is within the protected folder {PCLUSTER_BUCKET_PROTECTED_PREFIX}. Please use another folder.`

cli/tests/pcluster/cli/test_export_cluster_logs/TestExportClusterLogsCommand/test_helper/pcluster-help.txt
```diff
@@ -18,11 +18,16 @@ options:
...
-                        in the same region of the cluster
+                        in the same region of the cluster. If not specified,
+                        the ParallelCluster default bucket or the
+                        CustomS3Bucket (if specified in the config) will be
+                        used.
   --bucket-prefix BUCKET_PREFIX
                         Keypath under which exported logs data will be stored
                         in s3 bucket. Defaults to <cluster_name>-logs-<current
-                        time in the format of yyyyMMddHHmm>
+                        time in the format of yyyyMMddHHmm>, If not specify
```
gmarciani: Typo: "if not specify" => "if not specified"

cli/tests/pcluster/cli/test_export_cluster_logs/TestExportClusterLogsCommand/test_helper/pcluster-help.txt
```diff
@@ -18,11 +18,16 @@ options:
...
+                        `--bucket` option, cannot export logs to
+                        parallelcluster/ as it is a protected folder.
```
gmarciani, line 30: Let's use the constant here instead of `parallelcluster/`

cli/tests/pcluster/cli/test_export_image_logs/TestExportImageLogsCommand/test_helper/pcluster-help.txt
```diff
@@ -33,8 +34,13 @@ options:
...
-                        must be in the same region of the image
+                        must be in the same region of the image. If not
+                        specified, the ParallelCluster default bucket or the
+                        CustomS3Bucket (if specified in the config) will be
+                        used.
   --bucket-prefix BUCKET_PREFIX
                         Keypath under which exported logs data will be stored
                         in s3 bucket. Defaults to <image_id>-logs-<current
-                        time in the format of yyyyMMddHHmm>
+                        time in the format of yyyyMMddHHmm>. If not specify
```
gmarciani: Same as above for the typo and the use of constant

cli/tests/pcluster/models/dummy_s3_bucket.py
```diff
@@ -87,6 +87,7 @@ def mock_bucket_object_utils(
     delete_s3_artifacts_side_effect=None,
     upload_bootstrapped_file_side_effect=None,
     check_bucket_is_bootstrapped_side_effect=None,
+    check_bucket_is_bootstrapped_return_value=None,
```
gmarciani: why the check should return None, rather than True/False?

cli/tests/pcluster/cli/test_commands.py
```diff
@@ -333,3 +343,101 @@ def test_setup_bucket_with_resources_upload_failure(
...
+@pytest.mark.parametrize(
+    (
+        "head_object_side_effect",
+        "get_object_side_effect",
+        "get_object_return_value",
+        "expected_result",
+        "expected_exception",
+    ),
+    [
+        # Test case 1: The bootstrap file exists, content is in the new format, and contains all required features
```
gmarciani: Rather than describing the test with a comment, use the test id.
See example here: https://github.com/aws/aws-parallelcluster/blob/c84baa2955744e418eccc42c93c339131ef36a06/cli/tests/pcluster/aws/test_ec2.py#L168

cli/src/pcluster/utils.py
```diff
@@ -579,3 +579,19 @@ def retrieve_supported_regions():
             with open(pkg_resources.resource_filename(__name__, "/resources/supported-regions"), encoding="utf-8") as f:
                 retrieve_supported_regions.cache = f.read().split("\n")
     return retrieve_supported_regions.cache
+
+
+def get_service_principal(service_name: str, partition: str, region: str = None, regional: bool = False) -> str:
```
gmarciani, line 584: This new function should be covered by a unit test

cli/src/pcluster/utils.py
```diff
@@ -579,3 +579,19 @@ def retrieve_supported_regions():
...
+    domain_suffix = get_url_domain_suffix(partition)
+    if regional:
+        if not region:
+            raise ValueError("Region must be provided when 'regional' is True.")
+        return f"{service_name}.{region}.{domain_suffix}"
+    else:
+        return f"{service_name}.{domain_suffix}"
+
+
+def format_arn(partition: str, service: str, region: str, account: str, resource: str) -> str:
```
gmarciani, line 595: This new function should be covered by a unit test

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -148,12 +148,96 @@ def configure_s3_bucket(self):
...
+
+        # Create the combined bucket policy
+        bucket_policy = {
+            "Version": "2012-10-17",
+            "Statement": all_statements,
+        }
+
+        return bucket_policy
+
+    def _get_bucket_policy_for_cloudwatch_logs(self):
```
gmarciani, line 185: We should cover the generation of the policy with a unit test

cli/src/pcluster/models/s3_bucket.py
```diff
@@ -148,12 +148,96 @@ def configure_s3_bucket(self):
...
-            '{{"Id":"DenyHTTP","Version":"2012-10-17","Statement":[{{"Sid":"AllowSSLRequestsOnly","Action":"s3:*",'
-            '"Effect":"Deny","Resource":["arn:{partition}:s3:::{bucket_name}","arn:{partition}:s3:::{bucket_name}/*"],'
-            '"Condition":{{"Bool":{{"aws:SecureTransport":"false"}}}},"Principal":"*"}}]}}'
-        ).format(bucket_name=self.name, partition=self.partition)
-        AWSApi.instance().s3.put_bucket_policy(bucket_name=self.name, policy=deny_http_policy)
+
+        bucket_policy = self._generate_bucket_policy()
+        AWSApi.instance().s3.put_bucket_policy(bucket_name=self.name, policy=json.dumps(bucket_policy))
+
+    def _generate_bucket_policy(self):
```
gmarciani, line 155: We should cover the generation of this bucket policy with a unit test. I know this policy exist before your PR, but it's worth it to take the chance to cover it.

cli/src/pcluster/cli/commands/image_logs.py
```diff
@@ -43,26 +43,51 @@ def register_command_args(self, parser: ArgumentParser) -> None:  # noqa: D102
...
+                "If not specified, the ParallelCluster default bucket or "
+                "the CustomS3Bucket (if specified in the config) will be used."
+            ),
         )
         # Export options
         parser.add_argument(
             "--bucket-prefix",
             help="Keypath under which exported logs data will be stored in s3 bucket. Defaults to "
-            "<image_id>-logs-<current time in the format of yyyyMMddHHmm>",
+            "<image_id>-logs-<current time in the format of yyyyMMddHHmm>. If not specify `--bucket` option, "
```
gmarciani: Same as above: typo

gmarciani (changes requested):
We should implement a logic that prevents the deletion of objects from the protected folders

cli/tests/pcluster/cli/test_export_cluster_logs/TestExportClusterLogsCommand/test_helper/pcluster-help.txt
hehe7318: Done

cli/tests/pcluster/cli/test_export_cluster_logs/TestExportClusterLogsCommand/test_helper/pcluster-help.txt
hehe7318: This is a txt file for the unit test.

cli/tests/pcluster/cli/test_export_image_logs/TestExportImageLogsCommand/test_helper/pcluster-help.txt
hehe7318: Done for the typo. But this is a txt file for the unit test.

cli/tests/pcluster/models/dummy_s3_bucket.py
hehe7318: Fixed. Changed to True. Thanks for catching!

cli/tests/pcluster/cli/test_commands.py
hehe7318: Done, thanks for knowledge sharing.

cli/src/pcluster/cli/commands/image_logs.py
hehe7318: Done

cli/src/pcluster/utils.py
hehe7318: Done

cli/src/pcluster/utils.py
hehe7318: Done

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done, but seems like it's unnecessary, due to it doesn't have logic that worth to test. The generated policy statements depend on the return value of `get_service_principal`, `format_arn`, etc.

cli/src/pcluster/models/s3_bucket.py
hehe7318: Done

hehe7318:
> We should implement a logic that prevents the deletion of objects from the protected folders

I can't see why. We don't have a objects deletion logic in our codebase. Also we didn't add logic that put objects in protected folder. 

I suppose even if we should implement this logic it should not be included in this PR.

cli/src/pcluster/cli/commands/cluster_logs.py
```diff
@@ -65,18 +70,40 @@ def execute(self, args: Namespace, extra_args: List[str]) -> None:  # noqa: D102
...
+                # Skip validation if a bucket is specified via the --bucket CLI argument.
+                self._validate_bucket_prefix(args.bucket_prefix)
+
             return self._export_cluster_logs(args, args.output_file)
         except Exception as e:
             utils.error(f"Unable to export cluster's logs.\n{e}")
             return None
 
+    @staticmethod
+    def _validate_bucket_prefix(bucket_prefix: str) -> None:
```
gmarciani: This function is equal to the one used for the image builder part. Can we remove code duplication?

cli/src/pcluster/models/s3_bucket.py
gmarciani: A unit test is necessary even if the function returns a static string because it protects from regressions.

Example: what if I change the bucket policy version? That string is static, but still I want my unit test to capture the mistake.

Makes sense?

cli/tests/pcluster/cli/test_commands.py
```diff
@@ -333,3 +343,117 @@ def test_setup_bucket_with_resources_upload_failure(
...
+                None,
+                {"Body": Mock(read=Mock(return_value="{invalid json}".encode("utf-8")))},
+                False,
+                None,
+                id="The bootstrap file content cannot be parsed as JSON",
+            )
+        ),
+    ],
+)
+def test_check_bucket_is_bootstrapped(
```
gmarciani: This unit test should be moved to `cli/tests/pcluster/models/test_s3_bucket.py` because the original function 
`check_bucket_is_bootstrapped()` is in file `cli/src/models/s3_bucket`.

cli/tests/pcluster/test_utils.py
```diff
@@ -607,3 +607,114 @@ async def async_method(self, param):
...
+            utils.get_service_principal(service_name, partition, region, regional)
+    else:
+        result = utils.get_service_principal(service_name, partition, region, regional)
+        assert_that(result).is_equal_to(expected_result)
+
+
+@pytest.mark.parametrize(
+    "partition, service, region, account, resource, expected_result",
+    [
+        pytest.param(
```
gmarciani, line 659: *Testing Best Practice*
The function under test is joining input values following a given schema, but it does not make any assumption on the input values (and this is fine).

In this case, the test parameters should be generic:

Example: 

```
 pytest.param(
            "aws",
            "s3",
            "",
            "123456789012",
            "bucket_name",
            "arn:aws:s3::123456789012:bucket_name",
            id="Empty region",
        ),
``` 

should be:

```
 pytest.param(
            "ANY_PARTITION",
            "ANY_SERVICE",
            "",
            "ANY_ACCOUNT_ID",
            "ANY_BUCKET_NAME",
            "arn:ANY_PARTITION:ANY_SERVICE::ANY_ACCOUNT_ID:ANY_BUCKET_NAME",
            id="Empty region",
        ),
```

gmarciani (changes requested):

cli/src/pcluster/models/s3_bucket.py
hehe7318: Makes sense

cli/src/pcluster/cli/commands/cluster_logs.py
hehe7318: Done

cli/tests/pcluster/cli/test_commands.py
hehe7318: Done, thanks for catching this.

cli/tests/pcluster/test_utils.py
hehe7318: Done.

cli/tests/pcluster/test_utils.py
```diff
@@ -607,3 +607,114 @@ async def async_method(self, param):
...
+            "logs.cn-north-1.amazonaws.com.cn",
+            None,
+            id="Regional service principal in aws-cn partition",
+        ),
+    ],
+)
+def test_get_service_principal(mocker, service_name, partition, region, regional, expected_result, expected_exception):
+    """Test get_service_principal with various inputs."""
+    # Mock get_url_domain_suffix
+    domain_suffix = "amazonaws.com.cn" if partition == "aws-cn" else "amazonaws.com"
```
gmarciani: *Test Best Practice*
This if..else logic is not necessary.
Here you're unit testing the function `get_service_principal`, which does not make any assumption on the value returned by `get_url_domain_suffix`. So the mock should return a static value `AWS_DOMAIN_SUFFIX` and the test should assert that the function `get_url_domain_suffix`is called once

hehe7318: You are right, I should not mock `get_url_domain_suffix`. I re-checked the code and I decided to remove the mock logic and let the function itself handle the `domain_suffix`. Does it make sense to you?

cli/tests/pcluster/models/test_s3_bucket.py
```diff
@@ -301,3 +304,241 @@ def test_upload_file(mocker, content, file_name, file_type, s3_file_format, expe
...
+    )
+
+    log_group_arns = [
+        format_arn(partition, "logs", region, account_id, "log-group:/aws/parallelcluster/*"),
+        format_arn(partition, "logs", region, account_id, "log-group:/aws/imagebuilder/*"),
+    ]
+
+    policy_statements = bucket._get_bucket_policy_for_cloudwatch_logs()
+
+    assert_that(policy_statements).is_length(3)
```
gmarciani, line 338: You could have simplified this long set of assertions with a single assert that the policy is equal to the expected json, without asserting field by field.

cli/tests/pcluster/test_utils.py
gmarciani: It's ok, let's not block the Pr for this.
However, the best practice would be to let the get_url_domain_suffix return a static value `A_PARTITION` and the expected result is adapted accordigly.

cli/tests/pcluster/models/test_s3_bucket.py
```diff
@@ -301,3 +304,241 @@ def test_upload_file(mocker, content, file_name, file_type, s3_file_format, expe
...
+    # Verify DenyPutObjectOnReservedPath policy
+    deny_put_object_statement = statements_by_sid.get("DenyPutObjectOnReservedPath")
+    assert_that(deny_put_object_statement).is_not_none()
+    assert_that(deny_put_object_statement["Action"]).is_equal_to("s3:PutObject")
+    assert_that(deny_put_object_statement["Effect"]).is_equal_to("Deny")
+    assert_that(deny_put_object_statement["Resource"]).is_equal_to(f"{bucket_arn}/parallelcluster/*")
+    assert_that(deny_put_object_statement["Principal"]).is_equal_to({"Service": logs_service_principal})
+
+
+def test_generate_bucket_policy(mocker):
```
gmarciani, line 381: Same as above: reduce the assertion to an assert equal to a static json instead of asserting field by field.

hehe7318:
> > We should implement a logic that prevents the deletion of objects from the protected folders
> 
> I can't see why. We don't have a objects deletion logic in our codebase. Also we didn't add logic that put objects in protected folder.
> 
> I suppose even if we should implement this logic it should not be included in this PR.

Had a sync with Giacomo, we have validator and DenyPutS3Object(only for logs service) policy for the default pcluster bucket. So although we have objects deletion logic in our codebase(--keep-s3-objects false), it doesn't matter.

cli/tests/pcluster/models/test_s3_bucket.py
hehe7318: Done!

cli/tests/pcluster/models/test_s3_bucket.py
hehe7318: Done

cli/tests/pcluster/test_utils.py
hehe7318: Done

gmarciani (approved):

PR #6526, opened 2024-11-01
Move reboot step to be before installing kernel package during OS update

hgreebe:
### Description of changes
* Cherry-pick: https://github.com/aws/aws-parallelcluster/pull/6486

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
@@ -244,7 +244,12 @@ phases:
                 fi
                 apt-get --purge autoremove -y
               fi
-
+      - name: RebootStep
+        action: Reboot
+        onFailure: Abort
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -322,13 +327,6 @@ phases:
                  echo "SSM agent is installed by default"
               fi
 
-      - name: RebootStep
```
himani2411: There are other packages we install after the previous reboot step, so I dont think we should remove the last reboot step.

himani2411 (approved):

cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
hanwen-pcluste: The two steps should have different names

PR #6543, opened 2024-11-06
[Test] Simplify AD integ test

gmarciani:
### Description of changes
Simplified AD integ test by
1. Removing the use of EFS, FSxLustre, FSxOntap and FSxOpenZfs and keeping only EBS. In fact it is overkilling to validate AD with all the external shared storage. It's time consuming, costly and may introduce noise caused by failures on shared storage that are meant ot be detected by other tests.
2. Remove from cluster config the custom log retention.
3. Remove script to publish unused compute node metrics that were introduced during initial feature validation years ago and only used once.

### Tests
[ONGOING] Test `tets_ad_integration` succeeded

### References

### Checklist


hanwen-pcluste:
How about only removing FSxOntap, which is taking much time. Did we see sporadic failures because of shared storage in the test?

gmarciani:
> How about only removing FSxOntap, which is taking much time. Did we see sporadic failures because of shared storage in the test?

I think we should remove every EFS and FSx because:
1. the goal of this test is to validate the integration between PC and AD, not that EFS/FSx respect the user permissions (this is a Linux related thing that EFS/FSx is supposed to care about).
1. we observed test failures due to transient networking issues with FSx => we don't want such out of scope failures to impact the Ad test.

That said, I suggest to keep EBS simply because it's a time/money cheap data point on shared storage that typically has lower chances to fail the mounting.

hanwen-pcluste (approved):

PR #6547, opened 2024-11-06
[Integ-Test] Fix test_dcv intermittent failure by filter running instance

hehe7318:
### Description of changes
* Failure root cause: For this failed test, in aws console, LoginNodes AutoScalingGroup, we can see:
   ```
   Launching a new EC2 instance: i-09c9a5b6c8080d8e3. Status Reason: Instance failed to complete user's Lifecycle Action: Lifecycle Action with token 18dcffb8-3333-4510-b29a-65b97b862804 was abandoned: Heartbeat Timeout
   ```
* And so that a new LN instance replaced it. The `_get_initial_cluster_log_state` function didn't have a filter of 'running' instances. So that the first LN was added and use its not exist anymore private ip address to run the command: `/bin/bash --login -c 'ssh -o StrictHostKeyChecking=no -q ip-<wrong ip>.ec2.internal hostname -f'`. So that the test failed.
* Fix test_dcv intermittent failure by filter running instance

### Tests
* test_dcv passed

### Checklist


tests/integration-tests/tests/cloudwatch_logging/test_cloudwatch_logging.py
```diff
@@ -220,6 +220,9 @@ def _add_login_instance(self, instance):
     def _get_initial_cluster_log_state(self):
         """Get EC2 instances belonging to this cluster. Figure out their roles in the cluster."""
         for instance in cw_logs_utils.get_ec2_instances():
+            # Ensure the instance is in 'running' state
```
himani2411: If we loop using this condition, we can miss those instances which are not running but are part of the cluster.

For ex. a login node instance is not yet in running state, but its part of this cluster we create. Then we never track that instance as it will just go to the next iteration of the loop.

hehe7318: I have a doubt. If a cluster instance is not in running instance. Will this command work on HeadNode? 
`ssh -o StrictHostKeyChecking=no -q {} hostname -f".format(instance.get("PrivateDnsName")`
```
    def _add_compute_instance(self, instance):
        """Update the cluster's log state by adding a compute node."""
        compute_hostname = self._run_command_on_head_node(
            "ssh -o StrictHostKeyChecking=no -q {} hostname -f".format(instance.get("PrivateDnsName"))
        )
        self._cluster_log_state[COMPUTE_NODE_ROLE_NAME][compute_hostname] = {
            "node_role": COMPUTE_NODE_ROLE_NAME,
            "hostname": instance.get("PrivateDnsName"),
            "instance_id": instance.get("InstanceId"),
        }

    def _add_login_instance(self, instance):
        """Update the cluster's log state by adding a login node."""
        login_hostname = self._run_command_on_head_node(
            "ssh -o StrictHostKeyChecking=no -q {} hostname -f".format(instance.get("PrivateDnsName"))
        )
        self._cluster_log_state[LOGIN_NODE_ROLE_NAME][login_hostname] = {
            "node_role": LOGIN_NODE_ROLE_NAME,
            "hostname": instance.get("PrivateDnsName"),
            "instance_id": instance.get("InstanceId"),
        }
```
Basically the test_dcv failed because it try to run this command for a already terminated LoginNode instance.

himani2411: The Cluster would not be in CREATE_COMPLETE state if the HeadNode is not running.
Are we not checking if the cluster creation is completed before we log in. Also The HN is the one which is the 1st instance to come up during a cluster creation so we can log into the HN even before the cluster is fully created.

himani2411: Also this change will also affect test_cloudwatch_logging's tests

himani2411: Ohh I think I misinterpreted your question.

himani2411: Do we know why these instances are not in running state? I guess the solution makes sense that we do not want to log instances that are not in running state, but what I am concerned is that we miss the instance which is in pending or initializing state. Can you check if it affects the other test i mentioned?

himani2411: Plus I am not sure why we event run that command, as the hostname is the same as PrivateDnsName. If we dont get the PrivateDnsName whats the point of running that command.

hehe7318: > The Cluster would not be in CREATE_COMPLETE state if the HeadNode is not running.

In some cases, LoginNode may not be in running state. I explained in description.


> Do we know why these instances are not in running state?

Yes, I explained in description.

> Plus I am not sure why we event run that command, as the hostname is the same as PrivateDnsName. If we dont get the PrivateDnsName whats the point of running that command.

I am not sure either. I will check and reply later. Thank you!

> Can you check if it affects the other test i mentioned?

I checked, only affects test_dcv and test_cloudwatch_logging.

hehe7318: > Plus I am not sure why we event run that command, as the hostname is the same as PrivateDnsName. 

Seems there are some cases that the FQDN obtained by running `hostname -f` on the instance may differ from the `PrivateDnsName` provided by AWS EC2 instance metadata. I am not sure how to test it. Are you sure they will be totally the same in parallelcluster?

But anyway, if these two tests are stable previously, that means no instance not in `running` state was included. So it will be safe to add these lines. Or at least, I can add them under `if tags.get("Name", "") == "LoginNode":`. What do you think?

himani2411: Lets add the condition of LoginNode or move the running check condition to `self._add_login_instance(instance)`

hehe7318: Done!

himani2411 (approved):

PR #6623, opened 2024-12-13
Add validators to check head node instance type and shared storage type w.r.t cluster size

hanwen-cluster:
The requirements set in these validators are minimum. Users should leave more safety margin considering their workloads.

### Tests
* Manual testing with cluster configurations that trigger/not trigger the validators is successful

### DIFF
--- a/cli/src/pcluster/validators/cluster_validators.py
+++ b/cli/src/pcluster/validators/cluster_validators.py
@@ -1311,6 +1311,42 @@ class HeadNodeImdsValidator(Validator):
             )
 
 
+class HeadNodeMemorySizeValidator(Validator):
+    """
+    Head Node Memory Size Validator.
+
+    Verify if the Head Node has enough memory to manage compute nodes.
+    """
+
+    def _validate(self, head_node_instance_type: str, total_max_compute_nodes: int):
+        head_node_memory = (
+            AWSApi.instance().ec2.get_instance_type_info(head_node_instance_type).ec2memory_size_in_mib() / 1024
+        )
+        # Assume OS takes up 0.6GB memory. Only check upto 16GB memory to prevent usage of small instance types.
+        required_memory = min(total_max_compute_nodes / 25 + 0.6, 16)
+        if head_node_memory < required_memory:
+            self._add_failure(
+                f"Head node instance type {head_node_instance_type} has {head_node_memory} GB of memory. "
+                f"Please choose a head node instance type with at least {required_memory} GB of memory"
+                f" to manage {total_max_compute_nodes} compute nodes.",
+                FailureLevel.ERROR,
+            )
+
+
+class SharedEbsPerformanceBottleNeckValidator(Validator):
+    """Warn potential performance bottleneck of using Shared EBS."""
+
+    def _validate(self, total_max_compute_nodes: int):
+        if total_max_compute_nodes > 100:
+            self._add_failure(
+                "EBS shared storage is mounted on the head node and shared to the compute nodes. "
+                "Therefore, the head node network bandwidth is a network performance bottle neck "
+                "if the compute nodes rely on this shared storage. "
+                "Please use FSx and EFS for better performance.",
+                FailureLevel.WARNING,
+            )
+
+
 class ComputeResourceLaunchTemplateValidator(_LaunchTemplateValidator):
     """Try to launch the requested instances (in dry-run mode) to verify configuration parameters."""
 
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -88,6 +88,7 @@ from pcluster.validators.cluster_validators import (
     FsxArchitectureOsValidator,
     HeadNodeImdsValidator,
     HeadNodeLaunchTemplateValidator,
+    HeadNodeMemorySizeValidator,
     HostedZoneValidator,
     InstanceArchitectureCompatibilityValidator,
     IntelHpcArchitectureValidator,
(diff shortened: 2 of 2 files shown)


cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,41 @@ def _validate(self, imds_secured: bool, scheduler: str):
...
+            AWSApi.instance().ec2.get_instance_type_info(head_node_instance_type).ec2memory_size_in_mib() / 1024
+        )
+        # Assume OS takes up 0.6GB memory. Only check upto 16GB memory to prevent usage of small instance types.
+        required_memory = min(total_max_compute_nodes / 25 + 0.6, 16)
+        if head_node_memory < required_memory:
+            self._add_failure(
+                f"Head node instance type {head_node_instance_type} has {head_node_memory} GB of memory. "
+                f"Please choose a head node instance type with at least {required_memory} GB of memory"
+                f" to manage {total_max_compute_nodes} compute nodes.",
+                FailureLevel.ERROR,
```
hehe7318, line 1332: Why decide to use FailureLevel.Error here but not a Warning? I think customers have the right to decide what instance type they pay for, but we have a responsibility to warn them.

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,41 @@ def _validate(self, imds_secured: bool, scheduler: str):
             )
 
 
+class HeadNodeInstanceTypeValidator(Validator):
```
dreambeyondorange: This should probably be more specific that just `instance type` and mention memory in the name so folks know what its checking

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,41 @@ def _validate(self, imds_secured: bool, scheduler: str):
...
+        if head_node_memory < required_memory:
+            self._add_failure(
+                f"Head node instance type {head_node_instance_type} has {head_node_memory} GB of memory. "
+                f"Please choose a head node instance type with at least {required_memory} GB of memory"
+                f" to manage {total_max_compute_nodes} compute nodes.",
+                FailureLevel.ERROR,
+            )
+
+
+class SharedEbsClusterSizeValidator(Validator):
```
dreambeyondorange: Same naming comment as above, we should be able to understand what it does by it's name

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,41 @@ def _validate(self, imds_secured: bool, scheduler: str):
...
+    """Warn potential performance bottleneck of using Shared EBS."""
+
+    def _validate(self, total_max_compute_nodes: int):
+        if total_max_compute_nodes > 100:
+            self._add_failure(
+                "EBS shared storage is mounted on the head node and shared to the compute nodes. "
+                "This is a performance bottle neck if the compute nodes rely on this shared storage. "
```
dreambeyondorange: It's not clear why this is a performance bottleneck.  Performance of what components?

cli/src/pcluster/validators/cluster_validators.py
hanwen-cluster: Because the requirement is very relaxed.
For example, the validator allows t3.micro to manage 10 nodes, t3.medium to manage 85 nodes, t3.xlarge to manage infinite nodes. If the requirement is violated, it is very likely the scaling up would fail.
Also, the customer could suppress the validation anyway

hehe7318: Convinced me, I agree.

cli/src/pcluster/validators/cluster_validators.py
hanwen-cluster: Done

cli/src/pcluster/validators/cluster_validators.py
hanwen-cluster: Done

cli/src/pcluster/validators/cluster_validators.py
hanwen-cluster: Done

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,42 @@ def _validate(self, imds_secured: bool, scheduler: str):
             )
 
 
+class HeadNodeMemorySizeValidator(Validator):
+    """
+    Head Node Instance Type Validator.
```
dreambeyondorange: Probably need to change this to match

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1311,6 +1311,42 @@ def _validate(self, imds_secured: bool, scheduler: str):
...
+
+
+class SharedEbsPerformanceBottleNeckValidator(Validator):
+    """Warn potential performance bottleneck of using Shared EBS."""
+
+    def _validate(self, total_max_compute_nodes: int):
+        if total_max_compute_nodes > 100:
+            self._add_failure(
+                "EBS shared storage is mounted on the head node and shared to the compute nodes. "
+                "Therefore, the head node network bandwidth is a performance bottle neck "
```
dreambeyondorange: Maybe call it network performance? It’s still ambiguous like this

hanwen-cluster: Done

cli/src/pcluster/validators/cluster_validators.py
hanwen-cluster: Done

dreambeyondorange (approved):
Nice work!

PR #6645, opened 2025-01-23
[integ-test] rotate instance types for some integration tests

hanwen-cluster:
1. Unlike rotating OS, instance types are region dependent. Therefore, we cannot use general Jinja variables like `{{ OS_X86_1 }}`. We need to use region specific Jinja variables like `{{ US_EAST_1_INSTANCE_TYPE_0 }}`
2. For code efficiency, this commit only populates three large AWS regions. The code is extendable if more regions should be added.
3. This commit rotates instance types only on `test_essential_features` and `test_cluster_with_gpu_health_checks`. The code is extendable if more tests should be added.
4. Improve `test_cluster_with_gpu_health_checks` to be able to run on both x86 and arm

### Checklist

### DIFF
--- a/tests/integration-tests/framework/tests_configuration/config_renderer.py
+++ b/tests/integration-tests/framework/tests_configuration/config_renderer.py
@@ -13,6 +13,7 @@ import logging
 import os
 from datetime import date
 
+import boto3
 import yaml
 from jinja2 import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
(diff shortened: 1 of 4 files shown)


tests/integration-tests/framework/tests_configuration/config_renderer.py
```diff
@@ -60,6 +61,94 @@ def _get_os_parameters(config=None, args=None):
...
+            for index in range(len(xlarge_instances)):
+                instance_type = xlarge_instances[(today_number + index) % len(xlarge_instances)]
+                result[f"{region_jinja}_INSTANCE_TYPE_{index}"] = instance_type[: -len(".xlarge")]
+                availability_zones = instance_type_availability_zones[instance_type]
+                result[f"{region_jinja}_INSTANCE_TYPE_{index}_AZ"] = (
+                    availability_zones[0] if len(availability_zones) <= 2 else region
+                )
+            for index in range(len(gpu_instances)):
+                instance_type = gpu_instances[(today_number + index) % len(gpu_instances)]
+                result[f"{region_jinja}_GPU_INSTANCE_TYPE_{index}"] = instance_type[: -len(".xlarge")]
```
hgreebe, line 136: why are we removing ".xlarge" if we just add it back in the develop.yaml. Especially since we are only looking at xlarge instances.

hanwen-cluster: To make the approach more extensible.
For example `test_dcv_configuration` uses `.2xlarge`. If we lock down to `.xlarge`, this mechanism cannot be used by `test_dcv_configuration`

hgreebe: but right now we are already locked down to `.xlarge` due to this line `if instance_type["InstanceType"].endswith(".xlarge")`. If you really wanted to make it extendable, then shouldn't `_get_instance_type_parameters()` have a parameter that takes in a list of instance sizes and have the output be a map where the key is the instance size and the value is the list of instance types.

hgreebe (approved):

PR #6665, opened 2025-02-20
[Test] In test_build_image_custom_components replaced unnecessary complex component with hello world component to reduce noise

gmarciani:
### Description of changes
In test_build_image_custom_components replaced the component stig-build-linux-high with the hello-world-linux to reduce unnecessary complexity and noise. Also, validate both script and arn based components on all OSes. 
 
 The removal of the stig component seems an excessive relaxation, but it's not. The goal of the test is to validate that PC is able to build images with custom components, regardless their content. There is no reason to introduce such complexity in this test, causing noise.

### Tests
test_build_image_custom_components succeeds (tested on RHEL9, which was failing with the STIG component)


tests/integration-tests/tests/createami/test_createami/test_build_image_custom_components/image.config.yaml
```diff
@@ -2,12 +2,7 @@ Build:
   InstanceType: {{ instance_type }}
   ParentImage: {{ parent_image }}
   Components:
-    # Test arn custom component in eu-west-1 except Rocky and Ubuntu because stig-build-linux-high component doesn't support these ParallelCuster AMIs.
-    # Test script custom component in other regions.
-    {% if region == "eu-west-1" and "rocky" not in os and "ubuntu" not in os %}
     - Type: arn
-      Value: arn:{{ partition }}:imagebuilder:{{ region }}:aws:component/stig-build-linux-high/2024.2.3/1
-    {% else %}
+      Value: arn:{{ partition }}:imagebuilder:{{ region }}:aws:component/hello-world-linux/1.0.0/1
```
himani2411, line 6: is this available in all regions ?

himani2411: or the 2 regions we use in test?

gmarciani: The hello-world is available in all commercial regions

himani2411 (approved):

PR #6678, opened 2025-02-27
Fix cuda deviceQuery test

hgreebe:
### Description of changes
* Fix cuda deviceQuery test
* AL2 is still on cuda 12.8, so it still uses make instead of cmake
* ubuntu2004 does not have a high enough cmake version, it would need to be installed from the internet, this change just excludes running device query on ubuntu2004

### Tests
* Ran second stage build for all OSs

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
@@ -309,18 +309,32 @@ phases:
               echo "Correctly installed CUDA ${cuda_output}"
 
               echo "Testing CUDA with deviceQuery..."
-              if [ {{ validate.OperatingSystemArchitecture.outputs.stdout }} != 'arm64' ]; then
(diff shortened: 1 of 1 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml
```diff
@@ -309,18 +309,32 @@ phases:
               echo "Correctly installed CUDA ${cuda_output}"
 
               echo "Testing CUDA with deviceQuery..."
-              if [ {{ validate.OperatingSystemArchitecture.outputs.stdout }} != 'arm64' ]; then
-                /usr/local/cuda-${cuda_ver}/extras/demo_suite/deviceQuery | grep -o "Result = PASS"
-                [[ $? -ne 0 ]] && echo "CUDA deviceQuery test failed" && exit 1
+              if [ {{ validate.OperatingSystemName.outputs.stdout }} == 'ubuntu2004' ]; then
```
himani2411, line 312: So cant we install the latest version of cmake/make on Ubuntu2004?

himani2411 (approved):

PR #6699, opened 2025-03-11
[Integ Test] Increase stop_max_delay to fix test_error_handling

hehe7318:
### Description of changes
* The test_error_handling is failed because compute node launch time is longer than 5 minutes. But eventually it became fully functional. In our observation, the compute node is responded just 1 second after the timeout. So increase the timeout to 8 minutes to ensure will not cause a test failure.

### Checklist


hanwen-cluster (dismissed):

gmarciani:
What about increasing to 6m rather than 8m. Apperently it's enough and we should always minimize the relaxation to the bare minimum required.

gmarciani (approved):

PR #6722, opened 2025-03-14
[Storage] Restore default EBS volume size to 40GB. This is the size used for shared storage.

gmarciani:
### Description of changes
Restore default EBS volume size to 40GB. This is the size used for shared storage.
Set EBS root volume size to 45GB in ADC as we recently increased the AMI volume size.
This is required only in ADC because in these regions the root volume must be explicitly set. In Commercial regions instead, it is inherited from the AMI volume size.

### Tests
ONGOING test_ebs_single


cli/src/pcluster/constants.py
```diff
@@ -125,7 +125,9 @@
     "io2": 100,
     "gp3": 3000,
 }
-EBS_VOLUME_SIZE_DEFAULT = 45
+EBS_VOLUME_SIZE_DEFAULT = 40
+# This value must be equal to the AMI volume size
```
hanwen-cluster: This value must be greater or equal to the AMI volume size

hanwen-cluster (dismissed):

hanwen-cluster (approved):

PR #6727, opened 2025-03-14
[Test] Fix test shared home by setting the expected size for the shared ebs volume.

gmarciani:
### Description of changes
Fix test shared home by setting the expected size for the shared ebs volume.

We must match the value of 45Gb set in the cluster config:

```
    EbsSettings:
      Raid:
        Type: 1
      Size: 45
```

### DIFF
--- a/tests/integration-tests/tests/storage/test_shared_home.py
+++ b/tests/integration-tests/tests/storage/test_shared_home.py
@@ -123,8 +123,8 @@ def _check_shared_home(
             scheduler_commands, remote_command_executor, f"{mount_dir}/{get_username_for_os(os)}"
         )
     elif storage_type == "Ebs":
-        _test_ebs_correctly_mounted(remote_command_executor, mount_dir, volume_size=40)
-        _test_ebs_correctly_mounted(remote_command_executor_login_node, mount_dir, volume_size=40)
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/storage/test_shared_home.py
```diff
@@ -123,8 +123,8 @@ def _check_shared_home(
             scheduler_commands, remote_command_executor, f"{mount_dir}/{get_username_for_os(os)}"
         )
     elif storage_type == "Ebs":
-        _test_ebs_correctly_mounted(remote_command_executor, mount_dir, volume_size=40)
-        _test_ebs_correctly_mounted(remote_command_executor_login_node, mount_dir, volume_size=40)
+        _test_ebs_correctly_mounted(remote_command_executor, mount_dir, volume_size=45)
```
himani2411, line 126: Can you add change for this line too https://github.com/aws/aws-parallelcluster/pull/6725?

himani2411 (approved):

tests/integration-tests/tests/storage/test_shared_home.py
gmarciani: not needed, we reverted the default shared ebs size to 40

PR #6747, opened 2025-03-18
Add Validator to check OS & Capacity Reservation Platform match

himani2411:
### Description of changes
* Add Validator to check OS Capacity Reservation Platform match

### Tests
* Unit Tests and manual tests
```
{
    {
      "level": "ERROR",
      "type": "CapacityReservationValidator",
      "message": "Capacity reservation cr-003991131305dd2c3 has platform Linux/UNIX, which is not compatible with the cluster OS rhel9. Please use a reservation with platform Red Hat Enterprise Linux."
    },


```

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -49,6 +49,25 @@ NVIDIA_OPENRM_UNSUPPORTED_INSTANCE_TYPES = ["p3", "p3dn", "p2", "g3", "g3s", "g2
 SLURM = "slurm"
 AWSBATCH = "awsbatch"
 
+
+#  Capacity Reservation Platform types we support.
+CR_PLATFORM_LINUX_UNIX = "Linux/UNIX"
+CR_PLATFORM_UBUNTU_PRO = "Ubuntu Pro"
+CR_PLATFORM_RHEL = "Red Hat Enterprise Linux"
+
+
+CAPACITY_RESERVATION_OS_MAP = {
+    "alinux2": CR_PLATFORM_LINUX_UNIX,
+    "alinux2023": CR_PLATFORM_LINUX_UNIX,
+    "ubuntu2004": CR_PLATFORM_UBUNTU_PRO,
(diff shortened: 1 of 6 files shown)


cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -331,9 +335,22 @@ def _validate(
         is_flexible: bool,
         subnet: str,
         capacity_type: CapacityType,
+        os,
     ):
         if capacity_reservation_id:
             capacity_reservation = AWSApi.instance().ec2.describe_capacity_reservations([capacity_reservation_id])[0]
+            cr_platform = capacity_reservation.instance_platform()
+            if CAPACITY_RESERVATION_OS_MAP.get(cr_platform):
```
gmarciani: the way the validator fails, may be misleading. Example:
1. user wants to create cluster with OS AL23
2. user creates a reservation with platform WHATEVER_BUT_NOT_THE_CORRECT_ONE
3. the error message says to create a reservation with one of Linux/Unix, Ubuntu Pro, RHEL
4. it creates a RHEL reservation
5. now the validator fails with the helpful message, saying that AL23 requires Linux/Unix platform.

Proposed solution: fail with just one error message saying that the capacity reservation {CR_ID} has platform {PLATFORM}, which is not compatible with the cluster OS {OS}. Please use a reservation with platform {CORRECT_PLATFORM}

What do you think?

himani2411: will make changes

cli/src/pcluster/constants.py
```diff
@@ -49,6 +49,18 @@
 SLURM = "slurm"
 AWSBATCH = "awsbatch"
 
+CAPACITY_RESERVATION_OS_MAP = {
+    "alinux2": "Linux/UNIX",
```
gmarciani: Minor: can we define the reservation platforms as constants?
We should do the same for OS names, but this is a gap we have despite this PR.

gmarciani (dismissed):

gmarciani (approved):

PR #6771, opened 2025-04-01
[Build Image] Manage Kernel consistency through version locking

himani2411:
### Description of changes
* Pin Kernel in parallelcluster.yaml before we install any packages and unpin the kernel so that CX can update it on the AMI which is generated.

### Tests
rhel8 ami 

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -182,23 +182,6 @@ phases:
                 fi
               fi
 
-      - name: CreatingJsonFile
-        action: CreateFile
-        inputs:
-          - path: /etc/parallelcluster/image_dna.json
-            content: |
-              ${CfnParamChefDnaJson}
-            overwrite: true
-
-      - name: DisableKernelUpdate
-        action: ExecuteBash
-        inputs:
-          commands:
-            - |
-              set -v
-              DISABLE_KERNEL_UPDATE=$(cat /etc/parallelcluster/image_dna.json | jq -r '.cluster.disable_kernel_update.enabled')
-              echo "${!DISABLE_KERNEL_UPDATE}"
-
       - name: PinKernelVersion
         action: ExecuteBash
         inputs:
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/config/imagebuilder_config.py
```diff
@@ -237,6 +237,7 @@ class ImagebuilderDevSettings(BaseDevSettings):
     def __init__(
         self,
         disable_pcluster_component: bool = None,
+        disable_user_data: bool = None,
```
hanwen-cluster: This name doesn't seem to reflect the actual functionality.

Moreover, why is it necessary to have a separate parameter? This improvement could be added as the default behavior, right?

cli/src/pcluster/templates/imagebuilder_stack.py
```diff
@@ -385,8 +385,24 @@ def _add_imagebuilder_distribution_configuration(self, ami_tags, build_tags, lam
...
+        additional_instance_configuration_property = None
+        imagebuilder_user_data = os.path.join(
+            imagebuilder_utils.get_resources_directory(), "imagebuilder", "user_data.sh"
+        )
+        with open(imagebuilder_user_data, "r", encoding="utf-8") as user_data_file:
+            user_data_content = user_data_file.read()
+
+        additional_instance_configuration_property = (
+            imagebuilder.CfnImageRecipe.AdditionalInstanceConfigurationProperty(
+                user_data_override=Fn.base64(user_data_content)
```
hanwen-cluster: Do you think this override some default stuffs from image builder?
```
userDataOverride

    Use this property to provide commands or a command script to run when you launch your build instance.

    The userDataOverride property replaces any commands that Image Builder might have added to ensure that Systems Manager is installed on your Linux build instance. If you override the user data, make sure that you add commands to install Systems Manager, if it is not pre-installed on your base image.
```
https://docs.aws.amazon.com/imagebuilder/latest/APIReference/API_AdditionalInstanceConfiguration.html

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -182,6 +182,36 @@ phases:
...
+                fi
+
+                if [[ ${!OS} == "rocky8" ]] || [[ ${!OS} == "rocky9" ]] ; then
+                  yum versionlock rocky-release rocky-repos
+                elif [[ ${!OS} == "rhel8" ]] || [[ ${!OS} == "rhel9" ]] ; then
+                  yum versionlock redhat-release
+                fi
+              else
+                apt-mark hold linux-aws* linux-base* linux-headers* linux-image* 
+              fi
```
hanwen-cluster, line 212: Can we pin kernel packages only if the base AMI does not contain relevant pinning?

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -308,6 +332,38 @@ phases:
...
+                  yum versionlock delete redhat-release
+                fi
+              else
+                apt-mark unhold linux-aws* linux-base* linux-headers* linux-image* 
+              fi
+              if [[ "$(uname -a)" != ${!BEFORE_KERNEL_VERSION} ]] ; then
+                echo "Kernel was upgraded from ${!BEFORE_KERNEL_VERSION} to $(uname -a)"
+              fi
+              echo "Kernel version unlocked"
+
```
hanwen-cluster: Similarly, can we unpin only if the base AMI does not contain any pins

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
hanwen-cluster: You can use similar trick used to pin RHEL minor release:
```
                  if [[ ! -f /etc/yum/vars/releasever ]]; then
                    echo "yes" > /opt/parallelcluster/pin_releasesever
                    echo ${!VERSION} > /etc/yum/vars/releasever
                    yum clean all
                  fi
```

himani2411: My bad! i didnt push the latest chnages!

cli/src/pcluster/templates/imagebuilder_stack.py
himani2411: Will do!

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
himani2411: why? If we are doing this by default it shouldn't matter if the Parent AMI has some sort of pinning or not.

also the version lock we do for all the non-debian OSSes is on specific package names without wildcards.
I will take a look for debian again!

hanwen-cluster (dismissed):

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani: +1 on this, if the user uses a parent image with kernel pinning we should not silently remove it.

himani2411: https://github.com/aws/aws-parallelcluster/pull/6771#discussion_r2089146608

hanwen-cluster (approved):

PR #6860, opened 2025-06-04
Push performance data to DynamoDB and improve StarCCM test

hanwen-cluster:
### Description of changes
* See commit descriptions for details

### DIFF
--- a/tests/integration-tests/tests/performance_tests/common.py
+++ b/tests/integration-tests/tests/performance_tests/common.py
@@ -3,16 +3,20 @@ import logging
 import os as os_lib
 import pathlib
 import shutil
+import time
+import uuid
 from math import ceil
 from os import makedirs
 
 import boto3
 from assertpy import assert_that
+from framework.framework_constants import METADATA_DEFAULT_REGION, PERFORMANCE_METADATA_TABLE
+from framework.metadata_table_manager import MetadataTableManager
 from retrying import retry
 from time_utils import seconds
 from utils import get_username_for_os
 
-from tests.common.utils import fetch_instance_slots
+from tests.common.utils import fetch_instance_slots, get_installed_parallelcluster_version
 
 # Common settings used by all test scenarios
 # You can change these variables to adapt the performance test to your needs.
@@ -249,3 +253,33 @@ def _log_output_performance_difference(node, performance_degradation, observed_v
         f"Nodes: {node}, Baseline: {baseline_value} seconds, Observed: {observed_value} seconds, "
         f"Percentage difference: {percentage_difference}%, Outcome: {outcome}"
     )
+
+
+def push_result_to_dynamodb(name, result, instance, os):
+    reporting_region = METADATA_DEFAULT_REGION
+    logging.info(f"Metadata reporting region {reporting_region}")
+    # Create the metadata table in case it doesn't exist
+    MetadataTableManager(reporting_region, PERFORMANCE_METADATA_TABLE).create_metadata_table()
+    try:
+        # Create DynamoDB resource
+        dynamodb = boto3.resource("dynamodb", region_name=reporting_region)
+        table = dynamodb.Table(PERFORMANCE_METADATA_TABLE)
+
+        # Prepare item to be inserted
+        item = {
+            "id": str(uuid.uuid4().hex),
+            "name": name,
+            "instance": instance,
+            "os": os,
+            "timestamp": int(time.time()),
+            "result": str(result),
+            "pcluster_version": f"v{get_installed_parallelcluster_version()}",
+        }
+
+        # Put item in the table
+        table.put_item(Item=item)
+        logging.info(f"Successfully pushed result to DynamoDB with id: {item['id']}")
+
+    except Exception as e:
+        logging.error(f"Failed to push result to DynamoDB: {str(e)}")
+        raise
--- a/tests/integration-tests/tests/performance_tests/test_starccm.py
+++ b/tests/integration-tests/tests/performance_tests/test_starccm.py
@@ -6,7 +6,7 @@ import pytest
 from remote_command_executor import RemoteCommandExecutionError, RemoteCommandExecutor
 
 from tests.common.utils import assert_no_file_handler_leak, get_compute_ip_to_num_files
-from tests.performance_tests.common import _log_output_performance_difference
+from tests.performance_tests.common import _log_output_performance_difference, push_result_to_dynamodb
 
 # timeout in seconds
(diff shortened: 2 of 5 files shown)


tests/integration-tests/tests/performance_tests/common.py
```diff
@@ -249,3 +252,33 @@ def _log_output_performance_difference(node, performance_degradation, observed_v
         f"Nodes: {node}, Baseline: {baseline_value} seconds, Observed: {observed_value} seconds, "
         f"Percentage difference: {percentage_difference}%, Outcome: {outcome}"
     )
+
+
+def push_result_to_dynamodb(name, result, instance, os):
+    reporting_region = "us-east-1"
```
gmarciani: I suppose DDB table is deployed ion the region where Jenkins is, so it may not be us-east-1.
Can we retrieve it dynamically?

tests/integration-tests/tests/performance_tests/common.py
```diff
@@ -249,3 +252,33 @@ def _log_output_performance_difference(node, performance_degradation, observed_v
...
+    logging.info(f"Metadata reporting region {reporting_region}")
+    # Create the metadata table in case it doesn't exist
+    MetadataTableManager(reporting_region, "ParallelCluster-PerformanceTest-Metadata").create_metadata_table()
+    try:
+        # Create DynamoDB resource
+        dynamodb = boto3.resource("dynamodb", region_name=reporting_region)
+        table = dynamodb.Table("ParallelCluster-PerformanceTest-Metadata")
```
gmarciani: Let's define the table name in a constant. If we have multiple table in the DDB better to have a unique place where the table names are defined

tests/integration-tests/tests/performance_tests/test_starccm.py
```diff
@@ -99,33 +99,44 @@ def test_starccm(
...
+            for _ in range(parallelism):
+                parallel_runs.append(remote_command_executor.run_remote_command(run_command))
+            for run in parallel_runs:
+                result.append(
+                    calculate_observed_value(
+                        run, remote_command_executor, scheduler_commands, test_datadir, num_of_nodes
+                    )
+                )
+        if parallelism == 1:
+            for _ in range(2):
```
gmarciani: Let's avoid magic numbers. where possible. 2 is the size of the executions. let;'s have a constant for that

tests/integration-tests/tests/performance_tests/test_starccm.py
```diff
@@ -99,33 +99,44 @@ def test_starccm(
...
-        result_32, remote_command_executor, scheduler_commands, test_datadir, number_of_nodes[2]
-    )
+    max_node_num = max(number_of_nodes)
+    final_result = []
+    for num_of_nodes in number_of_nodes:
+        parallelism = int(max_node_num / num_of_nodes)
+        result = []
+        logging.info(f"Submitting StarCCM+ job with {num_of_nodes} nodes")
+        run_command = f'sbatch --ntasks={num_of_nodes * TASK_VCPUS} starccm.slurm.sh "{podkey}" "{licpath}"'
+        if parallelism > 1:
```
gmarciani: in this way we will have 4 parallel runs for 8 nodes, 2 for 16 nodes and 1 for 32 nodes. Is it intentional?

tests/integration-tests/tests/performance_tests/test_starccm.py
```diff
@@ -99,33 +99,44 @@ def test_starccm(
...
+            parallel_runs = []
+            for _ in range(parallelism):
+                parallel_runs.append(remote_command_executor.run_remote_command(run_command))
+            for run in parallel_runs:
+                result.append(
+                    calculate_observed_value(
+                        run, remote_command_executor, scheduler_commands, test_datadir, num_of_nodes
+                    )
+                )
+        if parallelism == 1:
```
gmarciani: why a specific if branch?  wouldn't it it be enough to do on line 113

```
num_executions = max(parallelism, 2)
for _ in range(num_executions):
```

tests/integration-tests/tests/performance_tests/common.py
hanwen-cluster: We can stably make it us-east-1. This behavior is consistent with the ParallelCluster-IntegTest-Metadata database.

tests/integration-tests/tests/performance_tests/test_starccm.py
hanwen-cluster: Yes, this is intentional, just to maximize the usage of nodes to give results as stable as possible

tests/integration-tests/tests/performance_tests/test_starccm.py
hanwen-cluster: When parallelism is 1, we need to execute the two runs of test sequentially. Merging them will cause the test runs to always run in parallel

tests/integration-tests/tests/performance_tests/test_starccm.py
hanwen-cluster: You mean changing 
```
for _ in range(2):
```
to
```
number_of_runs = 2
for _ in range(number_of_runs):
```
Why should we add another line and increase the complexity of the code?

tests/integration-tests/tests/performance_tests/common.py
hanwen-cluster: Done

tests/integration-tests/tests/performance_tests/test_starccm.py
gmarciani: Adding a line does not increase complexity, especially if it is meant to increase readability.
Avoiding magic numbers is a general clean code best practice.
This is a minor, not a blocking comment, but we should always keep this clean code best practice in mind.
Ref: https://www.pluralsight.com/tech-blog/avoiding-magic-numbers/

tests/integration-tests/tests/performance_tests/test_starccm.py
gmarciani: When parallelism=1 the two executions would be put in queue by the sbatch command, so in any case they would be sequential. Do you agree?

tests/integration-tests/tests/performance_tests/test_starccm.py
hanwen-cluster: Done

tests/integration-tests/tests/performance_tests/test_starccm.py
hanwen-cluster: Great discovery! Thank you! Done

gmarciani (approved):

PR #6887, opened 2025-06-17
Adjust error message to show that the capacity reservation does not exist

hgreebe:
### Description of changes
* Adjust error message to show that the capacity reservation does not exist.
* Orginal error can not be thrown [here](https://github.com/aws/aws-parallelcluster/blob/b1093071e72abddc9bf86838f736300678fb80e2/cli/src/pcluster/config/cluster_config.py#L2448-L2455) because then it causes a failure during cluster update.
* This change only catches the error during cluster creation validation

### Tests
* Tested that the appropriate error message occurs during cluster creating when a capacity reservation is specified but not the instance type
* Tested that the appropriate error message occurs during cluster creation when neither a capacity reservation or instance type is specified.
* Ran full set of integration tests.

### References
* Related to this change: https://github.com/aws/aws-parallelcluster/pull/6883

### DIFF
--- a/cli/src/pcluster/models/cluster.py
+++ b/cli/src/pcluster/models/cluster.py
@@ -500,6 +500,8 @@ class Cluster:
         except ConfigValidationError as e:
             raise e
         except Exception as e:
+            if "Invalid type for parameter InstanceTypes[0], value: None, type: <class 'NoneType'>" in e.__str__():
+                raise ConfigValidationError(f"Invalid cluster configuration: Capacity reservation does not exist.")
             raise ConfigValidationError(f"Invalid cluster configuration: {e}")
 
         return config, validation_failures


cli/src/pcluster/models/cluster.py
```diff
@@ -500,6 +500,8 @@ def _validate_and_parse_config(
         except ConfigValidationError as e:
             raise e
         except Exception as e:
+            if "Invalid type for parameter InstanceTypes[0], value: None, type: <class 'NoneType'>" in e.__str__():
```
gmarciani, line 503: The logici around exception raising should not be based on error messages, as it is a fragile approach. Error messages are meant to change over time.
The logic should be based on something more stable, such as exception types.

gmarciani: The best approach would be to raise a ConfigValidationError with a clear error message when we set the instancetype to None

PR #6898, opened 2025-06-22
Add unit test that checks that describe_capacity_reservation does not error when load a config file.

hgreebe:
### Description of changes
* Add unit test that checks that describe_capacity_reservation does not error when load a config file.
* Add unit test that checks that retrieving the instance type attribute of the old config does not occur during an update.
* During an update, both the old and the new config are loaded. The new config is then validated using the `_validate_and_parse_config` function. The loading of the old config should not cause an error.
* When creating a cluster, the config is also validated during the `_validate_and_parse_config` function.
* Validated that the unit test fails without the fix from this [PR](https://github.com/aws/aws-parallelcluster/pull/6896)

Behavior of the following use cases:
1. Create/Update a cluster with an input config presenting a non-existing or expired reservation and no instance type leads to an error telling the user that the CR does not exist. 
  a. This causes a validation error when `_validate_and_parse_config` is called.
  b. We have existing capacity reservation validators that make sure the capacity reservation is accessible

2. Create/Update cluster with an input config presenting a valid reservation but with the caller lacking describe_capacity_reservations permissions should display the relevant error message.
   a. We have existing unit tests that check this behavior.

3. Update cluster with the old config having an expired reservation. What happens if I provide the same config as input for an update? Will it fail saying the reservation for the input config is invalid?
  a. This unit test checks that loading the old config should not result in an error if the capacity reservation is expired. 
  b. If you provide the same config or an invalid config as input for an update, it will result in an error during the validation phase. This is because the old config is checked `_validate_and_parse_config` function.

### References
* Related to this change: https://github.com/aws/aws-parallelcluster/pull/6896


cli/tests/pcluster/config/test_config_patch.py
```diff
@@ -256,6 +258,103 @@ def test_single_param_change(
...
+)
+
+
+def test_capacity_reservation_patch(
+        mocker,
+        test_datadir,
+        pcluster_config_reader,
+        cr_id,
+        instance_type,
+        cr_exists,
```
hanwen-cluster: Where is this used?

hanwen-cluster (dismissed):

cli/tests/pcluster/config/test_config_patch.py
```diff
@@ -256,6 +258,96 @@ def test_single_param_change(
     _check_patch(src_conf.source_config, dst_conf.source_config, [expected_change], change_update_policy)
 
 
+@pytest.mark.parametrize(
```
gmarciani, line 261: 1. Are those new unit test able to capture the failure we are fixing in https://github.com/aws/aws-parallelcluster/pull/6896?
If so, please specify in the PR description. If not, we should introduce a unit test that protects us against regression. A possible approach is to run those unit test on the codebase without the fix and verify that they fail meaningfully.
2. n the fix we introduced a function that should not be called as part of the update path, otherwise it may cause a failure. Can we capture this in a unit test? We want to have a unit test that fails if that function is invoked as part of the update path.
3. What about unit tests for the happy case, i.e. when capacity reservation is set an exists?

cli/tests/pcluster/config/test_config_patch.py
```diff
@@ -256,6 +258,96 @@ def test_single_param_change(
...
+            "cr-12324398",
+            "c5.xlarge",
+            None,
+            None
+        )
+    ],
+)
+
+
+def test_capacity_reservation_patch(
```
gmarciani: We have the unit test on the update path. What about the unit test for the create path?

cli/tests/pcluster/config/test_config_patch.py
hgreebe: 1. It is able to capture the failure. The unit tests fail without the fix. I added this to the PR description.
2. I will see if I can capture this in a unit test.
3. For this unit test if shouldn't matter whether or not the capacity reservation exists because the describe_capacity_reservation function is never called.

cli/tests/pcluster/config/test_config_patch.py
hgreebe: This also applies to the create path because _load_config should work the same in both update and create. The config is validation is done in a different function and we have other unit tests that cover this.

cli/tests/pcluster/config/test_config_patch.py
gmarciani: Thanks for the iteration. On point (3) if we expect that function to never be called, then it would be better to add  an assertion to check that it's never called

cli/tests/pcluster/config/test_config_patch.py
gmarciani: If this test covers both the create and update path then what about calling the test `test_load_config_for_capacity_reservation` and specify in the comment that load_config is called both in the create and update path?

cli/tests/pcluster/config/test_config_patch.py
hgreebe: Sorry, I should have described it better. It is called in [`SlurmClusterConfig`](https://github.com/aws/aws-parallelcluster/blob/53eea7bf7b95e85abeb59e2b16b67d0d8bbf7928/cli/src/pcluster/config/cluster_config.py#L2929) but if there is an error it only logs a warning, so that didn't affect the loading process.
```
try:
  AWSApi.instance().ec2.describe_capacity_reservations(self.all_relevant_capacity_reservation_ids)
except AWSClientError:
  logging.warning("Unable to cache describe_capacity_reservations results for all capacity reservation ids.")
```

gmarciani: Ok, so we should make it clearer in the test description + assert that the function is called.

cli/tests/pcluster/cli/test_update_cluster.py
```diff
@@ -240,6 +247,86 @@ def test_no_nodejs_error(self, mocker, test_datadir):
...
+                }
+            ),
+            config=old_configuration,
+        )
+
+        mocker.patch("pcluster.aws.cfn.CfnClient.stack_exists", return_value=True)
+        mocker.patch("pcluster.aws.cfn.CfnClient.stack_exists", return_value=True)
+
+        # If describe_capacity_reservations is called, validate_update_request will fail.
+        cluster.validate_update_request(new_configuration, get_validator_suppressors(["ALL"]), force=True)
```
gmarciani: I agree with you that if describe_capacity_reservations()  would be called this line would fail, so this line would be enough. However, I suggest to add an explicit check (assert_not_called, see example of usage [here](https://github.com/aws/aws-parallelcluster/blob/11dc17f99de6bfe08bbc1855bca371fe88bae0dc/cli/tests/pcluster/cli/test_export_image_logs.py#L132) verifying that the mocked function describe_capacity_reservations is never called to make the test more robust, more explicit and more intentional.

gmarciani (dismissed):

hanwen-cluster (approved):
Approved based on @gmarciani 's approval

Issue #5914, opened 2023-12-04
(3.0.0-latest) Build image CloudFormation stacks fail to delete after images are successfully built 


hanwen-pcluste:
## The issue
Starting from September 19, 2023, build image CloudFormation stack ends in DELETE_FAILED status after the image is successfully built. The failures in the stack events look like:

| Timestamp | Logical ID | Status | Status reason |
| ----------- | ----------- | ----------- | ----------- |
| 2023-12-01 06:00:20 UTC-0800 | aws-parallelcluster-3-7-2-amzn2-hvm-arm64-202312011211 | DELETE_FAILED | The following resource(s) failed to delete: [DeleteStackFunctionExecutionRole]. |
| 2023-12-01 06:00:19 UTC-0800 | DeleteStackFunctionExecutionRole | DELETE_FAILED | Internal Failure |

The image is built correctly despite the stack is in DELETE_FAILED and you can use it as custom AMI for cluster creation.

## Affected versions
ParallelCluster versions 3.0.0-3.10.1 are affected.

## Mitigation

See details in Wiki https://github.com/aws/aws-parallelcluster/wiki/(3.0.0-and-later)-Build-image-CloudFormation-stacks-fail-to-delete-after-images-are-successfully-built

gwolski:
This issue still affect 3.13.1.  It would be great if the mitigation page provided a bit more of a recipe for a novice type user exactly what steps they need to take to implement the mitigation.
Should one just include the additional policy arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole or does one need to create a new policy as shown on the referenced AWS Lambda cleanup role section?

himani2411:
This issue was resolved in  https://github.com/aws/aws-parallelcluster/releases/tag/v3.14.0

PR #6912, opened 2025-07-09
[Develop] Introduce Global Cleanup IAM Role for ParallelCluster Build-Image

hehe7318:
### Description of changes
* Introduce ensure_cleanup_role() with revision tagging and idempotent creation / update logic.
  * 4-step safe update sequence documented, now only after the inline policy succeeds, set or bump the revision tag.
  * Lambda VPCAccess managed policy is attached only when LambdaFunctionsVpcConfig exists in the config
* Modify image_operations_controller to invoke ensure_cleanup_role when Build/Iam/CleanupLambdaRole is not provided and to fail fast on permission errors.
* Refactor imagebuilder_stack to remove all per-stack cleanup-role logic and wire Lambda to the global role by default.
* Update constants (role prefix / expected revision tag key & value).
* IamClient – add create_role, attach_role_policy, put_role_policy, tag_role

### Tests
* Deleted legacy test_imagebuilder_lambda_execution_role.
* Fixed the failing tests
* Added new unit tests to cover the new logic
* Update existing integration test to test the build-image stack can be fully self deleted
* Manually tested build-image command in following scenario and all passed with expected result:
  1.  no custom cleanup role specified, and no cleanup role pre-exist
  2. no custom cleanup role specified, and cleanup role already exists with expected revision tag
  3. no custom cleanup role specified, and cleanup role already exists, but with outdated revision tag
  4. custom cleanup role specified
  5. no cleanup role pre-exist, and run build-image command with no enough permission role
  6. cleanup role already exists with expected revision tag, and run build-image command with no enough permission role
  7. cleanup role already exists but with outdated revision tag, and run build-image command with no enough permission role
* Newly added and old unit tests all passed.

### References
* https://github.com/aws/aws-parallelcluster/issues/5914

### Checklist


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
 
 **BUG FIXES**
 - Fix an issue where Security Group validation failed when a rule contained both IPv4 ranges (IpRanges) and security group references (UserIdGroupPairs).
+- Fix a bug that caused build-image CloudFormation stacks fail to delete after images are successfully built.
```
gmarciani: 1. `after images are successfully built.` it fails despite the outcome of the build image. even if the build-image fails, the stack is not able to get auto-deleted
2. `fail to delete` I'd say fail to auto-delete. Otherwise it may seem that the user cannot even delete the stack manually.

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,31 @@ def build_image(
     validation_failure_level = validation_failure_level or ValidationLevel.ERROR
     dryrun = dryrun or False
 
+    raw_cfg_str = build_image_request_content["imageConfiguration"]
+    cfg_dict = yaml.safe_load(raw_cfg_str) or {}
+    # If CleanupLambdaRole exists in the config, skip ensure_cleanup_role
+    has_custom_cleanup_role = cfg_dict.get("Build", {}).get("Iam", {}).get("CleanupLambdaRole")
+    # If LambdaFunctionsVpcConfig exists in the config, attach the AWS-managed LambdaVPCAccess policy
+    has_lambda_functions_vpc_config = cfg_dict.get("DeploymentSettings", {}).get("LambdaFunctionsVpcConfig")
```
gmarciani: Why do we need the lambda function vpc configuraitons?

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,31 @@ def build_image(
     validation_failure_level = validation_failure_level or ValidationLevel.ERROR
     dryrun = dryrun or False
 
+    raw_cfg_str = build_image_request_content["imageConfiguration"]
+    cfg_dict = yaml.safe_load(raw_cfg_str) or {}
+    # If CleanupLambdaRole exists in the config, skip ensure_cleanup_role
+    has_custom_cleanup_role = cfg_dict.get("Build", {}).get("Iam", {}).get("CleanupLambdaRole")
```
gmarciani, line 114: [CodeStyle] Encapsulate the logic to check whether a custom role is set into a dedicated function and unit test it independently.

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,31 @@ def build_image(
...
+    # If LambdaFunctionsVpcConfig exists in the config, attach the AWS-managed LambdaVPCAccess policy
+    has_lambda_functions_vpc_config = cfg_dict.get("DeploymentSettings", {}).get("LambdaFunctionsVpcConfig")
+
+    if not has_custom_cleanup_role:
+        try:
+            account_id = AWSApi.instance().sts.get_account_id()
+            ensure_cleanup_role(
```
gmarciani: [CodeStyle] `ensure_cleanup_role` is a bit vague name for this function. 
The name should reflect that iwe are talking about the default cleanup role and what cleanup role we are talking about

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,31 @@ def build_image(
...
+                account_id, get_partition(), attach_vpc_access_policy=True if has_lambda_functions_vpc_config else False
+            )
+        except AWSClientError as e:
+            if e.error_code in ("AccessDenied", "AccessDeniedException", "UnauthorizedOperation"):
+                raise BadRequestException(
+                    "Current principal lacks permissions to create or update the ParallelCluster build-image "
```
gmarciani, line 127: [Design] Why do we need to update the role>? Didn't we agreed on having versioned roles specifically to not update the policy?

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,31 @@ def build_image(
...
+                    "cleanup IAM role. "
+                    "Either pass `Build/Iam/CleanupLambdaRole` or grant the missing permissions to continue. "
+                    "For detailed instructions, please refer to our documentation. "
+                    "https://docs.aws.amazon.com/parallelcluster/latest/ug/iam-roles-in-parallelcluster-v3.html#iam-rol"
+                    "es-in-parallelcluster-v3-user-policy-build-image"
```
gmarciani: [UX] We should not add links to specific pages in our public documentation. Those links may change overtime and we would end up with CLI messages containing broken links.

cli/src/pcluster/constants.py
```diff
@@ -335,3 +335,8 @@ class Operation(Enum):
 PCLUSTER_BUCKET_PROTECTED_FOLDER = "parallelcluster"
 PCLUSTER_BUCKET_PROTECTED_PREFIX = f"{PCLUSTER_BUCKET_PROTECTED_FOLDER}/"
 PCLUSTER_BUCKET_REQUIRED_BOOTSTRAP_FEATURES = ["basic", "export-logs"]
+
+PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_PREFIX = "PClusterBuildImageCleanupRole"
+# Tag key & expected revision (increment when policy widens)
+CLEANUP_ROLE_REVISION_TAG_KEY = "parallelcluster:cleanup-role-revision"
```
gmarciani: Better to be more specific about what cleanup role we are referring to.
What if tomorrow we need to add other cleanup roles?
Example: `parallelcluster:build-image-cleanup-role-revision`

cli/src/pcluster/constants.py
```diff
@@ -335,3 +335,8 @@ class Operation(Enum):
...
+PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_EXPECTED_REVISION = 1
```
gmarciani: [Nit] Why `EXPECTED`? I think we can name it `PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION`

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
     """Generate action in imagebuilder components."""
     action = {"name": action_name, "action": "ExecuteBash", "inputs": {"commands": [commands]}}
     return action
+
+
+def _hash(account_id: str) -> str:
```
gmarciani: A function to return an hash may be useful even elsewhere, not only for build image. 
What about adding it to a general utils module?

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+                },
+                {
+                    "Action": "imagebuilder:DeleteImageRecipe",
+                    "Resource": f"arn:{partition}:imagebuilder:*:{account_id}:image-recipe/parallelclusterimage-*/*",
+                    "Effect": "Allow",
+                },
+                {
+                    "Action": "imagebuilder:DeleteDistributionConfiguration",
+                    "Resource": f"arn:{partition}:imagebuilder:*:{account_id}:distribution-configuration/"
+                    f"parallelclusterimage-*",
```
gmarciani, line 122: f-string not needed here

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+                    "Effect": "Allow",
+                },
+                {
+                    "Action": ["imagebuilder:DeleteImage", "imagebuilder:GetImage", "imagebuilder:CancelImageCreation"],
+                    "Resource": f"arn:{partition}:imagebuilder:*:{account_id}:image/parallelclusterimage-*/*",
+                    "Effect": "Allow",
+                },
+                {
+                    "Action": "cloudformation:DeleteStack",
```
gmarciani, line 131: [Security] Can we scope this down using conditions on tags? For example, we known that every pcluster stack will have at least pcluster version tag

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+                    "Resource": f"arn:{partition}:cloudformation:*:{account_id}:stack/*/*",
+                    "Effect": "Allow",
+                },
+                # The below two permissions are required for the DeleteStackFunction Lambda to tag the
+                # created AMI with 'parallelcluster:build_status' and 'parallelcluster:parent_image' tags
+                {"Action": "ec2:CreateTags", "Resource": f"arn:{partition}:ec2:*::image/*", "Effect": "Allow"},
+                {"Action": "tag:TagResources", "Resource": "*", "Effect": "Allow"},
```
gmarciani, line 141: [Security] Why on all resources? Can't we scope this down to specific resources?

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+    assume_doc = {
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+                "Effect": "Allow",
+                "Principal": {"Service": "lambda.amazonaws.com"},
+                "Action": "sts:AssumeRole",
+                "Condition": {
+                    "ArnLike": {
+                        "aws:SourceArn": f"arn:{partition}:lambda:*:{account_id}:function:ParallelClusterImage-*"
```
gmarciani, line 197: [Security] This condition allows every function in the stack to assume this role, but we want only a specific function to be able to. Can we use a more specific name?

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+                },
+            }
+        ],
+    }
+
+    # Check whether the role already exists
+    try:
+        resp = iam.get_role(role_name=role_name)
+        tags = {t["Key"]: t["Value"] for t in resp["Role"].get("Tags", [])}
+        current_revision = int(tags.get(CLEANUP_ROLE_REVISION_TAG_KEY, 0))
```
gmarciani: why not using the revision number specified in PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_EXPECTED_REVISION?

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+    except AWSClientError as e:
+        if e.error_code == "NoSuchEntity":
+            logging.info("Creating global cleanup role %s", role_name)
```
gmarciani: Specify that we are creating it because it does not exists. May sound obvious but since this is a critical resource better to specify why we take some decisions

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+            iam.create_role(
+                RoleName=role_name,
+                Path=IAM_ROLE_PATH,
+                AssumeRolePolicyDocument=json.dumps(assume_doc),
+                Description="AWS ParallelCluster build-image cleanup Lambda execution role",
+            )
+            current_revision = 0
+        else:
+            raise e
+    # (re)attach AWSLambdaVPCAccessExecutionRole
```
gmarciani: Why re-attach?

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +75,173 @@ def _generate_action(action_name, commands):
...
+            role_name, f"arn:{partition}:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
+        )
+
+    if current_revision < PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_EXPECTED_REVISION:
+        # (re)attach AWSLambdaBasicExecutionRole
+        cleanup_role_basic_managed_policy = f"arn:{partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+        iam.attach_role_policy(role_name, cleanup_role_basic_managed_policy)
+
+        # Put (or overwrite) inline policy
+        iam.put_role_policy(
```
gmarciani: [Design] Do we have a way to avoid policy updates? I thought that by using versioned roles we could do that.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: In my approach, create role with tag=0. And the expected revision is 1. So that it can enter the if else block `if current_revision < PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_EXPECTED_REVISION` to update the policy.
See my function description, in this approach, it can prevent tag updated but policy not match the expectation.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Yes, we have, and it has been applied. See `if current_revision < PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_EXPECTED_REVISION:` and the function description for details.

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: See https://github.com/aws/aws-parallelcluster/blob/6d4e24d4968917b91dc57730753fb93a3f99ef48/cli/src/pcluster/templates/imagebuilder_stack.py#L741-L742

The cleanup role needs the `AWSLambdaVPCAccessExecutionRole` when [LambdaFunctionsVpcConfig](https://docs.aws.amazon.com/parallelcluster/latest/ug/DeploymentSettings-build-image-v3.html#DeploymentSettings-build-image-v3-LambdaFunctionsVpcConfig) exists in the config.

CHANGELOG.md
hehe7318: Done. Moved it to **CHANGES**.
`- Introduce default global build-image cleanup IAM versioning roles to prevent build-image CloudFormation stacks from failing to be automatically deleted after images are either successfully built or fail to build.`

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: Done, modified to `ensure_default_build_image_stack_cleanup_role`

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: Reached an agreement yesterday. Done. Changed the logic to create roles with revision in their name.

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: Deleted.

cli/src/pcluster/constants.py
hehe7318: Deleted. No longer applicable.

cli/src/pcluster/constants.py
hehe7318: done

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Done

cli/src/pcluster/imagebuilder_utils.py
hehe7318: It's needed. It's the same string. But due to length restriction so have to split to two lines.
```
                    "Resource": f"arn:{partition}:imagebuilder:*:{account_id}:distribution-configuration/"
                    f"parallelclusterimage-*",
```

cli/src/pcluster/imagebuilder_utils.py
hehe7318: No longer applicable. But the logic didn't change.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Done

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Synced yesterday. It's idempotent

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Changed to have revision in the role name and do not update the role.

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: I don't think it's necessary. It's a common dig logic and the logic to test if it can correctly dig is already include in the unit test `test_enable_cleanup_role_call_and_vpc_flag`.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: No we can't, the delete stack lambda function is named like "ParallelClusterImage-15a09820-6274-11f0-8162-0e903a4ee48d".

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html
I changed it to:
```
                {
                    "Action": "cloudformation:DeleteStack",
                    "Resource": f"arn:{partition}:cloudformation:*:{account_id}:stack/*/*",
                    "Condition": {
                        "ForAnyValue:StringLike": {
                            "cloudformation:ResourceTag/parallelcluster:image_id": "*"
                        }
                    },
                    "Effect": "Allow"
                },
```
Testing if it works.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: Checked Lambda function codes. Done.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
 - Support DCV on Amazon Linux 2023.
 - Upgrade Python runtime used by Lambda functions to python3.12 (from python3.9).
 - Remove `berkshelf`. All cookbooks are local and do not need `berkshelf` dependency management.
+- Introduce default global build-image cleanup IAM versioning roles to prevent build-image CloudFormation stacks from failing to be automatically deleted after images are either successfully built or fail to build.
```
gmarciani: We should capture the following details in the changelog entry:
1. the role is long-living (not bound to the stack lifecycle) and it is meant to persist the stack deletion.
2. the role name: since the role is meant to persist it's important to communicate its name so that users do not think it is a left over when they see it.
3. Since this is a long lasting issue in pcluster, it may be good to link the issue we are solving.

The fact that the role is versioned is an implementation details, more than something related to the user experience, so we can skip that detail.
```
The build-image command now deploys a global role that is used to automatically delete the build-image stack after images either succeed or fail the build. The role is meant to exists even after the stack has been deleted. This is to prevent build-image stack deletion failures, reported in https://github.com/aws/aws-parallelcluster/issues/5914
```

Not: since the changelog entry is long, use multilines to write it.

cli/src/pcluster/api/controllers/image_operations_controller.py
gmarciani: Makes sense, thank you. 
Nitpick: we need this line only if the user is using the default role, so you can move the line into the if branch below.

cli/src/pcluster/api/controllers/image_operations_controller.py
```diff
@@ -105,6 +108,29 @@ def build_image(
...
+    raw_cfg_str = build_image_request_content["imageConfiguration"]
+    cfg_dict = yaml.safe_load(raw_cfg_str) or {}
+    # If CleanupLambdaRole exists in the config, skip ensure_default_build_image_stack_cleanup_role
+    has_custom_cleanup_role = cfg_dict.get("Build", {}).get("Iam", {}).get("CleanupLambdaRole")
+    # If LambdaFunctionsVpcConfig exists in the config, attach the AWS-managed LambdaVPCAccess policy
+    has_lambda_functions_vpc_config = cfg_dict.get("DeploymentSettings", {}).get("LambdaFunctionsVpcConfig")
+
+    if not has_custom_cleanup_role:
+        try:
+            account_id = AWSApi.instance().sts.get_account_id()
```
gmarciani, line 120: if this sts call fails, we would return an error message saying that the principal is missing the permissions to create/update the role. Make sure to add sts permissions to our documentation

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -65,3 +74,175 @@ def _generate_action(action_name, commands):
     """Generate action in imagebuilder components."""
     action = {"name": action_name, "action": "ExecuteBash", "inputs": {"commands": [commands]}}
     return action
+
+
+def get_cleanup_role_name(account_id: str) -> str:
+    """Return the role name including a revision number."""
+    hashed_account_id = generate_string_hash(account_id)
+    return (
+        f"{PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_PREFIX}-{hashed_account_id}"
```
gmarciani: This role is supposed to not be deleted or modified by the user. 
What about adding a DO-NOT-DELETE suffix in the role name?
This is a standard approach, that we also adopted for the S3 bucket, which has a similar lifecycle.

What about `PClusterBuildImageCleanupRole-v{revision}-{hash}-do-not-delete`?
If it is too long, we can use `PClusterImageCleanupRole-v{revision}-{hash}-do-not-delete`.

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: You are right, thanks for catching it. Done.

CHANGELOG.md
hehe7318: Thanks for the advice! Applied.

hehe7318: You mentioned 
>the role name: since the role is meant to persist it's important to communicate its name so that users do not think it is a left over when they see it.

But I don't see it in your suggested line. Should I add it?

cli/src/pcluster/api/controllers/image_operations_controller.py
hehe7318: We are using the same call in other places when doing build-image. So no need to add it.

And ref: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

> GetCallerIdentity
Returns details about the IAM user or role whose credentials are used to call the operation.
Note
No permissions are required to perform this operation. If an administrator attaches a policy to your identity that explicitly denies access to the sts:GetCallerIdentity action, you can still perform this operation. Permissions are not required because the same information is returned when access is denied. To view an example response, see [I Am Not Authorized to Perform: iam:DeleteVirtualMFADevice](https://docs.aws.amazon.com/IAM/latest/UserGuide/troubleshoot_general.html#troubleshoot_general_access-denied-delete-mfa) in the IAM User Guide.

cli/src/pcluster/imagebuilder_utils.py
hehe7318: I think they are different. Delete the S3 bucket have impact to the running cluster because it stores some files required by the cluster. But deleting this user and running build-image again can create a new one without any impact.

And customer can customized the role policy. We added a `bootstrapped` tag for it so once it bootstrapped we will not update/override it again. It makes sense that customer customized it and take the responsibility. If their customization broke it, they can simply delete it and let pcluster create a new one.

gmarciani:
Why are we skipping the bad-url-suffix-check?

cli/src/pcluster/imagebuilder_utils.py
gmarciani: They are not so different: if a user deletes the role after submitting the build-image, all the ongoing build image stacks are going to fail the deletion. The bucket as well gets recreated if the user deletes it, nevertheless we decided to include such suffix to make sure users do not delete it.

A valid counter argument may be: this is true in general for whatever resource created by pcluster, so why adding do-not-delete to this role and not to other things?
And the answer would be: because it is a resource decoupled from the lifecycle of the stacks, the same way the bucket is.

As an alternative approach we can avoid the do-not-delete suffix in the role name and specify it in the description of the role. With buckets we do not have such possibility, but with roles we do. Now I'm thinking that it is even better so that we can communicate the requirement while keeping the role name shorter.

hehe7318:
> Why are we skipping the bad-url-suffix-check?

Because in trust policy we have:

```
"Principal": {"Service": "lambda.amazonaws.com"},
```

cli/src/pcluster/imagebuilder_utils.py
hehe7318: I agree with adding it in the role description. But I mean, after the bucket is re-created, the files can not be restored. So the cluster keeps failing. But if the role is re-created, everything runs correctly.

And the stack can be auto deleted after the image succeed/fail to build. So only the cleanup role is deleted when the stack is still there, the stack will fail to be deleted, and that's customer's responsibility. I don't think customer will do that when the stack hasn't been deleted because it's very clear in the role name that it's being used to cleanup the build-image resources. 

I agree with you to add do not delete in description. But include, do not delete it during the build-image is ongoing, or you will see the stack fail to be deleted.

gmarciani: Agreed: let's put a note in the description, but let's keep it simple: let's simply say that the role must not be deleted without expressing all the scenarios.

cli/tests/pcluster/cli/test_build_image.py
```diff
@@ -117,6 +129,178 @@ def test_nodejs_wrong_version_error(self, mocker, test_datadir):
             self.run_build_image_command(test_datadir)
         assert_that(exc_info.value.data.get("message")).matches("requires Node.js version >=")
 
+    def test_get_cleanup_role_name(self):
+        fake_account = "123456789012"
+        role = get_cleanup_role_name(fake_account)
+        assert_that(role).starts_with("PClusterBuildImageCleanupRole-")
+        assert_that(role).ends_with(f"-revision-{PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION}")
```
gmarciani: Revamping one of my previous comments: why not keeping it shorter using `-v{REVISION}` rather than `-revision-{REVISION}`

cli/tests/pcluster/cli/test_build_image.py
```diff
@@ -117,6 +129,178 @@ def test_nodejs_wrong_version_error(self, mocker, test_datadir):
             self.run_build_image_command(test_datadir)
         assert_that(exc_info.value.data.get("message")).matches("requires Node.js version >=")
 
+    def test_get_cleanup_role_name(self):
+        fake_account = "123456789012"
+        role = get_cleanup_role_name(fake_account)
+        assert_that(role).starts_with("PClusterBuildImageCleanupRole-")
```
gmarciani: [Test Fragility] Why asserting the prefix and suffix of the name rather than the full name? You have an account id so the hashing is deterministic. In that way the test will be more robust!

cli/tests/pcluster/cli/test_build_image.py
hehe7318: Thank for the advice! Done

cli/tests/pcluster/cli/test_build_image.py
hehe7318: Done!

gmarciani:
Thanks for the iteration!

gmarciani (approved):

Issue #6688, opened 2025-03-07
Option to set SharedStorageType: Efs encrypted=True


ghost:
I'm not going to paste all the config that's requested in the Issue template.  If that's a problem, please let me know- it's a lot of work to gather all that config info and it's not relevant to this Issue.

We tried to enable SharedStorageType: Efs  and create a new cluster, the CloudFormation fails (and rolls back).
In this case, I debugged the problem: our privileged user cannot create *unencrypted* Efs volumes, and the code to generate the CF stack looks like this:
https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/cluster_stack.py#L269
```

            internal_efs_storage_shared = SharedEfs(
                mount_dir="/opt/parallelcluster/init_shared", name="internal_pcluster_shared", throughput_mode="elastic"
            )


```

By adding "encrypted=True", I was able to create a cluster with SharedStorageType: Efs

The feature request here is to add some sort of ability to set options for the SharedStorageType: Efs including encryption, as well as the other standard Efs options.  This would probably match the existing support for these options in SharedStorage EFS Settings (https://docs.aws.amazon.com/parallelcluster/latest/ug/SharedStorage-v3.html)

hgreebe:
I have added this feature request to our backlog.

PR #6914, opened 2025-07-14
[Efs encryption] Support HeadNode Efs SharedStorage Encryption

Allenz5:
### Description of changes
* Add a `SharedStorageEfsSettings` section under the `HeadNode` section to include all settings related to Efs shared storage
* Add an `Encrypted` parameter under the `SharedStorageEfsSettings` section to allow users to specify encryption for EFS shared storage
* Add a `SharedStorageEfsSettingsValidator` to verify that `SharedStorageEfsSettings` should only be used when `SharedStorageType` is Efs

### Tests
* Add `test_shared_storage_efs_settings_validator` to test the behavior of `SharedStorageEfsSettingsValidator`
* Executed the updated integration tests `test_internal_efs` to verify EFS encryption

### References
* [Link to impacted open issues.](https://github.com/aws/aws-parallelcluster/issues/6688)
* [Link to impacted open issues.](https://issues.amazon.com/issues/PC-15403)
* [Link to documentation useful to understand the changes.](https://quip-amazon.com/XJEuAdRab6v2/Design-Efs-Encryption)

### Checklist


cli/src/pcluster/config/cluster_config.py
```diff
@@ -851,6 +852,14 @@ def __init__(self, allowed_ips: str = None, **kwargs):
         self.allowed_ips = Resource.init_param(allowed_ips)
 
 
+class SharedStorageEfsSettings(Resource):
+    """Represent the shared storage settings."""
```
himani2411: Please be more descriptive in comments. We need comment to specify the section as there is a `SharedStorage` Section in cluster-config.yaml and this can be confusing during an analysis

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -791,6 +792,15 @@ def make_resource(self, data, **kwargs):
         """Generate resource."""
         return HeadNodeSsh(**data)
 
+class SharedStorageEfsSettingsSchema(BaseSchema):
+    """Represent the schema of SharedStorageSettings."""
```
himani2411: Same as https://github.com/aws/aws-parallelcluster/pull/6914/files#r2205632778

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1336,6 +1336,23 @@ def _validate(self, head_node_instance_type: str, total_max_compute_nodes: int):
...
 
+class SharedStorageEfsSettingsValidator(Validator):
+    """
+    HeadNode SharedStorageEfsSettings Validator.
+
+    Verify HeadNode SharedStorageEfsSettings can only be used with Efs SharedStorageType.
+    """
+
+    def _validate(self, shared_storage_type: str, shared_storage_efs_settings):
+        if shared_storage_efs_settings and shared_storage_type != "Efs":
```
himani2411: So we show this error even if `Ebs` is used and `HeadNode/SharedStorageEfsSettings/Encrypted` is false?

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -2075,6 +2076,41 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
     assert_failure_messages(actual_failures, expected_message)
 
 
+@pytest.mark.parametrize(
+    "shared_storage_type, shared_storage_efs_settings, expected_message",
+    [
+        (
+                "Efs",
+                {
+                    "encrypted": True
```
himani2411: Can you add the use-case of `encrypted` is False? 
so that we cover https://github.com/aws/aws-parallelcluster/pull/6914/files#r2205669568

cli/src/pcluster/validators/cluster_validators.py
himani2411: Also why are we using `Efs` instead of using the `SharedStorageType` Enum?

himani2411:
Please mention that Integ tests that you will modify is ONGOING

cli/src/pcluster/validators/cluster_validators.py
Allenz5: > So we show this error even if `Ebs` is used and `HeadNode/SharedStorageEfsSettings/Encrypted` is false?

Yes, since `SharedStorageEfsSettings` is only for settings related to EFS.

cli/src/pcluster/config/cluster_config.py
Allenz5: Fixed.

cli/tests/pcluster/validators/test_cluster_validators.py
Allenz5: Added test cases of `encrypted` is False.

cli/src/pcluster/validators/cluster_validators.py
Allenz5: > Also why are we using `Efs` instead of using the `SharedStorageType` Enum?

Fixed.

cli/src/pcluster/schemas/cluster_schema.py
Allenz5: Fixed.

himani2411:
LGTM! As discussed will wait for the Integ tests changes before we merge the PR

cli/src/pcluster/templates/awsbatch_builder.py
```diff
@@ -22,7 +22,8 @@
 from aws_cdk.aws_ec2 import CfnSecurityGroup
 from aws_cdk.core import CfnOutput, CfnResource, Construct, Fn, Stack
 
-from pcluster.config.cluster_config import AwsBatchClusterConfig, CapacityType, SharedStorageType
+from pcluster.config.cluster_config import AwsBatchClusterConfig, CapacityType
+from pcluster.config.common import SharedStorageType
```
gmarciani, line 26: Here and in other part of the PR.
This change is motivated by a refactoring, unrelated to the main goal of this PR.
A best practice is to keep the PR minimal and avoid refactoring on components that are out of scope for the main goal. The refactoring you're suggesting here is correct, but I suggest to do it in a follow up PR.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -335,6 +331,19 @@ def _add_resources(self):
                 head_node_alarms=self.head_node_alarms,
             )
 
+    def _add_internal_efs_shared_storage(self):
+        if self.config.head_node.shared_storage_efs_settings:
+            encrypted = self.config.head_node.shared_storage_efs_settings.encrypted
+        else:
+            encrypted = False
```
gmarciani: [Single Responsibility / Maintainability] The resource `SharedStorageEfsSettings` should be the only responsible to determine the default value of the encryption. To this aim, you should define the default value into a constant `SharedStorageEfsSettings .DEFAULT_ENCRYPTION` and reference the constant here. In this way we do not risk to scatter default values in different places. This prevent scattering responsibilities in different places, makes the code easier to maintain as it avoids code duplications and risk of misalignments.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -335,6 +331,19 @@ def _add_resources(self):
...
+        internal_efs_storage_shared = SharedEfs(
+            mount_dir="/opt/parallelcluster/init_shared",
```
gmarciani, line 340: Why `init_shared` and not, for example, `shared`? I know this was already in the code, but do we have an understanding about it?

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1336,6 +1336,24 @@ def _validate(self, head_node_instance_type: str, total_max_compute_nodes: int):
...
+    """
+    HeadNode SharedStorageEfsSettings Validator.
+
+    Verify HeadNode SharedStorageEfsSettings can only be used with Efs SharedStorageType.
+    """
+
+    def _validate(self, shared_storage_type: str, shared_storage_efs_settings):
+        if shared_storage_efs_settings and shared_storage_type.lower() != SharedStorageType.EFS.value:
+            self._add_failure(
+                "SharedStorageEfsSettings is specified "
```
gmarciani, line 1349: The error message can be simplified to `SharedStorageEfsSettings can only be used when SharedStorageType is {SharedStorageType.EFS.value.capitalize()}.` 

I suggest to use the capitalize because, even if we are able to handle both `Efs` and `efs` as values, in our public doc we always use capitalized strings.

cli/tests/pcluster/validators/test_all_validators.py
```diff
@@ -392,6 +395,10 @@ def test_slurm_validators_are_called_with_correct_argument(test_datadir, mocker)
         ],
         any_order=True,
     )
+    shared_storage_efs_settings_validator.assert_has_calls(
+        [call(shared_storage_type="Ebs", shared_storage_efs_settings=None)],
```
gmarciani, line 399: [Test] This unit test would make mnore sense if we use a configuration that inject both shared_storage_type and shared_storage_efs_settings into the validators, otherwise you could have the doubt that there exists cases where shared_storage_efs_settings is not injected.

My suggestion is to cover two scenarios:
1. default case (the one you did): `shared_storage_type=Ebs` and `shared_storage_efs_settings` is not set
2. case where shared_storage_efs_settings is injected: `shared_storage_type=Efs` and  `shared_storage_efs_settings` is set.

cli/tests/pcluster/validators/test_cluster_validators.py
```diff
@@ -2075,6 +2076,50 @@ def test_mixed_security_group_overwrite_validator(head_node_security_groups, que
     assert_failure_messages(actual_failures, expected_message)
 
 
+@pytest.mark.parametrize(
```
gmarciani, line 2079: [Test] It's a best practice to define test cases with a self-explanatory id. This improves readability and troubleshooting.
See example: https://github.com/aws/aws-parallelcluster/blob/4f4c670c1ba7071e7a4543fe3523f940ade1da35/cli/tests/pcluster/api/controllers/test_cluster_operations_controller.py#L138-L146

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -335,6 +331,19 @@ def _add_resources(self):
...
+            name="internal_pcluster_shared",
+            throughput_mode="elastic",
+            encrypted=encrypted,
```
gmarciani, line 343: [Test] We should capture this property in a unit test. the unit test should verify that when encrypted is specified in the cluster config, then the cluster template has the EFS resource with that property set to the expected value.

See example: https://github.com/aws/aws-parallelcluster/blob/bf22eba84c3752ddeb52f6305904c92f0ae942cb/cli/tests/pcluster/templates/test_cluster_stack.py#L217-L249

cli/src/pcluster/validators/cluster_validators.py
Allenz5: We only support 'Efs' and 'Ebs', not 'EFS', 'EBS', 'ebs', or 'efs', so I should hardcode 'Efs' here.

cli/tests/pcluster/validators/test_cluster_validators.py
Allenz5: Added id for unit tests.

cli/src/pcluster/templates/cluster_stack.py
himani2411: I vaguely remember the reason for keeping init_shared is something related to how we take a back up the directories we move and to keep it separate from the other folders

cli/src/pcluster/templates/cluster_stack.py
Allenz5: I'm not supposed to set encrypted value in cluster_stack.py. Set to None instead.

cli/src/pcluster/templates/cluster_stack.py
Allenz5: I'm not sure about the reasoning.

cli/src/pcluster/templates/awsbatch_builder.py
Allenz5: The motivation of this change is to prevent circular import of SharedStorageType in cluster_validators.py.

cli/tests/pcluster/validators/test_all_validators.py
Allenz5: Covered these two scenarios in test_slurm_all_validators_are_called.

cli/src/pcluster/templates/cluster_stack.py
Allenz5: Added test_add_efs_shared_storage to capture this behavior

cli/tests/pcluster/templates/test_cluster_stack/test_efs_shared_storage_encryption/config-encrypted.yaml
```diff
@@ -0,0 +1,25 @@
...
+  Os: alinux2
+HeadNode:
+  InstanceType: t3.micro
+  Ssh:
+    KeyName: ec2-key-name
+  Networking:
+    SubnetId: subnet-12345678
+  SharedStorageType: Efs
+  SharedStorageEfsSettings:
+    Encrypted: true
```
gmarciani: [Non blocking comment] [Knowledge Sharing] You can also inject variables into the config from the test to avoid duplication of config files and make it more explicit what you're chagning in it. In this case you would inject the encrypted variable.

tests/integration-tests/tests/storage/test_internal_efs.py
```diff
@@ -17,22 +17,37 @@
...
     test_directory_correctly_shared_between_ln_and_hn,
     test_efs_correctly_mounted,
+    test_efs_correctly_encrypted,
     verify_directory_correctly_shared,
 )
 
-
+@pytest.mark.parametrize(
+    "encrypted",
+    [True, False],
```
gmarciani: Let's only test the encrypted scenario. Not wortit to have a dedicated test for the unencrypted

cli/src/pcluster/validators/cluster_validators.py
```diff
@@ -1336,6 +1336,24 @@ def _validate(self, head_node_instance_type: str, total_max_compute_nodes: int):
...
+                f"but the SharedStorageType is set to {shared_storage_type}. "
+                "SharedStorageEfsSettings can only be used when SharedStorageType "
+                f"is specified as Efs.",
```
gmarciani: LEt's use the enum rather than rewriting 'Efs'. Remember that you can capitalize the string.
Also, we can make it shorter and say 'SharedStoragerType is Efs' ratherr than 'is specified as'

tests/integration-tests/tests/storage/test_internal_efs.py
Allenz5: Removed false case

cli/src/pcluster/validators/cluster_validators.py
Allenz5: Fixed.

cli/tests/pcluster/templates/test_cluster_stack/test_add_efs_shared_storage/config-encrypted.yaml
```diff
@@ -0,0 +1,25 @@
...
+  Os: alinux2
+HeadNode:
+  InstanceType: t3.micro
+  Ssh:
+    KeyName: ec2-key-name
+  Networking:
+    SubnetId: subnet-12345678
+  SharedStorageType: Efs
+  SharedStorageEfsSettings:
+    Encrypted: true
```
gmarciani, line 11: [Non blocking comment] You're defining 3 different config files where the only difference is the use of encryption for EFS. In this case it's better to define a single config file where the value for the encryption is injected by the test. In this way not only you are reducing the number of files, but it's also immediately clear what are the differences between them. Example: https://github.com/aws/aws-parallelcluster/blob/c84baa2955744e418eccc42c93c339131ef36a06/cli/tests/pcluster/config/test_config_patch/test_single_param_change/pcluster.config.yaml#L18-L25

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
 - Support DCV on Amazon Linux 2023.
 - Upgrade Python runtime used by Lambda functions to python3.12 (from python3.9).
 - Remove `berkshelf`. All cookbooks are local and do not need `berkshelf` dependency management.
+- Add the configuration parameter `HeadNode/SharedStorageEfsSettings/Encrypted` to allow user create encrypted Efs shared storage.
```
gmarciani: [Nitpick] We should specify that we are talking about the internal shared storage, not the standard shared storage (the one in `SharedStorage` section) . I would rephrase to `...to enable encryption on the EFS file system used for the head node internal shared storage.`

gmarciani (approved):

PR #6916, opened 2025-07-14
[Test] Fix tests test_pcluster_configure and test_build_image

gmarciani:
### Description of changes
Fix the following integration tests:
1. **test_pcluster_configure:** the test has always assumed that the free tier only returns t*.micro instance types. However, new instance types have been recently dded to the free tier by AWS, so the test started failing intermittently because we cannot guarantee the t*.micro are returned as first element by EC2. The fix consists in making the test more flexible and succeed as long as the proposed instance type belong to the free tier, despite being a t*micro type.
2. **test_build_image:** the test started failing consistently on Rocky9 because the test was using our first stage AMI as parent image, which is Rocky9.5 that does not gets updates, ultimately leading to build failing when the cookbook attempts top install OS packages due to kernel pinning. With this change we are now validating the build image on the latest Rocky9 image (as of today Rocky9.6) also setting UpdateOsPackages to true.
Notice that Rocky 9.5 reached end os support on June 4th, 2025; so it's even better to start tracking the build-image on the new version, which is Rocky 9.6..

### Tests
* [SUCCESS] **test_pcluster_configure**
* [SUCCESS] **test_build_image** on rocky9

### DIFF
--- a/tests/integration-tests/tests/createami/test_createami.py
+++ b/tests/integration-tests/tests/createami/test_createami.py
@@ -132,7 +132,7 @@ def test_build_image(
         # Test Deep Learning AMIs
         base_ami = retrieve_latest_ami(region, os, ami_type="remarkable", architecture=architecture)
         enable_nvidia = False  # Deep learning AMIs have Nvidia pre-installed
-    elif "rhel" in os or "rocky" in os or "ubuntu" in os:
+    elif "rhel" in os or "ubuntu" in os or os == "rocky8":
         # Test AMIs from first stage build. Because RHEL/Rocky and Ubuntu have specific requirement of kernel versions.
         try:
             base_ami = retrieve_latest_ami(region, os, ami_type="first_stage", architecture=architecture)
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -132,7 +132,7 @@ def test_build_image(
         # Test Deep Learning AMIs
         base_ami = retrieve_latest_ami(region, os, ami_type="remarkable", architecture=architecture)
         enable_nvidia = False  # Deep learning AMIs have Nvidia pre-installed
-    elif "rhel" in os or "rocky" in os or "ubuntu" in os:
+    elif "rhel" in os or "ubuntu" in os or os == "rocky8":
```
hanwen-cluster, line 135: If we don't use first stage AMI for Rocky9, we need to make sure Lustre client is disabled for Rocky9

hanwen-cluster (approved):

PR #6884, opened 2025-06-16
[develop] Address cluster update failure when old capacity reservation has been deleted and changelog update for 3.13.2

hehe7318:
### Description of changes
* Cherry-pick of https://github.com/aws/aws-parallelcluster/pull/6867
* Cherry-pick of https://github.com/aws/aws-parallelcluster/pull/6896
* Update changlog for 3.13.2

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6928, opened 2025-08-01
[CLI][GB200] Add ultraserver instance(p6e-gb200) capacity block support

hehe7318:
### Description of changes
- Add ultraserver instance support with capacity block validation
  - Add CapacityBlockHealthStatusValidator for UltraServer instance check.
    - The validator also collect the capacity blocks that are in valid CAPACITY_BLOCK_INACTIVE_STATES. (for example: state `scheduled`). In this scenario we throw a warning and report the ids and the states.
  - Add describe_capacity_block_status api in ec2.py
  - Add logic to collect all ultraserver instance capacity block ids in a dict, key -> ULTRASERVER_INSTANCE_PREFIX, value -> list of capacity block ids
  - Add ULTRASERVER_INSTANCE_PREFIX_LIST and ULTRASERVER_CAPACITY_BLOCK_ALLOWED_SIZE_DICT constants
  - Implement capacity block size validation for p6e-gb200 instances (9, 18 nodes)
  - Add ultraserver-specific template generation and validation logic
  - Move get_instance_type_and_reservation_type_from_capacity_reservation to ec2.py

- Enhance template and configuration handling
  - Add capacity block support in cluster and queue stack templates
     - Pass ultrasever capacity block size to HeadNode dna.json. For p6e-gb200, "9", "18" or "9, 18".
     - Add has_ultraserver_instance to detect if the queue stack compute resources have ultraserver instance

### Tests
  - Add ultraserver capacity block test suites
  - Add ultraserver cluster and queue stack tests
  - Fix test mocks to return proper CapacityReservationInfo objects
  - Add describe_capacity_reservations method to dummy EC2 client
  - Update existing tests for refactored capacity reservation function


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
 **ENHANCEMENTS**
 - Echo chef-client log when a node fails to bootstrap. This helps with investigating bootstrap failures in cases CloudWatch logs are not available.
+- Add support for p6e-gb200 instances via capacity blocks.
```
gmarciani, line 8: [minor] Move this change to the top of enhancements as it is the major feature in this release

cli/src/pcluster/aws/ec2.py
```diff
@@ -577,3 +577,77 @@ def is_subnet_public(self, subnet_id):
...
+            kwargs["MaxResults"] = max_results
+
+        paginator = self._client.get_paginator("describe_capacity_block_status")
+        page_iterator = paginator.paginate(**kwargs)
+
+        statuses = []
+        for page in page_iterator:
+            statuses.extend(page.get("CapacityBlockStatuses", []))
+
+        return {"CapacityBlockStatuses": statuses}
```
gmarciani: [major] The function accepts multiple capacity blocks ids (great!), but it returns a dictionary that does not allow the caller to understand which status map to which reservation.
The caller expects the function to return something about statuses. It is redundant to return a dictionary where the results are mapped to a key `CapacityBlockStatuses`. It's like having a function `sum(a,b)` that instead of returning a+b returns `{"sumResult": a+b}`.

cli/src/pcluster/aws/ec2.py
```diff
@@ -577,3 +577,77 @@ def is_subnet_public(self, subnet_id):
...
+        reservation_type = None
+
+        if capacity_reservation_id:
+            capacity_reservations = self.describe_capacity_reservations([capacity_reservation_id])
+            if capacity_reservations:
+                reservation = capacity_reservations[0]
+                instance_type = reservation.instance_type()
+                reservation_type = reservation.reservation_type()
+
+        return instance_type, reservation_type
```
gmarciani, line 653: [major] The function is expected to describe the reservation, so specifying the reservation id is mandatory. Why returning `None` when the input is None rather than raising a ValueError or TypeError exception?
Letting None values flowing around should be avoided whenever it is possible because forces the logic to explicitly handle the case.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2409,7 +2415,16 @@ def instance_types(self) -> List[str]:
...
+            capacity_reservation_id = (
+                self.capacity_reservation_target.capacity_reservation_id if self.capacity_reservation_target else None
+            )
+            (
+                instance_type_from_capacity_reservation,
+                _,
+            ) = AWSApi.instance().ec2.get_instance_type_and_reservation_type_from_capacity_reservation(
+                capacity_reservation_id
+            )
+            self._instance_type = Resource.init_param(instance_type_from_capacity_reservation)
```
gmarciani, line 2427: The new functions you added can return None, so instance_type_from_capacity_reservation can be None. What are the consequences of returning a None instance type from this function?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2986,6 +2990,53 @@ def login_nodes_subnet_ids(self):
...
+
+        # Initialize empty lists for each supported ultraserver instance prefix
+        for ultraserver_instance_prefix in ULTRASERVER_INSTANCE_PREFIX_LIST:
+            self.__ultraserver_capacity_block_dict[ultraserver_instance_prefix] = []
+
+        # Iterate through all queues and compute resources to find ultraserver capacity blocks
+        for queue in self.scheduling.queues:
+            for compute_resource in queue.compute_resources:
+                cr_target = compute_resource.capacity_reservation_target or queue.capacity_reservation_target
+                if cr_target and cr_target.capacity_reservation_id:
```
gmarciani, line 3023: what if the config specifies a resource group ARN rather than a reservation id?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1358,6 +1369,12 @@ def _add_head_node(self):
                         else "false"
                     ),
                     "launch_template_id": launch_template_id,
+                    **(
+                        {"p6e_gb200_capacity_block_sizes": cluster_ultraserver_capacity_block_sizes_dict["p6e-gb200"]}
```
gmarciani: We are injecting into dna.json a mapping between instance type prefix and block size? Shouldn't the cookbook know exactly what block size a specific reservation has?
Also, given that we are getting closer to the template size limit, what about making the key shorter: Eg: p6egb200_block_sizes?

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -458,6 +458,69 @@ def get_capacity_reservations_per_az(
     return capacity_reservations_per_az
 
 
+class CapacityBlockHealthStatusValidator(Validator):
```
gmarciani, line 462: [major] When the cluster is created, the capacity block may be inactive and the status information may be unavailable. In this case, the validator should not fail because it is legit for the user to create the cluster when the capacity block is not active yet.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -369,6 +373,69 @@ def generate_launch_template_version_cfn_parameter_hash(queue, compute_resource)
...
+                if size is not None:
+                    cluster_ultraserver_capacity_block_sizes_dict.get(ultraserver_instance_prefix).append(size)
+
+        unique_sizes = sorted(set(cluster_ultraserver_capacity_block_sizes_dict.get(ultraserver_instance_prefix)))
+        if unique_sizes:  # Check if there are any sizes to validate
+            allowed_sizes_list = ULTRASERVER_CAPACITY_BLOCK_ALLOWED_SIZE_DICT.get(ultraserver_instance_prefix)
+            for size in unique_sizes:
+                if size not in allowed_sizes_list:
+                    raise BadRequestException(
+                        f"The capacity block sizes for ultraserver instance {ultraserver_instance_prefix} are "
```
gmarciani: [major] What is the user experience in this case? Will the error surfaced to the users allow them to immediately know what capacity reservation has the unsupported block size?

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -163,11 +167,16 @@ def _add_compute_resource_launch_template(
                 ),
             )
         ]
-
+        cr_target = compute_resource.capacity_reservation_target or queue.capacity_reservation_target
         for network_card in compute_resource.network_cards_list[1:]:
             compute_lt_nw_interfaces.append(
                 ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
-                    device_index=0 if network_card.maximum_network_interfaces() == 1 else 1,
+                    # Set device_index=0 for ultraserver instances or single-NIC network cards
```
hgreebe: device_index should be 0 for only Odd numbered device indexes. This is already covered in the existing logic because the odd device indexes have maximum_network_interfaces() = 1.

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -369,6 +373,69 @@ def generate_launch_template_version_cfn_parameter_hash(queue, compute_resource)
...
+        )
+
+    return cluster_ultraserver_capacity_block_sizes_dict
+
+
+def has_ultraserver_instance(cr_target):
+    """
+    Check if the compute resource uses ultraserver instances with capacity blocks.
+
+    Ultraserver instances (e.g., p6e-gb200) require special network interface configuration
```
hgreebe: This is not quite right. Only the odd indexed cards need the device_index = 0.

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2409,7 +2415,16 @@ def instance_types(self) -> List[str]:
...
         """Instance type of this compute resource."""
         if not self._instance_type:
-            self._instance_type = Resource.init_param(self._instance_type_from_capacity_reservation())
+            capacity_reservation_id = (
+                self.capacity_reservation_target.capacity_reservation_id if self.capacity_reservation_target else None
+            )
+            (
+                instance_type_from_capacity_reservation,
+                _,
+            ) = AWSApi.instance().ec2.get_instance_type_and_reservation_type_from_capacity_reservation(
```
hgreebe, line 2424: What are the benefits of creating an entirely new function rather than just using/building upon the existing `_instance_type_from_capacity_reservation()` function. It seems like we are repeating pre-existing logic.

CHANGELOG.md
hehe7318: Done

cli/src/pcluster/aws/ec2.py
hehe7318: The boto3 `describe_capacity_block_status` api call returns a dict: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_capacity_block_status.html#

Do we want to change the default behavior of this api call?

hehe7318: I got your point. The default behavior returns dict because it has key `NextToken`. But We have already handled it by paginator. Okay, I agree, I will let it directly return a list.

cli/src/pcluster/aws/ec2.py
hehe7318: Because before this PR, we have https://github.com/aws/aws-parallelcluster/blob/fe6a186e9957ba7e85f83c1774225d83326c93a4/cli/src/pcluster/config/cluster_config.py#L2407-L2413 and https://github.com/aws/aws-parallelcluster/blob/fe6a186e9957ba7e85f83c1774225d83326c93a4/cli/src/pcluster/config/cluster_config.py#L2456-L2466.

And now it's also used in the same property instance_type. I am not sure if we raise a TypeError exception here is correct. What do you think?

cli/src/pcluster/config/cluster_config.py
hehe7318: It comes from this PR: https://github.com/aws/aws-parallelcluster/pull/6884.

If returning a None instance type here, it means InstanceType is None, also CapacityReservationId is None. But we can't say the config is invalid. Because we still have InstanceTypes section and CapacityReservationResourceGroupArn.

cli/src/pcluster/aws/ec2.py
hehe7318: Please also see https://github.com/aws/aws-parallelcluster/pull/6928#discussion_r2251818862

cli/src/pcluster/config/cluster_config.py
hehe7318: We only accept capacity block for ultraserver instance. ResourceGroupArn is not considered here.

cli/src/pcluster/templates/cluster_stack.py
hehe7318: >We are injecting into dna.json a mapping between instance type prefix and block size? Shouldn't the cookbook know exactly what block size a specific reservation has?

This is a valid point. Yes, we are injecting.
Regarding g6e-gb200 instances specifically, the block size values are constrained to [9, 18]. According to Himani, these values can be combined. So, when performing the injection, we have three possible approaches:
- Single value "9"
- Single value "18"
- Combined value "9, 18"

hehe7318: >what about making the key shorter: Eg: p6egb200_block_sizes

Okay, done.

cli/src/pcluster/validators/ec2_validators.py
hehe7318: I had a talk with Himani and we reached an agreement on this. According to the design doc, this validator is needed. I think for ultraserver instance, it's not like general instance types. Can we keep it for now and have a talk with Himani to reach a final decision?

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: First, to clarify, when making capacity block reservation for ultraserver instance, you can only choose the block size from a fixed block size values list, for example, p6e-gb200 allowed block sizes are 9 and 18.

So usually this exception will not be raised. I added it just in case in the future something changes.

But if it's raised, it will tell customer your current capacity block sizes for {ultraserver_instance_prefix}(for example p6e-gb200) are like [9, 15, 18]. The allowed list is [9, 18]. So they can check their capacity block and find why there's a 15.

cli/src/pcluster/config/cluster_config.py
hehe7318: 1. To support ultraserver instance I need to also get the `reservation_type` to check if it's "capacity-block". 
2. I need this function in other places to handle the truth that the `capacity_reservation_target` section can under both queue level and compute resource level. 
3. Also I need this function when building stack, if keep it here, there's a circular dependencies issue.

cli/src/pcluster/templates/queues_stack.py
hehe7318: Thank you! Got it, it's a bit different from Himani's design, did you two reach an agreement?

cli/src/pcluster/config/cluster_config.py
gmarciani: is this a limitation captured in our code, i.e. we fail if we specify a capacity block and CapacityReservationResourceGroupArn  does the validation fail? At least in documentation it does not seems so: https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html#yaml-Scheduling-SlurmQueues-CapacityReservationTarget

cli/src/pcluster/templates/cdk_builder_utils.py
gmarciani: what if the user has multiple CB for the same instance type family and only one of those is misconfigured? In that case the user must go through every reservation to spot the problem. The best user experience would be to return an error message like `The capacity block {reservation_id} has invalid block size {detected_size}. Allowed values are {allowed values}.`

Is there any blocker to implement this in this PR?

cli/src/pcluster/validators/ec2_validators.py
gmarciani: I'm ok with returning a warning because it does not block the creation. However: we should return the warning the message should be more user friendly. Instead of `DescribeCapacityBlockStatus returned no entries for the provided Capacity Blocks; unable to verify health.` I'd say `Could not verify health status for capacity block {reservation_id}. Please check the capacity reservation.`.

cli/src/pcluster/templates/queues_stack.py
gmarciani: > device_index should be 0 for only Odd numbered device indexes. 

This is my understanding as well

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Okay, got it, implementing.

cli/src/pcluster/config/cluster_config.py
hehe7318: You mean specify CapacityBlock and and CapacityReservationResourceGroupArn at the same time? 
I think they can not be specified at the same time because CapacityReservationResourceGroupArn needs Instances section. https://docs.aws.amazon.com/parallelcluster/latest/ug/launch-instances-odcr-v3.html
```
      ComputeResources:
        - ...
          Instances:
            - InstanceType: instance
```
But if it's CapacityBlock, you have to use InstanceType section or leave it empty(optional). https://docs.aws.amazon.com/parallelcluster/latest/ug/launch-instances-capacity-blocks.html
```
   ComputeResources:
   - ...
     InstanceType: String (EC2 Instance type of the CB)
```

But even if it doesn't fail, why it matters in our case? This function is to collect `ultraserver_capacity_block_dict`, ultraserver instance can not in a CapacityReservationResourceGroupArn, right?

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Done

cli/src/pcluster/aws/ec2.py
hehe7318: Done

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Done:
- Update CapacityBlockHealthStatusValidator to support inactive capacity block states
  - Add CAPACITY_BLOCK_INACTIVE_STATES constant for scheduled, payment-pending, assessing, delayed states
  - Change validation logic to check for good interconnect status ('ok') instead of bad states
  - Replace TotalUnavailableCapacity check with TotalInstanceCount vs AvailableInstanceCount comparison
  - Add fallback to describe_capacity_reservations for inactive capacity blocks
  - Issue warnings for inactive capacity blocks instead of errors

cli/src/pcluster/templates/queues_stack.py
hehe7318: Removed.

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Removed

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -369,6 +373,61 @@ def generate_launch_template_version_cfn_parameter_hash(queue, compute_resource)
...
+        allowed_sizes_list = ULTRASERVER_CAPACITY_BLOCK_ALLOWED_SIZE_DICT.get(ultraserver_instance_prefix)
+
+        capacity_reservation_ids = cluster_ultraserver_capacity_block_dict.get(ultraserver_instance_prefix)
+        if capacity_reservation_ids:
+            statuses = AWSApi.instance().ec2.describe_capacity_block_status(capacity_reservation_ids)
+
+            for status in statuses:
+                size = status.get("TotalCapacity")
+                if size is not None:
+                    if size not in allowed_sizes_list:
```
gmarciani, line 399: [minor] why not a single if statement with and condition rather than nested if blocks?

cli/src/pcluster/templates/cdk_builder_utils.py
```diff
@@ -369,6 +373,61 @@ def generate_launch_template_version_cfn_parameter_hash(queue, compute_resource)
...
+                        raise BadRequestException(
```
gmarciani: [major] if you raise the exception here and you have multiple invalid statuses, you will only surface the first one to the user, whereas it could be useful to provide the full list of reservations with invalid status.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1265,6 +1266,16 @@ def _add_head_node(self):
         head_node_launch_template.add_metadata("Comment", "AWS ParallelCluster Head Node")
         # CloudFormation::Init metadata
 
+        # Process ultraserver capacity block information for DNA JSON
+        # This section collects capacity block sizes for ultraserver instances (e.g., p6e-gb200)
+        # and validates that they conform to allowed size configurations for Slurm topology
+        cluster_ultraserver_capacity_block_sizes_dict = {}
+        if self.config.scheduling.scheduler == "slurm":
```
gmarciani: [minor] Use the constants whenever it is possible.
https://github.com/aws/aws-parallelcluster/blob/edf576d22b8b1e1025d84dc9a704ae273b2481e5/cli/src/pcluster/constants.py#L48

cli/src/pcluster/constants.py
```diff
@@ -340,3 +340,11 @@ class Operation(Enum):
 # Tag key & expected revision (increment when policy widens)
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION = 1
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_BOOTSTRAP_TAG_KEY = "parallelcluster:build-image-cleanup-role-bootstrapped"
+
+ULTRASERVER_INSTANCE_PREFIX_LIST = ["p6e-gb200"]
```
gmarciani: [codestyle] This instance type prefix is reused many times in this PR. 
It may make sense to have a constant `P6E_GB200=p6e-gb200` defined here and reused everywhere in the code.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -458,6 +459,105 @@ def get_capacity_reservations_per_az(
...
+                # If describe_capacity_block_status fails, check if it's a scheduled capacity block
+                try:
+                    reservations = AWSApi.instance().ec2.describe_capacity_reservations([cb_id])
+                    if reservations and reservations[0].state().lower() in CAPACITY_BLOCK_INACTIVE_STATES:
+                        inactive_blocks.append(cb_id)
+                    else:
+                        # Not in inactive state, so it's a real error
+                        self._add_failure(
+                            f"Unable to retrieve status for Capacity Block {cb_id}. "
+                            f"Please verify the Capacity Block ID is valid and accessible.",
```
gmarciani, line 531: [major] This case may occur when the user specifies a non existing capacity block. I think we should surface a warning rather than an error because this case occurs if the user specifies a non existing reservation and:
1. in this case we already have a validator checking the existance of the reservation, so no reason to have two failures
2. this failure can block the cluster update. Example: the user tries to remove the reservation from the config via clusert update because the reservation does not exist anymore.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -458,6 +459,105 @@ def get_capacity_reservations_per_az(
...
+                except AWSClientError as e:
+                    self._add_failure(
+                        f"Unable to retrieve information for Capacity Block {cb_id}: {str(e)}",
+                        FailureLevel.ERROR,
+                    )
+
+        # Report inactive capacity blocks as warnings
+        if inactive_blocks:
+            self._add_failure(
+                f"Cannot verify health status for inactive Capacity Blocks: {', '.join(inactive_blocks)}. "
```
gmarciani: [Minor] `Capacity Block` -> `capacity block`

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -458,6 +459,105 @@ def get_capacity_reservations_per_az(
...
+                f"Health status validation will be performed when the capacity blocks become active.",
```
gmarciani: Not necessary to specify that health status would be checked when the reservation is active.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -458,6 +459,105 @@ def get_capacity_reservations_per_az(
...
+                FailureLevel.WARNING,
+            )
+
+        # Report unhealthy capacity blocks as errors
+        if unhealthy_details:
+            self._add_failure(
+                (
+                    "One or more Capacity Blocks are not healthy or have insufficient capacity: "
+                    + "; ".join(unhealthy_details)
+                    + ". Please ensure each Capacity Block reports InterconnectStatus='ok' and "
```
gmarciani: I'd rephrase to capacity block is healthy and all reserved capacity is available. The fact that we use a specific field to determine the healthiness is an internal detail. If you disagree and want to keep it, let's inject the `_GOOD_INTERCONNECT ` in the message at least

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1358,6 +1369,12 @@ def _add_head_node(self):
                         else "false"
                     ),
                     "launch_template_id": launch_template_id,
+                    **(
+                        {"p6egb200_block_sizes": cluster_ultraserver_capacity_block_sizes_dict["p6e-gb200"]}
+                        if "p6e-gb200" in cluster_ultraserver_capacity_block_sizes_dict
+                        and cluster_ultraserver_capacity_block_sizes_dict["p6e-gb200"]
```
himani2411: Can you add unit tests for checking this is generated as expected?
It would also be easier to understand the structure using unit tests

himani2411: IS there a reason why the name is `p6egb200_block_sizes`? why not a generaic name `topology_block_size` or just `block_size`?

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: Because if size not none and size in allowed_sizes_list, we do:
`cluster_ultraserver_capacity_block_sizes_dict.get(ultraserver_instance_prefix).append(size)`

cli/src/pcluster/templates/cdk_builder_utils.py
hehe7318: You are right, done.

cli/src/pcluster/templates/cluster_stack.py
hehe7318: Done

cli/src/pcluster/constants.py
hehe7318: Done

cli/src/pcluster/validators/ec2_validators.py
hehe7318: That doesn't convince me:

>in this case we already have a validator checking the existance of the reservation, so no reason to have two failures

That's not true. We need it. See [link](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-capacity-reservations.html#output) under `State -> (string)`. Think about when customer try to create a cluster with a CB that is in `unsupported` state. `describe_capacity_block_status` failed so we enter this except block. `describe_capacity_reservations` succeeds, and output is `State -> unsupported`. In this case we want to prevent customer from create the cluster because: The Capacity Reservation will not be delivered.

>this failure can block the cluster update. Example: the user tries to remove the reservation from the config via clusert update because the reservation does not exist anymore.

Correct me if I am wrong. It can not, when update, the cli checks the updated config, which means when customer remove the CB from the config, the update will not check the previous one.

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Done

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Done! Weird ^_^. I remembered I removed it.

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Make sense, you convinced me. Done!

cli/src/pcluster/templates/cluster_stack.py
hehe7318: >Can you add unit tests for checking this is generated as expected?
It would also be easier to understand the structure using unit tests

I believe I already added it in test_ultraserver_cluster_stack.py.

>IS there a reason why the name is p6egb200_block_sizes? why not a generaic name topology_block_size or just block_size?

Because in the future if other ultraserver instance types are introduced, don't we want another section in dna.json?

cli/src/pcluster/config/cluster_config.py
gmarciani: The use case we are talking about is: a compute resource with CAPACITY BLOCK, instance type specified and a resourcegroupArn rather than a reservationId. 

I think we should support this scenario. If we do not support it, let's verify that that there exist a validator to prevent it. If it does not exist, the next steps are:
1.If the team agrees on supporting the use case, let's address it in a follow up PR
2. if the team agrees on not supporting it, let;s add a validator to prevent it in a follow up PR.

cli/src/pcluster/templates/cdk_builder_utils.py
gmarciani: misread sorry :-)

cli/src/pcluster/validators/ec2_validators.py
gmarciani: Since this does not impact the update path, I'm ok with the error.

cli/src/pcluster/validators/ec2_validators.py
gmarciani: Same as above. If it does not impact the update, I'm ok with the error

gmarciani (approved):

cli/src/pcluster/config/cluster_config.py
hehe7318: I had a sync with Giacomo. We are not going to support it because: >Capacity Reservation Groups are designed to work with standard Capacity Reservations, not with Capacity Blocks. They allow you to group and manage multiple Capacity Reservations together, but this functionality does not extend to Capacity Blocks.

PR #6930, opened 2025-08-06
Change network interface setup logic to account for gb200

hgreebe:
### Description of changes
* gb200 requires the following setup:
  *  if `EfaEnabled: False`, all even indices are set up with the regular interface, the odd indices are not used
  * if `EfaEnabled: True`, all even indices need an efa interface type and all odd indicies need an efa-only interface type
* This change sets up the network interface configuration for gb200 without altering the exisitng configuration for other instance types
* Adds unit test that tests the network configuration for the launch template

### Tests
* Ran `test_multiple_nics` and `test_efa` integration tests

### References
* Cookbook change: https://github.com/aws/aws-parallelcluster-cookbook/pull/2999
* Public documentation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html

### DIFF
--- a/cli/src/pcluster/templates/queues_stack.py
+++ b/cli/src/pcluster/templates/queues_stack.py
@@ -150,33 +150,7 @@ class QueuesStack(NestedStack):
         instance_profiles,
         is_detailed_monitoring_enabled,
     ):
-        # LT network interfaces
-        compute_lt_nw_interfaces = [
-            ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
-                device_index=0,
-                network_card_index=0,
-                associate_public_ip_address=queue.networking.assign_public_ip,
-                interface_type="efa" if compute_resource.efa and compute_resource.efa.enabled else None,
-                groups=queue_lt_security_groups,
-                subnet_id=(
-                    queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None
-                ),
-            )
-        ]
-
-        for network_card in compute_resource.network_cards_list[1:]:
-            compute_lt_nw_interfaces.append(
-                ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
-                    device_index=0 if network_card.maximum_network_interfaces() == 1 else 1,
-                    network_card_index=network_card.network_card_index(),
-                    associate_public_ip_address=False,
-                    interface_type="efa" if compute_resource.efa and compute_resource.efa.enabled else None,
-                    groups=queue_lt_security_groups,
-                    subnet_id=(
-                        queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None
-                    ),
-                )
-            )
+        compute_lt_nw_interfaces = add_network_interfaces(queue, compute_resource, queue_lt_security_groups)
 
         conditional_template_properties = {}
         if compute_resource.is_ebs_optimized:
@@ -385,3 +359,51 @@ class QueuesStack(NestedStack):
         )
 
         return launch_template
+
+
+def add_network_interfaces(
+    queue,
+    compute_resource,
+    queue_lt_security_groups,
+):
+    """Generate launch template network interfaces list"""
+
+    is_gb200 = compute_resource.instance_types[0] == "p6e-gb200.36xlarge"
+    interface = "efa" if compute_resource.efa and compute_resource.efa.enabled and not is_gb200 else None
+
+    compute_lt_nw_interfaces = [
+        ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
+            device_index=0,
+            network_card_index=0,
+            associate_public_ip_address=queue.networking.assign_public_ip,
+            interface_type=interface,
+            groups=queue_lt_security_groups,
+            subnet_id=(queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None),
+        )
+    ]
+
+    for network_card in compute_resource.network_cards_list[1:]:
+        efa_enabled = True if compute_resource.efa and compute_resource.efa.enabled else False
+        even = network_card.network_card_index() % 2 == 0
+        # if efa is disabled, and we have a gb200 instance we skip configuring odd numbered indexes
+        if is_gb200 and not efa_enabled and not even:
+            continue
+
+        interface = "efa" if compute_resource.efa and compute_resource.efa.enabled else None
+        # if efa is enabled with a gb200 instance, even indexes are configured as efa and the odd as efa-only
+        if is_gb200 and efa_enabled:
+            interface = "efa" if even else "efa-only"
+
+        compute_lt_nw_interfaces.append(
+            ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
+                device_index=0 if network_card.maximum_network_interfaces() == 1 else 1,
+                network_card_index=network_card.network_card_index(),
+                associate_public_ip_address=False,
+                interface_type=interface,
+                groups=queue_lt_security_groups,
+                subnet_id=(
+                    queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None
+                ),
+            )
+        )
+    return compute_lt_nw_interfaces
--- a/cli/src/pcluster/templates/queues_stack.py
+++ b/cli/src/pcluster/templates/queues_stack.py
@@ -13,6 +13,7 @@ from pcluster.constants import (
     DEFAULT_EPHEMERAL_DIR,
     NODE_BOOTSTRAP_TIMEOUT,
     OS_MAPPING,
+    P6E_GB200,
     PCLUSTER_COMPUTE_RESOURCE_NAME_TAG,
     PCLUSTER_QUEUE_NAME_TAG,
     PCLUSTER_S3_ARTIFACTS_DICT,
@@ -366,9 +367,8 @@ def add_network_interfaces(
     compute_resource,
     queue_lt_security_groups,
 ):
-    """Generate launch template network interfaces list"""
-
-    is_gb200 = compute_resource.instance_types[0] == "p6e-gb200.36xlarge"
+    """Generate launch template network interfaces list."""
+    is_gb200 = compute_resource.instance_types[0].split(".")[0] == P6E_GB200
     interface = "efa" if compute_resource.efa and compute_resource.efa.enabled and not is_gb200 else None
 
     compute_lt_nw_interfaces = [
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+
+
+def add_network_interfaces(
+    queue,
+    compute_resource,
+    queue_lt_security_groups,
+):
+    """Generate launch template network interfaces list"""
+
+    is_gb200 = compute_resource.instance_types[0].split(".")[0] == P6E_GB200
```
hehe7318, line 372: ultraserver instance is only supported via capacity block. When capacity block is specified, InstanceTypes is not allowed. https://docs.aws.amazon.com/parallelcluster/latest/ug/launch-instances-capacity-blocks.html.

And InstanceType is optional. I have the logic in aws#6928 to identify if it's urtraserver instance.

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+    interface = "efa" if compute_resource.efa and compute_resource.efa.enabled and not is_gb200 else None
```
gmarciani: [minor] Here and below: this is not an interface, but an interface type. What about renaming it to `interface_type`?

gmarciani:
Please add in the Pr description a reference to official documentation driving this choice.

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+            interface_type=interface,
+            groups=queue_lt_security_groups,
+            subnet_id=(queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None),
+        )
+    ]
+
+    for network_card in compute_resource.network_cards_list[1:]:
+        efa_enabled = True if compute_resource.efa and compute_resource.efa.enabled else False
+        even = network_card.network_card_index() % 2 == 0
+        # if efa is disabled, and we have a gb200 instance we skip configuring odd numbered indexes
```
gmarciani: Please, include in the comment the reason why we skip odd indexes.

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+        if is_gb200 and not efa_enabled and not even:
+            continue
+
+        interface = "efa" if compute_resource.efa and compute_resource.efa.enabled else None
```
gmarciani: [minor] can be `= "efa" if efa_enabled else None"`

cli/tests/pcluster/templates/test_queues_stack.py
```diff
@@ -152,6 +154,77 @@ def test_compute_nodes_dna_json(
...
+    def maximum_network_interfaces(self):
+        return self._max_interfaces
+
+
+@pytest.mark.parametrize(
+    "efa_enabled, instance_type, network_cards_list, expected_interfaces",
+    [
+        (
+            True,
+            "p6e-gb200.36xlarge",
```
gmarciani: Our logic does not check for a specific size, but for whatever GB200 size. In this case it is better in unit test to use as instance_type value `p6e.gb200.WHATEVER`.

This approach has the following advantages:
1. (functional) it tests the actual logic, where the size is ignored
2. (documentation) it uses unit test to document our behaviour, making more explicit that we ignore the size

cli/tests/pcluster/templates/test_queues_stack.py
```diff
@@ -152,6 +154,77 @@ def test_compute_nodes_dna_json(
...
+            [NetworkCard(0), NetworkCard(1), NetworkCard(2, 2), NetworkCard(3), NetworkCard(4, 2)],
+            [
+                {"network_card_index": 0, "interface_type": None, "device_index": 0},
+                {"network_card_index": 2, "interface_type": None, "device_index": 1},
+                {"network_card_index": 4, "interface_type": None, "device_index": 1},
+            ],
+        ),
+        (
+            True,
+            "c6in.32xlarge",
```
gmarciani: Same as above. In this case it would be better to use as instance_Type `NOTp6e-gb200.WHATEVER_SIZE`

cli/tests/pcluster/templates/test_queues_stack.py
```diff
@@ -152,6 +154,77 @@ def test_compute_nodes_dna_json(
...
+            "p6e-gb200.36xlarge",
+            [NetworkCard(0), NetworkCard(1), NetworkCard(2, 2), NetworkCard(3), NetworkCard(4, 2)],
+            [
+                {"network_card_index": 0, "interface_type": None, "device_index": 0},
+                {"network_card_index": 2, "interface_type": None, "device_index": 1},
+                {"network_card_index": 4, "interface_type": None, "device_index": 1},
+            ],
+        ),
+        (
+            True,
```
gmarciani, line 195: We are missing the test followiong test case: `efa_enabled=False` and `instance_Type=NOTp6e-gb200.WHATEVER_SIZE`

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+
+
+def add_network_interfaces(
+    queue,
+    compute_resource,
+    queue_lt_security_groups,
+):
+    """Generate launch template network interfaces list."""
+    is_gb200 = compute_resource.instance_types[0].split(".")[0] == P6E_GB200
+    interface_type = "efa" if compute_resource.efa and compute_resource.efa.enabled and not is_gb200 else None
```
himani2411: I think we should mention that we skip the ODD indices if EFA is disabled for Gb200 instance in a validator as well as the public documentation?

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
         return launch_template
+
+
+def add_network_interfaces(
+    queue,
+    compute_resource,
+    queue_lt_security_groups,
+):
+    """Generate launch template network interfaces list."""
+    is_gb200 = compute_resource.instance_types[0].split(".")[0] == P6E_GB200
```
himani2411, line 371: why split at all if we know the instance type  `compute_resource.instance_types[0].split(".")[0]` ?

himani2411: Ohh ignore this I see its a constant which contains the prefix of the instance

hgreebe: so that we can check the prefix and not have to worry about the size

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+            network_card_index=0,
+            associate_public_ip_address=queue.networking.assign_public_ip,
+            interface_type=interface_type,
+            groups=queue_lt_security_groups,
+            subnet_id=(queue.networking.subnet_ids[0] if isinstance(compute_resource, SlurmComputeResource) else None),
+        )
+    ]
+
+    for network_card in compute_resource.network_cards_list[1:]:
+        efa_enabled = compute_resource.efa and compute_resource.efa.enabled
```
himani2411: If you are already using this as a constant why not use if above at line 372?

himani2411:
Can you also describe the test done for testing Gb200 launch template generation if you have done it?

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -385,3 +360,51 @@ def _add_compute_resource_launch_template(
...
+            continue
+
+        interface_type = "efa" if efa_enabled else None
+        # if efa is enabled with a gb200 instance, even indexes are configured as efa and the odd as efa-only
+        if is_gb200 and efa_enabled:
+            interface_type = "efa" if even else "efa-only"
+
+        compute_lt_nw_interfaces.append(
+            ec2.CfnLaunchTemplate.NetworkInterfaceProperty(
+                device_index=0 if network_card.maximum_network_interfaces() == 1 else 1,
```
himani2411, line 400: Is this true for p4d's too?or gb200?

Maximum Network interfaces of Gb200 is 17 so the device Index would be 1.

hgreebe: `MaximumNetworkInterfaces` for the network card is 1 for odd indexes and 2 for even indexes. The maximum network cards is 17.

This logic also works for all other instance types.

cli/tests/pcluster/templates/test_queues_stack.py
himani2411: It makes more sense to keep the instance type so that we have more confidence on what we are testing rather than keeping a generalized name like `NOTp6e-gb200.WHATEVER_SIZE`

gmarciani: I don't think so.

If you keep a specific value, your test verifies that the logic works on that specific value. If you use general values instead you cover the actual logic. This increases the reliability of the test and its value as a way to document behaviors. 

Example to demonstrate the advantages:
1. (functional) if in your logic you're taking the expected decision only for c6in.32xlarge, the test with the specific value will succeed, even if the logic is incorrect. If you use a generic value that is clearly not used in the logic to take decisions, your test will be able to capture bugs.
2. (documentation) tests are meant to validate the correctness of behaviors, but also to document them. Ideally, we should be able to infer our behavior from the test, rather than looking at the underlying source code. If the test uses a specific value `c6in.32xlarge`, I may think that you're taking a decision around that specific value. If you use `NOTp6e-gb200.WHATEVER_SIZE` it is clear that the described behavior is for whatever instance type that is not p6e-gb200.

cli/src/pcluster/templates/queues_stack.py
gmarciani: In documentation for sure, but I would not include it in a validator as it is a too low level detail for a validator.

gmarciani (dismissed):

cli/tests/pcluster/templates/test_queues_stack.py
himani2411: we are using `NOTp6e-gb200.WHATEVER_SIZE` or `p6e-gb200.WHATEVER_SIZE` which basically indicates that we are using p6e-gb200 for our logic. I dont really care about the naming convention here we could essentially specify the instance type as a comment, but i want the test to cover the different setup of NICs from the major multi-nic instances like hpc6id or p4d. 
If in future, there is an immediate need to check for this specific logic working for these specific instance type I dont want to go and check the InstanceTypeData of these instances and compare it with the test cases here to see that it works, which is why the instance type (or a comment) as the name gives more confidence rather than a generic name.

himani2411 (approved):

PR #6934, opened 2025-08-08
[Test] Add integration tests to validate support for GB200.

gmarciani:
### Description of changes
Add integration tests to validate support for GB200. In particular, it verifies the automated configuration of NVIDIA IMEX.

This test creates a cluster with the necessary custom actions to configure NVIDIA IMEX and verifies the following:
1. On the compute resource supporting IMEX (q1-cr1), the IMEX nodes file is configured by the prolog,
IMEX service is healthy and no errors are reported in IMEX's or prolog's logs.
Also, IMEX gets reconfigured when nodes belonging to the same compute resource get replaced
1. On the compute resource not supporting IMEX (q1-cr2), the IMEX nodes file is not configured by the prolog,
keeping the default values and IMEX is not started.

The test prints in test log the full IMEX status to facilitate troubleshooting.

**Important Notes**
* The test is added to the develop configuration for daily execution.
* The test uses instance type `g4dn` to simulate a `p6e-gb200` instance. This is a reasonable approximation for the test because the focus of the test is on IMEX configuration, which can be executed on `g4dn` as well.
* Will remove the compute node custom action once https://github.com/aws/aws-parallelcluster-cookbook/pull/3001 will be merged. at that point the test will replace the custom action with injecting the cookbook attribute to force the IMEX configuration.
* The test makes use of a compute node prolog, following the approach recommended by NVIDIA for per-Job Deployment. The advantage of this approach is that in this way IMEX can be easily restarted locally by each compute node at the right time. If the same would be done form the head node, an additional mechanism to orchestrate the IMEX reload should have been introduced.

**Limitations**
1. If we want to use p6e-gb200 instance types, we need to manually change the capacity reservation in the test because we do not have a way to automate such reservation today.

### Tests
* [SUCCESS] `test_gb200`

### References
* https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/overview.html
* https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/deployment.html


tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
```diff
@@ -0,0 +1,39 @@
+Image:
+  Os: {{ os }}
+  CustomAmi: ami-023512776531a9248
```
gmarciani: This will be removed after testing

tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
```diff
@@ -0,0 +1,38 @@
...
+    S3Access:
+      - BucketName: {{ bucket_name }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  SlurmQueues:
+  - Name: {{ queue_name }}
+    ComputeResources:
+      - Name: {{ compute_resource_with_imex }}
+        Instances:
+          - InstanceType: {{ instance }}
```
himani2411: Since we are using GPU instance type, I would suggest we add the enablement of GPU heath checks so we are sure that the IMEX prolog is the 1st to run as GB200 will fail GPU heath checks if IMEX is not setup. 
So both the IMEX prolog in silo (i.e disabled GPU heath check) and then an update to check if both prolog together without issues.

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
```diff
@@ -0,0 +1,147 @@
...
+    info "Skipping IMEX configuration because system service ${IMEX_SERVICE} is not enabled"
+    prolog_end
+  fi
+}
+
+function get_node_names() {
+  local _queue_name=$1
+  local _compute_resource_name=$2
+
+  ${SCONTROL_CMD} show nodes --json | \
```
himani2411: we already know that `scontrol show nodes` can slow down if a lot of nodes are being used in the cluster as it gives all the content and then we can filter it, so i would suggest either to use `scontrol show nodes <$SLURMD_NODENAME>`. I think we can get the nodename from Slurm environment itself, instead of filtering out using `jq` from the whole output of the command.

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
```diff
@@ -0,0 +1,147 @@
...
+        ) |
+        .name
+      ] |
+      join(",")'
+}
+
+function get_ips_from_node_names() {
+  local _nodes=$1
+  #TODO Exclude non IP values, e.g. hostnames (q1-st-cr1-2) for compute nodes that are initializing
+  ${SCONTROL_CMD} -ao show node "${_nodes}" | sed 's/^.* NodeAddr=\([^ ]*\).*/\1/'
```
himani2411: [Minor] Do we know which one is better or quicker in getting the IP, is it scontrol or IMDSv? Any throttling issues with IMDSv?

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
```diff
@@ -0,0 +1,147 @@
...
+  COMPUTE_RESOURCE_NAME=$(get_dna_parameter "scheduler_compute_resource_name")
+  LAUNCH_TEMPLATE_ID=$(get_dna_parameter "launch_template_id")
+  CR_NODES=$(get_node_names "${QUEUE_NAME}" "${COMPUTE_RESOURCE_NAME}")
+  IPS_FROM_CR=$(get_ips_from_node_names "${CR_NODES}")
+  IMEX_MAIN_CONFIG="/opt/parallelcluster/shared/nvidia-imex/config_${LAUNCH_TEMPLATE_ID}.cfg"
+  IMEX_NODES_CONFIG="/opt/parallelcluster/shared/nvidia-imex/nodes_config_${LAUNCH_TEMPLATE_ID}.cfg"
+
+  info "Queue Name: ${QUEUE_NAME}"
+  info "CR Name: ${COMPUTE_RESOURCE_NAME}"
+  info "CR Nodes: ${CR_NODES}"
```
himani2411: [Major]Can we add that this Prolog is run by HeadNode, either by adding the NodeType in the name itself or by adding a comment at the start of the file

himani2411:
I think it would be important to mention why we choose the Compute Node Prolog or HeadNode Prolog.

I know the reason by it would be helpful if we know this by reading the description or aleast specify it in a doc if we havent already!

tests/integration-tests/configs/develop.yaml
```diff
@@ -280,6 +280,13 @@ test-suites:
           instances: [{{ common.instance("instance_type_1") }}]
           oss: [{{ OS_X86_6 }}]
           schedulers: [ "slurm" ]
+  gb200:
```
himani2411: Can we keep the main section as UltraServer as thats the fature we are supporting and gb200 is the instance under ultraserver and in future we can add more instance under the same

tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
```diff
@@ -0,0 +1,38 @@
...
+    OnNodeStart:
+      Script: s3://{{ bucket_name }}/{{ head_node_start_script }}
+  Iam:
+    S3Access:
+      - BucketName: {{ bucket_name }}
+Scheduling:
+  Scheduler: {{ scheduler }}
+  SlurmQueues:
+  - Name: {{ queue_name }}
+    ComputeResources:
```
himani2411: [Major] We should also add a change in this file where we actually use a Capacity Block of Gb200 if it exist! As we want to use this code for testing later

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
himani2411: I think dna.json also has `slurm_nodename`  IIRC, because I do see the use of it in compute node cookbook recipes.

himani2411: Another `slurm_compute_nodename` in cookbook. And also we write the nodename in one of the files `/opt/parallelcluster/slurm_nodename`

tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
himani2411: A reason for keeping it if in future I change the GPU heath check I should run this test and be sure that I do not break this prolog

himani2411:
[Minor] Lets also add the NVIDIA deployment Model doc to specify the model we choose to support even though the scripts are not the same

tests/integration-tests/tests/gb200/test_gb200/test_gb200/nvidia-imex-status.job
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#SBATCH --job-name=nvidia-imex-status-job
+#SBATCH --ntasks-per-node=1
+#SBATCH --output=slurm-%j.out
+#SBATCH --error=slurm-%j.err
+
+IMEX_CONFIG_FILE="$1"
+
+srun bash -c "/usr/bin/nvidia-imex-ctl -N -j -c ${IMEX_CONFIG_FILE} > result_\${SLURM_JOB_ID}_\$(hostname).out 2> result_\${SLURM_JOB_ID}_\$(hostname).err"
```
himani2411: [Major] Why are we giving the `IMEX_CONFIG_FILE` as an option. I think we should not explicitly give this file if we are setting it as part of main configuration file. adding this `-c` option invalidates the end-to-end test as we will not verify the cookbook changes as per the config recipe  https://github.com/aws/aws-parallelcluster-cookbook/pull/2996 and https://github.com/aws/aws-parallelcluster-cookbook/pull/3001

tests/integration-tests/tests/gb200/test_gb200.py
```diff
@@ -0,0 +1,247 @@
...
+        f"compute resource {compute_resource_name}: {launch_template_id}"
+    )
+
+    submit_job_imex_status(rce, launch_template_id, queue_name, max_nodes)
+
+    assert_imex_nodes_config_is_correct(rce, launch_template_id, ["0.0.0.0", "0.0.0.0"])
+
+
+@pytest.mark.usefixtures("region", "os", "instance", "scheduler")
+def test_gb200(pcluster_config_reader, file_reader, clusters_factory, test_datadir, s3_bucket_factory, region):
```
himani2411: A change as per https://github.com/aws/aws-parallelcluster/pull/6934/files#r2264924160

tests/integration-tests/tests/gb200/test_gb200.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+
+    job_id = submit_job_imex_status(rce, launch_template_id, queue_name, max_nodes)
+
+    assert_imex_nodes_config_is_correct(rce, launch_template_id, ips)
+    assert_imex_status_ok(rce, job_id, ips)
+    assert_no_errors_in_logs(cluster, queue_name, compute_resource_name)
+
+
+def assert_imex_not_configured(cluster: Cluster, queue_name: str, compute_resource_name: str, max_nodes: int = 1):
```
himani2411: If we use the ExtraChefAttribute as per https://github.com/aws/aws-parallelcluster-cookbook/pull/3001 it would apply to these nodes too. I think we should add a check in terms of `assert_imex_status_is_not_ok` w.r.t at least that the status if NOT UP or is DOWN. 

If the prolog was extended to run on these node then the status would be DOWN but the `nvidia-imex-ctl` would be running.

tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
gmarciani: Good point. Gpu health check added for the compute resource simulating gb200.

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
gmarciani: Good point. Every compute node has its own node name stored in local file `/etc//parallelcluster/slurm_plugin/slurm_nodename`. However, here every compute node must retrieve the node names for all the other compute nodes belonging to the same compute resource, not only for itself

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
gmarciani: Probably IMDs is much faster. However, \they are meant for different reasons. With scontrol a single compute nodes can retrieve the IPs for all the node within the same compute resource, which is the goal of the prolog. With IMDS the compute ndoe can retrieve it only for itself

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
gmarciani: This prolog is not run by the head node, but by compute nodes. Anyway, added to the script comment

tests/integration-tests/tests/gb200/test_gb200/test_gb200/pcluster.config.yaml
gmarciani: Done (using a reservation placeholder atm). will add the replacement logic soon in this PR

tests/integration-tests/tests/gb200/test_gb200/test_gb200/nvidia-imex-status.job
gmarciani: `IMEX_CONFIG_FILE ` is the main IMEX config file, not the nodes config file.
if we do not specify the main config file the command would fail because the location where we store the main config file is not the default one:

```
/usr/bin/nvidia-imex-ctl -N -j
Failed to read node configuration file.  Unable to proceed.
```

tests/integration-tests/tests/gb200/test_gb200.py
gmarciani: Added

tests/integration-tests/tests/gb200/test_gb200.py
gmarciani: Done

tests/integration-tests/configs/develop.yaml
gmarciani: Done

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
himani2411: As per the discussion on slack, we agree that we should not change this 

* One of the node actually writes this file (write - 1 node)
* as each node dont actually write the `node_config*.cfg` file as we have a lock mechanism 
* But each node will verify that the content of this file is correct and hence they will do a verification (verify -all)

tests/integration-tests/tests/gb200/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
himani2411: Same as https://github.com/aws/aws-parallelcluster/pull/6934#discussion_r2268218633

tests/integration-tests/tests/gb200/test_gb200/test_gb200/nvidia-imex-status.job
himani2411: Then can we use `systemctl restart nvidia-imex` ? 
I dont see why we use this command as the default is what I replace in the cookbook. Will take a look again if we are missing something in cookbook

gmarciani: `systemctl restart nvidia-imex` is to restart nvidia-imex. `/usr/bin/nvidia-imex-ctl` it's not to restart nvidia-imex, is to read its status. I submit a job that reads nvidia-imex status to check what a job levering imex would perceive imex to be. Makes sense?

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
```diff
@@ -0,0 +1,150 @@
...
+
+  return 0 # Updated
+}
+
+function reload_imex() {
+  info "Stopping IMEX"
+  timeout ${IMEX_STOP_TIMEOUT} systemctl stop ${IMEX_SERVICE}
+  pkill -9 ${IMEX_SERVICE}
+
+  #TODO Improvement: rotate server port to prevent race condition
```
himani2411, line 109: [Non-Blocking]When do we plan to do this? Next Phase or another iteration in the coming weeks?

tests/integration-tests/tests/gb200/test_gb200/test_gb200/nvidia-imex-status.job
himani2411: If our configuration is correct we should not see the error `Failed to read node configuration file.  Unable to proceed.` as I have been using `/usr/bin/nvidia-imex-ctl -N ` command without issues when i was testing

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/90-nvidia-imex.prolog.sh
gmarciani: I suggest to address this improvement in a follow up PR once we fully understand the implications.
So far I have not observed any race condition, even re-executing the same integ test on an existing cluster multiple timnes, but it can be because we are using only 2 nodes and not a real cuda application

tests/integration-tests/tests/gb200/test_gb200/test_gb200/nvidia-imex-status.job
gmarciani: If the main configuration is not in the default location (ours is not), then it is expected for nvidia-imex-ctl to fail if you omit the config path. Could it be that you were testing using the default location?

himani2411 (dismissed):

himani2411 (approved):

PR #6944, opened 2025-08-21
[GB200][CLI] Fix ultraserver instance support

hehe7318:
### Description of changes
- Fix describe_capacity_block_status api call, filter cbr-id by cr-id
- Add UltraserverCapacityBlockSizeValidator to validate capacity block sizes against allowed values
- Move validation logic from cdk_builder_utils to dedicated validator in ec2_validators
- Add caching mechanism for describe_capacity_block_status API calls in EC2 client
- Update process_ultraserver_capacity_block_sizes to focus on processing
- Fix test cases

### Tests
* Unit tests passed

### Checklist


cli/src/pcluster/aws/ec2.py
```diff
@@ -599,22 +600,51 @@ def describe_capacity_block_status(
                 - TotalCapacity: Total number of instances in the capacity block
                 - TotalUnavailableCapacity: Number of unavailable instances
         """
-        kwargs = {}
+        # If no specific capacity_block_ids are requested, use cache key based on filters and max_results
```
gmarciani: [Code Style] We can move this comment to the pydoc of the function.

cli/src/pcluster/aws/ec2.py
```diff
@@ -579,7 +580,7 @@ def is_subnet_public(self, subnet_id):
         return False
 
     @AWSExceptionHandler.handle_client_exception
-    def describe_capacity_block_status(
+    def describe_capacity_block_status(  # noqa: C901
```
gmarciani, line 583: [Code Style][Not Blocking] The check C901 alerts when a function is getting too complicated. Why not splitting the logic in two functions instead of bypassing the check?

cli/src/pcluster/aws/ec2.py
```diff
@@ -599,22 +600,51 @@ def describe_capacity_block_status(
...
-
-        paginator = self._client.get_paginator("describe_capacity_block_status")
-        page_iterator = paginator.paginate(**kwargs)
-
-        statuses = []
-        for page in page_iterator:
-            statuses.extend(page.get("CapacityBlockStatuses", []))
-
-        return statuses
+            result = []
```
gmarciani: [Code Style] The outcome of this function is a list of statuses. Both branches of this logic return a list of statuses, so we should define the array of results before the if..else branch.
This improves also the readability because it's immediately clear that even if we have different branches of logic they all contributes to the same type of result.

Examples:

```
function returnArray():
   statuses = []
   if condition:
       statuses.extend(....)
   else:
      statuses.extend(....)
   return statuses
```

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -488,10 +490,13 @@ def _validate(self, capacity_reservation_ids: List[str]):  # noqa: C901
         for cb_id in capacity_reservation_ids:
```
gmarciani: [CodeStyle] We needed this PR because we got confused by the differences between capacity reservation ids and capacity block ids. We should prevent source of confusion around it in our code, even small ones. Example: here we are iterating over a list of CR ids, but the element is named `cb_id`, which makes refers to CB ids. Can you please adapt the names?

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -488,10 +490,13 @@ def _validate(self, capacity_reservation_ids: List[str]):  # noqa: C901
...
             try:
                 # Try to get capacity block status first
-                statuses = AWSApi.instance().ec2.describe_capacity_block_status([cb_id])
+                statuses = AWSApi.instance().ec2.describe_capacity_block_status()
```
gmarciani: [BLOCKING][Logic] Why should we call `describe_capacity_block_status()` withing the for loop?
We used to call `describe_capacity_block_status(cb_id)`, for every `cb_id`. which motivated the loop.
Now we are calling  `describe_capacity_block_status()`. So we are making here redundant calls to the same function. Why?

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -797,3 +802,35 @@ def _validate(self, placement_group, odcr, subnet, instance_types, multi_az_enab
...
+                needed_capacity_block_statuses = get_needed_ultraserver_capacity_block_statuses(
+                    statuses, capacity_reservation_ids
+                )
+
+                for status in needed_capacity_block_statuses:
+                    size = status.get("TotalCapacity")
+                    if size is not None:
+                        if size not in allowed_sizes_list:
+                            invalid_capacity_blocks.append(
+                                f"{status.get('CapacityBlockId')} (size: {size}, allowed: {allowed_sizes_list})"
```
gmarciani, line 846: [KUDOS:UX Best Practice | Responsive to Feedback]  Good user experience here. With this logic we guide the user in better identifying all the specific blocks causing the issue. Thank you

cli/tests/pcluster/validators/test_ec2_validators.py
```diff
@@ -1850,7 +1851,111 @@ def test_capacity_block_health_status_validator(
...
+                    "TotalCapacity": 16,
+                    "CapacityReservationStatuses": [
+                        {
+                            "CapacityReservationId": "cr-456",
+                            "TotalCapacity": 16,
+                        }
+                    ],
+                },
+            ],
+            "The following capacity blocks have invalid block sizes: cbr-456 \\(size: 16, allowed: \\[9, 18\\]\\).",
```
gmarciani, line 1968: [KUDOS:TestBestPractice | Responsive to Feedback] Adding ids to test cases is a very good practice to leverage uni testing as a better way to document behaviors and troubleshoot them in case of failures.  Thank you

cli/src/pcluster/aws/ec2.py
hehe7318: Makes sense. Done

cli/src/pcluster/aws/ec2.py
hehe7318: Because it's a completed api call, it's complicated, yes, but don't like other cases, putting the whole logic in the same function here, I think it increases the readability. I don't want to split it.

cli/src/pcluster/aws/ec2.py
hehe7318: Great suggestion. Done.

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Done

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Thank you for catching, fixed

cli/src/pcluster/validators/ec2_validators.py
hehe7318: Thank you!!

cli/tests/pcluster/validators/test_ec2_validators.py
hehe7318: Thank you!!

gmarciani (approved):

Issue #6811, opened 2025-05-05
(3.7.0 – 3.13.2) Cannot SSH into LoginNodes with a Key that differs from the HeadNode’s


hehe7318:
# Note

`LoginNodes/Pools/Ssh/KeyName` will be deprecated since ParallelCluster version 3.14.0.

# The issue

Users are unable to SSH into LoginNodes when the cluster has LoginNodes enabled and different SSH key pairs are specified for the HeadNode and LoginNodes.

# Root cause

Both HeadNode and LoginNodes write their public keys to `/home/$USER/.ssh/authorized_keys`. But `/home` is exported from the HeadNode and mounted by every LoginNode. So `authorized_keys` from the HeadNode overwrites those configured for LoginNodes and the LoginNode’s SSH daemon rejects authentication attempts that use the LoginNodes key.

# Affected Versions

ParallelCluster 3.7.0 - 3.13.2, all released versions support [LoginNodes](https://docs.aws.amazon.com/parallelcluster/latest/ug/LoginNodes-v3.html) are affected.

# Mitigation

You can find the mitigation of the problem in ParallelCluster Github Wiki: [(3.7.0 – 3.13.2) Cannot SSH into LoginNodes with a Key that differs from the HeadNode’s](https://github.com/aws/aws-parallelcluster/wiki/(3.7.0-%E2%80%93-3.13.2)-Cannot-SSH-into-LoginNodes-with-a-Key-that-differs-from-the-HeadNode%E2%80%99s)

PR #6948, opened 2025-08-25
[develop] Add deprecated feature warning message when LoginNodes SSH Key is explicitly set in the cluster config

hehe7318:
### Description of changes

This PR implements Phase 1 of the LoginNodes SSH Key deprecation plan by adding deprecation warnings for the `LoginNodes/Pools/Ssh/KeyName` configuration parameter, with a fix to ensure warnings are only shown when the SSH key is explicitly configured by users.

### Initial Implementation
- Added `LoginNodesSshKeyNameDeprecatedValidator` to warn about deprecated SSH KeyName
- Added tracking mechanism to distinguish between explicit user configuration and automatic defaults
- Added validator registration in `LoginNodesSsh` class
- Updated CHANGELOG.md with deprecation notice

### Behavior
- Case 1: User explicitly sets SSH KeyName - should show warning
   ```
   LoginNodes:
     Pools:
       - Name: pool1
         ...
         Ssh:
           KeyName: xuanqi-parallelcluster-dev-us-east-1-ED25519
           AllowedIps: x.x.x.x/x (Or without it)
   ```
   Cli returns:
   ```
       {
         "level": "WARNING",
         "type": "LoginNodesSshKeyNameDeprecatedValidator",
         "message": "LoginNodes/Pools/Ssh/KeyName is deprecated since ParallelCluster version 3.14.0. Please remove it from cluster configuration. The SSH key for the login nodes will be the one used by the head node."
       }
   ```
- User doesn't set SSH KeyName - should NOT show warning (SSH key automatically applied from head node)
   ```
   LoginNodes:
     Pools:
       - Name: pool1
         ...
         # No SSH KeyName specified - will use head node key automatically
   ```
- User sets other SSH settings but not KeyName - should NOT show warning (SSH key automatically applied from head node)
   ```
   LoginNodes:
     Pools:
       - Name: pool1
         ...
         Ssh:
           AllowedIps: x.x.x.x/x    # Only AllowedIps, no KeyName
   ```

### Tests

- All existing tests pass
- Manually test: Deprecation warning only appears for explicitly configured SSH keys

### References
- https://github.com/aws/aws-parallelcluster/issues/6811

### Checklist


gmarciani:
Can you please add to the description a sample output showing the warning? 
Also, in the description you identified 3 scenarios where this warning may (or may not) apply. 
I suggest to capture the config snippet fort every case to make it cristalline clear.

Example

1.User explicitly sets KeyName
LoginNodes:
  Pools:
    - Name: pool1
      ...
      Ssh:
        KeyName: xuanqi-parallelcluster-dev-us-east-1-ED25519
        AllowedIps: 0.0.0.0/0 (Or without it)
2.User doesn't set KeyName
LoginNodes:
  Pools:
    - Name: pool1
      ...
      Ssh:
        AllowedIps: 0.0.0.0/0
3.Empty SSH config
LoginNodes:
  Pools:
    - Name: pool1
      ...

cli/src/pcluster/validators/login_nodes_validators.py
```diff
@@ -0,0 +1,29 @@
...
+from pcluster.validators.common import FailureLevel, Validator
+
+LOGGER = logging.getLogger(__name__)
+
+
+class LoginNodesSshKeyNameDeprecatedValidator(Validator):
+    """Validator to warn about deprecated LoginNodes SSH KeyName."""
+
+    def _validate(self, key_name: str = None, key_name_explicitly_set: bool = False):
+        """Validate that LoginNodes SSH KeyName is deprecated."""
```
gmarciani: [Documentation] The validator does not "Validate that something is deprecated", but "Returns a deprecation warning when LoginNodes SSH KeyName is explicitly set"

cli/src/pcluster/validators/login_nodes_validators.py
```diff
@@ -0,0 +1,29 @@
...
+        if key_name and key_name_explicitly_set:
+            self._add_failure(
+                "LoginNodes/Pools/Ssh/KeyName is deprecated. Please remove it from cluster configuration. "
```
gmarciani: [Best Practice] Whenever a functionality gets deprecated, it's a best practice to specify the dates. Ideally, we should specify the date/version when the functionality is marked as deprecated and the date/version when is gonna be removed.

In our case, we do not specify dates, but version, and we cannot commit to a specific version for the retirement. So we must say "... is deprecated since ParallelCluster version 3.14.0".

Interesting article about it https://treblle.com/blog/best-practices-deprecating-api

tests/integration-tests/tests/pcluster_api/test_api/test_login_nodes/pcluster.config.yaml
```diff
@@ -27,5 +27,3 @@ LoginNodes:
       Networking:
         SubnetIds:
           - {{ public_subnet_id }}
-      Ssh:
-        KeyName: {{ key_name }}
```
gmarciani: [Test][Blocking] (Here and above) Why removing the parameter from the tests? The parameter is still exposed and we did not change its behavior. The change I would expect on the tests is to verify that the warning is returned.
We will remove the parameter from tests when the parameter will be removed form the cluster config schema.

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ CHANGELOG
 - Upgrade DCGM to version 4.2.3 (from 3.3.6) for all OSs except AL2.
 - Upgrade Python to 3.12.11 (from 3.12.8) for all OSs except AL2.
 - Upgrade Intel MPI Library to 2021.16.0 (from 2021.13.1).
+- Add deprecation warning for `LoginNodes/Pools/Ssh/KeyName` configuration parameter.
```
gmarciani: [Documentation][Blocking] Deprecations should be more highlighted in the changelog. I suggest to have a dedicated section **Deprecations** at the top of the changelog. The changelog entry should be something like `The configuration parameter [X] has been deprecated. The CLI now returns a warning message when it is used in the cluster configuration.`

Regarding the reference to the issue: I'm generally in favor of adding such references in the changelog as they may be helpful to give more context around specific decisions. However, the linked issue does not capture our decision to deprecate the feature. Let;'s be sure to mention it before releasing the changelog.

hehe7318: Thanks for the suggestion! I totally agree. Done

cli/src/pcluster/validators/login_nodes_validators.py
hehe7318: Done!

cli/src/pcluster/validators/login_nodes_validators.py
hehe7318: Thanks for the sharing! Done.

tests/integration-tests/tests/pcluster_api/test_api/test_login_nodes/pcluster.config.yaml
hehe7318: >Why removing the parameter from the tests?

Because we deprecated the feature, why keeping it? It should be set to the headnode key automatically. 

>The change I would expect on the tests is to verify that the warning is returned.

I added a unit test to test the validator and the warning message in the three cases. I think it's enough. Do you agree?

hehe7318:
> Can you please add to the description a sample output showing the warning? Also, in the description you identified 3 scenarios where this warning may (or may not) apply. I suggest to capture the config snippet fort every case to make it cristalline clear.
> 
> Example
> 
> 1.User explicitly sets KeyName LoginNodes: Pools: - Name: pool1 ... Ssh: KeyName: xuanqi-parallelcluster-dev-us-east-1-ED25519 AllowedIps: 0.0.0.0/0 (Or without it) 2.User doesn't set KeyName LoginNodes: Pools: - Name: pool1 ... Ssh: AllowedIps: 0.0.0.0/0 3.Empty SSH config LoginNodes: Pools: - Name: pool1 ...

Done!

tests/integration-tests/tests/pcluster_api/test_api/test_login_nodes/pcluster.config.yaml
gmarciani: I agree that the warning message must be tested with a unit test.
I disagree on removing the feature from the integ test because integ test must capture the contract we have with our users. This contract says that the feature still exists, even if it is marked as deprecated. We could evaluate case by case the removal of the test covering features marked as deprecated if it so expensive to not justify the coverage, but this is not the case.

gmarciani: Discussed offline. @hehe7318  agreed on keeping the integ tests.
Thanks for addressing this.

gmarciani (approved):

PR #6984, opened 2025-09-12
[integ-tests] Improve NCCL test to get better performance numbers

hanwen-cluster:
### Description of changes
1. Remove hard coding algorithm and proto to let OFI plugin pick the best values. This change aims at testing the best possible performance.
2. Specify NCCL_TESTS_SPLIT_MASK=0x0. This is consistent with other internal testing method. For the meaning of the parameter see https://github.com/NVIDIA/nccl-tests
3. Increase the minimal message size to 1024, and maximum size to 8G. It doesn't seem to be useful to monitor smaller message sizes, and bigger message sizes use the bandwidth fully.
4. Reduce the number of run from 100 to 20 (default). This makes the test running time stay the same after the message size is increased

### Tests
* test_efa has passed

### References

### Checklist


himani2411:
Can you sepcify the use of `NCCL_TESTS_SPLIT_MASK=0x0`?

hanwen-cluster:
Please see https://github.com/NVIDIA/nccl-tests for more details. Basically, it is what operation should be done when reducing the data

himani2411:
So we are not running any Parallel Operations as per https://github.com/NVIDIA/nccl-tests?tab=readme-ov-file#running-multiple-operations-in-parallel because we are only using the value 0x0

himani2411 (approved):

PR #6990, opened 2025-09-16
[B200] Install kernel-modules-extra before kernel pinning.

gmarciani:
### Description of changes
With this change we install the kernel-modules-extra package, which is required to support p6-b200 instance type. In particular, the package is required by the Infiniband packages installed in https://github.com/aws/aws-parallelcluster-cookbook/pull/3022.

This PR requires https://github.com/aws/aws-parallelcluster-cookbook/pull/3022

### Tests
* Manually built custom AMIs for all OSs executing the same steps of this PR + https://github.com/aws/aws-parallelcluster-cookbook/pull/3022 and validated that these steps are the ones required to make P6-B200 work (verified with dcgmi diagnosis)
* ONGOING Build image on all OSes
* PENDING integ tests: gpu_health_check

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -93,7 +93,7 @@ phases:
                   yum clean all
                 }
                 
-                PACKAGE_LIST="${!KERNEL_PACKAGES_PREFIX}-headers-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-devel-$(uname -r)"
+                PACKAGE_LIST="${!KERNEL_PACKAGES_PREFIX}-headers-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-devel-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-modules-extra-$(uname -r)"
                 Repository="BaseOS"
                 [[ ! $OS =~ (rocky8|rhel8) ]] && PACKAGE_LIST+=" ${!KERNEL_PACKAGES_PREFIX}-devel-matched-$(uname -r)" && Repository="AppStream"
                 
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -93,7 +93,7 @@ phases:
                   yum clean all
                 }
                 
-                PACKAGE_LIST="${!KERNEL_PACKAGES_PREFIX}-headers-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-devel-$(uname -r)"
+                PACKAGE_LIST="${!KERNEL_PACKAGES_PREFIX}-headers-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-devel-$(uname -r) ${!KERNEL_PACKAGES_PREFIX}-modules-extra-$(uname -r)"
```
himani2411, line 96: So we keep them pinned forever? Dont we want to unpin them?

hanwen-cluster: Should we pin the `${!KERNEL_PACKAGES_PREFIX}-modules-extra-$(uname -r)`?

gmarciani: @himani2411 we are not pinning the packages in `PACKAGE_LIST`. We are installing them, but not pinning. The pinning is executed [here](https://github.com/aws/aws-parallelcluster/blob/3687a3789e0a1abcd3de459e43c6f86ce54d1911/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml#L127)

In fact, I think we should pin it instead and unpin it at the end of the build. Doing...

hanwen-cluster (approved):

himani2411 (approved):

PR #7007, opened 2025-09-18
[B200] Make OSU and MPI Ring tests able to run on p6-b200

gmarciani:
#### OSU
We made test_osu able to be executed on p6-b200.48xlarge.
Made the test write the OSU output in the test workspace to facilitate consultation of results.
We also collected the OSU benchmark results for p6-b200 on all the supported OSs.

When executing OSU on IntelMPI, we figured out that the envars `LD_LIBRARY_PATH ` and `FI_EFA_USE_DEVICE_RDMA ` must be set by the job script because p6-b200 is a multi NICs instance type. The same variables are suggested by EFa in their public documentation https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start-nccl.html#nccl-start-base-enable

#### EFA
We made the MPI ring job time out proportional to the number of slots involved.

test_efa used to fail because the execution of the MPI ring job timed out after 20s.
The 20s timeout was set considering considering p4d.24xlarge instances, where the job takes ~11seconds.
However, the duration of the ring job is expected to be proportional to be number of slots, which in p6-b200.48xlarge are double of p4d.24xlarge.

### Tests
* test_osu on p6-b200.48xlarge both on OpenMPI and IntelMPI for all OSs
* test_efa (MPI RING section) on p6-b200.48xlarge on Ubuntu22.04.


tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -207,6 +244,8 @@ def _test_osu_benchmarks_multiple_bandwidth(
         "hpc6id.32xlarge": 23000,  # Equivalent to a theoretical maximum of a single 184Gbps card
         # 8 100 Gbps NICS -> declared NetworkPerformance 800 Gbps
         "trn1.32xlarge": 80000,  # Equivalent to a theoretical maximum of a single 640Gbps card
+        # 8 200 Gbps NICS -> declared NetworkPerformance 16000 Gbps
+        "p6-b200.48xlarge": 16000,  # Equivalent to a theoretical maximum of a single 640Gbps card
```
hanwen-cluster: Why is this number smaller than others?

gmarciani: Good point. It should be 160000.
8 *200 Gbps NICS -> theoretical max is 1600 Gbps = 200000MBps (80% is 160000MBps).

I took the 80% criteria from the previous instance we tested, i.e. trn1.32xlarge whose theoretical max is 800Gbps = 100000MBps (80% is 80000MBps).

For p4d we used a more conservative percentage of 60%.

hanwen-cluster (approved):

PR #7013, opened 2025-09-19
[GB200] Make IMEX prolog use local IMEx configurations + test fixes

gmarciani:
### Description of changes
In https://github.com/aws/aws-parallelcluster-cookbook/pull/3029 we moved from shared IMEx configurations to local ones. In this PR we adapt the prolog accordingly.
Also, we fixed an assertion made on IMEx logs, which used to check the logs in the head node, but it should check the compute nodes.

### Tests
[ONGOING] test_gb200

### DIFF
--- a/tests/integration-tests/tests/common/assertions.py
+++ b/tests/integration-tests/tests/common/assertions.py
@@ -439,8 +439,7 @@ def _assert_build_image_stack_deleted(stack_name, region, timeout_seconds=600, p
     pytest.fail(f"Timed-out waiting for stack {stack_name} deletion (last status: {last_status})")
 
 
-def assert_regex_in_file(cluster: Cluster, compute_node_ip: str, file_name: str, pattern: str, negate: bool = True):
-    rce = RemoteCommandExecutor(cluster, compute_node_ip)
+def assert_regex_in_file(rce: RemoteCommandExecutor, file_name: str, pattern: str, negate: bool = True):
     file_content = read_remote_file(rce, file_name)
     assertion = assert_that(bool(re.search(pattern, file_content, re.IGNORECASE)))
     assertion.is_false() if negate else assertion.is_fals()
--- a/tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/91_nvidia_imex_prolog.sh
+++ b/tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/91_nvidia_imex_prolog.sh
@@ -91,40 +91,9 @@ function write_file() {
     return 1 # Not Updated
   fi
 
-  # Try to acquire lock with timeout
-  (
-      if ! flock -x -w ${_lock_timeout_seconds} 200; then
-        # If timeout, assume deadlock and try to recover
-        info "Lock timeout after ${_lock_timeout_seconds}s, attempting deadlock recovery"
-        exit 1
-      fi
-      echo "${_content}" > "${_file}"
-  ) 200>"${_lock_file}"
-
-  local _lock_result=$?
-
-  if [[ ${_lock_result} -eq 0 ]]; then
-    return 0 # Updated successfully
-  fi
-
-  # Deadlock recovery: remove stale lock file and retry once
-  error "Potential deadlock detected for ${_file}, attempting recovery"
-  rm -f "${_lock_file}"
-  sleep 1  # Brief pause to avoid race conditions
-
-  (
-      if ! flock -x -w 10 200; then
-        exit 1
-      fi
-      echo "${_content}" > "${_file}"
-  ) 200>"${_lock_file}"
-
-  if [[ $? -eq 0 ]]; then
-    info "Lock acquired after deadlock recovery for ${_file}"
-    return 0 # Updated
-  fi
-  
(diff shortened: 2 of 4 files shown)


tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/91_nvidia_imex_prolog.sh
```diff
@@ -91,40 +91,9 @@ function write_file() {
     return 1 # Not Updated
   fi
 
-  # Try to acquire lock with timeout
```
himani2411, line 94: what if we do keep this as part of the prolog?

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/91_nvidia_imex_prolog.sh
```diff
@@ -171,8 +140,8 @@ function create_default_imex_channel() {
   COMPUTE_RESOURCE_NAME=$(get_compute_resource_name "${QUEUE_NAME}-st-" $SLURMD_NODENAME)
   CR_NODES=$(get_node_names "${QUEUE_NAME}" "${COMPUTE_RESOURCE_NAME}")
   IPS_FROM_CR=$(get_ips_from_node_names "${CR_NODES}")
-  IMEX_MAIN_CONFIG="/opt/parallelcluster/shared/nvidia-imex/config_${QUEUE_NAME}_${COMPUTE_RESOURCE_NAME}.cfg"
-  IMEX_NODES_CONFIG="/opt/parallelcluster/shared/nvidia-imex/nodes_config_${QUEUE_NAME}_${COMPUTE_RESOURCE_NAME}.cfg"
+  IMEX_MAIN_CONFIG="/etc/nvidia-imex/config.cfg"
```
himani2411, line 143: You also need to chnage the nvidia-imex-status.job file which points to using a config specific file.

gmarciani: good catch, done!

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/91_nvidia_imex_prolog.sh
gmarciani: Why?

himani2411: Its a deadlock prevention, even through we added it for shared file. 
Any scenario where more processes access this file and we end up in a deadlock scenario can be prevented if we keep it and we have logs showing that we were in deadlock

tests/integration-tests/tests/common/assertions.py
```diff
@@ -439,8 +439,7 @@ def _assert_build_image_stack_deleted(stack_name, region, timeout_seconds=600, p
     pytest.fail(f"Timed-out waiting for stack {stack_name} deletion (last status: {last_status})")
 
 
-def assert_regex_in_file(cluster: Cluster, compute_node_ip: str, file_name: str, pattern: str, negate: bool = True):
-    rce = RemoteCommandExecutor(cluster, compute_node_ip)
+def assert_regex_in_file(rce: RemoteCommandExecutor, file_name: str, pattern: str, negate: bool = True):
```
himani2411, line 442: why do we chnage this?

himani2411: This is in common. So there will be more places we need to change

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/nvidia-imex-status.job
```diff
@@ -8,6 +8,6 @@
 sleep 45
 QUEUE_NAME=$(cat "/etc/chef/dna.json" | jq -r ".cluster.scheduler_queue_name")
 COMPUTE_RES_NAME=$(cat "/etc/chef/dna.json" | jq -r ".cluster.scheduler_compute_resource_name")
-IMEX_CONFIG_FILE="/opt/parallelcluster/shared/nvidia-imex/config_${QUEUE_NAME}_${COMPUTE_RES_NAME}.cfg"
+IMEX_CONFIG_FILE="/etc/nvidia-imex/config.cfg"
```
himani2411: We can remove this file. We should no longer specify the configuration file if not needed!

himani2411 (dismissed):

tests/integration-tests/tests/common/assertions.py
gmarciani: it is common but used only by the gb200 test. Needed to change it because sometimes we need to check in the head node and some other times in the compute nodes. in this way we configure the rce outside of the function.

tests/integration-tests/tests/ultraserver/test_gb200/test_gb200/nvidia-imex-status.job
gmarciani: I remember it was necessary, but apparently it is not when the default location is used

himani2411 (approved):

PR #6994, opened 2025-09-16
[integ-tests] Run test_efa with p6-b200

hanwen-cluster:
### Description of changes
[integ-tests] Refactor get_ultraserver_capacity_reservation_id to be more generic considering instance type]
[integ-tests] Add p6-b200 to test_efa
See commits descriptions for details

### Tests
* test_efa with p6-b200 has passed on Ubuntu22

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7026, opened 2025-09-23
[Test] Add test runner option --capacity-reservation-id to reuse a specific reservation id for a testing.

gmarciani:
### Description of changes
Add test runner option --capacity-reservation-id to reuse a specific reservation id for a testing.
This is useful when we are retaining a cluster that makes use of capacity reservations.
Adapted OSU to make use of this new option.

### Tests
Executed OSU with the option

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -235,6 +235,10 @@ def pytest_addoption(parser):
         "--api-stack",
         help="Name of CFN stack providing the ParallelCluster API infrastructure.",
     )
+    parser.addoption(
+        "--capacity-reservation-id",
+        help="Use an existing capacity reservation.",
+    )
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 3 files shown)


himani2411:
why not use this functionality https://github.com/aws/aws-parallelcluster/pull/7006 and https://github.com/aws/aws-parallelcluster/pull/6994?

hanwen-cluster (approved):

gmarciani:
> why not use this functionality #7006?

That functionality retrieves the capacity reservation that has available capacity. That does not work in the following scenarios:
1. Rerunning a test with a retained cluster. In this case I want the cluster to use the reservation it is already using. The cluster is already consuming the capacity in the reservation, so that functionality would exclude the reservation. 
2. I have two reservations and the test are run in parallel. In that case they may end up using the same reservation and possibly fail due to ICEs.
3. The test takes ~8 minutes before selecting the reservation and you cannot wait that time

PR #7052, opened 2025-10-07
[integ-tests-framework] Add launch time and performance report generation for integration tests

hanwen-cluster:
### Description of changes
Add two new report generators to analyze integration test metrics from DynamoDB:

- generate_launch_time_report: Queries ParallelCluster-IntegTest-Metadata table to analyze cluster creation time and compute node launch times. Generates statistics grouped by OS and test name with time-windowed aggregation to achieve consistent result with os rotation. This is done every 5 days to avoid querying the whole test database everyday.

- generate_performance_report: Queries ParallelCluster-PerformanceTest-Metadata table to track performance data (OSU, StarCCM) by node count over time. This is done everyday because the database is separate and smaller. The execution time is around 30 seconds.

Both generators output Excel reports using pandas/openpyxl for easy analysis and visualization. Reports are automatically generated when JSON reports are requested via test_runner.py.

### Tests
We are able to generate excel files at the end of integration tests

### Checklist

### DIFF
--- a/tests/integration-tests/reports_generator.py
+++ b/tests/integration-tests/reports_generator.py
@@ -9,15 +9,24 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import ast
 import datetime
 import json
 import os
 import time
+from collections import defaultdict
 from typing import List
 
+import boto3
+import matplotlib.pyplot as plt
+import pandas as pd
 import untangle
 from framework.metrics_publisher import Metric, MetricsPublisher
 from junitparser import JUnitXml
+
+from pcluster.constants import SUPPORTED_OSES
+
+SECONDS_PER_YEAR = 365 * 24 * 60 * 60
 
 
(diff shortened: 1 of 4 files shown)


tests/integration-tests/reports_generator.py
```diff
@@ -143,3 +148,281 @@ def _put_metrics(
...
+            "FilterExpression": filter_expression,
+            "ExpressionAttributeNames": expression_attribute_names,
+            "ExpressionAttributeValues": expression_attribute_values,
+        }
+
+        # Add ExclusiveStartKey if we're not on the first iteration
+        if last_evaluated_key:
+            scan_params["ExclusiveStartKey"] = last_evaluated_key
+
+        response = dynamodb_client.scan(**scan_params)
```
himani2411, line 177: scanning the dynamodb table to get a year's worth of data every day seems like an overkill, especially when we are not analyzing it every day.

I would suggest to run this section of getting the data in XSLX maybe once a week or month?

hanwen-cluster: Great suggestion! I amended the code to get launch time Excel every 5 days. I will still get the performance data Excel everyday. Because the source dynamoDB table for launch time is much larger than the dynamoDB table for performance. For performance data report, it only takes 30 seconds. We can keep it everyday

himani2411 (approved):

PR #7064, opened 2025-10-22
Add kernel version to packages we hold for ubuntu

hgreebe:
### Description of changes
* Add kernel version to packages we hold for ubuntu
* This stops the efa installation from taking ~25 minutes
* If we hold all packages, without specifying the kernel, then all the different kernel version of the packages are held, which makes efa install packages for all the different kernels.

### Tests
* Ran first stage build and check that efa installation took ~3 minutes
* https://jenkins-commercial-dev.parallelcluster.ds9.nice.aws.a2z.com/job/build_first_stage/2065/

```
2025-10-22 15:29:46.825000	Stdout:                 == Installing EFA dependencies ==
2025-10-22 15:29:46.825000	Stdout:
2025-10-22 15:29:46.825000	Stdout:                 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
2025-10-22 15:29:46.825000	Stdout:
2025-10-22 15:29:46.825000	Stdout:                 Reading package lists...
2025-10-22 15:29:46.825000	Stdout:                 Building dependency tree...
2025-10-22 15:29:46.825000	Stdout:                 Reading state information...
2025-10-22 15:29:46.825000	Stdout:                 linux-headers-6.8.0-1040-aws is already the newest version (6.8.0-1040.42~22.04.1).
2025-10-22 15:29:46.825000	Stdout:                 0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.

...

2025-10-22 15:32:48.897000	Stdout:                 EFA installation complete.
2025-10-22 15:32:48.897000	Stdout:                 - Please logout/login to complete the installation.
2025-10-22 15:32:48.897000	Stdout:                 - Libfabric was installed in /opt/amazon/efa
2025-10-22 15:32:48.897000	Stdout:                 - Open MPI 4 was installed in /opt/amazon/openmpi
2025-10-22 15:32:48.897000	Stdout:                 - Open MPI 5 was installed in /opt/amazon/openmpi5
2025-10-22 15:32:48.897000	Stdout:                 ===================================================
```

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -129,7 +129,7 @@ phases:
                 [[ $OS =~ rhel ]] && yum versionlock redhat-release
               else
                 apt-get -y install linux-headers-$(uname -r) linux-modules-extra-$(uname -r)
-                apt-mark hold linux-aws* linux-base* linux-headers* linux-image* linux-modules-extra*
+                apt-mark hold linux-aws-$(uname -r)* linux-base-$(uname -r)* linux-headers-$(uname -r)* linux-image-$(uname -r)* linux-modules-extra-$(uname -r)*
               fi
               echo "Kernel version is $(uname -a)"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+3.15.0
+------
+
+**BUG FIXES**
+- Reduce EFA installation time for Ubuntu by ~20 minutes by only holding kernel packages for the installed kernel.
+
 3.14.0
 ------
 


cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -129,7 +129,7 @@ phases:
                 [[ $OS =~ rhel ]] && yum versionlock redhat-release
               else
                 apt-get -y install linux-headers-$(uname -r) linux-modules-extra-$(uname -r)
-                apt-mark hold linux-aws* linux-base* linux-headers* linux-image* linux-modules-extra*
+                apt-mark hold linux-aws-$(uname -r)* linux-base-$(uname -r)* linux-headers-$(uname -r)* linux-image-$(uname -r)* linux-modules-extra-$(uname -r)*
```
gmarciani, line 132: why keeping the `*` after `$(uname -a)`?

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani, line 127: Why shouldn't we specify the kernel version for RHEL?

himani2411:
What happens for if we update OS on a Pcluster AMI? So we start with pcluster official AMI and then just use `UpdateOsPackages/Enabled: true` would that break the EFA installation or will dkms will handle it? Dose it break current Efa installation?

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
hgreebe: Because some of the packages are named like this: `linux-headers-6.8.0-1040-aws`

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
hgreebe: Because the pinning works slightly differently for RHEL, so EFA does not install all the packages like it does for ubuntu

hgreebe:
> What happens for if we update OS on a Pcluster AMI? So we start with pcluster official AMI and then just use `UpdateOsPackages/Enabled: true` would that break the EFA installation or will dkms will handle it? Dose it break current Efa installation?

If update OS packacges is true then we don't pin all the packages so if does not cause the same issue.

gmarciani:
what about having an entry in the changelog? this is a bugfix that provides lot of value for build-image on Ubuntu

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani, line 132: [minor] We execute `uname -r` in many places (not only here, but across the whole component). 
What about having it into a variable rather than calling the command every time?

gmarciani (approved):

PR #7067, opened 2025-10-25
[IntegTest][develop] Fix Access Denied issue when mutiple AD integ tests run in parallel in the same region

hehe7318:
### Description of changes
Implement cross-worker directory stack sharing for AD integration tests

Add file-based registry with file locking to coordinate directory stack sharing across parallel pytest workers. This prevents premature deletion of shared AD directory stacks when multiple OS tests run concurrently in the same region.

- Add FileLock-based cross-process synchronization
- Implement atomic registry file operations in /var/tmp/.pcluster_tests
- Replace in-memory reference counting with persistent file registry
- Ensure directory stacks are only deleted when all tests complete

Fixes issue where first completing test would delete shared directory stack, causing subsequent tests to fail when accessing secrets.

### Tests
* Integ test passed, running 8 tests in parallel, 2 different directory types(SimpleAD and MicrosoftAD), 4 different OSs.

### Checklist


gmarciani (commented):
What about using the existing SharedFixture mechanism which implkemnents already the locking?

https://github.com/aws/aws-parallelcluster/blob/22a737f44e63ef72a865a1108a163cd67f0d9fe0/tests/integration-tests/framework/fixture_utils.py#L40-L42

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -226,55 +229,121 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
     ).is_true()
 
 
+class DirectoryStackSharedFixture(SharedFixture):
```
gmarciani: [BLOCKING] I think the easiest way to use SharedFixture is by using the decorated `@xdist_session_fixture`

See example: https://github.com/aws/aws-parallelcluster/blob/05eee1539124d84e3a55aeb9e90e83aa7bed337b/tests/integration-tests/conftest_networking.py#L285

Why can't we reuse the same approach we are currently using for shared VPC stacks?

gmarciani (changes requested):

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hehe7318: Thanks for pointing that out, you are right. Done. New test ongoing. This time I did exact the same thing as the vpc_stack. However, the behavior now is different.
Previously, before the latest change, when we need a stack, we create one.
Now, we pre-create stacks in all regions and all AD types, although we only need a few of them. And that's what vpc_stack do.
I am figuring if there's a way to create when we need the stack.

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -226,55 +229,133 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
...
+    def release_all(self):
+        # Release once per key so the last releaser triggers generator cleanup (deletion if managed=True).
+        for shared_fixture, _ in self._local.values():
+            try:
+                shared_fixture.release()
+            except Exception as e:
+                logging.warning("Error releasing shared fixture: %s", e)
+
+
+@xdist_session_fixture(autouse=True)
```
gmarciani: how do you prevent the creation of AD stack in every region if this is a session fixture that is autoused?

hehe7318: We have synced by call.

hehe7318: I will add comments in the code to explain as well

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -226,55 +229,114 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
...
-def directory_factory(request, cfn_stacks_factory, vpc_stack):  # noqa: C901
-    # TODO: use external data file and file locking in order to share directories across processes
-    created_directory_stacks = defaultdict(dict)
+def directory_factory(directory_shared_namespace, vpc_stacks_shared, cfn_stacks_factory, request):
+    """
+    Class-scoped factory that lazily creates per-(region, directory_type) SharedFixtures on first use.
+    For each key we 'acquire()' exactly once per worker, cache the stack name, and 'release()' at teardown.
+    """
+    base_dir = Path(directory_shared_namespace)  # shared path provided by the session fixture
+    local_cache = {}  # key -> (shared_fixture_obj, stack_name)
```
gmarciani: Here we have a dictionary where multiple processes read and writes. To prevent race conditions, we should make the read/write operation atomic. Example if I run multiple parallel integ tests in the same region with the same AD type, how can we ensure that only one AD stack is created?

If we do not address this limitation there is the risk that to address a race condition, we introduce another.

What do you think?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -226,55 +229,114 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
...
+
+        data = shared_fixture.acquire()
+        payload = data.fixture_return_value or {}
+        stack_name = payload.get("name")
+        local_cache[key] = (shared_fixture, stack_name)
+        return stack_name
+
+    yield _factory
+
+    # release once per key so the last releaser triggers generator cleanup when managed=True
```
gmarciani: How do we ensure that the AD stack gets deleted even if a worker crashes?

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hehe7318: Let's have a sync

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hehe7318: It's handled by https://github.com/aws/aws-parallelcluster/blob/bb0943bdb3100069e94c8281713d778c805874ae/tests/integration-tests/framework/fixture_utils.py#L85-L91 and https://github.com/aws/aws-parallelcluster/blob/bb0943bdb3100069e94c8281713d778c805874ae/tests/integration-tests/framework/fixture_utils.py#L151-L160

tests/integration-tests/tests/ad_integration/test_ad_integration.py
hehe7318: Every worker has its own local_cache. This local_cache is not shared cross worker. So every worker goes to create SharedFixture.

After create the SharedFixture, workers call acquire(). Inside the SharedFixture, the situation you mentioned is handled.
See
https://github.com/aws/aws-parallelcluster/blob/bb0943bdb3100069e94c8281713d778c805874ae/tests/integration-tests/framework/fixture_utils.py#L70-L82
and
https://github.com/aws/aws-parallelcluster/blob/bb0943bdb3100069e94c8281713d778c805874ae/tests/integration-tests/framework/fixture_utils.py#L196-L205

tests/integration-tests/tests/ad_integration/test_ad_integration.py
```diff
@@ -226,55 +229,114 @@ def _check_ssm_success(ssm_client, command_id, instance_id):
...
+                logging.info("Deleted directory stack %s in %s", name, region)
+            except Exception as e:
+                logging.warning("Failed deleting directory stack %s: %s", name, e)
+
+
 @pytest.fixture(scope="class")
-def directory_factory(request, cfn_stacks_factory, vpc_stack):  # noqa: C901
-    # TODO: use external data file and file locking in order to share directories across processes
-    created_directory_stacks = defaultdict(dict)
+def directory_factory(directory_shared_namespace, vpc_stacks_shared, cfn_stacks_factory, request):
```
gmarciani, line 282: Proposal to further simplify.
Replace the local cache `{"region:directory_Type" -> (SharedFixture, stack_name)}` with a simple reference to the shared fixture created by the worker. 
From the fixture payload you get the stack name.
After the ``yield` you do not need to iterate on the local_cache, but simply release the shared fixture.

hehe7318: Done. Removed the local cache based on the current assumption that each worker invokes this factory only once.

gmarciani (dismissed):

gmarciani (approved):

PR #6932, opened 2025-08-06
[develop] Upgrade connexion to ~=2.15.1, upgrade Werkzeug to ~=3.1 to address CVE-2024-34069

hehe7318:
### Description of changes
* Upgrade connexion to ~=2.15.1. 
* Upgrade Werkzeug to ~=3.1 to address [CVE-2024-34069](https://nvd.nist.gov/vuln/detail/cve-2024-34069). 
* Upgrade serverless_wsgi.py to the latest version. 
* Changes to encoder.py, flask_app.py and etc. to adapt the version bump.

### Tests
* Changes to tests to adapt the version bump.
* Unit tests all passed.

### References
* Link to impacted open issues.
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7078, opened 2025-11-04
[Dependencies] Constraint flask to version ~=3.1.0 (rather than ~=3.0) to reduce the risk of breaking changes.

gmarciani:
### Description of changes
Constraint flask to version ~=3.1.0 (rather than ~=3.0) to reduce the risk of breaking changes.

According to Flask release notes, there may be breaking changes in minor releases. See release notes for [3.1.0](https://github.com/pallets/flask/releases/tag/3.1.0).

Constraining to 3.1.x is also suggested by Dependabot in https://github.com/aws/aws-parallelcluster/pull/6568 and it is consistent with the approach we applied before the recent upgrade in https://github.com/aws/aws-parallelcluster/pull/6932. In fact, before that change we used to constraint to the minor (flask>=2.2.5,<2.3).

### Tests
This change does not alter the version of Flask tested in https://github.com/aws/aws-parallelcluster/pull/6932 (i.e. 3.1.2), but simply limit future updates within the same minor. No additional test required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ CHANGELOG
 **CHANGES**
 - Add validator that warns against the downsides of disabling in-place updates on compute and login nodes through DevSettings.
 - Upgrade Connexion to ~=2.15.1 (from ~=2.13.0).
-- Upgrade Flask to ~=3.0 (from >=2.2.5,<2.3).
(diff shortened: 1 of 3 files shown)


hehe7318 (approved):
The PR title should be ....(rather than ~=3.0)...

No blocking comment. Approved!

PR #7087, opened 2025-11-06
Add GetFunction and GetPolicy permissions to the build image cleanup role

hgreebe:
### Description of changes
* Add GetFunction and GetPolicy permissions to the build image cleanup role
* AccessDenied Errors were encountered when trying to delete the DeleteFunction Lambda function
* Cloudtrail events were encountering the following error:
```
User: arn:aws:sts::<ACCOUNT_ID>:assumed-role/PClusterBuildImageCleanupRole-*-v1/ParallelClusterImage-* is not authorized to perform: lambda:GetFunction on resource: arn:aws:lambda:us-west-2:<ACCOUNT_ID>:function:ParallelClusterImage-* because no identity-based policy allows the lambda:GetFunction action
```
```
User: arn:aws:sts::<ACCOUNT_ID>:assumed-role/PClusterBuildImageCleanupRole-*-v1/ParallelClusterImage-* is not authorized to perform: lambda:GetPolicy on resource: arn:aws:lambda:us-west-2:<ACCOUNT_ID>:function:ParallelClusterImage-* because no identity-based policy allows the lambda:GetPolicy action
```

### Tests
* Running test_build_image and checked CloudTrail to validate that `GetFunction` and `GetPolicy` actions no longer had AccessDenied errors.

### DIFF
--- a/cli/src/pcluster/constants.py
+++ b/cli/src/pcluster/constants.py
@@ -343,7 +343,7 @@ PCLUSTER_BUCKET_REQUIRED_BOOTSTRAP_FEATURES = ["basic", "export-logs"]
 
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_PREFIX = "PClusterBuildImageCleanupRole"
 # Tag key & expected revision (increment when policy widens)
-PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION = 1
+PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION = 2
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_BOOTSTRAP_TAG_KEY = "parallelcluster:build-image-cleanup-role-bootstrapped"
 
 P6E_GB200 = "p6e-gb200"
--- a/cli/src/pcluster/imagebuilder_utils.py
+++ b/cli/src/pcluster/imagebuilder_utils.py
@@ -140,7 +140,7 @@ def _expected_inline_policy(account_id: str, partition: str):
                 {"Action": "ec2:CreateTags", "Resource": f"arn:{partition}:ec2:*::image/*", "Effect": "Allow"},
                 {"Action": "tag:TagResources", "Resource": "*", "Effect": "Allow"},
                 {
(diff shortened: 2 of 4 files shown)


cli/src/pcluster/constants.py
```diff
@@ -343,7 +343,7 @@ class Operation(Enum):
 
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_PREFIX = "PClusterBuildImageCleanupRole"
 # Tag key & expected revision (increment when policy widens)
-PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION = 1
+PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_REVISION = 2
```
gmarciani, line 346: IIRC we need to increment only when we introduce breaking changes to the policy. This is an additive fix, not a breaking change. So I think we can keep the version to 1.

Can you please double check?
If this is the case, then please also fix the comment, as it could be misleading.

gmarciani: Nope, I was wrong. We need to increase the version at every change, because we do not modify the exisitng role
See code https://github.com/aws/aws-parallelcluster/blob/e2d21fc22b739019628383b8d263c3e12c23c1fa/cli/src/pcluster/imagebuilder_utils.py#L223-L224

cli/src/pcluster/imagebuilder_utils.py
```diff
@@ -140,7 +140,7 @@ def _expected_inline_policy(account_id: str, partition: str):
                 {"Action": "ec2:CreateTags", "Resource": f"arn:{partition}:ec2:*::image/*", "Effect": "Allow"},
                 {"Action": "tag:TagResources", "Resource": "*", "Effect": "Allow"},
                 {
-                    "Action": ["lambda:DeleteFunction", "lambda:RemovePermission"],
+                    "Action": ["lambda:DeleteFunction", "lambda:RemovePermission", "lambda:GetFunction", "lambda:GetPolicy"],
```
gmarciani: [Test] I would expect a unit test to be adapted accordingly. If there is no unit test covering this part, can we add it to cover this change?

gmarciani (dismissed):

gmarciani (approved):

PR #7134, opened 2025-12-02
Add validation to block updating order of tags during cluster update

hgreebe:
### Description of changes
* Add validation to block updating order of tags during cluster update
* Add new UpdatePolicy to show correct messaging if the update fails due to a change in tag order.
* Maintains current default behavior except for the case of order changes of tags
* Our code did not do an ordered comparison when checking for changes during an update, because of this, if the order of tags were changed, this was not picked up as part of the change set.
  * This would later on cause issues because EC2 would see the change in order and create a new launch template for the head node, causing the update to fail.
* This is an example error message that occurs when the order of tags is changed:
```
{
  "message": "Update failure",
  "updateValidationErrors": [
    {
      "parameter": "Tags",
      "requestedValue": "[{'Key': 'test2', 'Value': 'test2'}, {'Key': 'test1', 'Value': 'test1'}]",
      "message": "Update actions are not currently supported for the 'Tags' parameter. The order of tags has changed, but order changes are not supported. Restore the original order of 'Tags' to match the current configuration.",
      "currentValue": "[{'Key': 'test1', 'Value': 'test1'}, {'Key': 'test2', 'Value': 'test2'}]"
    }
  ],
  "changeSet": [
    {
      "parameter": "Scheduling.SlurmQueues[queue-0].ComputeResources[compute-resource1].MaxCount",
      "requestedValue": "3",
      "currentValue": "2"
    },
    {
      "parameter": "Tags",
      "requestedValue": "[{'Key': 'test2', 'Value': 'test2'}, {'Key': 'test1', 'Value': 'test1'}]",
      "currentValue": "[{'Key': 'test1', 'Value': 'test1'}, {'Key': 'test2', 'Value': 'test2'}]"
    }
  ]
```

* Overview of logic:
  * I first check if there is ONLY an order change and then check for other modifications to tags.
  * The reason I do this is because if did a simple list comparison, then I would not get the itemized changeset of each modification to the tags. 
  * I first check that the lists are not equal and then convert them to sets and check if they are equal.
    * If the tag list of current vs requested are not equal but the sets are equal then I know the only change is in order.
    * If the sets are not equal then I know that there is a modification that I want itemized in the changeset.

### Tests
* Ran unit tests
* Did manual tests with changes to order as well as additions and modifications to tags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ CHANGELOG
 - Upgrade jmespath to ~=1.0 (from ~=0.10).
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
 
+**BUG FIXES**
+- Add validation to block updates that change tag order.
+
 3.14.1
 ------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ CHANGELOG
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
 
 **BUG FIXES**
-- Add validation to block updates that change tag order.
+- Add validation to block updates that change tag order. Blocking such change prevents update failures.
 
 3.14.1
 ------
--- a/cli/src/pcluster/config/config_patch.py
+++ b/cli/src/pcluster/config/config_patch.py
@@ -203,7 +203,7 @@ class ConfigPatch:
                         data_key,
                         base_list,
                         target_list,
-                        change_update_policy,
+                        UpdatePolicy.UNSUPPORTED_ORDER_CHANGE,
                         is_list=True,
                     )
                 )
--- a/cli/src/pcluster/config/config_patch.py
+++ b/cli/src/pcluster/config/config_patch.py
@@ -176,19 +176,50 @@ class ConfigPatch:
     def _compare_list(self, base_section, target_section, param_path, data_key, field_obj, change_update_policy):
         """
         Compare list of nested section (e.g. list of queues) by comparing the items with the same update_key.
-
-        If update_key is not set we're considering Name as identifier.
         """
         update_key = field_obj.metadata.get("update_key")
 
+        # Get the lists
+        base_list = base_section.get(data_key, []) if base_section else []
+        target_list = target_section.get(data_key, []) if target_section else []
+
+        # For Tags, check if this is ONLY an order change (same content, different order)
+        if data_key == "Tags" and base_list != target_list:
+            # Check if it's an order-only change
+            try:
+                base_set = {frozenset(item.items()) for item in base_list}
+                target_set = {frozenset(item.items()) for item in target_list}
+                is_order_only_change = (base_set == target_set)
+            except (TypeError, AttributeError):
+                is_order_only_change = False
+
+            # If it's ONLY an order change, block it here and return
+            if is_order_only_change:
+                self.changes.append(
+                    Change(
+                        param_path,
+                        data_key,
+                        base_list,
+                        target_list,
+                        change_update_policy,
+                        is_list=True,
+                    )
+                )
+                return  # Don't process individual items
+            # Otherwise, fall through to item-by-item comparison below
+            # The reason I check if it is an order only change first is because if I just do I order dependent
+            # comparison and fail if there is a difference, then we don't get the item by item comparison in the
+            # change set if the modification was not the order.
+
+
         # Compare items in the list by searching the right item to compare through update_key value
         # First, compare all sections from target vs base config and mark visited base sections.
-        for target_nested_section in target_section.get(data_key, []):
+        for target_nested_section in target_list:  # Changed from target_section.get(data_key, [])
             update_key_value = target_nested_section.get(update_key)
             base_nested_section = next(
                 (
                     nested_section
-                    for nested_section in base_section.get(data_key, [])
+                    for nested_section in base_list  # Changed from base_section.get(data_key, [])
                     if nested_section.get(update_key) == update_key_value
                 ),
                 None,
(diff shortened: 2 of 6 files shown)


cli/src/pcluster/config/update_policy.py
```diff
@@ -682,6 +682,18 @@ def get_pool_name_from_change_paths(change):
     + ". If you need this change, please consider creating a new cluster instead of updating the existing one.",
 )
 
+UpdatePolicy.UNSUPPORTED_ORDER_CHANGE = UpdatePolicy(
+    name="UNSUPPORTED_ORDER_CHANGE",
+    level=1000,  # Same level as UNSUPPORTED since it should also block updates
+    fail_reason=lambda change, patch: (
+        f"Update actions are not currently supported for the '{change.key}' parameter. "
+        f"The order of {change.key.lower()} has changed, but order changes are not supported"
```
gmarciani: [minor] Missing period at the end of the sentence. Also, let's use a consistent case for the messages: here the change key is lowered, but it is not in the previous line. We do not change the case in other update policy messaging.

cli/src/pcluster/config/config_patch.py
```diff
@@ -181,14 +181,46 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
...
+
+        # For Tags, check if this is ONLY an order change (same content, different order)
+        if data_key == "Tags" and base_list != target_list:
+            # Check if it's an order-only change
+            try:
+                base_set = {frozenset(item.items()) for item in base_list}
+                target_set = {frozenset(item.items()) for item in target_list}
+                is_order_only_change = base_set == target_set
+            except (TypeError, AttributeError):
+                is_order_only_change = False
```
gmarciani, line 199: [BestPractice/ExceptionHandling] Here we are catching an exception without logging anything. I suggest to add a warning/info log line to facilitate troubleshooting in case of issues.

cli/src/pcluster/config/config_patch.py
```diff
@@ -181,14 +181,46 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
...
 
+        # Get the lists
+        base_list = base_section.get(data_key, []) if base_section else []
+        target_list = target_section.get(data_key, []) if target_section else []
+
+        # For Tags, check if this is ONLY an order change (same content, different order)
+        if data_key == "Tags" and base_list != target_list:
+            # Check if it's an order-only change
+            try:
+                base_set = {frozenset(item.items()) for item in base_list}
```
gmarciani, line 194: [Risk] What happens if the user specifies the same tag key twice in the cluster config? The list will be converted into a set, eliminating duplicates and the set comparison will not detect changes in the order. This approach would be ok only if we ensure that duplicates keys are not injected into the template

cli/src/pcluster/config/config_patch.py
```diff
@@ -181,14 +181,46 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
         """
         update_key = field_obj.metadata.get("update_key")
 
+        # Get the lists
+        base_list = base_section.get(data_key, []) if base_section else []
+        target_list = target_section.get(data_key, []) if target_section else []
+
+        # For Tags, check if this is ONLY an order change (same content, different order)
+        if data_key == "Tags" and base_list != target_list:
```
gmarciani: [Improvement] Here the logic seems specific to Tags, but the same logic can be applied to whatever parameter where the order is meaningful. If you agree with this statement, then I suggest to define a constant `ORDER_SENSITIVE_PARAMETERS = ["Tags"]` and change `data_key == "Tags"` to `data_key in ORDER_SENSITIVE_PARAMETERS`

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ CHANGELOG
 - Upgrade jmespath to ~=1.0 (from ~=0.10).
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
 
+**BUG FIXES**
+- Add validation to block updates that change tag order.
```
gmarciani: [Documentation] I'd specify why this bug fix is important from the user perspective. Blocking such change prevents update failures.

cli/src/pcluster/config/config_patch.py
hgreebe: You are not allowed to have duplicate keys. If you try to create a cluster with duplicate keys, you get the following error:
```
{
    {
      "level": "ERROR",
      "type": "ConfigSchemaValidator",
      "message": "[('Tags', ['Duplicate tag key (test1) detected. Tags keys should be unique within the Tags section.'])]"
    }
  ],
}

```

cli/src/pcluster/config/update_policy.py
hgreebe: If you look at the fail reason for other update policies, they don't have periods either. If I add a period, there is a double period in the actual message.

gmarciani: Interesting, thanks for pointing this out!

cli/src/pcluster/config/config_patch.py
gmarciani: That's great! Can you please add a comment in this line specifying that we can assume there are no duplicates thanks to that validator?

cli/src/pcluster/config/config_patch.py
```diff
@@ -181,14 +182,47 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
...
+                return  # Don't process individual items
+            # Otherwise, fall through to item-by-item comparison below
+            # The reason I check if it is an order only change first is because if I just do I order dependent
+            # comparison and fail if there is a difference, then we don't get the item by item comparison in the
+            # change set if the modification was not the order.
+
         # Compare items in the list by searching the right item to compare through update_key value
         # First, compare all sections from target vs base config and mark visited base sections.
-        for target_nested_section in target_section.get(data_key, []):
+        for target_nested_section in target_list:  # Changed from target_section.get(data_key, [])
```
gmarciani: Unnecessary AI comment here

gmarciani (approved):

PR #7146, opened 2025-12-09
[Release-3.14.1][Bug] Fix cfn-hup endless loop after rollback to cluster state older than 24 hours

hehe7318:
### Description of changes
When a cluster update fails and triggers a rollback to a state older than 24 hours, cfn-hup enters an endless loop on the head node. This happens because:

1. The rollback restores the launch template metadata to reference an expired wait condition handle (wait conditions expire after 24h)
2. cfn-signal fails to signal the expired handle and returns non-zero
3. cfn-hup sees the non-zero exit code and does not update its local metadata cache (metadata_db.json)
4. On the next polling interval, cfn-hup detects the same "change" and re-triggers the update recipe, creating an infinite loop

This fix appends `|| exit 0` to the update command, ensuring cfn-hup always updates its metadata cache regardless of whether cfn-signal succeeds or fails. This prevents the endless loop while still allowing CloudFormation to handle timeouts appropriately.

### Tests
* After 24 hours, manually test passed. 

### Checklist

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -1497,7 +1497,7 @@ class ClusterCdkStack:
                 "commands": {
                     "chef": {
                         # This command runs the update recipe and signals CloudFormation with the result.
-                        # The trailing "; exit 0" ensures cfn-hup always updates its local metadata cache
+                        # The trailing "|| exit 0" ensures cfn-hup always updates its local metadata cache
                         # (metadata_db.json) regardless of whether cfn-signal succeeds or fails.
                         # Without this, if cfn-signal fails (e.g., due to an expired wait condition handle
                         # after a rollback to a state older than 24h), cfn-hup would not update its cache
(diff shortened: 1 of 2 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1496,6 +1496,12 @@ def _add_head_node(self):
...
                 "commands": {
                     "chef": {
+                        # This command runs the update recipe and signals CloudFormation with the result.
+                        # The trailing "|| exit 0" ensures cfn-hup always updates its local metadata cache
+                        # (metadata_db.json) regardless of whether cfn-signal succeeds or fails.
+                        # Without this, if cfn-signal fails (e.g., due to an expired wait condition handle
+                        # after a rollback to a state older than 24h), cfn-hup would not update its cache
+                        # and would enter an endless loop, re-triggering the update recipe every minute.
                         "command": (
                             ". /etc/parallelcluster/pcluster_cookbook_environment.sh; "
```
gmarciani, line 1506: [FUTURE IMPROVEMENTS] In future we should consider to move this block of code to a script in the head node and simply use this metadata command to pass the wait condition handle as an argument. This would not only simplify the testing but also the mitigation on a running environment.

This is out of scope for the current PR and we do not want to change it in 3.14.1.
However, what do you think about it?

gmarciani:
LGTM, only need to address the code check failure

cli/src/pcluster/templates/cluster_stack.py
hehe7318: I think it makes a lot of sense. Better readable and maintenance. Also easier to test. Should I add a TODO in the PR?

gmarciani: Better to track it into our backlog of internal features

gmarciani (approved):

PR #7150, opened 2025-12-15
[Release-3.14.1][Test] Add integration test for the fixes of issues caused by cluster update and rollback failure

hehe7318:
### Description of changes
Add integration test to verify the following fixes work correctly:
- [F1] clustermgtd remains running after both update and rollback fail
           (expected when failure occurs after slurm reconfiguration, which is the safe section)
- [F2] cfn-hup does not enter an endless loop after rollback to a state older than 24h
- [F3] dna.json files are cleaned up after update and rollback failure

Test scenario:
1. Create cluster with 3 static compute nodes
2. Inject cfn-signal failure on head node (simulating expired wait condition)
3. Disable cfn-hup on CN1 before update (causes update to fail)
4. Trigger cluster update (add new queue)
5. Wait for CN2 to apply update, then disable its cfn-hup
6. Update fails (CN1 didn't update), rollback fails (CN2 won't rollback)
7. Verify: clustermgtd running, dna.json cleaned up, CN3 has correct config version, metadata_db.json updated, no cfn-hup endless loop

### Tests
* Running and debugging

### References

### Checklist


tests/integration-tests/configs/develop.yaml
```diff
@@ -825,6 +825,12 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: [{{ OS_X86_3 }}]
           schedulers: ["slurm"]
+    test_update_rollback_fixes.py::test_update_rollback_fixes:
```
gmarciani: What about naming the test `test_update_rollback_failure`?

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
```diff
@@ -0,0 +1,57 @@
...
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Iam:
+    AdditionalIamPolicies:
+      - Policy: arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
+Scheduling:
+  Scheduler: slurm
+  SlurmSettings:
+    ScaledownIdletime: 60
```
gmarciani: Here and in other parts of the config.
Is this meaningful for the test? If it is, why? If not, let's remove it and use the default value.

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
```diff
@@ -0,0 +1,57 @@
...
+  SlurmQueues:
+    - Name: queue1
+      ComputeSettings:
+        LocalStorage:
+          RootVolume:
+            Size: 45
```
gmarciani: Here and in other parts of the config. 
Is this meaningful for the test? If it is, why? If not, let's remove it and use the default value.

tests/integration-tests/tests/update/test_update_rollback_failure/test_update_rollback_failure/pcluster.config.update.yaml
```diff
@@ -0,0 +1,57 @@
...
+  Os: {{ os }}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+  Ssh:
+    KeyName: {{ key_name }}
+  Iam:
+    AdditionalIamPolicies:
+      - Policy: arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
```
gmarciani, line 11: Here and in other parts of the config.
Adding this policy is not required as it is automatically injected by our test framework.
See https://github.com/aws/aws-parallelcluster/blob/release-3.14/tests/integration-tests/conftest.py#L858-L866

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
```diff
@@ -0,0 +1,57 @@
...
+      ComputeResources:
+        - Name: queue2-cr1
+          Instances:
+            - InstanceType: c5.large
+          MinCount: 0
+          MaxCount: 5
+      Networking:
+        SubnetIds:
+          - {{ private_subnet_id }}
+Monitoring:
```
gmarciani: Here an in the other config files, why we need to explicitly set this to true? 
If there is not a valid reason, let's remove this params and rely on the default value, which is true.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+"""
+Integration tests for verifying fixes related to cluster update rollback scenarios.
+
+This test validates the following fixes:
+- [F1] clustermgtd remains running after both update and rollback fail
+- [F2] cfn-hup does not enter an endless loop after rollback to a state older than 24h
+- [F3] dna.json files are cleaned up after update failure
```
gmarciani: [minor] Worth mentioning :update and rollback failure

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+"""
+Integration tests for verifying fixes related to cluster update rollback scenarios.
+
+This test validates the following fixes:
+- [F1] clustermgtd remains running after both update and rollback fail
```
gmarciani: [minor] Worth mentioning: we expect it to be running if the update and rollback fail in the section of the update that we consider safe, which is after the slurm reconfiguration

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+6. During update's cluster readiness check (before rollback starts): inject failure on CN2
+   - CN2 has already applied the update at this point
+   - Disabling cfn-hup prevents CN2 from rolling back
+   - Cluster readiness check has 10 attempts (~15 min window)
+7. Rollback starts after update fails
+8. Rollback fails because CN2 won't rollback (still has update target config)
+9. Verify fixes:
+   - clustermgtd is running
+   - dna.json files are deleted
+   - CN3 (healthy node) has correct config version (source config before update)
```
gmarciani: LEt's also capture what we expect for the unhealthy nodes

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+@pytest.mark.usefixtures("os", "instance", "scheduler")
+def test_update_rollback_fixes(
+    region,
+    pcluster_config_reader,
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+):
+    """
+    Test that cluster update rollback fixes work correctly.
```
gmarciani: This is a repetition of what we documented in the comment aboie. Let's consolidate the comments into one comprehensive description to be put only here

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = SlurmCommands(remote_command_executor)
+
+    # Wait for all static nodes to be ready
+    _wait_for_static_nodes_ready(slurm_commands, expected_count=3)
+
+    # Get compute node hostnames
+    compute_nodes = slurm_commands.get_compute_nodes()
+    assert_that(len(compute_nodes)).is_greater_than_or_equal_to(3)
```
gmarciani: Why is this assertion required if we already have `_wait_for_static_nodes_ready(slurm_commands, expected_count=3)` above?
If not required, let's remove it

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+    # Map hostnames to instance IDs
+    cn1_instance_id = _get_instance_id_for_node(region, remote_command_executor, cn1, instance_ids)
+    cn2_instance_id = _get_instance_id_for_node(region, remote_command_executor, cn2, instance_ids)
+    cn3_instance_id = _get_instance_id_for_node(region, remote_command_executor, cn3, instance_ids)
+    logger.info(f"CN1: {cn1} -> {cn1_instance_id}")
+    logger.info(f"CN2: {cn2} -> {cn2_instance_id}")
+    logger.info(f"CN3: {cn3} -> {cn3_instance_id}")
+
+    # Get initial config version from DynamoDB (dna.json is cleaned up after successful create)
+    initial_config_version = _get_config_version_from_ddb(region, cluster.name, cn3_instance_id)
```
gmarciani, line 99: why retrieving it from the DDB record of a specific compute rather than from the head node?

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+    logger.info(f"Disabling cfn-hup on CN1 ({cn1}) before update...")
+    _disable_cfn_hup_on_compute_node(remote_command_executor, cn1)
+
+    # Step 4: Trigger cluster update with wait=False (non-blocking)
+    logger.info("Triggering cluster update (non-blocking)...")
+    updated_config_file = pcluster_config_reader(config_file="pcluster.config.update.yaml")
+
+    # Get the target config version from the update config file
+    # We'll use this to verify CN2 has applied the update before disabling its cfn-hup
+    cluster.update(str(updated_config_file), wait=False, raise_on_error=False, log_error=False)
```
gmarciani: Why `log_error=False`? even if we want to ignore the failure b/c the failure is expected, it could be useful to emit the error that triggers the expected failure.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+
+    # Step 6: Wait for CN2 to complete the update, then disable its cfn-hup
+    # CN2 needs to successfully apply the update first (have the new config version in DDB)
+    # Then we disable cfn-hup so it won't rollback (keeping the update target config)
+    logger.info("Waiting for CN2 to complete update before disabling its cfn-hup...")
+    _wait_for_node_config_version_change(
+        region, cluster.name, cn2_instance_id, initial_config_version, timeout_minutes=15
+    )
+
+    logger.info(f"CN2 has applied the update. Disabling cfn-hup on CN2 ({cn2})...")
```
gmarciani: Let's specify in this log line that we are doing this way to inject rollback failure. In this way it will be more clear the goal when we will go through the test logs.

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+    _wait_for_node_config_version_change(
+        region, cluster.name, cn2_instance_id, initial_config_version, timeout_minutes=15
+    )
+
+    logger.info(f"CN2 has applied the update. Disabling cfn-hup on CN2 ({cn2})...")
+    _disable_cfn_hup_on_compute_node(remote_command_executor, cn2)
+
+    # Wait for stack to reach UPDATE_ROLLBACK_COMPLETE state
+    logger.info("Waiting for stack to reach UPDATE_ROLLBACK_COMPLETE...")
+    final_status = _wait_for_stack_rollback_complete(cluster, region)
```
gmarciani, line 134: The stack reaches the state `UPDATE_ROLLBACK_COMPLETE` before the actual rollback completed in the head node. This is a known limitation of our rollback. We should consider this before executing the assertions below, otherwise we risk false positive.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+    # Note: supervisorctl status returns exit code 3 when process is STOPPED, so we use raise_on_error=False
+    result = remote_command_executor.run_remote_command(
+        f"srun -w {node_name} sudo {supervisorctl_path} status cfn-hup",
+        raise_on_error=False,
+    )
+    assert_that(result.stdout).contains("STOPPED")
+    logger.info(f"cfn-hup stopped on {node_name} ✓")
+
+
+def _wait_for_stack_rollback_complete(cluster, region, timeout_minutes=60):
```
gmarciani: Why do we need to nest functions here? If the only reason is to inject the timeout_minutes parameter, I suggest to simplify by removing the nested functions and use a static timeout that we think it is reasonable.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
```diff
@@ -0,0 +1,496 @@
...
+    logger.info("Verifying clustermgtd is running...")
+
+    # Find the supervisorctl path
+    result = remote_command_executor.run_remote_command(
+        "find /opt/parallelcluster -name supervisorctl -type f 2>/dev/null | head -1"
+    )
+    supervisorctl_path = result.stdout.strip()
+
+    if not supervisorctl_path:
+        supervisorctl_path = "/opt/parallelcluster/pyenv/versions/3.12.11/envs/cookbook_virtualenv/bin/supervisorctl"
```
gmarciani: This logic to call supervisorctl is duplicated in this test. Let's move it to an utility function.

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+
+    # If we reach here, no record was found - this is a test failure
+    pytest.fail(f"No DynamoDB record found for instance {instance_id}")
+
+
+def _verify_metadata_db_updated(remote_command_executor):
+    """Verify that metadata_db.json was updated (cfn-hup processed the change)."""
+    logger.info("Verifying metadata_db.json is updated...")
+
+    result = remote_command_executor.run_remote_command(
```
gmarciani: This is about checking existence of a file in a node. 
This is a logic that, if it does not exist yet, it would be helpful to other tests as well. 
I suggest to move to an utility function that can be used by other tests.

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+
+    _check_clustermgtd()
+    logger.info("clustermgtd is running ✓")
+
+
+def _verify_dna_json_cleaned_up(remote_command_executor):
+    """Verify that dna.json files are cleaned up after update failure."""
+    logger.info("Verifying dna.json files are cleaned up...")
+
+    result = remote_command_executor.run_remote_command(
```
gmarciani, line 336: This is about checking existence of a files in a node. 
This is a logic that, if it does not exist yet, it would be helpful to other tests as well. 
I suggest to move to an utility function that can be used by other tests.

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+                raise Exception(f"Config version not changed yet (still {old_version})")
+            logger.info(f"Instance {instance_id} config version changed to: {current_version}")
+            return current_version
+        except ValueError:
+            raise Exception(f"DynamoDB record not found for {instance_id}")
+
+    return _check_version()
+
+
+def _verify_compute_node_config_version_in_ddb(region, cluster_name, instance_id, expected_version):
```
gmarciani: This is about checking the config version deployed to a cluster node.
This is a logic that, if it does not exist yet, it would be helpful to other tests as well. 
I suggest to move to an utility function that can be used by other tests in future.

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+    """Verify that metadata_db.json was updated (cfn-hup processed the change)."""
+    logger.info("Verifying metadata_db.json is updated...")
+
+    result = remote_command_executor.run_remote_command(
+        "test -f /var/lib/cfn-hup/data/metadata_db.json && echo 'exists' || echo 'not found'"
+    )
+    assert_that(result.stdout.strip()).is_equal_to("exists")
+
+    # Also check the modification time is recent (within last hour)
+    result = remote_command_executor.run_remote_command("stat -c %Y /var/lib/cfn-hup/data/metadata_db.json")
```
gmarciani: This is about checking update time of a file in a node.
This is a logic that, if it does not exist yet, it would be helpful to other tests as well.
I suggest to move to an utility function that can be used by other tests.

tests/integration-tests/configs/develop.yaml
hehe7318: Makes sense!

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
hehe7318: Thanks for catching the test configuration file issues, I copy pasted these configs from some test_slurm tests config, I wanted to check them before push the changes to GH but forgot.:blush:

Removing

hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
hehe7318: Removed!

tests/integration-tests/tests/update/test_update_rollback_failure/test_update_rollback_failure/pcluster.config.update.yaml
hehe7318: Removed!

tests/integration-tests/tests/update/test_update_rollback_fixes/test_update_rollback_fixes/pcluster.config.update.yaml
hehe7318: Removed!

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: added!

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Added.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: done

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: What do you mean, I didn't get it. `dna.json` is cleaned up on HeadNode after cluster create, we can not get the config version from it.

hehe7318: Oh I got you, you mean get record from DDB, but with HeadNode id. What's the difference? The goal of the test is to verify whether the compute node (CN3) has the correct configuration version after the rollback. So, I think if we directly get the baseline value from the DDB record with CN3 ID, and then verify whether it remains unchanged, it makes the test logic clearer.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
gmarciani: I am a little bit concerned by the duration of the test. Can we shorten the 60min titomeout?

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    # Create cluster with initial configuration (3 static compute nodes)
+    init_config_file = pcluster_config_reader()
+    cluster = clusters_factory(init_config_file)
+
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    slurm_commands = SlurmCommands(remote_command_executor)
+
+    # Wait for all static nodes to be ready
+    _wait_for_static_nodes_ready(slurm_commands, expected_count=3)
```
gmarciani: We refer to `3` in multiple places. Can we define a variable, e.g. `n_static_nodes = 3` and use it everywhere we need it, including the cluster config?

tests/integration-tests/tests/update/test_update_rollback_failure/test_update_rollback_failure/pcluster.config.update.yaml
```diff
@@ -0,0 +1,57 @@
...
+      CapacityType: ONDEMAND
+      Iam:
+        AdditionalIamPolicies:
+          - Policy: arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
+      ComputeResources:
+        - Name: queue1-cr1
+          Instances:
+            - InstanceType: c5.large
+          MinCount: 3
+          MaxCount: 5
```
gmarciani: Why 2 dynamic nodes? It seems that the test only needs the 3 static ones

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Done, removed, thank you for pointing this out

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: Makes a lot of sense, I added a new logic that check the last line of chef-client log, to ensure chef finished.

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Removed

tests/integration-tests/tests/update/test_update_rollback_fixes.py
hehe7318: Shorten to 30 mins

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: It's just two lines and we have many options to check if a file exists, I don't think people will check util.py first before they write their own way to check if it exists. Let's skip this.

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: It's just two lines and we have many options to check if a file exists, I don't think people will check util.py first before they write their own way to check if it exists. Let's skip this.

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: Makes sense. Done

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: Done.

tests/integration-tests/tests/update/test_update_rollback_failure.py
hehe7318: Thank you! Done!

tests/integration-tests/tests/update/test_update_rollback_failure/test_update_rollback_failure/pcluster.config.update.yaml
hehe7318: Done

tests/integration-tests/tests/update/test_update_rollback_failure.py
gmarciani: I mean: the initial configure version is the same of the HN, so you can retrieve if reading from the head node itself at `/opt/parallelcluster/shared/cluster-config-version`. Anyway, not a blocker.

gmarciani (approved):

PR #7159, opened 2025-12-18
[BuildImage] Fix reboot failures on Ubuntu by preventing snap auto-refresh from interfering with build

gmarciani:
#### Problem
The build image for Ubuntu 22.04 and 24.04 intermittently fails the reboot step, which is executed before the execution of our cookbook. 

#### Root Cause
Our investigations show that such failure is related to snapd restarting the SSM Agent right before the reboot. snapd does it as part of its auto-refresh mechanism, which for tries to upgrade the packages in the background. We think this may interfere with the ssm agent reporting the successful status of the reboot.

#### Solution
Disable snap refresh for all snap packages during the build process.
Notice that we will do that with the following safe guards:
1. if snap is not installed, we skip the hold/unhold
2. we unhold only if it was us holding. So that if a user was setting a hold in their parent image, we leave it as it is.

This is how snap refresh hold/unhold works:

```
root@ip-172-31-0-49:~# snap refresh --time
timer: 00:00~24:00/4
last: today at 15:59 UTC
next: today at 22:03 UTC

root@ip-172-31-0-49:~# snap refresh --hold
Auto-refresh of all snaps held indefinitely

root@ip-172-31-0-49:~# snap refresh --time
timer: 00:00~24:00/4
last: today at 15:59 UTC
hold: forever
next: today at 22:03 UTC (but held)

root@ip-172-31-0-49:~# snap refresh --unhold
Removed auto-refresh hold on all snaps

root@ip-172-31-0-49:~# snap refresh --time
timer: 00:00~24:00/4
last: today at 15:59 UTC
next: today at 22:03 UTC
```

#### Q&A
1. Why re-enabling after the build process?
Because keeping it disabled would lower the security bar of the product for Ubuntu. Having snap auto-refresh enabled guarantess that packages installed with snap are automatically upgraded in the background

### Tests
* Verified fix on Ubuntu 22.04 and Ubuntu 24.04 builds that were previously failing
* Confirmed snap refresh hold is applied and removed correctly

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -34,6 +34,19 @@ phases:
                 *) echo "Unsupported OS: $RELEASE" && exit {{ FailExitCode }} ;;
               esac
 
+      # Disable snap refresh to prevent background updates during build (Ubuntu only)
+      - name: DisableSnapRefresh
+        action: ExecuteBash
+        inputs:
+          commands:
+            - |
+              set -x
+              OS="{{ build.OperatingSystemName.outputs.stdout }}"
+              if [[ $OS =~ ^ubuntu ]] && command -v snap &>/dev/null; then
+                snap watch --last=auto-refresh 2>/dev/null || true
+                snap refresh --hold && mkdir -p /opt/parallelcluster && touch /opt/parallelcluster/pcluster_build_image_snap_hold
+              fi
+
       # Initialize system information and URLs
       - name: SystemInfo
         action: ExecuteBash
@@ -258,6 +271,11 @@ phases:
                 rm -rf /tmp/imagebuilder_service/ssm_installed
               else
                  echo "SSM agent is installed by default"
+              fi
+              
+              if [[ -f /opt/parallelcluster/pcluster_build_image_snap_hold ]]; then
+                snap refresh --unhold 2>/dev/null || true
+                rm -rf /opt/parallelcluster/pcluster_build_image_snap_hold
               fi
               
               # Final cleanup
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ CHANGELOG
 - Reduce EFA installation time for Ubuntu by ~20 minutes by only holding kernel packages for the installed kernel.
 - Add GetFunction and GetPolicy permissions to PClusterBuildImageCleanupRole to prevent AccessDenied errors during build image stack deletion.
 - Fix validation error messages when `DevSettings` is null or `DevSettings/InstanceTypesData` is missing required fields.
+- Disable snap auto-refresh on Ubuntu during build image to prevent intermittent reboot failures.
 
 3.14.0
 ------


hanwen-cluster (dismissed):

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -260,6 +273,10 @@ phases:
                  echo "SSM agent is installed by default"
               fi
               
+              if [[ $OS =~ ^ubuntu ]]; then
```
himani2411: [Non-blocking] Lets add a comment to mention we remove the hold

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -34,6 +34,19 @@ phases:
                 *) echo "Unsupported OS: $RELEASE" && exit {{ FailExitCode }} ;;
               esac
 
+      # Disable snap refresh to prevent background updates during build (Ubuntu only)
```
himani2411, line 37: [Non-blocking] We recently added the chnage to reduce the size of this template where one of the tricks we used was combine a few steps; lets do that going forward so thta we dont keep facing the template size issues in future.

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani: Comments increase the template size, so I would add them only if the code is unclear. 
It seems to me this line is self-explanatory so we can avoid an extra line of comment.
Do you think this code needs a comment?

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani: This is a good point, thank you. 
However, we must disable snap auto-refresh as soon as possible to prevent the background upgrades and `SystemInfo` step did not seem good fit. I'm in favor of refactoring the component to mitigate the risk of template size exceeding, but will address out of this last minute fix.

himani2411: You dont really improve in terms of time by adding another step as the steps are sequential; so i disagree on `SystemInfo` being a bad fit since the code would be applied at the same time

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
himani2411: Agreed

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -34,6 +34,19 @@ phases:
...
+      - name: DisableSnapRefresh
+        action: ExecuteBash
+        inputs:
+          commands:
+            - |
+              set -x
+              OS="{{ build.OperatingSystemName.outputs.stdout }}"
+              if [[ $OS =~ ^ubuntu ]] && command -v snap &>/dev/null; then
+                snap watch --last=auto-refresh 2>/dev/null || true
+                snap refresh --hold && touch /tmp/pcluster_build_image_snap_hold
```
hanwen-cluster: Instead of using /tmp, we should use other folders like the code here https://github.com/aws/aws-parallelcluster/pull/7159/files#diff-365de0075f4659870e5ed9878016b28a29065c60bc7bb763da54670e5b626eb6R104-R105

hanwen-cluster (dismissed):

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
gmarciani: This comment is really impokrtant, thank you! In fact I just realized with testing that relying on /tmp files could also lead to unhold being skipped because the tmp file gets removed after the reboot.

Now using a file in `/opt/parallelcluster`

hehe7318 (approved):

himani2411 (approved):

PR #7180, opened 2025-12-29
[develop][cli] Improve curl retry for GitHub package downloads

hehe7318:
### Description of changes
- Increase retry count from 3 to 9
- Add `--retry-all-errors` to retry on SSL and HTTP/2 errors

If we retry 9 times, it will wait:
1 + 2 + 4 + 8 + 16 + 32 + 64 + 128 + 256 = 511 seconds = 8.51 minutes
It should can handle the China region network issue.

This fixes intermittent download failures in China regions caused by network instability when accessing GitHub.

### Tests
* build-image E2E test in China region ongoing

### References
* Cookbook PR for improving node package downloading: https://github.com/aws/aws-parallelcluster-cookbook/pull/3085

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -219,7 +219,7 @@ phases:
               
               # Download cookbook
               mkdir -p /etc/chef && sudo chown -R root:root /etc/chef
-              curl --retry 3 -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL"
+              curl --retry 10 --retry-all-errors -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL"
               
               cd /etc/chef && tar -xzf /etc/chef/aws-parallelcluster-cookbook.tgz --strip-components 1 && rm -f aws-parallelcluster-cookbook.tgz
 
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -219,7 +219,7 @@ phases:
               
               # Download cookbook
               mkdir -p /etc/chef && sudo chown -R root:root /etc/chef
-              curl --retry 3 -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL"
+              curl --retry 10 --retry-all-errors -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL"
```
himani2411: How many retires did it need for the test to pass in china? I think 17 mins is too high we add this 17 mins overall
Lets keep a smaller as the time to actually connect to the URL and then fail is still not counted. So far as you mentioned it never retired and failed within for example 2 mins.
So with every 1 + 2 + 4 + 8 + 16 + 32 + 64 + 128 + 256 + 512 = 1023 seconds = 17.05 minutes we would have to add the 2 mins in-between to get the actual increase in time

hehe7318: As discussed, decreased to 9 times. In total 511 seconds.

himani2411 (approved):

PR #7191, opened 2026-01-09
[Networking] Prevent duplicate security group rules in shared storage security group

gmarciani:
### Description of changes
Prevent duplicate security group rules in shared storage security group.

The deduplication is needed because when duplicate rules are requested, the CFN handler responsible for their creation goes into a path that is more susceptible to eventual consistency issue.
When such issue occur, cluster creation fail.

### Tests
SUCCEEDED integration tests: 
* test_fsx_lustre.py::test_multiple_fsx
* test_fsx_lustre.py::test_fsx_lustre_dra
* test_fsx_lustre.py::test_file_cache
* test_efs.py::test_multiple_efs
* test_internal_efs.py::test_internal_efs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ CHANGELOG
 - Fix LoginNodes NLB not being publicly accessible when using public subnet with implicit main route table association. 
   See https://github.com/aws/aws-parallelcluster/issues/7173
 - Fix cluster creation failure without Internet access when GPU instances and DCV are used.
+- Fix intermittent cluster creation failure caused by eventual consistency issues when head, compute and login nodes have the same security group.
 
 3.14.1
 ------
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -795,8 +795,17 @@ class ClusterCdkStack:
             "Storage": [storage_security_group.ref],
         }
 
+        # Deduplicate SGs to avoid creating identical rules for the same security group.
+        # This prevents EC2 duplicate rule errors which are susceptible to eventual consistency issues.
+        seen_sgs = set()
         for sg_type, sg_refs in target_security_groups.items():
             for sg_ref_id, sg_ref in enumerate(sg_refs):
+                # This conversion is required because sg_ref could either be a string or a CDK token reference.
+                # A CDK token must be converted to a string to make it comparable within the set.
+                sg_key = str(sg_ref)
+                if sg_key in seen_sgs:
+                    continue
+                seen_sgs.add(sg_key)
                 # TODO Scope down ingress rules to allow only traffic on the strictly necessary ports.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,12 +8,14 @@ CHANGELOG
...
 - Upgrade jmespath to ~=1.0 (from ~=0.10).
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
+- Scope down storage security group ingress rules to specific ports for all node types.
 
 **BUG FIXES**
 - Add validation to block updates that change tag order. Blocking such change prevents update failures.
 - Fix LoginNodes NLB not being publicly accessible when using public subnet with implicit main route table association. 
   See https://github.com/aws/aws-parallelcluster/issues/7173
 - Fix cluster creation failure without Internet access when GPU instances and DCV are used.
+- Fix cluster creation failure caused by eventual consistency issues when head, compute and login nodes have the same security group.
```
hanwen-cluster: Can you say the failure is intermittent?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -795,19 +795,26 @@ def _add_storage_security_group(self, storage_cfn_id, storage):
...
-                        ingress_port = EFS_PORT
-                    elif storage_type == SharedStorageType.FSX:
-                        ingress_protocol = "tcp"
-                        ingress_port = FSX_PORTS[LUSTRE]["tcp"][0]
+                if storage_type == SharedStorageType.EFS:
+                    ingress_protocol = "tcp"
+                    ingress_port = EFS_PORT
+                elif storage_type == SharedStorageType.FSX:
+                    ingress_protocol = "tcp"
+                    ingress_port = FSX_PORTS[LUSTRE]["tcp"][0]
```
hanwen-cluster: Only open up to 988 port might be prone to error because FSx Lustre actually require a few more ports. Official doc: https://docs.aws.amazon.com/fsx/latest/LustreGuide/limit-access-security-groups.html

How about not making this change? because this is not strictly related to the issue we are trying to fix

hanwen-cluster (changes requested):

cli/src/pcluster/templates/cluster_stack.py
gmarciani: This means that we have a bug on login nodes SG because for them we only open port 988.
I agree it is a different bug, so it is worth it to track the changes in two distinct commits, but I think we should fix it asap. Do you agree in fixing this thing in a separate commit in this PR?

CHANGELOG.md
gmarciani: Done

cli/src/pcluster/templates/cluster_stack.py
hanwen-cluster: I agree to not scope down the ports in this PR, and address this the port problem separately in another PR

hanwen-cluster (approved):

PR #7193, opened 2026-01-13
[Security] Scoped down the network access to managed EFS and FSx storage

gmarciani:
### Description of changes
Scoped down the network access to managed EFS and FSx storage from cluster nodes.
In particular, before this change all traffic used to be allowed from cluster nodes to the shared storage.
With this change this access is restricted to the strictly required ports.

For EFS:
  * Allow all traffic from within the storage security group itself.
  * Allow TPC traffic on port 2049 form cluster nodes.
    
For FSx:
  * Allow all traffic from within the storage security group itself.
  * Allow TPC traffic on ports 988 and 1018-1023 form cluster nodes.

#### Quotas
With this change we are increasing the number of rules in the shared storage security group for FSx. 
This is not a problem in terms of quotas (60 rules max per SG), not even on large clusters (10 pools, 50 queues). Here is the math for large cluster:

```
EFS Security Group: 13 rules, because 1 port range * (head node + compute nodes + 10 login pools) + all traffic for storage
FSx Security Group: 25 rules, because 2 port ranges * (head node + compute nodes + 10 login pools) + all traffic for storage
```

Notice that even adding 4 additional security groups on head node, each pool and each queue doers not change the total number of rules, because we do not include additional security groups as source in the shared storage security group as it is unnecessary.

#### Validators
For FSx we added the ports 10128-1023 to the security group we create. Those ports are suggested but not mandatory for basic functionalities, so we open them in the security group we create, but we do not enforce them when the security group is provided by the user to avoid the introduction of an unnecessary breaking change.

### Tests
* unit tests (modified/added to cover the changes)
* * manual test creating a cluster with the max number of resources (10 pools, 50 queues, 4 additional SGs per head node, each pool and each queue)
* SUCCEEDED integration tests: 
  * test_fsx_lustre.py::test_multiple_fsx
  * test_fsx_lustre.py::test_fsx_lustre_dra
  * test_fsx_lustre.py::test_file_cache
  * test_efs.py::test_multiple_efs
  * test_internal_efs.py::test_internal_efs

### References
1. https://docs.aws.amazon.com/fsx/latest/LustreGuide/limit-access-security-groups.html

### DIFF
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -806,25 +806,33 @@ class ClusterCdkStack:
                 if sg_key in seen_sgs:
                     continue
                 seen_sgs.add(sg_key)
-                # TODO Scope down ingress rules to allow only traffic on the strictly necessary ports.
-                #      Currently scoped down only on Login nodes to limit blast radius.
-                ingress_protocol = "-1"
-                ingress_port = ALL_PORTS_RANGE
-                if sg_type == "Login":
-                    if storage_type == SharedStorageType.EFS:
-                        ingress_protocol = "tcp"
-                        ingress_port = EFS_PORT
-                    elif storage_type == SharedStorageType.FSX:
-                        ingress_protocol = "tcp"
-                        ingress_port = FSX_PORTS[LUSTRE]["tcp"][0]
-                ingress_rule = self._allow_all_ingress(
-                    description=f"{storage_cfn_id}SecurityGroup{sg_type}Ingress{sg_ref_id}",
-                    source_security_group_id=sg_ref,
-                    group_id=storage_security_group.ref,
-                    ip_protocol=ingress_protocol,
-                    port=ingress_port,
+
+                # For Storage-to-Storage, allow all traffic.
+                # For Head/Compute/Login nodes, allow only the required ports.
+                storage_ports = {
+                    SharedStorageType.EFS: ("tcp", [EFS_PORT]),
+                    SharedStorageType.FSX: ("tcp", FSX_PORTS[LUSTRE]["tcp"]),
+                }
+                ingress_protocol, ingress_ports = (
+                    ("-1", [ALL_PORTS_RANGE])
+                    if sg_type == "Storage"
+                    else storage_ports.get(storage_type, ("-1", [ALL_PORTS_RANGE]))
                 )
-                rules.append(ingress_rule)
+
+                for rule_id, ingress_port in enumerate(ingress_ports):
+                    ingress_rule = self._allow_all_ingress(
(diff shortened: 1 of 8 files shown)


cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -834,9 +842,6 @@ def _add_storage_security_group(self, storage_cfn_id, storage):
                 rules.append(egress_rule)
 
                 if sg_type == "Storage":
-                    ingress_rule.cfn_options.deletion_policy = ingress_rule.cfn_options.update_replace_policy = (
-                        storage_deletion_policy
-                    )
```
hanwen-cluster, line 839: Why moving this code to another place?

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -806,25 +806,33 @@ def _add_storage_security_group(self, storage_cfn_id, storage):
...
-                    source_security_group_id=sg_ref,
-                    group_id=storage_security_group.ref,
-                    ip_protocol=ingress_protocol,
-                    port=ingress_port,
+
+                # For Storage-to-Storage, allow all traffic.
+                # For Head/Compute/Login nodes, allow only the required ports.
+                storage_ports = {
+                    SharedStorageType.EFS: ("tcp", [EFS_PORT]),
+                    SharedStorageType.FSX: ("tcp", FSX_PORTS[LUSTRE]["tcp"]),
```
hanwen-cluster, line 814: Although only Lustre can be created by ParallelCluster, using hardcoded `LUSTRE` here makes the code less extensible. We should retrieve the type from `storage` object

cli/src/pcluster/templates/cluster_stack.py
gmarciani: Because we are now creating multiple rules and we must apply the retention policy to all of them. So we moved this line into the for loop where rules are created.

cli/src/pcluster/templates/cluster_stack.py
gmarciani: I see your point and I agree that we should retrieve this info from the storage object. However, it seems to me that it is not worth it. To do this thing properly the storage object should expose an attribute that must have the same values used to index the constant `FSX_PORTS`. However, this is not the case. The storage object does not have a field to capture the FSx type, so to determine the FSx type we must retrieve it from the name of the class (SharedFsxLustre, ExistingFsxOntap, ....). So we will create a constraint between a class name and a key into a dictionary, which introduced fragility just to provide an extensibility we do not need.

Doing what you suggest (which is correct) in the robust way, would cost a refactoring that is not worth it

What do you think?

hanwen-cluster (approved):

PR #7207, opened 2026-01-22
Fix cuda device query

hgreebe:
### Description of changes
* Fix cuda device query
* When upgrading to Cuda 13, the device query test was in a different folder.

### Tests
* Ran build image

### References
* https://github.com/aws/aws-parallelcluster-cookbook/pull/3100


cli/src/pcluster/resources/imagebuilder/parallelcluster_test.yaml
```diff
@@ -252,7 +252,7 @@ phases:
               echo "Correctly installed CUDA ${cuda_output}"
 
               echo "Testing CUDA with deviceQuery..."
-              if [ {{ test.OSArchitecture.outputs.stdout }} != 'arm64' ]; then
+              if [ -f /usr/local/cuda-${cuda_ver}/extras/demo_suite/deviceQuery ]; then
```
hanwen-cluster, line 255: Can you explain this change?
Why was it checking architecture and now checking whether a file exists?

hgreebe: Because before, device query was in a different location for arm, but now it has been moved to the same location as arm. But al2 is still on the old cuda version so it still needs to look at the old location.

hanwen-cluster (approved):

PR #7210, opened 2026-01-27
Support b300 Instance

hgreebe:
### Description of changes
* Support b300 Instance
* Have network interface configuration default for b300 where primary interface use an ENA network interface and all others use efa-only (Configuration is use case 1 from [this documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html))
* Add support for running test_efa with b300 instance
* Upgrade NCCL version to `2.28.9`
* Upgrade OFI NCCL plugin version to `1.18.0`
* Upgrade Fabtest version to `2.4.0`

### Tests
* Ran test_efa with ubuntu2404, alinux2023, and rocky9
* Ran gpu health checks
* Compared NCCL benchmark results with OFI plugin team


cli/src/pcluster/templates/queues_stack.py
```diff
@@ -368,9 +369,11 @@ def add_network_interfaces(
     queue_lt_security_groups,
 ):
     """Generate launch template network interfaces list."""
-    is_gb200 = compute_resource.instance_types[0].split(".")[0] == P6E_GB200
+    instance_family = compute_resource.instance_types[0].split(".")[0]
+    is_gb200 = instance_family == P6E_GB200
+    is_b300 = instance_family == P6_B300
     efa_enabled = compute_resource.efa and compute_resource.efa.enabled
-    interface_type = "efa" if efa_enabled and not is_gb200 else None
+    interface_type = "efa" if efa_enabled and not is_gb200 and not is_b300 else None
```
gmarciani: [minor] Documentation: I suggest to make this logic more talkative by either:
1. add a comment here specifying that those instance families require the first interface to be ENA even if EFA is enabled.
2. define a constant `INSTANCE_TYPES_WITH_FIRST_INTERFACE_ENA = [P6E_GB200,P6_B300]` and reference the constant in this logic

tests/integration-tests/tests/efa/test_efa/test_efa/pcluster.config.yaml
```diff
@@ -13,14 +13,14 @@ Scheduling:
...
+      {% if capacity_reservation_id %}
       CapacityType: CAPACITY_BLOCK
       CapacityReservationTarget:
-        CapacityReservationId: {{ p6_b200_capacity_reservation_id }}
+        CapacityReservationId: {{ capacity_reservation_id }}
       {% endif %}
       Networking:
         PlacementGroup:
-          Enabled: {% if instance not in ["p4d.24xlarge", "p6-b200.48xlarge"] %}true{% else %}false{% endif %}
+          Enabled: {% if instance not in ["p4d.24xlarge", "p6-b200.48xlarge", "p6-b300.48xlarge"] %}true{% else %}false{% endif %}
```
gmarciani: [minor] In the test code we have the logic to determine the capacity reservation and placement group. We could improve config readability by putting here the variable `placement_group_enabled` and set this variable in the test code accordingly. 

I know that this approach was here before your PR, so not a blocker for this PR.

cli/src/pcluster/constants.py
```diff
@@ -347,6 +347,8 @@ class Operation(Enum):
 PCLUSTER_BUILD_IMAGE_CLEANUP_ROLE_BOOTSTRAP_TAG_KEY = "parallelcluster:build-image-cleanup-role-bootstrapped"
 
 P6E_GB200 = "p6e-gb200"
+P6_B300 = "p6-b300"
```
gmarciani, line 357: when we introduced GB200 we did for a limited set of OSs. Are we sure we do not have any OS limitation for B300? For instance, according to documentation for NVIDIA 580, AL2 is not mentioned as a supported OS: https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-580-126-09/index.html

tests/integration-tests/tests/efa/test_efa/test_efa/pcluster.config.yaml
himani2411: [Minor-NON Blocking] Lets change this to check for p family instance instead of the whole instance

cli/src/pcluster/templates/queues_stack.py
gmarciani: [Test] I suggest to cover the logic that determines the type of the interfaces with unit tests. To make it better I suggest to capture the logic into dedicated helper function and cover such function with unit test rather than the resulting template.

cli/src/pcluster/constants.py
hgreebe: I addded a validator to prevent al2 and b300 from being used together.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ CHANGELOG
 - Add validator that warns against the downsides of disabling in-place updates on compute and login nodes through DevSettings.
 - Upgrade jmespath to ~=1.0 (from ~=0.10).
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
+- Add support for p6-b300 instances by having a default network configuration for those instances.
```
gmarciani: We should surface in the changelog that the support is for all OS except for AL2

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -211,6 +213,12 @@ def _validate(self, instance_type: str, os: str):
                     " Please use one of the following OS: {2}".format(instance_type, os, SUPPORTED_OSES_FOR_P6E_GB200),
                     FailureLevel.ERROR,
                 )
+        if instance_type.startswith("p6-b300") and os in UNSUPPORTED_OSES_FOR_P6_B300:
```
gmarciani: [Test]  We should capture this change with unit tests.

gmarciani: [minor] We can refer to the constant P6_B300 rather than redefining the string.

cli/src/pcluster/validators/ec2_validators.py
```diff
@@ -211,6 +213,12 @@ def _validate(self, instance_type: str, os: str):
...
+            self._add_failure(
+                "The instance type {0} is not officially supported with OS {1}."
```
gmarciani: I know we use the term "officially" elsewhere. However, I think we should avoid it to prevent misunderstanding. Let;'s simply say that it is not supported.

cli/tests/pcluster/validators/test_ec2_validators.py
```diff
@@ -461,6 +461,17 @@ def test_instance_type_base_ami_compatible_validator(
             "rocky9",
             None,
         ),
+        (
+            "p6-b300.48xlarge",
```
gmarciani: [minor] If you use `p6-b300.WHATEVER_SIZE` rather than `p6-b300.48xlarge` it will be clearer that the behavior applies whatever the size is. Otherwiose there could be a doubt that it is something specific to 48xlarge.

You applied this testing best practice in the unit test above. I suggest to apply it here as well.

gmarciani (approved):

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -390,10 +395,18 @@ def add_network_interfaces(
...
 
-        interface_type = "efa" if efa_enabled else None
-        # if efa is enabled with a gb200 instance, even indexes are configured as efa and the odd as efa-only
-        if is_gb200 and efa_enabled:
-            interface_type = "efa" if even else "efa-only"
+        if efa_enabled:
+            if is_b300:
+                # if efa is enabled with a b300 instance, all network cards, except for the primary,
+                # are configured as efa-only
+                interface_type = "efa-only"
```
hanwen-cluster, line 402: Why is the interface type "efa-only", not "efa"? p6-b300 supports EFA and ENA on non-primary network cards, right?

PR #7221, opened 2026-02-05
[Test] Adapt integ test `test_update_rollback_failure` to validate that clustermgtd gets restarted even if the update fails before cluster readiness check.

gmarciani:
### Description of changes
Adapt integ test `test_update_rollback_failure` to validate that clustermgtd gets restarted even if the update fails before cluster readiness check. In particular, we inject a failure that occurs earlier in the update recipe, which causes the restart of slurmctld to fail. To inject such failure, we temporarily set slurmctld as non executable.

**Why we commented out the assertion on compute node CN3?**
In this PR we commented out the assertion on the config version deployed by compute node CN3 after the rollback failure. In this Pr we are moving the failure injected into the rollback to the earlier stage of the update recipe. When the rollback fails in its early stage, we cannot guarantee what config version compute nodes applied.
Depending on timing, they may have deployed the rolled back config (the correct one) or the config associated to the failed update (the wrong one). The test expected the compute node CN3 to deploy the rolled back config version after the rollback failure. This used to be true only because we used to inject a failure in rollback that occurs late in the rollback procedure (fails all attempts of cluster readiness checks), which gives time to CN3 to deploy the rolled back config. This is not true anymore because we now want to inject a failure in the rollback that occurs earlier (before cluster readiness check). The inject failure occurs ~1min after the rollback is triggered. As expected on rollback failure, the DNA json files are cleaned up as expected. When CN3 detects that a rollback is required, it does not start the rollback because it waits for the DNA json to be shared from the head node, but they have been already cleaned up.

### Tests
SUCCESS `test_update_rollback_failure`

### DIFF
--- a/tests/integration-tests/tests/update/test_update_rollback_failure.py
+++ b/tests/integration-tests/tests/update/test_update_rollback_failure.py
@@ -19,7 +19,11 @@ from assertpy import assert_that
 from remote_command_executor import RemoteCommandExecutor
 from retrying import retry
 from time_utils import minutes, seconds
-from utils import get_compute_nodes_instance_ids, get_file_mtime_age_seconds, verify_cluster_node_config_version_in_ddb
+from utils import (
+    get_compute_nodes_instance_ids,
+    get_file_mtime_age_seconds,
+    match_regex_in_log,
+)
 
 from tests.common.schedulers_common import SlurmCommands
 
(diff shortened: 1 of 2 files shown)


tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 import logging
+import re
```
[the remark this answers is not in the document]
gmarciani: fixed

tests/integration-tests/utils.py
```diff
@@ -1118,3 +1118,34 @@
     mtime = int(result.stdout.strip())
     current_time = int(remote_command_executor.run_remote_command("date +%s").stdout.strip())
     return current_time - mtime
+
+
+from remote_command_executor import RemoteCommandExecutor
```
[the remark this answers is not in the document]
gmarciani: fixed

tests/integration-tests/tests/update/test_update_rollback_failure.py
```diff
@@ -118,19 +122,15 @@ def test_update_rollback_failure(
...
-    _wait_for_node_config_version_change(
-        region, cluster.name, cn2_instance_id, initial_config_version, timeout_minutes=15
-    )
-
-    logger.info(
-        f"CN2 has applied the update. Disabling pcluster-check-update timer on CN2 "
-        f"({cn2}) to inject rollback failure..."
-    )
-    _disable_check_update_timer_on_compute_node(remote_command_executor, cn2)
+    # Step 6: Inject failure in slurmctld restart
```
hehe7318, line 125: >Extend integ test test_update_rollback_failure to validate that clustermgtd gets restarted even if the update fails before cluster readiness check. 

We are saying "extend", but do we still test if clustermgtd gets restarted when cluster readiness check failed after this change? If no, do you think it's okay and why?

gmarciani: Good question.
After this change we verify that clustermgtd gets restarted if the rollback fails before cluster readiness checks.  Covering both cases it is not worth it because:
1. We can safely assume that: if it is able to restart before cluster readiness check, then it is able to restart even if readiness check fails. To introduce an issue like that in the error handler, we should add a logic in the handler that explicitly prevent the restart under certain circumstances.
2. To verify both cases, we should verify that it gets restarted in between update failure (where we inject readiness check failure) and rollback start. Since we are not in control of the rollback trigger, the verification would be open to race conditions and false failures.

The only way to verify that it was restarted after cluster readiness check is to add an assertion on the chef-client logs that verifies that the error handler after readiness check actually triggered the restart. 
Do you think it is worth it?

hehe7318: I agree that it's not worth it. Thanks for the answer, approving the PR.

hehe7318 (approved):
LGTM

PR #7245, opened 2026-02-24
Make ami_type a fixture

hgreebe:
### Description of changes
* Make `ami_type` a fixture in order to fix `function uses no argument 'ami_type'` error

### Tests
* Ran `test_efa_unsupported` and `test_pcluster_configure`

### References
* This PR did not fix the issue because it just started failing for other tests: https://github.com/aws/aws-parallelcluster/pull/7244
* Error introduced through this PR: https://github.com/aws/aws-parallelcluster/pull/7223/changes

### DIFF
--- a/tests/integration-tests/conftest_networking.py
+++ b/tests/integration-tests/conftest_networking.py
@@ -134,6 +134,12 @@ def az_id():
     pass
 
 
+@pytest.fixture(autouse=True)
+def ami_type():
+    """Removes the need to declare the fixture in all tests even if not needed."""
+    pass
+
+
 def unmarshal_az_override(az_override):
     for region, regex in ZONE_ID_MAPPING.items():
         pattern = re.compile(regex)


tests/integration-tests/conftest_networking.py
```diff
@@ -134,6 +134,12 @@ def az_id():
     pass
 
 
+@pytest.fixture(autouse=True)
+def ami_type():
+    """Removes the need to declare the fixture in all tests even if not needed."""
+    pass
```
gmarciani, line 140: I agree the solution is to define the ami_type parameter as a fixture. 
However, why a no-op is enough?
I think the fixture must not be a no-op, but instead determine the ami_type from the values passed to the test runner.

hgreebe: When tests are parameterized via unmarshal_os_params, the actual value ("DLAMI" or "official") will override the fixture. When tests use `pcluster_config_reader`, it has `ami_type` in its signature and so it automatically gets added.

In the case of the `pcluster_configure` tests, it does not need it because it doesn't actually create a cluster. Which is why we were having failures with those tests and not other ones.

hgreebe: But if you think it is safer to add the default value, like `official`, i can do that

gmarciani (approved):

PR #7249, opened 2026-02-26
[Tag] Allow Tags to be updated during pcluster update-cluster

himani2411:
### Description of changes
* Removed volume `tag_specifications` from the HeadNode launch template and added Tags directly to HeadNode CFN Resource
* Enabled [`propagate_tags_to_volume_on_creation=True`](https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_ec2/CfnInstance.html#aws_cdk.aws_ec2.CfnInstance.propagate_tags_to_volume_on_creationThese) on the HeadNode EC2 instance. This allows tag updates to propagate to HeadNode volumes automatically via the instance's tags, instead of being statically defined in the launch template which doesn't pick up tag changes during cluster updates.
* NOTE: The `propagate_tags_to_volume_on_creation` will not support Tags to be updated on the HeadNode volume, it will only update Tags on the HeadNode Ec2 Instance during an update.
* Removed the validator and Update Policy on Tags which prevented us from Re-ordering/re-shuffling Tags. https://github.com/aws/aws-parallelcluster/pull/7134
* Added units tests and integ tests

### Tests
* Unit tests and integ tests

### References

### Checklist


hgreebe:
> will not support Tags to be updated on the HeadNode volume, it will not updat Tags on the HeadNode Ec2 Instance during an update.

If tags are not being updated on the head node instances or volume, where will this update of tags be visible?

cli/src/pcluster/config/config_patch.py
```diff
@@ -182,46 +181,16 @@ def _compare_list(self, base_section, target_section, param_path, data_key, fiel
...
-                        is_list=True,
-                    )
-                )
-                return  # Don't process individual items
-            # Otherwise, fall through to item-by-item comparison below
-            # The reason I check if it is an order only change first is because if I just do I order dependent
-            # comparison and fail if there is a difference, then we don't get the item by item comparison in the
-            # change set if the modification was not the order.
-
         # Compare items in the list by searching the right item to compare through update_key value
```
gmarciani, line 184: > Compare items in the list by searching the right item

Why you need to search the items in the list?
Wouldn't it be enough to compare the old Tags element with the new Tags element?

Since we support the tags update, I expect a tags reshuffling to be considered as a change in the changeset even if the tag value for a given tag key did not change.

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -1970,9 +1970,7 @@ class ClusterSchema(BaseSchema):
 
     monitoring = fields.Nested(MonitoringSchema, metadata={"update_policy": UpdatePolicy.IGNORED})
     additional_packages = fields.Nested(AdditionalPackagesSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
-    tags = fields.Nested(
-        TagSchema, many=True, metadata={"update_policy": UpdatePolicy.UNSUPPORTED, "update_key": "Key"}
-    )
+    tags = fields.Nested(TagSchema, many=True, metadata={"update_policy": UpdatePolicy.SUPPORTED, "update_key": "Key"})
```
gmarciani, line 1973: Related to https://github.com/aws/aws-parallelcluster/pull/7249/changes#r2859334403

Why we need to specify the `update_key`?
Being `Tags` a list of items, isn't it enough to compare the list without specifying the items key?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
 - Upgrade tabulate to <=0.9.0 (from <=0.8.10).
 - Add support for p6-b300 instances for all OSs except AL2.
 - Add alarm on missing clustermgtd heartbeat.
+- Move head node volume tagging from Launch Template to instance-level tag propagation, avoiding head node replacement on tag updates during cluster-updates.
```
gmarciani: I suggest to keep the changelog entry more focused on the end user experience, which "Add support for tags updates." The fact that we implemented this feature moving the tagging info is an implementation details.

himani2411:
> > will not support Tags to be updated on the HeadNode volume, it will not updat Tags on the HeadNode Ec2 Instance during an update.
> 
> If tags are not being updated on the head node instances or volume, where will this update of tags be visible?

Typo mistake! Updated the message to mention it will only update on HN Ec2 instance

cli/src/pcluster/config/config_patch.py
hgreebe: I think that this change is reverting to the old comparison that we did to check that the tag contents did not change. This comparison did not catch tag order changes. The deleted section was added to catch tag order changes. If we want tag order changes to be visibile in the changeset when we do a cluster update, then we would need to retain part of the deleted section. If we do not care about showing order changes in the changeset, then it is okay to just do that comparison by the key.

If there is no order change, but a key value change, looking at if the tag list changed as a whole does not give a clear changeset about if the value of different tags changed. It just shows that the entire tag section was modified.

So i guess the question is how specific do we want the changeset to be? I think we should still show if specific changes to the value have occurred.

himani2411: yes we are reverting, considering that we no longer see any failures due to reshuffling, do we want to keep the old change at all?

hgreebe: it is just a question of if we want a change in order to show up in the change set

himani2411: Considering we are doing an update I will make the change to show the detailed change-set (like a value change for an existing key and a new key-value pair added or removed) similar to any other sections of pcluster config.

But we wont show order/shuffling chnages in the changeset.

gmarciani (approved):

cli/src/pcluster/schemas/cluster_schema.py
himani2411: We need this as without the key the compasison of the whole List will be None here https://github.com/aws/aws-parallelcluster/blob/8822b35248fa31441f8f65616b31e14923561e04/cli/src/pcluster/config/config_patch.py#L188

```

    update_key_value = target_nested_section.get(update_key) # returns None
    base_nested_section = next(
        (
            nested_section
            for nested_section in base_section.get(data_key, [])
            if nested_section.get(update_key) == update_key_value
        ),
        None,
    ) ## This will become None==None
 

```

gmarciani (approved):

PR #7250, opened 2026-02-26
Make HeadNode Local Storage updates UNSUPPORTED

hgreebe:
### Description of changes
* Make HeadNode LocalStorage updates UNSUPPORTED
* This includes making RootVolume updates UNSUPPORTED
* With the previous behavior, all nested properties did not support an update, so you could not modify the head node local storage during an update, but you could add local storage if it was not there
* Adding local storage to the head node creates a new launch template which causes an update failure

### Tests
* Tried to do a cluster update that added LocalStorage, it failed

### DIFF
--- a/cli/src/pcluster/schemas/cluster_schema.py
+++ b/cli/src/pcluster/schemas/cluster_schema.py
@@ -1370,7 +1370,7 @@ class HeadNodeSchema(BaseSchema):
         HeadNodeNetworkingSchema, required=True, metadata={"update_policy": UpdatePolicy.UNSUPPORTED}
     )
     ssh = fields.Nested(HeadNodeSshSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
-    local_storage = fields.Nested(HeadNodeStorageSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
+    local_storage = fields.Nested(HeadNodeStorageSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
     shared_storage_type = fields.Str(
         required=False,
         metadata={"update_policy": UpdatePolicy.UNSUPPORTED},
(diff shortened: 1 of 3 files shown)


cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -276,7 +276,7 @@ def make_resource(self, data, **kwargs):
 class HeadNodeStorageSchema(BaseSchema):
     """Represent the schema of storage attached to a node."""
 
-    root_volume = fields.Nested(HeadNodeRootVolumeSchema, metadata={"update_policy": UpdatePolicy.SUPPORTED})
+    root_volume = fields.Nested(HeadNodeRootVolumeSchema, metadata={"update_policy": UpdatePolicy.UNSUPPORTED})
```
himani2411, line 279: Lets add a Bug fix entry for this change

cli/tests/pcluster/config/test_update_policy.py
```diff
@@ -2936,3 +2936,74 @@ def test_extra_chef_attributes_action_needed(mocker, old_value, new_value):
...
+                    "InstanceType": "t3.micro",
+                    "Networking": {"SubnetId": "subnet-12345678"},
+                    "LocalStorage": {"RootVolume": {"Size": 50}},
+                },
+            },
+            {
+                "HeadNode": {
+                    "InstanceType": "t3.micro",
+                    "Networking": {"SubnetId": "subnet-12345678"},
+                    "LocalStorage": {"RootVolume": {"Size": 100}},
```
himani2411: Will we see the messages as per ACTIONS needed specified for root volumne when we do this update, will we with this change? https://github.com/aws/aws-parallelcluster/blob/1e3f05742edf7673e330a97bb354bb6d0b48ad6a/cli/src/pcluster/schemas/cluster_schema.py#L151-L156

hgreebe: It will still show the action needed message. I added this case to the unit test

himani2411 (approved):

PR #7260, opened 2026-03-03
[develop][E2E Test] Add E2E test for Slurm 25.11 expedited requeue mode feature

hehe7318:
### Description of changes
Add `test_expedited_requeue` in test_slurm.py to validate that jobs submitted with `--requeue=expedite` are treated as highest priority after ICE recovery.

- The test uses recoverable ICE simulation via `create_fleet_overrides.json` (instead of the permanent overrides.py approach used by `test_fast_capacity_failover`).
- Write create_fleet_overrides.json with invalid InstanceTypes → create_fleet returns no instances → InsufficientInstanceCapacity → nodes go down
- Recover by changing InstanceTypes back to real ones (t3.medium, c5.large) → next launch succeeds
- Verify the expedited requeue job started before the normal job

### Tests
* Test passed.

### Checklist


tests/integration-tests/tests/schedulers/test_slurm/test_expedited_requeue/pcluster.config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  Scheduler: slurm
+  SlurmQueues:
+    - Name: queue
+      Networking:
+        SubnetIds:
+          - {{ private_subnet_id }}
+      ComputeResources:
+        - Name: ice-cr
+          Instances:
+            - InstanceType: t3.medium
```
gmarciani, line 19: [non blocking] In other tests we are leveraging the fixture to inject more flexible instance types dynamically. Example: https://github.com/aws/aws-parallelcluster/blob/febc83ef1ffd4543590bd2d0394bc24971a88f2a/tests/integration-tests/tests/storage/test_ebs/test_ebs_single/pcluster.config.yaml#L36-L39

tests/integration-tests/tests/schedulers/test_slurm/test_expedited_requeue/pcluster.config.yaml
```diff
@@ -0,0 +1,26 @@
...
+            - InstanceType: c5.large
+        - Name: normal-cr
+          InstanceType: {{ instance }}
```
gmarciani: Why using single instance type rather than flexible instance types?
The only value I see is to validate that the override works for both runinstances and create-fleet.
If this is the only value, the I suggest to move to flexible instance types because the goal of the test is not to test the override mechanism, so it's not worth it to risk ICEs.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -616,6 +620,175 @@ def test_fast_capacity_failover(
...
+        jid = scheduler_commands.submit_command_and_assert_job_accepted(
+            submit_command_args={
+                "command": f'echo "START_TIME=$(date +%s)"; sleep 30; echo "{job["label"]} done"',
+                "nodes": 1,
+                "partition": common_cluster_details["partition"],
+                "host": target_nodes[0],
+                "other_options": f"{requeue_opt}--exclusive",
+            }
+        )
+        time.sleep(20)  # Add a sleep to match manual submission of job
```
gmarciani: Why do we need to simulate the manual submission?
I thought we needed at some point to surface a specific misbehavior of the feature, that we then identified as expected. If this is the case, then we can remove the sleep.

tests/integration-tests/tests/schedulers/test_slurm/test_expedited_requeue/pcluster.config.yaml
hehe7318: We can't inject more flexible instance types dynamically here. In this test we have to simulate the ICE and then restore. The instance type must be determined.

tests/integration-tests/tests/schedulers/test_slurm/test_expedited_requeue/pcluster.config.yaml
hehe7318: I removed the queue `normal-cr`, it's not used in the test. This config was copied from `test_fast_capacity_failover`. Thanks for pointing this out!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Himani wrote this function, I agree with you, removed.

tests/integration-tests/tests/schedulers/test_slurm/test_expedited_requeue/pcluster.config.yaml
gmarciani: Yes, when you restore you can restore to use the dynamic flex instance types to reduce the risk of ICEs. why you cannot on the restore_

Iàll not block on this comment, PR approved. However, Iàd like to understand why we cannot do that. If it is a friction in the way we inject flexible instance type we may want to remove such friction to make that injection applicable everywhere.

gmarciani (approved):

PR #7261, opened 2026-03-03
Only run RCA if head node is accessible

hgreebe:
### Description of changes
* Only run RCA if head node is accessible
* Allows cluster creation error message to be clearly shown

### Tests
* Ran test with the following error: `Invalid cluster configuration: Cannot find official ParallelCluster AMI`

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -567,7 +567,15 @@ class ClustersFactory:
                         # in the case where the stack has been deleted.
                         stack_id = response.get("cloudformationStackArn")
                         events = get_cfn_events(stack_name=stack_id, region=cluster.region)
-                        rca_details = retrieve_rca_details(cluster)
+
+                        # Only attempt RCA if head node exists
+                        rca_details = None
+                        try:
+                            _ = cluster.head_node_ip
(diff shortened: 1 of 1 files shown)


tests/integration-tests/clusters_factory.py
```diff
@@ -567,7 +567,15 @@ def create_cluster(self, cluster, request, log_error=True, raise_on_error=True,
...
                         events = get_cfn_events(stack_name=stack_id, region=cluster.region)
-                        rca_details = retrieve_rca_details(cluster)
+
+                        # Only attempt RCA if head node exists
+                        rca_details = None
+                        try:
+                            _ = cluster.head_node_ip
+                            rca_details = retrieve_rca_details(cluster)
+                        except Exception as e:
+                            logging.warning("Skipping RCA, head node not accessible: %s", e)
```
gmarciani, line 577: [Non blocking] The long term solution is not to skip the RCA if head node is unaccessible, but instead return the reason of the failure. We can start simple and report as RCA conclusions that Head Node was not created.

This comment is non blocking because I got that the intention here is to avoid the RCA failure to cover the real error, but important to be aligned on the next steps, that could be in another PR.

gmarciani (approved):

PR #7262, opened 2026-03-03
Move EFA default from combined `efa` to `interface` + `efa-only`

hanwen-cluster:
### Description of changes
* Breaking change: when EFA is enabled, network interfaces now use `interface` + `efa-only`
instead of the combined `efa` type.

See individual commit descriptions for details.

### Tests
* test_efa has passed on c6in.32xlarge, hpc6id.32xlarge, p4d.24xlarge, c5n.18xlarge on Amazon Linux 2023, RHEL9, Ubuntu 24
* test_multiple_nics has passed  on c6in.32xlarge, hpc6id.32xlarge, p4d.24xlarge on RHEL9, Ubuntu 24

### Checklist


cli/src/pcluster/templates/queues_stack.py
himani2411, line 180: Do we need this now?

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -368,43 +371,95 @@ def add_network_interfaces(
...
     is_gb200 = instance_family == P6E_GB200
-    is_b300 = instance_family == P6_B300
     efa_enabled = compute_resource.efa and compute_resource.efa.enabled
-    # gb200 and b300 instances require the first interface to be ENA even if EFA is enabled
-    interface_type = "efa" if efa_enabled and instance_family not in INSTANCE_TYPES_WITH_FIRST_INTERFACE_ENA else None
+    use_legacy_efa = efa_interface_type == "efa"
+
+    # Determine NCI-0 MaxENIs to decide if we can fit 2 ENIs (interface + efa-only)
+    nci0_max_enis = (
+        compute_resource.network_cards_list[0].maximum_network_interfaces()
```
himani2411, line 395: Are e sure that the DescribeInstanceType API output is idempotent and it will return the NIC0 as the first item in the list? AFAIK none of Decscribe* API's in AWS guarantees this

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -368,43 +371,95 @@ def add_network_interfaces(
...
+        if compute_resource.network_cards_list
+        else 1
+    )
+
+    # Infer whether NCI-0 supports EFA from API data
+    max_efa = compute_resource.max_efa_interfaces
+    max_cards = compute_resource.max_network_cards
```
himani2411, line 402: the way we calculate `max_network_cards` if we have Multiple Instance Types is by taking the MaxNetworkCard of the Instance which has the lowest nos of NICs. would this affect the verification we do for NIC0 efa support?

cli/src/pcluster/config/cluster_config.py
```diff
@@ -2376,6 +2383,11 @@ def max_network_cards(self) -> int:
                 least_max_nics = min(least_max_nics, max_nics)
         return least_max_nics
 
+    @property
+    def max_efa_interfaces(self) -> int:
```
hgreebe, line 2387: This may cause issues if one instance supports efa on NIC0 and the other does not. 

We have validators like `InstancesCPUValidator` that check that the instances have the same number of cpus, we should add a similar validator to check that they support a similar network configuration.

cli/src/pcluster/templates/queues_stack.py
hanwen-cluster: Probably no. EFA is enabled by default. Can we exclude this change because it is not in the scope of my PR?

cli/src/pcluster/config/cluster_config.py
hanwen-cluster: Great suggestion! I added a validation https://github.com/aws/aws-parallelcluster/pull/7262/changes#diff-de6e1b9e7ca98522d0b9dcaf0df1697005dd0d3392c7c1975eb3eadfd395e556R211-R225

cli/src/pcluster/templates/queues_stack.py
hanwen-cluster: Great question! I added a validation https://github.com/aws/aws-parallelcluster/pull/7262/changes#diff-de6e1b9e7ca98522d0b9dcaf0df1697005dd0d3392c7c1975eb3eadfd395e556R211-R225

cli/src/pcluster/templates/queues_stack.py
hanwen-cluster: Great suggestion! I added a sort https://github.com/aws/aws-parallelcluster/pull/7262/changes#diff-b7df161aac3c801da17294f416af87e959f5363e969733154c52e286156ad1a3R271-R273

cli/src/pcluster/config/cluster_config.py
hgreebe: This still does not cover the case where one instance supports efa on NIC0 and one does not. 

If InstanceA has max_network_cards = 5 and max_efa_interfaces = 5
and InstanceB has max_network_cards = 6 and max_efa_interfaces = 5

Then when used as flexible instance types, max_network_cards would return 5 and max_efa_interfaces would return 5.

So `nci0_supports_efa = max_efa == max_cards` would be True, despite InstanceB not supporting efa on NCI0

Would it be worth is to add a validator that checks that the instances used in flexible instance types have a similar network configuration?

hanwen-cluster: Thank you for the detailed description!

I added another validation https://github.com/aws/aws-parallelcluster/pull/7262/changes#diff-de6e1b9e7ca98522d0b9dcaf0df1697005dd0d3392c7c1975eb3eadfd395e556R227-R243

Though I have to say this is a super small corner case

hgreebe (approved):

PR #7268, opened 2026-03-05
Add LaunchTemplateOverrides to cluster config

hgreebe:
### Description of changes
* Add LaunchSpecificationOverrides to cluster config
* Merge process: For each top-level property in the user's launch template, it calls `add_property_override` with the entire value. This means:
  * Lists (like NetworkInterfaces) - Replaced entirely
  * Dicts (like MetadataOptions) - Also replaced entirely at the top level
 * Add validator to only allow launch templates that only specify network interfaces
 * Added unit test to test merge logic and validators
 * Added integration test to validate merging logic 

### Tests
* Ran unit tests and `test_launch_template_overrides` integration test created in this pr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ CHANGELOG
 - Add alarm on missing clustermgtd heartbeat.
 - Support updates of `Tags` during cluster-updates.
 - Add `LaunchSpecificationOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
+  - This overrides the parallelcluster default using a shallow merge. 
 
 **BUG FIXES**
 - Add validation to block updates that change tag order. Blocking such change prevents update failures.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ CHANGELOG
 - Add support for p6-b300 instances for all OSs except AL2.
 - Add alarm on missing clustermgtd heartbeat.
 - Support updates of `Tags` during cluster-updates.
+- Add `LaunchSpecificationOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
 
 **BUG FIXES**
 - Add validation to block updates that change tag order. Blocking such change prevents update failures.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ CHANGELOG
 - Add support for p6-b300 instances for all OSs except AL2.
 - Add alarm on missing clustermgtd heartbeat.
 - Support updates of `Tags` during cluster-updates.
-- Add `LaunchSpecificationOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
+- Add `LaunchTemplateOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
   - This overrides the parallelcluster default using a shallow merge.
 
 **BUG FIXES**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ CHANGELOG
 - Add alarm on missing clustermgtd heartbeat.
 - Support updates of `Tags` during cluster-updates.
 - Add `LaunchSpecificationOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
-  - This overrides the parallelcluster default using a shallow merge. 
+  - This overrides the parallelcluster default using a shallow merge.
 
 **BUG FIXES**
 - Add validation to block updates that change tag order. Blocking such change prevents update failures.
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -683,3 +683,27 @@ class Ec2Client(Boto3Client):
                 reservation_type = reservation.reservation_type()
 
         return instance_type, reservation_type
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_launch_template_version(self, launch_template_id: str, version: str = "$Default"):
+        """
+        Describe a launch template version and return its data.
+
+        Args:
+            launch_template_id: The launch template ID (e.g., 'lt-0abc123def456')
+            version: The version number or '$Default'/'$Latest'
+
+        Returns:
+            dict: The LaunchTemplateData from the specified version
+        """
+        response = self._client.describe_launch_template_versions(
+            LaunchTemplateId=launch_template_id,
+            Versions=[str(version)],
+        )
+        versions = response.get("LaunchTemplateVersions", [])
+        if versions:
+            return versions[0].get("LaunchTemplateData", {})
+        raise AWSClientError(
+            function_name="describe_launch_template_versions",
+            message=f"Launch template {launch_template_id} version {version} not found",
+        )
--- a/cli/src/pcluster/aws/ec2.py
+++ b/cli/src/pcluster/aws/ec2.py
@@ -685,7 +685,7 @@ class Ec2Client(Boto3Client):
         return instance_type, reservation_type
 
     @AWSExceptionHandler.handle_client_exception
-    def describe_launch_template_version(self, launch_template_id: str, version: str = "$Default"):
+    def describe_launch_template_version(self, launch_template_id: str, version: str):
         """
         Describe a launch template version and return its data.
 
--- a/cli/src/pcluster/config/cluster_config.py
+++ b/cli/src/pcluster/config/cluster_config.py
@@ -878,6 +878,15 @@ class Efa(Resource):
         self.gdr_support = Resource.init_param(gdr_support, default=False)
 
 
+class LaunchSpecificationOverrides(Resource):
+    """Represent the Launch Specification Overrides configuration."""
+
+    def __init__(self, launch_template_id: str = None, version: int = None, **kwargs):
+        super().__init__(**kwargs)
+        self.launch_template_id = Resource.init_param(launch_template_id)
+        self.version = Resource.init_param(version)
+
+
 # ---------------------- Health Checks ---------------------- #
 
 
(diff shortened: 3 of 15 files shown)


cli/src/pcluster/config/cluster_config.py
```diff
@@ -878,6 +879,15 @@ def __init__(self, enabled: bool = None, gdr_support: bool = None, **kwargs):
         self.gdr_support = Resource.init_param(gdr_support, default=False)
 
 
+class LaunchSpecificationOverrides(Resource):
```
hanwen-cluster: Let's confirm with the team whether we want to use LaunchSpecificationOverrides or LaunchTemplateOverrides

cli/src/pcluster/aws/ec2.py
```diff
@@ -683,3 +683,27 @@ def get_instance_type_and_reservation_type_from_capacity_reservation(
                 reservation_type = reservation.reservation_type()
 
         return instance_type, reservation_type
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_launch_template_version(self, launch_template_id: str, version: str = "$Default"):
```
hanwen-cluster: How about removing the `= "$Default"`? It is cleaner to avoid spreading defaults in multiple files. Majority of the defaults should be in `cluster_config.py`. 

I know this default is not used by existing callers. So this comment is only about code style, not correctness

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -829,6 +830,24 @@ def make_resource(self, data, **kwargs):
...
+    """Represent the schema of LaunchSpecificationOverrides for a Compute Resource."""
+
+    launch_template_id = fields.Str(
+        required=True,
+        metadata={"update_policy": UpdatePolicy.QUEUE_UPDATE_STRATEGY},
+    )
+    version = fields.Int(
+        validate=validate.Range(min=1),
+        metadata={"update_policy": UpdatePolicy.QUEUE_UPDATE_STRATEGY},
+    )
```
hanwen-cluster, line 844: `version` should also be `required=True`

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ CHANGELOG
 - Add support for p6-b300 instances for all OSs except AL2.
 - Add alarm on missing clustermgtd heartbeat.
 - Support updates of `Tags` during cluster-updates.
+- Add `LaunchSpecificationOverrides` to cluster config to allow network interfaces to be customized by overriding the launch template of a compute resource.
```
hanwen-cluster: Shall we explain about:
1. This overrides pcluster defaults
2. The merge is a shallow merge

tests/integration-tests/configs/develop.yaml
```diff
@@ -291,6 +291,13 @@ test-suites:
           instances: [ "p6e-gb200.36xlarge" ]
           oss: [ "ubuntu2204", "ubuntu2404", "rocky9", "alinux2023", "rhel9" ]
           schedulers: [ "slurm" ]
+  launch_template_overrides:
+    test_launch_template_overrides.py::test_launch_template_overrides:
+      dimensions:
+        - regions: [{{ c5n_18xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_YESPG_OS_X86_0 }}]
```
hanwen-cluster: We can do `NOPG`. This test doesn't care about PG

tests/integration-tests/tests/launch_template_overrides/test_launch_template_overrides.py
```diff
@@ -0,0 +1,238 @@
...
+            "SubnetId": subnet_id,
+            "Groups": [security_group_id],
+        },
+        {
+            "DeviceIndex": 1,
+            "InterfaceType": "efa-only",
+            "SubnetId": subnet_id,
+            "Groups": [security_group_id],
+        },
+    ]
```
hanwen-cluster, line 49: The interface type and device index are the same as ParallelCluster default, right? Do you think we are able to test "override"? 
I know we are verifying the security group is applied, so think it is tested properly

tests/integration-tests/configs/develop.yaml
hanwen-cluster: We only need 1 hour reservation for this test

cli/src/pcluster/validators/instances_validators.py
```diff
@@ -343,17 +343,16 @@ class LaunchTemplateValidator(Validator):
     """Validate that the specified launch template exists, is accessible, and only contains allowed properties."""
 
     # Only these properties are allowed in the launch template
-    ALLOWED_PROPERTIES = {"NetworkInterfaces", "InstanceType"}
+    ALLOWED_PROPERTIES = {"NetworkInterfaces"}
```
hanwen-cluster, line 345: Why removing `InstanceType`. If users use EC2 web console, they need it.
An alternative is to allow it and have another validator check if the Instance Type match what's in the cluster config

tests/integration-tests/tests/launch_template_overrides/test_launch_template_overrides.py
hgreebe: The test sets efa enabled = false. So this overrides it to the default but throught the override rather than with efa enabled = true.

cli/src/pcluster/validators/instances_validators.py
hgreebe: The reason I disallowed it is because there are some other problems with using the console. For example, the console does not allow you to set the subnet id of NCI0, but the parallelcluster code expects the subnet to be set.

hgreebe: So essentially, I didn't want to have a work around to make the console work when there are other problems with it.

tests/integration-tests/tests/launch_template_overrides/test_launch_template_overrides.py
hanwen-cluster: Sounds great!

cli/src/pcluster/validators/instances_validators.py
hanwen-cluster: Sounds good!

hanwen-cluster (approved):

cli/src/pcluster/schemas/cluster_schema.py
```diff
@@ -829,6 +830,31 @@ def make_resource(self, data, **kwargs):
...
+    version = fields.Int(
+        validate=validate.Range(min=1),
+        required=True,
+        metadata={"update_policy": UpdatePolicy.QUEUE_UPDATE_STRATEGY},
+    )
+
+    @validates_schema
+    def validate_version_required(self, data, **kwargs):
```
gmarciani, line 847: The version is already marked as required. why do we need this validation?

cli/src/pcluster/aws/ec2.py
```diff
@@ -683,3 +683,27 @@ def get_instance_type_and_reservation_type_from_capacity_reservation(
                 reservation_type = reservation.reservation_type()
 
         return instance_type, reservation_type
+
+    @AWSExceptionHandler.handle_client_exception
+    def describe_launch_template_version(self, launch_template_id: str, version: str):
```
gmarciani, line 688: We execute this call at least twice. What about adding caching?

cli/src/pcluster/aws/ec2.py
```diff
@@ -683,3 +683,27 @@ def get_instance_type_and_reservation_type_from_capacity_reservation(
...
+        Args:
+            launch_template_id: The launch template ID (e.g., 'lt-0abc123def456')
+            version: The version number or '$Default'/'$Latest'
+
+        Returns:
+            dict: The LaunchTemplateData from the specified version
+        """
+        response = self._client.describe_launch_template_versions(
+            LaunchTemplateId=launch_template_id,
+            Versions=[str(version)],
```
gmarciani, line 701: version is already a string, why do we need to convert it?

cli/src/pcluster/validators/instances_validators.py
```diff
@@ -335,3 +336,42 @@ def _min_max_memory(self, instance_types_info: Dict[str, InstanceTypeInfo]):
...
+            )
+            return
+
+        # Check for disallowed properties
+        disallowed_properties = set(lt_data.keys()) - self.ALLOWED_PROPERTIES
+        if disallowed_properties:
+            self._add_failure(
+                f"Launch template '{launch_template_id}' in Compute Resource '{compute_resource_name}' contains "
+                f"properties that are not allowed: {', '.join(sorted(disallowed_properties))}. "
+                f"Only NetworkInterfaces is allowed.",
```
gmarciani, line 375: We define allowed properties in a constant. What about injecting that constant in the message rather than hard wiring the property name?

cli/tests/pcluster/templates/test_cluster_stack/test_cluster_config_limits/slurm.full_config.snapshot.yaml
```diff
@@ -168,6 +168,7 @@ Scheduling:
         Gpu:
           Enabled: null
       InstanceType: c5.xlarge
+      LaunchTemplateOverrides: null
```
gmarciani, line 171: I suggest to include this feature into the test `test_cluster_config_limits` by adding a launch template and version to every compute resource and make sure we do not breach any limit

cli/src/pcluster/templates/queues_stack.py
```diff
@@ -37,6 +37,31 @@
...
+
+    For each property in the user's launch template, applies it using add_property_override.
+    This replaces the specific property while CDK handles the rest normally.
+
+    Args:
+        launch_template: The CDK CfnLaunchTemplate construct
+        compute_resource: The compute resource configuration
+    """
+    launch_template_overrides = getattr(compute_resource, "launch_template_overrides", None)
+    if not launch_template_overrides or not launch_template_overrides.launch_template_id:
```
gmarciani, line 52: We return if those are none, which is correct. What about the version field? Is there any path where if lkaunch_template_overrides is there, version is not?

hgreebe: version field is required so if the version is not there than cluster creation validation will fail

PR #7275, opened 2026-03-10
[Test] Add integ test to capture fragilities in cluster update workflow

gmarciani:
### Description of changes
Cluster updates can encounter various race conditions, especially when compute and login nodes scale up and down during the update window. 

With this PR we add a new integ test to validate that the cluster is not exposed to the following race conditions:
1. **[Race Condition 1]** A node that completes the bootstrap after the update is not able to execute the update. We know this could happen when DNA files are removed at the end of a successful update.
2. **[Race Condition 2]** A node that completes the bootstrap close to the end of the readiness check can cause the readiness check to fail because it does not have enough time to complete the update.
3. **[Race Condition 3]** A node that faces a transient failure in the execution of the update does not retry the update.
4. **[Race Condition 4]** A node that faces a transient failure in the detection of the update does not retry the update.
5. **[Race Condition 5]** A login node that is started before the update but completes the config phase during the update will skip the update. This happens because cfn-hup will see the updated launch template and will think thereafter that it has deployed the updated config, even if it did not.

**Important Note**
Race conditions from 1 to 4 have been fixed in 3.15.0. Race condition 5 is not, so it would be a known issue causing the test failure. Instead of failing the test, we preferred to log the failure and let the test succeed. In this way we know that, if the test fails, it fails for unexpected reasons.

Additional improvements to common functionalities:
1. added the concept of cluster snapshot, which is a snapshot of `describe-cluster-instances` + some important diagnostic information for each cluster node, such as:
    * config version deployed on the node, according to its local dna
    * config version deployed on the node, according to DDB
    * timestamped order of execution of chef recipes
1. Add utility to wait for a set of acceptable stack statuses
3. Allow to search for patterns in logs with a timestamp constraint
4. Allow to retrieve dna.json from on login nodes
7. Allow to retrieve cluster config deployed on login nodes

### Tests
SUCCESS `test_update_race_conditions`.

### DIFF
--- a/tests/integration-tests/clusters_factory.py
+++ b/tests/integration-tests/clusters_factory.py
@@ -229,6 +229,15 @@ class Cluster:
             retry_on_result=lambda result: result["clusterStatus"] != expected_status,
         )(self.describe_cluster)()
 
+    def wait_cluster_stack_status(self, expected_statuses: list, stop_max_delay_minute: int = 15):
+        """Wait for the cluster stack to reach the desired status."""
+        retry(
+            wait_fixed=seconds(10),
+            stop_max_delay=minutes(stop_max_delay_minute),
+            retry_on_result=lambda result: result["cloudFormationStackStatus"] not in expected_statuses,
+        )(self.describe_cluster)()
+        return self.describe_cluster()["cloudFormationStackStatus"]
+
     def describe_compute_fleet(self):
         """Run pcluster describe-compute-fleet and return the result."""
         cmd_args = ["pcluster", "describe-compute-fleet", "--cluster-name", self.name]
--- a/tests/integration-tests/tests/common/utils.py
+++ b/tests/integration-tests/tests/common/utils.py
@@ -479,27 +479,33 @@ def wait_process_completion(remote_command_executor, pid):
         return result.stdout.strip()
 
 
-def get_deployed_config_version(cluster, compute_node_ip: str = None):
-    """Retrieves the cluster config version deployed on the cluster node from its dna.json
-    If 'compute_node_ip' is specified, the config version will be retrieved from the compute node;
-    otherwise, it will be retrieved from the head node.
+def get_deployed_config_version(cluster, compute_node_ip: str = None, login_node_ip: str = None, bastion: str = None):
+    """Retrieves the cluster config version deployed on a cluster node from its dna.json.
+
+    If 'compute_node_ip' is specified, the config version is retrieved from that compute node.
+    If 'login_node_ip' and 'bastion' are specified, it is retrieved from that login node via the bastion.
+    Otherwise, it is retrieved from the head node.
     """
-    dna_json = get_deployed_dna_json(cluster, compute_node_ip)
+    dna_json = get_deployed_dna_json(cluster, compute_node_ip, login_node_ip, bastion)
 
     return dna_json["cluster"]["cluster_config_version"]
 
 
-def get_deployed_dna_json(cluster, compute_node_ip: str = None):
-    """Retrieves the dna.json from the cluster node
-    If 'compute_node_ip' is specified, it will be retrieved from the compute node;
-    otherwise, it will be retrieved from the head node.
+def get_deployed_dna_json(cluster, compute_node_ip: str = None, login_node_ip: str = None, bastion: str = None):
+    """Retrieve the dna.json deployed on a cluster node via SSH.
+
+    By default the dna.json is retrieved from the head node.
+    If 'compute_node_ip' is specified, it will be retrieved from that compute node.
+    If 'login_node_ip' is specified, it will be retrieved from that login node.
+
+    Returns the parsed dna.json as a dict.
+
+    Raises:
+        RuntimeError: if the remote command fails (e.g. node is unreachable).
+        ValueError: if the response is not valid JSON or does not contain the expected 'cluster' key.
     """
     command = "sudo cat /etc/chef/dna.json"
-    rce = (
-        RemoteCommandExecutor(cluster, compute_node_ip=compute_node_ip)
-        if compute_node_ip
-        else RemoteCommandExecutor(cluster)
-    )
+    rce = RemoteCommandExecutor(cluster, compute_node_ip=compute_node_ip, login_node_ip=login_node_ip, bastion=bastion)
 
     try:
         result = rce.run_remote_command(command).stdout
@@ -522,6 +528,21 @@ def get_ddb_item(region_name: str, table_name: str, item_key: dict):
     """
     table = boto3.resource("dynamodb", region_name=region_name).Table(table_name)
     return table.get_item(Key=item_key).get("Item")
+
+
+def get_config_version_from_ddb(region, cluster_name, instance_id):
+    """Get the current cluster config version from DynamoDB."""
+    dynamodb = boto3.client("dynamodb", region_name=region)
+    table_name = f"parallelcluster-{cluster_name}"
+    ddb_key = f"CLUSTER_CONFIG.{instance_id}"
+
+    response = dynamodb.get_item(TableName=table_name, Key={"Id": {"S": ddb_key}})
+
+    if "Item" in response:
+        data = response["Item"].get("Data", {}).get("M", {})
+        return data.get("cluster_config_version", {}).get("S", "")
+
+    raise ValueError(f"No DynamoDB record found for instance {instance_id}")
 
 
 def get_compute_ip_to_num_files(remote_command_executor, slurm_commands):
--- a/tests/integration-tests/tests/update/test_update_race_conditions.py
+++ b/tests/integration-tests/tests/update/test_update_race_conditions.py
@@ -1,0 +1,549 @@
+# Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import json
+import logging
+import time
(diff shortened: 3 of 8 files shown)


tests/integration-tests/conftest.py
```diff
@@ -328,19 +333,19 @@
...
+#     try:
+#         region = session.config.getoption("regions") or get_all_regions(session.config.getoption("tests_config"))
+#         region = [unmarshal_az_override(az) for az in region]
+#         # Use the first element of the list of regions, since there must be at least one
+#         reporting_region = get_reporting_region(region[0])
+#         logging.info(f"Metadata reporting region {reporting_region}")
+#         # Setup the metadata table in case it doesn't exist
+#         MetadataTableManager(reporting_region, METADATA_TABLE).create_metadata_table()
+#     except Exception as exc:
+#         logging.info(f"There was a '{type(exc)}' error with '{exc}' when creating the table!")
```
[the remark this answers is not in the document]
gmarciani: This commit is just for testing and will be removed

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,584 @@
...
+    """
+    from datetime import datetime, timezone
+
+    target_status = "UPDATE_COMPLETE"
+    after_dt = datetime.strptime(after_utc, "%Y-%m-%d %H:%M:%S").replace(tzinfo=timezone.utc) if after_utc else None
+    client = boto3.client("cloudformation", region_name=region)
+    paginator = client.get_paginator("describe_stack_events")
+
+    # First, collect all events matching the logical resource id criteria
+    matching_events = []
```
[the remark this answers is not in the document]
gmarciani: Fixed

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,550 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import datetime
```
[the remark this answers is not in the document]
gmarciani: false positive

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,561 @@
...
+import json
```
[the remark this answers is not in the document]
gmarciani: Fixed

tests/integration-tests/tests/common/utils.py
```diff
@@ -479,27 +479,23 @@ def wait_process_completion(remote_command_executor, pid):
...
-    dna_json = get_deployed_dna_json(cluster, compute_node_ip)
+    dna_json = get_deployed_dna_json(cluster, compute_node_ip, login_node_ip)
 
     return dna_json["cluster"]["cluster_config_version"]
 
 
-def get_deployed_dna_json(cluster, compute_node_ip: str = None):
+def get_deployed_dna_json(cluster, compute_node_ip: str = None, login_node_ip: str = None):
     """Retrieves the dna.json from the cluster node
     If 'compute_node_ip' is specified, it will be retrieved from the compute node;
```
hgreebe: Could you update this description to include what is done with the login node ip?

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,411 @@
...
+    Test cluster update readiness check issues.
+
+    Two updates are performed:
+    - Update 1: Tests Issue 1. A dynamic node is bootstrapping during the update.
+      The update succeeds (blocked node is ignored). After the update, the node is
+      unblocked and verified to be stuck (unable to apply the update).
+    - Update 2: Tests Issues 2-4. All failure conditions are injected before the
+      update. The update is expected to fail. Logs are checked to confirm each issue.
+
+    Issues tested:
```
hgreebe: Can you expand the description a bit to explain why these things are a problem.
For example, 
Issue 1: DNA file deletion during active bootstrap. CN get stuck waiting for dna json file
...
Issue 4: cfn-hup stores updated LT instead of bootstrap LT on login node. LN does not see update and stays on old version

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,411 @@
...
+    clusters_factory,
+    test_datadir,
+    scheduler_commands_factory,
+    s3_bucket_factory,
+):
+    """
+    Test cluster update readiness check issues.
+
+    Two updates are performed:
+    - Update 1: Tests Issue 1. A dynamic node is bootstrapping during the update.
```
hgreebe: I find this line confusing: 
>  After the update, the node is unblocked and verified to be stuck

Since this has been fixed, aren't we validating that the node is not stuck?

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,411 @@
...
+    logger.info(f"Unblocking detection of update for compute node {CN_4}")
+    unblock_update_detection_on_compute_node(rce, cluster, CN_4)
+    logger.info(f"Unblocked detection of update for compute node {CN_4}")
+
+    logger.info("Waiting for the cluster stack to reach a final state")
+    #TODO Once we fix all the race condition, the only valid state will be UPDATE_COMPLETE
+    # We are leaving the other failed state for now so that we can use the verifications
+    # at the end of the test to facilitate troubleshooting
+    actual_cluster_stack_status = cluster.wait_cluster_stack_status(
+        expected_statues=["UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE", "UPDATE_ROLLBACK_FAILED"],
```
hgreebe: So after our changes, all issues should be fixed except for issue 4 right?

tests/integration-tests/tests/update/test_update_race_conditions.py
gmarciani: Revise the entire description to reflect the current scope

tests/integration-tests/tests/common/utils.py
gmarciani: Done

tests/integration-tests/tests/update/test_update_race_conditions.py
gmarciani: Was an old comment, thanks for spotting it. removed.

tests/integration-tests/tests/update/test_update_race_conditions.py
gmarciani: Correct. The last issue is about login nodes where cfnhup thinks it has deployed the correct config version. That fix is a nice to have for 3.15.0

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,485 @@
...
+    cluster = clusters_factory(config_file_1)
+    rce = RemoteCommandExecutor(cluster)
+
+    logger.info("MGIACOMO START ================================================================================")
+
+    cluster_snapshot = get_cluster_nodes_snapshot(cluster)
+    logger.info(f"Cluster snapshot:\n{json.dumps(cluster_snapshot, indent=2)}")
+    return
+
+    logger.info("MGIACOMO END ================================================================================")
```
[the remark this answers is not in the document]
gmarciani: Fixed

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -0,0 +1,485 @@
...
+    )
+
+    # Create cluster
+    config_file_1 = pcluster_config_reader(
+        output_file="pcluster.config.1.yaml",
+        login_nodes_count=1,
+        **common_cluster_config_args,
+    )
+    cluster = clusters_factory(config_file_1)
+    rce = RemoteCommandExecutor(cluster)
```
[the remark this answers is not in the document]
gmarciani: Fixedf

tests/integration-tests/clusters_factory.py
```diff
@@ -229,6 +229,15 @@ def wait_cluster_status(self, expected_status, stop_max_delay_minute=15):
             retry_on_result=lambda result: result["clusterStatus"] != expected_status,
         )(self.describe_cluster)()
 
+    def wait_cluster_stack_status(self, expected_statues: list, stop_max_delay_minute: int = 15):
```
hgreebe: is this supposed to be `expected_statuses`?

tests/integration-tests/tests/common/utils.py
```diff
@@ -479,27 +479,31 @@ def wait_process_completion(remote_command_executor, pid):
         return result.stdout.strip()
 
 
-def get_deployed_config_version(cluster, compute_node_ip: str = None):
+def get_deployed_config_version(cluster, compute_node_ip: str = None, login_node_ip: str = None, bastion: str = None):
     """Retrieves the cluster config version deployed on the cluster node from its dna.json
```
hgreebe: could you update this description to include `If 'login_node_ip' is specified`

tests/integration-tests/clusters_factory.py
gmarciani: It is expected. 

It is a "trick" to let the test collect more info in case the cluster fails the update.
We expect the cluster to succeed the update (status UPDATE_COMPLETE). 
However, if we specify here only UPDATE_COMPLETE and the update fails, the test will exit and we will not have all the important information collected by the soft assertions. In this way instead, we proceed in case of update failure and then we assert on the stack status and many other things.

tests/integration-tests/tests/common/utils.py
gmarciani: Fixed

tests/integration-tests/clusters_factory.py
gmarciani: (Typo fixed)

hgreebe (approved):

hgreebe (approved):

PR #7281, opened 2026-03-11
[integ-tests][OSU] Use rolling window for benchmark failure detection

hanwen-cluster:
### Description of changes
Replace the fixed threshold count (3 for collective, 4 for pt2pt) with a rolling window approach: a benchmark is flagged as failed only if 3 out of any 4 consecutive packet sizes exceed the tolerance.

The old method counted total failures across all packet sizes, treating scattered failures the same as clustered ones. Since benchmark characteristics vary by packet-size range, scattered failures are more likely noise while clustered failures indicate a real regression in a specific range.

Simulated against the 5 most recent DynamoDB runs with the new p75 thresholds:

Old detection: 10/30 test failures (33.3%)
Rolling window 4/3: 8/30 test failures (26.7%)
The 2 eliminated false positives had scattered failures across unrelated packet sizes. The 8 remaining failures all have dense clusters (e.g. 3+ consecutive sizes at ps 4-32 in osu_allreduce), which are stronger signals of actual issues.

### Tests
tests on develop.yaml have passed

### Checklist

### DIFF
--- a/tests/integration-tests/tests/performance_tests/test_osu.py
+++ b/tests/integration-tests/tests/performance_tests/test_osu.py
@@ -160,8 +160,6 @@ def _test_osu_benchmarks_pt2pt(
     # OSU pt2pt benchmarks cannot be executed with more than 2 MPI ranks.
     # Run them in 2 instances with 1 proc per instance, defined by map-by parameter.
     num_instances = 2
-    # Accept a max number of 4 failures on a total of 23-24 packet size tests.
-    accepted_number_of_failures = 4
 
     failed_benchmarks = []
     benchmark_group = "pt2pt"
@@ -178,10 +176,10 @@ def _test_osu_benchmarks_pt2pt(
             network_interfaces_count,
             test_datadir,
         )
-        failures = _check_osu_benchmarks_results(
+        failure_mask = _check_osu_benchmarks_results(
             test_datadir, output_dir, os, instance, mpi_version, benchmark_name, num_instances, output
         )
-        if failures > accepted_number_of_failures:
+        if _has_failure_cluster(failure_mask):
             failed_benchmarks.append(f"{mpi_version}-{benchmark_name}")
 
     return failed_benchmarks
@@ -200,9 +198,6 @@ def _test_osu_benchmarks_collective(
     slots_per_instance,
     partition=None,
 ):
-    # Accept a max number of 3 failures on a total of 19-21 packet size tests.
-    accepted_number_of_failures = 3
-
     failed_benchmarks = []
     benchmark_group = "collective"
     benchmark_names = ["osu_allgather", "osu_bcast", "osu_allreduce", "osu_barrier"]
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -178,10 +176,10 @@ def _test_osu_benchmarks_pt2pt(
             network_interfaces_count,
             test_datadir,
         )
-        failures = _check_osu_benchmarks_results(
+        failure_mask = _check_osu_benchmarks_results(
```
gmarciani, line 179: [nitpick] failure mask is misleading. Those are actual failures. We then process them to identify clustered failures

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -366,4 +362,25 @@ def _check_osu_benchmarks_results(
...
+    Instead of counting total failures across all packet sizes, look for clusters
+    of failures in consecutive packet sizes. This catches real regressions in a
+    specific packet-size range while ignoring scattered noise.
+
+    A benchmark is considered failed if any window of `window_size` consecutive
+    packet sizes contains at least `min_failures` failures.
+    """
+    if not failure_mask:
+        return False
+    if len(failure_mask) < window_size:
```
gmarciani, line 381: if the number of failures is less than the window, then it should return False.

tests/integration-tests/tests/performance_tests/test_osu.py
```diff
@@ -366,4 +362,25 @@ def _check_osu_benchmarks_results(
...
+        return sum(failure_mask) >= min(min_failures, len(failure_mask))
+    for i in range(len(failure_mask) - window_size + 1):
```
gmarciani, line 383: this logic assumes that the content of failure_mask is an array of elements 0,1 where 0 is success and 1 is failure, but I think this is not what _check_osu_benchmarks_results returns. Can we double check?

tests/integration-tests/tests/performance_tests/test_osu.py
hanwen-cluster: The structure of failure_mask is [0,1,1,0,0]. So it is indeed masks not real failures

tests/integration-tests/tests/performance_tests/test_osu.py
hanwen-cluster: Here is a corner case, when tests run on fewer packet sizes than the window size. For example, barrier test only has one entry, one failure should be reported

tests/integration-tests/tests/performance_tests/test_osu.py
hanwen-cluster: Logic is here https://github.com/aws/aws-parallelcluster/pull/7281/changes#diff-60b99ffc235e925350dc9f5ce1b9580076a5e5910a4638b037c68810fc3780afR353

gmarciani: Got it, so indeed it is a failure mask :-D

gmarciani (approved):

PR #7302, opened 2026-03-18
[Test] Prevent false positive failures in test  `test_update_race_condition`. 

gmarciani:
### Description of changes
Prevent false positive failures in test  `test_update_race_condition`. In particular:
1. do not fail when the cluster update fails only because of the known race condition on login node.
2. when cluster fails the update, determine the cluster snapshot used by the assertions only after the head node completed the update recipe, so that the snapshot contains reliable data.

### Tests
SUCCEEDED `test_update_race_conditions` on all OSs except for AL2 where there is a known issue about a logging file we need to fix in the product


tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -375,32 +388,46 @@ def assert_update_recipe_executed_on_old_nodes(snapshot, update_submitted_at, ex
...
     Returns False if there are no wrong records or if other instance IDs are present.
     """
-    found, output = match_regex_in_log(
+    found, lines = match_regex_in_log(
         rce,
         "/var/log/chef-client.log",
-        r"wrong records \(1\):.*",
+        r"Retrying execution of execute\[Check cluster readiness\], 0 attempt left",
         after_utc=after_utc,
+        nlines_after_match=50,
```
hehe7318, line 402: There is a mismatch.
In the comment: `Retrieves the last 100 lines`
Here: `nlines_after_match=50,`

hehe7318 (approved):

PR #7303, opened 2026-03-19
[Image] Add more verbosity when downloading Cookbook

himani2411:
### Description of changes
* Use the write-out flag to capture just the HTTP status code. On non-200, print the response body (which will be the S3 XML error like AccessDenied or ExpiredToken)
* Use `-sS` option to silence (no progress bar) but still prints curl errors like curl: `(6) Could not resolve host or curl: (28) Connection timed out.` Covers both transport failures and HTTP failure.
```
Cookbook download attempt 1 failed with HTTP 403. Response body:
2026-03-20 17:17:06.371000	Stdout: <?xml version="1.0" encoding="UTF-8"?>
2026-03-20 17:17:06.372000	Stdout: <Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>N9ANSFDAXHB1Z4NN</RequestId><HostId>6+XkwyfEcE8tbfrtF5HsHlw3YNVvAFBTpVwH6yhgnnSu5Zi0rxrBiiRUIna6FIpfKR4/g0+9Hgo=</HostId></Error>
2026-03-20 17:17:07.424000	Stdout: Cookbook download attempt 2 failed with HTTP 403. Response body:
2026-03-20 17:17:07.425000	Stdout: <?xml version="1.0" encoding="UTF-8"?>
2026-03-20 17:17:07.425000	Stdout: <Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>6KHWXEPXMTHKPZRM</RequestId><HostId>Q1dOT6L/MzAU44FaWHAS1GNovHW+7rZtpgqfRvyDNcucrUfN++L0ZF1BygmzPFn/KXZyQmEdgd394mWMHVlF5vfzzmQMukzK</HostId></Error>
2026-03-20 17:17:09.471000	Stdout: Cookbook download attempt 3 failed with HTTP 403. Response body:
2026-03-20 17:17:09.471000	Stdout: <?xml version="1.0" encoding="UTF-8"?>
```

### Tests
* Build image with above failure

### References

### Checklist

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
@@ -221,8 +221,14 @@ phases:
               mkdir -p /etc/chef && sudo chown -R root:root /etc/chef
               delays=(1 2 4 8 16 32 64 128 256)
               for i in {0..8}; do
-                curl -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL" && break
-                echo "Curl attempt $((i+1)) failed, retrying in ${!delays[$i]}s..."
+                HTTP_CODE=$(curl -sS -L -w '%{http_code}' -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL")
+                if [[ "$HTTP_CODE" -eq 200 ]]; then
+                  echo "Cookbook download succeeded (HTTP $HTTP_CODE)"
+                  break
+                fi
+                echo "Cookbook download attempt $((i+1)) failed with HTTP $HTTP_CODE. Response body:"
+                cat /etc/chef/aws-parallelcluster-cookbook.tgz
+                echo ""
                 sleep ${!delays[$i]}
               done
               


cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -221,8 +221,14 @@ phases:
               mkdir -p /etc/chef && sudo chown -R root:root /etc/chef
               delays=(1 2 4 8 16 32 64 128 256)
               for i in {0..8}; do
-                curl -L -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL" && break
-                echo "Curl attempt $((i+1)) failed, retrying in ${!delays[$i]}s..."
+                HTTP_CODE=$(curl -s -L -w '%{http_code}' -o /etc/chef/aws-parallelcluster-cookbook.tgz "$COOKBOOK_URL")
```
gmarciani: are we silencing now the output of curl (-s)  because it would otherwise prevent the capturing of the error code? What if curl fails? Would we be able to understand the failure?

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
```diff
@@ -221,8 +221,14 @@ phases:
...
+                if [[ "$HTTP_CODE" -eq 200 ]]; then
+                  echo "Cookbook download succeeded (HTTP $HTTP_CODE)"
+                  break
+                fi
+                echo "Cookbook download attempt $((i+1)) failed with HTTP $HTTP_CODE. Response body:"
```
gmarciani, line 229: If the curl returns a error code (say 403) we are still trying to untar the file. What about untarring the file only if the return code is 200?

himani2411: The loop breaks when we get HTTP_CODE of 200, and then we untar the file at line 235

We are doing `cat /etc/chef/aws-parallelcluster-cookbook.tgz` which is supposed to be the S3 XML file rather than a tgz file only when we do not see 200 code

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
himani2411: Good catch! will update

himani2411: Update to add `-sS` which should show the curl failure message but not the progress meter

https://curl.se/docs/manpage.html
> Use [--show-error](https://curl.se/docs/manpage.html#--show-error) in addition to this option to disable progress meter but still show error messages

cli/src/pcluster/resources/imagebuilder/parallelcluster.yaml
himani2411: Per offline discussion, we'll keep the untar step outside the loop since the build image step already exits on the untar failure. Adding extra parameters to handle the exit within the loop isn't worth the template size increase.

gmarciani (approved):

PR #7304, opened 2026-03-23
[Test] Make the test `test_update_race_condition` able to execute on AL2.

gmarciani:
### Description of changes
Make the test `test_update_race_condition` able to execute on AL2.

The test was assuming that pclustrer-check-update logs where pushed to /var/log/parallecluster/pcluster-check-update.log. On AL2 only, this is not the case: logs are pushed to /var/log/messages. In order to read /var/log./messages sudo permissions are required.

### Tests
SUCCESS `test_update_race_conditions` on AL2

### DIFF
--- a/tests/integration-tests/utils.py
+++ b/tests/integration-tests/utils.py
@@ -1183,7 +1183,7 @@ def match_regex_in_log(
             - str: If found, the full line containing the match plus the next `nlines_after_match` lines.
                    If not found, the last `nlines` lines of the log.
     """
-    result = rce.run_remote_command(f"cat {log_file}")
+    result = rce.run_remote_command(f"sudo cat {log_file}")
     log_content = result.stdout
 
     if after_utc:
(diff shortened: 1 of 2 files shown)


hehe7318 (approved):

tests/integration-tests/utils.py
```diff
@@ -1183,7 +1183,7 @@ def match_regex_in_log(
             - str: If found, the full line containing the match plus the next `nlines_after_match` lines.
                    If not found, the last `nlines` lines of the log.
     """
-    result = rce.run_remote_command(f"cat {log_file}")
+    result = rce.run_remote_command(f"sudo cat {log_file}")
```
hehe7318, line 1186: Why we need sudo permission here?

gmarciani: Because some critical logs can only be read by root. Example: /var/log/messages

PR #7312, opened 2026-03-25
[integ-tests-framework] Use Lazy import of dependencies in reports_generator

hanwen-cluster:
### Description of changes
Sometimes, only a few functions in reports_generator.py are called by a environment, which doesn't have to have all Python dependencies installed

### Tests
* test will run after the PR is merge

### Checklist

### DIFF
--- a/tests/integration-tests/framework/tests_configuration/config_renderer.py
+++ b/tests/integration-tests/framework/tests_configuration/config_renderer.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 import logging
 import os
+import random
 import re
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


tests/integration-tests/reports_generator.py
```diff
@@ -18,14 +18,9 @@
 from typing import List
 
 import boto3
-import pandas as pd
```
himani2411, line 21: why lazy import is required when we specify these dependencies in https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/requirements.txt?

gmarciani:
This is a **short term solution** to prevent alarms. 
The long term solution is to remove the dependency on SUPPORTED_OS constant and install pandas and other dependencies in Jenkins.

PR #7322, opened 2026-03-30
[Test] Improve debuggability, stability and coverage of `test_dcv_configuration` and `test_dcv_remote_access`

gmarciani:
### Description of changes
Improve stability, debuggability and coverage of `test_dcv_configuration` and `test_dcv_remote_access`
* debuggability: retrieve, print and analyze a comprehensive report of crashes (not only the crash filename, but the stack trace of the crash).  Also, moved from hard assertions to soft assertions to have a final report of all the observed failures.
* stability: prevent false positive failures, by ignoring harmless crashes related to gnome, unrelated to nvidia or dcv. Also fixed a gap that was causing failures when multiple instances of this test are executed in parallel by serializing the modifications to ssh known_hosts.
* coverage: the test is now able to detect crashes on all supported OSs, not only Ubuntu.

### Tests
1. `test_dcv_configuration` and `test_dcv_remote_access` executed on all OSs on both g5g.2xlarge and c5.xlarge. The test now fails only for relevant crashes reporting detailed information about those. The only relevant crash detected is related to an issue with DCV, so it is expected to report it.


himani2411:
[Non-blocking] Is it possible to keep this crash report irrespective of the DCV test i.e can we diagnose all the integ test clusters to run this crash report irrespective of which test we are running?

This can generate false positives (some of which you have already identified) during our test runs. The advantage i see is that we will be aware of other components which could cause crashes in our product.

This same script can be extended to detect Slurm related Coredumps  (/var/log, /var/spool/slurmd, /var/tmp/ or var/spool/abrt) and generate Backtrace

[1] https://slurm.schedmd.com/faq.html#backtrace
[2] https://slurm.schedmd.com/faq.html#core_dump



This can be de-scoped to another PR.

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -138,9 +186,40 @@ def _check_auth_ok(remote_command_executor, external_authenticator_port, session
...
+    """Check for crash files on the node and return a crash report dictionary.
+
+    Runs a script that scans crash locations across all pcluster-supported OSes
+    (Ubuntu, AL2, AL2023, RHEL8/9, Rocky8/9) and returns a JSON dictionary
+    mapping crash file paths to their human-readable content.
+
+    Returns an empty dict if no crashes found.
+    """
+    result = remote_command_executor.run_remote_script(str(DIAGNOSIS_SCRIPT_DIR / "get_crash_report.sh"), pty=False)
+    return json.loads(result.stdout)
```
himani2411: Can we wrap json.loads in a try/except and provide a meaningful error message including the raw stdout so that if there is a failure due to the script we know what the error is?

tests/integration-tests/tests/dcv/test_dcv.py
```diff
@@ -165,6 +245,24 @@ def _check_error_cases(remote_command_executor, dcv_authenticator_port):
...
     )
+    logging.info("Completed checks for authentication failure on %s", remote_command_executor.target)
+
+
+@contextlib.contextmanager
+def _temporary_known_host(hostname, host_keys_file, env):
+    """Add SSH host keys for hostname, yield, then remove them. Serialized via file lock across processes."""
+    lock_file = host_keys_file + ".lock"
+    with open(lock_file, "w") as lf:
+        fcntl.flock(lf, fcntl.LOCK_EX)
```
himani2411, line 270: We are adding a lock on a file which could be written by other test processes? Do we know any other tests which modifies this file?

gmarciani: Only the DCV tests write on this file. 
When multiple executions of the DCV tests are run in parallel, they interfere with each, causing intermittent integ test failure due to SSH connection errors.
The locking mechanism is to prevent this interference.

gmarciani:
> [Non-blocking] Is it possible to keep this crash report irrespective of the DCV test i.e can we diagnose all the integ test clusters to run this crash report irrespective of which test we are running?
> 
> This can generate false positives (some of which you have already identified) during our test runs. The advantage i see is that we will be aware of other components which could cause crashes in our product.
> 
> This same script can be extended to detect Slurm related Coredumps (/var/log, /var/spool/slurmd, /var/tmp/ or var/spool/abrt) and generate Backtrace
> 
> [1] https://slurm.schedmd.com/faq.html#backtrace [2] https://slurm.schedmd.com/faq.html#core_dump
> 
> This can be de-scoped to another PR.

This is already possible. The logic to build the crash report is fully contained into the script get_crash_report.sh introduced in this PR. That script does not depend on the test itself. when we will work on the diagnosis script, we will include this script in the tollset.

tests/integration-tests/tests/dcv/test_dcv.py
gmarciani: Good catch. done.

tests/integration-tests/tests/dcv/test_dcv.py
himani2411: Rephrasing the question, since this file is updated manually in the test_dcv but we could be doing other SSH commands in the entire test suite which leads to an automatic update of this file by ssh-client, for ex when we ssh to a client for the 1st time. Will we block this automatic update (i am assuming it doesnt but lets confirm)?

tests/integration-tests/tests/common/diagnosis/get_crash_report.sh
```diff
@@ -0,0 +1,215 @@
...
+    current_key = None
+    for raw in open(path, 'rb'):
+        line = raw.decode('utf-8', errors='replace')
+        if not line.startswith(' ') and ':' in line:
+            key, _, val = line.partition(':')
+            current_key = key.strip()
+            fields[current_key] = val.strip()
+        elif current_key and line.startswith(' '):
+            fields[current_key] += '\n' + line.rstrip()
+    keys = ['ProblemType','Package','ExecutablePath','Signal','Traceback',
```
himani2411: [Non-blocking] There is minor duplication of these keys and the formating of keys at the end of the function which can be addressed later.

tests/integration-tests/tests/dcv/test_dcv.py
gmarciani: It does not block

tests/integration-tests/tests/common/diagnosis/get_crash_report.sh
gmarciani: addressed

himani2411 (approved):

PR #7324, opened 2026-04-01
[Integ] Add integration test for Pcluster Build Image in proxied environment 

himani2411:
### Description of changes
Adds integration test which creates a proxied environment for testing ParallelCluster Build Image.
 Updated proxy.yaml by introducing EnableBuildImageProxy to distinguish build-time from cluster-time proxy configurations. Key changes include:

* Expands proxy allowlist for AWS services (CloudFront, S3), FSx repos, EFA installer, EFS-utils dependencies (Rust and Index) and OS repos.
* Adds snap store allowlist to resolve dpkg lock issues during DCV installation
* Grants AmazonS3ReadOnlyAccess for Node package presigned URL generation in cookbook recipe and use install_proxy_url Devsetting added as part of cookbook PR https://github.com/aws/aws-parallelcluster-cookbook/pull/3157
* We skip awsbatch CLI installation for this test as we will be deprecating this dependency. 
* Enables IP forwarding to prevent kernel packet drops in ConfigureSystem step
* Switches from GitHub to S3 URLs for aws-parallelcluster-cookbook downloads to avoid SSL connection failures through proxy

### Tests
* Successful Build image with ALINUX 2023, Rhel9/8, Rocky9/8 and Ubuntu22/24
* Integ test 
```
   proxy:
    test_proxy.py::test_proxy:
      dimensions:
        - regions: ["us-east-1"] # c5 instance type is not available in ap-southeast-5
          instances: ["c5.xlarge"]
          oss: [{{ OS_X86_7 }}]
          schedulers: ["slurm"]
```
### References

### Checklist


tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -12,6 +12,7 @@
 import datetime
 import json
 import logging
+import os
```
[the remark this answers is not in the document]
himani2411: Addressed

tests/integration-tests/tests/createami/test_createami/test_build_image_no_internet/image.config.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+Build:
+    InstanceType: {{ instance }}
+    ParentImage: {{ parent_image }}
+    SecurityGroupIds:
+        - {{ security_group_id }}
+    SubnetId: {{ subnet_id }}
+    Iam:
+        AdditionalIamPolicies:
+            - Policy: arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
```
gmarciani: We should inject the partition dynamically

cloudformation/proxy/proxy.yaml
```diff
@@ -121,6 +139,83 @@ Resources:
       InstanceId: !Ref Proxy
       RouteTableId: !Ref PrivateSubnetRT
 
+  # VPC ENDPOINTS (build-image mode only)
+  S3Endpoint:
```
gmarciani, line 143: I was imagining that an S3 endpoint was required also when deploying a cluster, not only on build-image. What makes this S3 endpoint different for the build image?

Same for logs, ec2, ssm

cloudformation/proxy/proxy.yaml
```diff
@@ -347,9 +497,11 @@ Resources:
 
   ProxyVerificationWaitConditionHandle:
     Type: AWS::CloudFormation::WaitConditionHandle
+    Condition: IsClusterProxy
 
   ProxyVerificationWaitCondition:
     Type: AWS::CloudFormation::WaitCondition
+    Condition: IsClusterProxy
```
gmarciani: Why not having the proxy verification ion the case of a proxy meant for build image?

cloudformation/proxy/proxy.yaml
```diff
@@ -297,9 +446,10 @@ Resources:
       SourceDestCheck: false
       SubnetId: !Ref PublicSubnet
 
-  # PROXY Client
+  # PROXY CLIENT (cluster proxy mode only)
```
gmarciani: This instances is meant to test the proxy functionality.
Why not keeping this verification for the build image case as well?

tests/integration-tests/tests/createami/test_createami/test_build_image_no_internet/image.config.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+DevSettings:
+    TerminateInstanceOnFailure: True
+    Cookbook:
+        ChefCookbook: {{ chef_cookbook }}
+        ExtraChefAttributes: '{"cluster": {"skip_awsbatch_cli_install": "true", "install_http_proxy_address": "{{ install_http_proxy_address }}"}}'
```
gmarciani, line 20: Why skip_awsbatch_cli_install ?

tests/integration-tests/tests/common/utils.py
```diff
@@ -652,3 +652,44 @@ def get_capacity_reservation_id(request, instance_type, region, count, os):
                         }
                     )
     return reservations_ids
+
+
+def upload_github_artifacts_to_s3(bucket_name, region, request):
```
gmarciani, line 668: What about wrapping this upload with retries to prevent test failures on temporary networking failures?

tests/integration-tests/tests/createami/test_createami/test_build_image_no_internet/image.config.yaml
himani2411: Updated! Thanks for catching this.

tests/integration-tests/tests/proxy/test_proxy.py
```diff
@@ -30,40 +30,63 @@ def proxy_stack_factory(region, request, cfn_stacks_factory):
...
+                    name=request.config.getoption("proxy_stack"),
+                    region=region,
+                    template=template_body,
+                )
+            else:
+                stack_name = generate_stack_name("integ-tests-proxy", request.config.getoption("stackname_suffix"))
+                stack_parameters = [
+                    {"ParameterKey": "Keypair", "ParameterValue": request.config.getoption("key_name")},
+                    {"ParameterKey": "VpcCidr", "ParameterValue": "10.0.0.0/16"},
+                    {"ParameterKey": "SSHCidr", "ParameterValue": "0.0.0.0/0"},
```
gmarciani: Not introduced by this PR, so we can address in a different PR, but this is allowing global access to SSH port and could cause alarms. What about restricting the access to the VPC CIDR instead?

tests/integration-tests/tests/createami/test_createami/test_build_image_no_internet/image.config.yaml
himani2411: I am removing this check since we will be deprecating this dependency and it does not make sense to test it now

cloudformation/proxy/proxy.yaml
himani2411: In the cluster mode, cluster nodes still has internet access. For Build Image I am blocking any internet access from the build instance. So if we want to add a verification I would have to do the same setup of HTTP ENV setup as we do in Cookbook PR for this verification step. In order to verify I would also have to hardcode the repo and dependencies URL's which we will have to keep updating in future if we add a new repo or dependency. I dont see the  value of modifying this verification step just for an earlier failure.

cloudformation/proxy/proxy.yaml
himani2411: Answered https://github.com/aws/aws-parallelcluster/pull/7324#discussion_r3030418214

cloudformation/proxy/proxy.yaml
himani2411: The proxy setup we have for the cluster nodes does not block internet access. So they can access these endpoints.

For Build Image I am blocking Internet access from the Build Image that Image builder will launch.

tests/integration-tests/tests/common/utils.py
himani2411: Good idea will update

cloudformation/proxy/proxy.yaml
himani2411: As discussed offline, the simple verification of `yum install` should be enough to see that Proxy setup is correct rather than calling each endpoint of FSX repos.

cloudformation/proxy/proxy.yaml
```diff
@@ -320,21 +469,30 @@ Resources:
...
+              echo "==> Testing HTTPS proxy (same as build instance uses via https_proxy env var)"
+              https_proxy="http://${ProxyPrivateIp}:${ProxyPort}" curl -v -o /dev/null https://api.snapcraft.io/v2/snaps/info/core 2>&1 || exit 1
+              echo "==> HTTPS transparent proxy test passed"
+              echo "==> Signaling success via curl"
+              curl -X PUT -H "Content-Type:" \
+                --data-binary '{"Status":"SUCCESS","Reason":"Proxy verification passed","UniqueId":"ProxyClient","Data":"OK"}' \
+                "${ProxyVerificationWaitConditionHandle}"
+            else
+              echo "==> Installing cfn-bootstrap"
+              pip3 install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz --break-system-packages
```
himani2411: NOTE TO REVIEWER: I needed to add `--break-system-packages` since I upgraded the AMI from ubuntu 20 to Ubuntu24, this OS version upgrade comes with PEP 668 (externally-managed-environment) enforcement. Since we dont really use this proxyClient apart from a verification I choose the option to use `--break-system-packages` instead of creating a new pyenv and installing the dependency for verification.

himani2411: Ignore this as i resolved another failure of dpkg lock issue by upgrading to ubuntu22 instead of ubuntu24

tests/integration-tests/tests/proxy/test_proxy.py
himani2411: One of the checks we do as part of the test_proxy test to to be able to login into the HeadNode using a Bastion instance from another VPC, which is why I will keep the current SSHCidr. Since we have had no issues so far and this is an ephemeral stack we can decrease the priority for this specific change.

gmarciani (dismissed):

gmarciani (approved):

gmarciani:
We agreed on addressing the compatibility with other partitions in a separate PR.

PR #7325, opened 2026-04-02
Handle BucketAlreadyOwnedByYou race condition in S3 bucket initialization

hanwen-cluster:
### Description of changes

The default S3 bucket name is deterministic per account+region, so concurrent cluster creation calls (e.g. parallel integration tests) can race: multiple callers get a 404 from head_bucket, then all attempt create_bucket. The first succeeds; the rest fail with BucketAlreadyOwnedByYou, which was previously unhandled and surfaced as "Unable to initialize s3 bucket."

Treat BucketAlreadyOwnedByYou as a success condition since it confirms the bucket exists and is owned by the caller, which is the desired end state. Log at info level and proceed normally.

### Tests
* Tests will run after PR is merged

### References

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+3.16.0
+------
+
+**BUG FIXES**
+- Fix sporadic S3 bucket (with name parallelcluster-*-v1-do-not-delete) creation failure when multiple create-cluster commands are running simultaneously in the same region.
+
 3.15.0
 ------
 
--- a/cli/src/pcluster/models/s3_bucket.py
+++ b/cli/src/pcluster/models/s3_bucket.py
@@ -478,8 +478,11 @@ class S3BucketFactory:
             try:
                 bucket.create_bucket()
             except AWSClientError as error:
-                LOGGER.error("Unable to create S3 bucket %s.", bucket.name)
-                raise error
+                if error.error_code == "BucketAlreadyOwnedByYou":
+                    LOGGER.info("S3 bucket %s was created shortly before by another process. Proceeding.", bucket.name)
+                else:
+                    LOGGER.error("Unable to create S3 bucket %s.", bucket.name)
+                    raise error
         else:
             LOGGER.error("Unable to check S3 bucket %s existence.", bucket.name)
             raise e


cli/src/pcluster/models/s3_bucket.py
```diff
@@ -478,8 +478,11 @@ def _create_bucket(cls, bucket: S3Bucket, e: AWSClientError):
             try:
                 bucket.create_bucket()
             except AWSClientError as error:
-                LOGGER.error("Unable to create S3 bucket %s.", bucket.name)
-                raise error
+                if error.error_code == "BucketAlreadyOwnedByYou":
+                    LOGGER.info("S3 bucket %s was concurrently created by another process. Proceeding.", bucket.name)
```
gmarciani: [minor] What about logging instead "Bucket ${BUCKET_NAME} is already owned by you". The fact the the bucket was created by a concurrent process is a low level detail we observed in our integ test, but this is a customer facing log line

gmarciani (commented):
[Docs] This bug fix deserves an entry in the changelog because it prevents cluster creation failures.

cli/src/pcluster/models/s3_bucket.py
hanwen-cluster: This detail is needed because the error wouldn't appear if the bucket existed long before cluster creation. This error is only triggered if the race condition happens

himani2411: This is message is part of the product not just the Integ test Infra so the messaging should be aligned with what the CX would see, which may or may not be multi-threaded cluster creation process.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+3.16.0
+------
+
+**BUG FIXES**
+- Fix sporadic S3 bucket creation failure when multiple create-cluster commands are running simultaneously in the same region.
```
himani2411: Do we have a name for the specific bucket like default bucket? or do-not-delete bucket? IF so I suggest we specify that in the changelog.

hanwen-cluster: Done

himani2411 (approved):

PR #7335, opened 2026-04-14
[integ-tests] Skip root volume tags check in us-iso regions

hanwen-cluster:
### Description of changes
* The tags are not propagated to root volumes in us-iso regions

### Tests
* Tests will run after the PR is merged

### References

### Checklist

### DIFF
--- a/tests/integration-tests/tests/tags/test_tag_propagation.py
+++ b/tests/integration-tests/tests/tags/test_tag_propagation.py
@@ -128,6 +128,7 @@ def _check_tag_propagation(
                 {"Name": "HeadNode", "parallelcluster:node-type": "HeadNode"},
             ),
             "tag_getter_kwargs": {"cluster": cluster, "os": os},
+            "skip": "us-iso" in cluster.region,
         },
         {
             "resource": "Compute Node",


tests/integration-tests/tests/tags/test_tag_propagation.py
```diff
@@ -148,7 +149,7 @@ def _check_tag_propagation(
                 {**config_file_tags, **queue_tags, **compute_resource_tags, **additional_config_tags},
             ),
             "tag_getter_kwargs": {"cluster": cluster, "os": os},
-            "skip": scheduler == "awsbatch",
+            "skip": scheduler == "awsbatch" or "us-iso" in cluster.region,
```
himani2411: why on Compute Node root Volume? This is not governed by the `propagate_tags_to_volume_on_creation` the same way as HN root volume is.

tests/integration-tests/tests/tags/test_tag_propagation.py
```diff
@@ -128,6 +128,7 @@ def _check_tag_propagation(
                 {"Name": "HeadNode", "parallelcluster:node-type": "HeadNode"},
             ),
             "tag_getter_kwargs": {"cluster": cluster, "os": os},
+            "skip": "us-iso" in cluster.region,
```
himani2411, line 131: Instead of removing the entire check which should not be the case since that reduces the coverage of the test. Can you remove the Tags that are specific to the HeadNode rather than the Volume itself.

tests/integration-tests/tests/tags/test_tag_propagation.py
hanwen-cluster: You are right, done!

tests/integration-tests/tests/tags/test_tag_propagation.py
himani2411: Ignore this comment as this should be skipped as originally intended.

himani2411 (approved):

PR #7350, opened 2026-04-22
[Test] Tolerate intermittent dcvsessionlauncher crash on g5g instances.

gmarciani:
### Description of changes
Tolerate the known dcvsessionlauncher SEGV (g_subprocess_send_signal -> on_read_startup_string_ready) on g5g instances.

This crash is caused by a DCV bug they are actively investigating. The crash is intermittent and even when it occurs its impact is intermittent and only on the first dcv connection. There is no point for us to alarm on this specific crash because, when it is impactful, the test will detect the failure in the dcv connection.

### Tests
SUCCESS `test_dcv_configuration`


himani2411 (dismissed):
LGTM!

himani2411 (approved):

PR #7357, opened 2026-04-29
[Test] Fix issues causing intermittent test failures: test_efa, test_ad_integration, test_queue_parameters_update, test_slurm, test_multiple_efs

gmarciani:
### Description of changes
Fixed issues causing intermittent failures on the following tests:
* `test_efa`: Replaced hpc5a.48xlarge on us-east-2 with hpc6a.48xlarge on eu-north-1 to reduce the risk of insufficient capacity exceptions.
* `test_ad_integration`: Fixed a race condition in the SSH key generation check by adding a sleep before the ls command inside the switch-user session, giving the PAM hook time to finish generating the user's SSH key and avoiding sporadic "Permission denied" failures.
* `test_queue_parameters_update`: Prevented the test from hanging indefinitely when dumping job output, caused by a weak condition and the lack of an SSH timeout. Hangs in this test could also trigger the watchdog to kill unrelated tests sharing the same VPC.
* `test_multiple_efs`: Switched the auxiliary instance that writes EFS data from AL2 to AL2023. AL2 reaches end of life in June 2026, and AL2023 should provide a more robust networking driver bootstrap, preventing the sporadic networking failures observed in this test.
* `test_slurm`: Removed an unnecessary dependency on NFS shared storage by replacing the shared-file round-trip with an srun that streams output back directly. NFS cross-client visibility is already covered by storage-specific tests and was just noise here.

Also, included the following improvements:
1. the RCA retrieval by including the chef client log captured in the head node console logs.
2. removed redundant use of fixture marks to make our tests able to run with pytest 8+ which is the version pulled in when we run the test with python 3.10+

### Tests
* SUCCESS `test_efa`
* SUCCESS `test_ad_integration`
* SUCCESS `test_queue_parameters_update`
* SUCCESS `test_slurm`
* SUCCESS `test_multiple_efs`

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -1734,15 +1734,17 @@ def _wait_compute_cloudinit_done(remote_command_executor, compute_node):
     assert_that(compute_cloudinit_status_output).contains("status: done")
 
 
-def _assert_mpi_process_completion(
-    remote_command_executor, slurm_commands, num_nodes, after_completion, check_proc_file
-):
-    result = slurm_commands.submit_command(
-        f'ps aux | grep "mpiexec.hydra.*sleep" | grep -v "grep" >> {check_proc_file}', nodes=num_nodes
+def _assert_mpi_process_completion(remote_command_executor, num_nodes, after_completion):
+    proc_track_result = remote_command_executor.run_remote_command(
+        f'srun -N {num_nodes} bash -c \'ps aux | grep "mpiexec.hydra.*sleep" | grep -v "grep" || true\''
     )
-    job_id = slurm_commands.assert_job_submitted(result.stdout)
-    slurm_commands.wait_job_completed(job_id)
-    proc_track_result = remote_command_executor.run_remote_command(f"cat {check_proc_file}")
+    logging.info(
+        "MPI process check (after_completion=%s) on %d nodes:\nstdout:\n%s\nstderr:\n%s",
+        after_completion,
+        num_nodes,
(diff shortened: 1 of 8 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -1734,15 +1734,17 @@ def _wait_compute_cloudinit_done(remote_command_executor, compute_node):
...
-    result = slurm_commands.submit_command(
-        f'ps aux | grep "mpiexec.hydra.*sleep" | grep -v "grep" >> {check_proc_file}', nodes=num_nodes
+def _assert_mpi_process_completion(remote_command_executor, num_nodes, after_completion):
+    # Run the process check inline on compute nodes via srun: stdout is streamed back over
+    # Slurm's RPC channel, avoiding the shared filesystem entirely. This keeps the test focused
+    # on Slurm behavior (process tracking and job termination) rather than incidentally exercising
+    # NFS cross-client visibility, which is validated separately by the storage tests
+    # (e.g. tests/storage/test_ebs.py).
+    # The trailing "|| true" ensures grep's exit code 1 on no match does not fail srun,
+    # so we get empty stdout when no matching processes are found.
```
himani2411: We get an empty output but debugging will be difficult if we do not get errors. With previous process we were printing the output file; lets print the slurm_,job_id>.out and .err files in this function.

gmarciani: addressed by logging both stdout and stderr of the srun job

himani2411 (dismissed):

himani2411 (approved):

PR #7361, opened 2026-04-30
[Update] Fix race condition in login nodes load balancer lookup causing cluster update failure in clusters with login nodes.

gmarciani:
### Description of changes
Fix race condition in login nodes load balancer lookup causing cluster update failure in clusters with login nodes.

With this change we now retrieve the load balancer for a given login node pool with a single call tp the Resource Groups Tagging API, instead of listing all account NLBs and filtering by tags.

The previous approach could fail sporadically when a concurrent cluster operation deleted another NLB between list_load_balancers and describe_tags.

This change requires the CLI user to have the extra permission `tag:GetResources`.

### Tests
1. Manually verified that the describe-cluster works as usual, retrieving the login node details. This operation triggers the changed code.
1. SUCCESS `test_create_disable_sudo_access_for_default_user `. This test makes use of this change and it was the test that detected the race condition we are fixing in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
 **CHANGES**
 - The validator `ClusterNameValidator` now enforces cluster names to be limited to 40 characters when using `ExternalSlurmdbd`, 
   consistent with the existing limit for `Database`. This prevents runtime failures caused by MySQL's table name length limit.
+- The CLI now requires the additional permission `tag:GetResources`.
 
 **BUG FIXES**
 - Fix sporadic S3 bucket (with name parallelcluster-*-v1-do-not-delete) creation failure when multiple create-cluster commands are running simultaneously in the same region.
@@ -14,6 +15,8 @@ CHANGELOG
   via custom Slurm settings or the cluster name contains upper-case letters.
 - Remove deprecated parameter `AccountingStorageUser` from Slurm configuration that was causing harmless error messages.
 - Fix clustermgtd failing to detect compute node bootstrap timeouts, which prevented the cluster from entering protected mode.
+- Fix race condition in load balancer lookup by using `tag:GetResources` to resolve load balancer ARNs by tags, 
+  which used to cause cluster update failure on clusters with login nodes.
 
 3.15.0
 ------
--- a/cli/src/pcluster/aws/elb.py
+++ b/cli/src/pcluster/aws/elb.py
@@ -41,6 +41,15 @@ class ElbClient(Boto3Client):
         return response["LoadBalancers"], response.get("NextMarker")
 
     @AWSExceptionHandler.handle_client_exception
+    def describe_load_balancer(self, load_balancer_arn: str) -> Any:
+        """Retrieve the load balancer matching the given ARN, or None if not found."""
+        if not load_balancer_arn:
+            return None
+        response = self._client.describe_load_balancers(LoadBalancerArns=[load_balancer_arn])
+        load_balancers = response.get("LoadBalancers", [])
+        return load_balancers[0] if load_balancers else None
+
+    @AWSExceptionHandler.handle_client_exception
     def describe_tags(self, load_balancer_arns: []):
         """Retrieve a list of tags associated to the load balancer arns provided as parameter."""
         """You can specify up to 20 load balancer arns in a single call."""
--- a/cli/src/pcluster/aws/resource_groups_tagging_api.py
+++ b/cli/src/pcluster/aws/resource_groups_tagging_api.py
@@ -1,0 +1,47 @@
+# Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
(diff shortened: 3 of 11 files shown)


cli/src/pcluster/aws/resource_groups_tagging_api.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
himani2411: Nitpick: Can we change this to 2026

cli/tests/pcluster/aws/test_resource_groups_tagging_api.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
himani2411: Nitpick: Can we change this to 2026

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
 - Fix cluster creation failure caused by Slurm accounting bootstrap failing when ClusterName is overridden
   via custom Slurm settings or the cluster name contains upper-case letters.
 - Remove deprecated parameter `AccountingStorageUser` from Slurm configuration that was causing harmless error messages.
+- Fix race condition in load balancer lookup that used to cause cluster update failure on clusters with login nodes.
```
himani2411: Can we specify the need of a new IAM policy in chnagelog? Example: `Fix race condition in load balancer lookup by using tag:GetResources to resolve load balancer ARNs by tags, which used to cause cluster update failure on clusters with login nodes.`

cli/tests/pcluster/aws/test_resource_groups_tagging_api.py
```diff
@@ -0,0 +1,170 @@
...
+    boto3_stubber("resourcegroupstaggingapi", mocked_requests)
+
+    return_value = ResourceGroupsTaggingApiClient().get_resources(
+        tag_filters={
+            "parallelcluster:cluster-name": "cluster-1",
+            "parallelcluster:login-nodes-pool": "login",
+        },
+        resource_type_filters=["elasticloadbalancing:loadbalancer"],
+    )
+    assert_that(return_value).is_equal_to([dummy_arn_1, dummy_arn_2])
```
himani2411, line 57: Nit: The tests cover the multi-ARN response path, but by design we expect exactly one NLB per pool — so taking the first ARN is correct. The previous approach had the same behavior. Not a blocker for this PR, but worth noting: if a customer happens to create their own NLB (or another cluster in process of deleting) with matching parallelcluster:cluster-name and parallelcluster:login-nodes-pool tags, we'd pick it up by mistake. As a future improvement, we could add a distinguishing tag filter (e.g. aws:cloudformation:stack-id) to make the lookup more precise.

cli/src/pcluster/aws/elb.py
```diff
@@ -40,6 +40,15 @@ def _describe_load_balancers(self, next_marker=None) -> Tuple[List[Any], str]:
...
         return response["LoadBalancers"], response.get("NextMarker")
 
+    @AWSExceptionHandler.handle_client_exception
+    def describe_load_balancer(self, load_balancer_arn: str) -> Any:
+        """Retrieve the load balancer matching the given ARN, or None if not found."""
+        if not load_balancer_arn:
+            return None
+        response = self._client.describe_load_balancers(LoadBalancerArns=[load_balancer_arn])
+        load_balancers = response.get("LoadBalancers", [])
+        return load_balancers[0] if load_balancers else None
```
himani2411, line 50: Same as https://github.com/aws/aws-parallelcluster/pull/7361/changes#r3173816740

cli/tests/pcluster/aws/test_resource_groups_tagging_api.py
gmarciani: This is a very good point and a weakness of both the previous and current approach. 
However, to trigger this weakness the user must either:
1. create a cluster with the same name in the same region, which is not possible.
2. create an ELB and use the `parallelcluster` tags. If a user does something like that on ELB, EC2 instances or any other resource, it could cause much worse issues on the cluster than an impact on this validation.

Also using `aws:cloudformation:stack-id` would reduce the risk, but would not solve, if we assume that the user sets "protected" tags on arbitrary resources.

Also, regarding the stack id tag, the ELB does not have the cluster stack id but the nested stack id. to retrieve that ID within the validation logic, we must add an extra call to CFN to determine the nested stack id. This would introduce complexity, without being a definitive solutionnto this weakness.

If you agree, I propose to keep the proposed approach as we are not worsening this wekaness

CHANGELOG.md
gmarciani: Good point. Done

cli/tests/pcluster/aws/test_resource_groups_tagging_api.py
gmarciani: done

cli/src/pcluster/aws/resource_groups_tagging_api.py
gmarciani: done

cli/src/pcluster/aws/elb.py
gmarciani: Same as above

cli/tests/pcluster/aws/test_resource_groups_tagging_api.py
himani2411: agreed

himani2411 (dismissed):

himani2411 (approved):

PR #7364, opened 2026-04-30
[Test] Add end2end test to validate SLURM Rest API.

gmarciani:
### Description of changes
Add end2end test to validate SLURM Rest API.
The cluster requires two custom actions:
* OnNodestart action: this is required to patch the SLURM installation so that slurmrestd can start. Without this patching action, slurmrestd would fail because it cannot find http-parser, which ia a required dependency.
* OnNodeConfigured action: this is the "standard" way to enable the SLURM Rest API, taken from the public AWS example.

the test validates that:
1. slurmrestd is running
2. SLIURM API is functional, by sending a request to the `ping` endpoint and checking the result is healthy. 

### Tests
SUCCESS `test_slurm_rest_api`

### DIFF
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -561,6 +561,12 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: [{{ OS_X86_0 }}]
           schedulers: ["slurm"]
+    test_slurm_rest_api.py::test_slurm_rest_api:
+      dimensions:
+        - regions: [ "eu-north-1" ]
+          instances: {{ common.INSTANCES_DEFAULT_X86 }}
+          oss: [{{ OS_X86_0 }}]
+          schedulers: [ "slurm" ]
   spot:
     test_spot.py::test_spot_default:
       dimensions:
--- a/tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/configure_slurmrestd.sh
+++ b/tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/configure_slurmrestd.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+#
+# OnNodeConfigured custom action for the test_slurm_rest_api integration test.
+#
+# Based on the upstream Slurm REST API postinstall script:
+#   https://raw.githubusercontent.com/aws-samples/aws-parallelcluster-post-install-scripts/main/rest-api/postinstall.sh
+#
+# Differences from upstream:
+#   - Uses a local slurm_rest_api.rb (uploaded to S3 by the test) instead of downloading
+#     it from GitHub. The local copy includes an `apt-get update` before installing nginx
+#     on Debian/Ubuntu to avoid stale package index 404 errors.
+#
+# Arguments:
+#   $1 - S3 URI of the adapted slurm_rest_api.rb (e.g. s3://bucket/scripts/slurm_rest_api.rb)
+
+set -ex
+
+SLURM_REST_API_RB_S3_URI="${1:?Usage: configure_slurmrestd.sh <s3-uri-of-slurm_rest_api.rb>}"
+
+# Copy Slurm REST API configuration files and scripts
+tmp_dir=/tmp/slurm_rest_api
+mkdir -p $tmp_dir
+
+source_path=https://raw.githubusercontent.com/aws-samples/aws-parallelcluster-post-install-scripts/main/rest-api
+files=(slurmrestd.service nginx.conf)
+for file in "${files[@]}"
+do
+    wget -qO- $source_path/$file > $tmp_dir/$file
+done
+
+rotate_jwt_path=/opt/parallelcluster/scripts/rotate_jwt.sh
+wget -qO- $source_path/rotate_jwt.sh > $rotate_jwt_path
+chmod +x $rotate_jwt_path
+
+# Download the adapted slurm_rest_api.rb from S3
+aws s3 cp "${SLURM_REST_API_RB_S3_URI}" $tmp_dir/slurm_rest_api.rb
+
+# Setup Slurm REST API
+sudo cinc-client \
+  --local-mode \
+  --config /etc/chef/client.rb \
+  --log_level auto \
+  --force-formatter \
+  --chef-zero-port 8889 \
+  -j /etc/chef/dna.json \
+  -z $tmp_dir/slurm_rest_api.rb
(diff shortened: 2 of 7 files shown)


tests/integration-tests/configs/develop.yaml
```diff
@@ -561,6 +561,12 @@ test-suites:
           instances: {{ common.INSTANCES_DEFAULT_X86 }}
           oss: [{{ OS_X86_0 }}]
           schedulers: ["slurm"]
+    test_slurm_rest_api.py::test_slurm_rest_api:
+      dimensions:
+        - regions: [ "us-east-1" ]
```
hanwen-cluster: minor: does the test have to be in a popular region? We should try to avoid regions used to reserve dynamic capacity reservation: https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/framework/tests_configuration/config_renderer.py#L302-L312

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/configure_slurmrestd.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+#
+# Verbatim copy of the upstream Slurm REST API postinstall script maintained in
+# aws-samples/aws-parallelcluster-post-install-scripts:
+#   https://raw.githubusercontent.com/aws-samples/aws-parallelcluster-post-install-scripts/main/rest-api/postinstall.sh
+#
+# Keep this file in sync with the upstream source; do not modify it locally.
+
+set -x
+set -e
```
hanwen-cluster: minor: how are these two lines differ from `set -ex`?

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/pcluster.config.yaml
```diff
@@ -0,0 +1,46 @@
...
+  Os: {{ os }}
+HeadNode:
+  InstanceType: {{ instance }}
+  Networking:
+    SubnetId: {{ public_subnet_id }}
+    AdditionalSecurityGroups:
+      - {{ database_client_security_group }}
+  Iam:
+    AdditionalIamPolicies:
+      - Policy: arn:{{ partition }}:iam::aws:policy/AmazonSSMManagedInstanceCore
```
hanwen-cluster: Is this necessary? I believe the test framework automatically adds this?

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/pcluster.config.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - Policy: arn:{{ partition }}:iam::aws:policy/SecretsManagerReadWrite # Required to read/write the JWT token from Secrets Manager
+    S3Access:
+      - BucketName: {{ bucket_name }} # Required to download custom actions
+  Ssh:
+    KeyName: {{ key_name }}
+  Imds:
+    Secured: {{ imds_secured }}
```
hanwen-cluster: Is this necessary? I believe the test framework automatically adds this?

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/rebuild_slurm.sh
```diff
@@ -0,0 +1,82 @@
...
+COOKBOOK_VENV=$(find_dir /opt/parallelcluster/pyenv/versions 3 cookbook_virtualenv)
+
+log "Activating cookbook virtualenv at ${COOKBOOK_VENV}"
+source "${COOKBOOK_VENV}/bin/activate"
+
+log "Uninstalling current Slurm installation"
+make uninstall
+
+# Install http-parser from local sources into /usr/lib64 (AL2023 only)
+if grep -q "Amazon Linux 2023" /etc/os-release; then
```
hanwen-cluster, line 51: What does the special case only applies to AL2023?

tests/integration-tests/configs/develop.yaml
gmarciani: Not at all, we can move. Will use eu-north-1

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/configure_slurmrestd.sh
gmarciani: No difference. I just copy pasted the original script. Now merged into single line.

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/pcluster.config.yaml
gmarciani: Yep, can be removed. Done

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/pcluster.config.yaml
gmarciani: Yep, can be removed. Done

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/rebuild_slurm.sh
gmarciani: AL23 is the only OS where http-parser is installed from source by us and it is installed on a non default path. On every other OS the package is installed from the OS repos and the library is already in the path loaded by default by the dynamic linker.

hanwen-cluster (dismissed):

tests/integration-tests/tests/schedulers/test_slurm_rest_api/test_slurm_rest_api/configure_slurmrestd.sh
```diff
@@ -0,0 +1,48 @@
...
+sudo cinc-client \
+  --local-mode \
+  --config /etc/chef/client.rb \
+  --log_level auto \
+  --force-formatter \
+  --chef-zero-port 8889 \
+  -j /etc/chef/dna.json \
+  -z $tmp_dir/slurm_rest_api.rb
+
+set +e
```
hanwen-cluster: What does this line do?

gmarciani: Left over from the original example. No point in keeping it, removed.

hanwen-cluster (approved):

PR #7374, opened 2026-05-04
[Test] Export image build logs to test outputs on failure.

gmarciani:
### Description of changes
When a build-image integration test fails, the image CloudWatch log group is deleted during cleanup, making post-mortem debugging difficult. This change exports the full image build log to the test output directory before the assertion fails, so the logs are preserved as Jenkins build artifacts.

The full build image logs will be stored in tess_outputs folder in Jenkins workspace. Example: `tests_outputs/1777929048.3008714.out/us-east-1/image_build_logs/integ-tests-build-image-ctud7246uq4iz1sp-mgbi-mgbi-logs.tar.gz`

Also removed the `image.keep_logs = True` which was basically dead code giving the illusion that logs were persisted, but it was actually unused.

### Tests
Executed `test_build_image` with a failure injected to trigger the export of logs on failure. Verified that the expected image logs are stored in the Jenkins workspace (See example above).

### DIFF
--- a/tests/integration-tests/tests/createami/test_createami.py
+++ b/tests/integration-tests/tests/createami/test_createami.py
@@ -12,6 +12,7 @@
 import datetime
 import json
 import logging
+import os
 import re
 import tarfile
 import tempfile
@@ -134,7 +135,7 @@ def test_build_image_no_internet(
     )
 
     image = images_factory(image_id, image_config, region)
-    _test_build_image_success(image)
+    _test_build_image_success(image, request.config.getoption("output_dir"))
 
 
 @pytest.mark.usefixtures("instance")
@@ -246,7 +247,7 @@ def test_build_image(
     )
 
     with soft_assertions():
-        _test_build_image_success(image)
+        _test_build_image_success(image, request.config.getoption("output_dir"))
         _test_build_instances_tags(image, image.config["Build"]["Tags"], region)
         _test_build_imds_settings(image, "required", region)
         _test_image_tag_and_volume(image)
@@ -330,7 +331,7 @@ def test_kernel4_build_image_run_cluster(
 
     image_id = generate_stack_name("integ-tests-build-image", request.config.getoption("stackname_suffix"))
     image = images_factory(image_id, image_config, region, **{"rollback-on-failure": False})
-    _test_build_image_success(image)
+    _test_build_image_success(image, request.config.getoption("output_dir"))
     _test_build_imds_settings(image, "required", region)
     _test_list_images(image)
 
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -705,11 +706,23 @@ def _test_build_image_failed(image):
         logging.info(pcluster_describe_image_result)
 
     if image.image_status == "BUILD_FAILED":
-        image.keep_logs = True
+        _export_image_logs(image, output_dir)
         _keep_recent_logs(image)
     assert_that(image.image_status).is_equal_to("BUILD_FAILED")
 
 
+def _export_image_logs(image, output_dir):
```
himani2411, line 714: Could you add the `wait_for_no_active_export_tasks` check before initiating an export? There's a limit of one export task per account per region.

Since all the `createami` tests run in the same region, if they all fail simultaneously, we could hit the resource limit. I'd suggest:

1. Adding a check for any active export task before starting a new one.
2. If possible, distributing some of the `createami` tests across different regions so we don't end up sequentially blocking all 7 `createami` operations in a single region.

himani2411: We also have the last 200 lines of the build image logs being printed as part of the teardown. Did those logs not help in identifying the root cause of failure? What is the reason that we need the entire logs and not just the end of image logs? 
And why not simply increase the number of lines from 200 to 500 to get more logs?

gmarciani: > Could you add the wait_for_no_active_export_tasks check before initiating an export?

Sure good catch! Done.

> If possible, distributing some of the createami tests across different regions so we don't end up sequentially blocking all 7 createami operations in a single region.

This requires a refactoring of our dynamic reservations. IK agree with your proposal, but I would do decoupled from this PR.

> Did those logs not help in identifying the root cause of failure?
And why not simply increase the number of lines from 200 to 500 to get more logs?

Unfortunately they were not enough in a recent case. Those lines were reporting the final error, but the actual root cause was way above in the logs. Probably in that case a ~700/800 lines would be enough. However, we used to be ok with 200, then we need to increase to 7/800, maybe tomorrow we need 1000, so at this point why not getting the full log? 

In terms of test coverage, the test_build_image is equivalent to a second stage plus some additional noise because we are not in full control of the input AMI; so if we consider required to have full build log for the second stage even when it succeeds, I believe it is useful to have the full log for test when it fails.

Do you agree?

himani2411: Agreed on the export and decoupling. Though I think we should remove the 200 lines export since its not required anymore!

himani2411 (approved):

PR #7376, opened 2026-05-05
[Test] Add test runner option `--retain-on-failure` to retain tests on failure

gmarciani:
### Description of changes
Add test runner option `--retain-on-failure [N]` to retain `N` tests on failure.
This test feature is useful to capture intermittent failures in a controlled way and retain the broken environments that are affected to facilitate investigations.

A recent example where it was necessary was the intermittent dcv session launcher failure, but it can also useful to capture the intermittent NFS port binding failure.

When the option is active, the test generates two files in the test workspace, that are used to both synchronize processes on the retention decision and also to facilitate the discovery of retained resources for the technician that is invetsigating the failure:
* `.retain_on_failure_tests`: contains the list of failed tests for which resources have been retained.
* `.retained_stacks`: contains the list of stack ARNs that have been retained.

**NOTES**
1. This PR requires a corresponding change in Jenkins to make it work so it will not cleanup the resources that are supposed to be retained.

### Tests
Executed 4 parallel instances of `test_dcv_configuration` with injected failures in all of them in the following scenarios:
  * w/o `--retain-on-failure`: failures are reported as usual, but no test stack is retained.
  * w/ `--retain-on-failure 1`: failures are reported as usual, cluster/vpc/iam stack retained for one single test failure.
  * w/ `--retain-on-failure 2`: failures are reported as usual, cluster/vpc/iam stack retained for two test failures.

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -41,7 +41,13 @@ from conftest_markers import (
 )
 from conftest_networking import unmarshal_az_override
 from conftest_tests_config import apply_cli_dimensions_filtering, parametrize_from_config, remove_disabled_tests
-from conftest_utils import add_filename_markers, get_reporting_region
+from conftest_utils import (
+    add_filename_markers,
+    get_reporting_region,
+    is_region_retained,
+    register_retained_stack,
+    retain_resources_on_teardown,
+)
 from constants import SCHEDULERS_SUPPORTING_IMDS_SECURED, NodeType
 from filelock import FileLock
 from framework.credential_providers import aws_credential_provider, register_cli_credentials_for_region
@@ -171,6 +177,12 @@ def pytest_addoption(parser):
         "--no-delete", action="store_true", default=False, help="Don't delete stacks after tests are complete."
     )
     parser.addoption(
+        "--retain-on-failure",
+        type=int,
+        default=0,
+        help="Retain cluster, VPC, IAM, and S3 bucket stacks for up to N failing tests (0=disabled, max 5).",
+    )
+    parser.addoption(
         "--retain-ad-stack", action="store_true", default=False, help="Retain AD stack and corresponding VPC stack."
     )
     parser.addoption("--benchmarks", action="store_true", default=False, help="enable benchmark tests")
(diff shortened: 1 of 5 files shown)


tests/integration-tests/conftest_utils.py
```diff
@@ -344,3 +345,164 @@ def get_reporting_region(region: str):
...
+            retained_stacks = set()
+
+        retained_stacks.add(stack_arn)
+        with open(stacks_path, "w") as f:
+            f.write("\n".join(retained_stacks) + "\n")
+
+    logging.info("Registered stack %s for retention", stack_arn)
+
+
+def _check_or_register_retain(request, test_id):
```
himani2411: I understand you you divided `resources` and `stacks` to handle the xdist workers and session level scope. 

Can you retain this naming convention in the functions names you are using? It will be easier for the reader to understand the scope.

tests/integration-tests/conftest.py
```diff
@@ -1178,6 +1215,8 @@ def _create_bucket(region):
     for bucket in created_buckets:
         if request.config.getoption("no_delete"):
             logging.info(f"Not deleting S3 bucket {bucket[0]}")
+        elif request.config.getoption("retain_on_failure") and is_region_retained(request, bucket[1]):
```
himani2411, line 1218: [Non-blocking] There's quite a bit of code duplication in this conftest file — not introduced by this PR, but worth cleaning up in a follow-up.

tests/integration-tests/conftest_utils.py
```diff
@@ -344,3 +345,164 @@ def get_reporting_region(region: str):
...
+def _current_test_failed(request):
+    """Return True if the current test item has failed during setup or call phases."""
+    for phase in ("setup", "call"):
+        rep = getattr(request.node, f"rep_{phase}", None)
+        if rep is not None and rep.failed:
+            return True
+    return False
+
+
+def _get_retain_counter_path(request):
```
himani2411, line 376: same as https://github.com/aws/aws-parallelcluster/pull/7376/changes#r3204061062

tests/integration-tests/conftest_utils.py
gmarciani: I agree on changes to improve readability, but I am not getting the specific change you are asking :-)

himani2411 (approved):

PR #7385, opened 2026-05-05
[Test] Prevent known_hosts from growing indefinitely from bastion SSH connections in integration tests.

gmarciani:
### Description of changes
Prevent known_hosts from growing indefinitely from bastion SSH connections in integration tests.
In particular, we added the SSH option UserKnownHostsFile=/dev/null so that the head node key is not persisted into the test runner's known_hosts file. Without this change, the known_hosts file grows indefinitely.

It is safe to add this option because such connections are ephemeral and we are already ignoring the verification of keys through the use of option StrictHostKeyChecking=no.

### Tests
SUCCESS `test_cluster_in_private_subnet`:
* before this fix, the execution of this test was adding 2 entries in the known_hosts file
* after this fix, the test succeeded without adding any new entry to the known_hosts file.

### DIFF
--- a/tests/integration-tests/remote_command_executor.py
+++ b/tests/integration-tests/remote_command_executor.py
@@ -83,11 +83,10 @@ class RemoteCommandExecutor:
             },
         }
         if bastion:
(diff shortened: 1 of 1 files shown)


hehe7318 (approved):
lgtm!

PR #7388, opened 2026-05-06
[Test] Reduce the risk of export-logs "Resource limit exceeded" failures in parallel tests.

gmarciani:
### Description of changes
Reduce the risk of `test_cli_commands` and `test_build_image` failures caused by "Resource limit exceeded" when exporting logs.

CloudWatch Logs allows only one active export task per account per region. When parallel tests compete for this slot, the previous approach; calling `wait_for_no_active_export_tasks` once before a retry loop; left a race window: a parallel test could start an export after the wait, causing every retry to fail without ever re-checking.

Changes:

- **Wait before every export attempt.** `wait_for_no_active_export_tasks` now lives inside `Cluster.export_logs` and `Image.export_logs`, right before the pcluster command. On retry, the wait re-executes, so each attempt verifies the slot is free before firing.

- **Increased retry budget from 3 to 10 minutes.** The old window was shorter than a typical export task duration (2–5 minutes), so retries would exhaust before the competing task finished.

- **Random jitter (10–20s) on both the wait polling and the retry delay.** Prevents thundering herd; parallel tests desynchronize naturally instead of colliding in lockstep on a fixed cadence.

- **Log every failure before retrying.** Each failed attempt logs the error, giving clear evidence in test output when retries occur.

### Tests
SUCCESS 2 parallel `test_cli_commands` + 2 `test_build_image` (so that we expose 4 instances competing for logs export).
Also tested the unhappy path in `test_build_image` where the logs are exported in case of failure.

### DIFF
--- a/tests/integration-tests/tests/createami/test_createami.py
+++ b/tests/integration-tests/tests/createami/test_createami.py
@@ -44,7 +44,6 @@ from tests.common.utils import (
     get_installed_parallelcluster_version,
     retrieve_latest_ami,
     upload_github_artifacts_to_s3,
-    wait_for_no_active_export_tasks,
 )
 from tests.proxy.test_proxy import proxy_stack_factory  # noqa: F401
 
@@ -254,9 +253,7 @@ def test_build_image(
         _test_list_image_log_streams(image)
         _test_get_image_log_events(image)
         _test_list_images(image)
-        wait_for_no_active_export_tasks(region)
         _test_export_logs(s3_bucket_factory, image, region)
-        wait_for_no_active_export_tasks(region)
         _test_export_logs(s3_bucket_factory, image, region, True)
 
     _test_cluster_creation(
@@ -437,6 +434,7 @@ def _set_s3_bucket_policy(bucket_name, partition, region):
 
 
 def _test_export_logs(s3_bucket_factory, image, region, use_pcluster_bucket=False):
+    bucket_name = None
     if not use_pcluster_bucket:
         bucket_name = s3_bucket_factory()
         logging.info("bucket is %s", bucket_name)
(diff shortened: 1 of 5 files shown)


tests/integration-tests/tests/createami/test_createami.py
```diff
@@ -451,14 +449,11 @@ def _test_export_logs(s3_bucket_factory, image, region, use_pcluster_bucket=Fals
...
-        if not use_pcluster_bucket:
-            ret = retry(wait_fixed=seconds(20), stop_max_delay=minutes(10))(image.export_logs)(
-                bucket=bucket_name, output_file=output_file, bucket_prefix=bucket_prefix
-            )
-        else:
-            ret = retry(wait_fixed=seconds(20), stop_max_delay=minutes(10))(image.export_logs)(
-                output_file=output_file, bucket_prefix=bucket_prefix
-            )
+        ret = image.export_logs(
+            bucket=bucket_name,
```
hehe7318, line 453: [Blocking] After removing the if else condition, plus you added `bucket_name = None`. Won't it pass the para `--bucket None` to `pcluster export-image-logs` command?

I think it will, because:
https://github.com/aws/aws-parallelcluster/blob/9440371e7d085ba7164312d0aff6a9ba69ebea7b/tests/integration-tests/images_factory.py#L114-L119

In the PR description I saw you mentioned:
>Executed 2 runs of test_cli_commands (we faced the export logs limit exceeded before this fix) in parallel with test_build_image (with injected failure to trigger logs export)

Can you also test `test_build_image` without failure to see the behavior?

gmarciani: Good catch, thank you! Fixed by making the export_logs function exclude the None values, the same way it is done for the cluster_factory.export_logs. Re-executing the tests with the successful scenario

hehe7318 (approved):

PR #7401, opened 2026-05-19
[Test] Extend test_essential_feature to also cover basic GPU workload leveraging CUDA Samples.

gmarciani:
### Description of changes
Extend test_essential_feature to also cover basic GPU workload leveraging CUDA Samples.

### Tests
* SUCCESS `test_essential_feature ` and verified that it is using the expected 5 equivalent flex instance types

### DIFF
--- a/tests/integration-tests/tests/basic/test_essential_features/test_essential_features/gpu_job.sh
+++ b/tests/integration-tests/tests/basic/test_essential_features/test_essential_features/gpu_job.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+#SBATCH --job-name=cuda-gpu-validate
+#SBATCH --output=cuda-gpu-validate-%j.out
+
+# Build and run a single CUDA sample (passed as a script argument) from the
+# pre-installed /usr/local/cuda-samples-13.0 tree. CUDA samples 13.x are
+# CMake-only and /usr/local/... isn't writable, so the script copies the
+# sample into a temp dir before building.
+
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/basic/test_essential_features/test_essential_features/gpu_job.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+SAMPLE_REL=$1
+SAMPLE_NAME=${SAMPLE_REL##*/}
+
+export PATH=/usr/local/cuda/bin:${PATH}
+
+echo "Node: $(hostname)"
+echo "Sample: $SAMPLE_REL"
+echo "SLURM_JOB_GPUS=${SLURM_JOB_GPUS:-unset}"
+echo "CUDA_VISIBLE_DEVICES=${CUDA_VISIBLE_DEVICES:-unset}"
```
hehe7318, line 25: [Minor] Do we need to explicitly set `--gres=gpu` to ensure GPU is visible?

hehe7318 (approved):
Approve with a comment.

tests/integration-tests/tests/basic/test_essential_features/test_essential_features/gpu_job.sh
gmarciani: not in this case because we are targeting the queue that only has cr with GPUs

PR #6655, opened 2025-02-10
[Scaling] Removing usage of cfn-init for compute fleet

himani2411:
### Description of changes
 Dependent on  and Detailed description in https://github.com/aws/aws-parallelcluster-cookbook/pull/2875

### Tests
```
test-suites:
  update:
    test_update.py::test_dynamic_file_systems_update:
      dimensions:
        - regions: [ "us-east-2" ]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: [ "alinux2023" ]
          schedulers: [ "slurm" ]
    test_update.py::test_update_slurm:
      dimensions:
        - regions: ["eu-central-1"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["rocky8"]
    test_update.py::test_update_compute_ami:
      dimensions:
        - regions: ["eu-west-1"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["rocky9"]
    test_update.py::test_update_instance_list:
      dimensions:
        - regions: ["ap-south-1"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["ubuntu2004"]
          schedulers: ["slurm"]
    test_update.py::test_queue_parameters_update:
      dimensions:
        - regions: [ "ap-south-1" ]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["alinux2"]
          schedulers: [ "slurm" ]
    test_update.py::test_dynamic_file_systems_update_rollback:
      dimensions:
        - regions: [ "us-east-2" ]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: [ "rhel8" ]
          schedulers: [ "slurm" ]
```

### References
* Link to impacted open issues.
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

[...]

PR #7409, opened 2026-05-27
[UpdateWorkflow] Make login node use head-node-driven orchestration for cluster updates

gmarciani:
**This PR depends on https://github.com/aws/aws-parallelcluster-cookbook/pull/3188**

### Description of changes
Make login node use head-node-driven orchestration for the update workflow.

With this change, login nodes do not depend on cfn-hup/cfn-init anymore and mirror the same update mechanism already adopted for compute nodes. With this change we expect the update workflow to be more resilient.

The changes in this PR mirrors the changes made for compute fleet in the following PRs:
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/3070: this is the core change that removes cfn-hup
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/2875: change in cookbook to remove cfn-init, required by change #1
1. https://github.com/aws/aws-parallelcluster/pull/6655: change in cli to remove cfn-init, required by change #1
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/3093: minor change related to logs

In particular:
1. files that used ot be written by cfn-init arer now written by user datab directives, such as dna.json.
2. parameters and permissions required by cfn-init are removed from the login nodes launch template
3. login nodes launch templates identifications are included in `launch-templates-config.json`
4. the validation of the chef attribute `in_place_update_on_fleet_enabled` is removed as the attribute is not used anymore (see https://github.com/aws/aws-parallelcluster-cookbook/pull/3188)

#### Notes
1. See comments "Note for the reviewer" for specific explanations about non obvious changes.

### Tests
* SUCCESS
```
test-suites:
  update:
    test_update.py::test_update_slurm:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update.py::test_dynamic_file_systems_update_rollback:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update.py::test_login_nodes_update:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update_rollback_failure.py::test_update_rollback_failure:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update_race_conditions.py::test_update_race_conditions:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ CHANGELOG
 3.16.0
 ------
 
+**ENHANCEMENTS**
+- Improve cluster update resiliency on login nodes by reusing the head-node-driven orchestration already in place on compute nodes,
+  removing the dependency on cfn-hup and cfn-init.
+
 **CHANGES**
 - The validator `ClusterNameValidator` now enforces cluster names to be limited to 40 characters when using `ExternalSlurmdbd`, 
   consistent with the existing limit for `Database`. This prevents runtime failures caused by MySQL's table name length limit.
@@ -19,6 +23,7 @@ CHANGELOG
 - Fix race condition in load balancer lookup by using `tag:GetResources` to resolve load balancer ARNs by tags, 
   which used to cause cluster update failure on clusters with login nodes.
 - Fail `pcluster build-image` early when the downloaded cookbook version does not match the ParallelCluster CLI version.
+- Fix login nodes not mounting `/opt/parallelcluster/shared` when EFS is used as the internal shared storage type.
 
 **DEPRECATIONS**
 - Amazon Linux 2 is no longer supported.
--- a/cli/src/pcluster/resources/login_node/user_data.sh
+++ b/cli/src/pcluster/resources/login_node/user_data.sh
@@ -48,6 +48,16 @@ datasource_list: [ Ec2, None ]
 output:
   all: "| tee -a /var/log/cloud-init-output.log | logger -t user-data -s 2>/dev/ttyS0"
 write_files:
+  - path: /tmp/dna.json
+    permissions: '0644'
+    owner: root:root
+    content: |
+      ${DnaJson}
+  - path: /tmp/extra.json
+    permissions: '0644'
+    owner: root:root
+    content: |
+      ${ExtraJson}
   - path: /tmp/bootstrap.sh
     permissions: '0744'
     owner: root:root
@@ -77,8 +87,6 @@ write_files:
       [ -f /etc/parallelcluster/pcluster_cookbook_environment.sh ] && . /etc/parallelcluster/pcluster_cookbook_environment.sh
 
       [ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh
-
-      $CFN_BOOTSTRAP_VIRTUALENV_PATH/cfn-init -s ${AWS::StackName} -v -c deployFiles -r ${LaunchTemplateResourceId} --region ${AWS::Region} --url ${CloudFormationUrl} --role ${CfnInitRole} || error_exit 'Failed to bootstrap the login node. Please check /var/log/cfn-init.log in the login node or in CloudWatch logs. Please refer to https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html#troubleshooting-v3-get-logs for more details on ParallelCluster logs.'
 
       # Configure AWS CLI using the expected overrides, if any.
       [ -f /etc/profile.d/aws-cli-default-config.sh ] && . /etc/profile.d/aws-cli-default-config.sh
@@ -129,7 +137,10 @@ write_files:
         vendor_cookbook
       fi
       cd /tmp
+      mkdir -p /etc/chef/ohai/hints
+      touch /etc/chef/ohai/hints/ec2.json
 
+      jq -s ".[0] * .[1]" /tmp/dna.json /tmp/extra.json > /etc/chef/dna.json || ( echo "jq not installed"; cp /tmp/dna.json /etc/chef/dna.json )
       {
         CINC_CMD="cinc-client --local-mode --config /etc/chef/client.rb --log_level info --logfile /var/log/chef-client.log --force-formatter --no-color --chef-zero-port 8889 --json-attributes /etc/chef/dna.json"
         pushd /etc/chef &&
--- a/cli/src/pcluster/templates/cluster_stack.py
+++ b/cli/src/pcluster/templates/cluster_stack.py
@@ -1592,6 +1592,22 @@ class ClusterCdkStack:
                     }
                 }
 
+        # LoginPools record Name (for DescribeLaunchTemplateVersions lookup) and LogicalId
+        # (filename suffix for the shared dna.json, must match the login node's own
+        # launch_template_id). Id/Version are omitted to avoid a CloudFormation circular dependency:
+        # head node LT metadata -> LoginNodes nested stack -> head node.
+        if self.config.login_nodes and self.config.login_nodes.pools:
+            lt_config["LoginPools"] = {
+                pool.name: {
+                    "LaunchTemplate": {
+                        "Name": f"{self._stack_name}-{pool.name}",
+                        "Version": "$Latest",
+                        "LogicalId": f"LoginNodeLaunchTemplate{create_hash_suffix(pool.name)}",
+                    }
+                }
+                for pool in self.config.login_nodes.pools
+            }
+
         return lt_config
 
     # -- Conditions -------------------------------------------------------------------------------------------------- #
--- a/cli/src/pcluster/templates/login_nodes_stack.py
+++ b/cli/src/pcluster/templates/login_nodes_stack.py
@@ -36,8 +36,6 @@ from pcluster.templates.cdk_builder_utils import (
 from pcluster.utils import (
     get_attr,
     get_http_tokens_setting,
-    get_resource_name_from_resource_arn,
-    get_service_endpoint,
     is_feature_supported,
 )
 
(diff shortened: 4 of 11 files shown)


cli/tests/pcluster/templates/test_cluster_stack.py
```diff
@@ -610,10 +616,7 @@ def assert_lt_properties(self, generated_template, resource_type):
                     # Managed instance profile
                     iam_instance_profile_name={"Ref": "InstanceProfileA50bdea9651dc48c"},
                     user_data_variables={
-                        "CloudFormationUrl": "https://cloudformation.us-east-1.amazonaws.com",
-                        "LaunchTemplateResourceId": "LoginNodeLaunchTemplateA50bdea9651dc48c",
-                        # Name of the managed role name within the managed instance profile
-                        "CfnInitRole": {"Ref": "RoleA50bdea9651dc48c"},
+                        "ProxyServer": "NONE",
```
gmarciani, line 619: Note for the reviewer: We added the check on ProxyServer because it's one of the user data variables that are always set and injected into the user data. Since we removed  `CloudFormationUrl`/`LaunchTemplateResourceId`/`CfnInitRole`, as consequence of removing dependency to cfn-init, we needed at least one variable to test that the injection is working.

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -129,79 +127,7 @@ def _add_login_nodes_pool_launch_template(self):
         ds_config = self._config.directory_service
         ds_generate_keys = str(ds_config.generate_ssh_keys_for_users).lower() if ds_config else "false"
 
-        if self._pool.instance_profile:
```
gmarciani, line 132: Note for the reviewer: we are removing this profile because it was declared just to be consumed by cfn-init. Since we removed cfn-init, there is no reason to keep it

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -299,65 +225,66 @@ def _add_login_nodes_pool_launch_template(self):
                     "launch_template_id": launch_template_id,
                 }
             },
-            indent=4,
+            indent=None,  # Keep indent as None for compact sizing and proper parsing in user_data.sh
         )
 
-        cfn_init = {
```
gmarciani, line 305: Note for the reviewer: we can entirely remove this because we removed dependency to cfn-init.

cli/src/pcluster/templates/cluster_stack.py
```diff
@@ -1592,6 +1592,23 @@ def _get_launch_templates_config(self):
                     }
                 }
 
+        # LoginPools entries intentionally avoid Fn::Ref/Fn::GetAtt to the LoginNodes nested
+        # stack: doing so would put this metadata (attached to the head node launch template)
+        # behind LoginNodes, which itself depends on the head node — creating a circular
+        # CloudFormation dependency. Instead, we emit deterministic identifiers (LaunchTemplateName
+        # and LogicalId) computed from the pool name; share_dna.py resolves the launch template
+        # at runtime via DescribeLaunchTemplates.
+        if self.config.login_nodes and self.config.login_nodes.pools:
```
gmarciani, line 1599: Note for the reviewer: we need this basic info about login nodes because the script `share_dna.py` in the cookbook (see https://github.com/aws/aws-parallelcluster-cookbook/pull/3188) needs it to retrieve the launch template user data and extract from it the dna.json body.

cli/src/pcluster/resources/login_node/user_data.sh
```diff
@@ -48,6 +48,16 @@ datasource_list: [ Ec2, None ]
 output:
   all: "| tee -a /var/log/cloud-init-output.log | logger -t user-data -s 2>/dev/ttyS0"
 write_files:
+  - path: /tmp/dna.json
```
gmarciani, line 51: Note for the reviewer: the insertion in this file are all coming from what cfn-=init used to do. Now that we removed cfn-init, we need the user data directive/body to do that.

tests/integration-tests/tests/update/test_update_race_conditions.py
```diff
@@ -386,46 +383,6 @@ def assert_update_recipe_executed_on_old_nodes(snapshot, update_submitted_at, ex
         ).is_not_empty()
 
 
-def is_update_failure_only_due_to_known_race_condition(rce, instance_ids: list, after_utc: str):
```
gmarciani, line 389: Note for the reviewer: we do not need this function anymore because by moving away of cfn-hup, we expect the so called Race Condition 5 to be solved, so there is no reason to recognize its failure mode and tolerate it.

CHANGELOG.md
```diff
@@ -19,6 +23,7 @@ CHANGELOG
 - Fix race condition in load balancer lookup by using `tag:GetResources` to resolve load balancer ARNs by tags, 
   which used to cause cluster update failure on clusters with login nodes.
 - Fail `pcluster build-image` early when the downloaded cookbook version does not match the ParallelCluster CLI version.
+- Fix login nodes not mounting `/opt/parallelcluster/shared` when EFS is used as the internal shared storage type.
```
gmarciani, line 26: Note for the review: this fix is required by the current change. The fix is shipped in aws/aws-parallelcluster-cookbook#3188

cli/src/pcluster/templates/cluster_stack.py
himani2411: Non-blocking: Can you update the comment to mention `CloudFormation deadlock`  or `CloudFormation circular dependency` rather than `CloudFormation cycle`? Ty!

himani2411: I understand that we do not want to add the deadlock; can you add the code snippet of where this deadlock happens? Why is the deadlock happening for LoginNode Stack and not for ComputeFleet?

If I am not wrong its because of https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/cluster_stack.py#L520-L521. Do we know why this is needed for LN and not ComputeFleet; from my understanding we need this dependency because of `head_eni=self._head_eni,` which is also needed in ComputeFleet stack, so why an explicit dependency is added only for LN Stack?

cli/src/pcluster/templates/login_nodes_stack.py
```diff
@@ -299,65 +225,66 @@ def _add_login_nodes_pool_launch_template(self):
...
+                                    get_attr(
+                                        self._config,
+                                        "dev_settings.timeouts.compute_node_bootstrap_timeout",
+                                        NODE_BOOTSTRAP_TIMEOUT,
+                                    )
+                                ),
+                                "AutoScalingGroupName": f"{self._login_nodes_stack_id}-AutoScalingGroup",
+                                "LaunchingLifecycleHookName": (
+                                    f"{self._login_nodes_stack_id}-LoginNodesLaunchingLifecycleHook"
+                                ),
```
himani2411, line 261: Note To Self: we removed below from LT because those were needed for Cfn-init command that we were running from `user_data.sh`; and since we are removing its usage we do not need it in LT.

```
"LaunchTemplateResourceId": launch_template_id,
                                "CloudFormationUrl": get_service_endpoint("cloudformation", self._config.region),
                                "CfnInitRole": instance_role_name,
```

cli/tests/pcluster/templates/test_cluster_stack.py
himani2411: Non-blocking: There is other Attributes we set like `LaunchingLifecycleHookName` or `AutoScalingGroupName`; can we add an assertion for those too?

cli/tests/pcluster/templates/test_login_nodes_stack.py
```diff
@@ -41,29 +43,47 @@ def test_login_nodes_dna_json(
...
-    assert_that(json.loads(login_node_extra_json["content"])).is_equal_to(expected_login_node_extra_json)
+    assert_that(json.loads(login_node_extra_json)).is_equal_to(expected_login_node_extra_json)
+
+    # Assertions on the cloud-init write_files directives that materialize dna.json and
+    # extra.json to check they are created with the correct ownership and permissions.
+    expected_owner = "root:root"
+    expected_permissions = "0644"
+
+    dna_json_directive = _get_write_files_directive(login_node_user_data_template, "/tmp/dna.json")
+    assert_that(dna_json_directive).is_not_none()
```
himani2411: Non-blocking: can we reduce code duplication by asserting the same checks inside of `_get_write_files_directive`?

tests/integration-tests/tests/update/test_update_race_conditions.py
himani2411: Its unclear based on the comment changes made for RaceCondition 5 whether the issue is solved after we remove reliance on cfn-hup. If we are not exposed to the racecondition can we update the comments to add a [SOLVED] Tag in the comments.

gmarciani: All the race conditions covered by this test are solved because we fixed them in 3.15.0 and with this change. The goal of the test is to verify that they are actually solved. Comment clarified

cli/tests/pcluster/templates/test_cluster_stack.py
gmarciani: Done

cli/tests/pcluster/templates/test_login_nodes_stack.py
gmarciani: Done

cli/src/pcluster/templates/cluster_stack.py
gmarciani: Both compute and login nodes, when started, require the head node to be running.
The difference between login nodes and compute nodes is when they are launched and this difference motivates the different dependency.

* Login nodes are launched as soon as the login nodes ASG is created. So when the ASG is created, the head node must be running. As such, the actual dependency is: ASG depends on Head Node. However, since ASG lives in nested stack, you cannot establish such cross stack dependencies (there may be hacky  tricks, but definitely out of scope for this PR). The result is that the entire login nodes nested stack must depend on the head node.
* Compute nodes are launched by the head node. The compute fleet nested stack creates the launch templates, that are then used by the head node to launch the compute nodes. So, in this case, the head node depends on the compute fleet stack.

himani2411 (approved):

PR #7416, opened 2026-06-01
Check that resolved path is still in dstdir for export logs

hgreebe:
### Description of changes
* Check that resolved path is still in dstdir for export logs
* Fixes bug where an attacker-controlled S3 object key could be used to escape the intended local export directory and overwrite an arbitrary file in the security context of the user or automation account running the export command.

### Tests
* Ran `pcluster export-cluster-logs`

### DIFF
--- a/cli/src/pcluster/models/common.py
+++ b/cli/src/pcluster/models/common.py
@@ -230,6 +230,10 @@ class CloudWatchLogsExporter:
             decompressed_path = decompressed_path.replace(
                 r"{unwanted_path_segment}{sep}".format(unwanted_path_segment=prefix, sep=os.path.sep), ""
             )
+            decompressed_path = os.path.realpath(decompressed_path)
+            if not decompressed_path.startswith(os.path.realpath(destdir) + os.sep):
+                LOGGER.warning("Skipping unsafe S3 key (path traversal detected): %s", archive_object.key)
+                continue
             compressed_path = f"{decompressed_path}.gz"
 
             LOGGER.debug("Downloading object with key=%s to %s", archive_object.key, compressed_path)


cli/src/pcluster/models/common.py
```diff
@@ -230,6 +230,10 @@ def _download_s3_objects_with_prefix(self, task_id, destdir):
             decompressed_path = decompressed_path.replace(
                 r"{unwanted_path_segment}{sep}".format(unwanted_path_segment=prefix, sep=os.path.sep), ""
             )
+            decompressed_path = os.path.realpath(decompressed_path)
+            if not decompressed_path.startswith(os.path.realpath(destdir) + os.sep):
+                LOGGER.warning("Skipping unsafe S3 key (path traversal detected): %s", archive_object.key)
```
gmarciani, line 235: "Path traversal" may be a bit cryptic/not-actionable for the end user. I suggest to make the message more explicit by mentioning the decompressed path and the destination path that it is escaping.

Also, I suggest to log this as an error because:
1. the effect is that the export does not complete, which is a pretty bad outcome for the user
2. the reason why it does not occur is related to a security issue

cli/src/pcluster/models/common.py
```diff
@@ -230,6 +230,10 @@ def _download_s3_objects_with_prefix(self, task_id, destdir):
             decompressed_path = decompressed_path.replace(
                 r"{unwanted_path_segment}{sep}".format(unwanted_path_segment=prefix, sep=os.path.sep), ""
             )
+            decompressed_path = os.path.realpath(decompressed_path)
```
gmarciani, line 233: [Testing] Can we cover this change with a unit test focused on provind the defense against the attack scenairo?

cli/src/pcluster/models/common.py
```diff
@@ -230,6 +230,10 @@ def _download_s3_objects_with_prefix(self, task_id, destdir):
...
+            if not decompressed_path.startswith(os.path.realpath(destdir) + os.sep):
```
gmarciani, line 234: What if destdir is passed by the user already with a trailing slash, e.g. /my/destination/dir/ ?
In that case even a legitimate decompressed path could fail the check.

hanwen-cluster:
Superseded by https://github.com/aws/aws-parallelcluster/pull/7441

PR #7418, opened 2026-06-02
[Test] Fix intermittent failure in test_dynamic_file_system_update.

gmarciani:
### Description of changes
Fix intermittent failure in test_dynamic_file_system_update.
The test could fail because the external EFS has chances to deploy mount targets in AZs that are not used by the cluster. 
With this fix we deploy the storage in the private subnets used by the compute nodes (one per AZ), so that the external EFS has a mount target in every AZ used by the cluster.

### Tests
SUCCEEDED
```
test-suites:
  update:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
```

### DIFF
--- a/cloudformation/storage/storage-stack.yaml
+++ b/cloudformation/storage/storage-stack.yaml
@@ -71,7 +71,8 @@ Parameters:
     Default: ''
   SubnetTwo:
     Description: ID of the Subnet (second Availability Zone) where the storage will be deployed.
-    Type: AWS::EC2::Subnet::Id
+    # The type has to be String to allow empty values.
+    Type: String
     Default: ''
   SubnetThree:
     Description: ID of the Subnet (third Availability Zone) where the storage will be deployed.
@@ -115,6 +116,9 @@ Parameters:
 Conditions:
   CreateEbs: !Equals [!Ref CreateEbs, 'true']
   CreateEfs: !Equals [!Ref CreateEfs, 'true']
+  CreateMountTargetResourceEfs0SubnetTwo: !And
+    - !Equals [!Ref CreateEfs, 'true']
+    - !Not [!Equals [!Ref SubnetTwo, '']]
   CreateMountTargetResourceEfs0SubnetThree: !And
     - !Equals [!Ref CreateEfs, 'true']
     - !Not [!Equals [!Ref SubnetThree, '']]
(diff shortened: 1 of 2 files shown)


hehe7318 (commented):
Looks not good to me to change none test file [cloudformation/storage/storage-stack.yaml](https://github.com/aws/aws-parallelcluster/pull/7418/changes#diff-fc385af4ca4c950e249e58d7341e8247dc093701674bea9159b10b790b9f77da), just for fixing intermittent test failure.
What do you think?

gmarciani:
> Looks not good to me to change none test file [cloudformation/storage/storage-stack.yaml](https://github.com/aws/aws-parallelcluster/pull/7418/changes#diff-fc385af4ca4c950e249e58d7341e8247dc093701674bea9159b10b790b9f77da), just for fixing intermittent test failure.
What do you think?

@hehe7318 that's an interesting point. I would agree with you if the change were not backward compatible. 
In this case, the change is backward compatible, we are just making the template able to deploy in single-AZ fashion.
We are making the template more flexible, we are not reducing its capabilities.

Also, let's think about the alternative approach where we do not make any change to the template. In this case we to deploy the storage stack with multi-AZ fashion forcefully, but using just one AZ, which basically means that we are paying to an EFS mount target that is never used.

What do you think?

hehe7318:
> > Looks not good to me to change none test file [cloudformation/storage/storage-stack.yaml](https://github.com/aws/aws-parallelcluster/pull/7418/changes#diff-fc385af4ca4c950e249e58d7341e8247dc093701674bea9159b10b790b9f77da), just for fixing intermittent test failure.
> > What do you think?
> 
> @hehe7318 that's an interesting point. I would agree with you if the change were not backward compatible. In this case, the change is backward compatible, we are just making the template able to deploy in single-AZ fashion. We are making the template more flexible, we are not reducing its capabilities.
> 
> Also, let's think about the alternative approach where we do not make any change to the template. In this case we to deploy the storage stack with multi-AZ fashion forcefully, but using just one AZ, which basically means that we are paying to an EFS mount target that is never used.
> 
> What do you think?

Convinced me! Approved!

hehe7318 (approved):

PR #7437, opened 2026-06-10
[Test] Allow execution of test dfsm at scale through the use of optional test dimension

gmarciani:
### Description of changes
Allow execution of test dfsm at scale through the use of optional test dimension.
When executed at scale the test deploys a cluster with:
* 10 login nodes pool, having 10 nodes each
* 50 queues with 10 nodes each

This is a medium cluster, not a large one (100 login nodes + 500 compute nodes). However, the critical scaling aspect we are trying to test is the retrieval of launch templates, which depends on the number of queues and pools not on the number of nodes. We are using hundreds of nodes just to stress more the system,. but the critical aspect for the update is the number of launch templates to retrieve.

With this PR we are introducing a new option al test config dimension `flags` that is meant to inject custom behaviors in tests. An example of usage is to execute a test at scale. 

### Tests
SUCCEEDED
```
test-suites:
  update:
      dimensions:
        - instances:
            - t3.large
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
          flags:
            - at_scale
```

### DIFF
--- a/tests/integration-tests/tests/update/test_update.py
+++ b/tests/integration-tests/tests/update/test_update.py
@@ -1068,9 +1068,10 @@ def remove_none_items(items: list):
 
 
 @pytest.mark.usefixtures("instance")
-def test_dynamic_file_systems_update(
+def test_dynamic_file_systems_update(  # noqa: C901
     region,
     os,
+    instance,
     pcluster_config_reader,
     ami_copy,
     clusters_factory,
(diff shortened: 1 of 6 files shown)


tests/integration-tests/tests/update/test_update.py
```diff
@@ -1082,12 +1083,32 @@ def test_dynamic_file_systems_update(
...
+    # When the "at_scale" flag is enabled, the test runs against a large cluster, i.e. 50 queues with 10 static
+    # compute nodes each and 10 login node pools with 100 login nodes per pool.
+    # Otherwise, it runs against a minimal cluster.
+    at_scale = "at_scale" in flags
+    login_nodes_pools_count = 10 if at_scale else 1
+    running_login_nodes_count = 10 if at_scale else 1
+    static_compute_min_count = 10 if at_scale else 1
+    static_compute_max_count = static_compute_min_count + 1
+    queue_count = 50 if at_scale else 2
+    head_node_instance_type = "c5.4xlarge" if at_scale else instance
```
hanwen-cluster, line 1099: If `at_scale` means 50 nodes, a `c5.xlarge` head node is enough

hanwen-cluster:
Great improvement on introducing "flags:"

hanwen-cluster (approved):

tests/integration-tests/tests/update/test_update.py
gmarciani: You are right. when I introduced the use of c4.xlarge I was using 2000 nodes. Then moved back to less nodes because of ICE and because that scale was overkilling. I'd leave the c4.xlarge so that when we need a bigger scale, the head node is already at the right size.

PR #7447, opened 2026-06-12
[DevSetting] Replace cinc_version/cinc_installer_url with CliAttributeOverrides which is similar to ExtraChefAttributes

himani2411:
### Description of changes
Replace cinc_version/cinc_installer_url with CliAttributeOverrides

* Model the CLI-side CINC overrides as a single freeform JSON DevSetting (CliAttributeOverrides), mirroring Cookbook.ExtraChefAttributes, instead of two scalar settings. The config exposes cinc_version/cinc_installer_url as read-only accessors that unfurl the blob, so imagebuilder_stack.py and parallelcluster.yaml keep populating CfnParamCincVersion/CfnParamCincInstaller unchanged.

* Add CliAttributeOverridesValidator for JSON well-formedness (object), with a KNOWN_CLI_ATTRIBUTE_OVERRIDES hook for optional future strict key validation.

* Add tests which cover CliAttributeOverridesValidator (valid object, empty/None, malformed JSON, non-object array/scalar) and the ImagebuilderDevSettings cinc_version/ cinc_installer_url accessors that unfurl the JSON blob (full, empty, partial, unknown-keys).

* CincVersion was added https://github.com/aws/aws-parallelcluster/pull/7420

### Tests
* Build Image

### References

### Checklist


cli/src/pcluster/schemas/imagebuilder_schema.py
himani2411, line 238: NOTE FOR REVIEWER: The cli_attribute_overrides can later be moved to BaseDevSettingsSchema when we add more overrides; since the 2 paramters are related to only Build Image I am keeping it in `ImagebuilderDevSettingsSchema`

cli/src/pcluster/validators/dev_settings_validators.py
```diff
@@ -19,6 +19,33 @@
...
+                f"Invalid value in {CLI_ATTRIBUTE_OVERRIDES_PATH}: must be a valid JSON string.",
+                FailureLevel.ERROR,
+            )
+            return
+        if not isinstance(attrs, dict):
+            self._add_failure(
+                f"Invalid value in {CLI_ATTRIBUTE_OVERRIDES_PATH}: must be a JSON object.",
+                FailureLevel.ERROR,
+            )
+        # TODO: optionally validate keys against KNOWN_CLI_ATTRIBUTE_OVERRIDES here.
```
himani2411: NOTE FOR REVIEWER: Keeping this as a future improvement since we do not validate all JSON content for ExtraChefAttributes as of today and the cinc attributes that I am replacing also do no have any validators. Any un-related attribute apart from the cinc* one's in the JSON will be a no-op.

cli/src/pcluster/validators/dev_settings_validators.py
```diff
@@ -19,6 +19,33 @@
 ATTR_CLUSTER_READINESS_CHECK_IGNORE_FAILURE = "cluster.cluster_readiness_check_ignore_failure"
 MIN_SLURM_RECONFIGURE_TIMEOUT = 300
 
+CLI_ATTRIBUTE_OVERRIDES_PATH = "DevSettings/CliAttributeOverrides"
+# Keys the CLI currently unfurls from the blob. Kept for an optional future
+# strict check; today validation is well-formedness only (like ExtraChefAttributes).
+KNOWN_CLI_ATTRIBUTE_OVERRIDES = {"cinc_version", "cinc_installer_url"}
```
gmarciani: I guess we can remove this constant because we are only interested in enforcing the json formatting of the `CliAttributeOverrides` without making eny assumption or enforcement about its field.

In Cookbook Chef Attribute we introduced the validation of specific attributes only because we decided to vend those attributes publicly to mitigate an impact, but in general we should not do it and keep it free form.

cli/src/pcluster/config/imagebuilder_config.py
```diff
@@ -251,14 +252,36 @@ def __init__(
         self.distribution_configuration = distribution_configuration
         self.terminate_instance_on_failure = Resource.init_param(terminate_instance_on_failure, default=True)
         self.disable_validate_and_test = Resource.init_param(disable_validate_and_test, default=True)
-        self.cinc_installer_url = Resource.init_param(cinc_installer_url, default="")
```
gmarciani, line 254: Removing this parameter from DevSettings schema is ok because we do not guarantee backward compatibility for DevSettings by definition. However, I suggest to document the removal in the changelog. the fact that we do not guarantee backward compatibility means it is ok to make this change, but does not mean we should not surface it.

himani2411: IF we never surface the addition of DevSettings in chnagelog; why should we surface their removal?

I propose we do not add this as part of changelog since we never document its existence.

cli/src/pcluster/validators/dev_settings_validators.py
himani2411: Updated

cli/src/pcluster/config/imagebuilder_config.py
gmarciani: The point of a changelog entry is to introduce opportunities and anticipate problems.
Even if we have never advertised the param in our changelog, nothing prevents a user from using it. 
Such user will face an error. Such error cannot be considered a breaking of a contract because it is an unofficial param, but still the entry will allow the user to avoid/motivate the failure.

If you disagree with that I understand because we are on a gray area.
In this case I suggest to let our PM have a call on this.

gmarciani (approved):

PR #7456, opened 2026-06-23
[Test] Add end2end test to cover cluster patching.

gmarciani:
### Description of changes
Add end2end test to cover cluster patching.

Minor changes made in this PR:
1. move the gpu workload script to a common folder so that it can be reused by this new test.

#### Test flow
1. Retrieve the AMI the cluster will use
2. Run EC2 instance with the source AMI, run patching commands and create a new patched AMI.
3. Create a cluster that uses the source AMI
4. Validate that the cluster is functional (GPU workload submitted from login node)
5. Stop the login nodes.
6. Update the cluster so that compute and login nodes use the patched AMI
7. Patch the head node in place and reboot it.
8. Start login nodes
9. Validate that the cluster is functional (GPU workload submitted from login node)

### Tests
ONGOING:

```
test-suites:
  patching:
      dimensions:
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_rhel8 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["rhel8"]
          schedulers: ["slurm"]
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_alinux2023 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_ubuntu2204 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["ubuntu2204"]
          schedulers: ["slurm"]
```


tests/integration-tests/tests/patching/test_patching.py
```diff
@@ -0,0 +1,232 @@
...
+      7.  Patch the head node in place and reboot it.
+      8.  Run the GPU workload again from the head node and login node, and assert the
+          kernel modules loaded before patching are still loaded on the head, compute
+          and login nodes.
+    """
+    ec2 = boto3.client("ec2", region_name=region)
+
+    # Retrieve the AMI the cluster will use *before* creation. Since the create
+    # config does not set a custom AMI, the cluster uses this default pcluster AMI.
+    base_ami = retrieve_latest_ami(region, os, ami_type="pcluster", request=request)
```
hanwen-cluster: Does this respect the "AMI_OWNER" specification in the testing infrastructure? If not, how about retrieving AMIs from the CloudFormation template of the first cluster creation?

tests/integration-tests/tests/patching/test_patching.py
```diff
@@ -0,0 +1,232 @@
...
+    wait_node_reachable(cluster, cluster.head_node_ip)
+
+    # GPU workload AFTER patching, from the head node and login node.
+    _run_gpu_workload(cluster, scheduler_commands_factory, use_login_node=False)
+    _run_gpu_workload(cluster, scheduler_commands_factory, use_login_node=True)
+
+    # Assert the kernel modules loaded before patching are still loaded on the head,
+    # compute and login nodes (which were all replaced or patched).
+    for node_type in NODE_TYPES:
+        _assert_kernel_modules_loaded(cluster, scheduler_commands_factory, node_type, kernel_modules_before)
```
hanwen-cluster: Are you sure the number of kernel modules don't change before and after the update?

gmarciani: Good point, what we want to enforce is that all the kernel modules loaded before the patching are still loaded after the patching (will fix the assertion). 

That said, it may be that this assertion will be anyways too hard because a kernel bump could legitimately invalidate some kernel modules. My proposal is to start with a hard assertion as long as it succeeds and we will relax if we face legitimate removals.

tests/integration-tests/tests/patching/test_patching.py
gmarciani: Godd catch, it may not. Fixing

hanwen-cluster (dismissed):

cloudformation/patching/ami-patching.yaml
```diff
@@ -0,0 +1,301 @@
...
+                    - uname -r
+                    - echo "Active kernel modules:"
+                    - lsmod
+              - name: ApplyPatches
+                action: ExecuteBash
+                inputs:
+                  commands:
+                    - aws s3 cp ${PatchScriptS3Uri} /tmp/patch_node.sh
+                    - chmod +x /tmp/patch_node.sh
+                    - /tmp/patch_node.sh
```
hanwen-cluster: Executing scripts in /tmp is not the best practice

gmarciani: Agree, moved to /usr/local/sbin

tests/integration-tests/tests/patching/test_patching.py
gmarciani: Fixed

hanwen-cluster (approved):

PR #7457, opened 2026-06-24
Extend MultiNetworkInterfacesInstancesValidator to cover single-card EFA instances

hehe7318:
### Description of changes

Since 3.15.0, single-network-card instances with EFA enabled (e.g. hpc6a, c5n) are launched with two network interfaces (a primary interface plus a dedicated efa-only interface), so AWS does not auto-assign them a public IP. Previously MultiNetworkInterfacesInstancesValidator only checked multi-network-card instances, so these queues passed validation and their compute nodes silently failed to bootstrap on public subnets without a NAT gateway.

Broaden the validator to also flag single-card EFA compute resources (mirroring the launch-template logic in queues_stack.add_network_interfaces, including the EfaInterfaceType: efa opt-out), and rework the failure messages to name the responsible instance types, explain why they launch multiple interfaces, and give a branch-specific workaround.
### Before

The validator only covered multi-network-card instances (`max_network_cards > 1`), so single-card EFA instances were **not flagged at all** — the cluster was created and compute nodes silently failed to bootstrap. When it did fire (multi-card instances), both branches emitted the same generic wording:

**`AssignPublicIp: true` branch:**
> The queue {queue_name} contains an instance type with multiple network interfaces however the AssignPublicIp value is set to true. AWS public IPs can only be assigned to instances launched with a single network interface.

**Public-IP subnet branch:**
> The queue {queue_name} contains an instance type with multiple network interfaces however the subnets {subnet_ids} is configured to automatically assign public IPs. AWS public IPs can only be assigned to instances launched with a single network interface.

### After

The message now names the responsible instance types, explains the reason per cause, reports both causes when a queue contains both, and gives a workaround that is correct for that branch.

#### Full messages by case

**Single-card EFA, public-IP subnet** (the reported customer case — `AssignPublicIp` not set, public subnet auto-assigns):
> The queue {queue_name} has EFA enabled on the single-network-card instance types {types}, so its compute nodes are launched with multiple network interfaces (a primary interface plus a dedicated efa-only interface), but the subnets {subnet_ids} auto-assign public IPs. AWS can't auto-assign a public IP to instances launched with more than one network interface. Use a private subnet with a NAT gateway to provide internet access to the compute nodes.

**Single-card EFA, `AssignPublicIp: true`:**
> The queue {queue_name} has EFA enabled on the single-network-card instance types {types}, so its compute nodes are launched with multiple network interfaces (a primary interface plus a dedicated efa-only interface), but AssignPublicIp is set to true. AWS can't auto-assign a public IP to instances launched with more than one network interface. Set AssignPublicIp to false and use a private subnet with a NAT gateway to provide internet access to the compute nodes.

**Multi-network-card, public-IP subnet:**
> The queue {queue_name} contains the instance types {types} with multiple network interfaces, but the subnets {subnet_ids} auto-assign public IPs. AWS can't auto-assign a public IP to instances launched with more than one network interface. Use a private subnet with a NAT gateway to provide internet access to the compute nodes.

**Multi-network-card, `AssignPublicIp: true`:**
> The queue {queue_name} contains the instance types {types} with multiple network interfaces, but AssignPublicIp is set to true. AWS can't auto-assign a public IP to instances launched with more than one network interface. Set AssignPublicIp to false and use a private subnet with a NAT gateway to provide internet access to the compute nodes.

**Both causes in the same queue, `AssignPublicIp: true`:**
> The queue {queue_name} contains the instance types {multi_card_types} with multiple network interfaces and has EFA enabled on the single-network-card instance types {efa_types}, so its compute nodes are launched with multiple network interfaces (a primary interface plus a dedicated efa-only interface), but AssignPublicIp is set to true. AWS can't auto-assign a public IP to instances launched with more than one network interface. Set AssignPublicIp to false and use a private subnet with a NAT gateway to provide internet access to the compute nodes.

### Checklist


charlie-chuang:
The added workaround is a real improvement — the existing message states the constraint but never tells the customer what to do. Let's keep that, but reduce to two variants. We don't need to differentiate single-card vs multi-card; "multiple network interfaces" is accurate for both and the fix is identical.

Subnet auto-assigns public IPs:

The queue {queue_name} contains instance types ({types}) that launch with multiple network interfaces, but the subnets {subnet_ids} auto-assign public IPs. AWS can't auto-assign a public IP to instances with more than one network interface. Use a private subnet with a NAT gateway to provide internet access to the compute nodes.

AssignPublicIp: true:

The queue {queue_name} contains instance types ({types}) that launch with multiple network interfaces, but AssignPublicIp is set to true. AWS can't auto-assign a public IP to instances with more than one network interface. Set AssignPublicIp to false and use a private subnet with a NAT gateway to provide internet access to the compute nodes.

Drop the single-card-specific and combined messages.

hehe7318:
> The added workaround is a real improvement — the existing message states the constraint but never tells the customer what to do. Let's keep that, but reduce to two variants. We don't need to differentiate single-card vs multi-card; "multiple network interfaces" is accurate for both and the fix is identical.
> 
> Subnet auto-assigns public IPs:
> 
> The queue {queue_name} contains instance types ({types}) that launch with multiple network interfaces, but the subnets {subnet_ids} auto-assign public IPs. AWS can't auto-assign a public IP to instances with more than one network interface. Use a private subnet with a NAT gateway to provide internet access to the compute nodes.
> 
> AssignPublicIp: true:
> 
> The queue {queue_name} contains instance types ({types}) that launch with multiple network interfaces, but AssignPublicIp is set to true. AWS can't auto-assign a public IP to instances with more than one network interface. Set AssignPublicIp to false and use a private subnet with a NAT gateway to provide internet access to the compute nodes.
> 
> Drop the single-card-specific and combined messages.

Thanks for the review and the advise! Hi I agree with most of your thought. Initially I have the same idea. The reason why I split single-card and multi-card is:
I want customer to know why a single-card instance have multiple network interfaces -> Normally by default it should have one.
So I split and explain -> `has EFA enabled on the single-network-card instance types {types}, so its compute nodes are launched with multiple network interfaces (a primary interface plus a dedicated efa-only interface)`
Or I think customer will be confused. Let me know your thought.

hanwen-cluster (approved):

PR #7461, opened 2026-06-29
[integ-tests-framework] Make capacity reservations for all instance types

hanwen-cluster:
### Description of changes
1. With https://github.com/aws/aws-parallelcluster/pull/7440, we started to make capacity reservations for {"c5.xlarge", "m6g.xlarge", "m6i.xlarge"}, and use other similar instance types if a capacity reservation fails to creat. This commit expands the logic to all instance types. 
1.1. With instance types <= .xlarge, we make duplicate capacity reservations because multiple tests in parallel could use the same instance types, therefore need multiple capacity reservations. With instance types >.xlarge, we make only one capacity reservation because tests with larger instance types usually make capacity reservations early in the test definition (e.g. test_efa in commercial makes capacity reservation in `develop.yaml`), therefore this second layer of capacity reservation shouldn't make duplicate capacity reservations. 
1.2. With instance types supporting EFA, create the capacity reservation in a placement group. With instance types not supporting EFA, create the capacity reservation without a placement group.
3. With this commit, resolve_instance_with_capacity allows specifying alternative_instance_types. Prior to this commit alternative_instance_types was always calculated with `get_similar_instance_types`, which could be too restrictive, so don't give too many alternatives for instance types like `c5n.18xlarge`
4. Improve test_efa in isolated_regions to take a flag to use any efa instances to avoid Insufficient Capacity Error. test_efa in commercial doesn't need this, because it could try out different regions. In isolated regions, the test has to run in a specific region.

### Tests
```
test-suites:
  efa:
    test_efa.py::test_efa:
      dimensions:
        - regions: ["ap-southeast-5"]
          instances: ["c5n.18xlarge"]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
          flags: ["any-efa-instances"]
        - regions: ["us-east-1"]
          instances: ["c5n.18xlarge"]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
```
In the above tests, the test in us-east-1 passed completely. The test in ap-southeast-5 failed some checks in fabtest because it was using g6.8xlarge. This failure is not a regression from this PR, and won't surface in isolated regions because fabtest is not run in isolated regions.

### Checklist


tests/integration-tests/tests/common/capacity_helpers.py
```diff
@@ -13,52 +13,126 @@
...
+        logging.warning("Could not list existing capacity reservations for %s: %s", instance_type, e)
+        return False
 
-def resolve_instance_with_capacity(region, az_id, instance_type, os, minutes=50, count=2):
-    """Try to reserve capacity for *instance_type* in *az_id*, falling back to alternatives on ICE.
+    for reservation in reservations:
+        if (
+            reservation.get("AvailabilityZoneId") == az_id
+            and reservation.get("InstancePlatform") == instance_platform
+            and bool(reservation.get("PlacementGroupArn")) == with_placement_group
```
hehe7318, line 36: [Blocking] Should we also check the CapacityResevations available capacity? Or is it handled in other place?

tests/integration-tests/tests/common/capacity_helpers.py
```diff
@@ -13,52 +13,126 @@
...
-    ``_try_reserve_head_node_instance`` in test_efa.py.  A successful reservation
-    guarantees capacity for the subsequent cluster launch and auto-expires after
-    *minutes* minutes.
+
+def _ensure_cluster_placement_group(ec2_client, name):
+    """Return the ARN of cluster placement group *name*, creating it if it does not exist."""
+    try:
+        return ec2_client.describe_placement_groups(GroupNames=[name])["PlacementGroups"][0]["GroupArn"]
+    except Exception:
+        return ec2_client.create_placement_group(GroupName=name, Strategy="cluster")["PlacementGroup"]["GroupArn"]
```
hehe7318, line 79: Where do we cleanup these PlacementGroup?

tests/integration-tests/tests/common/capacity_helpers.py
hanwen-cluster: This second layer of capacity reservation doesn't have visibility into how many instances are required by a test. So we just check a capacity reservation exists and don't care about the available capacity. If a test is peculiar on the available capacity, it should make explicit capacity reservation in `develop.yaml`

tests/integration-tests/tests/common/capacity_helpers.py
hanwen-cluster: We don't clean up. Since the naming of the placement groups is `{instance_type}_placement_group_{az_id}`, we won't create too many even without cleaning up. [AWS limits 500 placement groups per region](https://repost.aws/questions/QUV59mDpbkTNO3ZEGu1oRwmQ/placement-groups), which is safe for us

tests/integration-tests/tests/common/capacity_helpers.py
hehe7318: If multiple tests run in parallel using the same ODCR, is there a risk of consuming all capacity and causing an ICE error?

tests/integration-tests/tests/common/capacity_helpers.py
hehe7318: Okay, makes sense, so these PlacementGroup are long live, and can avoid re-creating everytime.

tests/integration-tests/tests/common/capacity_helpers.py
hanwen-cluster: Yes, there is the risk. So in `develop.yaml`, we should specify how many instances the test needs considering the parallelism

tests/integration-tests/tests/common/capacity_helpers.py
```diff
@@ -13,52 +13,126 @@
...
+                reservation.get("CapacityReservationId"),
+                instance_type,
+                az_id,
+                with_placement_group,
+            )
+            return True
+    return False
+
+
+def get_efa_instance_types(region):
```
hehe7318: As previously discussed, let's add a filter here to restrict ARM instance types.

hanwen-cluster: Done

hehe7318 (approved):

PR #7463, opened 2026-06-30
[Test] Add integ-test for clustermgtd instance-ID matching under EC2 eventual consistency

hehe7318:
### Description of changes
Add `test_clustermgtd_instance_id_matching`, a regression test for the EC2 DescribeInstances eventual-consistency issue where instances are returned with a missing PrivateIpAddress. clustermgtd must keep such instances, match them to Slurm nodes by InstanceId, and leave the healthy nodes in place instead of replacing them by IP matching.

The fault is simulated on the head node by setup_missing_private_ip_override.sh, which wraps common.ec2_utils.get_private_ip_address_and_dns_name to raise KeyError('PrivateIpAddress') for a chosen instance, exactly as a real DescribeInstances response with a missing PrivateIpAddress would. The wrapper is toggled via a trigger file and gated so the fault can be turned on/off without reinstalling.

Also add an assertion in the scaling stress test to guard against the 'not all EC2 info are available' discard path reappearing at scale.

### Tests

### References

### Checklist


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -441,6 +441,102 @@ def test_error_handling(
...
+    """
+    cluster_config = pcluster_config_reader()
+    cluster = clusters_factory(cluster_config)
+    remote_command_executor = RemoteCommandExecutor(cluster)
+    scheduler_commands = scheduler_commands_factory(remote_command_executor)
+    # Trigger file consumed by setup_missing_private_ip_override.sh to toggle the simulated fault.
+    trigger_file = "/tmp/ec2_eventual_consistency_instances"
+
+    partition = "queue1"
+    num_static_nodes = 2
```
gmarciani: [minor/not-blocking] inject `num_static_nodes` into the cluster config as jinja variable so that you make sure that the number of nodes is always aligned. If in future we must change the number of nodes, we will only need to change that variable, rather than that variable + the config file.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -441,6 +441,102 @@ def test_error_handling(
...
+    )
+
+    remote_command_executor.clear_clustermgtd_log()
+
+    # Restart is required because clustermgtd does not hot-reload python code.
+    remote_command_executor.run_remote_script(
+        str(test_datadir / "setup_missing_private_ip_override.sh"), run_as_root=True
+    )
+    remote_command_executor.run_remote_command(f"echo '{target_instance_id}' | sudo tee {trigger_file}")
+    remote_command_executor.run_remote_command("sudo systemctl restart supervisord")
```
gmarciani: Restarting supervisord would restart all the daemons it manages.
This is not necessayr. Better to keep this restart focused and restart only clustermgtd instead.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -441,6 +441,102 @@ def test_error_handling(
...
+
+    retry(wait_fixed=seconds(20), stop_max_delay=minutes(5))(assert_lines_in_logs)(
```
gmarciani: A healthy restart would happen in seconds. Let's reduce `stop_max_delay` to 1 minute.
If it takes more than 1 minute to restart, then there is something wrong going on and we do not want it to be masked by a logn retry.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -441,6 +441,102 @@ def test_error_handling(
...
+        remote_command_executor, ["/var/log/parallelcluster/clustermgtd"], ["ClusterManager Startup"]
+    )
+    retry(wait_fixed=seconds(20), stop_max_delay=minutes(5))(assert_lines_in_logs)(
+        remote_command_executor,
+        ["/var/log/parallelcluster/clustermgtd"],
+        [f"Incomplete EC2 info for instance {target_instance_id}, keeping it for instance-ID matching"],
+    )
+
+    # Let clustermgtd run several more iterations; a regression would have replaced the node by now.
+    time.sleep(4 * loop_time)
```
gmarciani: Why 4 iterations? Is there some other evidence we can wait rather than using a magic number, e.g. a specific log line?

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
```diff
@@ -0,0 +1,71 @@
...
+# See the License for the specific language governing permissions and limitations under the License.
+#
+# Simulate EC2 DescribeInstances eventual consistency. clustermgtd's describe_instances path has no
+# override hook, so we wrap common.ec2_utils.get_private_ip_address_and_dns_name to raise
+# KeyError("PrivateIpAddress") for any instance id listed in the trigger file, exactly as a real
+# DescribeInstances response with a missing PrivateIpAddress would.
+#
+# The wrapper is module-global (also used by fleet_manager) and toggled via the trigger file, which only
+# ever lists already-running instance ids. clustermgtd does not hot-reload python, so the caller must
+# restart supervisord once after this script; toggling the trigger file afterwards needs no restart.
```
gmarciani: [minor] it is not required to restart supervisord. it is only required to restart clustermgtd.

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
```diff
@@ -0,0 +1,71 @@
...
+# leave clustermgtd importing a broken module. The heredoc is unquoted to expand ${TRIGGER_FILE}/${MARKER},
+# so keep the Python below free of shell metacharacters (or escape them as \$).
+tmp_file=$(mktemp)
+sudo cat "${ec2_utils_path}" > "${tmp_file}"
+cat << EOF >> "${tmp_file}"
+
+${MARKER}
+import os as _ec_os  # noqa: E402
+
+_ec_original_get_private_ip_address_and_dns_name = get_private_ip_address_and_dns_name
```
gmarciani: what does the `_ec_` prefix mean?

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
```diff
@@ -0,0 +1,71 @@
...
+    # Read the target instance ids from a file so the fault can be toggled without reloading the module.
+    if _ec_os.path.exists(_EC_TRIGGER_FILE):
+        with open(_EC_TRIGGER_FILE) as _trigger:
+            _targets = _trigger.read().split()
+        if instance_info.get("InstanceId") in _targets:
+            raise KeyError("PrivateIpAddress")
+    return _ec_original_get_private_ip_address_and_dns_name(instance_info)
+EOF
+
+python3 -c "import ast, sys; ast.parse(open(sys.argv[1]).read())" "${tmp_file}"
```
gmarciani, line 71: [minor] We execute this line to validate that the resulting patched script is syntactically valid. That's a very good approach. Let's add a comment/echo describing what we are doing to clarify as it is not obvious.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -441,6 +441,102 @@ def test_error_handling(
...
+    clustermgtd_conf_path = retrieve_clustermgtd_conf_path(remote_command_executor)
+    remote_command_executor.run_remote_command(
+        f"sudo sed -i '/^loop_time /d' {clustermgtd_conf_path}; "
+        f"echo 'loop_time = {loop_time}' | sudo tee -a {clustermgtd_conf_path}"
+    )
+
+    remote_command_executor.clear_clustermgtd_log()
+
+    # Restart is required because clustermgtd does not hot-reload python code.
+    remote_command_executor.run_remote_script(
```
gmarciani, line 487: Let's make sure that the script gives an evidence of its progress. If the script fails, we want from the test logs to be clear where and why

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Thanks for catching this. Done. I extracted one existing supervisorctl-path lookup function into a shared `retrieve_supervisorctl_path` helper in `utils.py`.

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done! Thanks for catching!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done. Replaced the fixed sleep with waiting until the `Incomplete EC2 info for instance ...` line appears at least 4 times.

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
hehe7318: Done

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
hehe7318: Eventual consistency. I removed them.

tests/integration-tests/tests/schedulers/test_slurm/test_clustermgtd_instance_id_matching/setup_missing_private_ip_override.sh
hehe7318: Done. Added an `echo`.

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: The script already runs with `set -ex` so every command is traced, but I agree with you. Added progress echoes at each step (locate → syntax check → install).

gmarciani (approved):

PR #7460, opened 2026-06-29
[Release] Add upload-node.py to push aws-parallelcluster-node to S3 

himani2411:
### Description of changes

This is required as part of our release process as we move from uploading node package from s3 https://github.com/aws/aws-parallelcluster-cookbook/pull/3211

Multi-region S3 uploader for the node package, modeled on upload-cookbook.py. Uploads to s3://<region>-aws-parallelcluster/parallelcluster/<version>/node/, which is where the cookbook now downloads the node package from in all regions (see parallelcluster_node.rb). Invoked by the Jenkins push_node_s3 release job.

### Tests
* tested using Dry-run option
* Successful testing on personal S3 bucket

### References
* Link to impacted open issues.
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7476, opened 2026-07-07
[Release] Address code-duplication for release specific modules

himani2411:
### Description of changes
* Address code-duplication from https://github.com/aws/aws-parallelcluster/pull/7460/changes#r3501452428

### Tests
* Ran scripts using Dry run

### References

### Checklist

### DIFF
--- a/tests/integration-tests/conftest.py
+++ b/tests/integration-tests/conftest.py
@@ -796,6 +796,27 @@ def _inject_additional_iam_policies_for_nodes(
             _inject_additional_iam_policies(pool, policies)
 
 
+def _build_private_os_ami_filters(request):
+    """Build DescribeImages tag filters that pin a PRIVATE_OSES AMI to the git refs under test.
+
+    PRIVATE_OSES bypass AmiSearchFilters (CustomAmi wins), so without these filters
+    retrieve_latest_ami returns the newest pcluster AMI regardless of the refs under test.
+    Returns an empty list when no git refs are provided (released-version runs).
+    """
+    filters = []
+    for git_ref_option, tag_key in (
+        ("pcluster_git_ref", "build:parallelcluster:cli_ref"),
+        ("cookbook_git_ref", "build:parallelcluster:cookbook_ref"),
+        ("node_git_ref", "build:parallelcluster:node_ref"),
+    ):
+        git_ref = request.config.getoption(git_ref_option)
+        if git_ref:
+            filters.append({"Name": f"tag:{tag_key}", "Values": [git_ref]})
+    if filters:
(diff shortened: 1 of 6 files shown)


tests/integration-tests/conftest.py
```diff
@@ -845,34 +866,14 @@ def inject_additional_config_settings(  # noqa C901
...
         dict_add_nested_key(
             config_content,
             retrieve_latest_ami(
                 region,
                 config_content["Image"]["Os"],
                 ami_type="pcluster",
                 architecture=architecture,
-                additional_filters=additional_filters,
-                request=request,
-                allow_private_ami=True,
```
hehe7318, line 875: Why remove these two para? `request` and `allow_private_ami`

util/common.py
```diff
@@ -24,7 +24,18 @@
 
 def get_aws_regions(partition):
     ec2 = boto3.client("ec2", region_name=PARTITION_TO_MAIN_REGION[partition])
-    return set(r.get("RegionName") for r in ec2.describe_regions().get("Regions"))
+    return set(sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions")))
```
hehe7318, line 27: Why sort when add it in a set?

tests/integration-tests/conftest.py
himani2411: `request` and `allow_private_ami ` only affect one branch in retrieve_latest_ami — the one that appends the` is-public=true ` filter — and that branch also requires `os not in PRIVATE_OSES`. Since this call is inside the Os in PRIVATE_OSES block, that branch
  can never run here, so both params were no-ops for this call. The git-ref reading they used to enable now lives in _build_private_os_ami_filters(request), which builds the ref tag-filters directly. So removing them is a no-op cleanup, not a  behavior change.

util/common.py
himani2411: I picked the exact function as is from `util/upload-cookbook.py`  to keep the move faithful; agreed the sorted() is a no-op on a set — happy to drop it here.

```
def _get_all_aws_regions(region):
    return set(sorted(r.get("RegionName") for r in ec2.describe_regions().get("Regions")))
```

tests/integration-tests/conftest.py
hehe7318: Thanks for explaining, make sense.

util/common.py
hehe7318: Okay, keeping it does not affecting anything, we can keep.

hehe7318 (approved):
LGTM! Thanks for reduce the code duplication.

PR #7490, opened 2026-07-16
[IAM] Relax login nodes tag condition in public policies to prevent cluster creation/update failure

gmarciani:
#### Context
ParallelCluster relies on an IAM policy that has become too restrictive following a recent AWS service-side change to the CloudFormation ELBv2 Listener resource. When a cluster hits this issue Clusters that include login nodes fail to create or update when deployed through the ParallelCluster API or the CloudFormation Custom Resource.

### Changes
* Fix cluster creation and update failures on clusters with login nodes,  caused by the IAM policy requiring the `parallelcluster:cluster-name` tag to be the only tag instead of just being present.
* Add a login-nodes pool to the sample cluster deployed by the 1-click custom-resource template. This increases feature coverage so the 1-click path also provisions login nodes.

### Notes
This change is necessary and sufficient to resolve the impact described in https://github.com/aws/aws-parallelcluster/issues/7491. We'll consider a broader relaxation later for better future-proofing.

### Tests
* Manually validate deploying a cluster with login nodes through custom resource.
* SUCCESS
```
test-suites:
  pcluster_api:
    test_api.py::test_login_nodes:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["c5.xlarge"]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
  custom_resource:
    # Create a cluster via the pcluster CFN custom resource
    test_cluster_custom_resource.py::test_cluster_create:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["c5.xlarge"]
          oss: ["alinux2023"]
```

### DIFF
--- a/cloudformation/custom_resource/cluster-1-click.yaml
+++ b/cloudformation/custom_resource/cluster-1-click.yaml
@@ -82,6 +82,18 @@ Resources:
             Networking:
               SubnetIds:
               - !GetAtt [ PclusterVpc, Outputs.PrivateSubnetId ]
+        LoginNodes:
+          Pools:
+            - Name: login
+              InstanceType: t3.small
+              Count: 1
+              Networking:
+                SubnetIds:
+                  - !GetAtt [ PclusterVpc, Outputs.PublicSubnetId ]
+              Iam:
+                AdditionalIamPolicies:
+                  - Policy: !Sub arn:${AWS::Partition}:iam::aws:policy/AmazonSSMManagedInstanceCore
+              GracetimePeriod: 3
 
 Outputs:
   HeadNodeIp:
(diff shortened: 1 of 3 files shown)


cloudformation/custom_resource/cluster-1-click.yaml
```diff
@@ -82,6 +82,18 @@ Resources:
             Networking:
               SubnetIds:
               - !GetAtt [ PclusterVpc, Outputs.PrivateSubnetId ]
+        LoginNodes:
+          Pools:
+            - Name: login
+              InstanceType: t3.small
```
hanwen-cluster: Shall we use at least t3.medium to avoid intermittent failures on large OS (e.g. RHEL9 and Ubuntu 24)?

gmarciani: Good point. Just noticed we also use t3.micro in compute nodes for this template. Changed to t3.medium for both.

hanwen-cluster (approved):

PR #7501, opened 2026-07-23
[Test] Add integ-test for user Slurm maintenance reservations on static and dynamic nodes

hehe7318:
### Description of changes
Add test_slurm_maintenance_reservation verifying that a user-created Slurm reservation with flags=maint follows standard Slurm semantics: it keeps unreserved jobs off the reserved nodes but does not change node power state or prevent launching.

On a single cluster (1 static + 2 dynamic) the test checks both node types:
- static node: an unreserved job pinned to it stays pending, and when the node becomes unhealthy clustermgtd replaces it once and lets the replacement bootstrap back to idle while keeping its reservation (no kill/relaunch loop);
- dynamic node: an unreserved job stays pending and the node is not powered up, a `--reservation` job resumes the node and runs to completion;
- for both node types, once the reservation is deleted the node is no longer MAINTENANCE+RESERVED and the job that was left pending runs to completion.

Also adds `assert_msg_in_log_at_most` to the shared test assertions, as the counterpart to the existing `assert_msg_in_log_at_least`. The static-node section uses both to assert clustermgtd logs the replacement exactly once, which is what distinguishes a single legitimate replacement from a kill/relaunch loop.

### Tests
Verified on a real cluster (us-east-1, 1 static + 2 dynamic nodes):
* **Dynamic node section: all checks pass.** Reserved node stays powered down and blocks unreserved jobs, a `--reservation` job resumes the node and completes on it, the reservation is cleared after deletion, and the previously pending job then runs to completion.
* **Static node section: the pre-failure checks pass and the known bug (GitHub issue #7470) reproduces**, as the fix is not merged yet, the test is red now, and it should go green as soon as the fix is merged. The pending job left on the node did run to completion once the reservation was deleted.

### Checklist

### DIFF
--- a/tests/integration-tests/tests/schedulers/test_slurm.py
+++ b/tests/integration-tests/tests/schedulers/test_slurm.py
@@ -26,6 +26,7 @@ from utils import (
     check_status,
     get_compute_nodes_instance_ids,
     get_instance_info,
+    get_username_for_os,
     retrieve_clustermgtd_conf_path,
     retrieve_supervisorctl_path,
     set_protected_failure_count,
@@ -38,6 +39,7 @@ from tests.common.assertions import (
     assert_lines_in_logs,
     assert_msg_in_log,
     assert_msg_in_log_at_least,
+    assert_msg_in_log_at_most,
     assert_no_errors_in_logs,
     assert_no_msg_in_logs,
     assert_no_node_in_ec2,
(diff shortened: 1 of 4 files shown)


tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+    )
+
+    assert_no_errors_in_logs(remote_command_executor, "slurm", skip_ice=True)
+
+
+def _verify_maint_reservation_on_static_node(remote_command_executor, scheduler_commands, static_node, partition):
+    """Reserved static node keeps unreserved jobs off, but clustermgtd still replaces it once it goes unhealthy."""
+    clustermgtd_log = "/var/log/parallelcluster/clustermgtd"
+    reservation_name = "usermaint_st"
+    _create_user_maintenance_reservation(
```
hgreebe: I think you should move this function outside of `_verify_maint_reservation_on_static_node` function because it makes the flow of the test less readable. In `test_slurm_maintenance_reservation` it looks like you skip to verifying the maint reservation before creating it because the creation happens within the verify function.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+def _assert_normal_job_stays_pending_on_reserved_node(scheduler_commands, reserved_node, partition):
+    """Submit a job pinned to the reserved node with no --reservation and confirm it stays pending, then cancel."""
+    logging.info("Submitting a normal job that should not be scheduled onto reserved node %s", reserved_node)
+    result = scheduler_commands.submit_command("sleep 600", nodes=1, partition=partition, host=reserved_node)
+    job_id = scheduler_commands.assert_job_submitted(result.stdout)
+    # Poll a few times so a transient pending-before-scheduling does not pass for a lasting block.
+    for _ in range(4):
+        assert_that(scheduler_commands.get_job_info(job_id, field="JobState")).is_equal_to("PENDING")
+        time.sleep(15)
+    scheduler_commands.cancel_job(job_id)
```
hgreebe: Why do we need to cancel the job? Shouldn't we leave it pending and then when the reservation is disabled we check that the pending job gets scheduled.

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+    # The reservation must survive the replacement, not be dropped along with the old instance.
+    assert_that(recovered_flags).contains("MAINTENANCE", "RESERVED")
+
+    # clustermgtd logs this once per replacement round: expect exactly one legitimate replacement, whereas a
+    # kill/relaunch loop would log it many times.
+    replacement_msg = "Terminating instances backing unhealthy static nodes"
+    assert_msg_in_log_at_least(remote_command_executor, clustermgtd_log, replacement_msg, count=1)
+    assert_msg_in_log_at_most(remote_command_executor, clustermgtd_log, replacement_msg, count=2)
+
+    _delete_slurm_reservation(remote_command_executor, reservation_name)
```
hgreebe: This should also be moved outside of this function

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+
+
+def _verify_maint_reservation_on_dynamic_node(
+    remote_command_executor, scheduler_commands, cluster, region, dynamic_node, partition, num_static_nodes, user
+):
+    """Reserved dynamic node stays powered down for unreserved jobs but resumes for a --reservation job."""
+    reservation_name = "usermaint_dy"
+    # Authorize the submitting user so it can later target the reservation with --reservation.
+    _create_user_maintenance_reservation(
```
hgreebe: Same a previous comment about keeping this outside of the function

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+    )
+
+    # Deleting the reservation must restore normal scheduling on the node.
+    _delete_slurm_reservation(remote_command_executor, reservation_name)
+
+    @retry(wait_fixed=seconds(10), stop_max_delay=minutes(1))
+    def _assert_reservation_cleared():
+        assert_that(_get_node_state_flags(scheduler_commands, dynamic_node)).does_not_contain("MAINTENANCE", "RESERVED")
+
+    _assert_reservation_cleared()
```
hgreebe: why are we asserting reservation cleared for dynamic nodes but not static nodes?

tests/integration-tests/tests/schedulers/test_slurm.py
```diff
@@ -540,6 +542,212 @@ def _assert_slurm_nodes_have_instance_id(scheduler_commands, nodes, expected_ins
...
+    )
+    _verify_maint_reservation_on_dynamic_node(
+        remote_command_executor, scheduler_commands, cluster, region, dynamic_nodes[0], partition,
+        num_static_nodes, get_username_for_os(os)
+    )
+
+    assert_no_errors_in_logs(remote_command_executor, "slurm", skip_ice=True)
+
+
+def _verify_maint_reservation_on_static_node(remote_command_executor, scheduler_commands, static_node, partition):
```
hgreebe: We should check that the pending jobs run after the reservation is cleared

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Good catch, done!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done!

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Good catch, both paths assert it now.

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Done — asserted for both static and dynamic nodes.

tests/integration-tests/tests/schedulers/test_slurm.py
hehe7318: Good point, done — the job now stays pending and we assert it runs after the reservation is deleted.

hgreebe (approved):

PR #7510, opened 2026-07-29
Fix flaky failures during creating CFN stack from proxy.yaml

hanwen-cluster:
### Description of changes
The proxy.yaml CFN stack creation sporadically failed with "WaitCondition timed out. Received 0 conditions when expecting 1" on ProxyVerificationWaitCondition.

Root cause is a startup race. `ProxyClient` runs in the private subnet and reaches the internet only through the Proxy instance, whose transparent-proxy plumbing is set up by its UserData. `ProxyClient` is created after `Proxy`, which CloudFormation satisfies as soon as the `Proxy` instance is running, not when its UserData finishes.

When the `Proxy`'s own "apt-get update" is slow (e.g. a slow security.ubuntu.com mirror), `ProxyClient` would fail connecting to network.

Console logs confirmed the race: in a failing run the Proxy did not add the iptables REDIRECT rules until 04:51:31, while the ProxyClient's apt window ran 04:50:09-04:50:42 and failed ~49s too early. On runs where the mirror is fast, the Proxy finishes first and verification passes.

Fix: add ProxyReadyWaitConditionHandle/ProxyReadyWaitCondition. The Proxy signals this handle as the final step of its UserData. This removes the race regardless of how long the Proxy's apt takes.

### Tests
* test_proxy has been passed

### References

### Checklist

### DIFF
--- a/cloudformation/proxy/proxy.yaml
+++ b/cloudformation/proxy/proxy.yaml
@@ -467,6 +467,10 @@ Resources:
             echo iptables-persistent iptables-persistent/autosave_v4 boolean true | sudo debconf-set-selections
             echo iptables-persistent iptables-persistent/autosave_v6 boolean true | sudo debconf-set-selections
             apt-get $APT_RETRY install -y iptables-persistent
+
+            curl -X PUT -H "Content-Type:" \
+              --data-binary '{"Status":"SUCCESS","Reason":"Proxy ready","UniqueId":"Proxy","Data":"OK"}' \
+              "${ProxyReadyWaitConditionHandle}"
       Tags:
         - Key: Name
           Value: !Sub [ "Proxy-${StackIdSuffix}", {StackIdSuffix: !Select [1, !Split ['/', !Ref 'AWS::StackId']]}]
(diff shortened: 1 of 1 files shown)


cloudformation/proxy/proxy.yaml
```diff
@@ -479,13 +483,27 @@ Resources:
...
+      Timeout: 600
+
   # PROXY VERIFICATION CLIENT
   # In cluster mode: tests explicit proxy (http_proxy env var) by downloading from GitHub.
   # In build-image mode: tests transparent proxy (iptables REDIRECT) by curling allowlisted domains.
   ProxyClient:
     Type: AWS::EC2::Instance
     DependsOn:
       - Proxy
+      - ProxyReadyWaitCondition
```
gmarciani, line 506: [minor] It is enough now to depend on the wait condition only

gmarciani (approved):

cloudformation/proxy/proxy.yaml
hanwen-cluster: Yes, I thought about it. But why not to be robust to avoid another type of race condition?

PR #7512, opened 2026-07-29
[Test] In test_patching_cluster execute the lustre refresh on Ubuntu the same way we do on RHEL/Rocky, to prevent failures on kernel bump.

gmarciani:
### Description of changes
[Test] In test_patching_cluster execute the lustre refresh on Ubuntu the same way we do on RHEL/Rocky, to prevent failures on kernel bump.

### Tests
The test fails as expected now because the refresh finds not available Lustre client for the new kernel version

```
test-suites:
  patching:
      dimensions:
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_ubuntu2404 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["ubuntu2204","ubuntu2404"]
          flags: ["patching:minimal"]
          schedulers: ["slurm"]
```

### DIFF
--- a/tests/integration-tests/tests/patching/test_patching/test_patching_cluster/patch_node.sh
+++ b/tests/integration-tests/tests/patching/test_patching/test_patching_cluster/patch_node.sh
@@ -44,7 +44,7 @@ fix_rpmdb() {
 # If the upgrade no-ops (same version, different .ko) reinstall to swap the module, 
 # then fail if none resolves for the new kernel.
 # https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
-refresh_lustre_client() {
+refresh_lustre_client_rhel() {
     [[ -f /etc/yum.repos.d/aws-fsx.repo ]] || return 0
     local new_kernel
     new_kernel=$(rpm -q kernel --qf '%{VERSION}-%{RELEASE}.%{ARCH}\n' | sort -V | tail -n1)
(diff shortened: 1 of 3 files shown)


tests/integration-tests/tests/patching/test_patching/test_patching_cluster/patch_node.sh
```diff
@@ -58,6 +58,24 @@ refresh_lustre_client() {
...
+# On Ubuntu the FSx Lustre client kernel module ships as a per-kernel package
+# (lustre-client-modules-<uname-r>), so a patching kernel bump leaves no matching
+# module for the new kernel about to boot.
+# Before the reboot, install the target kernel's module package, mirroring the
+# cookbook's Ubuntu lustre setup.
+# See https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
+refresh_lustre_client_debian() {
+    [[ -f /etc/apt/sources.list.d/fsxlustreclientrepo.list ]] || return 0
+    local new_kernel
+    new_kernel=$(find /lib/modules -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort -V | tail -n1)
```
hehe7318: This comment comes from AI, can you double check if is positive?

One thing I'd double-check here: `/lib/modules/<ver>` existing doesn't really mean that kernel is installed. The lustre module packages create those dirs themselves — I unpacked `lustre-client-modules-6.17.0-1019-aws_2.15.6-1fsx32_amd64.deb` and its whole payload is:

```
./lib/modules/6.17.0-1019-aws/updates/kernel/fs/lustre.ko
...
```

And `lustre-client-modules-aws` (which we also install a couple of lines below) is just a meta package pointing at whatever is newest in the FSx repo — on noble right now that's `Depends: lustre-client-modules-6.17.0-1019-aws`, which may well not be the kernel on the box.

So after this function runs once, `/lib/modules` can be left with a dir for a kernel that isn't installed, and on a later run (or a retry) `sort -V | tail -n1` picks *that* one. We'd then install the module for a kernel that will never boot, the real incoming kernel still has no module, and the script exits 0 — so the test goes green with FSx broken after reboot.

The RHEL version avoids this by asking the package db (`rpm -q kernel`) instead of looking at the filesystem. Could we do the same on the apt side? Something like:

```bash
new_kernel=$(dpkg-query -W -f='${Package}\n' 'linux-image-*-aws' 2>/dev/null \
    | sed 's/^linux-image-//' | grep -E '^[0-9]' | sort -V | tail -n1)
```

That keeps the two functions consistent in what "the kernel about to boot" means.

gmarciani: > /lib/modules/<ver> existing doesn't really mean that kernel is installed.

This is true, but it is not a problem. Here our goal is to determine what is the kernel that will be activated after the reboot and rebuild the fsx kernel module for it.

However, for symmetry we can determine the kernel that will be installed in a more robust way using the package manager, the same way we already do for rhel.

gmarciani: done

hehe7318 (approved):

PR #7525, opened 2026-08-04
[Test] In test_patching_cluster, reduce the risk of ICEs and facilitate troubleshooting

gmarciani:
### Description of changes
In test_patching_cluster:
1. reduce the risk of ICEs by using c5.xlarge for headnode and login nodes rather than GPU instances.
2. increase the patching timeout for the head node to 3600s (from 1800s) to avoid the risk of timeouts when small instance types are used (such as c5.xlarge).
3. print the full ImageBuilder log when the patching fails to facilitate troubleshooting.
4. support new patching flavor `minimal-capped` and `full-capped`, where the patching is capped to the maximum kernel version supported by Lustre.
5. in both develop and released test config, execute full-capped on both Ubuntu and RHEL/Rocky, and full (uncapped) on AL2023, where capping is not required.
6. print stdout/stderr of patching script on the head node both in case of success and failure, to facilitate troubleshooting.
7. Prevent false positives on Rocky8 and RHEL8 caused by lazily loaded kernel modules.

#### Notes
1. `skip-bad-url-suffix-check`: safe to skip because the check would otherwise fail for the expected FSx Lsytre repo URL

### Tests
SUCCESS

```
test-suites:
  patching:
      dimensions:
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_NOPG_ubuntu2404 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["ubuntu2404"]
          flags: ["patching:minimal-capped"]
          schedulers: ["slurm"]
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_NOPG_ubuntu2404 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["ubuntu2404"]
          flags: ["patching:full-capped"]
          schedulers: ["slurm"]
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_NOPG_rhel9 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["rhel9"]
          flags: ["patching:full-capped"]
          schedulers: ["slurm"]
        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_NOPG_rhel8 }}]
          instances: ["g4dn.8xlarge"]
          oss: ["rhel8"]
          flags: ["patching:full-capped"]
          schedulers: ["slurm"]
```

### DIFF
--- a/cloudformation/patching/ami-patching.yaml
+++ b/cloudformation/patching/ami-patching.yaml
@@ -21,12 +21,16 @@ Parameters:
   PatchFlavour:
     Description: >-
       Patching flavour passed to the patch script. 'minimal' applies only security
-      errata (smallest set); 'full' applies all available package updates.
+      errata (smallest set); 'full' applies all available package updates. The
+      '-capped' variants additionally cap the kernel to the newest version the FSx
+      Lustre client supports.
     Type: String
     Default: minimal
     AllowedValues:
       - minimal
       - full
+      - minimal-capped
+      - full-capped
 
 Resources:
 
--- a/tests/integration-tests/configs/develop.yaml
+++ b/tests/integration-tests/configs/develop.yaml
@@ -176,17 +176,17 @@ test-suites:
   patching:
     test_patching.py::test_patching_cluster:
       dimensions:
-        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_ubuntu2404 }}]
+        - regions: [{{ g4dn_8xlarge_CAPACITY_RESERVATION_2_INSTANCES_2_HOURS_NOPG_ubuntu2404 }}]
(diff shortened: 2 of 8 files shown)


hanwen-cluster:
> reduce the risk of ICEs by using c5.xlarge for headnode and login nodes rather than GPU instances.

Minor: since this commit reduces the number of tests running in parallel, ICE won't be a problem. In the past, ICE happened because we made capacity reservation for 3 instances only, while we needed many more

hanwen-cluster (dismissed):

hanwen-cluster (dismissed):

hanwen-cluster (approved):

PR #7529, opened 2026-08-06
Add LambdaFunctionTimeout parameter to pcluster api

hgreebe:
### Description of changes
* Add LambdaFunctionTimeout parameter to pcluster api
* Provides workaround to lambda function timeouts during cluster validation phase

### Tests
* Ran `test_api_infrastructure` and checked timeout was being set to 60

### DIFF
--- a/api/infrastructure/parallelcluster-api.yaml
+++ b/api/infrastructure/parallelcluster-api.yaml
@@ -111,6 +111,15 @@ Parameters:
     Default: ''
     MaxLength: 10
 
+  LambdaFunctionTimeout:
+    Description: |
+      Timeout in seconds for the ParallelCluster Lambda function.
+      Increase this value if cluster operations time out during configuration validation.
+    Type: Number
+    Default: 30
+    MinValue: 30
+    MaxValue: 900
+
 Mappings:
   ParallelCluster:
     Constants:
@@ -122,7 +131,7 @@ Mappings:
 # More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
 Globals:
   Function:
-    Timeout: 30
+    Timeout: !Ref LambdaFunctionTimeout
     Tags:
       'parallelcluster:version': !FindInMap [ParallelCluster, Constants, Version]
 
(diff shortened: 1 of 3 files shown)


api/infrastructure/parallelcluster-api.yaml
```diff
@@ -111,6 +111,15 @@ Parameters:
     Default: ''
     MaxLength: 10
 
+  LambdaFunctionTimeout:
+    Description: |
+      Timeout in seconds for the ParallelCluster Lambda function.
+      Increase this value if cluster operations time out during configuration validation.
```
gmarciani, line 117: Let's just describe what the parameter does: `Timeout in seconds for the ParallelCluster Lambda function.`

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -111,6 +111,15 @@ Parameters:
...
+    Type: Number
+    Default: 30
+    MinValue: 30
+    MaxValue: 900
```
gmarciani, line 121: Is this the max value supported by Lambda?

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -111,6 +111,15 @@ Parameters:
     Default: ''
     MaxLength: 10
 
+  LambdaFunctionTimeout:
```
gmarciani, line 114: **BLOCKER -- PRODUCT STRATEGY**
I am not fully convinced by the direction.
We are adding a new parameter (one way door) to mitigate the impact of a design defect (sequential execution of config validators) which had after all these years impacted only 2 users afaik and for which there is anyways a workaround. Also, users are not asking us an additional flexibility, which would motivate the additional parameter, they are asking us to use a timeout that works, always.

What we should probably do is:
1. (short term) assuming we do not get rid of the design defect, set a timeout that works with it, always.
2. (long term) fix the design defect (parallelize validators), set the timeout that works always, the current 30s timeout is already a reasonable timeout for an API (actually already too long).

api/infrastructure/parallelcluster-api.yaml
```diff
@@ -122,7 +131,7 @@ Mappings:
 # More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
 Globals:
   Function:
-    Timeout: 30
+    Timeout: !Ref LambdaFunctionTimeout
```
gmarciani, line 134: [minor] This change makes me thing that we are currently not injecting the expected value here. We hard wire 30, but we should instead hard wire the global constant or omit it and let the global constant take effect.

PR #7537, opened 2026-08-10
Cache getPolicy API

hgreebe:
### Description of changes
* Cache getPolicy API
* Reduce the validation time for cluster creation and update when using the ParallelCluster API
* Prevents timeout during cluster creation and update caused when validating the cluster config

### Tests
* Created cluster with API that had multiple queues with the same set of 4 additional IAM policies
  * Without cacheing: 4*8 getPolicy calls which cause a timeout
  * With cache: 4 getPolicy calls - no timeout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ CHANGELOG
 - Prevent a path traversal when downloading exported cluster and image logs, so a CloudWatch log stream name containing `../`
   segments can no longer cause files to be written outside the export directory.
 - Return a clear error instead of an unexpected fatal exception when the configured AWS profile does not exist.
+- Cache getPolicy API calls to reduce validation time of clusters when using the ParallelCluster API.
 
 **DEPRECATIONS**
 - Amazon Linux 2 is no longer supported.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -93,6 +93,7 @@ CHANGELOG
 - Prevent a path traversal when downloading exported cluster and image logs, so a CloudWatch log stream name containing `../`
   segments can no longer cause files to be written outside the export directory.
 - Return a clear error instead of an unexpected fatal exception when the configured AWS profile does not exist.
+- Cache getPolicy API calls to reduce validation time of clusters when using the ParallelCluster API.
```
gmarciani: Why ParallelClusterAPI? This caching applies to both API and CLI.
Also surface in the message that the reduction of the validation is there only when additional iam policies is used

cli/src/pcluster/aws/iam.py
```diff
@@ -18,6 +18,7 @@ class IamClient(Boto3Client):
     def __init__(self):
         super().__init__("iam")
 
+    @Cache.cached
```
gmarciani, line 21: Can be addressed on a follow up PR: you were able to reproduce the timeout wiht a very simple reproducer. Why our tests were not able to catch it? Can we cover this with our tests?

hgreebe: I reproduced it by creating a cluster with 8 queues which each had 4 additional IAM policies. We could cover this in our tests by adding all the extra queues.

gmarciani (approved):

PR #7605, opened 2026-09-03
[BuildImage] Support build-image with UpdateOs=Enabled in proxied environments

gmarciani:
### Description of changes
Support build-image with UpdateOs enabled in proxied environments.

With this fix we now execute the proxy configuration in the ImageBuilder component update_and_reboot.yaml.
The proxy config steps mimic what we already do in the dedicated cookbook recipe.

#### Alternative approaches
As an alternative approach we could invoke the cookbook recipe to setup the proxy. In this way we would have the logic to configure the proxy in a single place. However, I am not doing it here because this would break the following two rules we have always complied with:
1. update&reboot component must not depend on any cookbook artifact. This rule can be revised b/c 
2. we do not call internal recipes. this is a weak argument b/c we could anyways create a new entrypoint just to configure the proxy.

### Tests
ONGOING

```
test-suites:
  createami:
    test_createami.py::test_build_image_no_internet:
      dimensions:
        # RHEL family (rhel8, rhel9) -> Red Hat Enterprise Linux reservation platform
        - regions: [{{ g4dn_2xlarge_CAPACITY_RESERVATION_2_INSTANCES_3_HOURS_NOPG_rhel9 }}]
          instances: ["g4dn.2xlarge"]
          oss: ["rhel8", "rhel9"]
        # Non-RHEL family (alinux2023, ubuntu2204, ubuntu2404, rocky8, rocky9) -> Linux/UNIX reservation platform
        - regions: [{{ g4dn_2xlarge_CAPACITY_RESERVATION_5_INSTANCES_3_HOURS_NOPG_ubuntu2404 }}]
          instances: ["g4dn.2xlarge"]
          oss: ["alinux2023", "ubuntu2204", "ubuntu2404", "rocky8", "rocky9"]
```

### DIFF
--- a/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
+++ b/cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
@@ -95,6 +95,68 @@ phases:
                 fi
               fi
 
+      - name: CreatingJsonFile
+        action: CreateFile
+        inputs:
+          - path: /etc/parallelcluster/image_dna.json
+            content: |
+              ${CfnParamChefDnaJson}
+            overwrite: true
+
+      # Configure the proxy for isolated (no-internet) builds, matching the setup_proxy.rb cookbook recipe.
+      - name: ConfigureProxy
+        action: ExecuteBash
+        inputs:
+          commands:
+            - |
+              set -x
+              PROXY_ENV_FILE=/etc/parallelcluster/pcluster_proxy_env.sh
+              PLATFORM='{{ build.PlatformName.outputs.stdout }}'
+              REGION=${AWS::Region}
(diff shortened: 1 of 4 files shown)


cli/src/pcluster/resources/imagebuilder/update_and_reboot.yaml
```diff
@@ -137,14 +203,6 @@ phases:
               set -x
               echo ${AWS::Region}
 
-      - name: CreatingJsonFile
```
hgreebe, line 140: Why is this being removed?

gmarciani: it;s moved up because we need the dna json earlier

hgreebe (approved):

PR #7625, opened 2026-09-16
Make stack id suffix use more of the UUID to prevent naming collisions

hgreebe:
### Description of changes
* Make stack id suffix use more of the UUID to prevent naming collisions
* The stack suffix was the `time_low` of the UUID, which is the creation timestamp at the millisecond resolution, if two stacks were created at the same time, this stack suffix can be the same.
* This fix uses `time_low` as well as the `clock_seq` and `node` fields which actually distinguish two concurrently generated UUIDs 
* This suffix remains under the capped chars for role name and policy name

### Tests
* Ran `test_cluster_create` and checked that the role names were properly generated.


cloudformation/policies/parallelcluster-policies.yaml
```diff
@@ -104,16 +104,20 @@ Conditions:
...
+  # Given a stack ID arn:aws:cloudformation:REGION:ACCOUNT:stack/STACK_NAME/8131d980-7fb5-11ee-9589-0a6424944f95,
+  # the resulting StackIdSuffix is 8131d98095890a6424944f95 (time_low + clock_seq + node).
   DefaultParallelClusterIamAdminPolicy:
     Type: AWS::IAM::ManagedPolicy
     Condition: EnableIamPolicy
     Properties:
       ManagedPolicyName: !Sub
         - ${IAMRoleAndPolicyPrefix}DefaultParallelClusterIamAdminPolicy-${StackIdSuffix}
-        - { StackIdSuffix: !Select [ 0, !Split [ '-', !Select [ 2, !Split [ '/', !Ref 'AWS::StackId' ] ] ] ] }
+        - StackIdSuffix: !Join
```
gmarciani, line 116: By increasing the suffix, we are reducing the maximum space for the iam role/prefix. Is the max length of the prefix still valid?

In the tests where we configure the iam prefix, let;s make sure we are using the maximum length allowed to cover the extreme case

hgreebe: `IAMRoleAndPolicyPrefix` is max 10 chars, `RoleName` is capped at 64 chars, Managed policy name is capped at 128 chars
A suffix built from `time_low` + `clock_seq` + `node` is 24 chars, which lands at 60(10 + 25 + 1 + 24= 60) for the role name and 71 for the longest policy name, which is within the limits.

gmarciani (approved):

PR #7626, opened 2026-09-16
Test that nodes being replaced were observed in a DOWN state

hgreebe:
### Description of changes
* Fix intermittent test failures occuring due to one node being replaced and in an idle state before the other down is registered as DOWN.
* `wait_for_compute_nodes_to_be_observed_in_states` checks that node was seen in DOWN at one point during the replacement process rather than asserting that both nodes being replaced are in a DOWN state at the same time.
* This reduces the flakiness of the test caused by different timing of replacement on each node.

### Tests
* Ran test_slurm_scaling


tests/integration-tests/tests/common/hit_common.py
```diff
@@ -79,6 +79,34 @@ def wait_for_compute_nodes_states(
     )
 
 
+def wait_for_compute_nodes_to_be_observed_in_states(
+    scheduler_commands,
+    compute_nodes,
+    expected_states,
+    wait_fixed_secs=20,
```
gmarciani, line 86: How did we decide 20s?
What is a node transition is faster than 20s?
If we think we can better avoid the flakiness with a shorted wait time, then let's use a shorter wait time.

gmarciani (approved):

tests/integration-tests/tests/common/hit_common.py
hgreebe: I used the same amount of seconds used in `wait_for_compute_nodes_states`: https://github.com/aws/aws-parallelcluster/blob/842b39edf29e50bc0dbb9ce98cfa600e331e029d/tests/integration-tests/tests/common/hit_common.py#L70

PR #7627, opened 2026-09-16
Have capacity reservation creation to try multiple candidate instance types 

hgreebe:
### Description of changes
*  Generalize reservation creation to try multiple candidate instance types for any indexed placeholder group (`<region>_<group>_INSTANCE_TYPE_<n>`, where `<group>` is optional, e.g. GPU)
* Build an ordered candidate list from the instance types already generated by `_get_instance_type_parameters()` for that placeholder group, rotated to start at the requested index and wrapping around.
* Try each candidate across all candidate regions/AZs before moving to the next candidate (instance type is the outer retry dimension).
* On success, overwrite the placeholder used by the test with the instance type actually reserved, in the form the config renders:
    - GPU placeholders render the full instance type (e.g. g6.4xlarge).
    - INSTANCE_TYPE placeholders that append the size in the config (e.g. {{ ..._INSTANCE_TYPE_0 }}.xlarge) render the family only, so the appended size is stripped.
    - Non-indexed/hardcoded instance types (e.g. c5_xlarge, g5g_2xlarge) keep the previous single-type behavior and never overwrite a placeholder.

* The placeholder-overwrite approach is safe as long as each indexed placeholder is used by a single reservation-backed test and is not referenced elsewhere; tests should use distinct placeholder indices. 
* If two tests have the same placeholder (same group and index) for the instance type, the first successfully created reservation will overwrite the instance type and then both tests will compete for the same reservation.
   * Modulo is used on the index so that there is no limit to the max index that can be specified.
* Adding a new group only requires generating its numbered keys in _get_instance_type_parameters().

### Tests
* Ran the following test config but added some create reservation failures to test that the fallback value works and that it tries and creates reservations for instance types with index >0 : 
```
test-suites:
  health_checks:
    test_gpu_health_checks.py::test_cluster_with_gpu_health_checks:
      dimensions:
        - regions: [{{ US_WEST_2_GPU_INSTANCE_TYPE_0_CAPACITY_RESERVATION_4_INSTANCES_2_HOURS_NOPG_OS_X86_5__c5_xlarge_CAPACITY_RESERVATION_5_INSTANCES_2_HOURS_NOPG_OS_X86_5 }}]
          instances: [{{ US_WEST_2_GPU_INSTANCE_TYPE_0 }}]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
  basic:
      dimensions:
        - regions: [{{ US_EAST_1_INSTANCE_TYPE_0_xlarge_CAPACITY_RESERVATION_8_INSTANCES_2_HOURS_NOPG_DCV_OS_X86_1__g5g_2xlarge_CAPACITY_RESERVATION_1_INSTANCES_1_HOURS_NOPG_DCV_OS_X86_1 }}]
          instances: [{{ US_EAST_1_INSTANCE_TYPE_0 }}.xlarge]
          oss: [{{ DCV_OS_X86_1 }}]
          schedulers: ["slurm"]
```


tests/integration-tests/framework/tests_configuration/config_renderer.py
```diff
@@ -297,13 +297,13 @@ def read_config_file(config_file, print_rendered=False, config=None, args=None,
     logging.info("Parsing config file: %s", config_file)
     os_parameters = _get_os_parameters(config=config, args=args)
     instance_type_parameters = _get_instance_type_parameters()
-    rendered_config = _render_config_file(
-        config_file,
+    render_parameters = {
```
hanwen-cluster: Why is this change necessary?

tests/integration-tests/framework/tests_configuration/config_renderer.py
```diff
@@ -380,11 +387,20 @@ def _check_or_create_capacity_reservations(config_file, os_parameters, instance_
...
             random.shuffle(candidate_regions)
-            if not _create_capacity_reservations(az_for_capacity_reservation, candidate_regions, specs, var):
+            successful_specs = _create_capacity_reservations(az_for_capacity_reservation, candidate_regions, specs, var)
+            if successful_specs:
+                for parsed, successful_spec in zip(parsed_placeholders, successful_specs):
+                    if parsed:
+                        prefix, selected_index, _ = parsed
+                        placeholder_key = f"{prefix}_{selected_index}"
+                        instance_types_for_capacity_reservation[placeholder_key] = _placeholder_render_value(
+                            parsed, successful_spec[0]
```
hanwen-cluster, line 397: Minor: I should've created an class/object for spec instead of relying on the index :)

tests/integration-tests/framework/tests_configuration/config_renderer.py
```diff
@@ -297,13 +297,13 @@ def read_config_file(config_file, print_rendered=False, config=None, args=None,
...
-    rendered_config = _render_config_file(
-        config_file,
+    render_parameters = {
         **kwargs,
         **os_parameters,
         **instance_type_parameters,
         **_check_or_create_capacity_reservations(config_file, os_parameters, instance_type_parameters),
-    )
+    }
+    rendered_config = _render_config_file(config_file, **render_parameters)
```
hanwen-cluster, line 306: Why is this change necessary?

tests/integration-tests/framework/tests_configuration/config_renderer.py
```diff
@@ -714,6 +749,74 @@ def _resolve_instance_type_and_os(instance_type, instance_type_parameters, os, o
...
+        placeholder = f"{prefix}_{index}" + (f"_{size_suffix}" if size_suffix else "")
+        resolved, _ = _resolve_instance_type_and_os(placeholder, instance_type_parameters, os, os_parameters)
+        candidates.append(resolved)
+        index += 1
+
+    if not candidates:
+        resolved, _ = _resolve_instance_type_and_os(
+            instance_type_expression, instance_type_parameters, os, os_parameters
+        )
+        return [resolved]
```
hanwen-cluster: Why is this necessary?

tests/integration-tests/framework/tests_configuration/config_renderer.py
hgreebe: When the region/reservation result and instance_type_parameters share a key (the placeholder- the instance type key can be also inside the key used for the region/reservation key), Python raises `TypeError: got multiple values for keyword argument 'US_WEST_2_GPU_INSTANCE_TYPE_0'` because duplicate keyword arguments are not allowed across separate ** unpackings in one call.

tests/integration-tests/framework/tests_configuration/config_renderer.py
hgreebe: It guards the case where a variable parses as `..._INSTANCE_TYPE_<n>` but no matching generated keys exist (e.g. a placeholder whose group prefix the parser accepts, but for which `_get_instance_type_parameters() `produced no `prefix_<n>` keys in that render.), avoiding an empty candidate list that would make the reservation loop fail. If we prefer to fail loudly instead, I can raise a clear error here rather than falling back.

hgreebe: I removed it

hanwen-cluster (approved):

